PR #206, opened 2010-09-01
Ability to render single pages or posts

mreid:
Hi,

I just added a simple feature that I find useful when writing posts. The usage is:

```
jekyll --single dir/_posts/2010-09-01-testing.md

jekyll --single dir/some-page.md
```

for posts and pages respectively.

What happens is Jekyll reads in your site's configuration, layouts, etc. and then renders just the file specified to the usual output directory. I typically use this to render a preview of the post I'm working without having to render my entire site.

I made a small change to the way Post handles directories along the way. Rather than have Post control the addition of the `_posts` directory, this is now done outside by Site. I think this new behaviour makes more sense and brings it into line with how Pages are handled.

Hope you find it useful. Let me know if you need any more details.

Mark.

PS. I also wanted an excuse to try the new pull requests. Looks great!

### DIFF
--- a/test/test_post.rb
+++ b/test/test_post.rb
@@ -2,7 +2,7 @@ require File.dirname(__FILE__) + '/helper'
 
 class TestPost < Test::Unit::TestCase
   def setup_post(file)
-    Post.new(@site, source_dir, '', file)
+    Post.new(@site, source_dir, '_posts', file)
   end
 
   def do_render(post)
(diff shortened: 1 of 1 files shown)


coneybeare:
This would be very useful for automated scripts. My specific use case: I have a daily newsletter generated upon build via special plugin which does not get deployed, but which I use the code for to make the HTML email newsletter.

This newsletter is automatically generated based on Google Analytics page view numbers, so it is likely to change even without a source folder content change.

I seek to automate my newsletter generation by doing a build via cronjob that will kick off the GA fetching code, build the newsletter, grab the HTML code for it and send the email newsletters. Doing this for a single page like this PR does would be much better for this scenario because I can build it quickly without messing with anything else that could potentially cause the cronjob to fail.

Please reconsider it's need.

envygeeks:
Hello.  This ticket is being locked because it is stale and further comment here is no longer beneficial. Please create a new ticket with your bug or feature request and link back to this ticket. This will help us better manage your issue. If this is a question then please ask your question at https://talk.jekyllrb.com rather than on this ticket system.

Thanks!

Issue #176, opened 2010-06-18
How to pass a variable to the include tag


StefanLiebenberg:
How can I implement something like

{% for entry in entries %}
  {% include $entry %}
{% endfor %}

Thank you

mryellow:
> #1495 is an attempt 

Being able to do the same with a plugin response would be handy, though nested plugins might get messy.

PR #593, opened 2012-07-03
Fix for #176

c-abird:
I found it very useful eg. for including subnavigations to be able to pass a variable as argument to the include tag.


mattgreenfield:
As far as I can tell, this is now possible. As of https://github.com/jekyll/jekyll/pull/1495 you can use `{% include {{variable}} %}`

PR #670, opened 2012-11-10
Use custom classifications like tags & categories

goodtouch:
We used to classify posts with tags & categories.

This patch adds a generic way to define & use your own classifications through `site.config[:classifications]` by adding `tags` like behavior to site & posts for added classifications.

Let's say you want classify Post by referring to Project(s), Customer(s) and Layout(s) you can now simply add the following to `_config.yml` :

```
classifications: 
  - projects
  - customers
  - layouts
```

And start using those classifications like you would with `categories` or `tags` :

Add the following to a Post:

```
project: bartender
layout:
  - grid
  - responsive
```

And then use the following in any template :

```
<h2>Projects</h2>
<ul>
{% for project in site.projects %}
  <li>{{ project | first }}</li>
{% endfor %}
</ul>
```

### DIFF
--- a/features/site_data.feature
+++ b/features/site_data.feature
@@ -79,6 +79,19 @@ Feature: Site data
     Then the _site directory should exist
     And I should see "Yuengling" in "_site/index.html"
 
+  Scenario: Use the custom site.projects classification variable
+    Given I have a _posts directory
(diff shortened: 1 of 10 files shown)


parkr:
@goodtouch Please open a new PR for collections if that's cool. Just use `Document`. Before Jekyll 3, posts will be `Document`-based, so it'll all fall into place. Thanks!

PR #686, opened 2012-12-09
Add --ip option to bind to specific ip address.

pvdvreede:
Use the --ip argument to specify the particular network interface to bind to.

Useful for things like cloud9 where you need to specifiy a certain IP address.

eg.

```
jekyll --ip '127.0.0.1' --server 8080
```

or in the case of cloud9...

```
jekyll --ip $IP --server $PORT
```


bradjones1:
For anyone who might be confused, this appears now to be -H (capital H)

PR #2294, opened 2014-05-03
Add option to avoid making directories for post permalinks that don't end in .html

akalin:
The current behavior is that if a post's permalink doesn't end in .html, a directory is made with the permalink path and the post is put into index.html in that directory. This is undesirable if the web server is already set up to redirect e.g. /foo to /foo.html.

So I added the option use_post_dirs to control this behavior. If true, it does the current behavior, but if false, it simply appends .html to the permalink path.

Also modified the serve command to emulate a web server that does the /foo -> /foo.html redirection when necessary.

Let me know if there is another better way to do this!

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -4,6 +4,24 @@ module Jekyll
     class Serve < Command
 
       class << self
+        require 'webrick'
+
+        class JekyllFileHandler < WEBrick::HTTPServlet::FileHandler
+          def initialize(server, root, options={}, default=WEBrick::Config::FileHandler)
+            @default_suffix = options['default_suffix']
+            super(server, root, options, default)
+          end
+
+          def search_file(req, res, basename)
+            found = super(req, res, basename)
+            # Redirect to .html if use_post_dirs is not set.
+            if !found && @default_suffix
+              basename += @default_suffix
+              found = super(req, res, basename)
+            end
+            return found
+          end
+        end
 
         def init_with_program(prog)
           prog.command(:serve) do |c|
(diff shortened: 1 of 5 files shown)


parkr:
Can you please write a Cucumber test for this as well to ensure both work?

willnorris:
So is this behavior planned for jekyll 3.0?  Writing extensionless files on disk?  As discussed on willnorris/jekyll-clean-urls#1, I think this is a bad idea and far more limiting than taking a multiviews approach, where the file on disk contains the `.html` extension, and webrick is patched to serve that file for extensionless requests.

akalin:
I don't know about the other patches, but mine definitely writes out the
files with .html and fixes up webrick to redirect.

willnorris:
@akalin ah, okay.  I may have been looking at the wrong thing then.... my apologies.

willnorris:
So it looks like changes to WEBrick to support content negotiation like behavior landed in #3452.  However, jekyll (3.0.0-beta1) is definitely writing extensionless files to disk when the permalink setting is extensionless.  @parkr, can you confirm if this is the intended behavior?  If so, what was the point of adding the WEBrick support above?

For what it's worth, I'm now running willnorris.com on jekyll 3 with [this simple plugin](https://github.com/willnorris/willnorris.com/blob/master/src/_plugins/file_extensions.rb) to ensure that destination files always include extensions.  This has replaced my jekyll-clean-urls plugin I was using with jekyll 2.

parkr:
> However, jekyll (3.0.0-beta1) is definitely writing extensionless files to disk when the permalink setting is extensionless. @parkr, can you confirm if this is the intended behavior? If so, what was the point of adding the WEBrick support above?

I know it's not a bug, but I'm still not sure I'm happy with the design decision. People want clean URLs – extension-less URLs – and #3452 provides this. What I think we do is always write with an extension, but offer a means of changing the Liquid-exposed URLs (e.g. `page.url`) to either be clean or be fancy.

Issue #2407, opened 2014-05-14
Collections ignoring permalink settings


mattes:
I would expect that the following in my `_config.yml` ...

```
collections:
  blog:
    output: true
    permalink: pretty
  docs:
    output: true
    permalink: pretty
```

... produces:

```
$ tree
.
├── 2014
│   └── 05
│       └── 15
│           └── hello-world
│               └── index.html
```

but it seems as if it ignores the `permalink` setting and produces:

```
$ tree
.
├── 2014-05-15-hello-world.html
```

[full dir structure](https://gist.github.com/mattes/25b8b6ad693059d20975)

Am I doing sth wrong?

PR #2418, opened 2014-05-17
Override collection url template

mathbruyen:
Adds the `permalink` metadata to collections in `_config.yml`. See #2407 (and #2293). The option is used as URL template for documents inside the collection, as opposed to the previously hardcoded value.

I only used currently allowed placeholders (`collection`, `basename` and `output_ext`) but others could be easily added. I wonder whether it would make sense to allow front-matter properties there, but since it would mostly require treatment (like spaces-to-dashes) it may not be as simple as saying it.

Shortnames like `pretty` cannot be used.

### DIFF
--- a/features/collections.feature
+++ b/features/collections.feature
@@ -29,6 +29,20 @@ Feature: Collections
     And I should see "Methods metadata: bar" in "_site/collection_metadata.html"
     And I should see "<p>Whatever: foo.bar</p>" in "_site/methods/configuration.html"
 
+  Scenario: Rendered collection at a custom URL
+    Given I have an "index.html" page that contains "Collections: {{ site.collections }}"
+    And I have fixture collections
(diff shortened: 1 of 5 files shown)


davebowker:
So, my collection is for example `style-guide`. Currently the URL generated for this is www.site.com/style-guide -- I'd like it to be www.site.com/s

In the `_config.yml` I have:
```
collections:
  style_guide:
    output: true
    permalink: /s
```

Which actually generates the single HTML file, however it's unstyled. If in the `_config.yml` file I put:
```
collections:
  style_guide:
    output: true
    permalink: /s.html
```

This actually creates the file fully styled.

Basically, I want to be able to go to `site.com/s` and have everything there styled, as opposed to going to `site.com/s.html`

Any pointers? Is this a bug in how it works?

Thanks

PR #2433, opened 2014-05-21
Honor default date.

### LINKED ISSUE #2422: Date doesn't work in front matter defaults
Hi there,
I want my posts to be the same date. So I put the following in my _config.yml file:

```
defaults:
  -
    scope:
      path: ""
      type: "post"
    values:
      date: "31.05.2014"
```

But when I build the site, instead of the date specified in front matter defaults, the posts get the date specified in the names of the files initially. Although when I add the date manually to the front matter of each post, it works.

I also tried to put the date in a different format like `date: 2014-05-31`, but it still doesn't work.

So can I set the date in front matter defaults? If yes, how can I do it?

yihangho:
Fix #2422.

Not sure if I should write tests or features, but in the end I decided to write features.

### DIFF
--- a/features/frontmatter_defaults.feature
+++ b/features/frontmatter_defaults.feature
@@ -28,6 +28,24 @@ Feature: frontmatter defaults
     And I should see "<p>some special data</p><div>Ben</div>" in "_site/2013/09/11/default-data.html"
     And I should see "just some special data by Ben" in "_site/index.html"
 
+  Scenario: Use default date
+    Given I have a _posts directory
+    And I have a "_posts/2014-05-21-default-date.textile" file that contains "---\ntitle: stuff\n---"
+    And I have a configuration file with "defaults" set to "[{scope: {path: ""}, values: {date: "2000-01-01"}}]"
+    When I run jekyll build
+    Then the _site directory should exist
+    And the "_site/2000/01/01/default-date.html" file should exist
+
+  Scenario: Honor date set in front-matter
+    Given I have a _posts directory
+    And I have the following post:
+      | title | date       | content |
+      | date  | 2014-05-21 | stuff   |
+    And I have a configuration file with "defaults" set to "[{scope: {path: ""}, values: {date: "2000-01-01"}}]"
+    When I run jekyll build
+    Then the _site directory should exist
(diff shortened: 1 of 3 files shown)


iloveip:
@parkr I have one more use case for this. As my site is in Russian, I need to localize the date. If I set something like `updatedate` in front matter defaults, such date can't be localized using i18n. Localization works only with `page.date`. So I would really like to be able to set the `date` for posts in front matter defaults.

parkr:
Posts are all meant to have different date values -- the idea is to have a series of chronologically sensitive items, each being created at a different moment in time. I can't think of why a _default_ date value is what we'd want.

iloveip:
@parkr Yes, I know. And I was going to use collections instead of posts for my knowledge base articles. But collections have other limitations, as no search at the moment and no `related_posts`. So I have to go back to posts. The only problem I have with posts is the date, as I need to update it from time to time and will have to do it manually in front matter of each post. That's why I would really like to be able to do it in one place.

parkr:
> @parkr Yes, I know. And I was going to use collections instead of posts for my knowledge base articles. But collections have other limitations, as no search at the moment and no related_posts. So I have to go back to posts. The only problem I have with posts is the date, as I need to update it from time to time and will have to do it manually in front matter of each post. That's why I would really like to be able to do it in one place.

It sounds like we need to fix collections then!

iloveip:
@parkr Any idea how long it may take?

parkr:
> @parkr Any idea how long it may take?

What do you mean by `search` in your previous comment? Not sure how long it will take, this is open source, we all do it on the side. If there were a PR, it'd infinitely speed up the process.

iloveip:
@parkr By `search` I mean that there are some search plugins for Jekyll, but at the moment they support only posts, not collections. 

I understand, sorry for any trouble. You are doing great job anyway. Thank you very much!

parkr:
> @parkr By search I mean that there are some search plugins for Jekyll, but at the moment they support only posts, not collections.

Ah, ok. Unfortunately we can't do much to change plugins except to make the interface the same so they don't have to add extra support. We're working on turning everything into a `Document` so we don't have different cases for `Post`, `Page` & `Layout`.

PR #2462, opened 2014-05-29
Default frontmatter defaults for permalinks not working properly

mcallan83:
I have the need for special permalinks for posts only. I attempted this by adding the following to my _config.yml:

```yaml
defaults:
-
  scope:
    path: ""
    type: "post"
  values:
    permalink: :categories/:title
```

If I have a post with a title of "testpost" and a category defined as "blog", I had hoped this setting would cause this file to be generated at `blog/testpost/index.html`. However, Jekyll is actually generating a folder called ":category/:title/" with a file called "index.html" inside it, which isn't close to what I would expect.

I'd expect permalinks defined this way (with template variables) to behave just like permalinks defined in the front matter of a post.

I have provided a failing test case to demonstrate this.

### DIFF
--- a/features/permalink_default.feature
+++ b/features/permalink_default.feature
@@ -1,0 +1,9 @@
+Feature: postpermalink defaults
+  Scenario: Define permalink default for posts
+    Given I have a _posts directory
+    And I have the following post:
+      | title          | date       | category | content | 
+      | testpost       | 2013-10-14 | blog     | blabla  |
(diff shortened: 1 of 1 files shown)


jochemkeller:
Hey Parker,

I have resolved this for the project I worked on when I posted previously, but I am indeed still interested in more _config.yml defaults!

parkr:
Cool. This PR is just the test. It should be relatively simply to fix it. Just check out `lib/jekyll/url.rb` and make sure `permalink` is processed.

Issue #2510, opened 2014-06-13
Highlight block in C language gets comment style


denilsonsa:
The following Jekyll markdown code (using redcarpet):

``````
{% highlight c %}
foobar
{% endhighlight %}
```c
foobar
```
``````

Renders roughly the following HTML code:

``` html
<div class="highlight">
    <pre>
        <code class="c"> <!-- LOOK HERE -->
            <span class="n">foobar</span>
        </code>
    </pre>
</div>
<div class="highlight">
    <pre>
        <code class="language-c" data-lang="c"> <!-- LOOK HERE -->
            <span class="n">foobar</span>
        </code>
    </pre>
</div>
```

The problem with the first block of highlighted code is that the language class `c` is the same class used for comments:

``` css
/* From default syntax.css */
.highlight .c { color: #999988; font-style: italic } /* Comment */
```

In other words, the first block of code will be in italic, when it clearly shouldn't.

I suggest changing that [jekyll source-code](https://github.com/jekyll/jekyll/blob/c546269485cd6891ef83f5c96e735c22ec51d4d7/lib/jekyll/tags/highlight.rb#L86) to generate the same HTML as redcarpet does. Yes, I know, this is behavior-changing, so it should probably be behind an option inside `_config.yml`.

PR #2511, opened 2014-06-13
Making `highlight` behave more like redcarpet

denilsonsa:
See issue #2510 for details.

Note: redcarpet does not escape the class name (like Jekyll does).

This means that `c++` in redcarpet generates `class="language-c++"`, while in Jekyll will be `c--`. Should I also remove this `gsub("+", "-")` call, so that both generate the same code? And why is it here anyway?

### DIFF
--- a/test/test_tags.rb
+++ b/test/test_tags.rb
@@ -91,11 +91,11 @@ CONTENT
     end
 
     should "render markdown with pygments" do
-      assert_match %{<pre><code class="text">test</code></pre>}, @result
+      assert_match %{<pre><code class="language-text" data-lang="text">test</code></pre>}, @result
     end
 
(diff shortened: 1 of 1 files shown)


leobalter:
Not that having changed class names from `html` to `language-html` represented a breaking change. I understand the examples with `c` and `c++` are enough arguments for this change but it would be easier to identify this change in a major version, in case you're using a `semver` standard.

Nothing to fix after for this comment but I ask to avoid breaking retrocompatibility on minor versions next time to prevent a time investigating what happened.

parkr:
@leobalter We do our best.

leobalter:
Sure, and Jekyll is a great tool! I'm sorry if was tough.

PR #2532, opened 2014-06-19
Support for hl_lines highlighting

alexmedearis:
This allows for key/value pairs of the form key="value1 value2 value3" to be passed as arguments to highlight.

In particular, hl_lines="3 4 5" etc. is valid highlight syntax, but previously did not parse correctly.

@parkr


nasht00:
I know this is old, but is there a way to make this work with the backtick syntax?

nasht00:
Does this work on Jekyll 3.0 with rouge?

PR #2633, opened 2014-07-24
Categories in the path are ignored

redox:
When using subdirs to reflect categories, the underlying `categories` attribute was not set. I've added a new unit test and updated some others.

### DIFF
--- a/lib/jekyll/post.rb
+++ b/lib/jekyll/post.rb
@@ -160,7 +160,7 @@ module Jekyll
     def process(name)
       m, cats, date, slug, ext = *name.match(MATCHER)
       self.categories ||= []
-      self.categories += (cats || '').split('/')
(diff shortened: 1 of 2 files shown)


SimplGy:
@bcomnes, and others who find this a little unclear:

This folder structure creates a category:

    /unicorns/_posts/2015-05-01-the-prettiest-one.md

This does not create a category, and it does not change the permalink:

    /_posts/2015/2015-05-01-the-prettiest-one.md

danijar:
@bcomnes Did you find a way to use subdirs as categories without breaking existing behavior?

I would like to have quite a few categories. But there are also other top-level folders, for example holding statically hosted files. So it becomes confusing which directories is are categories and which are not.

bodokaiser:
No, not really but it has been a while since I tried

PR #2638, opened 2014-07-28
Layout data should not override page data

### LINKED ISSUE #1934: page.layout should reflect the requested layout in all circumstances
If a layout itself has a layout, `page.layout` should still return the requested layout

## Scenario

* I have a blog with pages and posts
* I have a layout called `default.html` with my header and footer
* I have layouts called `post.html` and `page.html` which themselves specify `layout: default`
* I want to know if the given thing I'm rendering is a page or a post, such as to add a class to the body tag via `{{ page.layout }}`

## Expected

The post frontmatter says `layout: post`, I asked for `layout: post`, so when I query `layout`, return `post`

## Actual

The last rendered layout is `default`, so `page.layout` will return `default` under all circomstances for both posts and pages

ivantsepp:
Fixes #1934.

I removed the code that merged the layout data with the page data (while letting the layout data override the page data). I do not think layout's should have data. But in the case that assumption is wrong, another feasible patch would be:

```diff
diff --git a/lib/jekyll/convertible.rb b/lib/jekyll/convertible.rb
index 8e9700b..f0c7998 100644
--- a/lib/jekyll/convertible.rb
+++ b/lib/jekyll/convertible.rb
@@ -166,7 +166,8 @@ module Jekyll
       used = Set.new([layout])

       while layout
-        payload = Utils.deep_merge_hashes(payload, {"content" => output, "page" => layout.data})
+        payload = Utils.deep_merge_hashes({"page" => layout.data}, payload)
+        payload = Utils.deep_merge_hashes(payload, {"content" => output})

         self.output = render_liquid(layout.content,
                                          payload,
``

### DIFF
--- a/features/rendering.feature
+++ b/features/rendering.feature
@@ -33,3 +33,17 @@ Feature: Rendering
     When I run jekyll build
     Then the _site directory should exist
     And I should see "hey = 'for cicada';" in "_site/index.js"
+
+  Scenario: Don't let layout data override page data
+    Given I have a "yo-dawg.html" page with layout "simple" that contains "Yo dawg, I heard you liked <code>_layouts</code>? So I gave all your layouts a layout."
+    And I have a default layout that contains "{{ content }}"
+    And I have a "_layouts/simple.html" file with content:
+      """
+      ---
+      layout: default
+      ---
+      {{ content }} The layout is {{ page.layout }}
+      """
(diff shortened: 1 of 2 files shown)


parkr:
I'm really worried this will break a bunch of sites. Layouts can have data, so we can't take it away now.

Issue #2371, opened 2014-05-09
Defaults precedence doesn't seem to work


feuerbach:
I'm using jekyll from recent git master (8cad40a124de7e28dc199948ecf416893458b4bc).

I have the following defaults block in `_config.yml`:

```
defaults:
  -
    scope:
      path: "_posts"
    values:
      layout: "post"
  -
    scope:
      path: ""
    values:
      layout: "page"
```

However, it looks like the `page` layout is applied to posts, too.

PR #2774, opened 2014-08-19
Find variables in _config.yaml by searching with relative file path.

dziemian007:
This basically will fix https://github.com/jekyll/jekyll/issues/2371. Please check/reproduce there.

When FrontmatterDefaults tries to find matching sets from config file, lookup by path fails if it also contains "_posts" part. So instead of matching path like "/all/pages/2014-01-01-test.md" it will try to also match "/all/pages/_posts/2014-01-01-test.md".

Without the fix, we won't be able to customize via _config.yml files not assigned to any category (located at root /_posts)


parkr:
LGTM.

PR #2870, opened 2014-09-03
Support a new `relative_include` tag

gjtorikian:
This PR aims to solve a problem [reported here](https://github.com/jekyll/jekyll-coffeescript/issues/5#issuecomment-52993378) and elsewhere.

When it comes to assets, I'm perfectly happy with the following import strategy:

```

---

---

{% include assets/javascript/jquery.min.js %}
{% include assets/javascript/foo.js %}
{% include assets/javascript/whatever.js %}
```

What bothers me, though, is that I must place my assets in the __includes_ folder. Given that SASS has its own `import` notation, and that I want to be a good citizen and use Bower for imports, it becomes impossible to keep all my assets tidy and in a single location.

With `relative_include`, I should be able to include files _relative_ to the file they're defined in. I should not be able to do `../` and go somewhere else--that's bad. Most of this checking--along with evil symlinks--is already accomplished by the current `IncludeTag`, which is very nice. 

Please let me know how you feel about this. :heart_decoration:

### DIFF
--- a/lib/jekyll/tags/include.rb
+++ b/lib/jekyll/tags/include.rb
@@ -152,7 +152,7 @@ eos
 
       def resolved_includes_dir(context)
         page_path = context.registers[:page].nil? ? includes_dir : File.dirname(context.registers[:page]["path"])
-        File.join(File.realpath(context.registers[:site].source), page_path)
+        Jekyll.sanitized_path(context.registers[:site].source, page_path)
       end
     end
   end


ryanburnette:
Are there any docs for this?

parkr:
http://jekyllrb.com/docs/templates/#including-files-relative-to-another-file

ryanburnette:
Thanks!

Issue #2750, opened 2014-08-14
Add un-rendered content to `page` variable


steelywing:
Can I have a property like `page.raw` to access the un-rendered content ?
I want to create a 1-page that include some other pages, but when I access the other pages with `page.content`, some of them are rendered, some aren't, I know the docs said

```
rendered or un-rendered depending upon what Liquid is being processed and what `page` is. 
```

PR #2901, opened 2014-09-10
Expose page.unrendered_content and page.rendered_content to Liquid

nitoyon:
## Problem

In Liquid template, `page.content` variable represents both un-rendered content and rendered content (e.g. un-rendered content = `[Jekyll](http://jekyllrb.com)`, rendered content = `<p><a href="http://jekyllrb.com">Jekyll</a></p>`).

This behaviour is documented [as follows](http://jekyllrb.com/docs/variables/): 

> The content of the Page, rendered or un-rendered depending upon what Liquid is being processed and what page is.

By further code reading, I found how to refer to (un)rendered content.

                           | un-rendered content| rendered content
---------------------------|--------------------|--------------------------------------
un-rendered page           | `page.content`     | impossible
rendered & layouting page  | `page.content`     | `content` (If the page is already wrapped by other layout file, `content` isn't equal to rendered content)
rendered & layouted page   | ----               | `page.content`

It would be useful if we could get (un)rendered content anytime.

For example, I want to embed `<og:description>` tag to my new site generated by `jekyll new` command.

* Adding `<meta property="og:description" content="{{ content | strip_newlines | strip_html | escape}}">` to `_include/head.html` isn't good because `og:description` contains site name and post date.
* add `<meta property="og:description" content="{{ page.content | strip_newlines | strip_html | escape}}">` to `_include/head.html` isn't good because `og:description` contains raw markdown document.

I need a new Liquid variable to embed rendered content.

## Solution

I exposed `page.unrenderd_content` and `page.rendered_content` variable to Liquid. Now, it becomes obvious how to use (un)rendered content.

                           | un-rendered content    | rendered content
---------------------------|------------------------|----------------------------------
un-rendered page           | page.unrenderd_content | page.rendered_content (== nil)
rendered & layouting page  | page.unrenderd_content | page.rendered_content
rendered & layouted page   | page.unrenderd_content | page.rendered_content

The behavior of `page.content` isn't changed in order to keep the backward compatibility.

### DIFF
--- a/lib/jekyll/convertible.rb
+++ b/lib/jekyll/convertible.rb
@@ -49,6 +49,10 @@ module Jekyll
           self.content = $POSTMATCH
           self.data = SafeYAML.load($1)
         end
+
+        if self.respond_to?(:unrendered_content)
+          self.unrendered_content = self.content
+        end
       rescue SyntaxError => e
         Jekyll.logger.warn "YAML Exception reading #{File.join(base, name)}: #{e.message}"
       rescue Exception => e
@@ -232,6 +236,11 @@ module Jekyll
 
       self.content = render_liquid(content, payload, info) if render_with_liquid?
       self.content = transform
+
+      # tell rendered_content to layouts
+      if self.respond_to?(:rendered_content)
+        payload['page']['rendered_content'] = self.rendered_content = content
+      end
 
       # output keeps track of what will finally be written
       self.output = content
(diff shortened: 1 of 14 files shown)


parkr:
Let's revisit this when `Post` and `Page` are `Document` subclasses.

eloyesp:
#2750 implemented in #2759 fix this, but was never merged. I've also need this :(

PR #2982, opened 2014-10-08
Tweak page description

XhmikosR:
1. remove unneeded end tags
2. add `strip_newlines` in description (without this, I ended up with newlines inside the quotes)
3. move `X-UA-Compatible` right after `meta charset`

### DIFF
--- a/lib/site_template/feed.xml
+++ b/lib/site_template/feed.xml
@@ -7,7 +7,7 @@ layout: null
     <title>{{ site.title | xml_escape }}</title>
     <description>{{ site.description | xml_escape }}</description>
     <link>{{ site.url }}{{ site.baseurl }}/</link>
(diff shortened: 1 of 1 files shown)


amw:
`strip_newlines` removes the "\n" characters instead of replacing them with spaces. This creates word run-ins in markdown content. Source:

``` markdown
This is an example paragraph
that is broken into a few
lines. New lines render as
spaces in HTML.
```

This is rendered as `This is an example paragraphthat is broken into a fewlines. New lines render asspaces in HTML` in the meta tag.

I reproduced this on a brand new jekyll site by adding this as the first paragraph to the "welcome-to-jekyll.markdown" post.

I think the solution is to either keep the newlines inside the description meta tag or replace it with a custom liquid filter like this one:

``` ruby
  def newlines_to_spaces input
    input.to_s.gsub(/\r?\n/, ' '.freeze)
  end
```

amw:
I've opened a new issue since this PR has been closed for so long.

PR #3106, opened 2014-11-14
Statically serve php files with 'jekyll serve'

shuber2:
In order to integrate a "upcoming events" list to the frontpage of our website, we employ some webdav php code in index.php. However, "jekyll serve" does not ship php files for two reasons:
- The mime type for php is not set (to text/html), so index.php would be downloaded by the browser rather than displayed.
- index.php is not in the DocumentIndex when launching webrick.

This pull request fixes this. Note that by these little changes webrick still does not run a php interpreter, however, for most authoring tasks (adding static content, adapting layout, etc) it is fine if the php code is displayed rather than not serving php files at all. So this is a major improvement for our workflow.

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -77,7 +77,7 @@ module Jekyll
         def webrick_options(config)
           opts = {
             :BindAddress        => config['host'],
-            :DirectoryIndex     => %w(index.html index.htm index.cgi index.rhtml index.xml),
+            :DirectoryIndex     => %w(index.html index.htm index.cgi index.rhtml index.xml index.php),
             :DocumentRoot       => config['destination'],
             :DoNotReverseLookup => true,
             :MimeTypes          => mime_types,


cashewguy:
I *need* the option to generate my files into those that end with .php. I run `jekyll serve` and then sync _site to my server, which *can* run PHP. This doesn't conflict in any way with the "vision" of Jekyll, so I'm not sure why I can't serve out. .php documents.

Is there any way for me to do this yet?

parkr:
@cashewguy Nope, this was not accepted. We want to match GitHub Pages as closely as possible and we also don't want to be in the business of serving non-static content as static content. Accepting this patch would not render the PHP, and people would get confused about that. Check out something like MAMP or WAMP (or setup your own nginx/Apache instance) and generate your site into a directory that your php-enabled server knows about. Jekyll can definitely still generate the PHP files without a problem, but the built-in server cannot and will not handle the PHP rendering process. It's out of the scope of this project.

cashewguy:
Actually it won't. You either can't have Jekyll serve up .php files, or there isn't a documented way to do so. You can try, by putting `permalink: title.php`, but this just makes a directory `title.php/` with `index.html` inside of it. I can find no information on having jekyll generate `index.php` files - which makes PHP inside pretty useless.

Also, PHP even inside of `{% raw %}{% endraw %}` gets converted into HTML comments. Example:

Both `<?php phpinfo(); ?>` and `{% raw %}<?php phpinfo(); ?>{% endraw %}` generate this: `<!--?php phpinfo(); ?—>`

This doesn't have anything to do with rendering or executing the PHP - I don't want Jekyll to do that. I want it to be able to output .php files and leave the PHP inside untampered with so I can build a basic contact form. I already have my development setup to generate into the MAMP htdocs directory.

I certainly hope I'm wrong, but I can't find any information on this (other than this thread), and the Jekyll IRC channel equally had no help.

parkr:
Weird. Do you have a sample site you can share? If you have `mypage.php` with your php/html/liquid, it should just push out a valid php file because we have no converters for php. You can't use Markdown to generate PHP, just an FYI.

cashewguy:
No, it's all local at the moment, but you may have solved my problem. I didn't think to make a standalone page because the form is inside of a post. I was putting the PHP at the bottom of a .markdown document, expecting it to carry over uninterrupted/unchanged.

I'll try setting it up as an independent page. Danke,

parkr:
Bitte schön! :)

jdennes:
> Would you mind writing into support@github.com and asking if this can be changed on their end? I'd be happy to change it if GitHub will as well.

I've just deployed this change:

```
› curl -I http://pickardayune.com/misc/example.php
HTTP/1.1 200 OK
Server: GitHub.com
Content-Type: text/plain; charset=utf-8
...
```

parkr:
Thank you so much, @jdennes!

shuber2:
That is awesome! As you can see, I modified the original pull request to match text/plain. Thank you very much, that change really makes my life easier.

PR #3134, opened 2014-11-20
Allow collections YAML to end with dots

davidized:
Posts can have YAML front matter that ends with three dots instead of dashes since 52ac2b3 but that only applies to posts. This PR adds the ability for the YAML front matter of collections files to end with dots as well.

I don't have very much experience in writing ruby code and tests, but I managed to write one. It probably can be improved but I just don't have enough experience to know how to do that yet.

### DIFF
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -210,7 +210,7 @@ module Jekyll
             @data = defaults
           end
           @content = File.read(path, merged_file_read_opts(opts))
-          if content =~ /\A(---\s*\n.*?\n?)^(---\s*$\n?)/m
+          if content =~ /\A(---\s*\n.*?\n?)^((---|\.\.\.)\s*$\n?)/m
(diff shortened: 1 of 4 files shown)


parkr:
Yeah I changed my mind. I don't care about supporting dots.

lib/jekyll/document.rb
```diff
@@ -210,7 +210,7 @@ def read(opts = {})
             @data = defaults
           end
           @content = File.read(path, merged_file_read_opts(opts))
-          if content =~ /\A(---\s*\n.*?\n?)^(---\s*$\n?)/m
+          if content =~ /\A(---\s*\n.*?\n?)^((---|\.\.\.)\s*$\n?)/m
```
parkr: Let's put this in some const somewhere. It's used in `site.rb` as well.

parkr:
This'll need a rebase.

PR #3171, opened 2014-11-29
Add a Jekyll doctor warning for URLs that only differ by case

### LINKED ISSUE #3035: jekyll doctor should print a warning if two permalinks only differ by case
(This is a continuation from #3034)

If I'm on a case-insensitive file system and create two pages with permalinks that only differ by case, the site generation overwrites one of the pages. This should result in a `jekyll doctor` warning.

akoeplinger:
Those URLs are problematic on case-insensitive file systems because one of the URLs is overwritten by the other. Fixes #3035.

Note: I can't seem to find tests for doctor, I'm assuming there are none?

### DIFF
--- a/lib/jekyll/commands/doctor.rb
+++ b/lib/jekyll/commands/doctor.rb
@@ -66,9 +66,7 @@ module Jekyll
 
         def urls_only_differ_by_case(site)
           urls_only_differ_by_case = false
-          urls = {}
-          urls = collect_urls_case_insensitive(urls, site.pages, site.dest)
-          urls = collect_urls_case_insensitive(urls, site.posts, site.dest)
+          urls = case_insensitive_urls(site.pages + site.posts, site.dest)
           urls.each do |case_insensitive_url, real_urls|
             if real_urls.uniq.size > 1
               urls_only_differ_by_case = true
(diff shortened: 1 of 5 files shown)


mattr-:
domain names are case sensitive according to RFC 4353. The rest is up to how the server wants to handle GET requests, but I'd rather eliminate the possibility of any duplicates due to case-sensitivity, so :+1: on this.

@akoeplinger It looks like there are some missing tests to add. Would you mind taking care of those so we can merge this?

parkr:
> @akoeplinger It looks like there are some missing tests to add. Would you mind taking care of those so we can merge this?

:+1:

parkr:
Still interested in seeing this @akoeplinger?

akoeplinger:
@parkr ahh sorry guys, I totally missed the notifications here and was busy with other stuff. I'll add the tests tomorrow :)

parkr:
@akoeplinger Any updates?

parkr:
Merged in fdcd7613136fe39d1ccbda5b135e3eb96ccdb42c.

akoeplinger:
@parkr omg, no idea how this fell off my radar, sorry for the lack of updates :sob: Thanks for merging!

PR #3175, opened 2014-12-01
Read in _config_#{JEKYLL_ENV}.yml configuration file

### LINKED ISSUE #3173: Automatically read _config_{JEKYLL_ENV}.yml, too
Right now, we look for a _config.yml file, which represents user-specified configuration of the Jekyll site. It'd make the development process a bit easier if there were override configurations which were read in for each JEKYLL_ENV. This would allow you to, for example, specify your baseurl as `""` in your base _config.yml, then GitHub Pages would automatically read in _config_production.yml or _config_pages.yml, which could contain the proper override. It could be tested by setting the JEKYLL_ENV environment variable locally.

Thoughts, @jekyll/gh-pages, @balevine, @jekyll/core?

parkr:
For env-specific configuration tweaks.
Closes #3173.

- [x] Code
- [x] Tests
- [ ] Docs

### DIFF
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -85,7 +85,7 @@ module Jekyll
     # images and allows you to skip that when working in development.
 
     def env
-      ENV["JEKYLL_ENV"] || "development"
+      ENV["JEKYLL_ENV"] || "local"
     end
 
     # Public: Generate a Jekyll configuration Hash by merging the default
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -85,7 +85,7 @@ module Jekyll
     # images and allows you to skip that when working in development.
 
     def env
-      ENV["JEKYLL_ENV"] || "local"
+      ENV["JEKYLL_ENV"] || "dev"
     end
 
     # Public: Generate a Jekyll configuration Hash by merging the default
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -26,7 +26,6 @@ require 'set'
 # 3rd party
 require 'safe_yaml/load'
 require 'liquid'
-require 'kramdown'
 require 'colorator'
 
 SafeYAML::OPTIONS[:suppress_warnings] = true
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -153,7 +153,7 @@ module Jekyll
     #
     # Returns the filename for the JEKYLL_ENV-specific configuration file.
     def env_config_file
-      "_config_#{Jekyll.env.downcase}.yml"
+      "_config.#{Jekyll.env.downcase}.yml"
     end
 
     # Public: Read configuration and return merged Hash
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -92,11 +92,6 @@ module Jekyll
       override[config_key] || self[config_key] || DEFAULTS[config_key]
     end
 
-    # Public: Directory of the Jekyll source folder
-    #
-    # override - the command-line options hash
-    #
-    # Returns the path to the Jekyll source directory
     def source(override)
       get_config_value_with_override('source', override)
     end
@@ -110,6 +105,10 @@ module Jekyll
       get_config_value_with_override('verbose', override)
     end
     alias_method :verbose?, :verbose
+
+    def in_source(override, path)
+      Jekyll.sanitized_path(source(override), path)
+    end
 
     def safe_load_file(filename)
       case File.extname(filename)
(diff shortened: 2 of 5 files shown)


lib/jekyll.rb
```diff
@@ -78,7 +78,7 @@ class << self
     # images and allows you to skip that when working in development.
 
     def env
-      ENV["JEKYLL_ENV"] || "development"
+      ENV["JEKYLL_ENV"] || "local"
```
mattr-: I dislike this particular change. `local` doesn't have as heavy of a semantic meaning as `development` does. `development` is also seen in other places, such as `RACK_ENV` and `RAILS_ENV` and I think keeping that consistency is important.

lib/jekyll/configuration.rb
```diff
@@ -113,6 +113,10 @@ def source(override)
       override['source'] || self['source'] || DEFAULTS['source']
     end
 
+    def in_source(override, path)
```
mattr-: This method feels weird here. Don't we have a `Utils` namespace that this could go to instead?

mattr-:
I'm all for sites having multiple configuration files, but I'm not sure we should autoload any of them other than the default `_config.yml` file.

If we do decide to add this in, then this pull request is missing documentation.

lib/jekyll.rb
parkr: I thought about that, too. I think many of our users have never used Rails and they might not even know what Rack is. They think "here's my local machine and here are the override settings for my computer" and "there's github pages and here are the override settings for them". The idea of `development`, `test`, `production`, etc are all foreign. I think `local` is a more approachable term (and it has fewer characters!) for newcomers. I agree that it's a divergence from Rails and Rack, but my gamble is that the semantic benefit of `local` will help. So a site might have:

```text
.
| 
| - _config.yml
| - _config.local.yml
| - _config.github.yml
```

And that'd be it.

mattr-: I used examples from the Ruby world, but `development` and `production` are two terms that I would expect to be nearly ubiquitous amongst our userbase, which was the original point that I made poorly before.

parkr: I think I have a different feel for the user base, I guess. I see many more front end developers who have no concept of deployment environments. They know what is building their site (their local computer, github, or some other service), but `development` and `production` would be foreign ideas to them.

parkr: I could be totally wrong. //cc @benbalter for sanity.

benbalter: To me, it's a question of who you want to optimize for. "local" may be more intuitive for non-developer users, but for developers, "local" may be just as unintuitive as "development" is for non-developer users. I'm not strongly opinionated on the matter, but generally suggest not reinventing the wheel wherever possible, even if the wheel isn't 100% ideal.

parkr: Ok, let's default to `dev`.

parkr:
@gjtorikian @benbalter Any thoughts?

Ready to go on this one.

gjtorikian:
> If we do decide to add this in, then this pull request is missing documentation.

I agree with that sentiment.

I also would prefer "development" as the full string to match other projects/Rack,; for example Sinatra and Rails use `development?` as a check, not `dev`. That's my personal preference, though.

benbalter:
> If we do decide to add this in, then this pull request is missing documentation.

What do you mean by that?

parkr:
> I also would prefer "development" as the full string to match other projects/Rack,; for example Sinatra and Rails use development? as a check, not dev. That's my personal preference, though.

It's just so that you have `_config.dev.yml` instead of `_config.development.yml` which is the longest filename evar.

> If we do decide to add this in, then this pull request is missing documentation.

Yeah, this needs a `site/` update.

gjtorikian:
> It's just so that you have _config.dev.yml instead of _config.development.yml which is the longest filename evar.

Yeah I get that. I'm trying to think of scenarios where, for example, [Jekyll is crammed into a Rails site](https://blog.engineyard.com/2012/introducing-bloggy-a-simple-way-to-add-a-jekyll-blog-to-any-rails-application) and you want to be able to use the env name without having to resort to some `if` statements to get the right filename string.

But like I said: it's just my preference.

troyswanson:
@parkr This is a really cool idea. Thanks for this.

parkr:
> @parkr This is a really cool idea. Thanks for this.

Thanks!



> > It's just so that you have _config.dev.yml instead of _config.development.yml which is the longest filename evar.
>
> Yeah I get that. I'm trying to think of scenarios where, for example, Jekyll is crammed into a Rails site and you want to be able to use the env name without having to resort to some if statements to get the right filename string.
> 
> But like I said: it's just my preference.

I guess most people have just done `.dev` `.deploy` `.prod` or whatever instead of writing out the whole word. I don't see any reason not to use the whole word except for aesthetics so that's not really a good reason... I'll update my PR to read in `_config.development.yml` by default if it exists. :)

envygeeks:
TBH I wouldn't mind if it was supported but not actively mentioned because it does make more sense to use the whole word.  I think the only one I've ever shortened was "development" to "dev" but never anywhere external, it was mostly an internal thing for width.

PR #3189, opened 2014-12-05
Added basic microdata

robertoschiavone:


tkrotoff:
IMHO The HTML code with the microdata stuffs looks ugly.

Is there any added value to have `itemscope itemtype="http://schema.org/BlogPosting"`, `itemprop="name headline"`, `itemprop="datePublished"`...?
From what I have heard it does not help for SEO and the format might change in the future.

rodaine:
> From what I have heard it does not help for SEO and the format might change in the future.

Hm, where did you hear that? 

HTML alone has no context of what it contains, so annotating it with this information helps non-humans interpret what is on the page. At the surface level, I'd imagine the "SEO value" would be the improved [rich snippets](https://support.google.com/webmasters/answer/99170); personally, I like to see the date published on a news or blog article in the SRP so I know if it's still relevant. It also apparently [bumps up your rank on the SRP](http://searchengineland.com/from-microdata-schema-to-rich-snippets-markup-for-the-advanced-seo-162902), too.

As for format change, the spec is living just like HTML, et al. I'd be more concerned if it was stagnant and didn't evolve over time. 

Aesthetically, I will have to agree it's clunky and messy. However, the HTML is for the computer to read, just as minified CSS and uglified JS are as well.

tkrotoff:
> I will have to agree it's clunky and messy

Thanks for the links, I was not aware you could do all this with microdata. Kind of a necessary evil.

PR #3204, opened 2014-12-11
Sort static files just once.

parkr:
Ref: #2075.

### DIFF
--- a/lib/jekyll/renderer.rb
+++ b/lib/jekyll/renderer.rb
@@ -3,12 +3,12 @@
 module Jekyll
   class Renderer
 
-    attr_reader :document, :site, :payload
+    attr_reader :document, :site, :site_payload
 
-    def initialize(site, document, payload = nil)
-      @site     = site
-      @document = document
-      @payload  = payload
+    def initialize(site, document, site_payload = nil)
+      @site         = site
+      @document     = document
+      @site_payload = site_payload
     end
 
     # Determine which converters to use based on this document's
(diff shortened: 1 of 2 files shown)


parkr:
> It looks like YAML keys with digits in them are no longer allowed.

This could have to do with the strict parser. If you search for `strict` in lib/jekyll.rb and remove that line, does it work?

> Liquid seems to be parsing vendor/JS files, too

If these files are `include_relative`'d anywhere, then this'll happen as a result of the rendering process. :/

mattr-:
> > Liquid seems to be parsing vendor/JS files, too

> If these files are include_relative'd anywhere, then this'll happen as a result of the rendering process. :/

That would mean that we'd need to work around this change in behavior before releasing Liquid 3 in a Jekyll 2.x version, correct?

parkr:
> That would mean that we'd need to work around this change in behavior before releasing Liquid 3 in a Jekyll 2.x version, correct?

We'd have to have like `{% include_relative blah.js norender %}` – I think we manually render in the `IncludeTag` but not sure off the top of my head. I don't know if it affects Jekyll 2 – this is a Jekyll 3 PR.

gjtorikian:
> This could have to do with the strict parser. If you search for strict in lib/jekyll.rb and remove that line, does it work?

Yeeeeeep, that was it; turning `strict` off made it happy. Whatever, though, I'm fine with it not accepting numbers.

> We'd have to have like {% include_relative blah.js norender %}

*shrug* https://github.com/janl/mustache.js/pull/418 

In the meantime, it's just one character.

PR #3239, opened 2014-12-28
Only add 'index.html' if the path ends in a '/'

parkr:
Perhaps dealing with https://github.com/jekyll/jekyll/pull/2294

- [x] Implement
- [ ] Tests
- [ ] Docs

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -72,6 +72,15 @@ module Jekyll
               end
             end
           end
+
+          WEBrick::HTTPUtils.module_eval do
+            def mime_type(filename, mime_tab)
+              suffix1 = (/\.(\w+)$/ =~ filename && $1.downcase)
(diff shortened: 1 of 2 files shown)


mattr-:
I'm a bit out of the loop, but wouldn't this be a breaking change that we would need to leave out until 3.0? I guess it really depends on what we consider to be part of our API that would fall under semantic versioning. ¯\\\_(ツ)\_/¯

parkr:
> I'm a bit out of the loop, but wouldn't this be a breaking change that we would need to leave out until 3.0? I guess it really depends on what we consider to be part of our API that would fall under semantic versioning. ¯\_(ツ)_/¯

The `master` branch is headed to 3.0 now. 2.5.3 was the last 2.x release: http://jekyllrb.com/news/2014/12/22/jekyll-2-5-3-released/ We have a https://github.com/jekyll/jekyll/tree/stable-2-5 for maintenance.

mattr-:
ah, ok. then :+1: on this from me, but there are no documentation updates and no tests. :sob:

Issue #3202, opened 2014-12-11
Cache includes values by path


parkr:
Once an include has been read in, cache the value. This offers performance benefits by reducing I/O. The cache should be cleared when the file has changed, however.

PR #3243, opened 2014-12-29
Add a cache for the includes.

parkr:
https://github.com/jekyll/jekyll/issues/3202

Doesn't seem to make anything faster.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,6 +23,7 @@ gem 'minitest' if RUBY_PLATFORM =~ /cygwin/
 gem 'test-unit' if RUBY_PLATFORM =~ /cygwin/ || RUBY_VERSION.start_with?("2.2")
 
 if ENV['BENCHMARK']
+  gem 'benchmark-ips'
   gem 'rbtrace'
   gem 'stackprof'
 end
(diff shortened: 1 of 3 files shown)


mattr-:
If it doesn't make anything faster, then we don't need it. :smiley:

parkr:
It could just be that we are

mattr-:
Our tests aren't good benchmarking tools and if you were only looking at those, then I wouldn't expect to see any difference. :smiley:

gjtorikian:
I'm also not seeing a huge improvement from this branch over `master`. 

If it's not hurting anything, though, I'd still say merge it in. Maybe someone, somewhere needs it?

parkr:
Not impressive, but there's an improvement:

```text
              cached       634 i/100ms
            uncached       543 i/100ms
-------------------------------------------------
              cached     6570.6 (±3.0%) i/s -      32968 in   5.022063s
            uncached     5613.0 (±2.1%) i/s -      28236 in   5.032667s
```

parkr:
Just another thing we'd have to reset in a `--watch` cycle. For a marginal benefit, not much use. :/

PR #3245, opened 2014-12-29
This is probably a bad idea: optimize with C

parkr:
WIP. This is just boilerplate code.


mattr-:
I'm ambivalent about this. I know that this is probably a hot spot in our code base, but I'm not sure it's worth the tradeoff to write this in C, unless you have some measurements that really make a compelling case. :smiley:

parkr:
I was looking at the stackprof and the inefficiencies of `gsub` and was hoping to alleviate those worries. I want performance, dammit! :stuck_out_tongue_winking_eye:

jaybe-jekyll:
More speed is always good. 😃 However, the incremental regeneration experience has been positive and alleviates some of the draggy build feelings.

flyinprogrammer:
:scream: dat jruby [insert other interpreters] support :scream: would go out the window with this type of stuff.

parkr:
> dat jruby [insert other interpreters] support  would go out the window with this type of stuff.

We already have C extensions in our dependencies – I don't think JRuby runs Jekyll at the moment, just MRI. I think speed enhancement is worth it if this is the case.

PR #3255, opened 2015-01-04
Switch to shields.io for the README badges.

XhmikosR:
This way we have consistent badges.

Note that I didn't use the flat badges, which personally, I prefer, since shields.io doesn't support [hakiri (yet)](https://github.com/badges/shields/issues/270).


parkr:
Are these faster/smaller?

XhmikosR:
There should be no difference in size. At the moment they are consistent at least, plus we can use flat style etc.

PR #3256, opened 2015-01-04
Docs: Assorted accessibility tweaks.

XhmikosR:
jekkylrb.com:
jekyllrb-com

this branch:
access-branch

### DIFF
--- a/site/_includes/anchor_links.html
+++ b/site/_includes/anchor_links.html
@@ -3,7 +3,8 @@
     var anchor = document.createElement("a");
     anchor.className = "header-link";
     anchor.href      = "#" + id;
-    anchor.innerHTML = "<i class=\"fa fa-link\"></i>";
+    anchor.innerHTML = "<span class=\"sr-only\">Permalink</span><i class=\"fa fa-link\"></i>";
+    anchor.title = "Permalink";
     return anchor;
   };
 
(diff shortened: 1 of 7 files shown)


parkr:
I'm pretty happy with this. Not super excited about the darker background color for the footer and the code block but I'm open to it.

@cobyism Any thoughts here?

XhmikosR:
You can try http://squizlabs.github.io/HTML_CodeSniffer/ yourself.

parkr:
That's a cool tool! I'll have to look up why the contrast ratio is expected at that particular level.

cobyism:
This all seems fine to me. I think the darker background for the terminal window is a good move, however I’m not sure the other colour changes are really that much of a win. I don’t have any problems with them though, so :+1:

XhmikosR:
In case it's not clear, I tried to bump the colors enough so that they pass WCAG2AA, nothing more :)

XhmikosR:
As for the empty `span`, personally I'd rather add a new `sr-only` class and make use of it everywhere for the screen readers. Let me know if you agree with that and I'll make the change adapting the class from Bootstrap.

cobyism:
> I'd rather add a new sr-only class and make use of it everywhere for the screen readers.

Sounds good, yeah. Would another option be to add a `title` attribute or something similar to the icon element?

XhmikosR:
Not sure about title in this case, since the icon is just a helper IMO. I'll add the `sr-only` class and let you know.

XhmikosR:
Done. I added a `title` in the `a` element too.

Issue #2773, opened 2014-08-19
Symlink to /tmp causes Dropbox never to finish syncing


sunpig:
Don't know if this is just my setup - perfectly content to have this closed as weird edge case. You could also consider it a problem with Dropbox, not Jekyll.

OSX Mavericks 10.9.4, Dropbox 2.10.28

I cloned the Jekyll project into a directory inside my Dropbox. Down in the test directory, the Jekyll project contains a symlink to the `/tmp` directory (https://github.com/jekyll/jekyll/blob/master/test/source/_includes/tmp). Now Dropbox tries to sync everything in my local `/tmp` directory to the cloud. It doesn't have permissions on various items in there, but it doesn't give up! It keeps on retrying, which makes it look like it is stuck permanently in sync mode.

PR #3258, opened 2015-01-04
Symlink local tmp

### LINKED ISSUE #2831: test/source/_includes/tmp is a symlink to /tmp
https://github.com/jekyll/jekyll/blob/c8b22a19adee0d2dfe3418349b314c35dd8c1212/test/source/_includes/tmp

There has to be some way around symlinking to /tmp. It causes all sorts of permissions errors in certain environments and does not work at all with dropbox (dropbox tries to sync the whole /tmp) folder. Is there a more elegant solution?

omegahm:
Fixes #2831 and possibly #2773

The two tests regarding symlinks will use `tmp` in the root-dir instead of `/tmp`.


parkr:
Cool! We alread have a tmp directory so I'd rather use that. Kill the . and we're in business. Thanks!

omegahm:
@parkr I think we might need to have the `tmp` folder in Git with a `.keep` file. What's your take on that?

parkr:
> @parkr I think we might need to have the tmp folder in Git with a .keep file. What's your take on that?

I'd rather the `script/test` create the `tmp` directory than track it through Git.

omegahm:
@parkr Yup, I can see that that is actually also what is failing on Travis. :dizzy_face:

I have added `mkdir tmp` to `script/cibuild`.

XhmikosR:
This will fix 2 of the failing tests on Windows too hopefully.

parkr:
Great work. :shipit:

PR #3254, opened 2015-01-04
Update rake task site:publish to fix minor bugs.

flyinprogrammer:
*  Automate creating gh-pages directory
*  Ensure gh-pages dir always has correct branch checked out
*  Purge gh-pages of files so that a sync of site truly occurs
*  Ensure dot files are synced
*  Be exact with our exclusions to eliminate guess work

PR #3261, opened 2015-01-05
Use FileList instead of Dir.glob

flyinprogrammer:
@penibelst mentioned maybe using Rake::FileList in https://github.com/jekyll/jekyll/pull/3254

I think that's a smart move - plus any chance to use splat arguments is fun!


parkr:
Slick stuff. Thanks!

penibelst:
Nice.

PR #3264, opened 2015-01-07
Fix nav items alignment when on multiple rows

tkrotoff:
With the default theme, when you have multiple pages and thus "nav items", they span over multiple rows.
Problem: the nav items are misaligned

margin-left

Solution: use margin-right instead of margin-left

margin-right

### DIFF
--- a/lib/site_template/_sass/_layout.scss
+++ b/lib/site_template/_sass/_layout.scss
@@ -35,9 +35,9 @@
         color: $text-color;
         line-height: $base-line-height;
 
-        // Gaps between nav items, but not on the first one
-        &:not(:first-child) {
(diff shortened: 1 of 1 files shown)


kleinfreund:
:+1:

tkrotoff:
Question: is there a way to test the default template/theme using my local Jekyll git repo?
For this pull request I have manually tested by performing a `jekyll new test1` which is a bit cumbersome

parkr:
:+1:

tkrotoff:
Ooops I have introduced a bug with this commit when the menu is in "burger" mode:

"burger" mode bug

See pull request #3329 for the fix

Issue #223, opened 2010-10-22
add kramdown support


thesurlydev:
The following branch adds support for kramdown: http://kramdown.rubyforge.org/index.html

http://github.com/digitalsanctum/jekyll/tree/add_kramdown_support

Incentive for doing so from the website: "It is probably the fastest pure-Ruby Markdown converter available (June 2010), being about 4x faster than Maruku and about 9x faster than BlueFeather."

PR #3271, opened 2015-01-08
Change collection documentation based on #223

maban:
Clarifying what happens if no YAML front matter exists. The current explanation says that Jekyll will generate a file without Front Matter, but it appears this isn't the case.

I'm still not sure if this is the intended behaviour, but this clarifies how it currently works.

### DIFF
--- a/site/_docs/collections.md
+++ b/site/_docs/collections.md
@@ -37,7 +37,7 @@ collections:
 ### Step 2: Add your content
 
 Create a corresponding folder (e.g. `<source>/_my_collection`) and add documents.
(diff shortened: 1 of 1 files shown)


parkr:
Seems like the documentation described intended behavior.

kleinfreund:
@parkr The bit that files need front matter to be generated into the collection is one that can be easily missed, I think. The proposed change would clarify that.

parkr:
I think the implications are around binary static files.

parkr:
Let's fix the documentation for now and fix the code later.

PR #3274, opened 2015-01-09
Added per post excerpt_separator functionality

majioa:
Added per post excerpt_separator functionality, so you are able to specify :excerpt_separator (as well as just :excerpt) key direct inside the post YAML, to make an excerpt based on the value in the post. Tests
were also added.


parkr:
Nice work. I'm curious though: what's the use case for this?

majioa:
@parkr  I have a lot of documents in which I use the default separator, and some other docs, in which I use the specific one, but for that cases I don't want to specify the excerpt explicitly, just want to reuse the heap of those documents.

lib/jekyll/post.rb
```diff
@@ -118,6 +119,13 @@ def title
       data.fetch('title') { titleized_slug }
     end
 
+    # Public: the Post excerpt_separator, from the YAML Front-Matter or empty string
+    #
+    # Returns the post excerpt_separator
+    def excerpt_separator
+      data.fetch('excerpt_separator') { "" }
```
parkr: This should fall back on the site excerpt_separator and probably also on the front matter defaults.

parkr:
Any thoughts, @jekyll/core?

lib/jekyll/post.rb
majioa: @parkr do you mean instead of returning `""` it should return site `excerpt_separator`

parkr: I mean instead of just returning `""` straight away, it should check the defaults and the site first.

lib/jekyll/excerpt.rb
```diff
@@ -105,7 +105,9 @@ def inspect
     #
     # Returns excerpt String
     def extract_excerpt(post_content)
-      separator     = site.config['excerpt_separator']
+      separator     = !post.excerpt_separator.empty? &&
+                       post.excerpt_separator ||
+                       site.config['excerpt_separator']
```
alfredxing: The alignment here seems a bit weird: since the `!` only applies to `post.excerpt_separator.empty?`, the other two probably shouldn't be indented past the `!` (in other words, maybe unindent the above 2 lines by 1 space).

lib/jekyll/post.rb
alfredxing: @parkr The fallback is above in `excerpt.rb`, though I agree that the fallback should be here instead of there.

lib/jekyll/post.rb
```diff
@@ -307,7 +315,7 @@ def extract_excerpt
     end
 
     def generate_excerpt?
-      !(site.config['excerpt_separator'].to_s.empty?)
+      !site.config['excerpt_separator'].to_s.empty? || !excerpt_separator.empty?
```
alfredxing: If we move the fallback into `Post::excerpt_separator` (see above comments), checking `site.config` won't be necessary here.

alfredxing:
Added a few comments in the code. Nice job on this!

lib/jekyll/post.rb
majioa: @alfredxing @parkr   I've just made some changes to get separator proc.

lib/jekyll/post.rb
```diff
@@ -118,6 +119,14 @@ def title
       data.fetch('title') { titleized_slug }
     end
 
+    # Public: the Post excerpt_separator, from the YAML Front-Matter or site default
+    #         excerpt_separator value
+    #
+    # Returns the post excerpt_separator
+    def excerpt_separator
+      data.fetch('excerpt_separator') { site.config['excerpt_separator'].to_s }
```
alfredxing: I think it would seem a bit simpler to write `data.fetch('excerpt_separator', site.config['excerpt_separator'].to_s)` instead of using the block.

lib/jekyll/post.rb
alfredxing: @majioa Looks good! Just one more comment below.

lib/jekyll/post.rb
majioa: @alfredxing the form I've used just the same as the other fetchers. but I'd prefer the view: `data['excerpt_separator] || site.config['excerpt_separator'].to_s`

parkr: How come? They do different things.

majioa: @parkr yes, but I believe in that context they will have the similar behaviour, because I think, no one will define `excerpt_separator` as a `nil` or `false`. And in my case the all the jekyll test results dont change the state after applying this change =)

envygeeks: `||` is better safe than sorry, people are people and people make mistakes, it's possible.  `Hash#fetch` does not protect against basic cases of human editing, human assumption and human mistakes.  Trust me, it happens, I've done it before.

envygeeks: And by "done it before" I mean stripped out something leaving only the key, not thinking.

parkr: Whatever the solution, let's ensure this method always returns a string. Otherwise it'll blow up `generate_excerpt?`

majioa: @parkr well, i'l fix the method to return string

parkr:
Thanks!

Issue #3265, opened 2015-01-07
Leading whitespace is stripped from highlight block


kevinushey:
Following from the [lanyon](https://github.com/poole/lanyon) base Jekyll theme, I have Markdown in a file of the form (this being output from R):

```
{% highlight text %}
      [,1] [,2]
[1,] FALSE TRUE
[2,] FALSE TRUE
{% endhighlight %}
```

which, after render (through whatever machinery runs in `jekyll serve`), gets generated HTML as:

```html
<div class="highlight"><pre><code class="language-text" data-lang="text">[,1]
[1,]    0
[2,]    1
[3,]    2
[4,]    2</code></pre></div>
```

with the leading whitespace in front of `[,1]` removed.

Is there a way to ensure that leading whitespace is not stripped on render -- perhaps something I can specify in `_config.yml`, or otherwise? FWIW, I have:

```
markdown:            kramdown
highlighter:         pygments
```

in my `_config.yml`, alongside other (obstensibly) unrelated options.

My understanding is that [pygments](http://pygments.org/) is handling this translation, and the [documentation](http://pygments.org/docs/quickstart/) here suggests that this might (?) be controlled through the `stripall` argument, but I am not sure how I would get access to that (if that is indeed what I need).

Thanks!

troyswanson:
Ref https://github.com/jekyll/jekyll-help/issues/229

parkr:
We strip all this data so that you don't have a newline at the beginning (i.e. *extra* space at the beginning of your code block) – we've been doing this for a while. Looks like it needs to be trimmed down such that it only strips newlines.

kevinushey:
Wonderful, thanks! Hugely appreciated :+1:

parkr:
Totally! It'll be out with Jekyll 3. :+1:

PR #3275, opened 2015-01-09
only strip newlines when rendering highlight blocks (#3265)

kevinushey:
This modifies the `render` function to only strip leading / trailing newlines, rather than all whitespace.

(This change could be contentious for some, so I do wonder if it could / should be hidden behind a configuration setting in `_config.yaml`, perhaps...)

This PR is not complete (ie -- needs tests), but hopefully it can be finished off with help from the more experienced contributors.


parkr:
You'll also want to include `\r`. We'll need tests for sure.

PR #3278, opened 2015-01-10
Highlight: Only Strip Newlines/Carriage Returns, not Spaces

parkr:
This PR expands on the work @kevinushey did in #3275. It includes tests.


parkr:
Closes #3265.

PR #3292, opened 2015-01-12
Eliminate the need for prev_section and next_section metadata.

flyinprogrammer:
Not sure if we want this code...

This PR started with me looking into how we do all this awesome documentation/collection stuff, and then I saw that we keep track of nav metadata in two places, the yaml file and in the document as metadata, and that made me really sad. I also hadn't ever done work with liquid - it's kinda cool, but kinda gross [the creation of the doc list makes me sad]. Regardless, here we are, and believe it or not this code doesn't seem to increase build time all that much.

### DIFF
--- a/site/_includes/section_nav.html
+++ b/site/_includes/section_nav.html
@@ -1,25 +1,16 @@
 {% comment %}
-Lets turn our yaml into a structured list of documents without titles.
+Map grabs the doc sections, giving us an array of arrays. Join, flattens all
+the items to a comma delimited string. Split turns it into an array again.
 {% endcomment %}
-
-{% assign docs = site.data.docs | map: 'docs' %}
-{% capture document_array_string %}
-{% for doc in docs %}
-  {% assign last_last = forloop.last %}
-  {% for title in doc %}
-    {{title}}{% unless forloop.last and last_last %},{% endunless %}
-  {% endfor %}
-{% endfor %}
-{% endcapture %}
-{% assign document_array = document_array_string | strip_newlines | split: ',' %}
+{% assign docs = site.data.docs | map: 'docs' | join: ',' | split: ',' %}
 
 {% comment %}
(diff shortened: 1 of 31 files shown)


site/_includes/section_nav.html
```diff
@@ -1,22 +1,48 @@
...
+{% assign document_array = document_array_string | strip_newlines | split: ',' %}
+
+{% comment %}
+Because this is built for every page, lets find where we are in the document list by comparing url strings.
+Then if there's something previous or next, lets build links to it.
+Note: When arrays are accessed directly, they are zero based.
+{% endcomment %}
+
+{% for document in document_array %}
+  {% assign document_url = document | prepend:"/docs/" | append:"/" %}
```
parkr: It would be cool if you could do `site.docs | where: "url",document` or something.

parkr:
Not bad. Lots of fancy Liquid going on there, but looks pretty solid.

How does this impact build performance?

jaybe-jekyll:
@flyinprogrammer Wonderful to see someone focusing on such aspects. At the time, the metadata-driven navigation was an improvement; prior-- it was static. :1234: 

I look forward to exploring your concepts. I am also curious about the differential in build times.

flyinprogrammer:
Got to thinking about chained filters and low and behold, I could dry up some code :)

With out change:
0m11.518s
0m11.693s
0m11.559s

With latest (b31d8a1) change:
0m11.782s
0m11.684s
0m11.933s

It makes me sad that iterations are so expensive. That said though, I feel like trading some time for possible human error is worth it - but I might also be biased. :wink:

parkr:
Worth it. Thanks!

troyswanson:
Awesome job. I've been wanting to do this since 7f3b35191c541ed8dae36b38ef24d4538ab52d17 (exactly one year ago!) but couldn't figure out how to do it. Your solution is exactly what I should have come up with.

Thanks :+1:

Issue #3298, opened 2015-01-13
Issue with date_to_rfc822 filter


ryanburnette:
I'm using Jekyll 2.5.1 and I'm using Collections pretty heavily. In each item in a particular collection I have a `date` field in my YAML front-matter. In one of my templates I filter that date using the `date_to_rfc822` Liquid filter.

The YAML front-matter item looks like this
```yaml
date: 2014-12-07
```

The template item looks like this
```html
{{ item.date | date_to_rfc822 }}
```

And this is the error I get.

```
Invalid Date: '2014-12-07' is not a valid datetime.
```

Looking at [the source of this filter][1] it looks like it's falling through the case statement.

Note that if I set the date like this it works fine.

```yaml
date: "2014-12-07"
```

[1]: https://github.com/jekyll/jekyll/blob/a098dc80cdd18e58479f005facbb985244f0c3a2/lib/jekyll/filters.rb#L301-L313

parkr:
What is the type of the first option?

ryanburnette:
I don't know. I was about to test that and I just had to jump on the phone for a bit. Great question. It should be a String I thought.

parkr:
This is probably a bug:

```text
>> yaml = "date: \"2014-12-07\"\ndate_other: 2014-12-07"
=> "date: \"2014-12-07\"\ndate_other: 2014-12-07"
>> SafeYAML.load(yaml)
=> {"date"=>"2014-12-07", "date_other"=>#<Date: 2014-12-07 ((2456999j,0s,0n),+0s,2299161j)>}
>> SafeYAML.load(yaml).map(&:inspect)
=> ["[\"date\", \"2014-12-07\"]", "[\"date_other\", #<Date: 2014-12-07 ((2456999j,0s,0n),+0s,2299161j)>]"]
```

PR #3299, opened 2015-01-13
Date method in filters.rb should accept a Date object

ryanburnette:
As uncovered in #3298, the `date` method in `filters.rb` should accept a Date object.

### DIFF
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -302,6 +302,8 @@ module Jekyll
       case input
       when Time
         input
+      when Date
+        input.to_time
       when String
         Time.parse(input) rescue Time.at(input.to_i)
       when Numeric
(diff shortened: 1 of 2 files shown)


parkr:
Can you please add a test? Also, does this require we `require 'date'`?

ryanburnette:
Tests written and passing. I should have done that first. ;)

It was working, but I figured it couldn't hurt to add `require 'date'`. I put it at the top of `filters.rb`, was that the right location?

troyswanson:
:clap: Nice.

parkr:
> I put it at the top of filters.rb, was that the right location?

We historically have put them in `jekyll.rb`, but I think this is perfect. Each file should hypothetically be individually-requirable.

parkr:
Well done! Merging.

PR #3320, opened 2015-01-17
Allow variables (e.g `:categories`) in YAML front matter permalinks

parkr:
Replaces #2462.


parkr:
Turns out you can do this already.

Issue #2715, opened 2014-08-09
Update how highlighting works


mattr-:
Not sure where we want to keep potential 3.0 work, but wanted to capture this relevant bit of info from #2709 regarding some work that we need to do for highlighting.

> In general our syntax highlighting needs a huge overhaul. We have Pygments and Rouge injected into places within Jekyll that would make Robert Louis Stevenson blush. It's an absolute mess. Most of our markdown converters find all the `<pre>` tags and use gsub (can you believe it?!) to provide some sort of highlighting. Few of the markdown libraries we use offer support for plugins, so we're a little dead in the water. We've been asked to match what GitHub is doing and we've bent over backwards to accomplish that. Unfortunately, it's left us in this present quandary.

> I think the path going forward is to ask markdown libraries to give us some means to work with their internal AST's (or whatever it is they build) to manipulate the code and pre tags to our liking. I have a strong urge to rip out all syntax highlighting in everything but the {% highlight %} tag and ship highlighting when it's ready from the Markdown library side (i.e. kramdown, RDiscount, RedCarpet).

> We may be able to use HTML Pipeline to port GitHub's Markdown approach, but we'll need a bit of assistance from them to put it all together in a useful way.

envygeeks:
I have a question on this topic: Why don't we use something like ContentPipeline (my library) or HTMLPipeline (Github's) which would make rendering much cleaner for everybody? Mine normalizes everything from every client and gives  you access to highlighting, I don't know whether HTMLPipeline does, but in general we could clean up the whole mess that is rendering by just offloading it onto an external plugin that is designed to do content.

envygeeks:
BTW, I'd be willing to hand over the source to content-pipeline if you wanted to use that in Jekyll (we don't use it on many of my projects and it was explicitly designed for Jekyll pretty much...)

parkr:
@envygeeks I dig the idea of using a Pipeline. We're already using HTMLPipeline for jekyll-mentions and jemoji. Does it depend on Nokogiri, though? My big worry is including C extensions again.

PR #3323, opened 2015-01-18
Remove most runtime deps & move to Rouge as default highlighter

parkr:
This moves the following gems to as-wanted:

- pygments.rb
- redcarpet
- toml
- jekyll-paginate
- jekyll-gist
- jekyll-coffeescript
- classifier-reborn

~~I'm on the fence about also taking `jekyll-watch` away...~~ We're going to keep it.

/cc @jekyll/core

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,14 @@
 source 'https://rubygems.org'
 gemspec
 
+gem 'pygments.rb', '~> 0.6.0'
+gem 'redcarpet', '~> 3.1'
+gem 'toml', '~> 0.1.0'
+gem 'jekyll-paginate', '~> 1.0'
+gem 'jekyll-gist', '~> 1.0'
+gem 'jekyll-coffeescript', '~> 1.0'
+gem 'classifier-reborn', '~> 2.0'
+
 gem 'rake', '~> 10.1'
 gem 'rdoc', '~> 3.11'
 gem 'redgreen', '~> 1.2'
@@ -16,7 +24,6 @@ gem 'mime-types', '~> 1.5'
 gem 'activesupport', '~> 3.2.13'
 gem 'jekyll_test_plugin'
 gem 'jekyll_test_plugin_malicious'
-gem 'rouge', '~> 1.7'
 gem 'liquid-c', '~> 0.0.3'
 gem 'minitest' if RUBY_PLATFORM =~ /cygwin/
 gem 'test-unit' if RUBY_PLATFORM =~ /cygwin/ || RUBY_VERSION.start_with?("2.2")
--- a/features/markdown.feature
+++ b/features/markdown.feature
@@ -17,7 +17,10 @@ Feature: Markdown
     And I should see "<h1 id=\"my-title\">My Title</h1>" in "_site/index.html"
 
   Scenario: Markdown in pagination on index
-    Given I have a configuration file with "paginate" set to "5"
+    Given I have a configuration file with:
+      | key      | value             |
+      | paginate | 5                 |
+      | gems     | [jekyll-paginate] |
     And I have an "index.html" page that contains "Index - {% for post in paginator.posts %} {{ post.content }} {% endfor %}"
     And I have a _posts directory
     And I have the following post:
--- a/features/pagination.feature
+++ b/features/pagination.feature
@@ -4,7 +4,10 @@ Feature: Site pagination
   I want divide the posts in several pages
 
   Scenario Outline: Paginate with N posts per page
-    Given I have a configuration file with "paginate" set to "<num>"
+    Given I have a configuration file with:
+      | key      | value             |
+      | paginate | <num>             |
+      | gems     | [jekyll-paginate] |
     And I have a _layouts directory
     And I have an "index.html" page that contains "{{ paginator.posts.size }}"
     And I have a _posts directory
@@ -32,6 +35,7 @@ Feature: Site pagination
       | paginate      | 1                              |
       | paginate_path | /blog/page-:num                |
       | permalink     | /blog/:year/:month/:day/:title |
+      | gems          | [jekyll-paginate]              |
     And I have a blog directory
     And I have an "blog/index.html" page that contains "{{ paginator.posts.size }}"
     And I have a _posts directory
@@ -59,6 +63,7 @@ Feature: Site pagination
       | paginate      | 1                              |
       | paginate_path | /blog/page/:num                |
       | permalink     | /blog/:year/:month/:day/:title |
+      | gems          | [jekyll-paginate]              |
     And I have a blog directory
     And I have an "blog/index.html" page that contains "{{ paginator.posts.size }}"
     And I have an "index.html" page that contains "Don't pick me!"
--- a/features/rendering.feature
+++ b/features/rendering.feature
@@ -15,6 +15,7 @@ Feature: Rendering
   Scenario: Don't place asset files in layout
     Given I have an "index.scss" page with layout "simple" that contains ".foo-bar { color:black; }"
     And I have an "index.coffee" page with layout "simple" that contains "whatever()"
+    And I have a configuration file with "gems" set to "[jekyll-coffeescript]"
     And I have a simple layout that contains "{{ content }}Ahoy, indeed!"
     When I run jekyll build
     Then the _site directory should exist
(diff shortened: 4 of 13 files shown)


envygeeks:
I'm very much :+1: on this.

parkr:
Once this is merged, I'd like to get a beta out into the wild.

envygeeks:
@parkr can the beta wait like 2 days past tomorrow? I can definitely finish cleaning up our tests into a cleaner Minitest state by then if it can wait, and we can get that out the door with it.

parkr:
> @parkr can the beta wait like 2 days past tomorrow? I can definitely finish cleaning up our tests into a cleaner Minitest state by then if it can wait, and we can get that out the door with it.

Beta on Tuesday? Works for me.

parkr:
@envygeeks I'm also trying to figure out what tests are really slow because, yeah, some of them are terrible but I don't have a clear idea of which ones they are.

envygeeks:
@parkr I actually added that in (because Minitest has an option for that) it seems most of the problem isn't the actual testing, it's the generation of the sites over and over again, if you want though I can always set the threshold smaller so we can track anything really expensive: https://github.com/jekyll/jekyll/commit/d25963714afdc50c97604706c101f8dc480df488#diff-fa17210be88d091a513b7c0ed2cf5dafR27

parkr:
> it seems most of the problem isn't the actual testing, it's the generation of the sites over and over again

Ah! Ok cool. Some of it might also be that we're still testing plugins. :x

paulrobertlloyd:
I’d love to try this when it’s ready! Hopefully it will replace the current hacks I have in place to ensure kramdown and pygments play nice together.

parkr:
> kramdown and pygments

This simply changes the default highlighter to `rouge`, it doesn't alter the integration between kramdown and pygments, so I'm not sure what you mean.

paulrobertlloyd:
I mean, hopefully these new defaults will work better than the current defaults, which don't highlight code within fenced code blocks as expected (only code within `{% highlight %}` liquid blocks is parsed by pygments). See:

* https://github.com/jekyll/jekyll/issues/2715#issuecomment-58829496
* http://bloerg.net/2013/03/07/using-kramdown-instead-of-maruku.html
* https://github.com/navarroj/krampygs
* https://github.com/mvdbos/kramdown-with-pygments
* and my current hack: https://github.com/paulrobertlloyd/jekyll-pypedown
* …

This has been a long frustration of mine! Happy to test this to see if Rouge highlights fenced code blocks as expected.

alfredxing:
:+1:

I'd like to keep `jekyll-watch` in for now, since it's quite a core feature for the Jekyll CLI.

pathawks:
> I'm on the fence about also taking `jekyll-watch` away...

It’s *so* easy to add A Gem to a `Gemfile` as needed, what’s the big advantage to keeping it bolted to core?

envygeeks:
It's already in there so you are posing the wrong question: What is the big advantage of removing it?

parkr:
The advantage of removing it is there are fewer gems to download. The disadvantage is that it makes it more obtuse. With `jekyll serve` defaulting to using watch, that would also cause problems.

XhmikosR:
For what is worth I too believe watch should be part of core being that it's the default with `serve`.

Big :+1: about Rouge!

pathawks:
> For what is worth I too believe watch should be part of core being that it's the default with `serve`.

I guess since I never use `serve`, I wouldn't miss it.

Would decoupling it result in a lighter code review burden for GitHub Pages?

kleinfreund:
> I guess since I never use serve, I wouldn't miss it.

Huh. I use `jekyll s` all the time.

envygeeks:
I have to agree with @kleinfreund I rely on `jekyll s` to develop, it would drive me crazy to lose it out of the box, and I don't think it's loaded when Jekyll is not using it so I don't believe it impacts performance of production building at all.

parkr:
Cool, thanks for the input everyone. I have crossed that line out. We'll keep it. It doesn't impact the security pressure on GitHub for Pages support, because, as @envygeeks said, it's not required unless explicitly used, and they don't use it.

pattulus:
@paulrobertlloyd Sadly Jekyll 3.0 with `kramdown` and `rouge` doesn't support styling fenced code blocks. As before setting the liquid tag `{% highlight %}` is the only option as you can see in the demo code below:

    <p>fenced code block with backticks</p>
    <pre>
        <code class="language-ruby">def foo
          puts 'foo'
        end
        </code>
    </pre>

    <p>fenced codeblock with backticks</p>
    <pre>
        <code class="language-ruby">def foo
          puts 'foo'
        end
        </code>
    </pre>

    <p>liquid tags</p>
    <div class="highlight">
        <pre>
            <code class="language-ruby" data-lang="ruby">
                <span class="k">def</span>
                <span class="nf">foo</span>
                <span class="nb">puts</span>
                <span class="s1">'foo'</span>
                <span class="k">end</span>
            </code>
        </pre>
    </div>

I'd also like to see support for fenced code blocks. It would be one problem less on the list and shipping Jekyll with support for such a popular format to write code blocks out of the box would be amazing.

paulrobertlloyd:
@pattulus I’ve actually meaning to file bug against this… you _can_ get kramdown to highlight fenced code blocks with Rouge, but you have to set it as a separate config option, as follows:

```yaml
highlighter: rouge
kramdown:
  syntax_highlighter: rouge
```

(Strangely, this doesn't work if you wish to use Pygments).

Not sure if this is the expected behaviour, or even desirable (having to explicitly config a highlighter twice). Essentially `highlighter` relates to `{% highlight ruby %}`, while `kramdown: syntax_highlighter: rouge` relates to `~~~ruby`.

parkr:
@paulrobertlloyd Think we should enable the `kramdown.syntax_highlighter` automatically based on `highlighter`'s value?

paulrobertlloyd:
@parkr I think so… I mean, that’s the behaviour I expected, but I may be an outlier! I think that would be a sensible default (assuming you could override with `kramdown.syntax_highlighter`… or does that get even more complicated).

(Short answer: yes!)

pattulus:
@paulrobertlloyd That's a super-nifty trick. I hadn't set `highlighter: rouge` before. Now I have both of them set, like in your example. Still, using `~~~ruby` won't show any highlighting, nor output any of the spans in the code tag.

Any ideas if there's something missing in my config file? I'd really like to use fenced code blocks.

~~~yml
source:        _src
encoding:      utf-8
safe:          false
future:        false
unpublished:   false
lsi:           false
baseurl: "" # the subpath of your site, e.g. /blog/
url:           http://derubercast.link
permalink:     /podcast/:title.html

markdown: kramdown
highlighter: rouge
kramdown:
    syntax_highlighter: rouge
    input: GFM
    auto_ids: true
#   footnote_nr: 1
#   entity_output: as_char
    toc_levels: 1..3
    smart_quotes: lsquo,rsquo,ldquo,rdquo

gems: ['bourbon']
~~~

paulrobertlloyd:
@pattulus Ah, if you’re using `input: GFM` you should use three backpacks, not three tildes. Hopefully that should then work.

pattulus:
@paulrobertlloyd Mh. Already tried it and replace all tildes with backticks. I also tested both versions without the GFM option. Strange that it doesn't work here.

envygeeks:
Can we please move all this noise to a ticket of it's own?

pattulus:
Sorry. I didn't want to hijack this thing.
Last note: I figured it out, I had to use `bundle` to install `rouge` rather than doing `gem install` ([source](http://stackoverflow.com/questions/15191396/octopress-jekyll-syntax-highlighting-what-are-the-advantages-of-backticks#comment21507300_15196577))

paulrobertlloyd:
Moved discussion of syntax highlighting config woes to a new issue: https://github.com/jekyll/jekyll/issues/3371 /cc @parkr @pattulus

mrzool:
I guess then [the documentation](http://jekyllrb.com/docs/templates/#code-snippet-highlighting) isn't up to date on this?

screen shot 2015-11-20 at 18 01 22

envygeeks:
@mrzool yes, we no longer bundle pygments.rb, now we use rouge by default,  the surface of it applies but it could use a slight reworking to imply that pygments is not included but optional and rouge is used instead now.

PR #3325, opened 2015-01-18
Strip slashes on nil url tokens

imathis:
This will make it cleaner to generate urls with optional keys in the permalink template.

### DIFF
--- a/lib/jekyll/url.rb
+++ b/lib/jekyll/url.rb
@@ -61,7 +61,12 @@ module Jekyll
     def generate_url(template)
       @placeholders.inject(template) do |result, token|
         break result if result.index(':').nil?
-        result.gsub(/:#{token.first}/, self.class.escape_path(token.last))
+        if token.last.nil?
+          # Remove leading '/' to avoid generating urls with `//`
+          result.gsub(/\/:#{token.first}/, '')
+        else
+          result.gsub(/:#{token.first}/, self.class.escape_path(token.last))
+        end
       end
     end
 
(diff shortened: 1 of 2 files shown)


parkr:
Excellent, thank you! Would you mind adding a unit test? I believe they're in `test/test_url.rb`. If not, no problem. I'll add them at merge time! Thanks :smile:

imathis:
Request granted :)

parkr:
Thanks!

imathis:
@parkr. Done.

One other thing though. As I'm working with permalinks I'm finding that the default template names — `pretty`, `date`, etc — only work in the site's permalink config. Setting `permalink: date` in a post's front-matter puts the post at `/date/index.html` which is totally not what I was expecting.

Can you confirm that this is how it's supposed to work? That's not the impression I got from the docs at all.

parkr:
LGTM! :shipit: 

> As I'm working with permalinks I'm finding that the default template names — pretty, date, etc — only work in the site's permalink config. Setting permalink: date in a post's front-matter puts the post at /date/index.html which is totally not what I was expecting.

Total parity would be nice. The only problem is that `permalink` in a post is considered as-is. It's an experiment but I don't want it to go awry! Haha. @benbalter, do you think allowing the special url template names (e.g.  `permalink: pretty`) inside a post's YAML front matter is risky? Seems like a smart idea to me but I don't want to go being clever and inadvertently break things. :wink: Thanks!

parkr:
:tada: :confetti_ball:

alfredxing:
:pray:

PR #3329, opened 2015-01-19
Fix nav items alignment when in "burger" mode

tkrotoff:
I have introduced a bug with [this pull request](https://github.com/jekyll/jekyll/pull/3264) when the menu is in "burger" mode:

"burger" mode bug

Original CSS rule:
```CSS
.page-link {
    &:not(:first-child) {
        margin-left: 20px;
   }
}
```

[First pull request](https://github.com/jekyll/jekyll/pull/3264) to fix nav items on multiple lines:
```CSS
.page-link {
    &:not(:last-child) {
        margin-right: 20px;
    }
}
```

And we also need this to fix the "burger" menu:
```CSS
@include media-query($on-palm) {
    .page-link {
        margin-right: 0 !important;
        margin-left: 20px;
    }
}
```

### DIFF
--- a/lib/site_template/_sass/_layout.scss
+++ b/lib/site_template/_sass/_layout.scss
@@ -82,6 +82,9 @@
         .page-link {
             display: block;
             padding: 5px 10px;
+
+            margin-right: 0 !important;
+            margin-left: 20px;
         }
     }
 }


parkr:
We have a "burger" mode?

kleinfreund:
Vertically stacked navigational items that is. :P

lib/site_template/_sass/_layout.scss
```diff
@@ -82,6 +82,9 @@
         .page-link {
             display: block;
             padding: 5px 10px;
+
+            margin-right: 0 !important;
```
alfredxing: Not a big fan of `!important`. Any way to get around this?

parkr:
Whoops, sorry @alfredxing! Didn't see your comment. How would you propose getting around it? Maybe having a `@media` style?

alfredxing:
We would need the same level of specificity (`:not(:last-child)`) so the media query would override the original rule:

```scss
@include media-query($on-palm) {
    ...
    .page-link {
        display: block;
        padding: 5px 10px;
        margin-left: 20px;

        &:not(:last-child) {
            margin-right: 0;
        }
    }
}
```

@parkr Should I submit a PR or get @tkrotoff to update the branch?

tkrotoff:
@alfredxing :+1: tested and works fine

PR #3331, opened 2015-01-21
Fix site-nav alignment on small screens

tkrotoff:
Before:
30px

After:
$spacing-unit / 2

(Screenshots from iPhone 4s simulator)

Considering [the site title is 15px away from the left screen border](https://github.com/jekyll/jekyll/blob/v2.5.3/lib/site_template/_sass/_base.scss#L169) ($spacing-unit is 30px) on small screens:
```CSS
.wrapper {
    @include media-query($on-laptop) {
        padding-right: $spacing-unit / 2;
        padding-left: $spacing-unit / 2;
    }
}
```
site-nav (the "burger" menu) should be 15px from the right screen border instead of the hardcoded value of 30px.

### DIFF
--- a/lib/site_template/_sass/_layout.scss
+++ b/lib/site_template/_sass/_layout.scss
@@ -44,7 +44,7 @@
     @include media-query($on-palm) {
         position: absolute;
         top: 9px;
-        right: 30px;
+        right: $spacing-unit / 2;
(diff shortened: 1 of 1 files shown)


parkr:
:+1:

mattr-:
:+1:

PR #3347, opened 2015-01-24
Default to excluding common Ruby folders.

envygeeks:
This change implements a default exclude for Gemfile, Gemfile.lock and vendor because they are common Ruby files and they can hand up a user getting started quickly if following any recommendation for how to do Jekyll.  In that we express they should use Gemfile's but when you use `--path` with that you end up having an unbuildable site unless you manually set the exclude, which is bad.  We should default to excluding vendor and bundler files to just make life easier for everyone on the planet.


envygeeks:
Probably better not to exclude by default!

Issue #2346, opened 2014-05-07
List of collections not rendering


paulrobertlloyd:
I am trying to build a list of the collections I have defined in my _config.yml, thus:

``` yaml
collections:
  articles:
    output: true
    title: Articles
  essays:
    output: true
    title: Academic Essays
  presentations:
    output: true
    title: Presentations
  projects:
    output: true
    title: Projects
```

using the following Liquid template:

```
{% for collection in site.collections %}
    <section>
        <h1><a href="{{ collection.directory }}">{{ collection.title }}</a></h1>
        Label: {{ collection.label }}
    </section>
{% endfor %}
```

but the following is rendered:

```
<section>
    <h1><a href=""></a></h1>
    Label: 
</section>
<section>
    <h1><a href=""></a></h1>
    Label: 
</section>
<section>
    <h1><a href=""></a></h1>
    Label: 
</section>
<section>
    <h1><a href=""></a></h1>
    Label: 
</section>
```

The [documentation](http://jekyllrb.com/docs/collections/) says:

> …collections are also available under `site.collections`, with the metadata you specified in your `_config.yml`…

Is what I’m attempting correct? Or is this a bug?

Thanks—and congratulations on shipping v2!

bernardodiasc:
Hello, is this issue fixed? I'm still able to reproduce in Jekyll 2.5.3.

parkr:
> Hello, is this issue fixed? I'm still able to reproduce in Jekyll 2.5.3.

Can you please post a sample site or sample code to repro?

bernardodiasc:
This feature is not implemented yet in the project because this issue, but is pretty much identical as described here. Checking the [master branch (site.rb#L397-L398)](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/site.rb#L397-L398), seems the changes are not available yet in the 2.5.3 version, double checked in my Jekyll source files. Is that correct?

bernardodiasc:
```
{% for collection in site.collections %}
  {{ collection }}
{% endfor %}
```

return:

```
book#<Jekyll::Collection:0x0000000227a068>
concepts#<Jekyll::Collection:0x00000002279f28>
```

bernardodiasc:
Applying the change  `"collections"  => collections.values.map(&:to_liquid),` on `site.rb` works as expected, like already figured. My curiosity is about in what release is this fix (not on 2.5.3), and if works on GHP. Thanks!

parkr:
[...]

PR #3349, opened 2015-01-24
Collections output

bernardodiasc:
This fix the collections output as suggested in this thread https://github.com/jekyll/jekyll/issues/2346#issuecomment-42644189

### DIFF
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -394,7 +394,7 @@ module Jekyll
             "html_pages"   => pages.select { |page| page.html? || page.url.end_with?("/") },
             "categories"   => post_attr_hash('categories'),
             "tags"         => post_attr_hash('tags'),
-            "collections"  => collections,
+            "collections"  => collections.values.map(&:to_liquid),
             "documents"    => documents,
             "data"         => site_data
         }))


hubgit:
This small change makes collections work as I'd hoped they would (the list of collections can be iterated in templates, and their properties and items can be output). Is there another way to achieve this, or should the change be merged?

Currently it's possible to do this:

```liquid
<!-- figures -->
{% for item in site.figures %}
    <div id="figures">
        <h2>Figures</h2>
        {{ item.output }}
    </div>
{% endfor %}

<!-- tables -->
{% for item in site.tables %}
    <div id="tables">
        <h2>Tables</h2>
        {{ item.output }}
    </div>
{% endfor %}
```

but not this:

```liquid
<!-- figures and tables -->
{% for collection in site.collections %}
    <div id="{{ collection.label }}">
        <h2>{{ collection.title }}</h2>
        {% for item in collection.docs %}
            {{ item.output }}
        {% endfor %}
    </div>
{% endfor %}
```

I'd also be ok with knowing that the first example above is the correct way to do things, and that `site.collections` shouldn't be used in templates, if that is the case.

FloFra:
This is not a bug, but due to the fact that ```site.collections``` is a associated list and you are trying to use it like a normal array/list and therefore wrong. See my comment on #2346 for some examples. 

@hubgit: In your case just replace all ```collection``` with ```collection[1]``` and you should be fine.

parkr:
Continuing in #3670.

PR #3360, opened 2015-01-26
A lighter, smarter, font declaration

nternetinspired:
This PR introduces a new $base-font-weight variable, as we already have a var for other base font values, and then declares the `<body>` font properties with a single declaration in place of four.

Easier to configure, less css, same visual output.


XhmikosR:
:+1: for using the `font` shorthand property.

You should squash those two patches.

nternetinspired:
:+1: 

Squashed. Thanks @XhmikosR

PR #3361, opened 2015-01-26
Adds main sectioning element

nternetinspired:
This PR adds the `<main>` sectioning element. 

The rendered document already uses the html5 sectioning elements `<header>` and `<footer>`, this PR extends that by adding the semantic element for the main document body.

### DIFF
--- a/lib/site_template/_sass/_base.scss
+++ b/lib/site_template/_sass/_base.scss
@@ -9,6 +9,25 @@ dl, dd, ol, ul, figure {
 }
 
 
+/**
+ * Correct html5 display definition for older browsers
+ */
+article,
+aside,
+details,
+figcaption,
+figure,
+footer,
+header,
+main,
+menu,
+nav,
+section,
+summary {
+    display: block;
+}
+
+
 
 /**
  * Basic styling
--- a/lib/site_template/_layouts/default.html
+++ b/lib/site_template/_layouts/default.html
@@ -7,11 +7,11 @@
 
     {% include header.html %}
 
-    <div class="page-content">
+    <main class="page-content">
       <div class="wrapper">
         {{ content }}
       </div>
-    </div>
+    </main>
 
     {% include footer.html %}
 


nternetinspired:
Done in https://github.com/nternetinspired/jekyll/commit/a748c7620d91912f395023980744d2c804234c4d

parkr:
Can you link me to the docs for this element? I've never seen it before.

mattr-:
@parkr https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main

@nternetinspired You're missing the `role="main"` attribute for the element. Also, considering this isn't supported on **any** version of IE [according to caniuse](http://caniuse.com/#feat=html5semantic), I'm questioning the value of this particular change. In my mind, it's not important to make use of every single new HTML5 element that's out there, especially since it doesn't enhance the user experience in any way.

alfredxing:
Seems nice semantically but the IE compatibility really kills it for me.

kleinfreund:
Supporting the `main` element on Internet Explorer is as easy as adding `display: block;` CSS rule. It will work up from IE 9 then.

The `role` attribute might still be added, although the only issue here is that some browsers haven't yet mapped the `main` role to the `main` element.

nternetinspired:
@mattr- Thanks for adding a link. I'll add a link to the spec also:

http://www.w3.org/html/wg/drafts/html/master/grouping-content.html#the-main-element

The ARIA role was intentionally omitted, as my previous attempt to add ARIA Landmark Roles was unsuccessful: https://github.com/jekyll/jekyll/pull/2962. In this case it's really a nice-to-have, an accessibility bonus. It isn't a required attribute.

Not supported, in the case of old IE versions, is also not as bad as it sounds. As @kleinfreund rightly points out, all that is required to implement support for OldIE is `display: block;` and this is also required to ensure correct rendering in html5 elements already in use in the default template, so this PR extends backward compatibility to old versions of IE.

nternetinspired:
Normalize.css uses similar css to ensure html5 element rendering consistency back to IE8: 

https://github.com/necolas/normalize.css/blob/master/normalize.css#L26 with `<main>` being added more than 2yrs ago: https://github.com/necolas/normalize.css/commit/2eb80024f66dd5745cba5dcef5c9b0c466ecb8d5

XhmikosR:
@nternetinspired: why don't you just add the `main` element instead of copying all that block from normalize.css?

kleinfreund:
The majority of the HTML5 sectioning elements only needs the CSS rules for when HTML5 is not supported at all. We should only address the missing support in the latest It's here, I think.

XhmikosR:
I'm simply saying, either include normalize.css or just the needed selector. What's the point in including manually unused selectors?

kleinfreund:
Ah, btw I meant IE's. Not it's. I hate typing on mobile.

@xhmikosr That's what I'm saying.

XhmikosR:
:cool:

nternetinspired:
> @nternetinspired: why don't you just add the main element instead of copying all that block from normalize.css?

As already stated @XhmikosR that addition extends html5 compatibility, as the default template *already* makes widespread use of html5 elements, to older browsers. It seemed prudent to make the default template as widely supported as possible, given the minimal effort that requires. 

It's no problem to me if the consensus is against this change, as I don't use the default template, so feel free to shoot this PR down.

@kleinfreund, where would you suggest is the cut-off for latest IEs, IE10? In that case just:

    details,
    main,
    summary {
        display: block;
    }

XhmikosR:
Well, if that's the case @nternetinspired, then you should just include vanilla normalize.css and not just those selectors. Either go with the whole file, or just the needed selectors. Because I don't think the template is using `figcaption` etc...

Plus, this would be easier for maintenance and future updates.

Anyway, it's @park's and the other @jekyll/owners's decision to make, ultimately.

kleinfreund:
@nternetinspired That's actually something we should make a decision for. What browsers are we targeting with the site template? I'd say from IE10 is fine. IE9 user numbers are ingorable and supporting IE8 would be quite some extra work. However, there is also mobile browsers.

lib/site_template/_sass/_base.scss
```diff
@@ -9,6 +9,25 @@ dl, dd, ol, ul, figure {
...
+figure,
+footer,
+header,
+main,
+menu,
+nav,
+section,
+summary {
+    display: block;
+}
```
parkr: :scream_cat:

kleinfreund: @parkr As mentioned by @nternetinspired above, we probably only need these:

```
details,
main,
summary {
    display: block;
}
```

PR #3370, opened 2015-01-26
Update History.markdown.

XhmikosR:


parkr:
What's your motivation for this? I'm worried my `git blame` for each line will be less accessible for when we have to look back.

mattr-:
Thanks for the cleanup! having `History.markdown` look all nice and pretty is not necessarily a thing we're super concerned with and so I'm closing this on the grounds that preserving history is more important than 100% proper markdown. We as the core team will make every attempt to do better in the future. :heart:

PR #3376, opened 2015-01-29
Switch default font to Helvetica Neue

tkrotoff:
Helvetica Neue (1983) features multiple improvements over the original Helvetica font (1957), 'neue' means 'new' in German

If the OS does not include Helvetica Neue, then it falls back to Helvetica and if ...

FYI
- Bootstrap: [`"Helvetica Neue", Helvetica, Arial, sans-serif`](https://github.com/twbs/bootstrap/blob/v3.3.2/less/variables.less#L45) (here a discussion about it: https://github.com/twbs/bootstrap/issues/13823)
- Ionic: [`"Helvetica Neue", "Roboto", sans-serif !default`](https://github.com/driftyco/ionic/blob/v1.0.0-beta.14/scss/_variables.scss#L19)
- Foundation: [`"Helvetica Neue", "Helvetica", Helvetica, Arial, sans-serif !important`](https://github.com/zurb/foundation/blob/v5.5.0/doc/assets/scss/_global.scss#L45) (<- curious about why they double name Helvetica...)
- Disqus: `"Helvetica Neue", Helvetica, Arial, sans-serif`

Helvetica Neue is the default font for OS X >= 10.10 and iOS >= 7
Windows: Segoe UI, Android: Roboto

### DIFF
--- a/lib/site_template/css/main.scss
+++ b/lib/site_template/css/main.scss
@@ -6,7 +6,7 @@
 
 
 // Our variables
-$base-font-family: Helvetica, Arial, sans-serif;
+$base-font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
 $base-font-size:   16px;
 $small-font-size:  $base-font-size * 0.875;
 $base-line-height: 1.5;


kleinfreund:
I think this is a good idea.

However I think naming other sites that use a certain font stack is a rather bad argument for a font choice. Why is font choice related to what _big names_ use on the interwebz? It is not.

tkrotoff:
@kleinfreund 
> naming other sites [...] is a rather bad argument for a font choice

To check how others solve a problem similar to yours is never a bad idea. Of course it does not mean that you have to blindly follow the same path.

The "other sites" are not just random sites: this is the current top list of CSS frameworks. This list is relevant since we are talking here about the default micro CSS framework featured in Jekyll.

kleinfreund:
> The "other sites" are not just random sites: this is the current top list of CSS frameworks. This list is relevant since we are talking here about the default micro CSS framework featured in Jekyll.

For choosing a proper font, they are just that: random sites. Their font decisions can't be abstracted away to other sites only because they are in the same market. Anyways, this is offtopic.

tkrotoff:
Btw, on the font subject and not offtopic :)

I've noticed that my site title is not easy to read (with Helvetica bold 34px on OS X):

letter-spacing: -1px

Jekyll CSS defines `letter-spacing: -1px` several times: https://github.com/jekyll/jekyll/search?q=letter-spacing
It would be better to not specify `letter-spacing` and let the default font as is.

Without `letter-spacing: -1px`:

no letter-spacing

FYI (=> :-))
None of [Bootstrap](https://github.com/twbs/bootstrap/search?q=letter-spacing&type=Code), [Ionic](https://github.com/driftyco/ionic/search?q=letter-spacing) and [Foundation](https://github.com/zurb/foundation/search?q=letter-spacing) change `letter-spacing`.
It is thus probably wise to not do so with Jekyll either. If interested I can create a pull request.

kleinfreund:
:+1: for that as well.

parkr:
@jglovier, What do you think about this?

This is really splitting hairs for me, so I don't much care either way.

jglovier:
Honestly I could have sworn I specified "Helvetica Neue" in my original font stack. *scratches head*

Yeah, it's kinda splitting hairs, by def my kind of splitting hairs. :wink: :+1: :metal:

parkr:
Ok, cool. Thanks so much, @jglovier :heart:

m23mclaughlin:
How might one change the default font to "Courier"?

parkr:
@m23mclaughlin When you run `jekyll new <path>`, `cd` into `<path>` and edit the value in `css/main.scss`.

PR #3380, opened 2015-01-29
Factor out a `read_data_file` call to keep things clean

gjtorikian:
I actually need this for a monkey-patch I am planning. :wink: But it does look cleaner, no?

/cc @parkr

### DIFF
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -258,16 +258,23 @@ module Jekyll
         if File.directory?(path)
           read_data_to(path, data[key] = {})
         else
-          case File.extname(path).downcase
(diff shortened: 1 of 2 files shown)


parkr:
Looks good to me, but it needs a test. :smile:

gjtorikian:
Hrm, I would think the fact that nothing broke would be a succeeding test? :wink: Ok, ok...

alfredxing:
:+1: Would also be great if we could allow custom data formats in the future in the form of plugins.

parkr:
> Would also be great if we could allow custom data formats in the future in the form of plugins.

We could also start using converters...

PR #3379, opened 2015-01-29
Remove 'letter-spacing: -1px'

tkrotoff:
With `letter-spacing: -1px`:

Without `letter-spacing: -1px`:

(screenshots with Helvetica Neue/OS X)

FYI
None of [Bootstrap](https://github.com/twbs/bootstrap/search?q=letter-spacing&type=Code), [Ionic](https://github.com/driftyco/ionic/search?q=letter-spacing) and [Foundation](https://github.com/zurb/foundation/search?q=letter-spacing) change `letter-spacing`, it is thus probably wise to not do so with Jekyll either.

PR #3382, opened 2015-01-30
Optimize text rendering for legibility

alfredxing:
Add `text-rendering: optimizeLegibility` to enable kerning and ligatures. This applies to the site template as well as the jekyllrb.com site.

For the site template, this should produce a more desirable result for the site header when combined with #3379.

:sparkles: Kerning! :sparkles: Ligatures! :sparkles:

### DIFF
--- a/lib/site_template/_sass/_base.scss
+++ b/lib/site_template/_sass/_base.scss
@@ -18,6 +18,11 @@ body {
     color: $text-color;
     background-color: $background-color;
     -webkit-text-size-adjust: 100%;
+    -webkit-font-feature-settings: "kern" 1;
+    -moz-font-feature-settings: "kern" 1;
+    -o-font-feature-settings: "kern" 1;
+    font-feature-settings: "kern" 1;
+    font-kerning: normal;
 }
 
 
--- a/site/_sass/_style.scss
+++ b/site/_sass/_style.scss
@@ -13,6 +13,11 @@ body {
   border-top: 5px solid #fc0;
   @include box-shadow(inset 0 3px 30px rgba(0,0,0,.3));
   text-shadow: 0 1px 3px rgba(0,0,0,.5);
+  -webkit-font-feature-settings: "kern" 1;
+  -moz-font-feature-settings: "kern" 1;
+  -o-font-feature-settings: "kern" 1;
+  font-feature-settings: "kern" 1;
+  font-kerning: normal;
 }
 
 .clear {


kleinfreund:
:-1: 

- Ligatures work for font sizes smaller than 20 pixels and _certain_ fonts only.
- potential for performance issues (http://www.marco.org/2012/11/15/text-rendering-optimize-legibility)
- I had some issues with this property on my own page when using webfonts, but I don't recall what the exact pitfalls were

I think we should not add stuff to the basic site template unless there are good reasons for it. I doubt the benefits of `text-rendering: optimizeLegibility;` when applied to Helvetica/Arial.

parkr:
> I think we should not add stuff to the basic site template unless there are good reasons for it.

@kleinfreund I agree with this on principle, which is why the other PR's I received today (yesterday?) were mostly skimmed and deferred. They didn't seem important. The site template as-is is perfectly fine at the moment.

@alfredxing I have no problem with this change.

alfredxing:
@kleinfreund Was not aware of the performance issues, thanks for the link!

Do you think `font-feature-settings` would be better? At the very least I think kerning is needed.

jglovier:
Yeah, as @kleinfreund said the benefits of this property would only be realized with typefaces that we aren't using here by default. For our Helvetica stack, I don't think we'd really see much if any benefit.

>Do you think font-feature-settings would be better? At the very least I think kerning is needed.

@alfredxing I'm of the opinion that kerning web-fonts should always be handled manually through the use of the `letter-spacing` property. There may be a good argument against that of which I'm unaware, so feel free to push back if you think that's not a good approach.

alfredxing:
@jglovier Almost all fonts have kerning properties. Kerning is defined as letter spacing adjustments between certain pairs of letters, so using `letter-spacing` in general won't bring the same effect (unless you manually adjust `letter-spacing` for each pair of letters).

Here's a comparison of kerning vs no kerning (notice the `Yo` pair):

Comparison

parkr:
I'm really indifferent as long as it doesn't raise the barrier to entry. @bshackleford, ever used this property before? What do you think are its virtues?

bshackelford:
@parkr I'd recommend using 
```
font-feature-settings: "kern" 1;
font-kerning: normal;
```
instead of `text-rendering: optimizeLegibility` since it's not an official css property and could have performance implications. Here's a great article from Elliot Jay Stocks on the subject: http://www.elliotjaystocks.com/blog/kerning/

parkr:
Excellent, thank you!!

What do you think, @alfredxing?

nternetinspired:
I'd agree that font-feature is preferable, though this property always needs prefixing: http://caniuse.com/#feat=font-feature

> -webkit-font-feature-settings: "kern" 1,  "liga" 1;
      -moz-font-feature-settings: "kern" 1,  "liga" 1;
               font-feature-settings: "kern" 1,  "liga" 1;

alfredxing:
@parkr That looks great, I'll fix it up!

jglovier:
+:100: for `font-feature-settings` with prefixes.

tkrotoff:
@alfredxing In your commit you define font-feature-settings with -webkit, -moz and -o prefixes.
The rest of the code inside site_template does not use -moz and -o (so just -webkit for old Android and iOS devices I guess).

- -o is useless: recent versions of Opera (>= 15) uses -webkit, -o is for old versions of Opera (<= 12). Opera 27 (latest) uses the -webkit prefix for font-feature-settings and v12 does not even support font-feature-settings.
- -moz is useless: recent versions of Firefox (at least since v16) enable font-feature-settings by default (same for Apple browsers).
- -webkit is needed for Chrome/Opera even with the latest versions

I've created a demo: http://plnkr.co/edit/CT1pyW?p=preview (http://run.plnkr.co/plunks/CT1pyW/), check the results on BrowserStack: http://www.browserstack.com/screenshots/f0f9268efda5b4b9a58d802fc9061a08035fa324

So the proper code with clean indentation is:
```CSS
-webkit-font-feature-settings: "kern" 1;
        font-feature-settings: "kern" 1;
```

Anyway :+1: specially if it means removing [letter-spacing: -1px](https://github.com/jekyll/jekyll/pull/3379) :)

alfredxing:
@tkrotoff I usually look to MDN for this: https://developer.mozilla.org/en/docs/Web/CSS/font-feature-settings#Browser_compatibility. [Can I Use](http://caniuse.com/#feat=font-feature) lists `-webkit` and `-moz` as well, though it doesn't list `-o`. Do you know of any other resources for finding compatibility tables?

kleinfreund:
What we really need for these kinds of questions is which browser we want to target. /cc @parkr

Firefox has the unprefixed version since 34, so we can drop the `-moz-` prefix as soon as Firefox for Android has it too. Webkit-based browsers need it as well. Also we don't need the prefix for Opera since their latest version is webkit-based as well.

@alfredxing http://caniuse.com is widely regarded the go-to resource I believe. MDN is good as well.

tkrotoff:
@alfredxing
> Can I Use lists -webkit and -moz as well

@kleinfreund
> Firefox has the unprefixed version since 34


1. MDN and CanIUse are nice but not always accurate

2. [Experimentation](http://www.browserstack.com/screenshots/f0f9268efda5b4b9a58d802fc9061a08035fa324) shows that Firefox has font-feature-settings enabled by default since at least v16 on Windows and OS X (same for Apple)

3. Link given earlier by bshackelford: ["it’s worth noting that kerning is turned on by default in Firefox"](http://www.elliotjaystocks.com/blog/kerning/) (2014-04)

=> what the point to discuss about -moz again?

nternetinspired:
> What we really need for these kinds of questions is which browser we want to target.

:100: This would be really helpful for the project I think, a clear statement of what browsers the default template has been designed to be compatible with. In general I think that extending backward compatibility to older browsers  should be something that is left up to the site builder, but there are certain key components like html5 tags and media-queries that we need to be clear about.

It only needs to be a simple statement like: “Latest versions of major browsers and IE9/10.”

jglovier:
>:100: This would be really helpful for the project I think, a clear statement of what browsers the default template has been designed to be compatible with. In general I think that extending backward compatibility to older browsers should be something that is left up to the site builder, but there are certain key components like html5 tags and media-queries that we need to be clear about.

:+1: 

>1. MDN and CanIUse are nice but not always accurate

@tkrotoff curious if you have any examples of such inaccuracy? They've always been reliable in my experience.

---

Personally, I'm of the opinion that a project like this should simply support modern browsers. In the case where we want to do things that even modern browsers require prefixes for, then only should prefixes be used.

tkrotoff:
> any examples of such inaccuracy?

See my comments above (*). MDN and CanIUse don't mention that Firefox and Safari have kerning (`font-feature-settings: "kern" 1`) enabled by default (for quite some time now): there is a lack of precision here.

Also [MDN](https://developer.mozilla.org/en/docs/Web/CSS/font-feature-settings#Browser_compatibility) mentions that Opera >= 18 needs -o prefix for font-feature-settings and [CanIUse](http://caniuse.com/#feat=font-feature) mentions the -webkit prefix for Opera >= 15.

(It does not mean they are bad resources, just that you need to cross-check sometimes - with real tests for example).

> a clear statement of what browsers the default template has been designed to be compatible with

Please open a new issue in order to not pollute this one.

-- 
(*) Do people read them or are they unclear??

kleinfreund:
> See my comments above (*). MDN and CanIUse don't mention that Firefox and Safari have kerning (font-feature-settings: "kern" 1) enabled by default (for quite some time now): there is a lack of precision here.

That is not true. caniuse shows the browser support for certain features, not what user agent styles _might_ yield as a value in that browser. Firefox supports font feature settings from version 15 with prefix, and without a prefix up from version 34.

You put a lot of efforts in to your issues and I believe people are appreciating that, but in this case, I feel a bit of an aggression.

parkr:
Modern browsers. IE10, Safari, Chrome, Opera, Firefox. That said, this is a sample/example site. It's a starter, not meant to be the end product. So let's not get too crazy here.

nternetinspired:
> Modern browsers. IE10, Safari, Chrome, Opera, Firefox.

:+1:

alfredxing:
I think, because of the inconsistencies in prefixes, that I'll just keep all 3 prefixes. It's not going to hurt at all.

@parkr Any other suggestions for this PR?

parkr:
If you could please add `font-kerning: normal;` to the bottom of each and call it a day. :shipit:

jglovier:
:metal: :shipit:

Issue #3372, opened 2015-01-28
`:title` of collections parsed differently than `:title` of posts


beep:
Filed [a bug in jekyll-help about (possibly) inconsistent `:title` parsing between collections and posts](https://github.com/jekyll/jekyll-help/issues/253). @parkr [helpfully noted](https://github.com/jekyll/jekyll-help/issues/253#issuecomment-71773638) this might be a bug:

> Turns out [jekyll] [extracts the slug (used by `:title`) from the post filename](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/post.rb#L170), whereas collection documents' slugs [are generated either from the title element in the YAML front matter or, if it doesn't exist, by the file basename](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/document.rb#L135).
> 
> It would make sense to me to call the slug the input file's basename in all cases (with date data stripped, of course). What say you?

@parkr’s suggestion gets a big +1 from me. (He also asked me to [file the issue here](https://github.com/jekyll/jekyll-help/issues/253#issuecomment-71773638), so, uh, hello all.)

parkr:
Thank you for re-filing, @beep! We'll get this fixed right away. :sparkling_heart:

parkr:
In implementing this, I noticed that it un-does some of #2864, which allows the `title` attribute to override the filename. Hm.

parkr:
Fixed! It'll be a part of Jekyll 3. Thanks, again! :smile:

beep:
Amaaaaaazing. Thank you, @parkr!

PR #3383, opened 2015-01-30
The :title URL placeholder for collections should be the filename slug.

parkr:
This mimicks posts most closely. It can be overridden by the `slug` YAML front matter element.

Undoes some of #2864, but fixes a consistency problem shown in #3372.


gjtorikian:
No, we've since moved off of generating files from `title` and onto relaying on `path`, which made sense (in retrospect).

parkr:
Yay! Ok brillianté. :heart:

PR #3397, opened 2015-02-01
Fixes #3339 Add `pkill -f jekyll` to ways to kill.

### LINKED ISSUE #3339: Killing Detached Server?
Guys,

I might be missing something really obvious, but if you detach, and don't note down the PID in the output
````
Server detached with pid '4311'. Run `kill -9 4311' to stop the server.
````
how do you find and kill the process?

envygeeks:
This shows people how to kill Jekyll without knowing the PID using `pkill` (you could also do `kill -9 $(pgrep -f jekyll)` but that is just the long way around doing `pkill -f` so it shouldn't be shown in the Jekyll logger but can be documented here for people.

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -50,7 +50,7 @@ module Jekyll
           if options['detach'] # detach the server
             pid = Process.fork { s.start }
             Process.detach(pid)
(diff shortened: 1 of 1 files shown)


parkr:
Looks great to me :shipit:. If you merge this PR, please don't forget to add a line in the History.markdown file that references this PR and have that commit reference this PR too :) i have a tool i'll release that does this for you eventually...

PR #3402, opened 2015-02-02
Rake tasks site_template:preview and site_template:generate

tkrotoff:
I have some difficulties to modify and then test `site_template`. I don't want to run `jekyll new my_site_template` each time I experiment with `site_template`.

This pull request is not to be merged. It is to show how I have improved my workflow in order to work on `site_template` (I might doing it wrong).

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -125,6 +125,50 @@ task :console do
   sh "irb -rubygems -r ./lib/#{name}.rb"
 end
 
+
+#############################################################################
+#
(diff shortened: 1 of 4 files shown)


envygeeks:
What is the status/meaning of this?

parkr:
Thanks for this, but we don't require this at this time. Running `jekyll new test-site && cd test-site && jekyll serve` suffices. :smile:

PR #3404, opened 2015-02-03
Simple generic grid instead of footer specific code

tkrotoff:
There is already a concept of "grid" for the footer but the code is very specific and cannot be re-used.

The idea is to make the grid generic by defining `.row` instead of `.footer-col-wrapper` and `.col-*-*` instead of `.footer-col-*`.
The end result is a new file of 44 lines of code (_grid.scss) and _layout.scss reduced by 44 loc => 0 loc added, better semantics and a simple reusable grid.

## Example

Now one can write code like this:
```HTML
<div class="row">
  <div class="col-sm-6 col-md-4">
    <h2>Heading</h2>
    <p>...</p>
  </div>

  <div class="col-sm-6 col-md-4">
    <h2>Heading</h2>
    <p>...</p>
  </div>

  <div class="col-sm-12 col-md-4">
    <h2>Heading</h2>
    <p>...</p>
  </div>
</div>
```

- Result with window > 800px ($on-laptop):

- Result with 600px < window <  800px:

- Result with window < 600px ($on-palm)

## Current footer

As for the footer, the behavior is the same (3 columns then 2 then 1) at the exception of a few pixels: the grid contains 12 columns (customizable via $grid-columns) and cannot express the exact same % width as before.
Before:

After:

People can now adjust the widths based on their footer content by simply modifying the HTML instead of _layout.scss.

## Syntax

The syntax is identical to [Bootstrap grid](http://getbootstrap.com/css/#grid) at the exception of `col-lg-*` (> 1200px) (+ other advanced things) as it does not make sense with a $content-width of 800px. 
Thus file _grid.scss can be replaced by Bootstrap grid, the HTML will behave and look the same.

## Commits

The addition of the grid has been done in 2 commits: 256b8ad and 76b255c so it is easier to understand the mixins from the second commit.

## Testing

I've manually tested the grid by playing with several examples like the one from above under Chrome, Firefox, Safari (iOS 8.1 iPhone & iPad), Android 5 Chrome (Nexus 5), Opera 12, IE11 and IE9. (Under IE8 the columns stack up the way it does for the current footer - IE8 does not support media queries).

## More screenshots

- 1

- 2

- 3

### DIFF
--- a/lib/site_template/_includes/footer.html
+++ b/lib/site_template/_includes/footer.html
@@ -4,15 +4,15 @@
 
     <h2 class="footer-heading">{{ site.title }}</h2>
 
-    <div class="footer-col-wrapper">
-      <div class="footer-col footer-col-1">
+    <div class="footer-content row">
+      <div class="col-sm-6 col-md-4">
         <ul class="contact-list">
           <li>{{ site.title }}</li>
           <li><a href="mailto:{{ site.email }}">{{ site.email }}</a></li>
         </ul>
       </div>
 
-      <div class="footer-col footer-col-2">
+      <div class="col-sm-6 col-md-2">
         <ul class="social-media-list">
           {% if site.github_username %}
           <li>
@@ -45,7 +45,7 @@
         </ul>
       </div>
 
-      <div class="footer-col footer-col-3">
+      <div class="col-sm-12 col-md-6">
         <p class="text">{{ site.description }}</p>
       </div>
     </div>
--- a/lib/site_template/_sass/_base.scss
+++ b/lib/site_template/_sass/_base.scss
@@ -173,7 +173,6 @@ pre {
  * Clearfix
  */
 %clearfix {
-
     &:after {
         content: "";
         display: table;
@@ -187,7 +186,6 @@ pre {
  * Icons
  */
 .icon {
-
     > svg {
         display: inline-block;
         width: 16px;
--- a/lib/site_template/_sass/_grid.scss
+++ b/lib/site_template/_sass/_grid.scss
@@ -3,9 +3,7 @@
     @extend %clearfix;
 }
 
-.col-xs-1, .col-xs-2, .col-xs-3, .col-xs-4, .col-xs-5, .col-xs-6, .col-xs-7, .col-xs-8, .col-xs-9, .col-xs-10, .col-xs-11, .col-xs-12,
-.col-sm-1, .col-sm-2, .col-sm-3, .col-sm-4, .col-sm-5, .col-sm-6, .col-sm-7, .col-sm-8, .col-sm-9, .col-sm-10, .col-sm-11, .col-sm-12,
-.col-md-1, .col-md-2, .col-md-3, .col-md-4, .col-md-5, .col-md-6, .col-md-7, .col-md-8, .col-md-9, .col-md-10, .col-md-11, .col-md-12 {
+[class*="col-"] {
     -webkit-box-sizing: border-box;
             box-sizing: border-box;
 
--- a/lib/site_template/_sass/_grid.scss
+++ b/lib/site_template/_sass/_grid.scss
@@ -9,6 +9,9 @@
     -webkit-box-sizing: border-box;
             box-sizing: border-box;
 
+    // Prevent columns from collapsing when empty
+    min-height: 1px;
+
     padding-left: $spacing-unit / 2;
 }
 
--- a/lib/site_template/_sass/_grid.scss
+++ b/lib/site_template/_sass/_grid.scss
@@ -1,5 +1,7 @@
 .row {
-    margin-left: -$spacing-unit / 2;
+    margin-left: - $spacing-unit / 4;
+    margin-right: - $spacing-unit / 4;
+
     @extend %clearfix;
 }
 
@@ -10,7 +12,9 @@
     // Prevent columns from collapsing when empty
     min-height: 1px;
 
-    padding-left: $spacing-unit / 2;
+    // Space between columns: 15px = 2 * $spacing-unit / 4 = $spacing-unit / 2
+    padding-left: $spacing-unit / 4;
+    padding-right: $spacing-unit / 4;
 }
 
 
--- a/lib/site_template/_sass/_grid.scss
+++ b/lib/site_template/_sass/_grid.scss
@@ -30,8 +30,8 @@ $grid-columns: 12;
 }
 
 
-@include make-grid(xs);
 @include make-grid-columns-float(xs);
+@include make-grid(xs);
 
 @media (min-width: $on-palm) {
     @include make-grid-columns-float(sm);
--- a/lib/site_template/_sass/_grid.scss
+++ b/lib/site_template/_sass/_grid.scss
@@ -1,0 +1,65 @@
+.row {
+    margin-left: -$spacing-unit / 2;
+    @extend %clearfix;
+}
+
+.col-xs-1, .col-xs-2, .col-xs-3, .col-xs-4, .col-xs-5, .col-xs-6, .col-xs-7, .col-xs-8, .col-xs-9, .col-xs-10, .col-xs-11, .col-xs-12,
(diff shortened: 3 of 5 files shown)


kleinfreund:
I'm really not set on the idea to introduce a full grid system into the site template. It's meant to be super simple and the grid in the footer already seems somewhat overkilly. 

Against what every grid system states, I think they're not simple to use for new users.

gmile:
@kleinfreund what I believe @tkrotoff is trying to achieve here is simply standardizing the existing grid html/css code, making it more generic and supportable.

Changeset is `-51/+51` which means no significant amount of code to support is being introduced.

Also, the approach taken by @tkrotoff resembles https://github.com/twbs/bootstrap, which, to me, is always a good thing (I believe people at bootstrap know what they are doing).

:+1: to have this change.

tkrotoff:
To play with the grid, check http://pastebin.com/SQVrtLux (and http://pastebin.com/i4UDBnYs): it features a post that contains [the supported examples from Bootstrap documentation](http://getbootstrap.com/css/#grid) + adds `border: 2px dotted black` to `.col-*-*` in order to see what's going on.

And here the diff to replace _grid.scss by Bootstrap grid (SASS): http://pastebin.com/hzYTNZW6

> Changeset is -51/+51

Overall it adds 4 lines of code :)

> It's meant to be super simple

The grid system is 45 lines of code and does not have features like `.col-lg-*`, customizable nb of columns, `col-*-offset-*`, `col-*-push-*`, `.col-*-pull-*`...
Bootstrap grid system in SASS takes 313 loc (not bad considering the features, the number of comments and whitespaces - see the diff).

nternetinspired:
I do personally feel that introducing a generic, reusable, grid system is probably a good idea in general. I also thinks that grids are one area where it's preferable to place utility classes in html than extend a layout from the css. 

That said, I wouldn't hold up TWBS's grid as a good example of a grid system. It's over-specified, inflexible and far too limiting. Crucially, as @kleinfreund correctly points out, it is far from simple for new users and it requires considerable prior knowledge to use.

A class like `.col-sm-6` is utterly meaningless if you don't know it is part of a 12 column grid. What if I need a 5 column grid or a 7 column grid?

tkrotoff:
> I wouldn't hold up TWBS's grid as a good example of a grid system [...] far from simple for new users [...] What if I need a 5 column grid or a 7 column grid?

And the solution is...?

nternetinspired:
> And the solution is...?

The solution I prefer is an unambiguous syntax based on logical fractions. Human readable and understandable by anyone with elementary maths. An over-specified example to illustrate;

    <div class="grid">
	    <div class="grid-item one-half md-two-thirds lg-five-eighths">
		
	    </div>
	    <div class="grid-item one-half lg-one-third lg-three-eighths">
		
	    </div>
    </div>

Note that such an approach also separates grid logic from width logic.

nternetinspired:
> Changeset is -51/+51 which means no significant amount of code to support is being introduced.

I think it is also worth considering the compiled output.

tkrotoff:
@nternetinspired the code is [from you](https://github.com/nternetinspired/oneweb-sass/blob/master/_scss/patterns/_grid.scss). You use fractions like Pure.css with a different syntax.
1 code snippet is not enough to demonstrate "screen size proportions" grids are significantly better than "number of columns to cover" grids. It is a matter of taste.

It appears Bootstrap is by far the most popular framework/grid and is very well done.
That's why instead of re-inventing the wheel/a new syntax, I've used it.
I believe it is a pragmatic choice: site_template goal is not to innovate on the grid framework market.

The day a better grid or flexbox gets popular, I'm sure someone will write a pull request. Meanwhile this pull request is an enhancement over current situation: moving one step at a time.

I've compiled a list of the most popular grids: https://gist.github.com/tkrotoff/e15da98129e46db39bf6
Except Pure.css (fractions) and Ionic (flexbox), they all follow Bootstrap principle ("number of columns to cover" with a 12 columns grid).
Sometimes with differences: not responsive, no classes for small/medium screens, different syntax, different number of columns...

> compiled output

128 loc, see http://pastebin.com/vgqqvjc8

nternetinspired:
> It appears Bootstrap is by far the most popular framework/grid…

MacDonalds is by far the most popular restaurant.

> Meanwhile this pull request is an enhancement over current situation: moving one step at a time.

Sorry, I also disagree with you there. You replace readable, approachable, css with largely unfathomable mixins. I also think that if you compare compiled css output then this PR will not look so much like an enhancement.

Of course it is entirely for the maintainers to decide, but I believe the principle here is for a lightweight, basic, implementation, one that carries no more code than is required for this *demonstrative* implementation and presents a low barrier to entry.

> the code is from you.

Your point is? 

> 1 code snippet is not enough to demonstrate "screen size proportions" grids are significantly better than "number of columns to cover" grids.

I wasn't trying to demonstrate any such thing, though I did explain my reasoning behind such choices; approachable code with minimal assumptions. The basis of the grid is identical (both are proportional grids), I simply prefer semantic selectors. FWIW, a complete example of my own currently preferred implementation would be something like:

https://gist.github.com/nternetinspired/1d6469dd7527ed7693ba

In practice I usually comment out width ranges depending on my choice of grid, three-column, five-column, etc., so I don't output any more css than is required. Please note; I'm not suggesting we add this as it's overkill for the default template.

tkrotoff:
> [Bootstrap is by far the most popular framework/grid] McDonalds is by far the most popular restaurant

Comparing totally different markets and audiences is a difficult task (and an easy troll). Do you suggest Bootstrap is terrible and its success not deserved?

> You replace readable, approachable css with largely unfathomable mixins
> [...] the principle here is for a lightweight, basic, implementation

Meaning the code is fat, complex and unreadable because of the 2 loops of 3 lines each. Your wording is a bit strong. We can have [this](https://github.com/jekyll/jekyll/blob/256b8ad887e86fdbecd69f17ea2174c63a5630b0/lib/site_template/_sass/_grid.scss) instead (and lose $grid-columns variable).

> [The code is from you] Your point is?

You didn't cite your source, I did for you. It also means that you are not neutral: we always tend to prefer our own syntax (~NIH syndrome).

kleinfreund:
> Do you suggest Bootstrap is terrible and its success not deserved?

I believe he wanted to point out that a big user base doesn't necessarily come hand in hand with being flawless. It's also not the best strategy comparing Jekyll with Bootstrap here, because they are two different things.

Apart from that, the discussion here got quite out of hand. For the topics issue, I personally would favor a template without a grid system at all.

nternetinspired:
> Comparing totally different markets and audiences is a difficult task (and an easy troll). Do you suggest Bootstrap is terrible and its success not deserved?

I've used TWBS (Sass flavour) on plenty of projects, and I've eaten plenty of McD's. However, I'm not under any illusion that either is ‘the best’; though admittedly both have their place in the world.

> Meaning the code is fat, complex and unreadable because of the 2 loops of 3 lines…

Yes, to a newcomer who may not be familiar with either Sass or TWBS *that is complex code*; far more so than the code it replaces. This raises the bar to entry.

> You didn't cite your source, I did for you. It also means that you are not neutral: we always tend to prefer our own syntax.

I did state that it was my personal preference and I have never suggested the ‘source’ you cite for me be added to the default template.

> Apart from that, the discussion here got quite out of hand. For the topics issue, I personally would favor a template without a grid system at all.

I agree on both counts.

tkrotoff:
> It's not the best strategy comparing Jekyll with Bootstrap here, because they are two different things

That's why no such thing happened.

> I personally would favor a template without a grid system at all

If `row` and `col-*-*` classes is not a good idea for site_template, you will then agree that the existing ones `footer-col-wrapper`, `footer-col` and `footer-col-*` should be removed.

@nternetinspired 
> I agree on both counts

I though you liked the idea (you wrote *"introducing a generic, reusable, grid system is probably a good idea"*) but didn't like the syntax (because you prefer another) and implementation (because there are 2 loops).

parkr:
We'll go for simplicity here and leave it as-is, but thank you!

PR #3415, opened 2015-02-06
Show only two decimals in time to generate pages

eksperimental:
no need to display 10 decimals.
it fills up my screen with useless numbers! ;-p

### DIFF
--- a/lib/jekyll/commands/build.rb
+++ b/lib/jekyll/commands/build.rb
@@ -57,7 +57,7 @@ module Jekyll
           Jekyll.logger.info "Incremental build:", (full_build ? "disabled" : "enabled")
           Jekyll.logger.info "Generating..."
           process_site(site)
-          Jekyll.logger.info "", "done in #{Time.now - t} seconds."
+          Jekyll.logger.info "", "done in #{(Time.now - t).round 2} seconds."
         end
 
         # Private: Watch for file changes and rebuild the site.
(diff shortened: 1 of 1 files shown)


lib/jekyll/commands/build.rb
```diff
@@ -57,7 +57,7 @@ def build(site, options)
           Jekyll.logger.info "Incremental build:", (full_build ? "disabled" : "enabled")
           Jekyll.logger.info "Generating..."
           process_site(site)
-          Jekyll.logger.info "", "done in #{Time.now - t} seconds."
+          Jekyll.logger.info "", "done in #{(Time.now - t).round 2} seconds."
```
parkr: Is there a possible world which shows `0.00` seconds? Maybe 3?

Also, from a code-style standpoint, I'd prefer parentheses around the argument to `#round` :smile:

parkr:
Wonderful!

alfredxing:
:+1:

lib/jekyll/commands/build.rb
eksperimental: in the case of your possible world. then we need to use ceil(), so it will be 0.01  :-D

i thought no parenthesis was more aesthetic.
anyway.. fixed

parkr:
Moved it to 3 decimal places in 66f2a50249c5e2ebd05bac77c823c98d549830c7

parkr:
Weird git thing happened. This was merged:

- d795113e94003cab3144ea431068da0f57b4b3af
- dbc853006893f59311b1886c5f20ef70e31b4f22

Issue #3417, opened 2015-02-06
jekyll help build don't show some options


bogdan:
Current output of this command gives:

```
$ jekyll help build
jekyll build -- Build your site

Usage:

  jekyll build [options]

Options:
            --config CONFIG_FILE[,CONFIG_FILE2,...]  Custom configuration file
            --future       Publishes posts with a future date
            --limit_posts MAX_POSTS  Limits the number of posts to parse and publish
        -w, --[no-]watch   Watch for changes and rebuild
            --force_polling  Force watch to use polling
            --lsi          Use LSI for improved related posts
        -D, --drafts       Render posts in the _drafts folder
            --unpublished  Render posts that were marked as unpublished
        -q, --quiet        Silence output.
        -V, --verbose      Print verbose output.
```

That doesn't include `--destination --source` options that can only be found here: http://jekyllrb.com/docs/usage/

envygeeks:
Should be fixed with #3418

PR #3418, opened 2015-02-06
Added build --destination and --source flags

tarebyte:
From issue #3417

### DIFF
--- a/lib/jekyll/command.rb
+++ b/lib/jekyll/command.rb
@@ -49,8 +49,8 @@ module Jekyll
       # Returns nothing
       def add_build_options(c)
         c.option 'config',  '--config CONFIG_FILE[,CONFIG_FILE2,...]', Array, 'Custom configuration file'
-        c.option 'destination', '--destination <destination>', 'The current folder will be generated into <destination>'
-        c.option 'source', '--source <source> --destination <destination>', 'The <source> folder will be generated into <destination>'
+        c.option 'destination', '-d', '--destination <destination>', 'The current folder will be generated into <destination>'
+        c.option 'source', '-s', '--source <source> --destination <destination>', 'The <source> folder will be generated into <destination>'
         c.option 'future',  '--future', 'Publishes posts with a future date'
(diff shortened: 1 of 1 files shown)


lib/jekyll/command.rb
```diff
@@ -49,6 +49,8 @@ def configuration_from_options(options)
       # Returns nothing
       def add_build_options(c)
         c.option 'config',  '--config CONFIG_FILE[,CONFIG_FILE2,...]', Array, 'Custom configuration file'
+        c.option 'destination', '--destination <destination>', 'The current folder will be generated into <destination>'
+        c.option 'source', '--source <source> --destination <destination>', 'The <source> folder will be generated into <destination>'
```
parkr: ?

parkr:
Please also add the short versions, `-s` and `-d`.

lib/jekyll/command.rb
tarebyte: Is this not the correct place for these flags? Or are the flags not correct?

lib/jekyll/command.rb
```diff
@@ -49,6 +49,8 @@ def configuration_from_options(options)
       # Returns nothing
       def add_build_options(c)
         c.option 'config',  '--config CONFIG_FILE[,CONFIG_FILE2,...]', Array, 'Custom configuration file'
+        c.option 'destination', '-d', '--destination <destination>', 'The current folder will be generated into <destination>'
+        c.option 'source', '-s', '--source <source> --destination <destination>', 'The <source> folder will be generated into <destination>'
```
parkr: You can only have one switch per line, you cannot mix them.

parkr:
Please also use `SCREAMING_SNAKE_CASE` instead of `<brackets>` for the argument fillers.

tarebyte:
@parkr should be fixed now, thanks for the feedback!

tarebyte:
@parkr does this need anything else added to it? I think I caught everything you mentioned.

parkr:
These are redundant because they're global: https://github.com/jekyll/jekyll/blob/master/bin/jekyll#L22-L26

They should be removed from being global if we want to do this.

PR #3420, opened 2015-02-07
fix missing char in template docs

nicwest:
`The are as follows (with what they filter):`

should read:

`They are as follows (with what they filter):`


alfredxing:
Nice catch, thanks!

Issue #3421, opened 2015-02-07
Option draft is an alias for drafts


penibelst:
Jekyll 2.5.3 here. I noticed the command line option `--draft` without “s” works the same way as `--drafts`. Is this an intended behaviour? If not, it would be nice to deprecate the undocumented feature in the upcoming major release.

```shell
jekyll build --drafts
# same output
jekyll build --draft
```

parkr:
What output did you see for both? It should just be [`--drafts`](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/command.rb#L57).

penibelst:
Both render the drafts. No error is thrown.

zebron:
How do I get off the email thread?

pathawks:
> How do I get off the email thread?

Stop watching Jekyll?

parkr:
I have been all through Mercenary and OptionParser – it looks like it's a feature of OptionParser. Try this out:

```ruby
require 'optparse'

p ARGV
options = {}

optparser = OptionParser.new do |opts|
  opts.on('--drafts') do
    options['show_drafts'] = true
  end
end

p optparser.parse!
p options
```

It works with `-d`, `--draft`, and `--drafts`. We'd have to catch `--draft` ourselves, which I'm not really interested in at this point.

PR #3423, opened 2015-02-07
Add a test for wrong draft option

penibelst:
Showcase for #3421

### DIFF
--- a/features/drafts.feature
+++ b/features/drafts.feature
@@ -13,6 +13,16 @@ Feature: Draft Posts
     Then the _site directory should exist
     And I should see "Not baked yet." in "_site/recipe.html"
 
+  Scenario: Don't preview a draft with wrong option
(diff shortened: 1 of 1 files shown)


penibelst:
The test fails https://travis-ci.org/jekyll/jekyll/jobs/49900079#L283 This is a bug.

parkr:
I have been all through Mercenary and OptionParser – it looks like it's a feature of OptionParser. Try this out:

```ruby
require 'optparse'

p ARGV
options = {}

optparser = OptionParser.new do |opts|
  opts.on('--drafts') do
    options['show_drafts'] = true
  end
end

p optparser.parse!
p options
```

It works with `-d`, `--draft`, and `--drafts`. We'd have to catch `--draft` ourselves.

penibelst:
I’m surprised by the design choice of Option Parser.

PR #3426, opened 2015-02-08
Release jekyllrb.com as a locally-compiled site.

parkr:
In an effort to :dog2: food our features, let's compile jekyllrb.com with jekyll/jekyll HEAD rather than with Jekyll 2.4.0 or whatever GitHub Pages is using. If we're actively using the tool to build our documentation, we'll encounter the pain points and bugs and (hopefully) fix those before HEAD is released into the wild.

/cc @jekyll/core

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -203,26 +203,21 @@ namespace :site do
     end
 
     # Copy site to gh-pages dir.
-    puts "Copying site to gh-pages branch..."
-    copy_exclude = %w[
-      site/.
-      site/..
-      site/.jekyll-metadata
-      site/.sass-cache
-      site/_site
-    ]
-    FileList["site/{*,.*}"].exclude(*copy_exclude).each do |path|
-      sh "cp -R #{path} gh-pages/"
-    end
+    puts "Building site into gh-pages branch..."
+    ENV['JEKYLL_ENV'] = 'production'
+    require "jekyll"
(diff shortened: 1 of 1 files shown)


envygeeks:
Is there any reason to dog food pre-release versions? Dog fooding is what we already did now all we do is turn what should be stable documentation into something that could go down or end up broken which isn't such a good idea IMO.

parkr:
> Dog fooding is what we already did now

We didn't really; we had to wait for GHP to upgrade. I want to be able to use brand new features in Jekyll without waiting for GitHub's Security Team to give the green light.

> Is there any reason to dog food pre-release versions?

Easier to find and solve bugs when you hit them yourself. Also, we get to try out new features and can improve based on our own experiences. I really don't see the site going down unless I fuck up the Rakefile task.

envygeeks:
Using the case of solving bugs by hitting them yourself doesn't seem valid for me, because the "works for me" theory could and probably will become more prevalent even if people don't realize it and using a pre-release version on Jekyll's own website won't even cover a quarter of the cases unless we start mimicking what people do.

On the latter part, I don't really know what you mean since you should probably be using something like Codeship to pass or fail a deploy (or TravisCI) and the Rakefile should really have nothing to do with that IMO (unless you are using the Rakefile to deploy.)

Issue #2607, opened 2014-07-16
Invalid highlight html with rouge and linenos


kostrse:
Invalid HTML generated when highlighter set to *rouge* and `highlight` has option `linenos`.

For *kramdown* and *redcarpet* html is slightly different, but in both cases invalid.

**Original markdown:**

```
before

{% highlight cs linenos %}
abc
def
{% endhighlight %}

after
```
**Resulting html for redcarpet:**

``` html
<p>before</p>
<div class="highlight">
	<pre>
		<code class="language-cs" data-lang="cs">
			<table style="border-spacing: 0">
				<tbody>
					<tr>
						<td class="gutter gl" style="text-align: right">
							<pre class="lineno">
							    1
							    2
		</code>
							</pre>
						</td>
						<td class="code">
							<pre>
								<span class="n">abc</span>
								<span class="n">def</span>
								<span class="w"></span>
							</pre>
						</td>
					</tr>
				</tbody>
			</table>
	</pre>
</div>
<p>after</p>
```

Problem: *code* closed inside the table.

**Resulting html for kramdown:**

``` html
<p>before</p>
<div class="highlight">
	<pre>
		<code class="language-cs" data-lang="cs">
			<table style="border-spacing: 0">
				<tbody>
					<tr>
						<td class="gutter gl" style="text-align: right">
							<pre class="lineno">
							1
							2
							</pre>
						</td>
						<td class="code">
							<pre>
								<span class="n">abc</span>
								<span class="n">def</span>
								<span class="w"></span>
							</pre>
						</td>
					</tr>
				</tbody>
			</table>
			after
		</code>
	</pre>
</div>
```

Problem: "after" inside highlighter *div* block, should be inside *p* after the block.

**Reproduced with:**

```
jekyll 2.1.1 and 2.2.0 (@phocks)
rouge 1.5.1
```

envygeeks:
It seems this was fixed with #3435 is this the case?

aschamberger:
It's fixed.

adilinden:
Using markdown and rouge.

OS X 10.11.3
      jekyll (= 3.0.3)
      kramdown (= 1.10.0)
      rouge (= 1.10.1)

This is the markdown:

```md
{% highlight coffeescript linenos %}
# Objects:
math =
  root:   Math.sqrt
  square: square
  cube:   (x) -> x * square x
{% endhighlight %}
```

This is the HTML generated (I replaced actual line digits and code content with comments).

```html
[...]

PR #3435, opened 2015-02-09
Fix: Invalid highlight html with rouge and linenos #2607

EdMcBane:
Fixes broken HTML being generated when using rouge highlighter with linenos feature enabled.

Caused by a naive regex replace done on rouge output to add `<code></code>` tags within `<pre></pre>` tags, that assumes there will be just one such `<pre>` block. Rouge actually generates two when linenos is enabled, one for the line numbers and one for the actual code block.

Regex replaces the first `<pre>` tag with `<pre><code>` and the **first** `</pre>` tag with `</code></pre>`. Should be the **last**.

Fixes issue [#2067](https://github.com/jekyll/jekyll/issues/2607)

## Todo
* The fix is more of a workaround, since regex-replacing tags within third-party library output should definitely be avoided, but a major change would be required and this needs input from project mantainers
* The test for highlighting with rouge has been created as a separate file, may be merged into TestTags
* The way the test themselves are written could be improved, in this case I kept it similar to what was already there.

### DIFF
--- a/lib/jekyll/tags/highlight.rb
+++ b/lib/jekyll/tags/highlight.rb
@@ -111,10 +111,21 @@ eos
         "<div class=\"highlight\"><pre>#{h(code).strip}</pre></div>"
       end
 
+      def string_rsub(str, pattern, replacement)
+        last_match = nil
+        while true
+          pos = last_match.nil? ? 0 : last_match.end(0)
+          m = pattern.match(str, pos)
+          break if m.nil?
+          last_match = m
+        end
+        last_match ? (last_match.pre_match + replacement + last_match.post_match) : str
+      end
+ 
       def add_code_tag(code)
         # Add nested <code> tags to code blocks
         code = code.sub(/<pre>\n*/,'<pre><code class="language-' + @lang.to_s.gsub("+", "-") + '" data-lang="' + @lang.to_s + '">')
-        code = code.sub(/\n*<\/pre>/,"</code></pre>")
+        code = string_rsub(code, /\n*<\/pre>/,"</code></pre>")
         code.strip
       end
 
--- a/test/test_rouge.rb
+++ b/test/test_rouge.rb
@@ -1,0 +1,58 @@
+# coding: utf-8
+
+require 'helper'
+
+class TestRouge < Test::Unit::TestCase
+
+  def create_post(content, override = {}, converter_class = Jekyll::Converters::Markdown)
+    stub(Jekyll).configuration do
+      site_configuration({
+        "highlighter" => "rouge"
+      }.merge(override))
+    end
+    site = Site.new(Jekyll.configuration)
+
+    if override['read_posts']
+      site.read_posts('')
+    end
+
+    info = { :filters => [Jekyll::Filters], :registers => { :site => site } }
+    @converter = site.converters.find { |c| c.class == converter_class }
+    payload = { "highlighter_prefix" => @converter.highlighter_prefix,
+                "highlighter_suffix" => @converter.highlighter_suffix }
+
+    @result = Liquid::Template.parse(content).render!(payload, info)
+    @result = @converter.convert(@result)
+  end
+
+  def fill_post(code, override = {})
+    content = <<CONTENT
+{% highlight text %}
+#{code}
+{% endhighlight %}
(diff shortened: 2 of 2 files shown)


kleinfreund:
> The fix is more of a workaround, since regex-replacing tags within third-party library output should definitely be avoided

If this is an issue with how the Rouge highlighter is generating HTML, why is this not brought up over at https://github.com/jneen/rouge then?

EdMcBane:
It is actually an issue with how Jekyll does regex-replace on the output of rouge: rouge output is fine, it is Jekyll that is adding tags with a regex replace operation, and assuming that rouge will generate a single <pre> block.

There is not a clear interface between Jekyll and the highlighters. Pygments and rouge generate HTML, and Jekyll does regex-replace on it to inject a <code> block inside it.

IMHO the proper solution would be to "negotiate" with Pygments and rouge the structure of the output as an HTML document structure, and either:
- wrap the output inside <code> tags, if applicable
- or add parameters to pygments/rouge to include the <code> tags that Jekyll needs in the proper place

If this "negotiation" is not an option, it should be Jekyll responsibility to properly deal with whatever output both pygments and rouge generate, possibly avoiding fragile regex-replaces.

parkr:
> it should be Jekyll responsibility to properly deal with whatever output both pygments and rouge generate, possibly avoiding fragile regex-replaces.

I agree here – any changes we make to the output are our bugs to handle. 

As a *super hacky* but kind of clever way of handling this, StackOverflow suggests:

```ruby
new_str = old_str.reverse.sub(pattern.reverse, replacement.reverse).reverse
```

I don't think we'll be going with that, but it was too funny not to share.

lib/jekyll/tags/highlight.rb
```diff
@@ -111,10 +111,21 @@ def render_codehighlighter(code)
...
+          last_match = m
+        end
+        last_match ? (last_match.pre_match + replacement + last_match.post_match) : str
+      end
+ 
       def add_code_tag(code)
         # Add nested <code> tags to code blocks
         code = code.sub(/<pre>\n*/,'<pre><code class="language-' + @lang.to_s.gsub("+", "-") + '" data-lang="' + @lang.to_s + '">')
-        code = code.sub(/\n*<\/pre>/,"</code></pre>")
+        code = string_rsub(code, /\n*<\/pre>/,"</code></pre>")
```
parkr: @jneen We're trying to insert a `<code>` layer inside Rouge's your `<pre>`. This is a pretty crummy way of handling this. Would you consider adding support for another encasing tag (in our case, `code` with custom attributes)?

jneen: Hi @parkr, I'm working on a major refactor of this.  Likely what you'll want to do for rouge 1.x is pass `:wrap => false` to disable all wrapping, and provide the html context manually.

jneen: (`:wrap => false` is an option to Formatters::HTML that causes it to output just the highlighted spans and nothing else.)

parkr: @jneen Wonderful to hear. I'm sorry to have bothered you – it turns out we're wrapping it manually (i.e. Rouge isn't wrapping it) in order to maintain similarity to Pygments. I'm going to play around with `nowrap` and pygments and see what happens.

parkr:
Fixed in #3436, thank you for figuring this bug out!!

kusera:
I'm still experiencing this issue on Jekyll 2.5.3.

Generated tags:
tags

Source:
markdown

I'm building off an offshoot of [Type Theme](https://github.com/kusera/portfolio-iro) if that makes any difference.

jneen:
Are yall still using regexen on the output of rouge?

parkr:
We don't seem to be doing anything weird with rouge: https://github.com/jekyll/jekyll/blob/master/lib/jekyll/tags/highlight.rb

kusera:
It's strange, I see that you're using `<figure>` tags so I can't imagine why I'm getting `<div>` when serving the site with 2.5.3.

parkr:
It's change a little since 2.5.3: https://github.com/jekyll/jekyll/blob/v2.5.3/lib/jekyll/tags/highlight.rb#L107

PR #3436, opened 2015-02-10
highlight: fix problem with linenos and rouge.

parkr:
Found by @EdMcBane in https://github.com/jekyll/jekyll/pull/3435

The strange regexp we were doing to replace the `<pre><code></pre></code>`
bits in the Pygments output were wreaking havoc on Rouge output
because Rouge uses `<pre>`'s to wrap line numbers.

To be consistent, the output from `render_*` should *not* include
the wrapping `<div>` and `<pre>` tags. It should just be what was
inside. We can then wrap it in our own custom tags without using
any regular expressions, as God intended. Death to regular
expressions and HTML manipulation!

### DIFF
--- a/lib/jekyll/tags/highlight.rb
+++ b/lib/jekyll/tags/highlight.rb
@@ -77,8 +77,6 @@ eos
       def render_pygments(code, is_safe)
         Jekyll::External.require_with_graceful_fail('pygments')
 
-        @options[:encoding] = 'utf-8'
-
         highlighted_code = Pygments.highlight(
           code,
           :lexer   => @lang,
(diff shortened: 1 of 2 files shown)


pathawks:
This broke my site.

I had a plugin that was overriding `add_code_tag` to replace `<div class="highlight">` with some custom tags. Now that `<div class="highlight">` isn't in `code`, `add_code_tag` needs to be sure to add it.

Overall, I think this is much better than how it was being done in the past, but some documentation would be nice. I was pulling my hair out trying to track down the problem.

parkr:
@pathawks This probably isn't a huge concern as I can't imagine lots of people rewriting/reworking the code highlighting. I'm glad you (seemingly quickly) figured out what was wrong and fixed it. I don't think this change is appropriate for the Upgrading doc, but it's in the History 8-)

If you write a blog post about it, Google will probably put you front and center for anyone else having this problem :)

envygeeks:
Good thing this is beta not stable :smirk_cat:

pathawks:
> If you write a blog post about it, Google will probably put you front and center for anyone else having this problem :)

:+1:

> Good thing this is beta not stable

Yup, that's why I'm testing and reporting things that break :smiley:

Issue #2021, opened 2014-02-07
Remove html extension from pages on test server.


cirosantilli:
On Github Pages it already works: I can create a page with path `/pagetitle.html`, and use URL `/pagetitle` (without `.html`).

The problem is that to do that I have to setup all my links without the `.html`, which breaks the links on the test server.

For posts, this is already possible with `permalink: /:title`.

Is there any way to do it for pages also?

Workarounds:
- use posts instead of pages. Not satisfactory because:
  - it requires you to use an unwanted date prefix #484 , which is not going to change
  - it forces all your posts to be of `/:title` format.
- put the file in a folder with the desired name and rename it as `index.html`.
  
  Too much boilerplate.
- use Apache and play with `.htaccess`.
  
  Requires to install and setup Apache...

Related discussions:
- https://github.com/jekyll/jekyll/issues/156: seems to focus on Posts, not Pages as this issue, and was closed after the `permalink: /:title` solution came up, but this does not yet work for pages.
- https://github.com/holman/feedback/issues/231: stuck with the index workaround

Sayan98:
@cirosantilli See this [page](https://byparker.com/blog/2014/clearing-up-confusion-around-baseurl/)
and [this page for a bit of fun](http://sayan98.github.io/intellisults/)

Issue #156, opened 2010-04-17
Generating URLs without .html extensions


pedrocr:
I don't like having html extensions in urls. For a post named 2010-02-27-first-post.markdown I want to get:

URL: /2010/02/27/first-post
/2010/02/27/first-post.html

mod_rewrite or MultiViews takes care of serving that. By setting the permalink format to "pretty" or "/:categories/:year/:month/:day/:title" jekyll does:

URL: /2010/02/27/first-post
/2010/02/27/first-post/index.html

So I added a configuration option called no_post_dirs to get the first behavior. I did a branch (with test case) for it:

http://github.com/pedrocr/jekyll/tree/no_post_dirs

or more specifically the commit:

http://github.com/pedrocr/jekyll/commit/24b02d92766890956d7579d12ac488867d21d6c7

Cheers,

Pedro

PR #3452, opened 2015-02-13
Mimic GitHub Pages .html extension stripping behavior

rtomayko:
This modifies the WEBrick server started by `jekyll serve` such that a file "/foo.html" can be requested at "/foo". The behavior mirrors the GitHub Pages server configuration closely: first check for a file in the original location and then fallback to checking for file + `.html` extension.

The response's Content-Type ("text/html"), Content-Length, ETag, and Last-Modified headers are all set appropriately, as if the request was made with the `.html` suffix.

See the individual commits for additional remarks on the implementation.

/cc https://github.com/jekyll/jekyll/issues/156, https://github.com/jekyll/jekyll/issues/2021, https://github.com/holman/feedback/issues/231

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -40,7 +40,7 @@ module Jekyll
 
           s.mount(
             options['baseurl'],
-            FileHandler,
+            custom_file_handler,
             destination,
             file_handler_options
           )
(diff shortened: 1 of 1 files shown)


mattr-:
:sparkles: Nice! Thanks @rtomayko!

@parkr?

parkr:
Looks great, thanks Ryan. We'll release this in Jekyll 3.0. :+1:

rtomayko:
Beautiful, thanks!

PR #3456, opened 2015-02-14
add draft? method to identify if Post is a draft

willnorris:
This is useful in development scenarios where you are building with `--drafts`, and want to identify which posts are the drafts.  As an example, I have a [`/drafts`](https://github.com/willnorris/willnorris.com/blob/master/src/drafts.html) page that simply lists all of my current draft posts.

### DIFF
--- a/lib/jekyll/post.rb
+++ b/lib/jekyll/post.rb
@@ -24,6 +24,7 @@ module Jekyll
       content
       excerpt
       excerpt_separator
+      draft?
     ]
 
     # Post name validator. Post filenames must be like:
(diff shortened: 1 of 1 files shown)


parkr:
Hey, thanks for the PR. Not a bad idea at all :+1: A little concerned about the `?`.

@fw42 & @pushrax, is `thing.draft?` recommended/allowed syntax in Liquid 3 strict mode?

willnorris:
yeah, fair point.  I actually originally had it as `is_draft`, but thought that `draft?` was more idiomatic ruby.  I don't have strong feeling either way.

pushrax:
`?` is allowed at the end of identifiers.

PR #3464, opened 2015-02-15
write .jekyll-metadata even on full_rebuild

willnorris:
For a full rebuild, we certainly don't want to *read* from `.jeykll-metadata`, but we should still write it.  Otherwise, a subsequent incremental build would have to do a full rebuild again since there is no metadata file to start from.

/cc @alfredxing @mattr- to confirm


parkr:
:+1:

PR #3467, opened 2015-02-15
Clarify pagination works from within HTML files

jaybe-jekyll:
IRC question of confusion re: wording and meaning.

### DIFF
--- a/site/_docs/pagination.md
+++ b/site/_docs/pagination.md
@@ -13,8 +13,8 @@ folders you need for paginated listings.
   <h5>Pagination only works within HTML files</h5>
   <p>
     Pagination does not work from within Markdown or Textile files in your Jekyll site.
-    It will only work when called from within HTML files. For example, if you wanted
-    a paginated list of Posts, you would do so from within an HTML file, such as index.html.
+    It will only work when called from within HTML files; and at this time, only index.html.
+    For example, if you wanted a paginated list of Posts, you would do so from within the HTML file, index.html.
   </p>
 </div>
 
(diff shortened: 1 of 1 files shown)


site/_docs/pagination.md
```diff
@@ -12,9 +12,9 @@ folders you need for paginated listings.
 <div class="note info">
   <h5>Pagination only works within HTML files</h5>
   <p>
-    Pagination does not work with Markdown or Textile files in your Jekyll site.
-    It will only work when used within HTML files. Since you’ll likely be using
-    this for the list of Posts, this shouldn’t be an issue.
+    Pagination does not work from within Markdown or Textile files in your Jekyll site.
+    It will only work when called from within HTML files. For example, if you wanted
+    a paginated list of Posts, you would do so from within an HTML file, such as index.html.
```
parkr: At the moment, we actually *only* allow index.html, so this note is a little too forgiving.

parkr:
Thanks!!

site/_docs/pagination.md
jaybe-jekyll: That's a rather important detail and likely one that would result in hair being pulled out if trying to go about it with anything other than index.html. I'll adapt it slightly.

Does v3 change this behavior? Thanks, Parker!

parkr: v3 has no changes for pagination, as pagination is a separate plugin.

jaybe-jekyll: Thanks; I have updated the language; please review and merge if acceptable; thanks!

parkr:
Great! Maybe mention that `index.html` can be within subdirectories if you change your `paginate_path`?

PR #3469, opened 2015-02-15
docs: Add frontmatter date formatting information

jedd-ahyoung:
I saw [this question](http://stackoverflow.com/q/18362317/677526) on StackOverflow, and realized that having the answer in the documentation would be beneficial.


parkr:
It's actually a bit more flexible. We use `Time.parse` under the hood so it can be as little as the YYYY-MM-DD or the whole string you give. There are other formats as well! But this is a good rule of thumb format. :+1:

BM5k:
:+1:

parkr:
Slight technicalities, but let's get this up there and address those later.

envygeeks:
:heart:

PR #3305, opened 2015-01-14
Update Minitest, Allow multiple versions of ActiveSupport.

envygeeks:
TODO:

- [ ] Use describe.
- [x] Make sure cucumber still works.
- [x] Just use Minitest and rely always on latest.
- [ ] Remove ActiveSupport legacy dependency from tests.
- [ ] Fix an edge that Minitest discovered with collections and disappearing files.

PR #3472, opened 2015-02-16
explore removing activesupport

Ch4s3:
In ref to https://github.com/jekyll/jekyll/pull/3305 I pulled out activesupport and nothing blew up. I can't seem to find where any of activesupport is actually used.

*edit So it looks like this causes an issue with Rubies < 2.2. I'm investigating.


parkr:
Yeah I'd be down with this.

Ch4s3:
Cool. Yeah, activesupport is really weighty if it isn't needed. It seems like the failures for Ruby < 2.2 have something to do with minitest, but I haven't figured it out yet.

parkr:
@envygeeks Was looking into this.

Ch4s3:
Yeah I saw that on #3305, but wasn't sure if that issue was getting narrowed down or not.

envygeeks:
The errors on 2.1+ are because Minitest is no longer a part of StdLib so you'll need to add it to the Gemfile and then do some work on getting it done, I've not had any time to work on anything Jekyll the last few weeks since we have a new production build going out for a client but this weekend I'll see what I can do (if I can, it might be next weekend) about getting the rest of Minitest done so you can rebase this on that but as far as I remember I had already stripped out ActiveSupport entirely from everything Jekyll.

envygeeks:
Attaching this to this weekends work for me.  I'll merge this when I'm ready to deal with the random failing test and we can go from there.

PR #3475, opened 2015-02-17
Fix wording in code snippet highlighting section

markphelps:
Initial wording was kind of clunky and hard to read. Also there was an unnecessary semi-colon.

### DIFF
--- a/site/_docs/templates.md
+++ b/site/_docs/templates.md
@@ -324,9 +324,9 @@ on your system and set `highlighter` to `pygments` in your site's configuration
 file.
 
 Alternatively, you can use [Rouge](https://github.com/jayferd/rouge) to highlight
-your code snippets. It doesn't support as many languages as Pygments does but
-it should fit in most cases and it's written in pure Ruby ; you don't need Python
-on your system!
+your code snippets. It doesn't support as many languages as Pygments, however it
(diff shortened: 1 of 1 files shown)


mattr-:
LGTM! Thanks! :heart:

site/_docs/templates.md
```diff
@@ -324,9 +324,9 @@ on your system and set `highlighter` to `pygments` in your site's configuration
 file.
 
 Alternatively, you can use [Rouge](https://github.com/jayferd/rouge) to highlight
-your code snippets. It doesn't support as many languages as Pygments does but
-it should fit in most cases and it's written in pure Ruby ; you don't need Python
-on your system!
+your code snippets. It doesn't support as many languages as Pygments, however it
+should fit in most cases. Also, since [Rouge](https://github.com/jayferd/rouge)
```
parkr: `should fit in most cases` reads strangely to me. Maybe:

> It doesn't support as many languages as Pygments, but it should suit most use cases.

Something like this. `fit` sounds like it's a puzzle piece, not a tool. :)

parkr:
Thanks!

site/_docs/templates.md
markphelps: yeah, still didn't sound right to me either. Updated in 9a0db42

PR #3484, opened 2015-02-20
Return nil when comparing a `Post` with a non-`Post`.

justinweiss:
In Ruby < 2.2, `==` will swallow any exceptions raised by `<=>` and
return nil.

In Ruby > 2.2, `==` will surface any exceptions raised by `<=>`.

In Ruby 2.2, `==` will show an annoying warning message if `<=>` raises
an exception, which drowns out most of Jekyll's output.

This patch double-checks that the `date` and `slug` methods exist on
whatever you're comparing a `Post` to, and returns `nil` if they
don't. This keeps compatibility with Ruby 2.2+.

The Ruby change was discussed here:
https://bugs.ruby-lang.org/issues/7688, and this patch was inspired by
the one @tenderlove made to rdoc here:
https://github.com/rdoc/rdoc/commit/23c276de

Why LSI has nil documents it's comparing posts to is beyond me, but this
patch will do the right thing when it does!

### DIFF
--- a/lib/jekyll/post.rb
+++ b/lib/jekyll/post.rb
@@ -155,6 +155,9 @@ module Jekyll
     #
     # Returns -1, 0, 1
     def <=>(other)
+      return unless other.respond_to?(:date) &&
+                    other.respond_to?(:slug)
+
       cmp = self.date <=> other.date
       if 0 == cmp
        cmp = self.slug <=> other.slug
--- a/test/test_post.rb
+++ b/test/test_post.rb
@@ -57,6 +57,10 @@ class TestPost < Test::Unit::TestCase
       end
     end
 
+    should "return nil when compared with a non-Post" do
+      assert_nil Post.allocate <=> nil
+    end
+
     context "processing posts" do
       setup do
         @post = Post.allocate


parkr:
I'm cool with this. How does this affect sorting if a `nil` item is in the array? A test case would be nice for that.

@mattr- ?

lib/jekyll/post.rb
```diff
@@ -155,6 +155,9 @@ def relative_path
     #
     # Returns -1, 0, 1
     def <=>(other)
+      return unless other.respond_to?(:date) &&
```
mattr-: Can we rescue the exception instead?

```ruby
def <=>(other)
  # existing code
rescue
  nil
end
```

mattr-:
If we want to make sure we don't affect sorting, then we'd want to compact any arrays that have `nil` in them.

```
[41] pry(main)> ['a', 'b', nil, 'c'].sort
ArgumentError: comparison of String with nil failed
from (pry):40:in `sort'
[42] pry(main)> ['a', 'b', nil, 'c'].compact.sort
=> ["a", "b", "c"]
```

Granted, the example above is really contrived, but ¯\\\_(ツ)\_/¯

justinweiss:
So, after sleeping on it, I think this is the wrong solution -- it's hiding something that might actually be a bug. So I changed the line that threw the warning (from the LSI gem):

```ruby
proximity_array_for_content( doc, &block ).reject { |pair| pair[0] == doc }
```

to 

```ruby
proximity_array_for_content( doc, &block ).reject { |pair| (pair[0] <=> doc) == 0 }
```

to show the actual error. And it turns out it's *not* comparing to nil. It looks like it's trying to compare a `Post` object to a string (`Post` content).

I can try to take another stab at this and figure out why. But I'd appreciate any insight if you have it!

justinweiss:
Ah, this might be why. Jekyll is indexing posts:


```ruby
        site.posts.each do |x|
          lsi.add_item(x)
        end
```

But finding related posts based on content:

```ruby
    def lsi_related_posts
      self.class.lsi.find_related(post.content, 11) - [post]
    end
```

mattr-:
well, that definitely seems like a source of issues. :smiley: What happens if you change it to find related posts based on the post object itself?

justinweiss:
Changing `lsi_related_posts` to

```ruby
def lsi_related_posts
  self.class.lsi.find_related(post, 11)
end
```

doesn't *seem* to cause any problems. It's still excluding the current post from the related posts, and the related posts seem as reasonable as a magical black-box classifier can give you.

What kind of test would you want to see for a change like this? I could mock it, but that seems like it might be too implementation-focused.

justinweiss:
Two posts later, things still seem to be working well with that `lsi_related_posts` fix. Do you have any suggestions for tests I could write? I'd be happy to close this pull request and start another with the new code.

lib/jekyll/post.rb
parkr: :+1:

justinweiss: Looking into it further, it seems like there's a bug that was being masked in earlier versions of Ruby. Could you take a look at the pull request conversation? I have a fix, I'm just not sure how to put together a test for it.

justinweiss:
This PR is superseded by #3629.

PR #3485, opened 2015-02-20
Removed the trailing slash from the example "/blog/" basurl comment.

watkyn:
When I followed the example in the _config.yml for the base url:
baseurl: "" # the subpath of your site, e.g. /blog/

Jekyll put two slashes in the URL like this: http://yourdomain.com/blog//
I figured the comment might as well be updated.


parkr:
Yes!

Issue #3463, opened 2015-02-14
URL Structure changes


medicart:
How about having the option to remove the **trailing slash** from the url? 

This will make life easier for guys like us **migrating from Drupal** and want to keep the URL structure of the site without drowning in hundreds of 301s for search engines sake.

Look at example here: http://medicart.co.il

To achieve that, we had to use some tricky workarounds using **Amazon S3** capabilities and some unconventional usage of Jekyll pages rather than posts.

I can see how it can turn to be a big headache if one of the tricks will no longer function.

I think users will find this feature useful, and why not in the upcoming **version 3**? it shouldn't be that hard to add this, right?

penibelst:
Related to #3345 and many others.

@jekyll/owners I propose to introduce a new label for permalink design issues.

benbalter:
> How about having the option to remove the trailing slash from the url?... This will make life easier for guys like us migrating from Drupal and want to keep the URL structure of the site without drowning in hundreds of 301s for search engines sake.

In general, I'd rather Jekyll starts from the right way of doing things, and then working backward into functionality, rather than looking at how one piece of software does things, and working in backwards compatibility (that may or may not be right).

In this particular case (without starting a CMS flame war), other than those coming from Drupal, what's the value of slash-less URLs? Intuitively, and logically, it seems [trailing slashes are both the default and more defensible](https://cdivilly.wordpress.com/2014/03/11/why-trailing-slashes-on-uris-are-important/).

As for those migrating from Drupal, at least on GitHub Pages, we can see how hard it would be to redirect from `foo` to `foo/`, but I'm not sure how to do that without breaking at least some sites.

medicart:
@benbalter  agreed about the right way of doing things, I hated Drupal for forcing me to use that URL structure, however, I can think about at least one defensible reason for adding the trailing slash option:

I turns out that a lot of people will leave out the trailing slash when typing a url, even if exists, and unless you have a redirect rule in place, this will create 404. Nowadays, when many people turns to S3 or similar solutions for site hosting, **htaccess** will not be a viable solution. Shouldn't it be at least in a form of an option in Jekyll?

benbalter:
> I[t] turns out that a lot of people will leave out the trailing slash when typing a url

[...]

Issue #2691, opened 2014-08-04
'pretty' permalinks do not match URL TEMPLATE


bcomnes:
These two configuration settings do not result in the same behavior as indicated by the docs:
```yaml
#./_config.yml
- permalink: pretty
```
Results in `./processed.html - > ./processed/index.html`

```yaml
-permalink: /:categories/:year/:month/:day/:title/
```
Results in `./processed.html - > ./processed.html`
where it should actually result in:
```
./processed.html - > ./processed/index.html
```

fulldecent:
This issue is closed, but it is still referenced at http://jekyllrb.com/docs/permalinks/

adius:
Why is this issue closed? It hasn't been fixed yet, or has it?

jalcine:
@adius Probably because a PR is still pending [1](https://github.com/jekyll/jekyll/issues/2691#issuecomment-51929552) [2](https://github.com/jekyll/jekyll-help/issues/138) that addresses this.

adius:
Ah. I guess the issues shouldn't be closed until the PRs are merged!

bcomnes:
As far as I know, nobody has submitted a PR to address this.  The issue, as it stands, is settled which is why this issue is closed.  If someone wants this open, the best way to do that would be to submit a PR that actually addresses the issue.

adius:
^^ Interesting understanding of closed issues.

fulldecent:
Reopen this issue, it is still referenced at http://jekyllrb.com/docs/permalinks/

pathawks:
> A PR would be welcome. :smile:

fulldecent:
Open issues beget PRs

willnorris:
I'm going to reopen this, as I'm taking a stab at implementing a fix, either as part of #3490 or a separate PR.  Rather than introduce any new config settings, my intent is to just make the current behavior a little smarter with regard to URL endings for pages and collections based on the `permalink` setting, which is the confusion that @bcomnes initially described.

So my intended behavior for page URLs is something like:

- if you have `permalink: pretty`, then use the current logic.  That is:
  - html index pages use `/:path/`
  - other html pages use `/:path/:basename/`
  - all other pages use `/:path/:basename:output_ext`
- if you have `permalink: none` or `permalink: date`, use the current logic for that, which means all pages use `/:path/:basename:output_ext`
- otherwise, page URLs should match the ending style specified in `permalink`.  That is:
[...]

PR #3490, opened 2015-02-21
always include file extension on destination files

willnorris:
This is an initial take at always including file extensions on destination files written to disk.  This is a continuation of the discussion that's happened on #2294 and #3463, and coincides with the WEBrick changes in #3452.  

This should only have any impact when the site or post permalink setting is configured without either a trailing slash or the file extension for that file type.  In that case, the file written to disk will contain the correct extension for that type, while the URL will remain as specified in the permalink setting.

Notably, this does break the test that was added in #2925 to support files ending in ".htm" rather than ".html" (cc @pathawks to discuss), because this is using the `output_ext` value to determine the expected extension for the destination file.  I believe this is the correct behavior.  I think that any attempt to support alternate extensions (like ".htm") should make sure one way or another that `output_ext` returns the desired alternate extension.  Not doing so seems likely to cause more problems down the road.  Either way, we certainly need to resolve this breaking test before this can be safely merged.

### DIFF
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -4,7 +4,7 @@ module Jekyll
   class Document
     include Comparable
 
-    attr_reader   :path, :site, :extname
+    attr_reader   :path, :site, :extname, :output_ext
     attr_accessor :content, :collection, :output
 
     YAML_FRONT_MATTER_REGEXP = /\A(---\s*\n.*?\n?)^((---|\.\.\.)\s*$\n?)/m
@@ -19,6 +19,7 @@ module Jekyll
       @site = relations[:site]
       @path = path
       @extname = File.extname(path)
+      @output_ext = Jekyll::Renderer.new(site, self).output_ext
       @collection = relations[:collection]
       @has_yaml_header = nil
     end
@@ -130,7 +131,7 @@ module Jekyll
       {
         collection: collection.label,
         path:       cleaned_relative_path,
-        output_ext: Jekyll::Renderer.new(site, self).output_ext,
+        output_ext: output_ext,
         name:       Utils.slugify(basename_without_ext),
         title:      Utils.slugify(data['slug']) || Utils.slugify(basename_without_ext)
       }
@@ -164,6 +165,7 @@ module Jekyll
       dest = site.in_dest_dir(base_directory)
       path = site.in_dest_dir(dest, URL.unescape_path(url))
       path = File.join(path, "index.html") if url.end_with?("/")
+      path << output_ext unless path.end_with?(output_ext)
       path
     end
 
--- a/lib/jekyll/post.rb
+++ b/lib/jekyll/post.rb
@@ -280,6 +280,7 @@ module Jekyll
       # The url needs to be unescaped in order to preserve the correct filename
       path = site.in_dest_dir(dest, URL.unescape_path(url))
       path = File.join(path, "index.html") if self.url.end_with?("/")
+      path << output_ext unless path.end_with?(output_ext)
       path
     end
 
--- a/test/source/_posts/2015-02-20-extensionless-permalink.markdown
+++ b/test/source/_posts/2015-02-20-extensionless-permalink.markdown
@@ -1,0 +1,7 @@
+---
+layout: ~
+title:  Extensionless Permalink
+permalink: /:title
+---
+
+{{ page.url }}
--- a/features/permalinks.feature
+++ b/features/permalinks.feature
@@ -83,13 +83,3 @@ Feature: Fancy permalinks
     Then the _site directory should exist
     And the _site/custom/posts directory should exist
     And I should see "bla bla" in "_site/custom/posts/some.html"
-
-  Scenario: Use per-post ending in .htm
-    Given I have a _posts directory
-    And I have the following post:
-      | title     | date       | permalink               | content |
-      | Some post | 2013-04-14 | /custom/posts/some.htm  | bla bla |
-    When I run jekyll build
-    Then the _site directory should exist
-    And the _site/custom/posts directory should exist
-    And I should see "bla bla" in "_site/custom/posts/some.htm"
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -142,6 +142,7 @@ module Jekyll
     def destination(dest)
       path = site.in_dest_dir(dest, URL.unescape_path(url))
       path = File.join(path, "index.html") if url.end_with?("/")
+      path << output_ext unless path.end_with?(output_ext)
       path
     end
 
--- a/test/source/deal.with.dots.html
+++ b/test/source/deal.with.dots.html
@@ -1,6 +1,6 @@
 ---
 title: Deal with dots
-permalink: /deal.with.dots/
+permalink: /deal.with.dots
 ---
 
 Let's test if jekyll deals properly with dots.
--- a/test/test_document.rb
+++ b/test/test_document.rb
@@ -212,10 +212,15 @@ class TestDocument < JekyllUnitTest
       })
       @site.process
       @document = @site.collections["slides"].docs[0]
+      @dest_file = dest_dir("slides/test/example-slide-1.html")
     end
 
     should "produce the right URL" do
       assert_equal "/slides/test/example-slide-1", @document.url
+    end
+
+    should "produce the right destination" do
+      assert_equal @dest_file, @document.destination(dest_dir)
     end
   end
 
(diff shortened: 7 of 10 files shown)


lib/jekyll/document.rb
```diff
@@ -164,6 +164,8 @@ def destination(base_directory)
       dest = site.in_dest_dir(base_directory)
       path = site.in_dest_dir(dest, URL.unescape_path(url))
       path = File.join(path, "index.html") if url =~ /\/$/
+      output_ext = Jekyll::Renderer.new(site, self).output_ext
+      path += output_ext if path !~ /#{output_ext}$/
```
parkr: `if path.end_with?(output_ext)`?

lib/jekyll/document.rb
```diff
@@ -164,6 +164,8 @@ def destination(base_directory)
       dest = site.in_dest_dir(base_directory)
       path = site.in_dest_dir(dest, URL.unescape_path(url))
       path = File.join(path, "index.html") if url =~ /\/$/
+      output_ext = Jekyll::Renderer.new(site, self).output_ext
```
parkr: we should really be caching this value

parkr:
So dope. Thanks for the PR!

lib/jekyll/document.rb
willnorris: done.

lib/jekyll/document.rb
willnorris: done (I think the right way... double check)

parkr:
What about checking for the existence of a file extension before adding this one?

```ruby
>> File.extname("path/to/file.htm")
=> ".htm"
>> File.extname("path/to/file.htm").empty?
=> false
>> File.extname("path/to/file")
=> ""
>> File.extname("path/to/file").empty?
=> true
```

If it's empty, then add the output extension. Otherwise, pass.

Also, let's use `<<` instead of `+=`.

test/source/_posts/2015-02-20-extensionless-permalink.markdown
```diff
@@ -0,0 +1,7 @@
+---
+layout: ~
+title:  Extensionless Permalink
+permalink: /:title
+---
+
+Test
```
parkr: Can we output `{{ page.url }}` here to ensure the Liquid URL is being set properly?

willnorris:
> What about checking for the existence of a file extension before adding this one?

So I did think about that.  It will certainly work, but there's two things I didn't like about it: 

1. it prevents having a period in your post title if you have an extensionless permalink structure.  Perhaps a theoretical argument, as I'm not sure how often people actually use periods in post titles (I know I never do)
2. it doesn't address the issue of always being able to trust `output_ext` to accurately identify the extension of the destination file.  Based on how it's being used today, I don't actually see any immediate issues, but it just feels wrong to me.

That said, I'm happy to add it in if you'd like.

> Also, let's use << instead of +=.

Done.

test/source/_posts/2015-02-20-extensionless-permalink.markdown
willnorris: Done.

willnorris:
okay, thinking more about it, agree that respecting existing extensions in permalink is probably the way to go for right now at least.  This next commit should be safe to merge I think.

parkr:
> okay, thinking more about it, I agree that respecting existing extensions in permalink is probably the way to go for right now at least. This next commit should be safe to merge I think.

You make good points... people should be able to use dots in their files, and we don't want to limit file extensions... Hm...

parkr:
Would you mind adding a test for `Document` and `Page` as well? Just until they're all one class...

willnorris:
> > okay, thinking more about it, I agree that respecting existing extensions in permalink is probably the way to go for right now at least. This next commit should be safe to merge I think.
>
> You make good points... people should be able to use dots in their files, and we don't want to limit file extensions... Hm...

okay, yeah... I now see that there are test cases that specifically focus on files with dots in their name, so this is definitely something we'll want to support.  So now I'm back to feeling like we will need to back out the test case from #2925 and find another way to support that (if at all).  As @pathawks mentioned on that bug, he *could* setup redirects for the legacy URLs, which may be the better option here.

Right now, I'm fairly certain that Pages still end up with URLs contain the ".html" extension.  That never actually impacted me, so I sort of forgot about it.  I'll look into what it will take to fix that, and add tests for that and Document (as well as additional tests for dotted filenames).

parkr:
I think we should be able to support `htm` extensions, even if just to maintain backwards compatability with the hundreds of thousands (guessing, maybe more or less) of GitHub Pages sites that are presently out there... Perhaps we do check against a list of possible file extensions? That may slow down this method considerably...

One idea is to release this in a beta, have @pathawks take a look, and come up with a fix for that later.

> Right now, I'm fairly certain that Pages still end up with URLs contain the ".html" extension. That never actually impacted me, so I sort of forgot about it. I'll look into what it will take to fix that, and add tests for that and Document (as well as additional tests for dotted filenames).

Yes, I believe they are presently just as flexible as Posts, so it would be good to have them under test. Thank you!!

pathawks:
I'd be happy to test :+1: 

You want I should test this pull request?

parkr:
@pathawks Yes please.

willnorris:
> Yes, I believe they are presently just as flexible as Posts, so it would be good to have them under test. Thank you!!

Actually pages are the worst in terms of permalink flexibility (from _config.yaml that is).  If you're not using `:pretty`, then pages always use a template of `/:path/:basename:output_ext`.  This is a large part of #2691, which I'm trying to fix here as well (haven't uploaded that commit yet.. that's what prompted #3507)

parkr:
Ok then this PR looks good to me.

@pathawks, let me know what you figure out. It's hacky, but we could also check for `.htm`...

lib/jekyll/post.rb
```diff
@@ -280,6 +280,7 @@ def destination(dest)
       # The url needs to be unescaped in order to preserve the correct filename
       path = site.in_dest_dir(dest, URL.unescape_path(url))
       path = File.join(path, "index.html") if self.url =~ /\/$/
```
parkr: also this should totally be `.end_with?("/")`... not sure if that's for this PR but it's an optimization we should take advantage of. i'll write a benchmark to ensure this

willnorris:
@parkr: as previously noted, we do have tests that ensure that files with dots in them work, which this method of simply looking for a file extension doesn't work with.  As a result, this next commit switches back to enforcing `output_ext` exactly.  This also necessitates removing @pathawks' feature test for .htm permalinks.  Supporting non-standard file extensions like this will likely need to be done in a plugin, which I'll take a look at separately, just to show how it would be done for those that are interested.

This next commit also includes tests for posts and documents.

parkr:
What about allowing configuration of HTML files' file extension? is that crazy?

willnorris:
> What about allowing configuration of HTML files' file extension? is that crazy?

personally I think so, yes. :)

just make it a setting

Also keep in mind that I'm fairly certain @pathawks was only interested in having ".htm" extension for legacy, imported-in content.  I don't think there is any real demand for having .htm versus .html long-term, we just need a solution that is good enough to handle Pat's imported content (which I personally would still argue that just redirecting to the new URLs would be the much simpler approach)

parkr:
> Also keep in mind that I'm fairly certain @pathawks was only interested in having ".htm" extension for legacy, imported-in content. I don't think there is any real demand for having .htm versus .html long-term

:+1:

parkr:
also /cc @benbalter for that great image above. ben's always talking about our :christmas_tree: of options.

pathawks:
> I'm fairly certain @pathawks was only interested in having ".htm" extension for legacy, imported-in content.

Yup.

> just redirecting to the new URLs would be the much simpler approach

Yup.

willnorris:
@pathawks: in my brief testing, this plugin seems to do the trick of preserving the ".htm" extension when a post's custom permalink contains that extension:

```ruby
module Jekyll
  module Convertible
    def output_ext_with_htm
      ext = output_ext_without_htm
      ext = ".htm" if self.permalink.to_s.end_with?(".htm")
      ext
    end

    alias_method :output_ext_without_htm, :output_ext
    alias_method :output_ext, :output_ext_with_htm
  end
end
```

Or of course, you could redirect from within the web server itself.

Assuming that is okay and we're fine removing this behavior from jekyll core, then this should be ready for final review.  @parkr could you take another look (there's a few additional tests now as well)

parkr:
I'm fine with removing this feature. We'll need to include it on our upgrading docs page.

parkr:
This looks great to me! :shipit: 

Thanks, Will!

willnorris:
I don't think it necessarily needs mentioning in the upgrade docs, since this never actually worked until #3014, which has only shipped with 3.0.0.beta1.

(and just to make sure, "ship-it squirrel" == "go ahead and merge"?  Or should I wait for you or someone else to merge?)

pathawks:
:+1:

parkr:
> (and just to make sure, "ship-it squirrel" == "go ahead and merge"? Or should I wait for you or someone else to merge?)

Yep, :shipit: means I'm cool merging it. Thanks for checking :)

alfredxing:
Looks good! Though this only works for markdown and converted files, right?

So if I had a `a-page.html`:
````yaml
---
layout: default
permalink: /a-page
---
<strong>This is a page</strong>
````

It would output as `/a-page`, not `/a-page.html`?

willnorris:
It works just as well for non-converted files.  If there is no converter registered for the file extension, then `output_ext` returns the extension of the source file.  So in your example, the page would have a url of `/a-page`, but the destination file would be `/a-page.html`.

alfredxing:
@willnorris Right, forgot about that. Thanks!

snickers0129:
I like the idea of configuring the html extension for .html files as well...  I'm hosting on Amazon S3, which supports serving files that are extension-less and doesn't really offer configuration

willnorris:
@snickers0129 you could do a variation of https://github.com/jekyll/jekyll/pull/3490#issuecomment-76877725 to customize the extension out HTML output files.

willnorris:
err, you might actually need to patch the Markdown converter, but it would still be the `output_ext` method.

envygeeks:
Why not just do folder based navigation? The AWS docs states that's supported?

snickers0129:
@willnorris Thanks! I saw that... I just thought I'd cast my vote!

@envygeeks You mean putting folders with index.html in them?  That causes 302 redirects from url/page to url/page/

envygeeks:
Oh fair enough, I didn't notice it would do that crazy mess, but I guess I shouldn't be surprised it is a CDN.

PR #3392, opened 2015-01-31
Add Hooks, based on the Octopress::Hooks project

parkr:
@imathis did a splendid job with [octopress-hooks](https://github.com/octopress/hooks). It gives plugin authors the ability to create a hook for Jekyll content and perform some action within the pipeline.

In an effort to save us all some monkey-patching frustrations, I'm porting that work over into Jekyll proper and giving Brandon the commit credit for all the hard work he did on the project.

- [x] Site hooks
- [ ] Page hooks
- [ ] Post hooks (squash with document hooks?)
- [ ] Document Hooks
- [ ] "All" hooks (not sure what this is?)
- [ ] Tests, tests, tests!

PR #3498, opened 2015-02-22
#3392 copy page code from octopress-hooks

stevecrozz:
This commit adds support for all the page hooks from the octopress-hooks project

### DIFF
--- a/lib/jekyll/convertible.rb
+++ b/lib/jekyll/convertible.rb
@@ -230,19 +230,29 @@ module Jekyll
     #
     # Returns nothing.
     def do_layout(payload, layouts)
-      info = { :filters => [Jekyll::Filters], :registers => { :site => site, :page => payload['page'] } }
+      pre_render if respond_to?(:pre_render) && hooks
+
+      if respond_to?(:merge_payload) && hooks
+        layout_payload = merge_payload(payload.dup)
+      else
+        layout_payload = payload
+      end
+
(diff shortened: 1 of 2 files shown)


envygeeks:
I'm definitely :-1: on this, encapsulate it.  I'm not a fan of skynet classes.

envygeeks:
Would you be willing to encapsulate the hooks and then use that in the class?

stevecrozz:
@envygeeks sure. In that case, I'd like to just disregard https://github.com/jekyll/jekyll/pull/3392 entirely which I was following along with and restart this effort. I'll close this one now and target my next PR to master.

willnorris:
> I'm definitely :-1: on this, encapsulate it.  I'm not a fan of skynet classes.

@envygeeks: okay, just for my edification, what are "skynet classes"? :) I mean, I get the skynet reference, just not sure about the application here.

envygeeks:
@willnorris A Skynet class (to me) is for example what people often do with service classes or models in Rails when they first start (I'm not saying you or @stevecrozz are new at Rails btw, it's just a great example,) they put everything into the same class instead of encapsulating concerns.  The class gets too smart and too big for itself and ends up figuratively killing it's master by killing development and takes over everything.

willnorris:
got it :smile:  And I have zero experience with Rails (and very little with Ruby), so absolutely no offense taken.  All of my coding is in Go these days, so I'm happy to learn what it idiomatic in Ruby.

PR #3509, opened 2015-02-24
Add closing tags while generating post excerpts

light94:
While generating post excerpts, if another excerpt follows , the formatting of the previous excerpt is applied to the new one too. This happens because the <em> and <strong> tags are not closed in excerpts.

### DIFF
--- a/lib/jekyll/excerpt.rb
+++ b/lib/jekyll/excerpt.rb
@@ -107,6 +107,25 @@ module Jekyll
     def extract_excerpt(post_content)
       head, _, tail = post_content.to_s.partition(post.excerpt_separator)
 
+     
+      emCount = head.scan("<em>").count
+      strongCount = head.scan("<strong>").count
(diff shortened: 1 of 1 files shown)


kleinfreund:
This is not a issue with how excerpts are handled, or is it? Opening/closing HTML tags should be handled by the user and interpreted by the browser. We don’t want to interfer with that.

light94:
I agree with your view, although this commit proved to be helpful in my case, when I had imported my blog posts from wordpress and the excerpts did not look alike, due to em and strong tags not being closed.

kleinfreund:
In this case, I really think they should be closed manually in the related files.

parkr:
This is a piece of importing, not a problem here. I'f I were you, I'd run this change on all your files and rewrite to make sure the first paragraph (delineated with "\n\n") has the right tags. :smile:

Issue #3491, opened 2015-02-21
Liquid Exception: undefined method `sort' for nil:NilClass


luixxiul:
By running ``jekyll serve`` with ``--limit_posts``, the code below occurs the undefined method error;

````
{% assign something = site.categories.example | sort: "title" %}
{% for post in something %}
    <li><a href="{{ post.url }}">{{ post.title }}</a></li>
{% endfor %}  
````
Is there any way to resolve it without changing the number of ``--limit_posts``?

parkr:
Can you please run with `jekyll build --trace` and paste the output?

luixxiul:
Okay, here is the output:
````
C:\Users\Admin\Jekyll> bundle exec jekyll serve --limit_posts 1 --trace
DL is deprecated, please use Fiddle
  Liquid Exception: undefined method `sort' for nil:NilClass in map/index.html
C:/Users/Admin/Jekyll/vendor/bundler/ruby/2.1.0/gems/jekyll-2.5.3/lib/jekyll/filters.rb:248:in `sort': undefined method `sort' for nil:NilClass (NoMethodError)
        from C:/Users/Admin/Jekyll/vendor/bundler/ruby/2.1.0/gems/liquid-2.6.2/lib/liquid/strainer.rb:43:in `invoke'
        from C:/Users/Admin/Jekyll/vendor/bundler/ruby/2.1.0/gems/liquid-2.6.2/lib/liquid/context.rb:82:in `invoke'
        from C:/Users/Admin/Jekyll/vendor/bundler/ruby/2.1.0/gems/liquid-2.6.2/lib/liquid/variable.rb:50:in `block in render'
        from C:/Users/Admin/Jekyll/vendor/bundler/ruby/2.1.0/gems/liquid-2.6.2/lib/liquid/variable.rb:38:in `each'
        from C:/Users/Admin/Jekyll/vendor/bundler/ruby/2.1.0/gems/liquid-2.6.2/lib/liquid/variable.rb:38:in `inject'
        from C:/Users/Admin/Jekyll/vendor/bundler/ruby/2.1.0/gems/liquid-2.6.2/lib/liquid/variable.rb:38:in `render'
        from C:/Users/Admin/Jekyll/vendor/bundler/ruby/2.1.0/gems/liquid-2.6.2/lib/liquid/tags/assign.rb:26:in `render'
[...]

PR #3520, opened 2015-02-26
Sort will now raise error on nil object array input.

MartinRogalla:
Sort will now throw an error when a nil object array is given as input.
See issue #3491 for more information.

Signed-off-by: Martin Jorn Rogalla <martin@martinrogalla.com>

### DIFF
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -223,7 +223,7 @@ module Jekyll
     # Returns the filtered array of objects
     def sort(input, property = nil, nils = "first")
       if input.nil?
-          raise ArgumentError.new("Nil object array given. Sort cannot process an empty object array.")
+          raise ArgumentError.new("Sort: cannot sort a null object.")
       end
       if property.nil?
         input.sort
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -223,7 +223,7 @@ module Jekyll
     # Returns the filtered array of objects
     def sort(input, property = nil, nils = "first")
       if input.nil?
-          raise ArgumentError.new("Sort: cannot sort a null object.")
+          raise ArgumentError.new("Cannot sort a null object.")
       end
       if property.nil?
         input.sort
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -223,7 +223,7 @@ module Jekyll
     # Returns the filtered array of objects
     def sort(input, property = nil, nils = "first")
       if input.nil?
-          raise ArgumentError.new("Invalid object array given. Object array is null.")
+          raise ArgumentError.new("Nil object array given. Sort cannot process an empty object array.")
       end
       if property.nil?
         input.sort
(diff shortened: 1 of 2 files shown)


alfredxing:
Thanks! What does this mean in terms of output from the console?

parkr:
We definitely want tests for this :)

MartinRogalla:
@alfredxing 
Before:
```
Liquid Exception: undefined method `sort' for nil:NilClass in index.html
jekyll 3.0.0-beta1 | Error:  undefined method `sort' for nil:NilClass
```
After:
```
Liquid Exception: Liquid error: Invalid object array given. Object array is null. in index.html
jekyll 3.0.0-beta1 | Error:  Liquid error: Invalid object array given. Object array is null.
```


----


@parkr I understand. So write tests for the sort function and all functions dependent on the sort function?

parkr:
Hm, that new error is still vague. We should be catching them in Renderer/Convertible. Can we use an error class that will be caught?

MartinRogalla:
Hmm, I'm looking at it right now. I'm trying to find a way for liquid to also give back a line number in the offending document.

Right now I have:
```
Liquid Exception: Sort: null object array given. Sort cannot process an empty object array in index.html.
```

Does anyone know how to throw an error with the line number that liquid is currently trying to render? 

@parkr if I'm going in the wrong direction, could you give me an example of what kind of error you're expecting to see?

parkr:
> Does anyone know how to throw an error with the line number that liquid is currently trying to render?

I believe you can just ask a Liquid error for its `line_number`, but I'm not sure if we create a LiquidError here or not. https://github.com/Shopify/liquid/blob/3-0-stable/lib/liquid/errors.rb

I missed the `in index.html` – all the other errors have this and I was reviewing on my phone which doesn't have reliable side scrolling. My bad!

> @parkr I understand. So write tests for the sort function and all functions dependent on the sort function?

No, just a test that asserts that passing a `nil` to `sort` will raise an argument error that looks like this one. :)

lib/jekyll/filters.rb
```diff
@@ -222,6 +222,9 @@ def where(input, property, value)
     #
     # Returns the filtered array of objects
     def sort(input, property = nil, nils = "first")
+      if input.nil?
+          raise ArgumentError.new("Invalid object array given. Object array is null.")
```
parkr: Let's make this message `Cannot sort a null object.` or something?

MartinRogalla: Okay, just committed before this comment. Will update it now.

lib/jekyll/filters.rb
```diff
@@ -222,6 +222,9 @@ def where(input, property, value)
     #
     # Returns the filtered array of objects
     def sort(input, property = nil, nils = "first")
+      if input.nil?
+          raise ArgumentError.new("Sort: cannot sort a null object.")
```
parkr: don't you think `cannot sort a null object` tells you that you're in the `sort` function?

MartinRogalla:
True, don't know why I put it in there. Thanks!

MartinRogalla:
@parkr is it correct now or do you need any changes to this pull request?

parkr:
This LGTM! @mattr-?

mattr-:
:+1: Thanks!

PR #3319, opened 2015-01-17
Split off Textile into jekyll-textile-converter

parkr:
Textile is a marginal piece of our user base and isn't really worth having around in core at this point. In order to continue supporting it in Jekyll 3 and beyond, it's been turned into a plugin: https://github.com/jekyll/jekyll-textile-converter

Fixes #3316 to a certain degree.

PR #3528, opened 2015-02-28
switch remaining textile test files to markdown

### LINKED ISSUE #3507: several tests still use .textile files
Textile functionality was removed in #3319, and many of the tests were switched over to markdown at that time.  However, there still exist a large number of textile files in [tests/source/_posts](https://github.com/jekyll/jekyll/tree/master/test/source/_posts).  This is technically not a problem with the current tests as written, but is becoming a problem with the work I'm doing for #3490, since that now relies on `output_ext` returning the correct extension.  Since the new [jekyll-textile-converter](https://github.com/jekyll/jekyll-textile-converter) plugin isn't present when running tests, `output_ext` doesn't return the expected value of ".html" for textile files, breaking various tests.  So we either need to include the textile plugin when running the tests, or switch the remaining textile files over to markdown.  I'll be following up with  PR that does the latter.  

(feel free to assign this issue to me, or just consider it as such)

willnorris:
Textile support was removed from jekyll core in #3319, and most of the tests switched to markdown at that time.  This changes the remaining tests to use markdown as well.  The vast majority of the test cases were testing things in the file name or front matter, so it doesn't really matter what markup format they use.  The one test that was claiming to test that textile was transformed had actually been moved to markdown as well, just not renamed.

Fixes #3507


parkr:
This LGTM. @mattr-?

mattr-:
:metal: :sparkles: Thanks!

PR #3529, opened 2015-02-28
Remove adapters deprecation warning

fabianrbz:
Hi guys,

This pull request removes the following [deprecation warning](https://github.com/colszowka/simplecov/blob/fee9dcf1f990a57503b0d518d9844a7209db4734/lib/simplecov/configuration.rb#L124) from simplecov.

The new way is using [profiles](https://github.com/colszowka/simplecov#profiles). The warning comes from the gem 'simplecov-gem-adapter', which haven't been updated for the last 4 years.
In this pull request l removed the gem dependency and added a new profile with some of the settings that were in the gem. l thought it was better this way -instead of making a PR to the gem- because this way, we remove a dependency and it will be easier to customize the profile in the future.
What do you think?

Cheers!


lib/simplecov_custom_profile.rb
```diff
@@ -0,0 +1,10 @@
+require 'simplecov'
+
+SimpleCov.profiles.define 'gem' do
+  add_filter '/test/'
+  add_filter '/features/'
+  add_filter '/autotest/'
+
+  add_group 'Binaries', '/bin/'
+  add_group 'Libraries', '/lib/'
+end
```
parkr: Let's put this file in `test`.

test/helper.rb
```diff
@@ -1,5 +1,4 @@
-require 'simplecov'
-require 'simplecov-gem-adapter'
+require 'simplecov_custom_profile'
```
parkr: Please use `File.expand_path` to require this file.

parkr:
Great!!

fabianrbz:
@parkr thanks! what happened to the comments you made? were they implemented?

parkr:
> @parkr thanks! what happened to the comments you made? were they implemented?

Yep! fbe52bca85b66ad89040ecd1519dc9021ab5a532

fabianrbz:
sweet! thanks @parkr!

PR #3538, opened 2015-03-03
Improved permalinks for pages and collections

willnorris:
This updates the default permalink style for pages and collections to match the site-wide 'permalink' setting (more detailed notes in the commit message).

I'm not wild about the method name `Utils.append_permalink_ending`, but at least it's accurate.  I'm certainly open to suggestions for a better name there.

I also have not yet written the user-facing docs describing how this works.  I'll certainly do that as part of this PR, but wanted to go ahead and get some :eyes: on the implementation before then.  I also wasn't quite sure how we handle docs for features that aren't yet in the current stable release.  Do we just hold off pushing those live?  Or do we just say that the docs on jekyllrb.com correspond to the latest master branch, not necessarily any particular released version?

### DIFF
--- a/lib/jekyll/collection.rb
+++ b/lib/jekyll/collection.rb
@@ -170,7 +170,9 @@ module Jekyll
     #
     # Returns the URL template to render collection's documents at.
     def url_template
-      metadata.fetch('permalink', "/:collection/:path:output_ext")
+      metadata.fetch('permalink') do
+          Utils.add_permalink_suffix("/:collection/:path", site.permalink_style)
+      end
     end
 
     # Extract options for this collection from the site configuration.
--- a/lib/jekyll/utils.rb
+++ b/lib/jekyll/utils.rb
@@ -158,5 +158,45 @@ module Jekyll
         downcase
     end
 
+    # Add an appropriate suffix to template so that it matches the specified
+    # permalink style.
+    #
+    # template - permalink template without trailing slash or file extension
+    # permalink_style - permalink style, either built-in or custom
+    #
+    # The returned permalink template will use the same ending style as
+    # specified in permalink_style.  For example, if permalink_style contains a
+    # trailing slash (or is :pretty, which indirectly has a trailing slash),
+    # then so will the returned template.  If permalink_style has a trailing
+    # ":output_ext" (or is :none, :date, or :ordinal) then so will the returned
+    # template.  Otherwise, template will be returned without modification.
+    #
+    # Examples:
+    #   add_permalink_suffix("/:basename", :pretty)
+    #   # => "/:basename/"
+    #
+    #   add_permalink_suffix("/:basename", :date)
+    #   # => "/:basename:output_ext"
+    #
+    #   add_permalink_suffix("/:basename", "/:year/:month/:title/")
+    #   # => "/:basename/"
+    #
+    #   add_permalink_suffix("/:basename", "/:year/:month/:title")
+    #   # => "/:basename"
+    #
+    # Returns the updated permalink template
+    def add_permalink_suffix(template, permalink_style)
+      case permalink_style
+      when :pretty
+        template << "/"
+      when :date, :ordinal, :none
+        template << ":output_ext"
+      else
+        template << "/" if permalink_style.to_s.end_with?("/")
+        template << ":output_ext" if permalink_style.to_s.end_with?(":output_ext")
+      end
+      template
+    end
+
   end
 end
--- a/test/test_utils.rb
+++ b/test/test_utils.rb
@@ -166,4 +166,18 @@ class TestUtils < JekyllUnitTest
     end
   end
 
+  context "The \`Utils.add_permalink_suffix\` method" do
+    should "handle built-in permalink styles" do
+      assert_equal "/:basename/", Utils.add_permalink_suffix("/:basename", :pretty)
+      assert_equal "/:basename:output_ext", Utils.add_permalink_suffix("/:basename", :date)
+      assert_equal "/:basename:output_ext", Utils.add_permalink_suffix("/:basename", :ordinal)
+      assert_equal "/:basename:output_ext", Utils.add_permalink_suffix("/:basename", :none)
+    end
+
+    should "handle custom permalink styles" do
+      assert_equal "/:basename/", Utils.add_permalink_suffix("/:basename", "/:title/")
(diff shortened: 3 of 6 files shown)


test/test_utils.rb
```diff
@@ -166,4 +166,18 @@ class TestUtils < JekyllUnitTest
     end
   end
 
+  context "The \`Utils.append_permalink_ending\` method" do
+    should "handle built-in permalink styles" do
+      assert_equal "/:basename/", Utils.append_permalink_ending("/:basename", :pretty)
```
parkr: What if I have `"/:basename/", :pretty`?

lib/jekyll/utils.rb
```diff
@@ -158,5 +158,45 @@ def slugify(string, mode=nil)
...
+    #   # => "/:basename/"
+    #
+    #   append_permalink_ending("/:basename", "/:year/:month/:title")
+    #   # => "/:basename"
+    #
+    # Returns the updated permalink template
+    def append_permalink_ending(template, permalink_style)
+      case permalink_style
+      when :pretty
+        template << "/"
```
parkr: Dealing with user input here, it's possible this could result in `/:basename//`.

lib/jekyll/utils.rb
```diff
@@ -158,5 +158,45 @@ def slugify(string, mode=nil)
...
+    #   # => "/:basename:output_ext"
+    #
+    #   append_permalink_ending("/:basename", "/:year/:month/:title/")
+    #   # => "/:basename/"
+    #
+    #   append_permalink_ending("/:basename", "/:year/:month/:title")
+    #   # => "/:basename"
+    #
+    # Returns the updated permalink template
+    def append_permalink_ending(template, permalink_style)
```
parkr: Yeah, this is a bit fuzzy when just skimming the code. "Ending" is kind of vague and we could shorten `append` to `add`? Maybe `add_permalink_suffix`? Not quite sure what would be best here.

lib/jekyll/utils.rb
```diff
@@ -158,5 +158,45 @@ def slugify(string, mode=nil)
...
+      case permalink_style
+      when :pretty
+        template << "/"
+      when :date, :ordinal, :none
+        template << ":output_ext"
```
parkr: Same here, is it possible this could end up being `/:basename:output_ext:output_ext`?

lib/jekyll/collection.rb
```diff
@@ -170,7 +170,8 @@ def write?
     #
     # Returns the URL template to render collection's documents at.
     def url_template
-      metadata.fetch('permalink', "/:collection/:path:output_ext")
+      default = Utils.append_permalink_ending("/:collection/:path", site.permalink_style)
+      metadata.fetch('permalink', default)
```
parkr: Maybe the default should be lazy-loaded via a block?

parkr:
Woot! Looking pretty good, just a couple questions above.

test/test_utils.rb
willnorris: you wouldn't.  The value of template is not user-supplied, it's hardcoded in the method calls in `document.rb` and `page.rb` so we always control what the value is, and the docs for the method explicitly say that template should not have a trailing slash or extension.

lib/jekyll/utils.rb
willnorris: same as above, this shouldn't ever happen.  We could always add a check here to ensure that it doesn't, but that seems unnecessary to me.

lib/jekyll/utils.rb
willnorris: template isn't user input

lib/jekyll/utils.rb
willnorris: `add_permalink_suffix` sounds good to me.  done.

lib/jekyll/collection.rb
willnorris: and now I've reached the limit of my ruby knowledge :) I'm not sure I know what that would look like.

test/test_utils.rb
envygeeks: :-1: to hardcoding anything that can be user configurable.

lib/jekyll/collection.rb
envygeeks: @willnorris: http://ruby-doc.org//core-1.9.3/Hash.html#method-i-fetch

```ruby
metadata.fetch('permalink') do
  Utils.append_permalink_ending("/:collection/:path", site.permalink_style)
end
```

test/test_utils.rb
willnorris: huh?  This particular value isn't user configurable... it's the default template used for page/collection permalinks.  Look at the code it is replacing... it's already hardcoded.  Now we're just making the suffix match what the user has in their `permalink` setting in `_config.yaml`.  Users can always provide their own per-page `permalink` value in the front matter, or the per-collection `permalink` setting in `_config.yaml`

lib/jekyll/collection.rb
willnorris: ah, thanks.  fixed.

parkr:
I'm satisfied with this. Let's get it into the next beta so it can be thoroughly tested before we release it. :shipit:

willnorris:
what about user-facing docs (e.g. http://jekyllrb.com/docs/permalinks/)?  How do we handle that for features like this that are not in a stable release yet?

envygeeks:
@parkr @willnorris Do we not have edge docs?

willnorris:
@envygeeks we very well might, I'm just not completely up to speed on how all of that is setup.  I can write the docs as part of this PR, but just want to make sure they don't accidentally get published somewhere they shouldn't.

willnorris:
> I'm satisfied with this. Let's get it into the next beta so it can be thoroughly tested before we release it. :shipit: 

I'm going to go ahead and check this in so we can start testing it.  I'll do the user docs in a separate PR, and can figure out then what to do about edge docs.

parkr:
We don't have edge docs. I like `tip.golang.org`, but it's hard to keep fixes and changes due to upgrading features separate. We'd have to argue that once we release something, the docs are as-is and you just go to like `jekyllrb.com/2.5.3/` and it's the docs as they were at the `v2.5.3` tag.

We usually add a note, check out the bottom of this: http://jekyllrb.com/docs/home/

willnorris:
ah, started merging this pr just before seeing your comment :).  I'll add an unreleased note for the relevant changes to the permalink docs.

parkr:
Thank you! Please ref this pr in that commit.

Issue #3048, opened 2014-10-31
0.0.0.0:4000 unreachable in Chrome Canary


kdzwinel:
Few days ago I found out that I'm unable to open jekyll's default local URL (`0.0.0.0:4000`) in Chrome Canary (nightly version of Chrome). Instead, Chrome redirects me to the Google results for "0.0.0.0". I've alerted Chromium team about it ( http://crbug.com/428046 ) and I've learned that:
- `0.0.0.0` is a non-routable address
- they are going to push this change to stable
- I should use `127.0.0.1:4000` instead (and that it actually works)

My guess is that, when this change lands in stable Chrome, many people will be confused, as I was, about `0.0.0.0` being not accesible. To avoid that, it might be a good idea to replace default "Server address: http://0.0.0.0:4000/" with "Server address: http://localhost:4000".

ibnesayeed:
This change causes more inconveniences than it solves. The IP address `0.0.0.0` is not routable when resolved, that does not mean a service should not run on this address.

The issue it solves:

- The server says it is listening on 0.0.0.0:4000 but some browsers cant resolve that address.

The issues it causes:

- Running the server with default config is no longer accessible outside the localhost
- This affects deployment in Docker (with default config) as it will not allow accessing from outside the container until one gets the surprise revelation (or a some searching around) that a custom host must be supplied explicitly (which is against Ruby's philosophy of developer happiness and simpler fulfillment of expectations)

A possibly better approach:

- Keep the server listening on 0.0.0.0 but advertise local access differently, unless a custom hostname is provided

#3542 has some great suggestions on how to display a meaningful message.

parkr:
You can very easily run `jekyll serve --host 127.0.0.1`. Going back and forth with this IP address has caused nothing but headaches and I don't want to see it changed yet again, causing another influx of issues. Please use the configuration options we provide for your use case.

ibnesayeed:
[...]

PR #3542, opened 2015-03-04
Make 0.0.0.0 the default for listening.

### LINKED ISSUE #3540: Support JEKYLL_DOCKER env variable so we can alter defaults without shivs.
With people asking about installing Gems in the Docker Jekyll image it might be useful to accept a `JEKYLL_DOCKER` env variable so that we can default the source path to `/srv/jekyll` and the listen address to `0.0.0.0` if we detect it, which will allow people to use `bundler` easier and will allow us to stop shiv-ing the current Jekyll bin file.

/cc @jekyll/core

envygeeks:
Fixes #3540.

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -45,7 +45,7 @@ module Jekyll
             file_handler_options
           )
 
-          Jekyll.logger.info "Server address:", server_address(s, options)
+          Jekyll.logger.info "Server address:", server_address_info(s, options)
 
           if options['detach'] # detach the server
             pid = Process.fork { s.start }
@@ -125,15 +125,14 @@ module Jekyll
           WEBrick::HTTPUtils::load_mime_types(mime_types_file)
         end
 
-        def server_address(server, options)
-          baseurl = "#{options['baseurl']}/" if options['baseurl']
-          [
-            "http://",
-            server.config[:BindAddress],
-            ":",
-            server.config[:Port],
-            baseurl || ""
-          ].map(&:to_s).join("")
+        def server_address_info(server, options)
+          addr = server.config[:BindAddress].to_s
+          burl = "#{options['baseurl']}/" if options['baseurl']
+          ladr = addr == "0.0.0.0" ? "localhost" : a
+
+          rtn = "http://#{ladr}:#{server.config[:Port]}#{burl}"
+          rtn = "#{rtn} (listening on all interfaces)" if addr == "0.0.0.0"
+        rtn
         end
 
         # recreate NondisclosureName under utf-8 circumstance
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -126,12 +126,12 @@ module Jekyll
         end
 
         def server_address_info(server, options)
-          addr = server.config[:BindAddress].to_s
-          burl = "#{options['baseurl']}/" if options['baseurl']
-          ladr = addr == "0.0.0.0" ? "localhost" : a
+          bind_addr = server.config[:BindAddress].to_s
+          listen_addr = bind_addr == "0.0.0.0" ? "localhost" : bind_addr
+          base_url  = "#{options['baseurl']}/" if options['baseurl']
 
-          rtn = "http://#{ladr}:#{server.config[:Port]}#{burl}"
-          rtn = "#{rtn} (listening on all interfaces)" if addr == "0.0.0.0"
+          rtn = "http://#{listen_addr}:#{server.config[:Port]}#{base_url}"
+          rtn = "#{rtn} (listening on all interfaces)" if bind_addr == "0.0.0.0"
         rtn
         end
 
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -42,7 +42,7 @@ module Jekyll
       # Serving
       'detach'        => false,          # default to not detaching the server
       'port'          => '4000',
-      'host'          => '127.0.0.1',
+      'host'          => '0.0.0.0',
       'baseurl'       => '',
 
       # Backwards-compatibility options


parkr:
/cc @jekyll/core, for any concerns.

Does this run without root or sysadmin privileges?

envygeeks:
That's up to the system not up to us, by default no device is privileged and if the user chooses to make one we cannot help that or fix that as it was an explicit choice.  Linux by default only has the notion of privileged ports, of which we do not listen on.

troyswanson:
IIRC this was moved to 127.0.0.1 in another PR a while ago due to Google
Chrome not considering 0.0.0.0 a valid IP address. I'm on the road right
now but I will look for the related issue when I can.

envygeeks:
I'm on Google Chrome right now and it just worked for me on Chrome Stable.  Can the rest of you confirm that it works for you on Chrome Stable and that it works on OS X?

willnorris:
relevant chromium bug is https://code.google.com/p/chromium/issues/detail?id=428046

envygeeks:
Hmmmm... that can be easily fixed with a message, something of the sort of: `We are listening on 0.0.0.0 (all interfaces) but try http://localhost:4000`... that is, if we detect 0.0.0.0 as the listening device, but I don't know what they mean by QUERY (since there are dozens of things I can think of.) If they ask my system what or where 0.0.0.0 goes to that would explain why it works on my Linux box (kinda, because everything I do on 0.0.0.0 routes straight to 127.0.0.1.)

troyswanson:
Is there an inherent problem with 127.0.0.1?

parkr:
> Is there an inherent problem with 127.0.0.1?

@troyswanson It screws with serving via Jordon's docker container.

Thanks for remembering that! https://github.com/jekyll/jekyll/issues/3048 and https://github.com/jekyll/jekyll/pull/3053

willnorris:
can't the dockerfile just manually pass a host flag?

envygeeks:
@willnorris We don't entrypoint to Jekyll so that is out of the question (and we do that because the last thing I want to do is inhibit their ability to extend the image.)  That said, right now we shiv the system Jekyll command via https://github.com/jekyll/docker/blob/master/ubuntu/shared/usr/bin/jekyll but that causes an issue: How do we deal with `bundle install --path vendor/bundle` for persistence without persistence in docker.  My goal with this image is to make things with Jekyll as painless and as smooth as possible and this default will do that, we just have to find a way to deal with it.  Whether we include it by default and send a message the way that Rails does or whether we accept an `JEKYLL_ENV` we've to figure out some kind of solution so the docker image is painless for any workflow!

willnorris:
@envygeeks got it, that completely makes sense.  Then given that jekyll was listening on all interfaces anyway until very recently, I think switching back to that and just updating the startup message seems like a reasonable approach.

parkr:
Suggestion:

screen shot 2015-03-04 at 6 38 37 pm

jglovier:
How about:

`serving in browser at http://localhost:4000/ (listening on http://0.0.0.0:4000/)`

jglovier:
Or simply:

`Server address: http://localhost:4000 (listening on http://0.0.0.0:4000/)`

envygeeks:
I like the latter one the best, minus the trailing slash before the ')' for some reason that really bugs me!

envygeeks:
If everyone else can agree I'll implement the change in the morning.

willnorris:
I would caution against making an actual http URL with 0.0.0.0, since as noted in the chromium bug, it's technically an unroutable host and will fail on at least some browsers.  Instead, I would suggest something like "listening on port 4000" or "listening on *:4000" or "listening on all interfaces" or what-have-you.

willnorris:
or I guess even "listening on 0.0.0.0:4000" would be fine.  If it doesn't look like a URL, we're less likely to have people try to launch it in their browser and start reporting bugs again.

envygeeks:
IMO they should both be 'http` URL's because regardless of whatever Chrome did 0.0.0.0 still routes for me on Linux (on both Chrome 40 and 41,) I would hate to break known and expected behavior on Linux for the sake of somebody who just random clicks considering we tell them that they are both available.

alfredxing:
I think @jglovier's second suggestion is good. Chrome for me (Mac) refuses to let me type in `http://0.0.0.0:4000`, instead doing a Google search.

parkr:
If we get into the technicalities, listening on `0.0.0.0:4000` is "all hosts", so `0.0.0.0` is not one IP but all of them, right? That said, I'd like it to be something people can type into their browsers. If you can't type in `http://0.0.0.0:4000` on all browsers, let's make it something else. I use `localhost:4000` myself (not `127.0.0.1`) so I am comfortable with that. I'm not sure we ever really need to share the `0.0.0.0` do we? It says "listening on everything" but perhaps people thought we were doing that already with 127.0.0.1.

envygeeks:
I think that's an acceptable compromise:
`Server address: http://localhost:4000 (listening on all interfaces)`

jglovier:
:point_up_2: :+1: 

I like removing the `0.0.0.0` entirely like @envygeeks last wording. Eliminates possible confusion, but still makes the point.

parkr:
> I think that's an acceptable compromise:
> Server address: http://localhost:4000 (listening on all interfaces)

:+1: from me!

envygeeks:
I'll make the first thing I do this weekend when I start doing my todo's for Jekyll then!

parkr:
> I'll make the first thing I do this weekend when I start doing my todo's for Jekyll then!

:+1:

envygeeks:
It's done, make your complaints so I can merge if none.

lib/jekyll/commands/serve.rb
```diff
@@ -125,15 +125,14 @@ def mime_types
...
-          [
-            "http://",
-            server.config[:BindAddress],
-            ":",
-            server.config[:Port],
-            baseurl || ""
-          ].map(&:to_s).join("")
+        def server_address_info(server, options)
+          addr = server.config[:BindAddress].to_s
+          burl = "#{options['baseurl']}/" if options['baseurl']
```
parkr: why not just `baseurl`? a little more direct for a variable name.

lib/jekyll/commands/serve.rb
```diff
@@ -125,15 +125,14 @@ def mime_types
...
+          ladr = addr == "0.0.0.0" ? "localhost" : a
```
parkr: `ladr` is also pretty opaque

parkr: also, what is `a`?

envygeeks: No more opaque then the code that was there before.

lib/jekyll/commands/serve.rb
```diff
@@ -125,15 +125,14 @@ def mime_types
...
+
+          rtn = "http://#{ladr}:#{server.config[:Port]}#{burl}"
+          rtn = "#{rtn} (listening on all interfaces)" if addr == "0.0.0.0"
+        rtn
```
parkr: i'd prefer not to use `rtn`, but to name it something that specifies what it is, rather than what it'll be used for in this method.

parkr: you could also potentially say

```ruby
full_addr = "http://#{listen_address}:#{server.config[:Port]}#{burl}"
if addr == "0.0.0.0"
  "#{full_addr} (listening on all interfaces)"
else
  full_addr
end
```

lib/jekyll/commands/serve.rb
envygeeks: I've updated it to use clearer variable names.

parkr:
?

envygeeks:
Something probably happened when I was leaving for the day, sometimes I accidentally press buttons because I forget to turn off the touchpad and I've one of those pressure touchpads that everybody hates (for good reason.)

envygeeks:
After long deliberation I would prefer to leave this as it is to avoid issues with Chrome and other places who are moving away from routing 0.0.0.0 I have decided to just orchestrate our Docker ( which will allow people do as they please with a single YAML file without us having to adjust the way we work.  It will also facilitate other ideas I have for the docker image.

parkr:
Ok, thanks for the update, @envygeeks!

PR #3544, opened 2015-03-04
Refactored url.rb to be compliant with Ruby Style Guide.

MartinRogalla:
- Single Quotes
 - Fixed typo in variable name.
 - Removed Redundant Escaping in Regular Expression.

Signed-off-by: Martin Jorn Rogalla <martin@martinrogalla.com>

### DIFF
--- a/lib/jekyll/url.rb
+++ b/lib/jekyll/url.rb
@@ -29,7 +29,7 @@ module Jekyll
       @permalink    = options[:permalink]
 
       if (@template || @permalink).nil?
-        raise ArgumentError, "One of :template or :permalink must be supplied."
+        raise ArgumentError, 'One of :template or :permalink must be supplied.'
       end
     end
 
@@ -44,12 +44,12 @@ module Jekyll
     #
     # Returns the _unsanitized String URL
     def generated_permalink
-      (@generated_permlink ||= generate_url(@permalink)) if @permalink
+      (@generated_permalink ||= generate_url(@permalink)) if @permalink
     end
 
     # Generates a URL from the template
     #
-    # Returns the _unsanitized String URL
+    # Returns the unsanitized String URL
     def generated_url
       @generated_url ||= generate_url(@template)
     end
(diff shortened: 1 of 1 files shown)


parkr:
To which ruby style guide do you refer?

envygeeks:
Make it all double quotes if you're gonna do anything about quotes because we've all strong opinions on code style.  So the quotes part gotta go IMO because I don't care what the Ruby Style Guide has to say about quotes at all.

The rest is fine IMO.

MartinRogalla:
@parkr I believe I might have had the wrong style guide. https://github.com/bbatsov/ruby-style-guide#strings

So what exactly should be within single quotes and what should be in double quotes?

willnorris:
Jekyll follows GitHub's ruby style guide: https://github.com/styleguide/ruby

envygeeks:
IMO leave quotes well along unless you are directly working on the code itself, there is no reason to enforce one type of quote if people prefer other types because of muscle memory.  This isn't C/++.  I always personally use double quotes for the same reason as Github and only use Single quotes when it enhances the readability of the source `'this is an "example"'` is an example of when I would use single quotes.  However, my preference aside, I'm not about to go and tell somebody to use one or the other because that's something that can be fixed later when the code is directly worked on indirect of the actual fix happening.

MartinRogalla:
@envygeeks Okay, I won't change the quotes which are already in place, but don't really understand the following line as it seems inconsistent to me:

``` ruby
# Append a trailing slash to the URL if the unsanitized URL had one
url << "/" if in_url[-1].eql?('/')
```

I won't change any of the quotes which are in place and use double quotes wherever possible.

envygeeks:
@MartinRogalla Yeah that line can definitely change if you want to change it!

willnorris:
``` ruby
# Append a trailing slash to the URL if the unsanitized URL had one
url << "/" if in_url[-1].eql?('/')
```

We've switched to using `end_with?` elsewhere, which would be a lot easier to read than this

MartinRogalla:
@willnorris thanks for the suggestion! Should be fixed.

parkr:
Thanks!!

PR #3545, opened 2015-03-04
Extract reading of data from `site.rb` to reduce responsibilities.

MartinRogalla:
I believe that the `site.rb` class right now has too many responsibilities. In this pull request I want to try and take out all the data reading tasks. I took out every single method that I found to have the *reading of data* responsibility and put it in a new class called Reader(`reader.rb`).

My methodology per method:

1. Copy method from `site.rb` and fix references inside of the method.
2. Redirect old method to new method in `reader.rb`
3. Check if the test suite is passing
3. Update all references to the old `site.rb` method  and refer them to the new `reader.rb` method.
4. Run the test suite
5. Commit

Do you think this is a good idea at all or would you like it to be done differently? 

I'm open to any criticism you might have.

**Note**: *The usage of quotes is probably incorrect, I will fix these as soon as I know which style I should adhere to.*


MartinRogalla:
Hmm, I didn't see that the cucumber test failed. Working on it right now.

MartinRogalla:
Should be working now.

parkr:
I'd rather not move `in_source_dir` and `in_dest_dir` out – can you add a def delegator for that method? Otherwise we'd have to update a large number of plugins...

MartinRogalla:
After looking at those methods a bit more carefully I spotted that `in_(source/dest)_dir` is used by multiple classes and not just the reader. Therefore I moved those two functions back to `site.rb`. Solves the problem @parkr mentioned as well.

MartinRogalla:
Separated some more reader components into their own classes. I'm looking for some feedback on my progress. Don't hesitate to give me some critique ;)

lib/jekyll.rb
```diff
@@ -49,12 +49,19 @@ module Jekyll
...
   autoload :FrontmatterDefaults, 'jekyll/frontmatter_defaults'
   autoload :Layout,              'jekyll/layout'
-  autoload :LayoutReader,        'jekyll/layout_reader'
+  autoload :CollectionReader,        'jekyll/readers/dynamic/collection_reader'
+  autoload :DataReader,        'jekyll/readers/dynamic/data_reader'
+  autoload :LayoutReader,        'jekyll/readers/dynamic/layout_reader'
+  autoload :DraftReader,        'jekyll/readers/dynamic/draft_reader'
+  autoload :PostReader,        'jekyll/readers/dynamic/post_reader'
+  autoload :PageReader, 'jekyll/readers/dynamic/page_reader'
+  autoload :StaticFileReader,        'jekyll/readers/static/static_file_reader'
```
parkr: Please align all the strings vertically :smile:

lib/jekyll/reader.rb
```diff
@@ -0,0 +1,113 @@
...
+    end
+
+    # Read Site data from disk and load it into internal data structures.
+    #
+    # Returns nothing.
+    def read
+      @site.layouts = LayoutReader.new(site).read
+      read_directories
+      @site.data = DataReader.new(site).read(site.config['data_source'])
+      CollectionReader.new(site).read
```
parkr: why not sticking this into `@site.collections`?

test/test_tags.rb
```diff
@@ -13,7 +13,7 @@ def create_post(content, override = {}, converter_class = Jekyll::Converters::Ma
     site = Site.new(Jekyll.configuration)
 
     if override['read_posts']
-      site.read_posts('')
+      site.posts += PostReader.new(site).read('')
```
parkr: I'm not a big fan of `+=` generally... hm.

site/_docs/history.md
```diff
@@ -191,6 +191,7 @@ permalink: "/docs/history/"
 - Allow Travis to 'parallelize' our tests ([#2859]({{ site.repository }}/issues/2859))
 - Fix test for Liquid rendering in Sass ([#2856]({{ site.repository }}/issues/2856))
 - Fixing "vertycal" typo in site template's `_base.scss` ([#2889]({{ site.repository }}/issues/2889))
+- Convert remaining textile test documents to markdown ([#3528]({{ site.repository }}/issues/3528))
```
parkr: ? is this from a rebase? haha

parkr:
I'm generally pretty happy with this, actually. Nice work. 

/cc @jekyll/core, what do you think?

lib/jekyll.rb
MartinRogalla: Haha, sorry. Will be fixed.

site/_docs/history.md
MartinRogalla: Yes, sorry about that. Haha.

envygeeks:
I have no problems with this. :+1:

lib/jekyll/reader.rb
MartinRogalla: I don't understand. What do you mean? Do you mean to re-read the collections every time the `@site.collections` is called?

test/test_tags.rb
MartinRogalla: I replaced all the `+=` occurrences with concatenations. Apparently `+=` creates a new array every time it is used, while `concat` simply appends the new elements to the old array.

[Reference](http://stackoverflow.com/questions/23763215/ruby-array-concat-versus-speed)

envygeeks: In the context of optimization (as you implied) that matters very little to Jekyll (as in the type of app we are not the type of programmers we are but, really, it should be applied to both.)  We are a short lived process and our GC  time matters little  we probably don't even hit the exponential scale required of caring about the GC time and whether or not we are creating new array's and marking old objects.

parkr:
One last set of eyes. @mattr-, would you mind taking a look?

lib/jekyll.rb
```diff
@@ -49,12 +49,19 @@ module Jekyll
...
   autoload :FrontmatterDefaults, 'jekyll/frontmatter_defaults'
   autoload :Layout,              'jekyll/layout'
-  autoload :LayoutReader,        'jekyll/layout_reader'
+  autoload :CollectionReader,    'jekyll/readers/dynamic/collection_reader'
+  autoload :DataReader,          'jekyll/readers/dynamic/data_reader'
+  autoload :LayoutReader,        'jekyll/readers/dynamic/layout_reader'
+  autoload :DraftReader,         'jekyll/readers/dynamic/draft_reader'
+  autoload :PostReader,          'jekyll/readers/dynamic/post_reader'
+  autoload :PageReader,          'jekyll/readers/dynamic/page_reader'
+  autoload :StaticFileReader,    'jekyll/readers/static/static_file_reader'
```
mattr-: Why the distinction between dynamic vs. static? Why add that level of indirection? If we had more static readers, I wouldn't be asking about this, but it feels like YAGNI for just the one.

lib/jekyll/reader.rb
```diff
@@ -0,0 +1,113 @@
...
+    # that will become part of the site according to the rules in
+    # filter_entries.
+    #
+    # dir - The String relative path of the directory to read. Default: ''.
+    #
+    # Returns nothing.
+    def read_directories(dir = '')
+      retrieve_posts(dir)
+
+      # Obtain sub-directories in order to recursively read them.
```
mattr-: Comments in methods indicate to me code that is need of further refactoring towards expressiveness. What about a more functional style approach to filtering out the directories?

lib/jekyll/reader.rb
```diff
@@ -0,0 +1,113 @@
...
+      base = site.in_source_dir(dir)
+      dot = Dir.chdir(base) { filter_entries(Dir.entries('.'), base) }
+      dot_dirs = dot.select{ |file| File.directory?(@site.in_source_dir(base,file)) }
+      retrieve_dirs(base, dir, dot_dirs)
+
+      dot_files = (dot - dot_dirs)
+
+      # Obtain all the pages.
```
mattr-: Same suggestion with the comment as above.

lib/jekyll/reader.rb
```diff
@@ -0,0 +1,113 @@
...
+      retrieve_pages(dir, dot_pages)
+
+      # Assume the remaining files to be static files.
+      dot_static_files = dot_files - dot_pages
+      retrieve_static_files(dir, dot_static_files)
+    end
+
+    # Retrieves all the posts(posts/drafts) from the given directory
+    # and add them to the site and sort them.
+    #
```
mattr-: don't forget to document variables in your TomDoc. :smiley:

lib/jekyll/site.rb
```diff
@@ -300,7 +146,7 @@ def generate
     #
     # Returns nothing.
     def render
-      relative_permalinks_deprecation_method
+      deprecated_rel_permalink
```
mattr-: `deprecated_relative_permalinks` or `relative_permalinks_are_deprecated` are better name suggestions here, IMHO.

lib/jekyll/site.rb
```diff
@@ -447,29 +281,11 @@ def instantiate_subclasses(klass)
...
-      base = in_source_dir(dir, subfolder)
-      return [] unless File.exist?(base)
-      entries = Dir.chdir(base) { filter_entries(Dir['**/*'], base) }
-      entries.delete_if { |e| File.directory?(in_source_dir(base, e)) }
-    end
-
-    # Aggregate post information
-    #
-    # post - The Post object to aggregate information for
+    # Warns the user if permanent links are relative tot the parent
```
mattr-: typo here: `to` instead of `tot` :smiley_cat:

lib/jekyll/site.rb
```diff
@@ -479,16 +295,23 @@ def relative_permalinks_deprecation_method
       end
     end
 
+    # Get the to be written documents
+    #
+    # Returns an Array of Documents which should be written
```
mattr-: Love the new TomDoc! :heart:

lib/jekyll/site.rb
```diff
@@ -479,16 +295,23 @@ def relative_permalinks_deprecation_method
...
     def docs_to_write
       documents.select(&:write?)
     end
 
+    # Get all the documents
+    #
+    # Returns an Array of all Documents
```
mattr-: And here too! :purple_heart:

test/test_entry_filter.rb
```diff
@@ -19,7 +19,7 @@ class TestEntryFilter < JekyllUnitTest
       files = %w[index.html site.css .htaccess vendor]
 
       @site.exclude = excludes + ["exclude*"]
-      assert_equal files, @site.filter_entries(excludes + files + ["excludeA"])
+      assert_equal files, @site.reader.filter_entries(excludes + files + ["excludeA"])
```
mattr-: Can we not just instantiate a reader directly in these tests?

mattr-:
This is great stuff. I see places where you could push the refactoring further, but I think more pull requests can do that vs. holding this one up. I wrote a few questions, but I don't think they block merging this in.

I know you ended up moving stuff around quite a bit, but there's a bit of unnecessary noise in the diff (like the moving around of `limit_posts` for example). Not a huge issue, but something to be aware of, as that can make things harder to review.

All in all, I say :shipit:

lib/jekyll/reader.rb
mattr-: The other readers have their output assigned, whereas `CollectionReader` manipulates `@site.collections` directly. That's probably a function of how collections are written more than anything else, but it does break the pattern in this code.

parkr:
Thank goodness for @mattr- :pray:

lib/jekyll.rb
MartinRogalla: Good point. I thought there were going to be more static readers. Therefore initially I split it up in the two directories. But yes you're right it's a bit overdone like this. I'll change it now.

lib/jekyll/site.rb
MartinRogalla: Fixed.

lib/jekyll/site.rb
MartinRogalla: Fixed.

lib/jekyll/reader.rb
MartinRogalla: Fixed.

test/test_entry_filter.rb
MartinRogalla: This would require decoupling of the Reader and EntryFilter from the site. I do think this should be done, but should be left for future work, as it is quite some work and this pull request will become too large in my opinion.

mattr-: Agreed. Was just something to think about for the future. :smiley:

MartinRogalla:
Made the corrections as they were suggested by @mattr- in 63a1ec8.

mattr-:
:heart: :sparkling_heart: :blue_heart: :purple_heart: :green_heart:

:shipit:

alfredxing:
Looks good! :cool:

MartinRogalla:
Is there anything I should do on my end, before this can get merged? :blush:

parkr:
:tada:

PR #3546, opened 2015-03-04
Improve consistency in site/_docs/plugins.md

capnfabs:
Three plugin strategies are listed, but then only two were referenced in the following note. I've updated this doc so that all three are referenced.


parkr:
Thanks!

PR #3547, opened 2015-03-05
Resolve warnings

yous:
I noticed Jekyll generates some warnings on Travis build: https://travis-ci.org/yous/yous.github.io/builds/52976809

- Remove duplicated range from regex

    `/\w/` is equivalent to `[a-zA-Z0-9_]`. See http://ruby-doc.org/core-2.2.0/doc/regexp_rdoc.html#label-Character+Classes.
- Replace `File.exists?` with `File.exist?`

    `File.exists?` is deprecated method. See http://ruby-doc.org//core-2.2.0/File.html#exists-3F-method.


envygeeks:
:+1:

alfredxing:
:+1: Thanks!

PR #3552, opened 2015-03-07
Use rspec-mock instead of rr

### LINKED ISSUE #3549: Remove rr gem
RR is not supported anymore =(
https://github.com/rr/rr/commit/885885ca0cd426ea5c35fa3851abefe3d3a78428

I am trying to remove it from rubygems.org too, not sure if the best option is to use mocha. Anyhow, we probably should remove it from jekyll too.

arthurnn:
`rr` gem is not maintained anymore. We can replace it with rspec-mock, and stop using mocking in a few spots.

review @parkr @envygeeks 
[fixes #3549]

### DIFF
--- a/test/test_command.rb
+++ b/test/test_command.rb
@@ -4,7 +4,7 @@ class TestCommand < JekyllUnitTest
   context "when calling .add_build_options" do
     should "add common options" do
       cmd = Object.new
-      mock(cmd).option.with_any_args.at_least(1)
+      expect(cmd).to receive(:option).at_least(:once)
       Command.add_build_options(cmd)
     end
   end
@@ -12,7 +12,7 @@ class TestCommand < JekyllUnitTest
     context "when fatal error occurs" do
       should "exit with non-zero error code" do
         site = Object.new
-        stub(site).process { raise Jekyll::Errors::FatalException }
+        def site.process; raise Jekyll::Errors::FatalException; end
         error = assert_raises(SystemExit) { Command.process_site(site) }
         refute_equal 0, error.status
       end
--- a/test/test_related_posts.rb
+++ b/test/test_related_posts.rb
@@ -3,11 +3,7 @@ require 'helper'
 class TestRelatedPosts < JekyllUnitTest
   context "building related posts without lsi" do
     setup do
-      stub(Jekyll).configuration do
-        Jekyll::Configuration::DEFAULTS.merge({'source' => source_dir,
-                                               'destination' => dest_dir})
-      end
-      @site = Site.new(Jekyll.configuration)
+      @site = fixture_site
     end
 
     should "use the most recent posts for related posts" do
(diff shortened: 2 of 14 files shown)


test/test_command.rb
```diff
@@ -12,7 +12,7 @@ class TestCommand < JekyllUnitTest
     context "when fatal error occurs" do
       should "exit with non-zero error code" do
         site = Object.new
-        stub(site).process { raise Jekyll::Errors::FatalException }
+        def site.process; raise Jekyll::Errors::FatalException; end
```
parkr: ?!

test/test_related_posts.rb
```diff
@@ -3,11 +3,9 @@
 class TestRelatedPosts < JekyllUnitTest
   context "building related posts without lsi" do
     setup do
-      stub(Jekyll).configuration do
-        Jekyll::Configuration::DEFAULTS.merge({'source' => source_dir,
+      config = Jekyll::Configuration::DEFAULTS.merge({'source' => source_dir,
                                                'destination' => dest_dir})
-      end
-      @site = Site.new(Jekyll.configuration)
+      @site = Site.new(config)
```
parkr: This can all be simplified down to:

```ruby
@site = fixture_site
```

it uses `source_dir` and `dest_dir` by default.

test/test_related_posts.rb
```diff
@@ -24,23 +22,20 @@ class TestRelatedPosts < JekyllUnitTest
...
-      stub(Jekyll).configuration do
-        Jekyll::Configuration::DEFAULTS.merge({'source' => source_dir,
+      config = Jekyll::Configuration::DEFAULTS.merge({'source' => source_dir,
                                                'destination' => dest_dir,
                                                'lsi' => true})
-      end
-      any_instance_of(Jekyll::RelatedPosts) { |i| stub(i).display }
-      @site = Site.new(Jekyll.configuration)
+      allow_any_instance_of(Jekyll::RelatedPosts).to receive(:display)
+      @site = Site.new(config)
```
parkr: Then this can be

```ruby
@site = fixture_site({"lsi": => true})
```

test/test_command.rb
arthurnn: no need to `stub` here, you can just re-define the method and make it raise.

arthurnn: btw, we do that a lot on Rails test suite.

test/test_tags.rb
```diff
@@ -5,12 +5,7 @@
 class TestTags < JekyllUnitTest
 
   def create_post(content, override = {}, converter_class = Jekyll::Converters::Markdown)
-    stub(Jekyll).configuration do
-      site_configuration({
-        "highlighter" => "rouge"
-      }.merge(override))
-    end
-    site = Site.new(Jekyll.configuration)
+    site = fixture_site({"highlighter" => "rouge"}.merge(override))
```
parkr: this won't deep-merge, not sure if it's needed

test/test_tags.rb
```diff
@@ -681,11 +672,10 @@ def highlight_block_with_opts(options_string)
 
   context "relative include tag with variable and liquid filters" do
     setup do
-      stub(Jekyll).configuration do
+      configuration =
         site_configuration({'pygments' => true})
-      end
 
-      site = Site.new(Jekyll.configuration)
+      site = Site.new(configuration)
```
parkr: let's use `fixture_site` here as well :)

parkr:
Thank you, @arthurnn!!!!!!

test/test_tags.rb
arthurnn: I just kept the way it was before.
So you are saying we dont need the `{"highlighter" => "rouge"}` part?

parkr: i'm an idiot, there's nothing to deep-merge :) don't mind me...

test/test_command.rb
parkr: is this reset later?

arthurnn: actually that method will be only defined to that instance. `site` is not shared with other tests so it is fine.

arthurnn:
You are welcome.. thanks for Jekyll !
:tada: everything is :green_heart:  now.

test/test_command.rb
parkr: This is totally new to me, very awesome!!

parkr:
LGTM! :shipit:

envygeeks:
Sorry for saying thanks late! :heart:

PR #3553, opened 2015-03-07
prototype of jekyll hooks, encapsulated

stevecrozz:
This is a second attempt at #3498 (adding hooks to Jekyll). In my last attempt @envygeeks had some complaints about the lack of encapsulation. This PR is an attempt to address those concerns and try out a new idea. This is definitely not ready to merge, but I wanted to get some eyes on it before I implement and write tests for the whole thing. The hooks concept is inspired by, but in this instance not perfectly compatible with octopress-hooks, so it's a bit of a departure from #3392 by @parkr, but I think it nicely encapsulates the hook logic.

Here's a practical example of how you could use one of these hooks to do something useful: https://github.com/stevecrozz/featherblade

### DIFF
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -175,8 +175,6 @@ module Jekyll
     #
     # Returns nothing.
     def write(dest)
-      Jekyll::Hooks.trigger self, :post_render
-
       path = destination(dest)
       FileUtils.mkdir_p(File.dirname(path))
       File.open(path, 'wb') do |f|
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -175,11 +175,15 @@ module Jekyll
     #
     # Returns nothing.
     def write(dest)
+      Jekyll::Hooks.trigger self, :post_render
+
       path = destination(dest)
       FileUtils.mkdir_p(File.dirname(path))
       File.open(path, 'wb') do |f|
         f.write(output)
       end
+
+      Jekyll::Hooks.trigger self, :post_write
     end
 
     # Returns merged option hash for File.read of self.site (if exists)
--- a/lib/jekyll/hooks.rb
+++ b/lib/jekyll/hooks.rb
@@ -1,0 +1,53 @@
+module Jekyll
+  module Hooks
+    # initial empty hooks
+    @registry = {
+      Jekyll::Site => {
+        reset: [],
+        post_read: [],
+        pre_render: [],
+        post_write: [],
+      },
+      Jekyll::Page => {
+        post_init: [],
+        pre_render: [],
+        post_render: [],
+        post_write: [],
+      },
+      Jekyll::Post => {
+        post_init: [],
+        pre_render: [],
+        post_render: [],
+        post_write: [],
+      },
+    }
+
+    NotAvailable = Class.new(RuntimeError)
+
+    # register a hook to be called later
+    def self.register(klass, event, &block)
+      unless @registry[klass]
+        raise NotAvailable, "Hooks are only available for the following " <<
+          "classes: #{@registry.keys.inspect}"
+      end
+
+      unless @registry[klass][event]
+        raise NotAvailable, "Invalid hook. #{klass} supports only the " <<
+          "following hooks #{@registry[klass].keys.inspect}"
+      end
+
+      @registry[klass][event] << block
+    end
+
+    # interface for Jekyll core components to trigger hooks
+    def self.trigger(instance, event, *args)
+      # proceed only if there are hooks to call
+      return unless @registry[instance.class]
+      return unless @registry[instance.class][event]
+
+      @registry[instance.class][event].each do |hook|
+        hook.call(instance, *args)
+      end
+    end
+  end
+end
--- a/lib/jekyll/hooks.rb
+++ b/lib/jekyll/hooks.rb
@@ -20,7 +20,7 @@ module Jekyll
     # initial empty hooks
     @registry = {
       :site => {
-        reset: [],
+        after_reset: [],
         post_read: [],
         pre_render: [],
         post_write: [],
@@ -48,23 +48,23 @@ module Jekyll
     @hook_priority = {}
 
     NotAvailable = Class.new(RuntimeError)
+    Uncallable = Class.new(RuntimeError)
 
-    # register hook(s) to be called later
-    def self.register(owners, event, priority: nil, &block)
+    # register hook(s) to be called later, public API
+    def self.register(owners, event, priority: DEFAULT_PRIORITY, &block)
       Array(owners).each do |owner|
-        register_one(owner, event, priority: priority_value(priority), &block)
+        register_one(owner, event, priority_value(priority), &block)
       end
     end
 
     # Ensure the priority is a Fixnum
-    def self.priority_value(priority=nil)
-      return DEFAULT_PRIORITY unless priority
+    def self.priority_value(priority)
       return priority if priority.is_a?(Fixnum)
       PRIORITY_MAP[priority] || DEFAULT_PRIORITY
     end
 
-    # register a single hook to be called later
-    def self.register_one(owner, event, priority: nil, &block)
+    # register a single hook to be called later, internal API
+    def self.register_one(owner, event, priority, &block)
       unless @registry[owner]
         raise NotAvailable, "Hooks are only available for the following " <<
           "classes: #{@registry.keys.inspect}"
@@ -73,6 +73,10 @@ module Jekyll
       unless @registry[owner][event]
         raise NotAvailable, "Invalid hook. #{owner} supports only the " <<
           "following hooks #{@registry[owner].keys.inspect}"
+      end
+
+      unless block.respond_to? :call
+        raise Uncallable, "Hooks must respond to :call"
       end
 
       insert_hook owner, event, priority, &block
--- a/lib/jekyll/hooks.rb
+++ b/lib/jekyll/hooks.rb
@@ -1,51 +1,101 @@
 module Jekyll
   module Hooks
+    # Helps look up hooks from the registry by owner's class
+    OWNER_MAP = {
+      Jekyll::Site => :site,
+      Jekyll::Page => :page,
+      Jekyll::Post => :post,
+      Jekyll::Document => :document,
+    }.freeze
+
+    DEFAULT_PRIORITY = 20
+
+    # compatibility layer for octopress-hooks users
+    PRIORITY_MAP = {
+      low: 10,
+      normal: 20,
+      high: 30,
+    }.freeze
+
     # initial empty hooks
     @registry = {
-      Jekyll::Site => {
+      :site => {
         reset: [],
         post_read: [],
         pre_render: [],
         post_write: [],
       },
-      Jekyll::Page => {
+      :page => {
         post_init: [],
         pre_render: [],
         post_render: [],
         post_write: [],
       },
-      Jekyll::Post => {
+      :post => {
         post_init: [],
+        pre_render: [],
+        post_render: [],
+        post_write: [],
+      },
+      :document => {
         pre_render: [],
         post_render: [],
         post_write: [],
       },
     }
 
+    # map of all hooks and their priorities
+    @hook_priority = {}
+
     NotAvailable = Class.new(RuntimeError)
 
-    # register a hook to be called later
-    def self.register(klass, event, &block)
-      unless @registry[klass]
+    # register hook(s) to be called later
+    def self.register(owners, event, priority: nil, &block)
+      Array(owners).each do |owner|
+        register_one(owner, event, priority: priority_value(priority), &block)
+      end
+    end
+
+    # Ensure the priority is a Fixnum
+    def self.priority_value(priority=nil)
+      return DEFAULT_PRIORITY unless priority
+      return priority if priority.is_a?(Fixnum)
+      PRIORITY_MAP[priority] || DEFAULT_PRIORITY
+    end
+
+    # register a single hook to be called later
+    def self.register_one(owner, event, priority: nil, &block)
+      unless @registry[owner]
         raise NotAvailable, "Hooks are only available for the following " <<
           "classes: #{@registry.keys.inspect}"
       end
 
-      unless @registry[klass][event]
-        raise NotAvailable, "Invalid hook. #{klass} supports only the " <<
-          "following hooks #{@registry[klass].keys.inspect}"
+      unless @registry[owner][event]
+        raise NotAvailable, "Invalid hook. #{owner} supports only the " <<
+          "following hooks #{@registry[owner].keys.inspect}"
       end
 
-      @registry[klass][event] << block
+      insert_hook owner, event, priority, &block
+    end
+
+    def self.insert_hook(owner, event, priority, &block)
+      @hook_priority[block] = "#{priority}.#{@hook_priority.size}".to_f
+      @registry[owner][event] << block
     end
 
     # interface for Jekyll core components to trigger hooks
     def self.trigger(instance, event, *args)
-      # proceed only if there are hooks to call
-      return unless @registry[instance.class]
-      return unless @registry[instance.class][event]
+      owner_symbol = OWNER_MAP[instance.class]
 
-      @registry[instance.class][event].each do |hook|
+      # proceed only if there are hooks to call
+      return unless @registry[owner_symbol]
+      return unless @registry[owner_symbol][event]
+
+      # hooks to call for this owner and event
+      hooks = @registry[owner_symbol][event]
+
+      # sort and call hooks according to priority and load order
+      hooks.sort_by { |h| @hook_priority[h] }.each do |hook|
         hook.call(instance, *args)
       end
     end
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -160,6 +160,7 @@ module Jekyll
         collection.docs.each do |document|
           if regenerator.regenerate?(document)
             document.output = Jekyll::Renderer.new(self, document, payload).run
+            Jekyll::Hooks.trigger document, :post_render
           end
         end
       end
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -75,6 +75,8 @@ module Jekyll
       if limit_posts < 0
         raise ArgumentError, "limit_posts must be a non-negative number"
       end
+
+      Jekyll::Hooks.trigger self, :reset
     end
 
     # Load necessary libraries, plugins, converters, and generators.
@@ -132,6 +134,7 @@ module Jekyll
     def read
       reader.read
       limit_posts!
+      Jekyll::Hooks.trigger self, :post_read
     end
 
     # Run each of the Generators.
@@ -150,6 +153,9 @@ module Jekyll
       relative_permalinks_are_deprecated
 
       payload = site_payload
+
+      Jekyll::Hooks.trigger self, :pre_render, payload
+
       collections.each do |label, collection|
         collection.docs.each do |document|
           if regenerator.regenerate?(document)
@@ -158,7 +164,6 @@ module Jekyll
         end
       end
 
-      payload = site_payload
       [posts, pages].flatten.each do |page_or_post|
         if regenerator.regenerate?(page_or_post)
           page_or_post.render(layouts, payload)
@@ -183,6 +188,7 @@ module Jekyll
         item.write(dest) if regenerator.regenerate?(item)
       }
       regenerator.write_metadata
+      Jekyll::Hooks.trigger self, :post_write
     end
 
     # Construct a Hash of Posts indexed by the specified Post attribute.
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -76,7 +76,7 @@ module Jekyll
         raise ArgumentError, "limit_posts must be a non-negative number"
       end
 
-      Jekyll::Hooks.trigger self, :reset
+      Jekyll::Hooks.trigger self, :after_reset
     end
 
     # Load necessary libraries, plugins, converters, and generators.
--- a/site/_docs/plugins.md
+++ b/site/_docs/plugins.md
@@ -520,7 +520,7 @@ The complete list of available hooks is below:
         <p><code>:site</code></p>
       </td>
       <td>
-        <p><code>:reset</code></p>
+        <p><code>:after_reset</code></p>
       </td>
       <td>
         <p>Just after site reset</p>
--- a/site/_docs/plugins.md
+++ b/site/_docs/plugins.md
@@ -491,16 +491,16 @@ custom functionality every time Jekyll renders a post, you could register a
 hook like this:
 
 {% highlight ruby %}
-Jekyll::Hooks.register Jekyll::Post, :post_render do |post|
+Jekyll::Hooks.register :post, :post_render do |post|
   # code to call after Jekyll renders a post
 end
 {% endhighlight %}
 
-Jekyll provides hooks for <code>Jekyll::Site</code>, <code>Jekyll::Page</code>
-and <code>Jekyll::Post</code>. In all cases, Jekyll calls your hooks with the
-container object as the first callback parameter. But in the case of
-<code>:pre_render</code>, your hook will also receive a payload hash as a
-second parameter which allows you full control over the variables that are
+Jekyll provides hooks for <code>:site</code>, <code>:page</code>,
+<code>:post</code>, and <code>:document</code>. In all cases, Jekyll calls your
+hooks with the container object as the first callback parameter. But in the
+case of <code>:pre_render</code>, your hook will also receive a payload hash as
+a second parameter which allows you full control over the variables that are
 available while rendering.
 
 The complete list of available hooks is below:
@@ -517,7 +517,7 @@ The complete list of available hooks is below:
   <tbody>
     <tr>
       <td>
-        <p><code>Jekyll::Site</code></p>
+        <p><code>:site</code></p>
       </td>
       <td>
         <p><code>:reset</code></p>
@@ -528,7 +528,7 @@ The complete list of available hooks is below:
     </tr>
     <tr>
       <td>
-        <p><code>Jekyll::Site</code></p>
+        <p><code>:site</code></p>
       </td>
       <td>
         <p><code>:pre_render</code></p>
@@ -539,7 +539,7 @@ The complete list of available hooks is below:
     </tr>
     <tr>
       <td>
-        <p><code>Jekyll::Site</code></p>
+        <p><code>:site</code></p>
       </td>
       <td>
         <p><code>:post_render</code></p>
@@ -550,7 +550,7 @@ The complete list of available hooks is below:
     </tr>
     <tr>
       <td>
-        <p><code>Jekyll::Site</code></p>
+        <p><code>:site</code></p>
       </td>
       <td>
         <p><code>:post_write</code></p>
@@ -561,7 +561,7 @@ The complete list of available hooks is below:
     </tr>
     <tr>
       <td>
-        <p><code>Jekyll::Page</code></p>
+        <p><code>:page</code></p>
       </td>
       <td>
         <p><code>:post_init</code></p>
@@ -572,7 +572,7 @@ The complete list of available hooks is below:
     </tr>
     <tr>
       <td>
-        <p><code>Jekyll::Page</code></p>
+        <p><code>:page</code></p>
       </td>
       <td>
         <p><code>:pre_render</code></p>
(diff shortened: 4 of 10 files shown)


parkr:
Don't worry about the circleci failure.

This looks pretty good! @envygeeks?

envygeeks:
:-1: I like it.

envygeeks:
Sorry about the :-1: I actually meant to do :+1:

stevecrozz:
Great. I'll flesh out the rest of it and add some test coverage.

stevecrozz:
This is a bit more fleshed out now, although still lacking in tests and docs. I did do some one-off tests and it all seems to be working pretty well. Every hook receives the object being "hooked into" as the first argument.

One major difference from octopress-hooks worth pointing out is how I'm dealing with the payload objects just before rendering.

In octopress-hooks, there's a special hook for merging in new payload params. But I took a different approach and just pass the payload as an argument to each of the pre_render hooks. Anyone who implements a pre_render hook will receive the payload object and can modify it there if necessary as well as do any other things that need doing before rendering. pre_render is the only hook that receives a second argument at the moment. @imathis: let me know if you have any thoughts on this, especially if you already rejected this idea for some reason.

Next up is tests and docs, but I'll give this PR a few days to bake in case there's more feedback.

stevecrozz:
I'm working on cucumber tests now. How thorough do these tests need to be? Do we want each hook for every container to have its own test scenario?

envygeeks:
I say just write them and if there is a problem we'll point it out!

stevecrozz:
@envygeeks / @parkr I've got cucumber tests and docs written. Do you have any feedback?

site/_docs/plugins.md
```diff
@@ -469,6 +469,179 @@ module Jekyll
 end
 {% endhighlight %}
 
+## Hooks
+
```
parkr: Can you add an "unreleased" note here? Check out #3556 for an example.

parkr:
Unless you have some thoughts, @envygeeks, let's merge and get this out there in a new beta so people can start building for it.

stevecrozz:
Amended commit to include "unreleased" note in the docs

alfredxing:
:+1: Nice work on this! We're missing hooks for Documents but that shouldn't be a huge issue for now.

imathis:
I'm sorry I'm a bit late to this.

Unless I've overlooked it, this does not include:
- Site: pre_read, merge_payload
- Page/Post: merge_payload

For reference here's Octopress's [hook timeline](https://github.com/octopress/hooks#hook-timeline).

stevecrozz:
@imathis That's correct. I removed pre_read from my PR while writing the test cases because I couldn't come up with a use case to test that wasn't already covered. There may be some good ones, but I didn't think of any.

As for merge_payload, I did depart from the octopress-hooks approach which is why I was interested in your feedback. I tried to cover the use case of altering the payload using pre_render which passes the payload to the callback as its second parameter. There's less code and less to test in this approach and it seems more flexible, but I'd be interested to know if I missed a good reason to implement it the way you already did.

parkr:
@imathis ^^?

imathis:
I'm going to create a branch of octopress-hooks that translates hooks to be based on this. If my tests pass. :+1: Let's see how long this takes :)

envygeeks:
Can we focus more on what we need and less on what we have?

imathis:
@envygeeks I'm focusing on what I need. I built the features in Octopress hooks because I needed them. If I can keep that functionality, I'll be good to go. Otherwise I'll have to keep monkey patching my own hooks.

envygeeks:
@imathis Well if you run into trouble then please do post the hooks that are failing and where they originate and are used at so I might be able to help you two come to compromise by possibly reworking both sides to see if we can fit your hooks into his hook system or if we need to truly have him change his hook system.

imathis:
@envygeeks Thanks. I don't anticipate having to rework this system. I think most of the stuff I'm doing should map to these pretty well. I sure like this code better though. I do have one question though. The way I originally wrote the hooks, they were extended from `Jekyll::Plugin`. This meant I could set priority for each hook, a feature I've found pretty useful for some of the plugins I've written. As far as I can tell, I'd set the priority on the plugin. If I needed different priority hooks, I'd just have to create separate plugins for them.

imathis:
So far this is working beautifully. I've had one issue, though. After I set up a test for page hooks, I discovered that the incremental regeneration cache was preventing the hook from affecting the page. I cleared the cache and then it worked. I have a feeling that, "Have you tried deleting your Jekyll cache?" may become a common troubleshooting question.

imathis:
I've rewritten my Octopress Hooks tests to use this instead of Octopress Hooks and it's awesome. The two things I lack, which I'd really like are:

1. Collection hooks.
2. All hooks (a way to write a hook that runs on any page type (Page, Post, Collection)

I definitely think Collection hooks are necessary, "All Hooks" are somewhat less necessary, but in practice the ability to tell Jekyll to process all content, avoids duplication and reduces bugs in plugins. I'm not sure how you'd implement it, but it would be really useful.

Thanks so much for this. It's way nicer to use than Octopress Hooks. :+1:

alfredxing:
Incremental regeneration doesn't render or write pages/posts if they aren't changed. We can move hooks like `post_render` and `post_write` so they get triggered, or add new hooks that get called even when posts/pages are not rendered.

parkr:
> "All Hooks" are somewhat less necessary, but in practice the ability to tell Jekyll to process all content, avoids duplication and reduces bugs in plugins

Eventually `Document` will be the main class; `Page` and `Post` will no longer exist at that point. Adding a hook to `Document` will take care of this. What will be harder, is getting more specific. How do you attach a filter to just one collection? Have the method only run if the doc's collection is X? Perhaps that will work for now.

stevecrozz:
@imathis : @envygeeks and I were just talking about the priority idea. In order to do it the way that seems to make the most sense to us, we'll need a way to apply a name to a hook. So we have some changes to this PR in mind to support this. Roughly, hooks would go back to being classes, as they are in octopress-hooks, and they would implement a 'run' method and have a name (the name being the name of the class):

```ruby
class MyHook
  def run(page); end
end

class MyOtherHook
  def run(page); end
end

Jekyll::Hooks.register Jekyll::Page, :post_render, MyHook

# insert one hook before another, hook specified as a symbol here in case MyHook
# is not yet defined (due to not yet having been loaded)
Jekyll::Hooks.register Jekyll::Page, :post_render, MyOtherHook, before: :MyHook
```

I agreed to start working on this idea. As always I do appreciate feedback on this. Let me know if this looks good or bad to any of you.

@imathis I have no objections to collection hooks. If you can contribute to the cucumber tests / implementation / docs here I think we can include those.

imathis:
I'd be happy to contribute the Collections stuff. 

On the priority, I'm looking for a way of nudging global priority like how Jekyll Plugins have the `:highest, :high, :normal, :low, :lowest`. I use this for [octopress-paginate](https://github.com/octopress/paginate/blob/master/lib/octopress-paginate/hooks.rb#L4) to ensure that other hooks have run their transformations before the paginate plugin merges the page's payload with the pagination values.

I use them again in [octopress-multilingual](https://github.com/octopress/multilingual/blob/master/lib/octopress-multilingual/hooks.rb#L4) where I try to merge page and site payloads before other hooks, so that other hooks can work with multilingual features.

I'm sure users could manually set priority by setting the load order of their gems, but it would be ideal if people didn't have to think about load order in order to use a plugin, but that would be the fallback strategy if plugins had competing priority which caused a conflict. Additionally being able to set priority like this allows user plugins written in the `_plugins` directory to participate in the priority chain.

Do you think this will take a lot of effort to do? Does it seem reasonable?

envygeeks:
Doing it your way would drastically simplify the way the code has to work compared to the way we discussed last night where we have named hook points and a before or after, if we just have a numbered priority set then sure it's tons easier to select on.

stevecrozz:
It would certainly be an easier change if hooks can remain as plain Ruby callables and we don't wrap them in classes. Should I just add a 'priority' keyword argument to `Jekyll::Hooks::register`?

envygeeks:
@stevecrozz I think so, easier to do, less work and suites his needs without breaking much!

stevecrozz:
One more thing. I'm not certain the names `:highest, :high, :normal, :low, :lowest` are clear because I can't tell if high priority means it should run earlier or later than the others. I have another suggestion which is `:earliest, :early, :normal, :late, :latest`. Any thoughts on these? Am I alone in my confusion?

parkr:
The problem with all of this is really that you can't know the way the hooks will be run, e.g. what if two plugins have `:earliest` or `:highest`? Which one is run first?

Maybe we should order them more systematically, like with ints or something.

envygeeks:
We already had an elaborate plan for not only out-of-order plugin priority but generic plugin priority which was what @stevecrozz mentioned yesterday, if what we want is the ability to insert yourself before another hook then the plan from yesterday is better than the plan from today.

imathis:
@parkr: I'm fine with integer ranking or symbols. The Jekyll priorities are [currently number based](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/plugin.rb#L3), but use symbols as human friendly labels.

@envygeeks: Thinking about it more, I don't see your solution from yesterday working out very well. In the case of my multilingual plugin, I really do want it to run before most plugins, setting it to `:high` priority ensures that I can do that. This means that other plugins don't have to be aware of multilingual features to work properly. It would be unfortunate if all other plugins had to *opt-in* by registering their hooks to run after octopress-multilingual.And of course it would be impossible for me to know which plugin should be run after and keep updating my plugin, adding them to the list of hooks that should be registered after.

I prefer an agnostic global order approach.

envygeeks:
@imathis your multilingual plugin shouldn't be a hook, it should be a generator tbh.

envygeeks:
Or, if it's modifying things like the nav, why does it need to run first or before everything? I would assume that as translator it would want to run absolutely last so that it has access to the final modified source no?

imathis:
@envygeeks Perhaps it could be done with a generator, I don't know. Will a generator let me merge the payload of the site and all the pages? As far as I understand, I need hooks for that. But this is an aside.

What do you think about a global priority? I know it's more of a club than a scalpel, but it also seems they could coexist. People could use the simple global priority system, or if they know of another plugin's hooks they want to proceed or follow after, they could use that instead. This way hooks could be shipped with a simple priority system, but as validating use-cases arise, your more advanced priority system could be added.

envygeeks:
I wouldn't think of it as a global priority but in that scenario I would vote for it to just be :top or :bottom, you either insert yourself at the top of the list or you insert yourself at the bottom of the list if you don't want to hook before or after a specific plugin (with the default being :bottom) that would have already been the case to begin with (always inserting at the bottom) but inserting at the top might require an extra conditional on the sort but that's no big deal.

imathis:
I'm fine with a three level priority system. `:top`, `:normal`, `:bottom`. I think it's useful to be able to indicate a higher priority as well as a lower priority than normal.

pathawks:
It’d be neat if the labels could have *slightly* more descriptive names.

ie: I know my site feed needs to wait until posts have been generated.

stevecrozz:
I'm willing to implement whatever we end up with as a consensus, but I'm starting to think the right solution is the simplest one: Load plugin hooks in the order you want them to run.

This could pose a problem only when order matters and you don't control the order. But the other solutions have their own problems too, and comparatively this problem seems easy to understand and manage.

parkr:
> Load plugin hooks in the order you want them to run.

@stevecrozz How would you achieve this? Some are required in the source code here, some in other plugins, etc. All you have control over is your Gemfile and your config `gems` directive. I'm worried that won't work nicely.

parkr:
@stevecrozz Let's have Hooks inherit from `Jekyll::Plugin` so they can be given a priority. Once all the plugins have been required, sort them accordingly.

stevecrozz:
Personally, I do it like this: https://github.com/stevecrozz/lithostech.com/blob/master/_plugins/ext.rb

But my approach certainly has its limitations and I'm fine with doing it differently. @parkr, can we chat in real-time about this on IRC? I'm stevecrozz on freenode...

parkr:
@stevecrozz I can't do today, but perhaps Sunday or Monday night? Real-time is tough. Feel free to express here – async is easier :)

stevecrozz:
I just don't have a clear picture of what you're asking for. I have a lot of questions, but maybe the easiest way for you to explain would be for you to write the documentation for how you want this to work. Would you be willing to do that?

parkr:
Hooks should be able to be given a priority. It would make sense to register a hook which is its own subclass of `Jekyll::Plugin`.

```ruby
class MyFunPlugin << Jekyll::Hook
  priority :highest # run me first

  def pre_render(site)
    # do mah thang
  end
end
```

Maybe?

parkr:
@stevecrozz What do you think?

stevecrozz:
@parkr It looks good. Did you want hooks to be subclasses of Jekyll::Hook or Jekyll::Plugin? And in your example `MyFunPlugin`, how would you specify that this is a Page hook and not a Site hook?

stevecrozz:
Also, if we have one class per hook, it seems cleaner to make the method name something generic like 'run'. Since all plugins are not hooks, I'm assuming you want to stick with subclassing Jekyll::Hook which can be both hook declaration and registration in one step. An instance of MyFunPlugin is never needed, so the 'run' method can be a stateless, class method.
```ruby
class MyFunPlugin < Jekyll::Hook
  priority :highest # run me first
  container Jekyll::Page
  hook :pre_render

  def self.run(page, payload)
    # do mah thang
  end
end
```

But if the only thing we really want to add to this PR is priority, we can do it pretty easily without making such a big change (easier for me to implement):
```ruby
Jekyll::Hook.register Jekyll::Page, :post_render, priority: :highest do |page|
end
```

If the hook author really wants classes, that still works:
```ruby
Jekyll::Hook.register Jekyll::Page, :post_render, priority: :highest, &MyFunPlugin.method(:run)
```

parkr:
@imathis Any preference? :point_up:

imathis:
I'm fine with passing the priority as an option when rendering. I like that people can write simple hooks with a single line and very few options.

Looking at this again, I have an unrelated suggestion. Rather than pass a class like `Jekyll::Page` why not accept symbols like this?

```ruby
Jekyll::Hook.register :page, :post_render do |page|
end
```

The register method can determine the appropriate classes and when Jekyll eventually makes everything a collection we can just update the hooks module to use a method which filters collection by type of page.

Also, I'd like to suggest that this could be an array of page types, meaning that hooks would be run on each, for example:

```ruby
Jekyll::Hook.register [:page, :post, :collection], :post_render do |item|
end
```

Thoughts?

StevenBlack:
I have long experience with hooks.  Some thoughts on the [proposed implementation](https://github.com/stevecrozz/jekyll/blob/jekyll-hooks/lib/jekyll/hooks.rb).

* Consider making Hooks a self-referential class.  When Hooks *are* Hooks, then we can build pipe-and-filter, chain-of-responsibility, composite, and strategy pattern implementations with surprising ease.

In the [proposed implementation](https://github.com/stevecrozz/jekyll/blob/jekyll-hooks/lib/jekyll/hooks.rb) we register Hooks with Jekyll.

In an expanded view of that, we could **also** register Hooks with Hooks.  In so dong, chained logic structures are possible.

All that's needed for this expanded view is to add a `@hook` array to the Hook class, and we iterate the `@hook` array after the business of `self` is done.

**Sidebar:** in the past I've had much cleaner results when Hooks themselves decide whether they apply to the current context.  An easy way to implement this is wrap `Hook::call()` or `Hook::execute()` in the equivalent of `if hookAppliesToThisSituation{ ... }`.

StevenBlack:
Another note.

I like the currently proposed implementation.

I'd like to propose a thought mode.  There are two ways of looking at hooks.

* **Parser hooks** are Hooks that modify userland content at various junctures.
* **Process hooks** — which we haven't discussed much yet — are ways the Jekyll build process could be extended and enhanced at various junctures.

Both parser hooks and process hooks should have the same interface and both should be self-referenced classes.  

In other words Hooks, regardless of whether they are parser or process hooks, can delegate to other Hooks at predictable junctures common to all Hook processes.

imathis:
@StevenBlack I like the idea of making a Hook a self referential class. Especially since that allows a user to customize internal Hook methods for each implementation, rather than having to monkey patch Jekyll::Hooks.

StevenBlack:
@imathis exactly!

Something like this:  Ignore the third row classes because that's a higher-order discussion but basically we put a `HookAnchor` in Jekyll and we chain userland hooks on this anchor.

stevecrozz:
I think a good next step would be to document the proposed interface from the outside so we can all give it a thumbs up, and then I will work to provide a minimal implementation. If anyone here would like to document their preferred interface, it could help provide a good reference for these discussions. Just publish a gist with new docs and link it here if you decide to help out in this way. I'll do my best to translate this thread into some new documentation also.

In general, I'm on board with these plans, but I would recommend a lot of restraint in the first implementation. From my perspective, we just need a solid foundation to build upon, something that can grow with us as we add features, but also something that doesn't aim to address use cases that we don't have. And on that note, if you do have some real use cases that you want covered, please do consider contributing some cucumber specs. Even without making them pass, they'll be very helpful for me to review.

parkr:
We might be getting ahead of ourselves. I'm generally :+1: on what @stevecrozz said:

> I would recommend a lot of restraint in the first implementation. From my perspective, we just need a solid foundation to build upon, something that can grow with us as we add features, but also something that doesn't aim to address use cases that we don't have.

Let's move forward with more of an MVP that fits our general use cases, then build upon that.

envygeeks:
This shit escalated fast.

imathis:
@parkr @stevecrozz :+1: I'm happy with proceeding without hooks being self-referential classes until we decide we need them. I really would like [this](https://github.com/jekyll/jekyll/pull/3553#issuecomment-93606719) though.

stevecrozz:
@imathis and I just discussed this and we decided to make a few changes to this PR. This is just here so anyone watching can see what we're planning to do:

1. register hooks with owners that are symbols (not direct references to classes)
2. during registration, you can specify owner as an array to register the same hook to multiple owners
3. add optional priority during registration as a symbol (:lowest, :low, :normal, :high, :highest)
4. implement hooks for collections as they are in octopress-hooks

envygeeks:
Why does there need to be 5 instead of 3 levels of hooking?  The wording is what gets to me really not the levels because highest compared to high makes no sense in my head as far as programming is concerned.  Why not make them numbers and support legacy names until 3.1?

stevecrozz:
@envygeeks I'm good with that. Are you, @imathis?

imathis:
Sure, three should be fine. Five was just mimicking the Jekyll pattern for Plugins.

StevenBlack:
> ... levels of hooking.

Careful here.  

There's a class of parsing hooks that 1) Seed content in the `pre_` hook and 2) then unseed the content in the `post_` hook.

Bearing in mind that it's hooks, plural, then an early content-seeding `pre_` hook typically needs to be a late `post_` hook for unseeding that content.  I can't think of a single instance of an early `pre_` hook that also wants to be an early `post_` hook.

If hooks *are* hooks, as I'm advocating, then the sequence of hooks is determined in userland by the way they are composed or the sequence they are either chained-together, or loaded into a flat iterable collection.

Consider that the Jekyll pattern for plugins doesn't necessarily map to hooks.

envygeeks:
Leaving aside you want to unseed in a one time run.... unless you have absolute control over every instance of the hook system you can't guarantee anything so your logic is broken to begin with.  You cannot ensure that every hook author will not just use :highest, so there is no point in even including it because as you just pointed out, every Jim Bob programmer is going to just use :highest to "unseed" content making 5 levels moot to begin with.

StevenBlack:
@envygeeks I recommend you [start here](http://www.amazon.com/Design-Patterns-Object-Oriented-Software-Development/dp/0201422948/ref=sr_1_1?ie=UTF8&qid=1430326570&sr=8-1&keywords=Wolfgang+pree) and then maybe we can talk about "broken logic" with some sort of sensible basis for evaluation for the design of hookable software.

parkr:
@StevenBlack Please don't be snarky. @envygeeks is a very good developer. 

I agree that the wording could be better. `:first`, `:unspecified`, and `:last` might make even more sense if we go with 3 levels. What do you think?

imathis:
@parkr, I'm fine with those labels. For this other stuff, I'm staying out of it. :palm_tree:

envygeeks:
@StevenBlack I've reviewed your code, I don't think you should be handing out advice any time soon.

@parkr @imathis my only concern is the way people tend to always behave, :highest will probably be the highest common usage because everybody wants to be important so if we end up with everybody wanting to be the most important why not just reduce the levels and you can either choose to be not-important, don't-care or just plain old important.  I like the way you labled those new labels.

StevenBlack:
@envygeeks for the record, I wrote my first-ever .rb file 11-days ago, first commit 9-days ago.

I can tell you that extensibility in general, and hook operations in particular, are easy to get-wrong.

imathis:
If you two want to fight, meet at the flag pole after school but keep this kind of thing out of issues.

stevecrozz:
This should address all the feedback @imathis. Just in time for JekyllConf!

lib/jekyll/document.rb
```diff
@@ -175,11 +175,15 @@ def destination(base_directory)
     #
     # Returns nothing.
     def write(dest)
+      Jekyll::Hooks.trigger self, :post_render
```
parkr: Why not put this at the end of render?

lib/jekyll/hooks.rb
```diff
@@ -0,0 +1,103 @@
...
+      },
+    }
+
+    # map of all hooks and their priorities
+    @hook_priority = {}
+
+    NotAvailable = Class.new(RuntimeError)
+
+    # register hook(s) to be called later
+    def self.register(owners, event, priority: nil, &block)
```
parkr: We don't use named parameters anywhere else in the code base. Advantages of this are that you can specify the block without specifying the priority?

lib/jekyll/hooks.rb
```diff
@@ -0,0 +1,103 @@
...
+      Array(owners).each do |owner|
+        register_one(owner, event, priority: priority_value(priority), &block)
```
parkr: What if `block` is nil?

lib/jekyll/hooks.rb
```diff
@@ -0,0 +1,103 @@
+module Jekyll
+  module Hooks
+    # Helps look up hooks from the registry by owner's class
+    OWNER_MAP = {
+      Jekyll::Site => :site,
+      Jekyll::Page => :page,
+      Jekyll::Post => :post,
+      Jekyll::Document => :document,
+    }.freeze
```
parkr: What if I want to make a hook for my custom plugin? Like post_sass or something?

lib/jekyll/site.rb
```diff
@@ -158,7 +164,6 @@ def render
         end
       end
 
-      payload = site_payload
```
parkr: This line is non-negotiable. Totally changes the speed of regen. For one site, it went from 300 seconds to 80 seconds as a result of this line. I'd like to keep it at all costs or cache the site payload on first call.

lib/jekyll/hooks.rb
envygeeks: I personally don't like this usage of keyword args.  Why not have it default the value to the internal API's instead of having nil seep through and then deciding on the outside, have the outside default and the internal API force a value or collapse.  That's just my opinion: the public API's set the usage and defaults.

envygeeks: Let me add, I'm not saying remove the `**kwd` I'm saying, lets build an API where the defaults surface.

site/_docs/plugins.md
```diff
@@ -470,6 +470,220 @@ module Jekyll
...
+      <th>Called</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>
+        <p><code>:site</code></p>
+      </td>
+      <td>
+        <p><code>:reset</code></p>
```
parkr: After_reset?

parkr:
Thanks for all your work on this, @stevecrozz!! :sparkles:

Any other thoughts, @jekyll/core and @imathis?

envygeeks:
My only complaint is that defaults aren't on the surface API, we are seeping nil through the stack into internal API's and then setting defaults.  I think that's obscure design and we should surface every default we have on the layer that is public making everything obvious.

parkr:
I would tend to agree, Jordon! Which defaults, specifically?

lib/jekyll/document.rb
stevecrozz: That's fine with me. It's here just to match the behavior of octopress hooks:
https://github.com/octopress/hooks/blob/master/lib/octopress-hooks.rb#L256

I'm fine with moving it though. Do you still want me to do that?

lib/jekyll/hooks.rb
stevecrozz: Adding an exception when block does not resond to 'call'

lib/jekyll/hooks.rb
stevecrozz: Good idea. I think it's safe to proceed with this frozen constant for now because its just an internal API. If we need a public API for adding hook points we can change this storage mechanism easily enough.

lib/jekyll/site.rb
stevecrozz: Not sure I understand this. The same line: ```payload = site_payload``` is just 10 lines above. If it does really need to be here, do you have another suggestion on how we can allow the :site, :pre_render hook to modify the payload?

stevecrozz: @parkr: I probably need to clarify the problem a bit more. I removed this line because it overwrites the value of the local variable 'payload' which can be altered by site pre_render hooks. There's a bunch of possible solutions to this problem but in order to suggest one, I think I need to understand this performance issue.

lib/jekyll/document.rb
parkr: Yeah, I think it'd make more sense. :)

lib/jekyll/site.rb
parkr: Hm. I was doing this to update the contents of the payload. If it still works as expected (they're all pointers I guess) then it should be fine. I'm sorry to be cross – I thought you removed the caching! Thanks, Steve.

envygeeks:
@parkr it was originally on line 53 of hooks.rb but it looks @stevecrozz caught what I meant and correct it so :tada: I'm liking this so far!

imathis:
Lookin' good! I'll give another go at rebuilding some plugins based on this.

lib/jekyll/site.rb
stevecrozz: No prob. I just don't want to accidentally introduce a big performance issue

parkr:
@gjtorikian: would you mind trying out this branch on a big site you have access to? Interested in performance change, if any. :heart:

stevecrozz:
@imathis You could potentially author a shim to make your current octopress-hooks based plugins work with these new hooks. Here's a totally untested start at something like that.
https://gist.github.com/stevecrozz/ca294104609efd8205bb

lib/jekyll/site.rb
lgp36018: hi

Stephen Crosby <notifications@github.com> wrote:
>> @@ -158,7 +164,6 @@ def render
>>          end
>>        end
>>  
>> -      payload = site_payload
>
>No prob. I just don't want to accidentally introduce a big performance
>issue
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/jekyll/jekyll/pull/3553/files#r29559695

-- 
Sent from my Android device with Gmail Plus. Please excuse my brevity.

imathis:
@stevecrozz I'll give it a try, thanks for the guidance.

gjtorikian:
> would you mind trying out this branch on a big site you have access to? Interested in performance change, if any.

I kept trying it out and thought it would be a quick discovery, but kept running into problem. First, most plugins aren't Jekyll 3 compatible yet; I've logged [a](https://github.com/jekyll/jekyll-sitemap/issues/74) [few](https://github.com/jekyll/jemoji/issues/17) issues for them. But even with them commented out, I discovered so of my own plugins/hacks are incompatible, too. It'll take me some time to upgrade those, in order to be able to test performance on Jekyll 3/this branch.

parkr:
@stevecrozz Is this ready to go?

parkr:
Merging. Feel free to add further feedback as we approach the final release!

imathis:
I just pushed an [octopress-multilingual pull-request](https://github.com/octopress/multilingual/pull/14) for merging Jekyll hooks support. It worked really well and was nice to integrate. Thanks for your work on this @stevecrozz and everyone else who helped! :+1:

parkr:
Hooray!!! :tada:

stevecrozz:
Very cool. Thanks for all the help, everyone. I know there were a bunch of other ideas up there that didn't make it into this first round but it doesn't have to end here. I'll look forward to seeing and reviewing more hook-related work.

PR #3566, opened 2015-03-11
Smartify

penibelst:
*WIP, do not merge*

Based on @parkr’s comment https://github.com/benbalter/benbalter.github.com/pull/222#issuecomment-75596864:

> Yeah, we can add a smartify or smartify_quotes filter.

I’ve decided to implement the *smartify* filter. It would use Kramdown’s engine to convert plain ASCII characters into their corresponding typographic symbols. Usage in Jekyll:

```
{{ 'The Files "in" the Computer --- Zoolander' | smartify }}
=> The Files “in” the Computer — Zoolander
```

Should I continue?

### DIFF
--- a/lib/jekyll/converters/smartypants.rb
+++ b/lib/jekyll/converters/smartypants.rb
@@ -1,0 +1,21 @@
+module Jekyll
+  module Converters
+    class Smartypants < Converter
+      safe true
+
+      priority :lowest
+
+      def matches(ext)
+        true
+      end
+
+      def output_ext(ext)
+        ext
+      end
+
+      def convert(content)
+        content
+      end
+    end
+  end
+end
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -166,6 +166,7 @@ end
 require_all 'jekyll/commands'
 require_all 'jekyll/converters'
 require_all 'jekyll/converters/markdown'
+require_all 'jekyll/converters/smartypants'
 require_all 'jekyll/generators'
 require_all 'jekyll/tags'
 
--- a/lib/jekyll/converters/smartypants/kramdown_parser.rb
+++ b/lib/jekyll/converters/smartypants/kramdown_parser.rb
@@ -1,0 +1,11 @@
+module Jekyll
+  module Converters
+    class Smartypants
+      class KramdownParser
+        def convert(content)
+          content
+        end
+      end
+    end
+  end
+end
--- a/lib/jekyll/converters/smartypants/redcarpet_parser.rb
+++ b/lib/jekyll/converters/smartypants/redcarpet_parser.rb
@@ -1,0 +1,13 @@
+require 'redcarpet'
+
+module Jekyll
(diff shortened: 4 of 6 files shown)


parkr:
I think it's worthwhile. @jekyll/core?

alfredxing:
:+1:

pathawks:
Cool!

I've been using a custom filter to call `Redcarpet::Render::SmartyPants.render` for quite some time, but building it into Jekyll using the default Markdown engine would be neat.

penibelst:
@pathawks Thank you for the hint.

I see two ways to render the input:

1. Use the smarty pants method from users current markdown engine. (difficult)
1. Markdownify and than strip HTML. (easy)

Should I implement the second way now?

parkr:
@benbalter – thoughts?

ideally, it'd be some idempotent method that doesn't do any html stuff. we've run into problems doing that in the past and it always blows up in our face.

pathawks:
> Markdownify and then strip HTML. (easy)

Woe to the user who can't figure out why `smartify` is ruining hash tags.

`#twitter` becomes `twitter`

penibelst:
I see your points. The first implementation just uses Redcarpet’s renderer. From what I see in the markdownify filter, I need to create a converter class to decide what engine is to use. I can not figure out, how to do it. Can anyone collaborate here please?

lib/jekyll/converters/smartypants.rb
```diff
@@ -0,0 +1,21 @@
+module Jekyll
+  module Converters
+    class Smartypants < Converter
+      safe true
+
+      priority :lowest
+
+      def matches(ext)
+        true
```
parkr: !!!!!!!! :( this applies this to _every_ convertible file. definitely don't want this. what if it's JSON or something?

lib/jekyll/converters/smartypants.rb
```diff
@@ -0,0 +1,21 @@
...
+      end
+
+      def output_ext(ext)
+        ext
+      end
+
+      def convert(content)
+        content
```
parkr: I'd like to just do any smarty stuff here. We don't need a bunch of different converters for it, right?

parkr:
I'm starting to think this should just be a plugin... it's getting big for inclusion in the core. If it were one or two functions, that'd be fine. But it's a beefy addition.

penibelst:
> I'm starting to think this should just be a plugin... it's getting big for inclusion in the core. If it were one or two functions, that'd be fine. But it's a beefy addition.

You are right. It increases the complexity without huge gain. Until now people, including myself, can live without the filter. I close my attempt.

pathawks:
Maybe I should plugin-ify my custom filter, and we can maybe revisit this in the future if a clear path forward presents itself.

pathawks:
[**jekyll-smartify**](https://github.com/pathawks/jekyll-smartify)

penibelst:
@pathawks Nice reminder.

Issue #3573, opened 2015-03-15
Remove script/test or have it just rake test.


envygeeks:
Is there any real reason to use it? `bundle exec rake test` works just as well.  It would remove the hacky loader.rb from our files which tbh probably serves no real purpose at the end of the day, it seems like a legacy piece and it's better to remove it now before we head into RSpec than to deal with it later.

Link: #3543

parkr:
I'd like to keep all the script/ stuff. I have them in all my projects -- it's a nice abstraction so new users don't get lost trying to run a project because they don't know the language's standards.

mattr-:
:+1: to keeping the `script` folder.

parkr:
It could just shell out to bundle exec rake test, but i'd like to keep it :)

envygeeks:
While I don't agree that it's a nice abstraction (if there honestly isn't anything that happens that's special) I never removed `script/` inside of my pull request (and never will intend to remove `script/`, because it's not my purpose to disrupt) I removed `loader.rb` and did in-fact have it rely on bundle exec and the way minitest works.

PR #3574, opened 2015-03-15
Remove loader.rb and "modernize" script/test.

envygeeks:
Fixes: #3573


envygeeks:
Yeah I didn't mean "utitlize modernize".  I meant just modernize.

parkr:
I remember adding the loader in because rake wasn't available in Ruby 2.2 (just the lib files weren't there).

script/test
```diff
@@ -1,23 +1,19 @@
 #! /bin/bash
-#
+set -xe
```
parkr: Could we live without `x`?

parkr:
:+1:

parkr:
> I remember adding the loader in because rake wasn't available in Ruby 2.2 (just the lib files weren't there).

Seems to have been fixed!

Issue #3577, opened 2015-03-16
Documentation for adding variable parameters to an include


daviewales:
Currently, the [documentation for includes](http://jekyllrb.com/docs/templates/#includes) explains how to pass string parameters to includes, but gives no explanation of how to pass variable parameters to includes.

Specifically, the documentation says:

    {% include footer.html param="value" %}

However, it doesn't explain that variables can be passed this way:

    {% include footer.html param="value" variable-param=page.variable %}

It took lots of searching to figure this out.

parkr:
Agreed, that should be in there! Feel free to throw that in there.

tomjohnson1492:
I'm trying to understand the documentation that you added around variable parameters, but I'm coming up blank. I added a question on Jekyll Talk trying to understand this section of the documentation. Can you clarify it a bit? Here's more detail on my Jekyll Talk post. [https://talk.jekyllrb.com/t/do-parameters-for-includes-allow-liquid-syntax/2536](https://talk.jekyllrb.com/t/do-parameters-for-includes-allow-liquid-syntax/2536)

Ultimately I want to store a product variable in my configuration file, and then use it in a parameter in my include.

PR #3581, opened 2015-03-17
Document addition of variable parameters to an include.

daviewales:
This fixes issue #3577.

### DIFF
--- a/site/_docs/templates.md
+++ b/site/_docs/templates.md
@@ -286,7 +286,7 @@ root of your source directory. This will embed the contents of
   </p>
 </div>
 
-You can also pass parameters to an include. Note that the variable parameter is NOT surrounded with the usual liquid curly brackets:
+You can also pass parameters to an include. Note that the variable parameter is **not** surrounded with the usual Liquid curly brackets:
 
 {% highlight ruby %}
 {% raw %}{% include footer.html param="value" variable-param=page.variable %}{% endraw %}
--- a/site/_docs/templates.md
+++ b/site/_docs/templates.md
@@ -286,10 +286,10 @@ root of your source directory. This will embed the contents of
   </p>
 </div>
 
-You can also pass parameters to an include:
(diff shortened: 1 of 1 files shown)


alfredxing:
:+1: Though we might want to use bold or italic instead of all caps?

site/_docs/templates.md
```diff
@@ -286,10 +286,10 @@ root of your source directory. This will embed the contents of
   </p>
 </div>
 
-You can also pass parameters to an include:
+You can also pass parameters to an include. Note that the variable parameter is NOT surrounded with the usual liquid curly brackets:
 
 {% highlight ruby %}
-{% raw %}{% include footer.html param="value" %}{% endraw %}
+{% raw %}{% include footer.html param="value" variable-param=page.variable %}{% endraw %}
```
troyswanson: It is totally possible that I'm wrong, but I *think* dashes are not allowed in Liquid variables...

daviewales: Dashes seem to be working on my end.

daviewales:
@alfredxing I've changed it from ALL CAPS to **strong**.

site/_docs/templates.md
```diff
@@ -286,10 +286,10 @@ root of your source directory. This will embed the contents of
   </p>
 </div>
 
-You can also pass parameters to an include:
+You can also pass parameters to an include. Note that the variable parameter is **not** surrounded with the usual Liquid curly brackets:
```
parkr: I don't understand "variable parameter" in this case. What do you think about:

> Omit the quotation marks if you would like to send a variable's value, and do not include the curly brackets.

daviewales: That seems OK. I would probably say:

> Omit the quotation marks if you would like to send a variable's value. Liquid curly brackets should not be used here.

daviewales: I've changed it to say

> Omit the quotation marks to send a variable's value. Liquid curly brackets should not be used here:

PR #3583, opened 2015-03-17
Option to force document rendering

sonnym:
Some time ago, a change was made to the way collections are processed that treats files without front matter as being static files (#2737).  While this assumption makes sense in most cases (I haven't seen anyone complain about it yet!), these behave so much differently from the normal `Jekyll::Document` that they are much more difficult to work with in cases where the consumer of the collection does still want them to be processed.  In my particular case, I am using two of my plugins in conjunction (sonnym/jekyll_example_embed && sonnym/jekyll_elm) to render and embed examples.  Having the front matter in files written in a different language is unsightly, so I added an option that preserves the old behavior.

I also documented how static files work, as the original pull request that added them as a part of collections did not include any changes to the documentation.

Please let me know if you would like to see any additional work to this!  Thanks!

### DIFF
--- a/lib/jekyll/collection.rb
+++ b/lib/jekyll/collection.rb
@@ -36,7 +36,9 @@ module Jekyll
     def read
       filtered_entries.each do |file_path|
         full_path = collection_dir(file_path)
+
         next if File.directory?(full_path)
+
         if Utils.has_yaml_header? full_path
           doc = Jekyll::Document.new(full_path, { site: site, collection: self })
           doc.read
@@ -46,6 +48,7 @@ module Jekyll
           files << StaticFile.new(site, site.source, relative_dir, File.basename(full_path), self)
         end
       end
+
       docs.sort!
     end
 
--- a/lib/jekyll/collection.rb
+++ b/lib/jekyll/collection.rb
@@ -39,7 +39,7 @@ module Jekyll
 
         next if File.directory?(full_path)
 
-        if Utils.has_yaml_header? full_path
+        if Utils.has_yaml_header?(full_path) || metadata['render']
           doc = Jekyll::Document.new(full_path, { site: site, collection: self })
           doc.read
           docs << doc if site.publisher.publish?(doc)
--- a/site/_docs/collections.md
+++ b/site/_docs/collections.md
@@ -47,8 +47,7 @@ Matter is provided, Jekyll will not generate the file in your collection.
 Note: the folder must be named identically to the collection you defined in
 your `_config.yml` file, with the addition of the preceding `_` character.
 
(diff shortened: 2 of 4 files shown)


parkr:
I like to documentation but I think we shouldn't have this overriding behaviour. Not having the YAML front matter means it's not a convertible page, and that holds for all things Jekyll (afaict) except layouts. Content is only renderable with the `---\n---\n` and I think that makes it simpler to reason about. Convenience of importing other content or unsightliness is not really something I think we should use to make design decisions about the functionality and behaviour of Jekyll.

I like the docs tho!

sonnym:
My particular usecase is that I have a custom converter that turns documents from the Elm language into HTML files, and I have a liquid tag that embeds examples in the form of source code followed by an iframe pointing to the generated file.

Using static files, I would need to recreate the conversion step in the the example embed plugin.  Using documents, I would need to have empty, boilerplate in my elm files.

Is there another way to get that desired behavior?

parkr:
> Is there another way to get that desired behavior?

Insert the YAML front matter in a build step that runs before Jekyll?

sonnym:
That doesn't sound particularly reusable for a public plugin! I guess, in terms of recreating the conversion step, I would just have to map the files into documents, which I don't think will be too difficult.  That is what I am going to try, at least!

Feel free to take and discard what you want from this pull request :smile:

sonnym:
In case you are curious, I was able to hack together a workaround.  It is so dangerous looking, dubios, and probably downright fragile that I almost feel bad about it, but it works!

https://github.com/sonnym/jekyll_example_embed/blob/fa19b06c1f8029d627736682403fa52d9e95aaf8/lib/jekyll_example_embed.rb#L39-L65

I'd be interested to hear any critiques about this method.  It looks like collections have been stable enough for long enough that this shouldn't be *too* problematic.

sonnym:
Sweet!  Thanks for adding in the documentation and the (as always) prompt responses!

parkr:
You bet, thanks for the PR :+1:

sonnym:
No worries, dude.  I actually just saw that you went to Cornell, and I feel like a complete ass, hah!  I am class of 2008 and still live in Ithaca!  I would have gladly bought you a beer for all you've done, had I known before today.  If you come back to visit in the next few months, let me know!

parkr:
> No worries, dude. I actually just saw that you went to Cornell, and I feel like a complete ass, hah! I am class of 2008 and still live in Ithaca! I would have gladly bought you a beer for all you've done, had I known before today. If you come back to visit in the next few months, let me know!

Hah, no problem man. Yeah, I just left last summer! I'm sure I'll be back. Enjoy the winter for me! :wink: :snowflake:

PR #3590, opened 2015-03-18
Each test that requires the `tmp` directory should ensure its existence

nickburlett:
So that `bundle exec rake test` from a fresh clone:
* Create 'tmp' dir for test_tags if it doesn't exist  
* Don't also create that 'tmp' dir in script/test

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -91,12 +91,8 @@ multitask :default => [:test, :features]
 
 task :spec => :test
 
-task :setup_for_tests do
-  FileUtils.mkdir_p("tmp") unless File.directory?("tmp")
-end
-
 require 'rake/testtask'
-Rake::TestTask.new(:test => :setup_for_tests ) do |test|
+Rake::TestTask.new(:test) do |test|
   test.libs << 'lib' << 'test'
   test.pattern = 'test/**/test_*.rb'
   test.verbose = true
--- a/Rakefile
+++ b/Rakefile
@@ -90,8 +90,13 @@ end
 multitask :default => [:test, :features]
 
 task :spec => :test
+
+task :setup_for_tests do
+  FileUtils.mkdir_p("tmp") unless File.directory?("tmp")
+end
+
 require 'rake/testtask'
-Rake::TestTask.new(:test) do |test|
+Rake::TestTask.new(:test => :setup_for_tests ) do |test|
   test.libs << 'lib' << 'test'
   test.pattern = 'test/**/test_*.rb'
   test.verbose = true
--- a/script/test
+++ b/script/test
@@ -5,10 +5,6 @@ set -e
 #   script/test
 #   script/test <test_file>
 
-if [ ! -d tmp ]
-  then mkdir tmp
-fi
-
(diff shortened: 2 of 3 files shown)


parkr:
So to do it in the rakefile, you'd want to do

```ruby
task :setup_for_tests do
  Dir.mkdir_p("tmp") unless File.directory?("tmp")
end

require 'rake/testtask'
Rake::TestTask.new(:test => :setup_for_tests) do |test|
  test.libs << 'lib' << 'test'
  test.pattern = 'test/**/test_*.rb'
  test.verbose = true
end
```

parkr:
I don't think the tests should do this directly, so let's make it a pre-req. A plus is that it'll work on Windows, too.

nickburlett:
ok, I'll update the Rakefile tonight and try it again.

nickburlett:
Ok, the tmp dir is now made in the Rakefile as a test prerequisite.

envygeeks:
Putting it as a rake task isn't such a good idea.  That serves only a single purpose and a single purpose solely tied to testing.  It should be put into the test setup itself not as a task on top of Rake.  We are promoting one time use encapsulation that serves no purpose, it's like C++ all over again :cry:

nickburlett:
@envygeeks I don't disagree... the tests themselves seem like they should set up their own pre-requisites. However, since the `script/test` shell script was originally responsible for creating the directory, I can see the argument for moving it into the rake file. 

I'm happy to have the code go either way, I just want the tmp directory created :->

envygeeks:
Even though I'm resisting I'll concede to @parkr because my personal CI already makes up for the shortfalls in jekyll testing (because I like you prefer rake test) so it doesn't affect me whichever route you guys go but it could impact how people intuitively follow our code and a rake task is pretty uninituive to the testers mind.

parkr:
So the tests themselves shouldn't setup their prerequisites, that's not their concern. They're concerned with ensuring the functionality is as expected. The tests should be grouped based on context, and the context defines the prerequisites.

The `tmp` dir is generally good to have and could be used for lots of tests. The problem here is that `tmp` doesn't exist, so when it goes to create `tmp/pages-test`, it just says "nope" and dies. Why `-p` isn't default in all `mkdir` calls is a mystery to me.

I'd be cool if the creation and deletion of the directories needed for these tests went in `setup` and `teardown` blocks in a context specific to this prerequisite.

nickburlett:
@parkr should I back out the Rakefile and move the mkdir to a setup block? There's already a `context "with symlink'd include"` block, so adding a setup block would be trivial

parkr:
> @parkr should I back out the Rakefile and move the mkdir to a setup block? There's already a context "with symlink'd include" block, so adding a setup block would be trivial

yes please! sorry to send you on this crazy wild :bird: chase.

nickburlett:
@parkr Pull request updated to use a setup block

parkr:
Sweet! One last thing: would you mind squashing your commits into one?

nickburlett:
Closing this in favor of #3609 that contains squashed commits

Issue #3591, opened 2015-03-18
Incremental regeneration issues


nickburlett:
The incremental regeneration doesn't work properly when using serve or watch for three reasons:
1. it only caches the mtime of the first dependency it discovers, so for posts/pages with N multiple dependencies (i.e., every layout file used to render them), it continues to regenerate the post/page N times, at which point it's seen all of the dependencies. 
2. it also incorrectly remembers which files are unchanged, meaning that once it decides a file is unchanged it won't ever regenerate that file until the jekyll command is restarted. 
3. it doesn't regenerate files (other than static files) if the destination file is removed

nickburlett:
Test for removed files (it fails):

````diff
diff --git a/test/test_site.rb b/test/test_site.rb
index 7888f33..e19341e 100644
--- a/test/test_site.rb
+++ b/test/test_site.rb
@@ -495,6 +495,25 @@ class TestSite < JekyllUnitTest
         assert_equal mtime3, mtime4 # no modifications, so remain the same
       end
 
+      should "regnerate files that have had their destination deleted" do
+        contacts_html = @site.pages.find { |p| p.name == "contacts.html" }
+        @site.process
+
+        source = @site.in_source_dir(contacts_html.path)
+        dest = File.expand_path(contacts_html.destination(@site.dest))
+        mtime1 = File.stat(dest).mtime.to_i # first run must generate dest file
+
+        # simulate file modification by user
+        File.unlink dest
+        refute File.exist?(dest)
+
+        sleep 1
+        @site.process
+        assert File.exist?(dest)
+        mtime3 = File.stat(dest).mtime.to_i
+        refute_equal mtime2, mtime3 # must be regenerated 
+      end
+
     end
 
   end
````

alfredxing:
@nickburlett The regenerator doesn't check for the existence of destination files. If you delete destination files, you should delete `.jekyll-metadata` as well (or just use the `jekyll clean` command)

nickburlett:
@alfredxing I guess was expecting behavior similar to StaticFile. In test_site, there's a test for `should "write static files if not modified but missing in destination"`. However, it looks like this test is actually part of a full_rebuild site test, so I guess that wouldn't apply. 

Are you against the idea of it checking for the existence of destination files? It seems like it would be a helpful feature and make incremental build less error-prone if files go missing or if the config changes the output path.

nickburlett:
[...]

PR #3592, opened 2015-03-18
Clear the regenerator cache every time we process

nickburlett:
To address part of #3591, clear the regenerator's cache every time the
site is processed. This ensures that the regenerator doesn't incorrectly
believe a file hasn't changed based on stale information.

### DIFF
--- a/lib/jekyll/regenerator.rb
+++ b/lib/jekyll/regenerator.rb
@@ -59,6 +59,14 @@ module Jekyll
       @cache = {}
     end
 
+
+    # Clear just the cache
+    #
+    # Returns nothing
+    def clear_cache
+      @cache = {}
+    end
+
     # Checks if a path's (or one of its dependencies)
     # mtime has changed
     #
--- a/lib/jekyll/regenerator.rb
+++ b/lib/jekyll/regenerator.rb
@@ -9,7 +9,7 @@ module Jekyll
       read_metadata
 
       # Initialize cache to an empty hash
-      @cache = {}
+      clear_cache
     end
 
     # Checks if a renderable object needs to be regenerated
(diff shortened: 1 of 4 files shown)


nickburlett:
I'm not entirely sure how to deal with the other part of #3591 (where it doesn't cache or update metadata for all dependencies), but this at least fixes the bigger problem of not regenerating files that do need regeneration.

lib/jekyll/regenerator.rb
```diff
@@ -59,6 +59,14 @@ def clear
       @cache = {}
```
parkr: if we go with this, this should just use that helper

lib/jekyll/site.rb
```diff
@@ -68,6 +68,7 @@ def reset
       self.static_files = []
       self.data = {}
       @collections = nil
+      @regenerator.clear_cache()
```
parkr: in ruby, you don't include the `()` unless it takes arguments :)

nickburlett: Ah, yes... can you tell it's been a while since I last coded in ruby? (^_^)
I'll fix this and regenerator clear tonight.

nickburlett:
I updated the code to use the clear_cache method instead of manually assigning `@cache = {}` and I've added a test that would fail without the cache clearing in `site.reset`

parkr:
Cool. @alfredxing

alfredxing:
:+1: Thanks! I can't believe I didn't notice this myself earlier...

lib/jekyll/site.rb
parkr: should this clear everything, or just the cache?

/cc @alfredxing

nickburlett: I don't think we want to clear the metadata... it should still be a correct representation of what was generated previously. It's just the cache of whether or not a path needs to be regenerated that's now stale.

PR #3598, opened 2015-03-20
Added "liquid-md5" plugin

pathawks:
Added a link to my **[liquid-md5](https://github.com/pathawks/liquid-md5)** filter plugin.


envygeeks:
:heart:

PR #3600, opened 2015-03-20
RR replaced with RSpec Mocks

pathawks:
With #3552, Jekyll is now using **[RSpec Mocks](https://github.com/rspec/rspec-mocks/)** instead of **[RR](https://github.com/btakita/rr/)**.

This updates the documentation to reflect that change.


envygeeks:
:heart:

Issue #2208, opened 2014-04-05
Rename _config.yml to _jekyll.yml


benbalter:
What do we think about changing the default/suggested configuration file for Jekyll sites from `_config.yml` to `_jekyll.yml`? Here's why:

Take a look at the folder structure of [Jekyll](https://github.com/jekyll/jekyll) or [Bootstrap](https://github.com/twbs/bootstrap). It's a common design pattern to have things like a `travis.yml` file, `cucumber.yml`, or a `bower.json` file in the project root.

Couple that with the fact that many large projects maintain project documentation along side the code in a `site` or `docs` subfolder, and deploy with a Rake task or similar scripting.

Rather than encouraging projects to put a generic `_config.yml` file in the site root, which really doesn't explain what it is (and may conflict with another file), or to break the design pattern used by other projects and place the config file in a sub folder, starting with Jekyll 2.0.0, lets encourage users to create `_jekyll.yml` files.

We'd still maintain full backwards compatibility (look to `_config.yml` if `_jekyll.yml` doesn't exist) but in documentation and the starter site, we'd create `_jekyll.yml` files going forward.

Thoughts?

/cc @mdo in case he can shed some light on the Bootstrap workflow.

PR #3605, opened 2015-03-22
_config.yml isn't significant and can clash

krainboltgreene:
The configuration file for a jekyll site is currently called `_config.yml`. However due to the nature of jekyll it's feasible that jekyll will be in the same directory space as another framework, library, or program. Once we concede that jekyll isn't in a vacuum it's very possible that the config file could be confused with another program or user's config file since the name isn't significant.

Common practice with things like Travis and Circle is to create a significant yml file: `.travis.yml` and `circle.yml`. In fact, Jekyll uses these files in it's own repository showing the maintainer's acceptance of this practice. The use of `dot file` names has been around for quite a while and is supported by a lot of programs as program metadata. In addition it makes it a lot easier to identify the configuration file for your jekyll site.

To prevent this problem I've created this pull request where in all places except `History.markdown` I rename `_config.yml` to `.jekyll.yml`. I messed with zero logic or configuration, requiring no real tests to be added as the current test are sufficient.

As this is a rather large change I suspect it'll probably be passed up, however in lieu of getting this merged (I do want it merged) I want to point out two problems I encountered doing this:

  1. It was not *easy*. I should have been able to change at max maybe 3 lines since this contains two default constants at best: CONFIG_FILE_NAME and CONFIG_FILE_EXTENSION. Neither of these exist and it meant careful search/replace of the entire repository with at least once having to manually read code. This suggests the value is a "magic number" and could use with a little refactoring.

  2. After pulling down, bundle installing, and running the tests the project refused to run the features. I have not tested these changes against the features for this reason. It suggests there's a problem with the feature tests that people already setup more than described aren't seeing:

```
$ bundle exec rake features
/Users/krainboltgreene/.rvm/rubies/ruby-2.1.5/bin/ruby -S bundle exec cucumber  --profile travis
cucumber.yml was not found.  Current directory is /Users/krainboltgreene/Code/krainboltgreene/jekyll.  Please refer to cucumber's documentation on defining profiles in cucumber.yml.  You must define a 'default' profile to use the cucumber command without any arguments.
Type 'cucumber --help' for usage.
```


pathawks:
Duplicate of #2208

parkr:
This was suggested in #2208 and ultimately rejected.

A dot file isn't good because our average user will have never encountered a dotfile in their life. It has to be shown by default. Plus, dotfiles cause trouble on Windows machines. So no dotfiles.

It must start with an underscore so it's not consumed as a jekyll page or post or whatever. So that leaves us with `_jekyll.yml`, which is what #2208 suggested.

> However due to the nature of jekyll it's feasible that jekyll will be in the same directory space as another framework, library, or program.

Jekyll was created with the idea that a given directory would house the Jekyll site, and that it would remain separate from everything else. If you wanted to bring stuff in from other directories, that was accomplished with symlink support. Generally, supporting a Jekyll site that's mixed in with a bunch of other things is just asking for anger and frustration.

krainboltgreene:
Damn, I swore I looked really hard. Github's search can be such a pain.

krainboltgreene:
@parkr What about the two other points?

parkr:
> I should have been able to change at max maybe 3 lines since this contains two default constants at best: CONFIG_FILE_NAME and CONFIG_FILE_EXTENSION.

It is a "magic string", you're right. It could be refactored, but that'd be a bunch of unneeded work at this point.

> After pulling down, bundle installing, and running the tests the project refused to run the features.

Did you try `script/cucumber`?

krainboltgreene:
No, because that's not the documented command. But it also fails:

```
▶ script/cucumber

features/collections.feature:6  Scenario: Unrendered collection .....x
  uninitialized constant Minitest::Assertion (NameError)
  ./features/step_definitions/jekyll_steps.rb:186:in `/^I should see "(.*)" in "(.*)"$/'
  features/collections.feature:12:in `And I should see "Collections: <p>Use <code>Jekyll.configuration</code> to build a full configuration for use w/Jekyll.</p>\n\n<p>Whatever: foo.bar</p>\n<p>Signs are nice</p>\n<p><code>Jekyll.sanitized_path</code> is used to make sure your path is in your source.</p>\n<p>Run your generators! default</p>\n<p>Page without title.</p>\n<p>Run your generators! default</p>" in "_site/index.html"'
```

krainboltgreene:
Actually, on further consideration I don't get

> It must start with an underscore so it's not consumed as a jekyll page or post or whatever. So that leaves us with _jekyll.yml, which is what #2208 suggested.

There's a default configuration with a default include `[]`, just like there is a default keep_files `['.git','.svn']`, and a default include `['.htaccess']`. Why not make the default exclude be `[.jekyll.yml]`?

Issue #3602, opened 2015-03-20
3.0.0.pre.beta2 breaks permalinks on existing sites


kaatt:
> mkdir bugtest
> cd bugtest
> echo "gem 'jekyll', '3.0.0.pre.beta2'" > Gemfile
> mkdir foo
> cd foo
> echo -e '---\npermalink: /foo/baz.html\n---' > bar.htm
> cd ..
> bundle exec jekyll b
> find .
###
.
./.jekyll-metadata
./_site
./_site/foo
**./_site/foo/baz.html.htm**
./_site/Gemfile
./_site/Gemfile.lock
./foo
./foo/bar.htm
./Gemfile
./Gemfile.lock

beta1 has no problem and generates `./_site/foo/baz.html`

willnorris:
Thanks @kaatt for the detailed steps to reproduce... that helps a lot.  And just to link this to the appropriate PR, this is actually related to #3490.

so strictly speaking, this is working as intended, though I agree that it seems somewhat unintuitive in this particular case.  The change that #3490 made was to ensure that output files always include the correct file extensions, where "correct" is determined by the return value of [output_ext](https://github.com/jekyll/jekyll/blob/a5b9df728766ce052d13429b36eb51ba237c1ddb/lib/jekyll/convertible.rb#L80).  

For file types that Jekyll has a converter for, the converter itself determines the extension.  For example, [the markdown converter returns ".html"](https://github.com/jekyll/jekyll/blob/a5b9df728766ce052d13429b36eb51ba237c1ddb/lib/jekyll/converters/markdown.rb#L57).  For file types that Jekyll does **not** have a converter for, we use the extension of the original source file.

**So why does Jekyll ignore the file extension in the permalink variable?**

Jekyll doesn't know that it's intended to be a file extension.  Jekyll supports permalinks that include dots in them.  So imagine that we had a markdown page with the front matter:

``` yaml
permalink: /i.like.to.separate.words.with.dots
```

What should the destination file's extension be?  Should it be ".dots"?  It seems very unlikely that that was what the author actually intended.  Since markdown files are rendered as HTML, it wouldn't be served properly by most web servers anyway.  Instead, we append ".html" to the permalink when outputting the file, and then [provide instructions](http://jekyllrb.com/docs/permalinks/#extensionless-permalinks) on how to configure a web server appropriately (or in the case of GitHub pages or WEBrick, no extra config is necessary).

So there are a few possible options for you:

[...]

Issue #3324, opened 2015-01-18
3.0 RELEASE GAMEPLAN


parkr:
I have [written about what I'd like to see for Jekyll 3.0](https://byparker.com/blog/2014/jekyll-3-the-road-ahead/). That post is now a month and a half old. I'd like to get Jekyll 3.0 out the door in Q1 2015 to bring all the amazingness we have been cooking out to you excellent Jekyll users and developers. Check lists usually help me more than anything else, so here's what I'd like to accomplish:

- [x] Incremental regeneration ( #3116 )
- [x] Integrate basic hooks (potentially via @imathis's [`Octopress::Hooks`](https://github.com/octopress/hooks) work) ( ~~#3392~~, #3553. )
- [x] Switch over to Rouge as default syntax highlighter, dropping hard dependency on Python ( #3323 )
- [x] Drop hard dependency on ExecJS runtime (e.g. Node) ( #3323 )
- [x] Make most dependencies optional and handle their absence gracefully ( #3323 )
- [x] Upgrading to Liquid 3 & adding optional liquid-c support ( #3002 )
- [x] Make `_posts` a default collection (& therefore simplify the internals) ( #3169, #3690 )
-  ~~Remove `Convertible`.~~
- ~~Automated Windows tests? ( #3115 )~~

Anything else you'd like to see?

imathis:
Everything here looks so great. I'll be especially excited to hooks come to Jekyll. Anything to get rid of all the monkey-patching we do for plugins!

paulrobertlloyd:
This looks *great*. In terms of supporting plug-ins, it would be nice-to-have a more fully featured `jekyll-paginate` be released around the same time (i.e. the having the ability to paginate practically any group of posts). Pagination I think is the only area from a templating perspective where Jekyll still suffers slightly.

pepelsbey:
+1 for pagination issues.

parkr:
Pagination chat should be had over at: https://github.com/jekyll/jekyll-paginate But thanks for the input.

iainspad:
Rouge seems like a nice idea, but frankly, I'd like to see if my `code` blocks can be written out as lists and not using `table`. Does Rouge support that or are there other methods it provides?

alfredxing:
We should probably get working on `Make _posts a default collection (& therefore simplify the internals) & remove Convertible`. Related issue is #3169.

parkr:
[...]

PR #3607, opened 2015-03-22
Upgrade guide for Jekyll 3

chrisfinazzo:
Per #3324 and comments in #3602, docs changes related to the upgrade process for 3.0 can land here. In their current state, there is some duplication on the [Upgrading][] page and the information is (obviously) in need of an update.

[Upgrading]: http://jekyllrb.com/docs/upgrading/

### DIFF
--- a/site/_docs/upgrading.md
+++ b/site/_docs/upgrading.md
@@ -87,7 +87,7 @@ to one or more config files (comma-delimited, no spaces).
 
 * `--no-server`
 * `--no-auto`
-* `--auto` (now `--watch` and `--no-watch`)
+* `--auto` (now `--watch` and `--no-watch`, defaults to `--watch`)
 * `--server`
 * `--url=`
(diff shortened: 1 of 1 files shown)


chrisfinazzo:
We're repeating ourselves when describing the behavior of Absolute v. Relative Permalinks. Does this still need to be prefaced with a "Warning" label?

chrisfinazzo:
Sure, I'll update that.

See also - #3610 - I think that should cover most of `_docs`.

site/_docs/upgrading.md
```diff
@@ -85,7 +87,7 @@ to one or more config files (comma-delimited, no spaces).
 
 * `--no-server`
 * `--no-auto`
```
parkr: this is actually `--no-watch` not the line below

chrisfinazzo:
👍

PR #3608, opened 2015-03-22
Process metadata for all dependencies

nickburlett:
When adding a dependency, also add the dependency to the metadata hash.

Addresses part 1 of #3591. Prior to this fix, the regnerator only paid attention the mtime of the first dependency it checked, so for posts/pages with N multiple dependencies (i.e., every layout file used to render them), it continues to regenerate the post/page approximately N times, at which point it's seen all of the dependencies.

### DIFF
--- a/lib/jekyll/regenerator.rb
+++ b/lib/jekyll/regenerator.rb
@@ -104,7 +104,10 @@ module Jekyll
     def add_dependency(path, dependency)
       return if (metadata[path].nil? || @disabled)
 
-      metadata[path]["deps"] << dependency unless metadata[path]["deps"].include? dependency
+      if !metadata[path]["deps"].include? dependency
+        metadata[path]["deps"] << dependency
+        add(dependency) unless metadata.include?(dependency)
+      end
       regenerate? dependency
     end
 
--- a/test/test_regenerator.rb
+++ b/test/test_regenerator.rb
@@ -188,6 +188,20 @@ class TestRegenerator < JekyllUnitTest
       assert @regenerator.modified?(@path)
     end
 
+    should "not regenerate again if multiple dependencies" do
+      multi_deps = @regenerator.metadata.select {|k,v| v['deps'].length > 2}
+      multi_dep_path = multi_deps.keys.first
+
+      assert @regenerator.metadata[multi_dep_path]["deps"].length > 2
+
+      assert @regenerator.modified?(multi_dep_path)
+
+      @site.process
+      @regenerator.clear_cache
+
+      refute @regenerator.modified?(multi_dep_path)
+    end
+
     should "regenerate everything if metadata is disabled" do
       @site.config["full_rebuild"] = true
       @regenerator.clear


alfredxing:
Hmm... I feel like there's a reason I didn't `add` the dependencies, but I can't think of what it was. I'll look into this more.

parkr:
Ok! Let me know if you need anything. Thanks alfred!!

nickburlett:
@alfredxing Have you had a chance to look into this? 

I've been using it for a while now, and it's been helpful. I haven't been making heavy use of my jekyll project, but what work I've been able to do has been made much easier with this change. I'm using a variation on the fauno/jekyll-pandoc-multiple-formats plugin that generates PDFs from various pages on the site, and this adds several minutes to the regeneration time. With the changes in d4b8f0d it takes much less time, as it only updates the document I'm changing.

I'd love to see this merged into the official branch!

alfredxing:
@nickburlett I've thought about it but haven't come up with anything. It's great to hear that it's working well for you in use, so we might as well merge this in.

@parkr Anything need to be done here before merge?

nickburlett:
@alfredxing To be fair, I've probably only used it for about ten hours of editing over the past couple of weeks, and I'm not sure if my work load is representative of what others need.

lib/jekyll/regenerator.rb
```diff
@@ -104,7 +104,10 @@ def modified?(path)
     def add_dependency(path, dependency)
       return if (metadata[path].nil? || @disabled)
 
-      metadata[path]["deps"] << dependency unless metadata[path]["deps"].include? dependency
+      if !metadata[path]["deps"].include? dependency
+        metadata[path]["deps"] << dependency
+        add(dependency) unless metadata.include?(dependency)
```
parkr: should `add()` be idempotent, @alfredxing?

lib/jekyll/regenerator.rb
```diff
@@ -104,7 +104,10 @@ def modified?(path)
     def add_dependency(path, dependency)
       return if (metadata[path].nil? || @disabled)
 
-      metadata[path]["deps"] << dependency unless metadata[path]["deps"].include? dependency
+      if !metadata[path]["deps"].include? dependency
```
parkr: maybe it should be a `Set` instead of just an `Array`? Does that slow it down unacceptably? That would eliminate the need for the check here.

lib/jekyll/regenerator.rb
alfredxing: Ideally it should be. I think I ended up doing things like the `unless metadata.include?(dependency)` above, which isn't really good.

parkr: What do you mean, "ideally"? :wink:

alfredxing: Well, it *should* be, but it's not a bug or anything. So, yes, we should change it, but maybe as a separate refactor/documentation issue/PR instead?

lib/jekyll/regenerator.rb
alfredxing: A set sounds good. Wouldn't it actually speed it up, since it's constant time find (instead of linear time)?

lib/jekyll/regenerator.rb
parkr: :+1:

lib/jekyll/regenerator.rb
parkr: I'm honestly not sure.

PR #3609, opened 2015-03-22
Create 'tmp' dir for test_tags if it doesn't exist

nickburlett:
Rather than use script/test to create the tmp directory, create it in a setup block for the appropriate context in the `TestTags` test.

### DIFF
--- a/script/test
+++ b/script/test
@@ -5,10 +5,6 @@ set -e
 #   script/test
 #   script/test <test_file>
 
-if [ ! -d tmp ]
-  then mkdir tmp
-fi
-
 if [ -d test/dest ]
   then rm -r test/dest
 fi
--- a/test/test_tags.rb
+++ b/test/test_tags.rb
@@ -4,6 +4,10 @@ require 'helper'
 
 class TestTags < JekyllUnitTest
 
+  def setup
+    FileUtils.mkdir_p("tmp")
+  end
+
   def create_post(content, override = {}, converter_class = Jekyll::Converters::Markdown)
     site = fixture_site({"highlighter" => "rouge"}.merge(override))
 


nickburlett:
@parkr ... crud, I missed a case that doesn't come up in any of the orderings tried on my system. Two questions:

1. Is there a way to pass the seed value to the test script or the rake command? I can't find one that works.
2. Is there a better way to squash the commit that fixes the test failure without making a new branch?

envygeeks:
1. What seed value are you trying to pass?  "tmp" is hardcoded so there is no reason to pass that?
2. `git rebase -i` after you make a commit and then `+branch` when you push and it will rebase and update the current pull request with the latest data from your feature branch.

If you aren't using a feature branch,  now you know why they ask us to use feature branches :wink:

nickburlett:
@envygeeks 

1. I'm trying to pass a seed value because it seems that the ordering of tests being run depends on it. There are two tests that require the tmp directory to be made, but in this version I accidentally made it so only one of the two creates the tmp directory, making the ordering important. 
2. I had read that `git rebase -i` was not a good idea if you had already pushed to a public repository. But I guess since no one is likely to have pulled from my patch branch, it shouldn't be a problem 

Also,

>There is no need for File.directory? when doing FileUtils.mkdir_p because it silently ignores existing folders.

Good point. 

I'll commit the fix, rebase, and push.

envygeeks:
1. Try moving the setup into the base parent describe.  If setup is not available there, try before without any arguments (as I do believe that is the Minitest::Spec way.)

2. Normally it's not okay to rebase on a public repo (unless all are well aware of the history rewrite - and we are) but in the case of pull requests sometimes we do like to rebase and often if we want to squash we do ask for a rebase (quite often I ask people to squash multiple commits into a single more readable history chain) because you are right, it's no big deal on a pull request because your history is independent of ours until we merge but we would never rebase on Jekyll because that could cause chaos.

envygeeks:
OH, it looks like we aren't fully utilizing minitest spec, we are hybrid, so put the setup at the very top of the class.  I forgot we were still hybrid until I finished going through the tests.

nickburlett:
@envygeeks So something like:

````diff
diff --git a/test/test_tags.rb b/test/test_tags.rb
index 0460f3f..34d8437 100644
--- a/test/test_tags.rb
+++ b/test/test_tags.rb
@@ -4,6 +4,10 @@ require 'helper'
 
 class TestTags < JekyllUnitTest
 
+  def setup
+    FileUtils.mkdir_p("tmp")
+  end
+
   def create_post(content, override = {}, converter_class = Jekyll::Converters::Markdown)
     site = fixture_site({"highlighter" => "rouge"}.merge(override))
 
@@ -745,7 +749,6 @@ CONTENT
     end
 
     context "with symlink'd include" do
-
       should "not allow symlink includes" do
         File.open("tmp/pages-test", 'w') { |file| file.write("SYMLINK TEST") }
         assert_raises IOError do
````

envygeeks:
Yessum :heart:

nickburlett:
@envygeeks Done. 

@parkr I think/hope the goose-chase is over...

envygeeks:
:heart:

parkr:
:tada: !

PR #3610, opened 2015-03-23
Wrap lines

chrisfinazzo:


envygeeks:
Can I get you to squash this into a single commit?   If you don't know how that's fine but it would be preferable to be a single commit!

parkr:
> Can I get you to squash this into a single commit?

Yes please :+1:

chrisfinazzo:
Yeah, about that... `git rebase` doesn't want to cooperate.

screen shot 2015-03-23 at 11 32 40 am

screen shot 2015-03-23 at 11 33 01 am

screen shot 2015-03-23 at 11 32 19 am

Long story short, I _can_ do the work, but BBEdit won't save 😡.

To be fair, this is further than I got in #3567 (where I have no clue what's going on inside of the history and when I try  to squash things, the commits disappear 😖. I suspect that there's an ugly `git reset` required to sort this out, but since these are already public, I'd rather not screw up the log _too_ badly)

pathawks:
> but since these are already public, I'd rather not screw up the log too badly

The whole point is to rewrite the history.

Yes, it's "public" in the sense that you have published this pull request, but it's not "*public*" in the sense that others have been working from your branch.

Rewriting history on the **jekyll/jekyll master** branch would be a huge no-no. Rewriting history on your pull request feature branch is exactly what *squash* means :smile:

parkr:
Try `EDITOR=vim git rebase -i HEAD~3`, maybe? Maybe it's a shortcoming of BBEdit.

chrisfinazzo:
Probably, I've seen this bug come up before in earlier versions, so it's not uncommon. I filed it with Support, so hopefully they will be able to reproduce this.

envygeeks:
Any reason why this was closed?

chrisfinazzo:
Not quite sure. I did run into an error while trying to squash things together, but it seems like this was just because my `.gitconfig` had changed unexpectedly and I got into a situation where I couldn't save the rebase results.

It's fixed now (See commit d492004). If this gets re-opened, will the new commit get picked up? If not, let me know and I will submit a new PR.

envygeeks:
@chrisfinazzo unfortunately I'm unable to reopen it because it was squashed while closed, can you please reopen it as another pull so we can just work from the new squash.

chrisfinazzo:
See #3616.

PR #3616, opened 2015-03-24
Wrap lines at 80 characters

chrisfinazzo:


site/_docs/collections.md
```diff
@@ -52,7 +53,8 @@ your <code>_config.yml</code> file, with the addition of the preceding <code>_</
   </p>
 </div>
 
-### Step 3: Optionally render your collection's documents into independent files
+### Step 3: Optionally render your collection's documents into independent
+files
```
parkr: this breaks the header i think

site/_docs/configuration.md
```diff
@@ -394,7 +424,11 @@ defaults:
       author: "Mr. Hyde"
 {% endhighlight %}
 
-With these defaults, all posts would use the `my-site` layout. Any html files that exist in the `projects/` folder will use the `project` layout, if it exists. Those files will also have the `page.author` [liquid variable](../variables/) set to `Mr. Hyde` as well as have the category for the page set to `project`.
+With these defaults, all posts would use the `my-site` layout. Any html files
+ that exist in the `projects/` folder will use the `project` layout, if it
+  exists. Those files will also have the
+  `page.author` [liquid variable](../variables/) set to `Mr. Hyde` as well as
+   have the category for the page set to `project`.
```
parkr: ?

site/_docs/configuration.md
```diff
@@ -360,7 +384,11 @@ defaults:
       layout: "default"
 {% endhighlight %}
 
-Here, we are scoping the `values` to any file that exists in the scopes path. Since the path is set as an empty string, it will apply to **all files** in your project. You probably don't want to set a layout on every file in your project - like css files, for example - so you can also specify a `type` value under the `scope` key.
+Here, we are scoping the `values` to any file that exists in the scopes path.
+ Since the path is set as an empty string, it will apply to **all files** in
+  your project. You probably don't want to set a layout on every file in your
+   project - like css files, for example - so you can also specify a `type`
+    value under the `scope` key.
```
parkr: ?

parkr:
some weird indentation stuff and some formatting things go wrong. looks like we can't just broad-stroke it, have to be careful about what we change

site/_docs/configuration.md
```diff
@@ -32,7 +32,8 @@ class="flag">flags</code> (specified on the command-line) that control them.
     <tr class="setting">
       <td>
         <p class="name"><strong>Site Source</strong></p>
-        <p class="description">Change the directory where Jekyll will read files</p>
+        <p class="description">Change the directory where Jekyll will
+        read files</p>
```
MartinRogalla: I don't see how this should improve the readability?
Personally I would like to see it much more like this:
``` html
<p class="description">Change the directory where Jekyll will read files</p>
```
or if the line *has* to be shorter, like this:
``` html
<p class="description">
    Change the directory where Jekyll will read files
</p>
```

Same comment for a few other occurrences.

chrisfinazzo:
I know there are at least a few places where it's not possible to move lines around because it's inside of a code block or part of a long URL. A bit surprised that I missed it considering that these are just regular paragraphs and a header, but I'll look into it and try to shape things up. 

For the record, inline links are hideous. Did someone on the Core team make that decision or did this just happen organically over time because GitHub encourages long URLs in their Markdown docs?

I _would_ create a PR to "fix" that, but I imagine it would be massive as it touches almost everything inside of `site`.

site/_docs/configuration.md
chrisfinazzo: Yeah, I need to look more closely at the tables, code blocks, and other things inside of `<pre>` tags. My original intent was just to get everything under 80 characters for each line, but I can see there is more work that needs to be done.

envygeeks: Paragraphs are a block tag, there must *always* be a new line after the `<p>` just so you know, those are just the unwritten style rules for HTML.  If it's a block level element imply it's block level by giving it a new line, if it's inline then imply it's inline without a new line.

envygeeks:
For what it's worth my vote will always go towards whatever makes the source of the docs more readable, so if you think it makes things more readable I think you should put it in if you like and see what everyone else thinks.  It never hurts to improve readability in all places!

parkr:
I guess in principle we'd want this, but it isn't worth it at this point. Thank you though!! :heart:

PR #3619, opened 2015-03-26
bin/jekyll: with no args, exit with exit code 1

### LINKED ISSUE #3584: Calling “jekyll” with no arguments gives a zero return value
I just did several rounds of editing a file followed by

    jekyll && cat _site/2015/03/17/test.html

before I realized that Jekyll hadn’t actually built the site: calling `jekyll` with no arguments prints the help message and quits. It’s good that Jekyll automatically prints some help information when this happens but I’m not sure that a zero exit code is appropriate. If the user intentionally wrote `jekyll --help` then this would make sense, but just calling plain `jeykll` is really an error and the resulting `$?` should reflect that.

parkr:
fixes #3584

### DIFF
--- a/bin/jekyll
+++ b/bin/jekyll
@@ -31,6 +31,7 @@ Mercenary.program(:jekyll) do |p|
     if args.empty?
       Jekyll.logger.error "A subcommand is required."
       puts p
+      abort
     else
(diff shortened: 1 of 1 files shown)


envygeeks:
:+1:

PR #3623, opened 2015-03-27
Should this have the "Warning" label?

chrisfinazzo:


parkr:
The entire section is called `Gotchas` so not sure if it makes sense here. :)

chrisfinazzo:
I know, this is kind of iffy. I wonder if we really need an entire section, as there's only one item there (Just stop at 3). I put it out here because the consequences of not following this instruction in a Travis setup is so bad. The only other place I could see putting this would be under *Troubleshooting*, but I'm not sure it belongs there.

parkr:
> I wonder if we really need an entire section, as there's only one item there (Just stop at 3).

:+1:

parkr:
Can you also put the `highlight` in the body of that?

chrisfinazzo:
Like this?

screen shot 2015-04-08 at 3 57 57 pm

That's...OK. I guess that's one way to do it. 😕

parkr:
oh ew.

parkr:
what does it look like now?

PR #3627, opened 2015-03-30
Removed minitest-profile

mdenhoedt:
I have removed the `minitest-profile` gem from the `Gemfile` and the `require minitest/profile` from the `test/helper.rb` file, because they are not used. They were added in [this commit](https://github.com/jekyll/jekyll/commit/f1edf9e692389f445eabbf45198722758026dd9d), but the `--profile` flag has been removed in [this commit](https://github.com/jekyll/jekyll/commits/master/script/test).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,7 +24,6 @@ gem 'jekyll_test_plugin_malicious'
 gem 'liquid-c', '~> 3.0'
 gem 'minitest'
 gem 'minitest-reporters'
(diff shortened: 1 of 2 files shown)


envygeeks:
We would rather have profiling so if it was accidentally removed readd it, do not remove it entirely.

mdenhoedt:
Ok, thank you for your quick response! I will see what I can do to add profiling back in :smile:

PR #3629, opened 2015-03-30
Use `Jekyll::Post`s for both LSI indexing and lookup.

justinweiss:
When looking for related posts, Jekyll was *indexing* `Jekyll::Post`
objects, but *finding* related posts based on `Jekyll::Post#content`. This
caused two problems:

1. Ruby 2.2 will warn on `==` if `<=>` throws an exception (and future Ruby
versions will surface that exception). Because `String`s can't be
compared with `Jekyll::Post`s, this warning was appearing all the time
while searching for related posts.

2. LSI won't return a post itself when searching for related posts. But
LSI could never tell that we were searching on a post, since Jekyll
passed post content, not a post object. With this fix, we can remove the
`- [post]` from `Jekyll::RelatedPosts#find_related`.

This is a more accurate fix for #3484.

### DIFF
--- a/lib/jekyll/related_posts.rb
+++ b/lib/jekyll/related_posts.rb
@@ -42,7 +42,7 @@ module Jekyll
     end
 
     def lsi_related_posts
-      self.class.lsi.find_related(post.content, 11) - [post]
+      self.class.lsi.find_related(post, 11)
     end
 
     def most_recent_posts
--- a/test/test_related_posts.rb
+++ b/test/test_related_posts.rb
@@ -22,12 +22,27 @@ class TestRelatedPosts < JekyllUnitTest
     setup do
       allow_any_instance_of(Jekyll::RelatedPosts).to receive(:display)
       @site = fixture_site({"lsi" => true})
-    end
-
-    should "use lsi for the related posts" do
       @site.reset
       @site.read
       require 'classifier-reborn'
+      Jekyll::RelatedPosts.lsi = nil
+    end
+
+    should "index Jekyll::Post objects" do
+      @site.posts = @site.posts.first(1)
+      expect_any_instance_of(::ClassifierReborn::LSI).to receive(:add_item).with(kind_of(Jekyll::Post))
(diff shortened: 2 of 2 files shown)


envygeeks:
:+1:

parkr:
Not sure if lsi should be this coupled to the model, considering that `Jekyll::Post` will be a goner by the launch of 3.0.

@mattr-, thoughts?

justinweiss:
Any other thoughts on this?

lib/jekyll/related_posts.rb
```diff
@@ -42,7 +42,7 @@ def build_index
     end
 
     def lsi_related_posts
-      self.class.lsi.find_related(post.content, 11) - [post]
+      self.class.lsi.find_related(post, 11)
```
parkr: LSI should really be based on the content of the post, no? Doesn't this change the functionality of the indexer?

justinweiss: [It shouldn't](https://github.com/jekyll/classifier-reborn/blob/133487722c311ff6c7f560cca40d393f36f70468/lib/classifier-reborn/lsi.rb#L47) -- both `add_item` and `lsi_related_posts` will call `#to_s` on what you pass into it. I'd assume you could also leave this function the same, and change `build_index` to call `add_item(x.content)`, but this way seemed clearer to me. 

Regardless, I feel like `build_index` and `lsi_related_posts` should be using the same kind of object -- either both Posts or both Strings. That way, you won't have to ` - [post]` at the end, because classifier-reborn is [already doing that on its side](https://github.com/jekyll/classifier-reborn/blob/133487722c311ff6c7f560cca40d393f36f70468/lib/classifier-reborn/lsi.rb#L237).

parkr: Oh, ok great! :+1:

justinweiss:
Thanks!

parkr:
Thank YOU! :heart: Sorry it took so long.

PR #3632, opened 2015-04-01
Removed unused method StaticFile::extname

mdenhoedt:
I have removed the unused method `StaticFile::extname`. The method is not used by any test and doesn't add value. Everywhere `File.extname(path)` is called directly, therefore `StaticFile::extname is useless`.

### DIFF
--- a/lib/jekyll/static_file.rb
+++ b/lib/jekyll/static_file.rb
@@ -25,10 +25,6 @@ module Jekyll
       File.join(*[@base, @dir, @name].compact)
     end
 
-    def extname
-      File.extname(path)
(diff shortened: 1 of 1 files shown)


parkr:
We should cache the return value of File.extname(filename).

mdenhoedt:
@parkr as you did in pull request [#3596](https://github.com/jekyll/jekyll/pull/3596)? I think that is a good idea, but still this method remains unused, or am I missing something?

parkr:
@mdenhoedt Yes! Want to submit a PR to my PR? :)

Issue #3626, opened 2015-03-30
Missing tests for static_file.rb


mdenhoedt:
There are no unit tests for the [StaticFile](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/static_file.rb) class in the map `/test`, are there? Is it a good idea to add `/test/test_static_file.rb` with unit tests?

envygeeks:
It is but it would probably be wasted effort since we are rewriting tests for RSpec anyways.

PR #3633, opened 2015-04-01
Added tests for Jekyll:StaticFile

mdenhoedt:
Added tests for `Jekyll:StaticFile` this should solve issue [#3626](https://github.com/jekyll/jekyll/issues/3626).

### DIFF
--- a/test/test_static_file.rb
+++ b/test/test_static_file.rb
@@ -1,0 +1,95 @@
+require 'helper'
+
+class TestStaticFile < JekyllUnitTest
+    def make_dummy_file(file_name)
+        temp_file = File.new(file_name, "w")
+        temp_file.puts("some content")
+        temp_file.close
+    end
+
+    def modify_dummy_file(file_name)
+        temp_file = File.open(file_name, "w")
+        temp_file.puts("more content")
+        temp_file.close
+    end
+
+    def remove_dummy_file(file_name)
+        File.delete(file_name)
+    end
+
+    def setup_static_file(base, dir, name)
+        @static_file = StaticFile.new(@site, base, dir, name)
+    end
+
+    context "A StaticFile" do
+        setup do
+            clear_dest
+            @site = Site.new(site_configuration)
+        end
+
+        should "have a source file path" do
+            static_file = setup_static_file("root", "dir", "file_name.html")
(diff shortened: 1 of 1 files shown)


envygeeks:
Are you sure this is not covered by a higher level test in features/ ?

mdenhoedt:
The test coverage for `StaticFile` is already nearly 100%. However, I think it is a good idea to have unit tests for it too, but it's up to you :smile:.

envygeeks:
I think it's a good idea to cover it.  Anybody else have any objects? /cc @jekyll/core

test/test_static_file.rb
```diff
@@ -0,0 +1,95 @@
+require 'helper'
+
+class TestStaticFile < JekyllUnitTest
+    def make_dummy_file(file_name)
+        temp_file = File.new(file_name, "w")
+        temp_file.puts("some content")
+        temp_file.close
```
parkr: Why not:

```ruby
File.open(filename, "w") do |f|
  f.puts "some content"
end
```

?

test/test_static_file.rb
```diff
@@ -0,0 +1,95 @@
...
+    end
+
+    def setup_static_file(base, dir, name)
+        @static_file = StaticFile.new(@site, base, dir, name)
+    end
+
+    context "A StaticFile" do
+        setup do
+            clear_dest
+            @site = Site.new(site_configuration)
```
parkr: I believe you can use `fixture_site` here!

parkr:
I also think it's a good idea to cover it. :)

parkr:
There's definitely an opportunity to DRY up this file (a lot of repetition right now). Would you mind tackling that?

mdenhoedt:
Thanks for the feedback. I am not an experienced Ruby programmer and I will definitely try to improve my pull request.

test/test_static_file.rb
envygeeks:
```ruby
File.write(filename, "some content")
```

parkr: ^^ that also works :)

parkr:
No problem! Let us know if you have any questions. :)

envygeeks:
:heart:

PR #3637, opened 2015-04-04
Removed inconsistency with period

Sourdoughh:


parkr:
Thanks!

lgp36018:
Please help me install new rom and restore recovery.because i see signal code kenel seandriod not enforce.

PR #3638, opened 2015-04-04
Expand collections documentation.

adunning:
The current documentation refers to a location below for a discussion of accessing document variables, but does not provide anything concrete. I have persevered with the albums example, and given something slightly complex in order to show how to access nested attributes (since I found it difficult to find documentation on this).

I also noted a method for setting default variables in collections and fixed a minor typo.

It would also be useful to add an example of how to access nested attributes on the current collection item without using a for loop (I am embarrassed to say that I don't know how to do it myself).

### DIFF
--- a/site/_docs/collections.md
+++ b/site/_docs/collections.md
@@ -37,6 +37,17 @@ collections:
     foo: bar
 {% endhighlight %}
 
+Default attributes can also be set for a collection:
+
+{% highlight yaml %}
+defaults:
+  - scope:
+      path: ""
+      type: my_collection
+    values:
+      layout: page
+{% endhighlight %}
+
 ### Step 2: Add your content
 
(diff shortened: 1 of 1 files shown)


site/_docs/collections.md
```diff
@@ -319,3 +330,47 @@ file, each document has the following attributes:
...
+        duration: "7:47"
+      - title: "Agnus Dei I, II & III"
+        duration: "6:49"
+{% endhighlight %}
+
+Every album in the collection could be listed on a single page with a template:
+
+{% highlight html %}
+{% raw %}
+{% for album in site.albums %}
```
parkr: Does this work as expected?

adunning: It does with Jekyll 2.5.3. I'm just not sure about the best way to output only the attributes for the current collection item (i.e. in its individual output); the solutions I have used seem rather unwieldy.

PR #3642, opened 2015-04-06
Reference links

chrisfinazzo:
Based on comments from #3616.

### DIFF
--- a/site/_docs/datafiles.md
+++ b/site/_docs/datafiles.md
@@ -8,9 +8,13 @@ In addition to the [built-in variables](../variables/) available from Jekyll,
 you can specify your own custom data that can be accessed via the [Liquid
 templating system](https://wiki.github.com/shopify/liquid/liquid-for-designers).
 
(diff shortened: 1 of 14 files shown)


parkr:
Thanks so much for all your hard work, @chrisfinazzo. I'm so sorry, but I think we're going to leave it how it is for now. I like the closeness of the link and the text – nothing is ever confusing when reading it. Having to search for the link reduces the usability of the Markdown, so we'll leave it as-is.

chrisfinazzo:
¯\\\_(ツ)\_/¯

As I said before, this was really an exercise to satisfy my own neuroses about how it should be done. The closest thing I could think of is a style guide for Jekyll (Comments, documentation, etc.) To some extent, this already exists inside of `CONTRIBUTING.markdown` and would only amount to a few lines; e.g, "When creating docs for a given feature, please use these conventions." As much as it would satisfy _me_ I understand if it's not something that we can realistically ask people to do.

I do still maintain that reference links look more "correct" if you're viewing Markdown source, but that's just me 😉

PR #3647, opened 2015-04-09
Add WOFF2 font format

nhoizey:


parkr:
@azizshamim Is this in the GHP mime types file? We're trying to maintain parity.

envygeeks:
I don't know how he would know that because he doesn't have access to their servers?

alfredxing:
I created a test repo, and it seems that WOFF files are served as `application/x-font-woff` (as expected), but WOFF2 files are still being served as `application/octet-stream`, which probably means GitHub hasn't added this on their end yet.

parkr:
> I don't know how he would know that because he doesn't have access to their servers?

@envygeeks Aziz works on GitHub's infrastructure. He knows. :)

@alfredxing Ok, then it's time for a support request! I'll send it now.

parkr:
On hold at the moment. Got a response from GitHub, but the W3C spec doesn't specify this MIME Type (and is vague about which one to use).

envygeeks:
The spec states `application/font-woff2` which will more than likely be what it will land as in the final spec.

parkr:
@envygeeks Are you looking at http://dev.w3.org/webfonts/WOFF2/spec/#IMT ?

envygeeks:
https://code.google.com/p/chromium/issues/detail?id=391289

parkr:
I'm convinced! Thanks. :smile:

nhoizey:
\o/

Thanks!

envygeeks:
:heart:

jdennes:
We've deployed the corresponding change on the GitHub side. GitHub Pages will now serve `.woff2` files with a `Content-Type` header value of `application/font-woff2`.

parkr:
Thank you so much @jdennes!!!!!!!

PR #3649, opened 2015-04-10
Fix header for RSS feeds with UTF8 characters.

schneems:
```
$ curl http://localhost:4000/feed.xml -I
HTTP/1.1 200 OK
Etag: 64af8c-2356-552805aa
Content-Type: text/xml; charset=utf-8
Content-Length: 9046
Last-Modified: Fri, 10 Apr 2015 17:17:30 GMT
Server: WEBrick/1.3.1 (Ruby/2.2.1/2015-02-26)
Date: Fri, 10 Apr 2015 17:17:34 GMT
Connection: Keep-Alive
```

Originally contributed to https://github.com/fitztrev/jekyll-utf8/pull/3

### DIFF
--- a/lib/jekyll/mime.types
+++ b/lib/jekyll/mime.types
@@ -1,8 +1,8 @@
 -# These are the same MIME types that GitHub Pages uses as of 26 January 2014
 
-text/html                             html htm shtml
+text/html;charset=utf-8               html htm shtml
 text/css                              css
-text/xml                              xml rss xsl xsd
+text/xml;charset=utf-8                xml rss xsl xsd
 image/gif                             gif
 image/jpeg                            jpeg jpg
 application/x-javascript              js


schneems:
Not sure if this is the best way to do this. It works...but it seems like a one off. I'm also not sure if you wanted "html" specified as well.

parkr:
Idea: can this be specified in the `mime.types` file? Can it be global for all text files?

parkr:
> I'm also not sure if you wanted "html" specified as well.

It's a good idea. :smile:

parkr:
Turns out this can be done in the `mime.types` file:

```text
text/html;charset=utf-8               html htm shtml
```

produces:

```ruby
{"html"=>"text/html;charset=utf-8", "htm"=>"text/html;charset=utf-8", "shtml"=>"text/html;charset=utf-8"}
```

:raised_hands: 

Thoughts, @schneems?

pathawks:
```xml
<?xml version="1.0" encoding="utf-8"?>
```

The encoding specified in the document will override the encoding in the HTTP header. Fortunately, [the encoding specified in the document is already UTF-8](https://github.com/jekyll/jekyll/blob/master/site/feed.xml#L5).

schneems:
> The encoding specified in the document will override the encoding in the HTTP header

It won't set a valid charset though. My feed wasn't being parsed as valid by w3c even with that https://github.com/schneems/schneems/blob/master/feed.xml

@parkr I like that approach better. Want me to update this PR?

parkr:
> @parkr I like that approach better. Want me to update this PR?

That would be amazing! :smile:

pathawks:
> My feed wasn't being parsed as valid by w3c even with that https://github.com/schneems/schneems/blob/master/feed.xml

What you have [here](http://dbtek.github.io/dbyll/feed.xml) validates for me.

[](http://validator.w3.org/feed/check.cgi?url=http%3A%2F%2Fdbtek.github.io%2Fdbyll%2Ffeed.xml)

schneems:
@pathawks that was after https://github.com/schneems/schneems/commit/4c70a5d77db132f914432790515cad3cff2b061b it was failing before that.

schneems:
Updated PR

pathawks:
It looks like it is being served by GitHub pages; plugins are disabled.

parkr:
> It looks like it is being served by GitHub pages; plugins are disabled.

@pathawks he runs it on heroku I think :)

pathawks:
[`http://dbtek.github.io/dbyll/feed.xml`](http://dbtek.github.io/dbyll/feed.xml)

Not served by `github.io`? :no_mouth:

parkr:
Thank so much @schneems!!!!!!

envygeeks:
My condolences to whoever uses WEBRick to server files on Heroku.  I think it's time I build a Jekyll pack so that that kind of insanity is stopped before it causes a train wreck for somebody (worse than it already has.) It can't be that hard to run jekyll build and then boot Nginx on that.

schneems:
```
$ dig www.schneems.com

; <<>> DiG 9.8.3-P1 <<>> www.schneems.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 34722
;; flags: qr rd ra; QUERY: 1, ANSWER: 3, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;www.schneems.com.		IN	A

;; ANSWER SECTION:
www.schneems.com.	1252	IN	CNAME	schneems.herokuapp.com.
```

I work for Heroku.

envygeeks:
Whether you work for Heroku has no bearing on what I said at all.

envygeeks:
Unless you guys don't want a build pack for Jekyll+Nginx for Heroku, then by all means say it.

parkr:
@schneems Are you using WEBrick? Isn't Puma recommended? Or is it just not that big of a deal for you?

envygeeks:
@parkr Puma can't serve static files sadly.  It's WEBRick that is the problem but if we built a proper pack we could crush most of these problems before they happen because it has proper mimetypes.

schneems:
@envygeeks there's already a dedicated jekyll buildpack i've not used it. I would instead prefer a micro-buildpack of some kind. So you can use multi-buildpack with the regular Ruby buildpack, then run the microbuildpack after it. For serving assets you could use ActionDispatch::Static middleware if you really wanted.

Honestly latency hasn't been that bad, my perc 95 is around 150ms



I'm also running it on more than one dyno, so that helps.

envygeeks:
I didn't know that was supported so I will definitely do that when I look into building a buildpack.

PR #3652, opened 2015-04-13
Security fix: OSVDB-120415 - Upgrade redcarpet to 3.2

christianv:
*Note*: Please release a new gem version of jekyll after merging this.

More information at: http://osvdb.org/show/osvdb/120415

`redcarpet Gem for Ruby contains a flaw that allows a cross-site scripting (XSS) attack. This flaw exists because the parse_inline() function in markdown.c does not validate input before returning it to users. This may allow a remote attacker to create a specially crafted request that would execute arbitrary script code in a user's browser session within the trust relationship between their browser and the server.`

https://github.com/diaspora/diaspora/commit/9fc00d0814807064e2dc7343fc330f7cdcf8112a
https://github.com/fiedl/your_platform/commit/8e707ebb947dd4277704849bcc4a383401d4bd0e
http://social.schiessle.org/display/b38b1460c2b201329b1f4860008dbc6c
https://gemnasium.com/gems/redcarpet/versions/3.2.3

/cc @parkr @envygeeks

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 gemspec
 
 gem 'pygments.rb', '~> 0.6.0'
-gem 'redcarpet', '~> 3.1'
+gem 'redcarpet', '~> 3.2.3'
 gem 'toml', '~> 0.1.0'
 gem 'jekyll-paginate', '~> 1.0'
 gem 'jekyll-gist', '~> 1.0'


parkr:
Hey @christianv! Would you mind explaining to me how this affects Jekyll, vendors that run Jekyll, and Jekyll-created sites? Each site is static, so generally they don't accept user input except by the site creator herself or himself.

christianv:
@parkr indeed, this would only happen in a use case where the site creator e.g. copy/pastes something themselves and pushes it up.

This is probably not super critical but I would suggest upgrading nonetheless.

Gemfile
```diff
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 gemspec
 
 gem 'pygments.rb', '~> 0.6.0'
-gem 'redcarpet', '~> 3.1'
+gem 'redcarpet', '~> 3.2.3'
```
parkr: This unnecessarily locks the user to using 3.2.x, instead of 3.x. I believe you can declare this as `~> 3.2, >= 3.2.3`.

parkr:
Keep in mind that this Gemfile is for Jekyll 3.x and is not shipped with the gem. That would be the 2.5-stable branch, I believe.

envygeeks:
This does affect our other services (that are not static.) so they are going down right now to be upgraded with the latest version of Redcarpet we can pull.  Thank you for the security report, I have also forwarded it to Ubuntu Security Team.

PR #3660, opened 2015-04-16
fixed install instructions

falkorichter:
as tested in https://github.com/sensorberg-dev/sensorberg-dev.github.io/blob/master/.travis.yml

### DIFF
--- a/site/_docs/continuous-integration.md
+++ b/site/_docs/continuous-integration.md
@@ -76,6 +76,11 @@ an explanation of each line.
 language: ruby
 rvm:
 - 2.1
+
+before_install:
+- travis_retry gem install html-proofer
+- travis_retry gem install jekyll
+
 script: ./script/cibuild
 
 # branch whitelist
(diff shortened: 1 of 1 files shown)


falkorichter:
But your code does not work... 

I followed the instructions and I´m getting an error:
https://travis-ci.org/sensorberg-dev/sensorberg-dev.github.io/builds/59362782

envygeeks:
@falkorichter because you don't have a proper Gemfile which is pretty much  the standard for Ruby and CI's but if you do not wish to use one there are directions on how to do what you wish to do without needing a Gemfile using .travis.yml.

falkorichter:
Then we should include a gemfile in the instructions. I just created a new jekyll site using "jekyll new" and there is no gemfile.

envygeeks:
That is not a relevant point... if you wish to use a CI then it's up to you to add the Gemfile for the CI.  We also don't include a .travis.yml file, you added that and Travis CI requires either a Gemfile that lists your dependencies or you to install those dependencies with their hook system so you must add that.  We do not make the assumption that you wish to default with a CI so we don't add a Gemfile by default.  We expect people will be using Github Pages mostly.

envygeeks:
The most basic Gemfile will do for most cases (that don't use external dependencies and what not):

```ruby
source "https://rubygems.org"
gem "jekyll"
```

falkorichter:
I will fix the pull request and mention a working gemfile. I find a CI integration jekyll for github page very useful and I was happy to read that there was even documentation on it. 

Why should the documentation not be complete, so that everyone can follow the instructions without any additional knowledge and it will work?

envygeeks:
Please open a new pull instead of adjusting this one.  Thanks.

PR #3663, opened 2015-04-17
Trim keys when reading CSV

mneumegen:
It took me a while to figure out that Jekyll doesn't trim the space between headings in a CSV. 

If I have this heading in my CSV: 
```name, github ```

To access the second column I have to use this hash: [' github'] instead of what I expected: ['github']

I added a trim to keys on a csv and a test case.

### DIFF
--- a/test/test_site.rb
+++ b/test/test_site.rb
@@ -410,6 +410,7 @@ class TestSite < JekyllUnitTest
         site = Site.new(site_configuration('safe' => false))
         site.process
 
+
         file_content = SafeYAML.load_file(File.join(source_dir, '_data', 'products.yml'))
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: test/test_site.rb [hunk 1]
```diff
@@ -410,6 +410,7 @@ def convert(*args)
         site = Site.new(site_configuration('safe' => false))
         site.process
 
+
```
pathawks: ?

envygeeks:
:-1: While I appreciate what you are doing, this is a problem with the CSV library, why was this not filed there and instead taken here?

pathawks:
Ruby/Ruby#575

mneumegen:
Yep makes sense

PR #3665, opened 2015-04-19
Set log level to debug when verbose flag is set

purp:
Adds Jekyll::LogAdapter#adjust_verbosity which ensures that --quiet always wins. Also refactors Jekyll::Configuration minorly to dry up the #source, #quiet?, and new #verbose? methods.

Tests included; all tests pass.

### DIFF
--- a/lib/jekyll/commands/build.rb
+++ b/lib/jekyll/commands/build.rb
@@ -23,7 +23,8 @@ module Jekyll
         # Build your jekyll site
         # Continuously watch if `watch` is set to true in the config.
         def process(options)
-          Jekyll.logger.log_level = :error if options['quiet']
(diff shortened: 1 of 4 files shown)


lib/jekyll/commands/build.rb
```diff
@@ -23,7 +23,8 @@ def init_with_program(prog)
         # Build your jekyll site
         # Continuously watch if `watch` is set to true in the config.
         def process(options)
-          Jekyll.logger.log_level = :error if options['quiet']
+          # Adjust verbosity quickly
```
parkr: Why the comment?

parkr:
Rock solid. :metal:

PR #3672, opened 2015-04-21
Remove extra space

chrisfinazzo:


envygeeks:
Punctuation-ally you just broke English. http://en.wikipedia.org/wiki/Sentence_spacing

parkr:
@envygeeks That wikipedia article says it's "widely debated," so removing the second space is only wrong for some people. I grew up using one space between sentences, and was chastised by Microsoft Word when I tried to use two instead of one. So, I'd say one it is. All the other sentences in our docs use one space only between them, so now it's just an argument of consistency. :)

envygeeks:
It's not widely debated, it's taught in all American schools...

envygeeks:
And to add, the *only* thing that is widely debated is whether or not single space is right or wrong.  Not whether or not double spacing is proper English... because double spacing is proper English.

troyswanson:
:+1: for single spacing, since that's how web browsers render everything, regardless of how many spaces you put after a period.

chrisfinazzo:
Thanks @parkr, clearly, you are one of my people 😄.

While I was taught "use 2 spaces" early on, college and grad school drilled it into me that this convention is misguided. APA and MLA Style both permit the use of two spaces, but in formal settings (Read: Academia) this is generally frowned upon in my experience.

Regarding AP Style, the rule is clear: using 2 spaces is wrong, don't do it.

Issue #2744, opened 2014-08-13
Deprecation notice for relative permalinks in noisy


zpao:
The notice says "Starting in 2.0...", however I'm now using 2.2 and explicitly opting out of absolute permalinks (not depending on the previous default).

Should this message get removed now that we're past 2.0? Or should it at least be relaxed when specifically opting out?

parkr:
@alfredxing, do you have time for this in the next week or so?

alfredxing:
@parkr Sure, I have exams but can probably work on this when I'm taking breaks from studying.

I found the underlying code, but I'm wondering whether we *should* remove the warning, if its sole purpose is to be annoying to get users to switch to absolute permalinks.

parkr:
Hm, that's true. I'm not sure if we can say opt-in or opt-out anymore -- maybe we just remove relative permalinks support altogether... i dig the feature myself (easily move files around etc) but it is prone to confusion and conflict.

alfredxing:
Addressed in above PR!

parkr:
:tada:

PR #3679, opened 2015-04-28
Remove relative permalinks

alfredxing:
Remove relative permalinks support for 3.0. Still provides an error message when the configuration option is set (this can be removed in a couple more minor/patch versions).

Ref #2744.


penibelst:
I like this breaking change because it will prevent some beginners from creating deep nested clusters.

parkr:
:clap:

PR #3683, opened 2015-05-01
Use the proper names

chrisfinazzo:
To me, it's still Mac OS X. Old habits die hard...


envygeeks:
Well why didn't you change 1.9.1 to 2.0.0 too then? :smiling_imp:

chrisfinazzo:
@envygeeks,

It's updated now, is this what you were looking for? Looking through `rbenv install --list` I don't see anything about a patch level, so I would assume this is correct...

parkr:
:shipit:

alfredxing:
:tada:

envygeeks:
:heart:

PR #3692, opened 2015-05-10
Add script/rubyprof to generate cachegrind callgraphs

fw42:
Inspired by [script/stackprof](https://github.com/jekyll/jekyll/blob/63792ad322183f159c530e9f1112b94bc333d200/script/stackprof), this adds a script to generate [ruby-prof](https://github.com/ruby-prof/ruby-prof) profiles in KCachegrind calltree format (can be opened using qcachegrind under Mac OS X).

Output looks something like this:

screen shot 2015-05-09 at 10 32 43 pm

Thoughts @parkr?

### DIFF
--- a/script/rubyprof
+++ b/script/rubyprof
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+export BENCHMARK=1
+
+TEST_SCRIPT="Jekyll::Commands::Build.process({'source' => 'site', 'full_rebuild' => true})"
+
+RUBY=$(cat <<RUBY
+  require 'ruby-prof'
+  result = RubyProf.profile{ ${TEST_SCRIPT} }
+  printer = RubyProf::CallTreePrinter.new(result)
+  filename = "tmp/ruby_prof_#{rand 10000}"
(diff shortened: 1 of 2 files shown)


parkr:
This is _super_ cool.

script/rubyprof
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/ruby
+
+require 'ruby-prof'
+require 'jekyll'
```
parkr: can we lock to `bundle exec` here?

script/rubyprof
```diff
@@ -0,0 +1,18 @@
...
+
+RubyProf.start
+
+at_exit do
+  result = RubyProf.stop
+  printer = RubyProf::CallTreePrinter.new(result)
+  filename = "/tmp/ruby_prof_#{rand 10000}"
```
parkr: can we write to local `tmp` directory? this would vomit on windows :(

fw42:
Updated. It's closer to stackprof now. btw I think stackprof can generate those files too. Not sure how exactly the output differs.

PR #3696, opened 2015-05-11
Make Jekyll::Cleaner#existing_files more lazy

fw42:
Small micro-optimization. Will probably not make a big difference for most sites, but on our ~6000 files site, it shaves off about a second off build time.

The optimization is to not call the `keep_file_regex` and `keep_dirs` methods in every single iteration of the block but only once at the beginning. Also use `Regexp.union` instead of that ugly hack.

@parkr


envygeeks:
My only complaint is the commit message, it's misleading because nothing here is actually lazier.

fw42:
updated the commit message

PR #3697, opened 2015-05-12
Don't check binary files for yaml header

fw42:
This is another optimization for large sites. We have a site with about ~6000 files, about ~4000 of which are images. I noticed that even when running with incremental regeneration, starting the build process takes up to 5 seconds (which imho is way too long, considering that no time at all is spent in Liquid in that case).

Turns out some part of that time is due to checking if a file has a YAML header (to determine if it is a page or a static file). To do that, the file has to be opened and read (which is at least three system calls per file). With 4000 files, this can be noticeable.

This PR adds a check that prevents Jekyll from reading files that have a file extension commonly associated with binary files, since those files will most likely not have a valid YAML header.

This shaves off about 500-750ms of the time it takes to run `Jekyll::Reader#read` on my site (on an old-ish Macbook). Not sure if this is worth it for most people or not, but for me it's a ~10% improvement in start-up time, so I thought I'd suggest it.

@parkr

### DIFF
--- a/lib/jekyll/reader.rb
+++ b/lib/jekyll/reader.rb
@@ -32,7 +32,7 @@ module Jekyll
       dot = Dir.chdir(base) { filter_entries(Dir.entries('.'), base) }
       dot_dirs = dot.select{ |file| File.directory?(@site.in_source_dir(base,file)) }
       dot_files = (dot - dot_dirs)
-      dot_pages = dot_files.select{ |file| Utils.has_yaml_header?(@site.in_source_dir(base,file)) }
+      dot_pages = dot_files.select{ |file| !Utils.binary_file_extension?(file) && Utils.has_yaml_header?(@site.in_source_dir(base,file)) }
       dot_static_files = dot_files - dot_pages
 
       retrieve_posts(dir)
--- a/lib/jekyll/utils.rb
+++ b/lib/jekyll/utils.rb
@@ -108,6 +108,11 @@ module Jekyll
       !!(File.open(file, 'rb') { |f| f.read(5) } =~ /\A---\r?\n/)
     end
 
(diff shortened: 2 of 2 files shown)


envygeeks:
Your code is superfluous and assumptive... and missing hundreds of other "binary extensions".

I'm :-1: on this.  Extension does not necessarily correlate with file type, and doing this kind of test to begin with (or even accepting it) gives our source the feeling that we are doing it wrong everywhere... that's now how we should be doing things whether on a static generator or not (especially in the not.)

Setting aside any benches that are done (because they are considered invalid too since the source is considered invalid) this kind of thing should be part of a larger optimization that uses thread pools to actually make the source truly faster and consume as much IO and CPUTime as we can that or this or at minimum it should be noted how superfluous this source is and give an option to disable it entirely (though IMO that's just bad for business.)

envygeeks:
What I mean by all the above as a TLDR: Any check like the check you did is consider a huge no-no.

fw42:
Fair enough. Thanks for taking a look.

parkr:
Thoughts on a build cache here? Touching every single file multiple times definitely slows us down. It'd be good to reduce the calls here as much as possible. We are only reading 5 bytes afaik but if we could somehow optimize this that would help a whole awful lot. I'd be weary of keeping a file extension blacklist but if there were a way to reliably tell if something wasn't a text file without having to read it, that would be excellent!

envygeeks:
I like the idea of a file-system cache that keeps track of attributes and mimes, it would bring in the improvement @fw42 needs while allowing us to stay away from the assumption that extension will match actual file type.

fw42:
> Thoughts on a build cache here?

You mean cache if a file had the header or not? And if not, don't look for a header again the next time?

Issue #3624, opened 2015-03-28
Auto-generation silently failing


cjerdonek:
I'm a new Jekyll user on Mac OS X.  I'm using Jekyll with [this project](https://github.com/sfbrigade/sfbrigade.github.io), but I can't seem to get auto-generation working (e.g. when changing files in the `_includes` directory).

I'm using Ruby 2.1.1 with Jekyll 2.4.0, which are the versions specified here: https://pages.github.com/versions.json (which is what appears in the project's Gemfile).  I've spent time looking through the troubleshooting areas of the docs for ideas on this, as well as at old issues in the tracker (though not exhaustively).  But I couldn't find anything.  Thanks in advance for any help.

pathawks:
For troubleshooting, you should check out [talk.jekyllrb.com](https://talk.jekyllrb.com)

cjerdonek:
@pathawks I wasn't able to find any relevant discussions on that site.

I'm not experiencing any error messages.  The feature simply doesn't work.  Another person using the same project says it works for them, which suggests that it's not a configuration issue.

cjerdonek:
Okay, so it seems to regenerate if `--force_polling` is used, as suggested in https://github.com/jekyll/jekyll/issues/2926.  But that issue still doesn't explain why it's not working without `--force_polling`.  It seems like a bug somewhere.

cjerdonek:
I suspect it's a case-sensitivity issue.  Jekyll reports the following:

    Source: /Users/chris/dev/codeforsf/sfbrigade.github.io

But running the following gives--

    $ pwd
    /Users/chris/Dev/codeforsf/sfbrigade.github.io

pathawks:
> The feature simply doesn't work.

WFM :poop:

I cloned sfbrigade/sfbrigade.github.io, ran `bundle install`, `bundle exec jekyll serve`, and edited the file `_includes/open-graph.html`  
The site was updated to reflect the changes I made.

My path also has a mix of upper and lowercase letters:

```sh
$ pwd
/Users/pat/GitHub/sfbrigade.github.io
```

Jekyll recognizes this:

```text
$ bundle exec jekyll serve
Configuration file: /Users/pat/GitHub/sfbrigade.github.io/_config.yml
            Source: /Users/pat/GitHub/sfbrigade.github.io
       Destination: /Users/pat/GitHub/sfbrigade.github.io/_site
      Generating... 
                    done.
[...]

PR #3704, opened 2015-05-16
Fix #3653: Add a doctor helper to test pwd's.

### LINKED ISSUE #3653: Ensure the Dir.pwd and the actual working directory line up
Sometimes `Dir.pwd` returns the `$(pwd)` in downcased or otherwise incorrect form. In `jekyll doctor`, we should check this and it should be Windows-compatible as well.

Ref: https://github.com/jekyll/jekyll/issues/3624.

envygeeks:
This is just a basic test (as implied through a lot of the conversation) and will probably not suffice for all but according to what I have read, this basic test can detect most anomalies so we can warn you but it's not bulletproof by any means.

Ref: #3624
/cc @cjerdonek

### DIFF
--- a/lib/jekyll/commands/doctor.rb
+++ b/lib/jekyll/commands/doctor.rb
@@ -30,6 +30,7 @@ module Jekyll
 
         def healthy?(site)
           [
+            fsnotify_buggy?(site),
             !deprecated_relative_permalinks(site),
             !conflicting_urls(site)
           ].all?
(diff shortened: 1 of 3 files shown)


lib/jekyll/commands/doctor.rb
```diff
@@ -59,8 +60,24 @@ def conflicting_urls(site)
           conflicting_urls
         end
 
-        private
+        def fsnotify_buggy?(site)
+          return true if !Utils::Platforms.osx?
+          if Dir.pwd != `pwd`.strip
+            Jekyll.logger.error "  " + <<-STR.strip.gsub(/\n\s+/, "\n  ")
+              We have detected that there might be trouble using using
```
cjerdonek: "using using"

cjerdonek:
Thanks for doing this.

lib/jekyll/commands/doctor.rb
envygeeks: Fixed, and because of that you forced me to read the rest of it as a double check and I noticed I did `docker` instead of `doctor` in my platforms addition.

parkr:
:shipit:

Issue #3705, opened 2015-05-16
[Performance] Sorting files in Jekyll::Reader


fw42:
When running Jekyll with incremental regeneration on a site with about 6000 files (about 4000 of which are images, the rest is pages), it takes about ~5 seconds on my machine for Jekyll to complete the build (no changes were made, so that's 5 seconds to realize that nothing needs to be done - those 5 seconds run again on every file modification when using watch mode).

I noticed that `Jekyll::Reader` sorts the list of pages and the list of static files that it generates after scanning the filesystem and after profiling it, I saw that it spends a considerable amount of time comparing strings. Removing those calls to `sort_by!` shaves off about 1-1.5 seconds of the mentioned 5 seconds for me (without changing the generated output!).

I couldn't find a reason why the file lists need to be sorted. Nothing in Jekyll seems to depend on it. What am I missing? What's the use case for this?

If this isn't necessary, I can make a PR to remove it.

@parkr @envygeeks thoughts?

envygeeks:
It probably meant something back in Mad Max Times (Ruby 1.8) but now days it's probably meaningless.  Personally (and as always my opinion doesn't reflect on everyone) I say it should go if it provides performance and doesn't break any tests (that that our tests are infallible.)

envygeeks:
Oh, I forgot to put the obligatory :+1: from me.

fw42:
Cool. Will make a PR then.

fw42:
Hm I guess this can potentially change the generated output. Was just a coincidence that it didn't for me.

You can do stuff like
```
{% for file in site.static_files %}
  ...
{% end %}
```

My PR would possibly change the output of that.

Maybe there is another way to fix this. Will think about it some more.

envygeeks:
@parkr what do you think of just adding a sort helper (if we don't already have one) so that users can opt to do that themselves if they wish.

fw42:
maybe we could do something like overwrite the `Site#static_files` method to do the sorting there, on demand, so that at least you only sort things if your template actually references this variable...

parkr:
> maybe we could do something like overwrite the Site#static_files method to do the sorting there, on demand, so that at least you only sort things if your template actually references this variable...

That's not a bad idea at all!

PR #3707, opened 2015-05-16
Performance: Sort files only once

fw42:
Fixes https://github.com/jekyll/jekyll/issues/3705.

Turns out the reason that it was slow for me was not that it was sorting at all but that it was sorting the same list of files over and over again.

Solution: Sort only once at the end of `#read`, not every time we load more files from a directory (those functions are called recursively via `#read_directories`). This also ensures backwards compatibility for the regular code path (unless you have a plugin that directly calls `#retrieve_posts` etc. or stuff like that).

@envygeeks @parkr @alfredxing

### DIFF
--- a/lib/jekyll/reader.rb
+++ b/lib/jekyll/reader.rb
@@ -15,8 +15,16 @@ module Jekyll
     def read
       @site.layouts = LayoutReader.new(site).read
       read_directories
+      sort_files!
       @site.data = DataReader.new(site).read(site.config['data_source'])
       CollectionReader.new(site).read
+    end
+
+    # Sorts posts, pages, and static files.
+    def sort_files!
+      site.posts.sort!
+      site.pages.sort_by!(&:name)
+      site.static_files.sort_by!(&:relative_path)
     end
 
     # Recursively traverse directories to find posts, pages and static files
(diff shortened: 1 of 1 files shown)


envygeeks:
:+1:

alfredxing:
:+1:

What's the performance improvement you're seeing with this?

lib/jekyll/reader.rb
```diff
@@ -15,10 +15,18 @@ def initialize(site)
     def read
       @site.layouts = LayoutReader.new(site).read
       read_directories
+      sort_files
```
parkr: `sort_files!`?

fw42:
> What's the performance improvement you're seeing with this?

not crazy much.. about 500ms of ~5s (for initial build and for each regeneration cycle) on a site with about 6000 files. but I guess you have to start somewhere :-)

parkr:
:shipit:

envygeeks:
Pending responses it'll merge!

fw42:
Updated: Renamed the method

fw42:
Would you mind releasing a beta5 release to Rubygems?

envygeeks:
It'll happen sometime in the next few days because I have a few more changes I would like to get into beta5 (aside from our new JRuby support.) Like thread pooling and possibly even collection fixes.

envygeeks:
If you need though I can cut a new release of our master image for Docker.

parkr:
Why not just call those beta6, @envygeeks?

PR #3713, opened 2015-05-19
Don't crash when reading/writing Marshal

fw42:
@parkr @envygeeks, missed a small detail in the other PR. Should have tested this better :-(

### DIFF
--- a/lib/jekyll/regenerator.rb
+++ b/lib/jekyll/regenerator.rb
@@ -130,7 +130,7 @@ module Jekyll
     #
     # Returns nothing.
     def write_metadata
-      File.open(metadata_file, 'w') do |f|
+      File.open(metadata_file, 'wb') do |f|
         f.write(Marshal.dump(metadata))
       end
     end
@@ -164,6 +164,9 @@ module Jekyll
           Marshal.load(content)
         rescue TypeError
           SafeYAML.load(content)
+        rescue ArgumentError => e
+          Jekyll.logger.warn("Failed to load #{metadata_file}: #{e}")
+          {}
         end
       else
         {}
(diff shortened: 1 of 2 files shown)


lib/jekyll/regenerator.rb
```diff
@@ -130,7 +130,7 @@ def add_dependency(path, dependency)
     #
     # Returns nothing.
     def write_metadata
-      File.open(metadata_file, 'w') do |f|
+      File.open(metadata_file, 'wb') do |f|
```
fw42: Marshal is binary data, so this can crash on some contents :-( Sorry I missed that.

lib/jekyll/regenerator.rb
```diff
@@ -164,6 +164,9 @@ def read_metadata
           Marshal.load(content)
         rescue TypeError
           SafeYAML.load(content)
+        rescue ArgumentError => e
```
fw42: This makes it so that we don't crash the whole process and leave the user wondering if we ever fail to read the metadata for some reason (for example because it crashed because of bugs like the above).

lib/jekyll/regenerator.rb
fw42: For reference, the error that you would get is:
```
lib/jekyll/regenerator.rb:138:in `write': "\xD2" from ASCII-8BIT to UTF-8 (Encoding::UndefinedConversionError)
```

envygeeks:
Do you mind adding regression tests for these please? :dart:

fw42:
Added a test for reading corrupted files. Not sure how to test the other thing (but I don't really think a test is necessary for that).

envygeeks:
I'm +1 on this fix, I didn't get this problem but it's nice that you noticed it before it became a problem.

alfredxing:
:+1: I didn't have any issues with just `w`, but I guess it might have broken slightly more sensitive systems.

PR #3719, opened 2015-05-20
Fix #3718: Add support for Title Casing.

### LINKED ISSUE #3718: Port Rails title case as filter
Currently we only have `{{ variable | capitalize }}` which applies sentence case.

`{{ variable | title }}` should upcase the first letter of every word. Not available in Liquid, needs a new tag utilising Rails title case.

See: https://talk.jekyllrb.com/t/liquid-capitalize-tag-behaviour/760

envygeeks:


penibelst:
I won’t agree. This is a language-specific filter. Many natural languages have no casing.

envygeeks:
The last time I checked nobody was forcing you to use this method? I could be wrong though, did I enforce usage of this somewhere and not notice it?

parkr:
This is fine with me. All of Jekyll right now is English-centric so this being the same isn't a regression in our language support.

envygeeks:
This isn't necessarily language-specific unless you make it so, it intentionally takes a naive approach and just capitalizes all words regardless, it does not use title case from any given language, it does not ask you to use it, claiming this is language centric is like claiming that "downcase" or "upcase" (in ruby-core) is language specific when they do the same thing, they do as told without any specificity.

parkr:
A title has a language-specific meaning, though, so it's not the same as simply downcase or upcase. Even in English, titles differ: some news organizations, for example, only capitalize the first character of the title of an article, whereas others capitalize the first character of only "important" words (i.e. not "a", "an", "of", etc). So maybe the proper understanding is that there are multiple ways of doing titles even within one language, so it's not language-centric because languages can have multiple standards.

penibelst:
> it intentionally takes a naive approach and just capitalizes all words regardless

@envygeeks What is the use case for the filter? I don’t know any language that requires such a hard capitalization. Even in English only some words are capitalized:

> the first word and all nouns, pronouns, adjectives, verbs and adverbs, but generally not articles, conjunctions and short prepositions

In @parkr’s [blog titles](https://byparker.com/), you can see a human capitalization approach. His writing style is different from the new filter. English writers will raise issues to improve this behaviour.

Maybe you can add a language shortcode to make clear, what the rules are?

```
{{ "the-immediate-future-of-jekyll" | split: "-" | join: " " | titlecase: "en" }}
=> The Immediate Future of Jekyll
```

parkr:
> Maybe you can add a language shortcode to make clear, what the rules are?

@penibelst As I mention in my comment, it's really not language-specific.

penibelst:
> As I mention in my comment, it's really not language-specific.

What is the use-case for this filter? Where could you use it?

envygeeks:
> What is the use-case for this filter? Where could you use it?

Wherever the user wants to use it, we are not dictators, I was doing what a user asked for, if you don't like it that's really not a concern I'm mithered with because I refuse to dictate how people use a feature.

envygeeks:
I also want to mention that this pull was put on hold because of the same concerns were already brought up in the originating ticket, we don't allow people to customize which words are capitalized so they can define the semantics of their language or ruleset and it was agreed we did not want to add yet another option to Jekyll for that so until somebody could come up with a better idea this was just put on hold.  I should probably close it so that people don't keep bringing up this argument.

PR #3732, opened 2015-05-25
Fix it so that 'blog.html' matches 'blog.html'

envygeeks:


parkr:
Looks good to me. /cc @jekyll/core for another pair of :eyes:

alfredxing:
:+1:

PR #3733, opened 2015-05-25
Add support for blog.html > blog/ but < blog/index.html like Nginx.

envygeeks:
This adds support for blog/ being less than blog.html when routing but not greater than blog/index.html so that if you have a URL blog.html and blog/archive/page.html then you don't end up with a directory listing, you end up with blog.html like you probably expected... the way that NGINX works when you do simple try_file routing.

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -99,17 +99,28 @@ module Jekyll
           opts
         end
 
-        # Custom WEBrick FileHandler servlet for serving "/file.html" at "/file"
-        # when no exact match is found. This mirrors the behavior of GitHub
-        # Pages and many static web server configs.
+        # Allows files to be routed in a pretty URL in both default format
+        # and in custom page/index.html format and while doing so takes into
+        # consideration importance of blog.html > blog/ but not > blog/index.html
+        # because you could have URL's like blog.html, blog/archive/page.html
+        # and in a normal circumstance blog/ would be greater than blog.html
+        # breaking your entire site when you are playing around, and I
+        # don't think you really want that to happen when testing do you?
+
         def custom_file_handler
           Class.new WEBrick::HTTPServlet::FileHandler do
             def search_file(req, res, basename)
-              if file = super
-                file
+              if (file = super) || (file = super req, res, "#{basename}.html")
+                return file
+
               else
-                super(req, res, "#{basename}.html")
+                file = File.join(@config[:DocumentRoot], req.path.gsub(/\/\Z/, "") + ".html")
+                if File.expand_path(file).start_with?(@config[:DocumentRoot]) && File.file?(file)
+                  return ".html"
+                end
               end
+
+              nil
             end
           end
         end


parkr:
Does this change the tests? Can you add one for this, please?

envygeeks:
If there are tests then they are obscure as hell so I'm going to assume there aren't any tests to begin with and that Jekyll doesn't even do that kind of high level testing.

lib/jekyll/commands/serve.rb
```diff
@@ -98,17 +98,22 @@ def webrick_options(config)
...
-        # Custom WEBrick FileHandler servlet for serving "/file.html" at "/file"
-        # when no exact match is found. This mirrors the behavior of GitHub
-        # Pages and many static web server configs.
         def custom_file_handler
           Class.new WEBrick::HTTPServlet::FileHandler do
             def search_file(req, res, basename)
-              if file = super
-                file
+              if (file = super) || (file = super req, res, "#{basename}.html")
+                then file
```
parkr: `then`?

lib/jekyll/commands/serve.rb
```diff
@@ -98,17 +98,22 @@ def webrick_options(config)
...
               else
-                super(req, res, "#{basename}.html")
+                return ".html"
```
parkr: what does this do?

lib/jekyll/commands/serve.rb
envygeeks: I use then throughout a lot of my code, it's actually a part of Ruby.

lib/jekyll/commands/serve.rb
envygeeks: That's actually a feature that takes advantage of what might be (a bug?) in Webrick.  If you  return ".html" it will actually concatinate the path with ".html" so if you visit /blog and it knows you are at /blog and you return ".html" it will give you "blog.html"

lib/jekyll/commands/serve.rb
parkr: Haha I know, I was just wondering why you chose to include it. We never use `then` in any of the other code in Jekyll

lib/jekyll/commands/serve.rb
```diff
@@ -98,17 +98,22 @@ def webrick_options(config)
...
 
-        # Custom WEBrick FileHandler servlet for serving "/file.html" at "/file"
-        # when no exact match is found. This mirrors the behavior of GitHub
-        # Pages and many static web server configs.
         def custom_file_handler
           Class.new WEBrick::HTTPServlet::FileHandler do
             def search_file(req, res, basename)
-              if file = super
-                file
+              if (file = super) || (file = super req, res, "#{basename}.html")
```
parkr: did you mean to do `"#{basename}/index.html"` here?

envygeeks:
I've updated this pull with the current version my client sites currently use that does not cause webrick to 500 error and allows it to still 404 properly.

lib/jekyll/commands/serve.rb
envygeeks: No, because that first part was copied directly from the *original* source.

envygeeks:
I think I might update this one last time to prevent path traversal, even though it's supposed to be a development thing I have this funny feeling some people might use it in production so I should probably validate the path.

parkr:
:shipit:

envygeeks:
Gonna merge this into the other pull I mentioned and ship a new pull with both in one.

PR #3690, opened 2015-05-09
initial implementation of posts as a collection

parkr:
This PR moves posts functionality into `Collection` and `Document`. Still very much a work-in-progress.

To-do list:

- [ ] Permalink style/structure consistency (dates!)
- [ ] Liquid consistency
- [x] Filename validation (right now, not customizable)
- [x] `published?`
- [ ] Categories from subdirs
- [x] Categories from metadata
- [x] Tags from metadata
- [ ] Excerpts (auto-generate only when asked for (Liquid::Drop?))
- [ ] Auto-generated titles from slug
- [ ] Custom `<=>` with the date in there (maybe we get that for free with the `YYYY-MM-DD-` prefix?)
- [x] Permalink style shortcuts (currently only allowed for "dated" documents in the posts collection)
- [ ] `id` ~> `slug`
- [ ] `previous`, `next`
- [ ] `draft?`
- [ ] Reading in drafts at all.

Fixes #3169.

/cc @benbalter

Issue #3169, opened 2014-11-29
Make pages and posts instances of Jekyll::Document.


parkr:
Can't take credit for this idea. @benbalter said it. The idea is that posts are a sort of collection, and pages are as well if you think kind of tilt your head and squint a little. It would make sense for them all to share a common interface. We abolish the Convertible module and make anything that was once convertible a Document, et voila.

afeld:
Ha, just came to this same epiphany independently... which I guess makes me the Leibniz (the less well-known or handsome one) to your Newton :chart_with_upwards_trend: 

Really, I think *everything* could represented as collections (pages, posts, and data). Content is data! Everything is data!

![data all the things](http://cdn.meme.am/instances/500x/30793439.jpg)

Making the APIs between those existing data types completely consistent is really appealing, and it would simplify the internals a lot too. Might be worth thinking through what that implementation (rewrite?) would look like, and layering on APIs (maybe via a plugin gem?) to provide backwards compatibility.

alfredxing:
I'm experimenting with this (posts only right now), and got it to somewhat work. A big issue I came across is how we are going to handle all of the post-specific features (like categories, tags, dates, permalink styles, etc.)?

parkr:
> post-specific features (like categories, tags, dates, permalink styles, etc.)?

For categories & tags, we need an idea of `group_by` filters. Essentially that's what they do. The permalink styles could be applied with front matter defaults (or even just like them, i.e. collection name-based). As for dates, we'd need to bring over the `MATCHER` and `valid?` ideas to Document.

fulldecent:
I'm chiming in as a user IMHO to discuss our API.

 **The `_` prefix is good, it clearly demarcates Jekyll-land and asset-land**

For example, just looking at it, I know that `/assets/bear.jpg` will wind up at `example.com/assets/bear.jpg`. Consider a world without `_` prefixes, and you are editing index.html and wanted to link to `cars/redCar.html`... you would need to first check `_config.yml` to see if `cars/` is a "special Jekyll folder".

**"Pages" and "Folders" are preferred terminology, with "Posts" being a specific type of "Page"**

This is intuitive and keeps convention web authors have used for decades. Here's an example of something that makes sense:

    {{ site.books | where:"ISBN","4109147509740" | map:"url" }}

This searches `/path/to/site/_books/*` and the Front Matter of each page there. Which is no more special than searching JSON files in the `_data/` directory:

[...]

PR #3735, opened 2015-05-25
Configurable way do define default sort order for collections

### LINKED ISSUE #3539: Sort on nested values in collections not working
Given a collection that has documents with a pattern that look something like:

```yaml
time:
  start: 2015-05-11 18:00:00 -8
  end: 2015-05-12 16:00:00 -8
```

And a layout that has something like this:

```html
<ul id="calendar">
  {% assign events = site.events | sort: 'time.start' %}
  {% for page in events %}
  <li class="event">
    <a href="{{ page.url }}">
      <span class="month">{{ page.time.start | date: "%B" }}</span>
      <span class="day">{{ page.time.start | date: "%-d" }}</span>
    </a>
  </li>
  {% endfor %}
</ul>
```

Jekyll just doesn't sort them using that nested value at all.

If I move the values out into top-level properties, like:

```yaml
date_start: 2015-05-11 18:00:00 -8
date_end: 2015-05-12 16:00:00 -8
```

And modify the layout to use:

```html
{% assign events = site.events | sort: 'date_start' %}
```

It works as expected.

### LINKED ISSUE #3720: Configurable way do define default sort order for collections
In order to :
- help reduce the gap between document collections and posts (#3169 and #3690)
- help fixing #3539
- port #670 for collection (:grin:)

I started working on configurable way to define default sort order for collections in the metadata
(thus allowing to sort collections by reverse date provided that the documents store a date field in the data Hash).

I used the following `SQL/Rails` syntax in `_config.yml` :

```yaml
my_date_sorted_collection:
  output: true
  order: date DESC, slug
```

The string format supports multiple comma separated properties and nested fields (`time.start ASC`) (with limit to prevent abuse).

## Implementation

I created a `Jekyll::Sorter` module with an `order` method.

It returns and caches a `lambda` that can be used with the `Enumerable#sort` method. ie:

```ruby
document_array.sort(&:Jekyll::Sorter.order('date DESC, slug'))
```

I made Post and Collection use this instead of the default `sort!` + `<=>` methods.

The properties are fetched from the data attribute of the object elements in the array.

## WIP

Not enough to start a PR, but it should be a nice starting point for a discussion.

You can:
- check out progresses [here](https://github.com/goodtouch/jekyll/tree/add-configuration-for-collection-order)
- or quick-start a blog with a vendored gem using this branch for those who want to test & hack things:

(OMG I should have done this a long time ago!)

```
git clone --branch add-configuration-for-collection-order https://github.com/goodtouch/jekyll-dev-blog.git
cd jekyll-dev-blog
./script/bootstrap
./script/server
```

## To-do list:

- [X] Add Jekyll::Sorter module implementing the `order` method
- [X] Support nested fields (`{'time' => {'start' => 'whatever'}}`, Jekyll::Sorter.order("time.start ASC"))
- [X] Validate format of the order string
- [X] Limit nesting level (2) and number of fields (2) to prevent GitHub Page potential abuses (easy to change constants)
- [ ] Gracefully handle nil fields
- [X] Make Posts and Collections use the same Jekyll::Sorter.order (with different default arguments) when populating liquid hashes
- [X] in-code documentation
- [X] add a jekyll filter (temporarily named order for now as it's not backward compatible with the sort filter)
      this one should fix #3539 if using order: 'time.start' instead of sort
- [ ] add doc in site/_docs
- [ ] add some tests
- [X] add some benchs

goodtouch:
In order to :
- help reduce the gap between document collections and posts (#3169 and #3690)
- help fixing #3539
- port #670 for collection (:grin:)

I started working on configurable way to define default sort order for collections in the metadata
(thus allowing to sort collections by date for example).

You can use the following syntax in `_config.yml` :

``` yaml
my_date_sorted_collection:
  output: true
  order: data.date DESC, path
```

The string format supports multiple properties and nested fields (`data.time.start ASC`) (with limit to prevent abuse).

It generates static methods to prevent string manipulation performance issues (see benchmarks in #3720)

If you want to test & hack things, you can quick-start a blog with a vendored gem using this branch with:

```
git clone --branch add-configuration-for-collection-order https://github.com/goodtouch/jekyll-dev-blog.git
cd jekyll-dev-blog
./script/bootstrap
./script/server
```
## To-do list:
- [X] Support nested fields (Jekyll::Sorter.order(docs, "data.time.start ASC"))
- [X] Validate format of the order string
- [X] Limit nesting level (2) and number of fields (2) to prevent GitHub Pages potential abuses
    (easy to change constants)
- [X] Make Posts and Collections use Jekyll::Sorter.order when populating liquid hashes
- [X] Add a Liquid filter (named order for now as it's not backward compatible with the sort filter)
    this one should fix #3539 if using order: 'time.start' instead of sort
- [X] Add validations
- [X] Count and limit maximum generated order methods to prevent abuse.
- [X] Option to restrict allowed top level fields for sorting (`date`, `slug`, `path`, ...) and fields that can be used with nesting (`data`)
- [X] Fix existing tests
- [X] Code documentation
- [X] Benchmarks (see #3720)
- [X] Site doc
- [X] Add tests for Sorter
- [X] Add tests for Liquid filter
- [X] Add a Cucumber feature?
- [X] Handle nil values
- [X] Handle `NULLS FIRST` and `NULLS LAST`

Fixes #3720.

### DIFF
--- a/features/collections.feature
+++ b/features/collections.feature
@@ -122,6 +122,34 @@ Feature: Collections
     Then the _site directory should exist
     And I should see "Item count: 2" in "_site/index.html"
 
+  Scenario: Ordered Collections
+    Given I have an "index.html" page that contains "1. of {{ site.methods.size }}: {{ site.methods.first.output }}"
+    And I have fixture collections
+    And I have a "_config.yml" file with content:
+    """
+    collections:
+      methods:
+        output: true
+        order: data.title
+    """
+    When I run jekyll build
+    Then the _site directory should exist
+    And I should see "1. of 7: <p>Page without title.</p>" in "_site/index.html"
+
+  Scenario: Ordered Collections with NULLS specified
+    Given I have an "index.html" page that contains "1. of {{ site.methods.size }}: {{ site.methods.first.output }}"
+    And I have fixture collections
+    And I have a "_config.yml" file with content:
+    """
+    collections:
+      methods:
+        output: true
+        order: data.title DESC NULLS FIRST
+    """
+    When I run jekyll build
+    Then the _site directory should exist
+    And I should see "1. of 7: <p>Page without title.</p>" in "_site/index.html"
+
   Scenario: Sort by title
     Given I have an "index.html" page that contains "{% assign items = site.methods | sort: 'title' %}1. of {{ items.size }}: {{ items.first.output }}"
     And I have fixture collections
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -67,6 +67,7 @@ module Jekyll
   autoload :RelatedPosts,        'jekyll/related_posts'
   autoload :Renderer,            'jekyll/renderer'
   autoload :Site,                'jekyll/site'
+  autoload :Sorter,              'jekyll/sorter'
   autoload :StaticFile,          'jekyll/static_file'
   autoload :Stevenson,           'jekyll/stevenson'
   autoload :URL,                 'jekyll/url'
--- a/lib/jekyll/collection.rb
+++ b/lib/jekyll/collection.rb
@@ -46,7 +46,18 @@ module Jekyll
           files << StaticFile.new(site, site.source, relative_dir, File.basename(full_path), self)
         end
       end
-      docs.sort!
+      Jekyll::Sorter.order!(docs, self.order, {
+        allow_nil: true,
+        allowed_fields: Document::ATTRIBUTES_FOR_SORTING,
+        nestable_fields: Document::ATTRIBUTES_FOR_NESTED_SORTING
+      })
+    end
+
+    # Get the sort order from the metadata or use the default one
+    #
+    # Returns the configured sorter method or the default
+    def order
+      @metadata['order'] ||= 'path ASC'
     end
 
     # All the entries in this collection.
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -8,6 +8,12 @@ module Jekyll
 
     YAML_FRONT_MATTER_REGEXP = /\A(---\s*\n.*?\n?)^((---|\.\.\.)\s*$\n?)/m
 
+    # Attributes that can be used when sorting posts
+    ATTRIBUTES_FOR_SORTING = %w[path]
+
+    # Attributes that can be used with nesting when sorting posts
+    ATTRIBUTES_FOR_NESTED_SORTING = %w[data]
+
     # Create a new Document.
     #
     # site - the Jekyll::Site instance to which this Document belongs
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -253,6 +253,26 @@ module Jekyll
       end
     end
 
+    # Sort an array of objects according to one or more properties.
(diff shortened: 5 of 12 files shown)


envygeeks:
I think this is gonna have to wait until after the 3.0 release unless somebody on @jekyll/core wants to review it but we are close to the release window for 3.0 so this won't get enough exposed real world testing.

parkr:
Hey, @goodtouch! How are you doing? I'm just going through all the old issues and pull requests and deciding what to pursue and what to kindly thank-but-no-thank. In your case, I would like to see sorting be configurable. 

I like your implementation, but I think it has some security concerns. Specifically, building code and executing it could be very dangerous. We have [a `sort` method as part of the Liquid filters](https://github.com/jekyll/jekyll/blob/b5e459430bb510f283bdb3935b3a220b81819442/lib/jekyll/filters.rb#L228-L265) whose code does something similar to what you're doing here. Think it could be co-opted to this cause? I wonder if there's a way to incorporate `#reverse` into the sorter in a safe way. Would this suffice?

```yaml
collections:
  posts:
    output: true
    sort_by: date
    sort_order: reverse
```

Additionally, are you still interested in working on the code for this? I would like to ship this with v3.2 if you're up for it.

goodtouch:
Hi @parkr! I've been quite busy lately, sorry for the delay 😅 .

Yeah I've seen the `sort liquid filter` before working on this one, but if I remember well it doesn't support multiple properties order (ie. `date DESC, slug DESC`), or nested properties (ie. `time.start` as in #3539).

The first implementation I made wasn't using code generation, but as @envygeeks was concerned about performance, I tested some different ways to do it, benchmarked them and went with the parsing + generated `class_eval` one since it was much faster (you can see #3720 for the discussion and [goodtouch/sorter-bench](https://github.com/goodtouch/sorter-bench) for the benchmarks).

I added some safety guards ([validation Regexp](https://github.com/jekyll/jekyll/pull/3735/commits/271c76470eb58023780dd21b68ef316fd6900896#diff-820b949601f83e81f2439f6745f05f57R26) and a customizable [limit](https://github.com/jekyll/jekyll/pull/3735/commits/271c76470eb58023780dd21b68ef316fd6900896#diff-820b949601f83e81f2439f6745f05f57R17) of generated sort methods) to prevent evaluation of arbitrary code and high memory/CPU consumption.

I can add more safety checks or fallback to any non code-generating version if you prefer, it's really up to you (and maybe the average number of items to sort in collections hosted on GitHub Pages).
Using a method to access the hash like [this one](https://github.com/goodtouch/sorter-bench/blob/master/sorter.rb#L182-L193) works, is safe, but is also much slower.

Whatever configuration format you prefer is also fine by me as long as it supports multiple and nested properties sorting (all of them are safe and simple to parse, it's really the `order` method that matters). Using `sort_by` and `sort_order` or `order`, `asc`, `desc`, `reverse` or `natural` doesn't make much of a difference to me and should be quite easy to change once decided (naming is always the hardest part 😁  ).

jhabdas:
For those arriving here looking to sort a collection by date here's a breadcrumb: https://github.com/jekyll/jekyll/issues/4687

PR #3739, opened 2015-05-26
website/docs/configuration: Fix missing category in sample

Thannoy:
Text after the sample extract says this sample gives the default value "project" for the key "category", but the sample extract did not contain this.

### DIFF
--- a/site/_docs/configuration.md
+++ b/site/_docs/configuration.md
@@ -392,6 +392,7 @@ defaults:
     values:
       layout: "project" # overrides previous default layout
       author: "Mr. Hyde"
+      category: "project"
 {% endhighlight %}
 
 With these defaults, all posts would use the `my-site` layout. Any html files that exist in the `projects/` folder will use the `project` layout, if it exists. Those files will also have the `page.author` [liquid variable](../variables/) set to `Mr. Hyde` as well as have the category for the page set to `project`.


kleinfreund:
It implicitly adds `project` to the `category`, because the files lives in the directory `/project`. It's not necessary to expliclitly say it.

That said, I think it doesn't hurt adding it here, since it makes things more obvious.

Thannoy:
Isn't the base directory `/projects`? (taken from the scope/path)

kleinfreund:
Yes, and that's why the `project` category is automatically applied. I have a directory named `en` which automatically applies the categories `en` to the pages and posts inside of it.

parkr:
Pages aren't automatically assigned categories... I'm confused by what you said @kleinfreund.

Thannoy:
@kleinfreund : There is the tailing **s** in the path prefix `/projects` which is not in the category name `project`.
@parkr : Oh, maybe. This is absolutely out of my current skiils :)

kleinfreund:
> Pages aren't automatically assigned categories...

@parkr Is that really true? Why is this behavior so inconsistent then?

envygeeks:
My interest just got peaked, how is it inconsistent?  Much more elaboration is needed IMO.

kleinfreund:
Well, the behavior is inconsistent in that posts get an automatic category when residing in `dir` but pages do not?

envygeeks:
That's not inconsistent, they are two different document types with two different rule sets and that behavior is standard for any system when you imply the defacto standard: pages don't have categories and posts do.

envygeeks:
Let me also clarify that if you are adding categories you are the one who is creating the inconsistency.

parkr:
Thanks for your pull request, but this isn't needed at this time. Thank you!

PR #3743, opened 2015-05-28
Docs now reflect correct redcarpet version

mxstbr:
Fixed the documentation for redcarpet, as Jekyll updated to v3.2.x many more extensions are supported.


envygeeks:
:heart:

PR #3758, opened 2015-06-04
Regex escape keep_files prior to building regex

benbalter:
This avoids some strange edge cases that may cause regex errors when keep files have strange characters in them. 

/cc https://github.com/jekyll/jekyll-feed/pull/54

### DIFF
--- a/lib/jekyll/cleaner.rb
+++ b/lib/jekyll/cleaner.rb
@@ -98,7 +98,9 @@ module Jekyll
     #
     # Returns the regular expression
     def keep_file_regex
-      Regexp.union(site.keep_files)
+      # Escape user input to avoid regex errors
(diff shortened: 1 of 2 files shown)


envygeeks:
I'm :+1: on this even without taking into considering my feeback!

parkr:
cool!

benbalter:
Turns out unlike [what we're using on 2.5](https://github.com/jekyll/jekyll/blob/stable-2-5/lib/jekyll/cleaner.rb#L88-L92) (where I experienced the issue), `Regexp.union` auto escapes when passed strings.

```
[1] pry(main)> array = ["test", "test$"]
=> ["test", "test$"]
[2] pry(main)> Regexp.union(array)
=> /test|test\$/
```

envygeeks:
There is a 2.5-stable so why not port this into that branch?

PR #3769, opened 2015-06-09
Update install docs (Core dependencies, Windows reqs, etc)

chrisfinazzo:


envygeeks:
:+1:

envygeeks:
Though I iterate again, since nodejs isn't required anymore maybe that should be updated in parallel.

chrisfinazzo:
OK I'll update that.

As you indicated in the other thread, the distinction needs to be made that many of the dependencies that Jekyll uses today are "opt-in" for 3.0.

Over time, those changes can land here.

PR #3777, opened 2015-06-10
Rubocop

fw42:
@parkr, is this something that you think would be useful for Jekyll? Do you like Rubocop?

This PR does:
* Basically copy the Rubocop config from Liquid
* Auto-correct violations of that config (mostly changes related to indentation, spacing, hash syntax, ...)
* Auto-generate a TODO config for everything that's not covered yet

This PR doesn't yet make Rubocop part of CI runs, but I think we should add that too (follow-up PR).


envygeeks:
This wasn't aimed at me but these changes are counterproductive and cleanly highlight Ruby's syntax issues with the hash rockets, I'm :-1: without changes to the way Rubocop behaves.

1. Lots of people will clash on hash rockets, namely those who program in multiple languages on a daily basis where syntax is more consistent and features are more clearly defined.  I am one of those, I always use a hash rocket for hashes so that it's clear when I use a keyword argument in a method even if Ruby itself internally does not technically (at the end) make that distinction itself when it comes to the syntax (because you can flip flop on keyword arguments the way you do on hashes.)

2. The change from `kind_of?` to `is_a?` is unacceptable because there is no difference between the two both performance and behavior wise.  This is an enforcement of an opinion that doesn't matter when it comes down to technical implementation.

3. Again, the switch from %w[] to %w() is another example of an enforced opinion that does not matter when it comes to technical implementation.  This is not `%W` vs `%w` this is delimiter vs delimiter an enforced change that again, brings no benefit.

I could keep going if I read more of the changes but I have work to do.

fw42:
The idea of Rubocop (or this branch) is not to enforce this specific style/config, but to enforce SOMETHING. I don't want to bikeshed about the changes in this PR, that's all totally up for debate, but I think having some guideline and consistency (whatever the rules are going to be) is useful. You disagree?

envygeeks:
I do not disagree.  I just think some of the changes don't matter :100:

parkr:
I like the idea of conforming to some spec, but I would want to find a slightly more archaic spec (for example, we generally prefer hash rockets even with symbol keys).

fw42:
I'm gonna close this for now

parkr:
After 3.0, we can consider getting something like this back in. Thanks for proposing it, @fw42!

envygeeks:
I'm coming back to this, CodeClimate now supports Rubocop and it's implicit rather than explicit, it's not in anybodies way, we can configure it and tailor it to our needs and nobody needs to worry about it or care about it.  I'll work with @jekyll/core and setting that up  since we already use CodeClimate for other stuff.

parkr:
@envygeeks Noted.

Issue #2684, opened 2014-08-03
Make _includes directory configurable


minad:
All the other directories are configurable

fniephaus:
@parkr I don't really understand why adding more configuration settings is a problem for beginners. They can still start with a very basic `_config.yml`.
On the other hand, things like this limit power users that like to be in full control over their site's structure.

Maybe Jekyll needs two documentations, one for beginners and one for advanced users. As an advanced user, I wouldn't even mind if I had to read [rubydoc](http://www.rubydoc.info/github/mojombo/jekyll/Jekyll) for finding out more about advanced/hidden features.

parkr:
My comment was from last August. I think it'd be fine to add, as long as it's rock-solid and has no security problems.

liquidvisual:
This would be immensely useful. +1

PR #3782, opened 2015-06-11
Make includes dir configurable and rename dirs for consistency

fniephaus:
This addresses #2684 and makes the `_includes` directory configurable.


lib/jekyll/configuration.rb
```diff
@@ -11,7 +11,8 @@ class Configuration < Hash
       'destination'   => File.join(Dir.pwd, '_site'),
       'plugins'       => '_plugins',
       'layouts'       => '_layouts',
-      'data_source'   =>  '_data',
+      'data_source'   => '_data',
+      'includes_dir'  => '_includes',
```
parkr: Perhaps just `includes` to match `plugins` and `layouts`?

fniephaus: I just wanted to avoid confusions, because there already is an `include`.
What would you prefer?

parkr: Ah, yes, that might be confusing. For consistency, it would be nice to have `plugins_dir`, `layouts_dir`, and so forth as well. But We'd need the backwards-compatibility layer (in Jekyll::Configuration so should be easy). Want to do that? Would be great to have :)

fniephaus: Sure, I could do that. So you want all directory configuration options to have a `_dir` suffix and a backwards-compatibility layer that maps old names to new names?

parkr: Yeppers! Examples of backwards-compatibility are in `Jekyll::Configuration`. Just add whatever is needed to that method. :)

fniephaus: No worries. I'll update this pull request next weekend if that's ok!

parkr: Sure. We're going to release 3.0 soonish, but this can be 3.1.

fniephaus: Ok, so maybe I can finish this on Thursday if that helps

fniephaus: @parkr turns out I'd also need to update `jekyll-sass-converter` if we rename `source` to `source_dir`. Is that still something you would want?

parkr: Keep source and dest/destination as-is.

fniephaus:
@parkr this should be it. let me know if you want me to change anything else. :smiley:

envygeeks:
@fniephaus can you please squash those commits. Thanks.

fniephaus:
@envygeeks Done!

PR #3785, opened 2015-06-14
Created ubuntu Guide, linked in Install notes

Lewiscowles1986:


envygeeks:
tl;dr:

* Use `ruby2.0` not `ruby`
* Tell them they don't need nodejs for a default jekyll install on 3.0.0+ it's optional.
* Use `build-essential` not `make` and `gcc` directly.  Add `libxml2-dev libxslt1-dev libyaml-dev`.
* Clarify the language of your introductions.

---

I don't actually know how often @parkr releases the documentation since I don't necessarily pay attention to that but at this point I'm going to assume that docs are on hold until the 3.0.0 release because a lot of our docs directly target that version so based on that:

* Don't target the `ruby` package, it points to ruby1.9... That's the moving target package of Ubuntu. We do not support ruby1.9 anymore.  You should also directly target a specific ruby package for that reason (ruby2.0 Ubuntu 14.04 - 14.10, ruby2.1 Ubuntu 15.04 -- heck, it could even stay 2.0 for 15.04 I don't care, just target a modern Ruby we support directly.)
* Attach a Jekyll version to your recommendation on NodeJS because 3.0.0 does not require a JavaScript runtime.  I would also mention that they can vendor (Gemfile) therubyracer or therubyrhino if they don't want Nodejs -- this is explicitly what we do on our server installs and what we do on the Docker images to keep them small-ish.  You don't need to mention the gem vesions of JavaScript runtimes but they are nice, the only thing that matters here is mention that on 3.0.0 they probably won't need NodeJS unless they are using Coffeescript.
* Please have them install build-essential not `make` and `gcc` directly and add libxml2-dev libxstl1-dev and libyaml-dev because chances are one day they'll try something (like gem update,) lets make it easy on them from the get go by making sure they have *everything* they need for most Ruby gem installs, (g++, gcc, zlib1g-dev, libxslt1-dev, libxml2-dev, libyaml-dev.  A lot of gems also use C++ too, which means g++ is very important for them to have.

Other notes:

* Don't target LTS releases, mention them sure, but target Ubuntu/Debian in general because they really do share the same guide for this type of thing.  Every step you gave (with the changes I asked for) will work on both current Debian and Ubuntu 14.04+
* "sudo-capable user account" is not necessary, `gem --user-install` and `bundle` exists to solve this problem.  I am for mentioning these methods as they teach people how to Ruby the way people currently recommend doing Ruby but I'll digress on this to the opinion of the rest of @jekyll/core.

Lewiscowles1986:
Hi :smile: , 

Thanks for the feedback. Just as an FYI, this guide is tested for the current release, and works, which is why I posted it... I accept your critique, I will do my best to play around and verify in my spare time, and agree that `build-essential` is a nicer package than two isolated dependencies. I am very new to Jekyll so please bear with me.

Personally I'd hope for more linux install articles as a windows guide is posted, and I'd like to have some easy install steps to help people who want to use OSS stacks (hence the guide).

Can I also ask that if you are using docker, you try to link an article on that, and your setup, so that people wanting to production deploy, and not hobby dev / get into or use gh-pages can get into. Once I am better versed I'd :heart: to author more articles that help people; and I accept your feedback as someone who uses this professionally will be invaluable in steering me towards accessible and long-term stable articles.

:+1:

envygeeks:
We provide OSS Docker images (https://github.com/docker/docker-jekyll) under jekyll/jekyll on Docker hub thought you are right that I should probably build a Docker install guide for people who wish to use Docker without having to learn to Ruby, because we keep that image updated with the latest Ruby (as in 2.2) and latest version's of Jekyll with everything you need.  For noobs this is perfect because Docker is as easy as `docker run jekyll/jekyll:beta` and it will do the rest.  I'll see what I can do sometime this week.

envygeeks:
Sorry the right repo is: https://github.com/jekyll/docker-jekyll

Lewiscowles1986:
Right, I have adjusted the guide to include instructions on installing Jekyll stable, or Jekyll 3 beta, and have made it clear that for 3.x therubyracer can be used as a drop-in replacement for node.

Lewiscowles1986:
Mainly revisiting because I went to run Jekyll 3 on Ubuntu 14.04LTS, and as @envygeeks mentioned originally, Jekyll 3 would not run from the original guide

envygeeks:
I'm :shipit: for this now, we can revisit it in the next LTS and remove brightbox but for now that is the best version of Ruby that for LTS that doesn't bring in all the crap that pre-15.04 fixed from Debian unstable, but that's a couple years away so no biggie or even a worry there, maybe we could sideline that sometime but it's not something that should stall this.

parkr:
I think our ubuntu guide should be:

1. Install Docker
2. Pull down jekyll/stable
3. Run jekyll/stable for your site

I mostly don't want to have to maintain a guide we won't use and therefore won't have an incentive to keep up-to-date. Thoughts?

envygeeks:
I kind of agree with that,  it seems lots of people use our Docker images and that might be the route we start to recommend to people to avoid issues (much like other projects are starting to do,) but at the same sense we really don't need to maintain an Ubuntu guide past 15.04 since they've normalized Ruby (finally, it took a long time but it finally got done!) and on Debian unstable so I don't think this will require much work at all tbh if we target it generally.

parkr:
@envygeeks Ok. As long as it's what we can say the core team recommends, I'm cool with it.

parkr:
Thanks for this guide, @Lewiscowles1986. I think I'm going to point folks to our Docker containers when hoping to run Jekyll on Linux. In any event, this PR will remain here for folks looking in the future.

Thanks again!

Lewiscowles1986:
:+1: ha I forgot all about it; I am still using Jekyll right now, it's not massively hard anyway. 

The original idea was to give back to OS in small ways for as many cool projects as possible, with easy to achieve things, low hanging fruit to wean me into the process, learning every day.

PR #3787, opened 2015-06-14
Change non-existent hl_linenos to hl_lines to allow passthrough in safe mode

### LINKED ISSUE #3776: Pygments' hl_lines option is sanitized incorrectly
#2642 introduced correct support for Pygments `hl_lines` in the first commit, and then immediately [broke it in the next one](https://github.com/jekyll/jekyll/commit/3cb2e74b5c93e837d08182d735ebb8a5fd793ced#diff-7dbeb1c0a6ca43d2d580a95ed86deb26R66). /cc @parkr

Pygments has [`hl_lines`](http://pygments.org/docs/formatters/), but you whitelisted [`hl_linenos`](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/tags/highlight.rb#L67) which leads to **line highlighting not available in safe mode**.

```html
<style>.hll, .special { color: red !important; }</style>
{% highlight text linenos=table hl_linenos=2 %}
hl_linenos=2
line2
line3
line4
{% endhighlight %}

{% highlight text linenos=table hl_lines=2 %}
hl_lines=2
line2
line3
line4
{% endhighlight %}

{% highlight text linenos=table linenospecial=2 %}
linenospecial=2
line2
line3
line4
{% endhighlight %}

{% highlight text linenos=table hl_linenos=2 hl_lines=3 linenospecial=4 %}
hl_linenos=2 hl_lines=3 linenospecial=4
line2
line3
line4
{% endhighlight %}
```

image

So `hl_lines` works only in non-safe mode. There was probably never a `hl_linenos` option, you invented that I guess. `linenospecial` is useless because it somehow pollutes the previous highlight as well and can only highlight one line.

TWiStErRob:
Fixes #3776 by changing to the correct name for whitelisting.

### DIFF
--- a/lib/jekyll/tags/highlight.rb
+++ b/lib/jekyll/tags/highlight.rb
@@ -64,7 +64,7 @@ eos
         if is_safe
           Hash[[
             [:startinline, opts.fetch(:startinline, nil)],
-            [:hl_linenos,  opts.fetch(:hl_linenos, nil)],
+            [:hl_lines,    opts.fetch(:hl_lines, nil)],
             [:linenos,     opts.fetch(:linenos, nil)],
             [:encoding,    opts.fetch(:encoding, 'utf-8')],
             [:cssclass,    opts.fetch(:cssclass, nil)]
--- a/test/test_tags.rb
+++ b/test/test_tags.rb
@@ -114,9 +114,9 @@ CONTENT
       assert_equal true, sanitized[:linenos]
     end
 
-    should "allow hl_linenos" do
-      sanitized = @tag.sanitized_opts({:hl_linenos => %w[1 2 3 4]}, true)
-      assert_equal %w[1 2 3 4], sanitized[:hl_linenos]
+    should "allow hl_lines" do
+      sanitized = @tag.sanitized_opts({:hl_lines => %w[1 2 3 4]}, true)
+      assert_equal %w[1 2 3 4], sanitized[:hl_lines]
     end
 
     should "allow cssclass" do


envygeeks:
Did you manually test this works? Also it seems you broke tests, can you fix those please? :+1:

TWiStErRob:
I manually modified my installed version of jekyll in the Ruby gems directory and tested it on my site if it goes through. I don't want to set up a whole environment for jekyll dev, it was to do it as a jekyll user. It's probably the same fix in the test as well, just changing the constant. I'll do it tomorrow and see what CI says.

envygeeks:
Without the tests working merging this is an automatic :-1:, we cannot accept anything you change if you break tests and don't fix those tests that defeats the purpose of testing.  I understand it's hard to modify software as a non-programmer but that does not change that we need software to have quality.

TWiStErRob:
Yep, I agree with testing. I'm a dev too, just in the strongly typed world. I'm just lazy to set up yet another development environment for changing a typo.

envygeeks:
For Ruby it's easy to isolate all that mess.  When you clone our repo do `bundle install --path vendor/bundle` so you can isolate all the dependencies and then `script/test` will do the rest for you.  When you are done remove vendor/bundle and you'll have removed everything we use in dev.

parkr:
:+1: to what @envygeeks suggested; it's very easy to develop Jekyll.

Even if you don't want to set it up locally, you can use Travis's output to fix the tests.

TWiStErRob:
> `bundle install --path vendor/bundle`

Thanks, I like that better than installing gems into my ruby I'll never use. Just like `npm`'s `node_modules`.

> and then `script/test`

Exactly my point. Executing that would require me to install cygwin or a virtual machine. Luckily I was able to decypher that I need `ruby -S bundle exec rake test`, which obviously thrown an `SIGSEGV` (https://bugs.ruby-lang.org/issues/10978), why would it work?!..., then I tried the other option `ruby -S bundle exec ruby -Itest` which didn't use any processor for minutes, so I closed it. To get `rake` work I would have to compile ruby myself, because my 2.2.2 is 500 revs behind... I hope you see what I mean now by not wanting to set up the env.

> you can use Travis's output to fix the tests.

Yep, that's the only way now; that was my original idea when I said "see what CI says.". @envygeeks' description got my hopes up again that it's really "that simple".

TWiStErRob:
Any idea for this failure? it looks unrelated.

envygeeks:
@TWiStErRob You could use boot2docker on Windows to solve all of that, I should get around to building a dev image for people but I've been putting it off since I have an active project.

That said I retriggered the test and I'll update you once I see if it's a Travis fluke or consistent.

envygeeks:
Seems to me it was a fluke, all is well. :+1:

parkr:
What was the error?

TWiStErRob:
JRuby test VS _config.yml location, like this: https://travis-ci.org/jekyll/jekyll/jobs/65915927

envygeeks:
It's a fluke that happens every once in a while on Travis CI only.  I've not been able to reproduce it at all on my local build (but I also don't use RVM... but it happens and I've been trying to get it reproduce ever since.)

envygeeks:
Is there anything hold this back or can we go ahead and merge it? I'm :+1:

parkr:
Do we have a test for this that actually runs it through Pygments? I'm fine on changing it but I'd like a test so we don't break it again.

envygeeks:
That's what stalled me, I tried to do this myself on a branch before @TWiStErRob did it and when I added in a test to actually test the output, it always remained unchanged.  I do agree though, we probably need a regression tester so it doesn't happen again.

TWiStErRob:
What's up with this? Is it waiting on me?

envygeeks:
@TWiStErRob if I recall right this is waiting for you to add a test to ensure it's ran properly.

TWiStErRob:
I updated the commit, it has the test-change too, ci tests were failing at first, because I didn't modify the tests. As for those regression tests, I'm not sure how to even start.

parkr:
> ci tests were failing at first, because I didn't modify the tests. As for those regression tests, I'm not sure how to even start.

Ensure you have rebased on the latest `master`, too.

TWiStErRob:
Rebased, but I still have Windows...

parkr:
Yep windows will cause problems. As long as Travis passes, I think we're good.

PR #3792, opened 2015-06-17
Add ssl support add serve tests

envygeeks:
/cc @parkr @jekyll/core

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -1,161 +1,198 @@
-# -*- encoding: utf-8 -*-
 module Jekyll
   module Commands
     class Serve < Command
-
       class << self
+        COMMAND_OPTIONS = {
+          "detach" => ["-B", "--detach", "Run the server in the background (detach)"],
+          "port" => ["-P", "--port [PORT]", "Port to listen on"],
+          "host" => ["host", "-H", "--host [HOST]", "Host to bind to"],
+          "skip_initial_build" => ["skip_initial_build", "--skip-initial-build",
+            "Skips the initial site build which occurs before the server " \
+              "is started."],
+
+          "ssl_key" => ["--ssl-key [KEY]", "X.509 (SSL) Private Key."],
+          "ssl_cert" => ["--ssl-cert [CERT]", "X.509 (SSL) certificate."],
+          "baseurl" => ["-b", "--baseurl [URL]", "Base URL"]
+        }
 
         def init_with_program(prog)
           prog.command(:serve) do |c|
-            c.syntax 'serve [options]'
-            c.description 'Serve your site locally'
+            c.description "Serve your site locally"
+            c.syntax "serve [options]"
             c.alias :server
             c.alias :s
 
             add_build_options(c)
+            COMMAND_OPTIONS.each do |k, v|
+              c.option k, *v
+            end
 
-            c.option 'detach', '-B', '--detach', 'Run the server in the background (detach)'
-            c.option 'port', '-P', '--port [PORT]', 'Port to listen on'
-            c.option 'host', '-H', '--host [HOST]', 'Host to bind to'
-            c.option 'baseurl', '-b', '--baseurl [URL]', 'Base URL'
-            c.option 'skip_initial_build', '--skip-initial-build', 'Skips the initial site build which occurs before the server is started.'
-            c.option 'open_url', '-o', '--open-url', 'Opens the local URL in your default browser'
+            c.action do |_, opts|
+              unless opts.key?("watch")
+                opts["watch"] = true
+              end
 
-            c.action do |args, options|
-              options["serving"] = true
+              opts["serving"] = true
               options["watch"] = true unless options.key?("watch")
-              Jekyll::Commands::Build.process(options)
-              Jekyll::Commands::Serve.process(options)
+              Build.process(opts)
+              Serve.process(opts)
             end
           end
         end
 
         # Boot up a WEBrick server which points to the compiled site's root.
+
         def process(options)
           options = configuration_from_options(options)
-          destination = options['destination']
+          destination = options["destination"]
           setup(destination)
 
-          s = WEBrick::HTTPServer.new(webrick_options(options))
-          s.unmount("")
-
-          s.mount(
-            options['baseurl'],
-            custom_file_handler,
-            destination,
-            file_handler_options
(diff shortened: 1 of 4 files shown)


lib/jekyll/commands/serve.rb
```diff
@@ -1,142 +1,192 @@
-# -*- encoding: utf-8 -*-
 module Jekyll
   module Commands
     class Serve < Command
-
       class << self
+        COMMAND_OPTIONS = {
+          "detact" => ["-B", "--detach", "Run the server in the background (detach)"],
```
parkr: s/detact/detach/

lib/jekyll/commands/serve.rb
```diff
@@ -1,142 +1,192 @@
...
 
             add_build_options(c)
+            COMMAND_OPTIONS.each do |k, v|
+              c.option k, *v
+            end
+
+            c.action do |a, o|
+              unless o.key?("watch")
+                o["watch"] = \
+                  true
```
parkr: Would you put the `true` on the line above please? I think it'll read a bit better.

lib/jekyll/commands/serve.rb
envygeeks: That's been fixed locally I should probably push up the new changes with the new handler.

lib/jekyll/commands/serve.rb
```diff
@@ -1,142 +1,192 @@
...
+            c.syntax "serve [options]"
             c.alias :server
             c.alias :s
 
             add_build_options(c)
+            COMMAND_OPTIONS.each do |k, v|
+              c.option k, *v
+            end
+
+            c.action do |a, o|
```
parkr: `a` and `o` are cryptic – would you mind changing them back to abbreviated words? I don't think `args` is used, so you can `_` that, and call `o` `opts`.

lib/jekyll/commands/serve.rb
```diff
@@ -1,142 +1,192 @@
...
-
-          s.mount(
-            options['baseurl'],
-            custom_file_handler,
-            destination,
-            file_handler_options
+          server = WEBrick::HTTPServer.new(
+            webrick_options(
+              options
+            )
```
parkr: Was the original really unreadable to you? I thought it was pretty straight-forward.

lib/jekyll/commands/serve.rb
```diff
@@ -1,142 +1,192 @@
...
           end
         end
 
+        private
         def webrick_options(config)
           opts = {
-            :BindAddress        => config['host'],
-            :DirectoryIndex     => %w(index.html index.htm index.cgi index.rhtml index.xml),
-            :DocumentRoot       => config['destination'],
+            :MimeTypes => mime_types,
```
parkr: I prefer all the `=>` rockets to be aligned, please.

lib/jekyll/commands/serve.rb
envygeeks: It was a little unreadable in that it took me a few seconds to get a grip on the file which is why some of the changes landed but I also split off a bit of it to fix that problem so maybe we should review the new changes I made because they are quite a bit different.

lib/jekyll/commands/serve.rb
```diff
@@ -1,142 +1,192 @@
...
           }
 
-          if config['verbose']
-            opts.merge!({
-              :Logger => WEBrick::Log.new($stdout, WEBrick::Log::DEBUG)
-            })
+          enable_ssl(opts, config)
+          enable_verbosity(
+            opts, config
+          )
```
parkr: I think this `enable_verbosity` call can be on one line.

parkr:
No substantive questions here, all style fixes. Thanks, Jordon!

lib/jekyll/commands/serve.rb
```diff
@@ -1,142 +1,192 @@
...
           end
         end
 
+        private
         def mime_types
           mime_types_file = File.expand_path('../mime.types', File.dirname(__FILE__))
-          WEBrick::HTTPUtils::load_mime_types(mime_types_file)
+          WEBrick::HTTPUtils::load_mime_types(
+            mime_types_file
+          )
```
parkr: I think this can be on one line.

lib/jekyll/commands/serve.rb
```diff
@@ -1,142 +1,192 @@
...
         end
 
+        private
         def start_callback(detached)
           unless detached
-            Proc.new { Jekyll.logger.info "Server running...", "press ctrl-c to stop." }
+            proc do
+              Jekyll.logger.info(
+                "Server running...", "press ctrl-c to stop."
+              )
```
parkr: I think this logger call can be on one line.

lib/jekyll/commands/serve.rb
```diff
@@ -1,142 +1,192 @@
...
+        def file_handler
           Class.new WEBrick::HTTPServlet::FileHandler do
             def search_file(req, res, basename)
               if file = super
-                file
-              else
-                super(req, res, "#{basename}.html")
+                then file
+                else super(
+                  req, res, "#{basename}.html"
```
parkr: I think this can be on one line.

lib/jekyll/commands/serve.rb
```diff
@@ -1,142 +1,192 @@
...
+            )
           else
-            opts.merge!({
-              :AccessLog => [],
-              :Logger => WEBrick::Log.new([], WEBrick::Log::WARN)
-            })
+            opts[:AccessLog] = []
+            opts[:Logger] = WEBrick::Log.new(
+              [], WEBrick::Log::WARN
+            )
```
parkr: Oh, is this faster than creating the hash and merging it in?

alfredxing:
:+1:

SSL support is good in general but I really don't want to act like we are promoting running this as a production server :wink:

parkr:
How's this PR doing? Why do we want SSL for local testing, again?

envygeeks:
The SSL was for somebody who requested SSL because they were working with API's that required SSL and stuff like that, it's a good idea to allow it just in-case.  I'll work on getting it cleaned up to get merged tomorrow morning since I have time.

envygeeks:
I'm gonna drop this pull and merge two pulls I have open with the same file into a single pull. Plus this rebase messed everything up and I'm in no mood to fight with Git, I've a lot of Jekyll stuff planned for today.

PR #3795, opened 2015-06-17
Update dependencies.

envygeeks:


Gemfile
```diff
@@ -20,13 +20,13 @@ if RUBY_PLATFORM =~ /cygwin/ || RUBY_VERSION.start_with?("2.2")
 end
 
 gem 'rake', '~> 10.1'
-gem 'rdoc', '~> 3.11'
+gem 'rdoc', '~> 4.2'
```
parkr: I don't think we use `rdoc` anymore. Is this still necessary?

parkr:
Pulling down to check everything. `mime-types` isn't tested so bump could cause issues.

parkr:
Somehow, the update from cucumber 1.3 to 2.x didn't break anything.

envygeeks:
Yeah that confused me too.  I thought for sure Cucumber would break but it didn't.

PR #3798, opened 2015-06-18
Jekyll/add support for webrick file precedence

boa0332:


envygeeks:
What is this? This looks like my commit without any changes and a hard merge.

PR #3818, opened 2015-06-24
Update windows.md with Ruby version info

mgiuffrida:
Jekyll dependency hitimes does not support Ruby 2.2 on Windows yet


parkr:
:shipit:

PR #3820, opened 2015-06-25
Add missing flag to disable the watcher

chrisfinazzo:


envygeeks:
I personally like the way that CLI's (well some do some dont) where it's `--[no-]watch` but :shipit:

chrisfinazzo:
To each their own, I guess...

See #3607.

site/_docs/configuration.md
```diff
@@ -180,7 +180,7 @@ class="flag">flags</code> (specified on the command-line) that control them.
         <p class="description">Enable auto-regeneration of the site when files are modified.</p>
       </td>
       <td class="align-center">
-        <p><code class="flag">-w, --watch</code></p>
+        <p><code class="flag">-w, --watch, --no-watch</code></p>
```
parkr: I agree with @envygeeks. Would love to see this be `--[no-]watch` or similar.

chrisfinazzo: Sure, will update.

Out of curiosity, is there a short equivalent of the `--no-watch` flag? Seems odd to only have `-w`.

parkr: > Out of curiosity, is there a short equivalent of the --no-watch flag? Seems odd to only have -w.

Nope.

PR #3821, opened 2015-06-25
We don't need sudo here

chrisfinazzo:
Tl;dr - People probably shouldn't be using `sudo` to install gems or managing them directly. Wouldn't it be safer to encourage a Ruby manager (rbenv, RVM) + Bundler instead?

### DIFF
--- a/site/_docs/configuration.md
+++ b/site/_docs/configuration.md
@@ -180,7 +180,7 @@ class="flag">flags</code> (specified on the command-line) that control them.
         <p class="description">Enable auto-regeneration of the site when files are modified.</p>
       </td>
       <td class="align-center">
-        <p><code class="flag">-w, --watch</code></p>
+        <p><code class="flag">-w, --watch, --no-watch</code></p>
       </td>
     </tr>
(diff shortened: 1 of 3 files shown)


envygeeks:
This does not encourage anything, it simply enforces the idea that sudo is bad, and it's not. :-1:

chrisfinazzo:
Not categorically bad, but used only as appropriate. I don't mean to say that it shouldn't be supported, but there's a reason why contributors are instructed to install Bundler. The reasoning for this does become less important w/ Jekyll 3, as dependencies are opt-in, but in many other places, direct manipulation of gems is discouraged (Docker images, Windows installs, etc.)

I will admit that I was cautiously optimistic about Bundler when I first tried it, but it simplifies things to such an extent that I can't imagine not using it. 

I don't think this should be an outright requirement, but more along the lines of "Trust us, things will go smoothly and you'll stay sane if you follow these instructions"

@jekyll/core

chrisfinazzo:
At least on OS X, this _can_ work. I can't comment on Linux. I'll update those...

PR #3823, opened 2015-06-25
Adapt StaticFile for collections, config defaults

mbland:
This enables files such as images and PDFs to show up in the same relative output directory as other HTML and Markdown documents in the same collection.

It also enables static files to be hidden using defaults from _config.yml in the same way that other documents in the same collection and directories may be hidden using `published: false`.

Extracted from 18F/hub#332.

cc: @afeld @gboone @jeremiak @dhcole

### DIFF
--- a/lib/jekyll/static_file.rb
+++ b/lib/jekyll/static_file.rb
@@ -37,7 +37,7 @@ module Jekyll
 
     def destination_rel_dir
       if @collection
-        @dir.gsub(/\A_/, '')
+        File.dirname(url)
       else
         @dir
       end
@@ -61,9 +61,10 @@ module Jekyll
 
     # Whether to write the file to the filesystem
     #
-    # Returns true.
+    # Returns true unless the defaults for the destination path from
+    # _config.yml contain `published: false`.
     def write?
-      true
+      defaults.fetch('published', true)
     end
 
     # Write the static file to the destination directory (if modified).
@@ -99,6 +100,42 @@ module Jekyll
         "modified_time" => modified_time,
         "path"          => File.join("", relative_path)
       }
+    end
+
+    def placeholders
+      {
+        collection: @collection.label,
+        path: relative_path[
+          @collection.relative_directory.size..relative_path.size],
+        output_ext: '',
+        name: '',
+        title: '',
+      }
+    end
+
+    # Applies a similar URL-building technique as Jekyll::Document that takes
+    # the collection's URL template into account. The default URL template can
+    # be overriden in the collection's configuration in _config.yml.
+    def url
+      @url ||= if @collection.nil?
+        relative_path
+      else
+        ::Jekyll::URL.new({
+          template:  @collection.url_template,
+          placeholders: placeholders,
+        })
+      end.to_s.gsub /\/$/, ''
+    end
+
+    # Returns the type of the collection if present, nil otherwise.
+    def type
+      @type ||= @collection.nil? ? nil : @collection.label.to_sym
+    end
+
+    # Returns the front matter defaults defined for the file's URL and/or type
+    # as defined in _config.yml.
+    def defaults
+      @defaults ||= @site.frontmatter_defaults.all url, type
     end
   end
 end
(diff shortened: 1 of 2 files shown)


envygeeks:
I like it :+1:

mbland:
@envygeeks Thanks for the rapid feedback. All comments incorporated!

envygeeks:
Do you mind rebasing the two commits into a single commit and forcing it up again so that other reviewers can see both the changes and everything in on swift review? Thanks!

mbland:
No problem! Done.

dhcole:
Nice work @mbland. Would be great to see this land in Jekyll core.

lib/jekyll/static_file.rb
```diff
@@ -100,5 +101,40 @@ def to_liquid
...
+          placeholders: {
+            collection: @collection.label,
+            path: self.relative_path[
+              @collection.relative_directory.size..self.relative_path.size],
+            output_ext: '',
+            name: '',
+            title: '',
+          },
+        }).to_s
+        @url = @url.gsub /\/$/, ''
```
parkr: this is effectively `@url.gsub!` – why not use the bang method here?

lib/jekyll/static_file.rb
```diff
@@ -100,5 +101,40 @@ def to_liquid
         "path"          => File.join("", relative_path)
       }
     end
+
+    # Applies a similar URL-building technique as Jekyll::Document that takes
+    # the collection's URL template into account. The default URL template can
+    # be overriden in the collection's configuration in _config.yml.
+    def url
+      return @url unless @url.nil?
```
parkr: haven't seen this in a while – why not something like

```ruby
@url ||= if @collection.nil?
  relative_path
else
  Jekyll::URL.new({
    # etc
  })
end.to_s.gsub!(/\/$/, '')
```

lib/jekyll/static_file.rb
```diff
@@ -100,5 +101,40 @@ def to_liquid
...
+      if @collection.nil?
+        @url = self.relative_path
+      else
+        @url = ::Jekyll::URL.new({
+          template:  @collection.url_template,
+          placeholders: {
```
parkr: would you please create a separate method which creates the `placeholders` hash?

parkr:
Looking pretty good! Thanks, @mbland.

Looks like you're coming from a Python background, based on all those `self.`'s. In Ruby, we only use them when setting an attribute on a class instance, such as `self.url = "some new url"`. When we access the attribute, however, we do not include the `self.` If you would please remove each instance of `self.` whenever accessing instance variables or methods, I would be most appreciative!

lib/jekyll/static_file.rb
parkr: can we leverage any of `#to_liquid`?

lib/jekyll/static_file.rb
mbland: Done.

I must've been thinking about how `gsub!` will return nil if nothing was changed, and how that would not work as a return value. But it's not a return value here.

lib/jekyll/static_file.rb
mbland: `placeholders` method: Done.

Regarding `#to_liquid`, I don't see any overlap between the values.

lib/jekyll/static_file.rb
mbland: Wow, had no idea that was possible... Done, though I did have to remove the `!` from `gsub!`, as that will return `nil` if no substitutions are performed.

mbland:
Thanks, @parkr! Took out all the `self.`s, though it's actually @afeld's fault I started using them in Ruby. :-P

envygeeks:
Can you please rebase all this into a single commit so that it can be re-reviewed in one swift shot.

Thanks.

mbland:
Yep! Done.

parkr:
Cool! :shipit:

mbland:
Hey gang, not to nag, but is there further review pending, or is this ready to merge?

envygeeks:
This is ready! Sorry I didn't notice @parkr said to :shipit:

mbland:
_Sweeet,_ thanks, @envygeeks!

PR #3828, opened 2015-06-30
Add build options (fixes #3744)

### LINKED ISSUE #3744: `jekyll clean` doesn't understand `--config`
I'm using 3.0.0.pre.beta5

`jekyll clean -s src -d build` doesn't work either.

---

`config.yml` is at the root of the source tree; the commands are executed from the root as well.

superlukas:
See #3744


parkr:
:shipit:

PR #3830, opened 2015-07-01
Statically serve php files with 'jekyll serve'

shuber2:
As described in pull request #3106, I would like to serve php files with a proper mime type, that is, text/plain. Github did not use to se such a mime type, so request #3106 was denied, but that changed in the meantime, see the comment section there. Hence, here comes my request again, now conforming with github behavior.

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -77,7 +77,7 @@ module Jekyll
         def webrick_options(config)
           opts = {
             :BindAddress        => config['host'],
-            :DirectoryIndex     => %w(index.html index.htm index.cgi index.rhtml index.xml),
+            :DirectoryIndex     => %w(index.html index.htm index.cgi index.rhtml index.xml index.php),
             :DocumentRoot       => config['destination'],
             :DoNotReverseLookup => true,
             :MimeTypes          => mime_types,
--- a/lib/jekyll/mime.types
+++ b/lib/jekyll/mime.types
@@ -1,6 +1,7 @@
 # These are the same MIME types that GitHub Pages uses as of 26 January 2014
 
 text/html;charset=utf-8               html htm shtml
+text/plain;charset=utf-8              php
 text/css                              css
 text/xml;charset=utf-8                xml rss xsl xsd
 image/gif                             gif


envygeeks:
Your first commit contradicts your second IMO.

shuber2:
@envygeeks I do not think so, unless you provide more details.

envygeeks:
It should be obvious that most index pages are *not* plain text.

shuber2:
@envygeeks Your comment does not help. Commit 8a5727862 adds the mime type for php files to match github's behavior. The second commit is independent from the exact choice for the mime type of php files, and this choice was made by github anyhow. Have you followed the comments of pull request #3106?

envygeeks:
I have, the question is have you? And if you have, perhaps you should re-read it, but to put it simple: http://envygeeks.github.io v. http://envygeeks.github.io/index.php

envygeeks:
Like I said, the first contradicts the second, the first is fine, I think the second should go.

shuber2:
@envygeeks I adapted the commits too quickly and did not think it through. The issue is that the HTML in the php file wouldn't be rendered if the mimetype is text/plain. So yes, I agree, the second commit makes little sense with mimetype text/plain.

So what I would need is that github actually switches from text/plain to text/html, as mentioned in the original pull request.

Sorry.

parkr:
I think this is fine – I don't mind having a `text/plain` index page if it helps use Jekyll better. In this case, it's a very small thing to give and it'll match GHP.

envygeeks:
I don't think GHP supports it because the links above (which were just created) don't work the way you would expect, but based on the vote for it if I think about it, adding it isn't a big deal anyways because I don't think GHP supports CGI index pages either.

parkr:
@envygeeks You're right that `index.php` isn't a directory index. http://envygeeks.github.io/index.php

> So what I would need is that github actually switches from text/plain to text/html, as mentioned in the original pull request.

@shuber2 Let's nix the `index.php` from the DirectoryIndex for now and revisit it if GitHub Pages adds the support for it. Thank you!

parkr:
I can't rationalize this at the moment. If you want to use dynamic pages, please use a different server.

PR #3837, opened 2015-07-06
Fixed an unclear code comment

jensechu:
There is a line that has an unclear code comment. I'm not entirely sure what the line is really supposed to imply but I made my best guess.

If that's not accurate please let me know and I can reword! :+1:


parkr:
:+1: Thanks!

Issue #3839, opened 2015-07-07
Redcarpet Vulnerability (CVE-2015-5147)


wjdp:
The following came through on Gemnasium on a project using Jekyll

![image](https://cloud.githubusercontent.com/assets/1690934/8536544/e37a8cb0-244a-11e5-8130-ab8f23145185.png)

envygeeks:
:shipit:

wjdp:
FYI just spotted only affects v3.3.0 - v3.3.1 (https://security-tracker.debian.org/tracker/CVE-2015-5147)

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least
three months.

The resources of the Jekyll Core team are limited, and so we are asking for your help.

If you can still reproduce this error on the `3.0-stable` branch or on `master`,
please reply with all of the information you have about it in order to keep the issue open.

Thank you for all your contributions.

envygeeks:
Recarpet is optional now days so the version is on the user.

PR #3840, opened 2015-07-07
Redcarpet 3.2.3 -> 3.3.2 because CVE-2015-5147

wjdp:
Re #3839, to see what tests do


parkr:
I think we're ok here. All relevant services have been notified.

PR #3841, opened 2015-07-07
Redcarpet v2 patch

wjdp:
Same as #3840 but for stable-2.5, updates redcarpet because of #3839 vulnerability.

I think the stable-2.5 branch is a commit behind the v2.5.3 tag so a couple of other bits are showing as changed below.


wjdp:
Just spotted only affects v3.3.0 - v3.3.1 (https://security-tracker.debian.org/tracker/CVE-2015-5147), as v2.5.x jekyll uses redcarpet v3.1 this PR is unnecessary

PR #3848, opened 2015-07-13
Detailed instructions for rsync deployment method

vitalyrepin:
Suggest to provide more extended documentation on rsync-approach. It also mentions rrsync wrapper script which restricts access for rsync to the server. Based on my blog post here: http://vrepin.org/vr/JekyllDeploy/

### DIFF
--- a/site/_docs/deployment-methods.md
+++ b/site/_docs/deployment-methods.md
@@ -89,11 +89,73 @@ Setup steps are fully documented
 Another way to deploy your Jekyll site is to use [Rake](https://github.com/jimweirich/rake), [HighLine](https://github.com/JEG2/highline), and
 [Net::SSH](https://github.com/net-ssh/net-ssh). A more complex example of deploying Jekyll with Rake that deals with multiple branches can be found in [Git Ready](https://github.com/gitready/gitready/blob/cdfbc4ec5321ff8d18c3ce936e9c749dbbc4f190/Rakefile).
 
+
+### scp
+
+Once you’ve generated the `_site` directory, you can easily scp it using a `tasks/deploy` shell script similar to [this deploy script here](https://github.com/henrik/henrik.nyh.se/blob/master/script/deploy). You’d obviously need to change the values to reflect your site’s details. There is even [a matching TextMate command](http://gist.github.com/214959) that will help you run this script from within Textmate.
+
 ### rsync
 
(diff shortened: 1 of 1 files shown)


envygeeks:
Why exactly did the old method need to be removed?

gynter:
The old method does not describe the process using `rsync`, but using `scp`.

vitalyrepin:
Rsync based method also suggests to limit write access only to the site directory (using rrsync wrapper) which is more secure than enabling certificate-based ssh access to the hosting account.

envygeeks:
Your method uses SSH all over the place so what does your last statement mean?  I personally think the old way could have been retagged instead of removed... your method is not necessarily better (in the way you do it) for the reasons you state but for the reasons you don't state that are obvious to the end-user.  I'm not against this, I just think the old way didn't need to be removed, people work in different ways with different tools that they have at the time.

vitalyrepin:
Changed my commit per your feedback - put back old method and retagget it to 'scp'.

My last statement was about usage rrsync wrapper and the restriction to certificate-based authorization (in .ssh/authorized_keys):

```
command="$HOME/bin/rrsync <folder>",no-agent-forwarding,no-port-forwarding,no-pty,no-user-rc,no-X11-forwarding ssh-rsa <cert>
```

In this configuration, if the user is authorized with certificate he/she is allowed to execute **only** rrsync command (restricted rsync).

envygeeks:
:shipit: /cc @jekyll/core

parkr:
Sure, feel free to merge and push out.

vitalyrepin:
Any update on this? I see "All checks have failed" - is it something caused by my changes? I suspect, this is something else.

parkr:
@vitalyrepin Would you mind rebasing on latest `master`?

vitalyrepin:
Rebased. Now checks are passed.

envygeeks:
:shipit:

chrisfinazzo:
@vitalyrepin See comments in #4048. I'm not familiar with this approach...can you prefix the install location as `usr/local`?

envygeeks:
He used the Linux approach, not the single-user Mac approach.  On Linux ~/bin is a common directory (even if a lot of us move it to `~/.local/bin`  acutally, bin is such a common directory that a lot of distros have a default .profile that checks if it exists so it can place it in the path.

vitalyrepin:
Just for the reference:

It is not "/bin" but it is "~/bin".

"~" is expanded by the shell to the home directory of the user. Works in OS X as well.

Instruction is written with assumption that the user does not have root access (hence, can not put anything to the system-wide directories, including /usr/local/bin). That's why "~" (home dir) is used.

And this assumption is very realistic - this part of instructions describes server-side setup. E.g., it can be hosting's shell. User is usually granted chrooted environment there.

chrisfinazzo:
Ah, I see. I've updated #4048 with the proper path. cc/ @vitalyrepin

Issue #3847, opened 2015-07-13
Working with drafts over midnight causes unexpected URL changes.


shinkondo:
Let's Say I created a draft, and opened it up in a browser at `/2015/7/12/{post-name}.html` on 7/12/2015.

If made change to the post on 7/13/2015 and regenerated the site, the updated post shows up in `/2015/7/13/{post-name}.html`.

This would be confusing because I can still reload the old post at `/2015/7/12/{post-name}.html`, which does not show any change I made.

I use Windows 7, Cygwin, Jekyll 3.0.0.pre.beta8, gulp, and browser-sync.

I think the following could be possible solutions:
- Give drafts future date that is long enough and do not require change (but may conflict with the `future: false` option ).
- Delete outdated drafts in `_site`.
- Put caution about this in [documentation](http://jekyllrb.com/docs/drafts/)

I would appreciate your advice/ suggestion/ comments.

kleinfreund:
> Delete outdated drafts in _site.

Can you build your site with `full_rebuild: true` in the _config.yml, please? This way, everytime you change something (or run `jekyll build`) the contents of the `_site` directory are wiped.

envygeeks:
I agree with @kleinfreund what happens when you do what he asked? Because outdated drafts *should* be deleted as they should end up in the outdated files list because they are no longer in theory being generated because technically drafts have no date, the date is the current date given on a whim.

shinkondo:
I cannot check in the exact situation until midnight, but I realized that any renamed file under `_site` is not deleted after build regardless of `full-rebuild` option.

And it is probably due to the bug of `keep_files`, combined with my blog folder name containing ".git" as a substring. Please see #3849 for detail.

This fix prevents me from reloading an outdated file when a post is renamed. It is good enough as a reminder for me.

Thank you for your help.

PR #3849, opened 2015-07-14
Cleaner: `keep_files` should only apply to the beginning of paths, not substrings with index > 0

shinkondo:
`keep_files` could prevent more files than intended from being removed. 

For example, if someone is working under a folder something like `username.github.io`, `keep_files:   [".git", ".svn"]` matches any files under `username.github.io/_sites`, leaving the destination folder uncleaned.

This would be confusing when a file is renamed while it is viewed on a browser. An user can be unaware of the fact that he is looking at outdated file (see #3847).

According to the [documentation](http://jekyllrb.com/docs/configuration/#global-configuration), the paths in `keep_files` should be "relative to the destination" rather than being keywords.

For these reasons, I made a change so that keep_files are interpreted as paths under the destination folder.


envygeeks:
What is the status of this?

parkr:
> For example, if someone is working under a folder something like username.github.io, keep_files: [".git", ".svn"] matches any files under username.github.io/_sites, leaving the destination folder uncleaned.

The example here doesn't apply to the feature. `keep_files` is a way of inhibiting the removal of files and folders inside of your compiled site directory (e.g. `_site`). In your example, it sounds like you think of `keep_files` as a way to keep your site's source from being cleaned up.

Am I misreading you here?

shinkondo:
First, thank you for your feedback.

I am afraid you are. I think of `keep_files` as a way to keep a compiled site directory, not source.
If you would take a look at #3847, you might get some ideas about my intent; I was confused by the fact that compiled outdated html was not removed.  

Let me know if you are still unsure about my post. I know I am not a good writer.

test/test_cleaner.rb
```diff
@@ -37,6 +37,40 @@ class TestCleaner < JekyllUnitTest
...
+      @site = fixture_site
+      @site.keep_files = ['.git']
+
+      @cleaner = Cleaner.new(@site)
+      @cleaner.cleanup!
+    end
+
+    teardown do
+      FileUtils.rm_rf(dest_dir('.git'))
+      FileUtils.rm_rf(dest_dir('.username.github.io'))
```
parkr: This should not include the preceding `.`, correct?

shinkondo: That is correct. Thank you for pointing it out.

parkr:
@shinkondo I see, thank you for your patience.

> my blog folder name containing ".git" as a substring

This seems to have explained it for me. You have `keep_files: ".git"` with a subdirectory of `username.github.io` and you find that this subdirectory, once created, is not cleaned?

I am 👍 on this PR. Just one comment above.

shinkondo:
I think what you described captures my issue, but please don't pull this request as of now.

It seems to me that the situation has changed because of bd2c337e5bd1a3532f42776f55f80f4e273cfeac. My code in pull-request would not work. Now regex generated from `keep_files` would match against relative paths under `_sites`, not absolute paths. My code will generate regex with absolute paths. I bet it would not match anything practically.

parkr:
@shinkondo You're welcome to merge `origin/master` into this branch and let Travis CI take a crack at it. bd2c337 doesn't fix your issue, though, does it?

shinkondo:
No, bd2c337e5bd1a3532f42776f55f80f4e273cfeac does not fix my issue.

As Travis CI showed, I was wrong to say that my code would not work. It seems that the code is using absolute paths again because of bd2c337e5bd1a3532f42776f55f80f4e273cfeac.

In any case, I can fix my mistake in the test. Is there anything else I should do? I can try to squash/rebase the git log if that is preferable.

parkr:
This looks just fine to me. Thank you!

@jekyllbot: merge +bug

PR #3852, opened 2015-07-16
Try to organize dependencies into dev and test groups.

envygeeks:


Gemfile
```diff
@@ -1,47 +1,55 @@
...
 gem 'jekyll-paginate', '~> 1.0'
-gem 'jekyll-gist', '~> 1.0'
 gem 'jekyll-coffeescript', '~> 1.0'
+gem 'jekyll-gist', '~> 1.0'
 
 platform :ruby, :mswin, :mingw do
-  gem 'pygments.rb', '~> 0.6.0'
   gem 'rdiscount', '~> 2.0'
-  gem 'classifier-reborn', '~> 2.0'
+  gem 'pygments.rb', '~> 0.6.0'
```
parkr: We trying to test pygments on Windows now?

envygeeks: I didn't even notice that but it looks like it was already there, do you want me to remove from Windows?

parkr: I think it's OK. If we can get a Windows box to test this on, then all the better, but if not, we can just remove it. I think most Windows users use Rouge, anyway.

PR #3862, opened 2015-07-20
Add casesensitive_categories option

### LINKED ISSUE #3285: Lower-case category in permalink breaks all inbound links
Hello.  I've had been using an older version of Jekyll and recently upgraded to the latest version as part of setting up a new box.  It seems that in the latest version, `:categories` in permalink resolves to the lowercase version of whatever categories are in my post instead of the category as specified (my posts use first-letter-upper-case).  That has effectively broken all inbound links to my site because URLs are case sensitive and inbound links are using URLs as I originally published them.  Needless to say, this really sucks.  Is there a way to get a configuration option that will respect the case of categories so that Jekyll is backward compatible?

x-way:
Add an option to leave the capitalization of the categories untouched by jekyll.

Fixes #3285.

### DIFF
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -45,6 +45,9 @@ module Jekyll
       'host'          => '127.0.0.1',
       'baseurl'       => '',
 
+      # Backwards-compatibility options
+      'casesensitive_categories' => false,
+
       # Output Configuration
       'permalink'     => 'date',
       'paginate_path' => '/page:num',
--- a/lib/jekyll/post.rb
+++ b/lib/jekyll/post.rb
@@ -230,7 +230,7 @@ module Jekyll
         :title       => slug,
         :i_day       => date.strftime("%-d"),
         :i_month     => date.strftime("%-m"),
-        :categories  => (categories || []).map { |c| c.to_s.downcase }.uniq.join('/'),
+        :categories  => (categories || []).map { |c| site.config['casesensitive_categories'] ? c.to_s : c.to_s.downcase }.uniq.join('/'),
         :short_month => date.strftime("%b"),
         :short_year  => date.strftime("%y"),
         :y_day       => date.strftime("%j"),


envygeeks:
If I have to scroll to view your code it's too big, move it to it's own method please. /cc @jekyll/core

parkr:
I was scrolling through twitter the other day and came across a quote that went something like:

> If someone proposes adding an option for something, it means something probably went wrong somewhere. Options are rarely the solution.

I like this idea in general. Adding further to the :christmas_tree: of options is usually a symptom of a larger problem (e.g. tags weren't well-thought out).

3.0 can break things. Let's decide what the best default is and if it's absolutely crucial to have a backwards-compatibility option like this one (for example, if most webservers are going to scream and cry if the case of a URL changes). Ultimately, I'd rather just make it what people expect and leave it optionless. Options can be good, but they shouldn't be the cure-all we use for every situation. Just look at where Mutt ended up with that mentality.

benbalter:
> 3.0 can break things. Let's decide what the best default is and if it's absolutely crucial to have a backwards-compatibility option like this one (for example, if most webservers are going to scream and cry if the case of a URL changes). Ultimately, I'd rather just make it what people expect and leave it optionless.

+ :100:

envygeeks:
It's not the web servers that will complain because they are by default case-sensitive and I believe it's only Apache that can be configured to break that defacto rule if I'm honest I'm gonna vote for anything that follows the (not official) golden rule that pretty much everybody follows: everything in the URL should be lowercase.  So when it comes to writing tags and categories I will always vote for normalizing them like slugs and enforcing a lowercase.

parkr:
> If I'm honest I'm gonna vote for anything that follows the (not official) golden rule that pretty much everybody follows: everything in the URL should be lowercase. So when it comes to writing tags and categories I will always vote for normalizing them like slugs and enforcing a lowercase.

Where do you see this golden rule? I wasn't aware of it. Is it default for all webservers?

envygeeks:
@parkr It's ***unofficial*** and I consider it golden because it's used all over the place and probably because of the way Google and Yahoo will treat inconsistent URL's with different content.

envygeeks:
Let me clarify my comments from earlier.  The complainer in this case would be Google and Yahoo,  if you have multiple authors making their tags and one uses Heroku and the other heroku, you could fix that easily sure, but how long until you notice that mistake? Did Google notice it first? (And they will notice Heroku vs heroku) but if we gave an option (or enforced, or provided a hook) this would not be a problem because their inconsistency can be made consistent.  This is where the unofficial rule comes into play, just lowercase everything in the URL for consistency (there are obvious exceptions to this rule -- but IMO tags is not one of them) but it's easier to just lowercase it all and normalize a slug so it's easy to read, type and blah blah blah.

gynter:
There is no such *golden* rule, official or unofficial. Everything in the Internet, [includig URLs](http://tools.ietf.org/html/rfc1738), work according to RFCs and other official standards.

But I agree, case sensitivity should be *enabled* by default, since lets be honest, most hosting providers use *nix systems, which are case sensitive. One could always force both tags and categories lowercase using `downcase` Liquid filter. Also everything else (pages, posts, collections) are currently case sensitive, which means they are accessible via case sensitive URLs, why should categories or tags behave differently?

envygeeks:
I'm pretty sure "I consider" marks it as a personal opinion and a golden rule to me and I believe you are looking for https://tools.ietf.org/html/rfc3986 Intro to 6.2.4 and 6.2.4 which raises the same concerns I did, which is why I have my unofficial golden rule.

gynter:
I don't disagree with you on the described above, search engines are not happy about mixedcase URLs and those can mess things up quite bad.

Maybe add a configuration option for lowercase URLs which would force everything (including the template variables) related to pages, posts, and collections and categories lowercase? Thou that might raise an issue what to do if You have two pages with same name but different case, how should we deal with those?

envygeeks:
I think hooks are a good compromise, the collective gets what they want, and I get what I want because well I'm more than willing to make a hook based plugin to normalize tags and categories.  In the case of an option if there were clashing names I would raise an InconsistentError because they wanted consistency and now they broke that want but I still like the idea of merging this if @parkr want and us just adding a hook later so the outlier's can get what we want.

benbalter:
:+1: for a hook. As an added datapoint, GitHub Pages is case sensitive. A breaking change in Jekyll 3.0 will likely break URLs for a lot of sites.

parkr:
I really want there to be a way to do this without a configuration option. Hm.

x-way:
Why was this PR closed?

parkr:
Hey @x-way! Sorry about this, my bot deleted the master branch and GitHub closes all PR's who point to that base when it's deleted. I have restored the branch.

Case sensitive categories depends on the filesystem as well. If you have `Apple` and `apple` on a Mac, a post with either category can be requested via `/Apple/my-post.html` or `/apple/my-post.html`, so the case-sensitivity becomes total moot.

But let's give this another shot and maybe include it in 3.1 as an option. If enough folks like it, we can keep it around for a while :) 

Could you please add tests and documentation for this new option and rebase on the latest master? Thanks!

parkr:
Closed as stale.

Issue #3870, opened 2015-07-27
hooks don't work on subclasses


stevecrozz:
Currently Jekyll triggers hooks by matching the class of the triggering instance against a map of classes to symbols. This causes a problem described in octopress/paginate#7 where hooks are not triggered because the instance in question is a subclass of Page and not an instance of Page.

We need to come up with a way to support this use case.

imathis:
Thanks so much!

parkr:
We should use `is_a?` in my opinion instead of `class` if possible. That may require a slight API change, though, yeah? Instead of `:document`, we'd pass `Jekyll::Document` (Symbol being replaced by a Class).

stevecrozz:
@parkr: I think what it comes down to is what we want to do here:
https://github.com/jekyll/jekyll/blob/master/lib/jekyll/hooks.rb#L92

It would be a bigger change than just swapping `class` for `is_a?` because in order to do it we'd have to enumerate the options and call `is_a?` for each one. We can do it without even an external interface change, but I hesitate because Hash lookups are so much faster than calling is_a? in a loop.

We could complicate the internal trigger interface by requiring callers to pass the owner symbol in addition to their current parameters. That would help keep hide the complexity from end users and could also be done without changing the external interface. I think this is along the lines of what @envygeeks proposed to me, but I'd like to confirm with him.

envygeeks:
We might have to take a double approach unless we want to force people to reintegrate.  Here is an example:

```ruby
Jekyll::Hooks.trigger :site, :post_render, data
Jekyll::Hooks.trigger data, :post_render
```

We are late in the 3.0 cycle but there is a need for hooks to be more flexible, there are many cases where we would need to highly reconfigure pieces of Jekyll's core to get hooks integrated into them and if we don't accept both we break one or the other way late in our release schedule.

Lets take for example my want to add a `:post_init` hook for `:configuration` so that plugins can attach themselves and create default configurations, or other various hook places.  Right now this can be complicated or overly loose because what we had assumed was a Configuration being passed ended up being a Hash being passed, but if we had the ability to trigger like the above we could solve this problem.

The decision has to be made to either make triggers generic allowing anybody to call or to make them based on class which simplifies your code, if it comes down to it, my vote is going to be accepting `Jekyll::Hooks.trigger :site, :post_render, data` because it ultimately simplifies everything and does not have any specific tie.

stevecrozz:
[...]

PR #3871, opened 2015-07-27
#3870 trigger hooks by owner symbol

stevecrozz:
Addresses #3870 by changing the trigger interface to work with symbol owners rather than object owners, making the hooks system more flexible.

### DIFF
--- a/lib/jekyll/convertible.rb
+++ b/lib/jekyll/convertible.rb
@@ -144,6 +144,15 @@ module Jekyll
       end
     end
 
+    # returns the owner symbol for hook triggering
+    def hook_owner
+      if is_a?(Post)
+        :post
+      elsif is_a?(Page)
+        :page
+      end
+    end
+
     # Determine whether the document is an asset file.
     # Asset files include CoffeeScript files and Sass/SCSS files.
     #
@@ -236,7 +245,7 @@ module Jekyll
     #
     # Returns nothing.
     def do_layout(payload, layouts)
-      Jekyll::Hooks.trigger self, :pre_render, payload
+      Jekyll::Hooks.trigger hook_owner, :pre_render, self, payload
       info = { :filters => [Jekyll::Filters], :registers => { :site => site, :page => payload['page'] } }
 
       # render and transform content (this becomes the final content of the object)
@@ -250,7 +259,7 @@ module Jekyll
       self.output = content
 
       render_all_layouts(layouts, payload, info) if place_in_layout?
-      Jekyll::Hooks.trigger self, :post_render
+      Jekyll::Hooks.trigger hook_owner, :post_render, self
     end
 
     # Write the generated page file to the destination directory.
@@ -264,7 +273,7 @@ module Jekyll
       File.open(path, 'wb') do |f|
         f.write(output)
       end
-      Jekyll::Hooks.trigger self, :post_write
+      Jekyll::Hooks.trigger hook_owner, :post_write, self
     end
 
     # Accessor for data properties by Liquid.
--- a/lib/jekyll/hooks.rb
+++ b/lib/jekyll/hooks.rb
@@ -1,13 +1,5 @@
 module Jekyll
   module Hooks
-    # Helps look up hooks from the registry by owner's class
-    OWNER_MAP = {
-      Jekyll::Site => :site,
-      Jekyll::Page => :page,
-      Jekyll::Post => :post,
-      Jekyll::Document => :document,
-    }.freeze
-
     DEFAULT_PRIORITY = 20
 
     # compatibility layer for octopress-hooks users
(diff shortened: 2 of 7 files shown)


envygeeks:
Does this require a documentation update?

/cc @parkr @imathis

stevecrozz:
The old docs still apply since the only real change was for the 'trigger' method which is undocumented.

lib/jekyll/convertible.rb
```diff
@@ -144,6 +144,17 @@ def type
       end
     end
 
+    # returns the owner symbol for hook triggering
+    def hook_owner
+      if is_a?(Draft)
```
parkr: This is superfluous, I believe. `Draft.is_a? Post` should be true.

stevecrozz: You are right. I'll make that change.

lib/jekyll/hooks.rb
```diff
@@ -88,19 +80,17 @@ def self.insert_hook(owner, event, priority, &block)
...
+      return unless @registry[owner][event]
 
       # hooks to call for this owner and event
-      hooks = @registry[owner_symbol][event]
+      hooks = @registry[owner][event]
 
       # sort and call hooks according to priority and load order
       hooks.sort_by { |h| @hook_priority[h] }.each do |hook|
-        hook.call(instance, *args)
+        hook.call(*args)
```
parkr: this will require a doc update, right?

lib/jekyll/hooks.rb
```diff
@@ -88,19 +80,17 @@ def self.insert_hook(owner, event, priority, &block)
     end
 
     # interface for Jekyll core components to trigger hooks
-    def self.trigger(instance, event, *args)
-      owner_symbol = OWNER_MAP[instance.class]
-
+    def self.trigger(owner, event, *args)
```
parkr: i think we can keep using instance, and just accept either a symbol or an instance of something.

stevecrozz: The situation that prompted this change was one where a user had subclassed Page. Let's say you do that by declaring ```class MyPage < Jekyll::Page; end```. The expected behavior is at certain stages of the build, Jekyll should trigger hooks.

The problem is Jekyll doesn't currently know that MyPage should trigger :page hooks. MyPage relies on its superclass and potentially other internal Jekyll components to trigger hooks at the right times. But if what we have is an instance of MyPage, how can we translate that into the right hash key, :page in this case, to look up the hooks.

I had a few ideas and this PR represents the simplest, but maybe also the ugliest from the caller's perspective. But here's a list of the ideas I've had so far. Let me know if you have any others:

1) Require the trigger caller to pass they key in the trigger method call (this PR)

2) Give every type of owner an instance method, class method or constant (something heritable) that calls out the owner symbol. Take this change to Jekyll::Page for example:
```ruby
module Jekyll
  class Page
    def self.hook_owner; :page; end
  end
end
```
Instances of Jekyll::Page and instances of Jekyll::Page descendants could now just ask the instance of MyPage or instance of Jekyll::Page for its .class.hook_owner and know how to find the hooks. This could essentially replace the lookup table at the top of Jekyll::Hooks and make each hook owner declare such a method.

The idea is the same, but there could be some more sugary syntax:
```ruby
module Jekyll
  class Page
    ...
  end
  Hooks.add_owner :page, Page
end
```

3) Make the Hooks class search the class hierarchy of the instances passed in. For example if you pass in an instance of MyPage (subclass of page), Hooks would quiz that instance to see if it is_a?(Jekyll::Page) and if so trigger :page hooks. If not, ask to see if it is_a?(Jekyll::Post) and so on until it has checked all the possible hook owners it could be.

envygeeks: The cleanest and most direct IMO is option one.  It's not just that area we had trouble, it was within Jekyll itself too.  The problem within Jekyll's own source is what actually started this entire conversation weeks before this one spawned the change.

Where at one point we were trying to add in hooks for Jekyll's internal configurations so we could hook out dozens of points, we couldn't get a subclass because it was Hash because the code did not speak for itself. That's too broad, so we needed to make things open and explicit in that a user should have to pass an owner and a hook, this makes it so that anybody can call any hook at any point for any reason.

It's also not just us we need to worry about, it's plugins too, the hook system is only as valuable as it's usage so we have really two problems here:

1. How do we trigger hooks.
2. How do we let plugins have their own hooks so this: https://github.com/jekyll-assets/jekyll-assets/blob/master/lib/jekyll/assets/hook.rb doesn't have to happen.

IMO there should be no basis for resistance against changing the API to the more explicit version because hooks haven't been around long enough to gain usage, there was never any option for plugins to use them directly (which means a plugin can't be broken) and the messaging to hooks was only used internally so this shouldn't in theory and in testing break anything like Octopress, because it's only asking to add a hook to a point... we are the ones who are messaging the hooks on the point... not Octopress.

parkr:
Ok, I'm fine moving forward with this PR. Anything else we need?

PR #3877, opened 2015-07-30
Update test_filters.rb

tasken:


envygeeks:
Now you've updated the test without the changes you made, both of those changes (the old and the new) need to be in the same pull request.  You need not send a new pull request, just force push the changes.  So if you edit a filter, edit the test.

tasken:
could you specify a little tutorial for that? I looked online how to add files to a commit and always get errors when rebasing or force pushing to jekyll/jekyll.

all comits made on tasken/jekyll:

b880b4c98c356959eab1cb950b64816a06df7a21
ea5957e03ae03335eb2b0542a096170ca0fc8391
43c38dc899ca0e1f1d43e4ac222eee51b15ddf86

tasken:
Ok, i think i've managed to do it!

parkr:
@tasken What's your motivation for this change? I prefer the [oxford comma](http://www.grammarly.com/blog/2015/what-is-the-oxford-comma-and-why-do-people-care-so-much-about-it/) there for clarity's sake.

tasken:
i think that should be optional as per default AP style is more common and straightforward.

to get a csv from page.tags double replace is needed to cover 2 and +3 tags cases-> array_to_sentence_string | replace:' and ',', ' | replace:',,',','

parkr:
To get a CSV from `page.tags` you can do `{{ page.tags | join "," }}`. I don't think that's a good reason to change this personally.

envygeeks:
I'm confused as to what AP style list output has to do with CSV to begin with.  I would do exactly as @parkr suggested, the methods you edited are for humanized output.

tasken:
A little confusion maybe, at first thought that was an error, the tags was an example of the problem I found. In spanish we don't use the oxford comma, and hence the confusion.

PR #3880, opened 2015-08-01
Add site.html_files to variables docs

probins:


envygeeks:
Please fix your html formatting! Thanks.

probins:
how do you mean? I didn't format anything. Just did a copy/paste in the online editor.

site/_docs/variables.md
```diff
@@ -130,6 +130,14 @@ following is a reference of the available data.
       </p></td>
     </tr>
     <tr>
+      <td><p><code>site.html_files</code></p></td>
+      <td><p>
+
+        A list of all static HTML files.
```
parkr: This might be a little vague. What do you think about `A list of all static files which end in .html`?

probins: I just copied what was in `html_pages`. It might be better if they both said something like that - make it clear that `html_pages` is a subset of `pages` and `html_files` is a subset of `static_files`. This wasn't clear to me from the current docs, though it's clear once you loop through them to see what they contain.

parkr: Sure, you could include that piece.

probins:
amended as discussed, and changed commit msg

Issue #3882, opened 2015-08-01
Jekyll's future posts command line switch is kind of weird


nateberkopec:
So, it seems that Jekyll's default value for the `future` option is `true`. So, by default, posts with a future date are published. OK, fair enough, I might not have chose that value, but you've probably thought about it longer than I have.

What's weird, though, is that there's also a command line switch `--future`. So, by default, `jekyll build --future` doesn't do anything - `future` is already true.

Is it weird to have a switch that, by default, does nothing? Are the only people getting use out of the `--future` switch those of us who set `future` to false in our configs? Is there some way to `jekyll build --future=false` that I'm missing?

alfredxing:
I agree! Having the `future` options is not very useful IMO, especially since we already have `published`...

There is also a comment beside the `future` option in Configuration saying that it should be removed, and I think now is a good time to do it since it will technically break compatibility.

/cc @jekyll/core

envygeeks:
I'm :+1: on dropping the option to false by default.

nateberkopec:
@alfredxing I want to disagree with removing `future` entirely, because `future` enables a very cool workflow for me:

1. I write a post with a date in the future in the `_posts` folder. With `future` set to false, it is not built into the site. 
2. I have a script running every hour that rebuilds my site and uploads it to S3.
3. Thus, when the posts date comes around, my post is automatically added to the site.

That sort of publish-at-this-time workflow is, AFAIK, not possible with just `published` or the current drafts feature.

@envygeeks Is there any backwards-compat concerns for changing default config options like this? We could be dropping posts from people's sites, right? Just wanted to bring it up.

parkr:
`future` should definitely default to `false`... it was probably just overlooked in a previous release. Would you update that for us? Then the flag will make more sense... :smile:

nateberkopec:
Sweet. Can do.

fizerkhan:
+1

mykagrove:
+1
With the current version it always seems to publish posts with future dates no matter what I do.
That is to say, files named similar to: ```2017-01-01-test-post.md```
Compiling with any of:
```
jekyll build
jekyll build --future
jekyll build --future false
```

nateberkopec:
Fixed in #3892.

d3netxer:
so I spent a lot of time trying to figure out why some of my posts were not displaying. There was no warning when running Jekyll that my posts were not being published. Therefore I support for future being set to true as default.

[...]

PR #3892, opened 2015-08-05
Future should be set to false in the default config

nateberkopec:
Solves #3882


lib/jekyll/configuration.rb
```diff
@@ -26,7 +26,7 @@ class Configuration < Hash
       # Filtering Content
       'show_drafts'   => nil,
       'limit_posts'   => 0,
-      'future'        => true,           # remove and make true just default
```
parkr: Oh interesting... it looks like we were going to remove it anyway. Looks like #690 is where we wrote that comment. I'm cool on keeping it.

parkr:
Cool. Just needs a test, please!

nateberkopec:
@parkr I'm a little confused. You're asking me to test that Jekyll::Configuration::DEFAULTS['future'] == false? I can't find any similar tests that test what the DEFAULTS constant is set to. Can you give me some guidance on what kind of test you're looking for?

parkr:
> You're asking me to test that Jekyll::Configuration::DEFAULTS['future'] == false?

Yes I am, but not directly. I would like to see a test that ensures that a post in the future is not in the output site. This can be tested in a number of ways, one of which is a cucumber feature. The test you changed, in fact, is a good place to start. Write a post that has a year of 2020 and ensure it isn't written by default and is written when `--future` is true.

nateberkopec:
@parkr ping!

nateberkopec:
@parkr alright actual ping. Testing that a future post isn't written by default + it is written with jekyll --build.

features/create_sites.feature
```diff
@@ -112,16 +112,16 @@ Feature: Create sites
...
     And I have a _posts directory
     And I have the following posts:
       | title  | date       | layout | content             |
       | entry1 | 2007-12-31 | post   | content for entry1. |
-      | entry2 | 2020-01-31 | post   | content for entry2. |
+      | entry2 | 2008-01-01 | post   | content for entry2. |
     When I run jekyll build
     Then the _site directory should exist
-    And I should see "URL: /2020/01/31/entry2/" in "_site/index.html"
+    And I should see "URL: /2008/01/01/entry2/" in "_site/index.html"
```
parkr: Maybe just negate this? "I should not see..."

parkr:
Merged in 559cd6c7e70f96bbda7cac3d36aee13a03206a39

Issue #3893, opened 2015-08-05
Contributing.md feature command doesn't work


nateberkopec:
Contributing.md suggests `bundle exec rake features` to run the feature tests. This did not work for me:

```
$ bundle exec rake features
/Users/censored/.rbenv/versions/2.2.2/bin/ruby -S bundle exec cucumber  --profile travis
cucumber.yml was not found.  Current directory is /Users/censored/Documents/Code/jekyll.  Please refer to cucumber's documentation on defining profiles in cucumber.yml.  You must define a 'default' profile to use the cucumber command without any arguments.
Type 'cucumber --help' for usage.
```

That's sort of annoying. I just want to run the feature tests, not set up Cucumber. `script/cucumber` does seem to work, but that isn't the command documented in Contributing.md.

Also noted in #3605

parkr:
Can you update Contributing.md to refer to `script/cucumber` please?

PR #3894, opened 2015-08-05
Contributing.md should refer to script/cucumber

nateberkopec:
See #3893 

I believe the bundle exec here is probably significant if the user has other versions of cucumber installed.


parkr:
Closes #3893.

PR #3902, opened 2015-08-10
jekyll-docs intall instructions improvement

idmarjr:
Making it clear to the user that is necessary to use `--pre` when installing `jekyll-docs` gem.


parkr:
Should be fixed now. Will release in tandem with Jekyll from now on. https://github.com/jekyll/jekyll-docs/releases/tag/v2.5.3

PR #3908, opened 2015-08-11
Update structure.md

jaybe-jekyll:
Add in references to _data formats and extensions options of .csv, .json, etc., consistent with _docs/datafiles.md

### DIFF
--- a/site/_docs/structure.md
+++ b/site/_docs/structure.md
@@ -133,8 +133,8 @@ An overview of what each of these does:
         <p>
 
           Well-formatted site data should be placed here. The jekyll engine
-          will autoload all yaml files (ends with
-          <code>.yml</code> or <code>.yaml</code>) in this directory. If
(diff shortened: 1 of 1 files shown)


parkr:
Thanks! Not sure if you meant to remove the styling but unfortunately I don't believe the backticks work there for anything more than plain ol' backticks.

envygeeks:
AFAIW all Markdown processors recognize '` as `<code>` because it's listed in the official page.

pathawks:
I believe that, because it is listed inside a `<p>` tag, Markdown processors will ignore everything and just treat it as raw HTML, which is why the `<code>` tag was there.

envygeeks:
I didn't even notice that, thanks for pointing that out.

parkr:
Merged in 4c1427c. Thanks, @jaybe-jekyll

jaybe-jekyll:
Thanks; in my haste I did not update to match existing styling; and it should match. Thanks!

PR #3917, opened 2015-08-16
Added another site for learning purposes

leo:
A week ago, I asked @parkr via email if he could add my site here (mostly because I thought it's too cheeky to just propose a pull request). But now he told me that it's better to just do it here:

I'm asking because I spend a huge amount of time and effort on making it great and usefully structured for people who're just getting started with Jekyll. Therefore it's also great as a forked starting-point, if you ask me. Besides keeping the code clean, I also spend much time on making the site as fast as possible. There's not much CSS in use, the HTML is clearly structured and I've did a lot for the performance.

There's also a lot of "include"-thinking happening for things like embedded Tweets, images or iFrames - which most people just inline in each post. When making a significant change, I also always make sure to write a few paragraphs about why I exactly did it as a commit message. And when it comes to really big updates, I write entire posts too (explaining all improvements and their benefits to the site's performance/look).

Here's an recent example: http://leo.im/notes/v2/

I'm definitely sure that many people could get something out of it. Don't you think so too?

### DIFF
--- a/site/_docs/sites.md
+++ b/site/_docs/sites.md
@@ -20,6 +20,8 @@ learning purposes.
     ([source](https://github.com/rsms/rsms.github.com))
 - [Scott Chacon](http://schacon.github.com)
     ([source](https://github.com/schacon/schacon.github.com))
+- [Leonard Lamprecht](http://leo.github.io)
+    ([source](https://github.com/leo/leo.github.io))
 
 If you would like to explore more examples, you can find a list of sites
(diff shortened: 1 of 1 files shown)


parkr:
Leo emailed me a couple times. I haven't taken the time to really dig into his site but it seems ok from an eagle-eyed view. I won't accept everyone's website, but this one seems OK for now as long as there's no offensive content, etc.

leo:
@parkr Thanks! There are no such things in my posts.

PR #3918, opened 2015-08-17
I think you made a typo there

gangsthub:
as seen on:
http://blog.davepoon.net/2015/01/19/setting-up-sass-with-jekyll/
http://markdotto.com/2014/09/25/sass-and-jekyll/#create-a-sass-stylesheet

### DIFF
--- a/site/_docs/assets.md
+++ b/site/_docs/assets.md
@@ -73,7 +73,7 @@ You may also specify the output style with the `style` option in your
 
 {% highlight yaml %}
 sass:
-    style: :compressed
+    style: compressed
 {% endhighlight %}
 
 These are passed to Sass, so any output style options Sass supports are valid


parkr:
LGTM.

TWiStErRob:
I think it's actually those articles that are wrong. I removed this colon: https://github.com/TWiStErRob/twisterrob.github.io/blob/master/_config_dev.yml#L22 and the generated CSS style is not expanded any more (from `_config_dev.yml`), but the compressed (the one defined in `_config.yml`), so without colon the setting is being ignored.

parkr:
@TWiStErRob [We strip out the leading `:`](https://github.com/jekyll/jekyll-sass-converter/blob/83101840b53c57d660d41eae27a059755dcdc07e/lib/jekyll/converters/scss.rb#L29), so Chris is correct for the latest code. What version are you using?

TWiStErRob:
Ah the latest. I use GH-Pages gem, maybe 2.6.0... I wish I could use edge on GH.

envygeeks:
@TWiStErRob you can, just build it and push the HTML (`_site`) instead of letting Github build it.

pathawks:
I can confirm that the colon is *not* necessary on Pages Gem v39 (Jekyll 2.4.0)

PR #3925, opened 2015-08-20
Add a Code of Conduct.

### LINKED ISSUE #3924: Adopt a code of conduct.
/cc @jekyll/owners @jekyll/core

I like many in the community think we need a code of conduct but I like a lot of the community think that some parts of the Open Code of Conduct by the TODO group is a bit ambiguous and hostile so in keeping with good faith I think we should adopt the https://github.com/rails/rails/blob/master/CODE_OF_CONDUCT.md (except name it CONDUCT.md or COC.md.) I am proposing that we adopt a conduct like a lot of the community but pick one that everyone can agree on.

parkr:
Closes #3924.

/cc @jekyll/core

### DIFF
--- a/CONDUCT.md
+++ b/CONDUCT.md
@@ -1,0 +1,22 @@
+# Contributor Code of Conduct
+
+As contributors and maintainers of this project, and in the interest of fostering an open and welcoming community, we pledge to respect all people who contribute through reporting issues, posting feature requests, updating documentation, submitting pull requests or patches, and other activities.
+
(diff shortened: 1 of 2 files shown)


envygeeks:
I'm totally :+1: on this without the alteration.  I've thought about it and there is no need to adjust it.

mattr-:
Cool. :sunglasses: 

Stand back! I'm going to do science! (aka merge a pull request)

alfredxing:
:+1:

I've always believed that bad/disrespectful behaviour is always punishable with or without a code of conduct, but at least having one gives us something concrete we can point to.

Issue #3914, opened 2015-08-14
Scssify filter can result in BOM in HTML


benburwell:
I'm using the `scssify` filter to spit out a `:compressed` stylesheet in a `<style>` block. Since my SCSS file contains non-ASCII characters, [Sass places a byte-order mark at the beginning of the compressed output](https://github.com/sass/sass/issues/1449). So effectively, I have:

```
<style>

[BOM]a{color:red}

</style>
```

Having the BOM immediately before a selector appears to be an issue, as that rule is not interpreted in Chrome. Changing the Sass mode to `:compact` replaces the BOM with a `@charset` rule, which works fine. I'm not sure if this would be an issue with Sass or with Jekyll (or with Chrome, as FF appears to ignore the BOM), but I figured I'd look for some feedback here.

FWIW, the [W3C CSS Validator](https://jigsaw.w3.org/css-validator/#validate_by_input) also chokes up on BOM occurring after beginning-of-input:

```
    ﻿a { color: red; }
```

(there is a BOM just before the `a`, so copy-and-paste carefully)

From what I've read, the CSS spec requires implementations to interpret the BOM at the beginning of the file, but I haven't seen anything about it occurring in the middle.

envygeeks:
This isn't a bug in Chrome, the opposite would be a bug in FF IMO because ignoring mistakes and proceeding is whacky behavior, but it's not something that is even important.  At minimum I think this would be a bug in SASS since we pass it onto them to deal with AFAIW we could work around it in Jekyll though I don't know if that's worth it, all things considered, I don't think we heavily lock SASS so if they rolled out a fix in theory you could get it as fast as they rolled it out as long as it's not major.

benburwell:
@envygeeks, I'm totally with you on the browser issue. I'm not too familiar with the internals of Jekyll, so please correct me if I'm wrong. It seems that the `scssify` filter and the SCSS converter both (reasonably) use the same Sass plugin.

I was looking through the Sass source and I noticed a distinction between the `compile` and `compile_file` functions. Perhaps adjusting `compile` to not insert the BOM but leaving it in `compile_file` would be the most sensible thing. I'm not sure whether that would work well within the Sass architecture, but I'll see what they have to say.

My next question is, would this actually result in the desired behavior in Jekyll? It looks to me like `scssify` uses the same converter that compiling a SCSS file does. Would this need to be changed as well?

envygeeks:
[...]

PR #3934, opened 2015-08-26
Add BOM substitution and tests for Sass and SCSS filters

benburwell:
Addresses #3914. The tests feel a little clunky, what with creating a new JekyllFilter, but I didn't want to convert the existing test cases to `:compressed` mode without an explicit go-ahead/request to do so.

### DIFF
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -38,6 +38,7 @@ module Jekyll
       'highlighter'   => 'rouge',
       'lsi'           => false,
       'excerpt_separator' => "\n\n",
+      'add_charset'   => true,
 
       # Serving
       'detach'        => false,          # default to not detaching the server
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -23,7 +23,7 @@ module Jekyll
     def sassify(input)
       site = @context.registers[:site]
       converter = site.find_converter_instance(Jekyll::Converters::Sass)
-      converter.convert(input)
+      converter.convert(input).sub(/^\xEF\xBB\xBF/, '@charset "UTF-8";')
     end
 
     # Convert a Scss string into CSS output.
@@ -34,7 +34,7 @@ module Jekyll
     def scssify(input)
       site = @context.registers[:site]
       converter = site.find_converter_instance(Jekyll::Converters::Scss)
-      converter.convert(input)
+      converter.convert(input).sub(/^\xEF\xBB\xBF/, '@charset "UTF-8";')
     end
 
     # Slugify a filename or title.
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -23,7 +23,10 @@ module Jekyll
     def sassify(input)
       site = @context.registers[:site]
       converter = site.find_converter_instance(Jekyll::Converters::Sass)
-      converter.convert(input).sub(/^\xEF\xBB\xBF/, '@charset "UTF-8";')
+      output = converter.convert(input)
+      add_charset = @context.registers[:site].config['add_charset']
+      replacement = add_charset ? '@charset "UTF-8";' : ''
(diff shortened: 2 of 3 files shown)


envygeeks:
I don't know that we should add `@charset` more than just entirely strip it, unless there is something I'm missing?   Other than that :shipit:

parkr:
Fixes #3914.

parkr:
> I don't know that we should add `@charset`

What's the rationale here?

benburwell:
Rationale for adding the charset was that the BOM is intentionally inserted by Sass instead of the charset rule under particular circumstances to signal to the CSS client that it's a UTF-8 encoding. With the substitution, the charset rule would only be added if a BOM is removed.

If we are guaranteed to be generating a UTF-8 file, (with a BOM itself, or a HTML meta tag, or served with that encoding header), we could safely drop this according to (my reading of) the spec.

parkr:
Is this a setting of the Sass converter? Maybe we could tweak it over at jekyll-sass-converter.

I could see people not wanting the charset injected into their snippets

benburwell:
Basically, when the Sass/SCSS source contains non-ASCII characters, [Sass inserts either a BOM or the charset rule](https://github.com/sass/sass/blob/master/lib/sass/tree/visitors/to_css.rb#L131). If the compressed mode is being used, it's the BOM, otherwise it uses the charset rule.

So the only way there would be a BOM to begin with is if the file contains non-ASCII characters, in which case Sass would be adding the charset rule anyway.

parkr:
I'm fine with adding in the charset if we think they'll only end up in CSS files or style blocks. I'd hate to have someone have some highlight block that has that header thrown in when they don't want it.

benburwell:
That makes sense. I'm not too familiar with all of Jekyll's internals -- is there a place to differentiate in what context the converter is being called? I looked around a bit but I didn't find anything. May easily have missed it though.

parkr:
> is there a place to differentiate in what context the converter is being called?

No sir, this is not available. So we'll have to choose something and decide it fits for all circumstances. I don't see `scssify` being used in snippets all that often, so let's just go ahead with what you have and we'll see what feedback you get.

@envygeeks, any thoughts?

envygeeks:
Jekyll filters support multiple arguments so we could always create an undocumented API argument to switch the context and if people notice it... they can abuse it otherwise it's there for us to differentiate internally without having to dup code.

envygeeks:
Any update on this? We are coming to close of the 3.0 window so if you want to see this in the release.

parkr:
> Jekyll filters support multiple arguments so we could always create an undocumented API argument to switch the context and if people notice it... they can abuse it otherwise it's there for us to differentiate internally without having to dup code.

They're instantiated with the site config, so that could be used here. Nevertheless, let's strip the BOM. Add an option to include the CHARSET declaration.

benburwell:
Okay, I'll update the PR to use a config option for whether to add the charset rule. Probably not until tomorrow night, but possibly later this evening.

envygeeks:
Just noticed it was updated /cc @jekyll/core -- Should be ready for re-review.  I'm :shipit:

lib/jekyll/configuration.rb
```diff
@@ -38,6 +38,7 @@ class Configuration < Hash
       'highlighter'   => 'rouge',
       'lsi'           => false,
       'excerpt_separator' => "\n\n",
+      'add_charset'   => true,
```
parkr: What do you think about making this a sass rule?

```yaml
sass:
  style: compressed
  add_charset: true
```

then we can take care of it over in jekyll-sass-converter. i think this shouldnt' be in the core now that i have had more time to think on it.

benburwell:
I agree. I was actually thinking about that last night when I was putting this in. I'll submit a new PR in jekyll-sass-converter tonight. I think this one can be closed then.

parkr:
Thanks, @benburwell!

PR #3940, opened 2015-08-27
Fix the formatting

chrisfinazzo:


envygeeks:
I don't think this fixes anything, but I do think that it makes the formatting feel off.

envygeeks:
I guess it would be fine if they weren't forced onto their own line, but if you want to make it inline but when it being on it's own line, it should be a block like it was, unless it was an accident it was made "`" instead of "```"

parkr:
I like the way the 4-space indents give these lines a significance and separation that is easier on the eye and the brain. Thanks for the contribution, but I think I'm going to leave it as is. :+1:

parkr:
The style could use an update tho:

screen shot 2015-08-27 at 1 25 54 pm

chrisfinazzo:
@parkr, that's what I was getting at, for whatever reason these were not indented properly. I'll look at master again to see what's actually going on (my local branch has obviously diverged from this).

chrisfinazzo:
Huh, indented by 4 spaces and...nothing. No text styles whatsoever. 😒

envygeeks:
That's because 4 spaces and triple grave are treated different than {% highlight %} but you can't use {% highlight %} on that document because it's actually meant for Github (it's linked through by every ticket creation to nag you to follow the rules)

parkr:
@chrisfinazzo Our Markdown converter should add a `<pre>` – maybe we can add the style to that on this page?

envygeeks:
Manually wrap it: https://github.com/jekyll/jekyll/blob/master/lib/jekyll/tags/highlight.rb#L116

envygeeks:
With only the `<figure>` the rest of it is already created.

Issue #3938, opened 2015-08-26
Include directive with variable will not guess extension


fulldecent:
## Test case

**_includes/theInclude.htm**

    I'm in

**_layouts/page.htm**

    {% include {{ page.whichInclude }} %}

**aPage.htm**

    ---
    layout: page
    whichInclude: theInclude
    ---

## Expected

The resulting `aPage.htm` will contain `I'm in`.

## Actual

 > Liquid Exception: Included file '_includes/theInclude' not found in _layouts/page.htm

## Discussion

Tested and failed on Jekyll 2.5.3. The expected outcome WILL happen if we substitute `{{ page.whichInclude }}` to be `theInclude`.

parkr:
> The expected outcome WILL happen if we substitute {{ page.whichInclude }} to be theInclude.

Weird... I definitely didn't expect that myself.

What is the value of `{{ page.whichInclude }}` in your template?

fulldecent:
It is `theInclude`

fulldecent:
Confirmed still broken on jekyll 3.0.1

parkr:
@fulldecent Do you mean 3.1.6? 3.0.1 is not something we're still working on.

envygeeks:
I'm confused, is this a feature or a bug?

fulldecent:
I have tested again in 3.1.6 and found that:

 * `{% include {{ page.whichInclude }} %}` does not work (same since originally reported)
 * `{% include theInclude %}` does not work (different since reported, not sure when changed)

This behavior is now consistent. Thank you, closing issue.

PR #3947, opened 2015-08-30
Added support for including files without extension

j-allard:
This adds support for including files without specifying the extension.

```{% include head %}``` where ```head``` is in fact ```head.#{ext}```

AFAIK this doesn't affect the inclusion of files with an extension ```{% include head.html %}``` and if both ```head``` and ```head.#{ext}``` are present, an exception is thrown to tell the user that the filename is too ambiguous.
Potentially fixing https://github.com/jekyll/jekyll/issues/3938
Any thoughts (besides the lack of tests) ?

### DIFF
--- a/benchmark/guess-full-file
+++ b/benchmark/guess-full-file
@@ -1,0 +1,35 @@
+#!/usr/bin/env ruby
+require "benchmark/ips"
+require "find"
+
+def file_without_extension?(file)
+  File.basename(file, ".*") == File.basename(file)
+end
+
+def matching_process(path)
+   if file_without_extension?(path)
+     dir = File.dirname(path)
+     match = Dir["#{dir}/#{File.basename(path)}.{html, htm}"][0]
+     path = match unless match.nil?
+   end
+end
+
+Benchmark.ips do |x|
+  x.report "Matching without extension" do
+    40.times do
+      matching_process("/path")
+    end
+  end
+
+  x.report "Matching with extension" do
+    40.times do
+      matching_process "/prout.htm"
+    end
+  end
+
+  x.report "Matching already existing file without extension" do
+    40.times do
+      matching_process "/home"
+    end
+  end
+end
--- a/lib/jekyll/tags/include.rb
+++ b/lib/jekyll/tags/include.rb
@@ -1,4 +1,5 @@
 # encoding: UTF-8
+require 'find'
 
 module Jekyll
   module Tags
@@ -53,6 +54,10 @@ module Jekyll
           params[match[1]] = value
         end
         params
+      end
+
+      def file_without_extension?(file)
+        File.basename(file, ".*") == File.basename(file)
       end
 
       def validate_file_name(file)
@@ -112,6 +117,10 @@ eos
         validate_file_name(file)
 
         path = File.join(dir, file)
+        if file_without_extension?(path)
+          match = Dir["#{dir}/#{File.basename(file, ".*")}.{html, htm}"][0]
+          path = match unless match.nil?
+        end
         validate_path(path, dir, site.safe)
 
         # Add include to dependency tree
--- a/test/test_tags.rb
+++ b/test/test_tags.rb
@@ -600,6 +600,22 @@ CONTENT
       end
     end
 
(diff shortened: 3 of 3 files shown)


envygeeks:
I think telling the user they are being ambiguous when your code is being ambiguous is bad behavior IMO.  If you want so called "naked" files then the rule needs to be held that explicitness is far more valuable than implicitness.  So if I have head and head.html then head best be the one selected without an error being thrown because I asked for head and head existed.   Code needs to follow common sense and throwing an error in that case is not common sense.

In general though I'm personally :+1: on this, I'm just :-1: on the behavior of it.

j-allard:
Yeah you're probably right.
I rewrote the PR so direct matches take priority over partial matches and I added a test.

```{% include head %}``` would now match ```head``` over ```head.html```  (provided that they are both present).

j-allard:
Are there any other objections to merge this ?

envygeeks:
:shipit:

leo:
👍

parkr:
Looks pretty solid to me. Would love to see some benchmarks before we merge. Worried about slowing this tag down.

j-allard:
For what this is worth, here's a benchmark showing the performance of the matching stuff
image

Did you have something more specific in mind ?

I changed the PR again to make sure it only tries to match the file when it *doesn't* exist.

1) If you try to include ```head``` and its already present, it will just skip the call to .```match_file_without_extension()``` (and this is probably the most expensive one) and use the ```head``` file directly.

2) If you try to include ```head.html```, there's an extension to the file so ```match_file_without_extension()``` is skipped here too.

3) If you try to include ```head``` and it's *not* present, it will try to match it against something, there will be a slow down because of the match, but I guess that's better than an error because otherwise the file doesn't exist ?

If you have any more pointers about how to possibly make it better, I'm very interested :)

envygeeks:
Do you mind switching that  to use benchmark/ips.  Also I would simply do: `Dir["{head,head.*}"][0]` of course you'll want to use Jekyll's from_source_dir (or include_dir) but that simplifies the code drastically because now you are putting it on Ruby to do the lifting.

envygeeks:
The latter part where I said "simply do" was me saying what i would do, not me saying to switch it!

j-allard:
So I switched the benchmark to benchmark/ips and I changed the find loop for Dir.glob.
I didn't see why I would need from_source_dir (or include_dir) since we already know where the included file is.
I assumed we only want to match against files in the same dir, or else it could be confusing (?)

envygeeks:
I didn't notice that dir was being constrained ahead of time (at least if my quick read of the entire file gave me the right hints.) I made that comment on the assumption it might be done, so you are right there is no need for you to do it!

:shipit: /cc @jekyll/core

mattr-:
What happens if more than one file matches? i.e. I've got (for whatever reason) a head.html and a head.md

In general, i'm :-1: on this. I don't see how the loss of typing a few extra characters is worth the added ambiguity in the markup that this adds. "include head? head what?"

j-allard:
It currently only matches against *.html* and *.htm* files, so in your example *head* would only match *head.html*.
If there's more than one match, it simply pick the *first one* it finds. There's currently no sorting happening, and the default order of the files returned by *Dir.glob* seems unspecified.

This patch only affects cases where a file is included without any extension.
There won't be any matching magic invoked if you'd try to include, let's say, *head.html*.

Is there any particular reason why this wasn't merged yet ? I'd still be very happy to work on it if anything else needs to be improved !

envygeeks:
It's not merged because it's waiting for a consensus, only small changes are merged without it.

parkr:
I'm willing to hear more argument, but in general I don't see the need for this. More ambiguity doesn't help us in Jekyll, I think. As new people try to pick this up, the more direct a site's source is to read, the better off they are.

parkr:
Not sure why jekyllbot just closed this. Some coding error, probably. :(

Extensions help users understand which file is included because it more directly maps to what is on their system and it's readily identifiable as a filename. Thanks for the pull request, but we're not ready to make this change at this time. :heart:

Issue #3958, opened 2015-09-02
.htaccess Files are ignored from exclude: filter in _config.yml


htmelvis:
The documentation states that dotfiles (.htaccess) should be placed in the `include:` filter in your config file. By default dotfiles are ignored.

 .htaccess is compiled and placed in the jekyll `destination:`  even if placed in the `exclude:` filter  

The issue stems from #742 and I believe it was brought back here: 
<img width="973" alt="screen shot 2015-09-02 at 2 18 10 pm" src="https://cloud.githubusercontent.com/assets/1952926/9640200/580bf248-517e-11e5-9959-f0c9c54f391b.png">

kenzierocks:
+1, this is a big problem if you have a rewrite in `.htaccess`.

ashmaroli:
To *exclude* `.htaccess` from being processed into `_site/`, insert an empty `include:` list in your config file:
```yaml
# _config.yml

# the empty 'include' list below overrides the default
# setting `include: [".htaccess"]`
include: []
```

Crunch09:
@ashmaroli Any further action we need to take here before we can close this one? Just wondering as it's open for two years.

PR #3959, opened 2015-09-02
Removed .htaccess from the include filter in the main configuration. …

htmelvis:
…As stated in the docs you must add this file manually if you want to include. Exclude would not work on .htaccess.  This solves Issue #3958 
screen shot 2015-09-02 at 12 57 04 pm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ tmp/*
 .jekyll-metadata
 /vendor
 /test/source/file_name.txt
+.idea/
--- a/features/create_sites.feature
+++ b/features/create_sites.feature
@@ -123,11 +123,11 @@ Feature: Create sites
     Then the _site directory should exist
     And I should see "URL: /2008/01/01/entry2/" in "_site/index.html"
 
-  Scenario: Basic site with whitelisted dotfile
-    Given I have an ".htaccess" file that contains "SomeDirective"
-    When I run jekyll build
-    Then the _site directory should exist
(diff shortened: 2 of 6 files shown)


pathawks:
Looks like you accidentally modified `.gitignore` ?

envygeeks:
Looks like it's using Intellij, sounds like a good idea to leave it there just in-case somebody else does in the future so we don't get commits of the settings.

htmelvis:
agh! intelliJ ignores it's project files in my project

envygeeks:
It looks like you caused tests to fail.

htmelvis:
My Bad, I should have read the complete contributing doc. I will fix this and resubmit.

htmelvis:
It looks like .htaccess has been added to almost every test as an input file to check against. I am still working on my version to not expect .htaccess and I am unsure if you want changes merged back in altered that way. It may be easier to change the docs (if your default dotfile makes sense) to reflect the dotfile being present in your test project.

htmelvis:
Removed tests to watch for .htaccess file creation in the final build. I have commented out the main test "watch the dotfile" in ```create_sites.feature```

parkr:
Hey @htmelvis, we can't accept this until 4.x as it would be considered a breaking change. One thing we could do is remove it if `.htaccess` is in `exclude`.

htmelvis:
Ok, that sounds good, completely forgot about this.

Issue #3956, opened 2015-09-01
Move to POSIX compliant so tiny/embedded systems can use our script/*


envygeeks:
Some systems just don't install Bash and nothing we do strictly requires Bash, we should move to POSIX compliant /bin/sh scripts (some systems point this to bash -- ahem Arch) but most actually point it to POSIX shellscript interpreters (like dash or busybox sh)

Even though Bash is widely available to move is a effortless design change.

parkr:
This should be pretty easy, though I'm ok if some of the scripts can't work in POSIX. Most folks are using bash or zsh for development.

Are these needed to build the jekyll docker containers?

envygeeks:
One will be within a few weeks when we start enforcing a passing Jekyll build to pass a build for our Docker images that way we don't accidentally push out a bad master build, it's part of my plan to get us on automatic builds for the `jekyll/master` image.

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least
three months.

The resources of the Jekyll Core team are limited, and so we are asking for your help.

If you can still reproduce this error on the `3.0-stable` branch or on `master`,
please reply with all of the information you have about it in order to keep the issue open.

Thank you for all your contributions.

PR #3962, opened 2015-09-05
Rewrite shell scripts so they are compilant with POSIX

j-allard:
i tested them all with *dash* and AFAIK they're all working as expected.
Tell me if I missed something.
Should fix (or help at least) this https://github.com/jekyll/jekyll/issues/3956


envygeeks:
I don't know where I stand on this since I already have all this done on my local Jekyll along with it all being linted and a few of those removed plus a new one added.

The only thing I can honestly say right now is that in one part you removed the ANSI stuff which was necessary and still left the "$1" in printf when it should have been adjusted to use printf properly.

j-allard:
Ha ! Well since you already got this working locally, I guess there's no need for me to play with this PR anymore. :-)

parkr:
Sorry about that, @j-allard! Thanks for the pr :)

Issue #3963, opened 2015-09-06
installation docs fail to document required version of Ruby


oresmus:
Jekyll apparently has system requirements which are not mentioned on any of these pages:

    http://jekyllrb.com/docs/quickstart/

    http://jekyllrb.com/docs/installation/

    http://jekyllrb.com/docs/troubleshooting/

Specifically, it uses a subpackage which says it requires Ruby version >= 1.9.3.

##### Background and details: 

I am running Mac OS X 10.7.5. It has Ruby 1.8.7, apparently from Apple. I have successfully installed RubyGems in the past.

I first solved a minor problem, namely that "gem install jekyll" hung with no error message, consuming 100% cpu until I killed it after a few minutes, by running

    % sudo gem update --system

as advised by the troubleshooting page. I then got a permission error from `gem install jekyll`, so I tried

    % sudo gem install jekyll

which started working, but then reported the system requirement mentioned above:

    % sudo gem install jekyll
    Fetching: fast-stemmer-1.0.2.gem (100%)
    Building native extensions.  This could take a while...
    Successfully installed fast-stemmer-1.0.2
    Fetching: classifier-reborn-2.0.3.gem (100%)
    ERROR:  Error installing jekyll:
    	classifier-reborn requires Ruby version >= 1.9.3.
    Exit 1
    % 

I will of course try to install a newer Ruby, but so far (as a Ruby beginner) I find multiple ways of doing this on OSX, no clear indication of which is best, and that many of them don't list their system requirements regarding required version of OSX.

##### Suggested fix:

The jekyll installation pages should list explicitly whatever oldest version of Ruby (and of Mac OS X) is supported.

Also, if practical, they should suggest a specific way of installing a newer Ruby on OS X which is known to work well with jekyll. (This may be helpful only to Ruby beginners, but Jekyll is useful for anyone wanting to use gh-pages, and many of them may have no other need for Ruby.)

jaybe-jekyll:
Updating the (OSX) system gem with sudo is an absolute Nono. Never mess with OSX internals in any way. 

If you desire a custom Ruby, use homebrew or the likes to achieve it. Rbenv is a optio. And works well. 

Furthermore I prefer to install Jekyll locally as a user and [only] in user space. That can be accomplished with something such as the following:

'gem install --no-rdoc --no-ri -n ~/bin --user-install --verbose jekyll'

envygeeks:
[...]

PR #3968, opened 2015-09-09
replace broken link; add other Mac-specific info

oresmus:
The link for 'use Xcode to install Command Line Tools' seems to be broken. I don't know what it contained, but I have replaced it with a short piece of text which describes how to do the same thing. I have also added a bit more advice which was helpful for me.

FYI, the existing Mac advice on this page includes using sudo. I have been told that using sudo in these commands is not advisable, but since I have no direct knowledge of the pros and cons of that (in this specific case), I am not completely removing that advice, only adding the alternative of not using sudo.

FYI, I am submitting this after the discussion in #3963 but it is not about the same doc bug. It covers most (but maybe not all) of the changes I might suggest to this page. (I would also like advice on where to put new info specific to Mac/ruby newbies, about installing ruby before trying the "quickstart".)

### DIFF
--- a/site/_docs/troubleshooting.md
+++ b/site/_docs/troubleshooting.md
@@ -40,19 +40,25 @@ export PATH=$PATH:/home/private/gems/bin
 export RB_USER_INSTALL='true'
 {% endhighlight %}
 
-On Mac OS X, you may need to update RubyGems:
+On Mac OS X, you may need to update RubyGems (using `sudo` only if necessary):
 
 {% highlight bash %}
 sudo gem update --system
 {% endhighlight %}
 
-If you still have issues, you may need to [use Xcode to install Command Line
-Tools](http://www.zlu.me/ruby/os%20x/gem/mountain%20lion/2012/02/21/install-native-ruby-gem-in-mountain-lion-preview.html)
-that will allow you to install native gems using the following command:
+If you still have issues, you can download and install new Command Line Tools (such as `gcc`) using the command
+
+{% highlight bash %}
+xcode-select --install
+{% endhighlight %}
+
+which may allow you to install native gems using this command (again using `sudo` only if necessary):
 
 {% highlight bash %}
 sudo gem install jekyll
 {% endhighlight %}
+
+Note that upgrading MacOS X does not automatically upgrade Xcode itself (that can be done separately via the App Store), and having an out-of-date Xcode.app can interfere with the command line tools downloaded above. One way to mitigate this without installing a newer Xcode is to simply rename `/Applications/Xcode.app`. (The effect of this can be checked in the output of `xcode-select -p`.)
 
 To install RubyGems on Gentoo:
 


site/_docs/troubleshooting.md
```diff
@@ -40,20 +40,26 @@ export PATH=$PATH:/home/private/gems/bin
...
+xcode-select --install
+{% endhighlight %}
+
+which may allow you to install native gems using this command (again using `sudo` only if necessary):
 
 {% highlight bash %}
 sudo gem install jekyll
 {% endhighlight %}
 
+Note that upgrading MacOS X does not automatically upgrade Xcode itself (that can be done separately via the App Store), and having an out-of-date Xcode.app can interfere with the command line tools downloaded above. One way to mitigate this without installing a newer Xcode is to simply rename `/Applications/Xcode.app`. (The effect of this can be checked in the output of `xcode-select -p`.)
```
parkr: What does renaming accomplish?

parkr:
Thanks!

oresmus:
@parkr , based on my experience/inference, renaming `/Applications/Xcode.app` prevents shell commands like `gcc` from using the binaries inside that version of `Xcode.app`, causing them instead to use other binaries (I am not sure if those are the same binaries which `xcode-select --install` downloads -- they might be, or they might have come from my recent upgrade of OS X from 10.7.5 to 10.10.5). 

My evidence for this is that before I renamed it, `xcode-select -p` reported a path inside `/Applications/Xcode.app` and `gem install jekyll` had compilation errors (of the sort that using an old Xcode might explain, and in the very first gem of many), but after doing it, `xcode-select -p` reported a path not inside any Xcode.app, and `gem install jekyll` compiled many gems and completed without errors.

parkr:
Ok. I don't think renaming is a great way to mitigate issues with an out-of-date Xcode. The best way is to update Xcode or to use gcc from Homebrew or something.

oresmus:
Renaming is equivalent to deinstalling (but more easily reversible), and the old version was causing problems, so it might not be the best way, but it is sometimes useful.

Anyway, feel free to accept this PR and further edit it; or if you would prefer me to modify the PR, I could either remove the last sentence about renaming (but leave the useful info before it), or I could modify the last sentence to not actually recommend renaming, but still carry the useful info (which I did not know before and was important for me to find out) about how the command-line tools work differently depending on what is present at the file path `/Applications/Xcode.app`.

envygeeks:
Personally I think this is better suited in the gotcha's with a strong warning that it should be an absolute last attempt because as a systems guy I would be highly annoyed if I seen a system that did some hacky broken stuff like this when there are other more viable solutions.

parkr:
Thanks!

oresmus:
@parkr, You're welcome -- good solution.

(I don't have time to write a "mac gotchas page" for now, but your revised commit already contains the main things such a page would have needed to say, so maybe a new page is not needed. The only other info I could add, that might be useful and that I know enough to say correctly, is "Mac OS 10.7.5 is probably too old to be worth trying this on", and "no need for rvm or rbenv, just use brew to upgrade to latest ruby (and use the advice here about Xcode and command line tools), then install jekyll as documented". Some things I don't know, and wish I did, are exactly what difference sudo makes in the documented commands that suggest using it -- does it only make some writes succeed rather than fail, or does it change in which directories those writes are attempted?)

Issue #3984, opened 2015-09-21
gem install jekyll failed on Mac OS X 10.11


chmaynard:
$ gem install jekyll
ERROR:  While executing gem ... (Gem::FilePermissionError)
    You don't have write permissions for the /Library/Ruby/Gems/2.0.0 directory.

Mac OS X 10.11 (15A282a)
Xcode 7.0 (7A220)

jaybe-jekyll:
I prefer...

`$ gem install --user-install --bindir ~/bin --no-document --pre --verbose jekyll`

Then... i.e....

- `$ ~/bin/jekyll new ~/Sites/testjekyll`
- `$ ~/bin/jekyll serve --watch --verbose ~/Sites/testjekyll`

And or add `~/bin` to your `$PATH` shell variable to avoid requiring full path to `jekyll`.

Installing this way installs for and within the user account [only], instead of system-wide.

envygeeks:
This is not a Jekyll bug, please ask your question at https://talk.jekyllrb.com as this medium is for bugs.

chmaynard:
@jaybe-jekyll Brilliant, thanks.

Successfully installed jekyll-3.0.0.pre.beta8
17 gems installed

FYI, I think Apple has locked down the system directories in OS X 10.11, so the easy-install instructions on http://jekyllrb.com may need to change. Even using sudo won't work.

chmaynard:
@envygeeks Is there a process for reporting documentation bugs?

envygeeks:
What do you think is a bug @chmaynard?

chmaynard:
The easy-install instructions for jekyll don't seem to work in OS X 10.11 (El Capitan). Apple plans to release this new version of OS X in a few weeks, so you folks might want to investigate.

envygeeks:
If you are using RVM or Rbenv to do the install ( -- they installed ruby and rubygems -- ) via non-root userland you'll need to file a bug with them, if you are using system Ruby, it never worked the way you expected it to because you need to `sudo` to do system installs.  If it did work before that then that is a glaring security bug for OS X.

envygeeks:
We also hint at t he need to sudo in the troubleshooting: http://jekyllrb.com/docs/troubleshooting/#installation-problems it could use a reworking though and if I recall somebody submitted or is going to submit the changes.

chmaynard:
[...]

PR #3999, opened 2015-09-27
Instructions for El Capitan

chrisfinazzo:


site/_docs/troubleshooting.md
```diff
@@ -60,12 +60,43 @@ which may allow you to install native gems using this command (again using
...
+that were previously writable are now considered system locations and are no
+longer available. As a result, it is recommended that you choose one of a
+number of available Ruby environments ([RVM][], [rbenv][], [chruby][], etc.) in
+which to install Jekyll.
+
+[RVM]: https://rvm.io
+[rbenv]: http://rbenv.org
+[chruby]: https://github.com/postmodern/chruby
+
+If you elect to use a method other than Homebrew to install Ruby, it may be
```
parkr: Home brew seems to come out of the blue here. Is it mentioned above? I'm on mobile and can't check easily.

Additionally, we could say "other than Homebrew to install Ruby, such as those listed above, ..." To clarify that they are ruby managers

parkr:
In general, :+1:!

site/_docs/troubleshooting.md
chrisfinazzo: I realized almost as soon as I wrote the words that something just didn't sound right. Homebrew is mentioned without really any introduction or explanation at all. My thinking came out of @chmaynard's [comment][] and a follow up I added. While I think most people who get this far likely would understand it, I agree it's really not supported well and needs to be rephrased.

To your second point, that seems like a much clearer transition - I'll give it some thought and see how this can be incorporated...

[comment]: https://github.com/jekyll/jekyll/issues/3984#issuecomment-143149777

alfredxing:
:+1:
&nbsp;
> it is recommended that you choose one of a number of Ruby version managers

Though my personal experience of using Ruby on OS X (I'm running El Cap on one of my machines) is that using a Ruby manager just isn't worth the hassle for most people. There are some people who need multiple Ruby versions but for someone who only needs one version at a time, Homebrew works great. It's always nice to get Ruby updates when you run the occasional `brew upgrade`, rather than needing a whole separate Ruby manager.

Just my opinion!

chmaynard:
I have been following this thread for the last week, and I agree that installing into ~/bin is a good approach. However, I still strongly recommend using the Homebrew package manager. IMHO, The typical Jekyll user should or will be using Homebrew anyway, especially if there are several accounts on the same machine.

envygeeks:
From a systems standpoint that logic is broken, if you have multiple users on the same computer using the same software it would be more prudent to do a system wide install and have them install gems in their home directory and if jekyll is a common denominator then install Jekyll system wide too, but either way we are not on about logical systems management anyways we are on about OS X and their changed systems protection mechanisms that go a bit wild.

parkr:
Let's do this:

Simply recommend Homebrew and give the copy-and-paste command to install and setup Homebrew and Ruby via Homebrew. Mention for "advanced users" that Ruby Version Managers are also acceptable. Ensure `gem install` can be run _without_ `sudo`.

chmaynard:
Here's my personal checklist for installing Jekyll on OS X:

- [ ] Install Xcode command line utilities `xcode-select --install`
- [ ] Install Homebrew (instructions at http://brew.sh)
- [ ] Modify $PATH to use Homebrew `export PATH=/usr/local/bin:$PATH`
- [ ] Install the latest Ruby `brew install ruby`
- [ ] Install the latest Jekyll `gem install jekyll`

chrisfinazzo:
@parkr, OK, will update.

When finished, is it preferable that I squash this into 1 commit for the merge?

parkr:
> When finished, is it preferable that I squash this into 1 commit for the merge?

If you're comfortable with that. I'm ok merging commits that have detailed/useful commit messages and encapsulate useful changes.

envygeeks:
True that, the only time I think a squash needs to happen is if there are simple corrections, otherwise additions can be their own commits to help people see evolution.

connorshea:
+1 especially now that El Capitan has been released to the public.

r3id:
So does this mean that it's no longer a simple install process like under Yosemite?

envygeeks:
@r3id it seems that is the case because of the new over zealous system protections.

chrisfinazzo:
@envygeeks Overzealous? That's a bit harsh don't you think? Was it ever a good idea or reasonable security practice to allow an entity other than Apple to modify system files, inject code into another process, or attach unknown runtimes?

The answer to all of these questions is unequivocally no.

envygeeks:
@chrisfinazzo yes:

1. It's my system, I will do as I please, when I'm pleased to do it.
2. It's my system, I will do as I please, when I'm pleased to do it.
3. It's my system, I will do as I please, when I'm pleased to do it.
4. It's my system, I will do as I please, when I'm pleased to do it.
5. It's my system, I will do as I please, when I'm pleased to do it.
6. dylib, heard of it?  I can still inject code if you don't protect against it yourself.
7. There ways to stop process attachment without going overzealous and preventing advanced users from doing what they want. The Linux kernel does it, and has for at least as long as I can remember (as far as like 8 years ago is when I first encountered the ability to prevent unpriviledged unowned process attachment.)  I can sudo into root and attach to what I want and that's what should be allowed.  If it's not, it's overzealous.  I don't know what I said has to do with this anyways because this is probably still possible with sudo.

If none of those can be worked around for advanced users, it's overzealous, it's limiting and it's not very fostering to what computers initially existed for in the first place and it's a slap in the face.  I will say though I'm not convinced you can't sudo out of it, I've not played with it yet to see what the state of their new system protection is but I do know that unless they do some serious hardening a few of the things you listed will still be possible within that users land unless they remove core pieces of the languages they rely on.

parkr:
Personal feelings about systems aside, this PR is looking pretty good.

The process is still the same once a non-system Ruby is installed on your El Capitan machine.

envygeeks:
I'm :+1:

chrisfinazzo:
Woah.

Although this is resolved now, I should clarify. To your 3 categories of complaints:

* Fine. It's your machine, go to town. Just be careful out there.

* I have. This was [fixed][] in August as part of the 10.10.5 update and (I assume) rolled into 10.11. If it should pop up again, at the very least it is a known quantity. There is [another][] recent exploit which messes with privilege escalation, so they aren't perfect, but at these these issues aren't being ignored.

[fixed]: http://www.macrumors.com/2015/08/13/os-x-10-10-5-release-dyld-fix/

[another]: http://arstechnica.com/security/2015/09/drop-dead-simple-exploit-completely-bypasses-macs-malware-gatekeeper/

* To your last point, there'a pretty [deep dive][] into things as part of Ars' El Capitan review which should address most of those concerns. The basic gist of it is that there is no GUI for this and you have to boot the Recovery OS to get a terminal which will accept the relevant commands - including one which will disable the feature if you so desire.

[deep dive]: http://arstechnica.com/apple/2015/09/os-x-10-11-el-capitan-the-ars-technica-review/8/#h1

chmaynard:
From p. 6 of Apple's *System Integrity Protection Guide*:

> Developers using Perl, Python, Ruby, or any other scripting languages that ship with OS X, are encouraged to manage their own installations of the language and dependencies in ```/usr/local/```.

In other words, Apple is locking down ```/usr``` and encouraging users to install into ```/usr/local/```. Hopefully, a clean installation of El Capitan contains ```/usr/local/```. If not, Apple screwed up.

chrisfinazzo:
@chmaynard,

I've read through the guide and unfortunately I can't confirm whether or not `/usr/local` is present if you're coming from 10.10 (haven't upgraded yet). However, it seems that Apple is subtly encouraging people to do one of two things.

You may:

1. Use Homebrew to get `/usr/local`. The last time I did a setup (on 10.9), the install script still needed to do `chown` on this location, but with SIP in play, it seems like this is acceptable behavior and Apple won't mess with it in future updates. Never mind that the README practically [begs][] you to install the standard way...

[begs]: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/FAQ.md#why-does-homebrew-insist-i-install-to-usrlocal-with-such-vehemence

2. Create the structure manually if it doesn't exist already. It's a known convention on UNIX and elsewhere, so there shouldn't be issues with doing this.

chmaynard:
@chrisfinazzo In case you haven't seen it, here's a [Homebrew article on El Capitan][1]. Not sure why Homebrew decided to recommend ```chown /usr/local``` instead of simply changing the permissions. 

[1]: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/El_Capitan_and_Homebrew.md

chrisfinazzo:
Yeah, I saw that too, good to know that the maintainers already have a procedure for this. Reading it again makes me think that Option 2 might not work even if it is a known location - which is kind of a shame for people who like to do things using `make` where you might prefix `/usr/local` as the destination.

`chmod` is really the more appropriate choice, guess I just forgot...like I said, it's been a while since I did this last.

DomT4:
> Not sure why Homebrew decided to recommend chown /usr/local instead of simply changing the permissions.

Homebrew doesn't mandate `sudo` and the permissions on OS X El Cap for `/usr/local` seem to be set as `root:wheel` in line with the rest of `/usr`, albeit with the caveat that users can actually change the permissions on `/usr/local`.

envygeeks:
@DomT4 `/usr/local` is supposed to be owned by root because it's in the default path and changing the permissions or owner could put your system at risk because then anybody who has direct write to that folder can shiv system binaries with bad ones easily.  That to me seems like a security risk.

DomT4:
Most Macs are also single-user with that user having `admin` rights, which somewhat takes the edge off an argument around "Gaining access to root could be a disaster", certainly *prior* to El Cap. Unless you're using OS X 10.11 there's also at least two exploits out there that'll give you full root access with little effort, as @chrisfinazzo references.

Apple really hasn't pushed privileged user separation in the same way Linux has encouraged, and SIP is almost the first big acceptance that people running as the admin user constantly is an issue, but we're barely into the SIP timetable and documentation is minimal. Trying to read Apple's mind on the direction of the changes going forwards is a fairly fruitless endeavour until we actually get to drive down this road a bit.

Homebrew doesn't recommend installing it there if the system is multi-user, and as ever people get to pick their prefixes by just `git clone`-ing it somewhere else. Homebrew doesn't even block the user from running the entire party as root, if they really want to. There may be a default, but people aren't bound by it.

PR #4000, opened 2015-09-28
Empty permalink now raises an error

### LINKED ISSUE #3936: Empty string permalink make a stack overflow
Hi, I just discovered that when I create a file with an empty permalink like
```
---
permalink: ""
---
```
Jekyll fails to create the page and tell me : 
```
/Library/Ruby/Gems/2.0.0/gems/jekyll-2.5.3/lib/jekyll/cleaner.rb:59: stack level too deep (SystemStackError)
```
I'm actually on OS X (I don't know if it's MacOS specific bug).

If you want more informations, just ask :)

pheuko:
Error messages for empty permalinks. Fix #3936.

I created a Utils.permalink method to check the permalink for Documents, Pages and Posts.

Please check if it is a good solution.

### DIFF
--- a/lib/jekyll/convertible.rb
+++ b/lib/jekyll/convertible.rb
@@ -61,7 +61,7 @@ module Jekyll
         Jekyll.logger.abort_with "Fatal:", "Invalid YAML front matter in #{File.join(base, name)}"
       end
 
-      unless valid_permalink?(self.data['permalink'])
+      if self.data['permalink'] && empty_permalink?(self.data['permalink'])
         Jekyll.logger.error "Error:", "Invalid permalink in #{File.join(base, name)}"
       end
 
@@ -275,8 +275,8 @@ module Jekyll
     # permalink - the data permalink
     #
     # Returns true if the permalink is valid, false if otherwise
-    def valid_permalink?(permalink)
-      permalink.nil? || permalink.length > 0
+    def empty_permalink?(permalink)
+      permalink.length == 0
     end
 
     # Write the generated page file to the destination directory.
--- a/lib/jekyll/convertible.rb
+++ b/lib/jekyll/convertible.rb
@@ -61,6 +61,10 @@ module Jekyll
         Jekyll.logger.abort_with "Fatal:", "Invalid YAML front matter in #{File.join(base, name)}"
       end
 
+      unless valid_permalink?(self.data['permalink'])
+        Jekyll.logger.error "Error:", "Invalid permalink in #{File.join(base, name)}"
+      end
+
       self.data
     end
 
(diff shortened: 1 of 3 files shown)


parkr:
Congrats on being our 4,000th issue or pull request! :tada: 

I'm not a huge fan of `Utils.permalink` from a naming perspective. Let me think on this some.

alfredxing:
4000! Woohoo! :tada: 

Maybe `validate_permalink`?

envygeeks:
`valid_permalink?` because it's not a ActiveModel action.

pheuko:
What about a Data module? Which could be included in Documents, Pages and Posts.

```ruby
module Jekyll
  module Data
    def permalink
      return data['permalink'] if data && valid_permalink?(data['permalink'])
    end

    def valid_permalink?(input, msg = 'Input does not have a valid permalink.')
      return true if input.nil? || input.length > 0
      fail Errors::FatalException.new, "Invalid permalink '#{input}': " + msg
    end
  end
end
```

envygeeks:
I don't know if we need another module.  I'm on the fence about that personally.

pheuko:
I just updated the PR with a simple solution and a better method name as suggested.

parkr:
I wouldn't want a function that ends in a `?` to throw an error – it usually denotes returning some BOOL value. Maybe `validate_permalink`? I don't think we want a separate module for this – I'm not sure how I feel about sharing code yet. Just duplicate in Page, Post and Document where needed, and use the Utils function.

envygeeks:
You are right, I didn't even think about that,  it should probably be "validate_permalink!" which will either return `nil` or throw.

pheuko:
I already updated this PR entirely a few days ago with a much cleaner solution, and used the ```validate_permalink?``` method returning a boolean as expected. I think you didn't see that.

envygeeks:
Well I think this is ready to :ship:

lib/jekyll/convertible.rb
```diff
@@ -61,6 +61,10 @@ def read_yaml(base, name, opts = {})
         Jekyll.logger.abort_with "Fatal:", "Invalid YAML front matter in #{File.join(base, name)}"
       end
 
+      unless valid_permalink?(self.data['permalink'])
```
parkr: Check to make sure the permalink key was given.

parkr: And please add a test to ensure this output is not written if the user did not specify a permalink in the front matter.

pheuko:
Sorry, I forgot about it. But I think I did it now. Please check if it is ok, thanks.

parkr:
@pheuko I checked it and decided to throw an error and did some other restructuring too in #4361. Your commits are still in that branch, have no fear! :smile:

pheuko:
Thank you for your help @parkr, I appreciate it.

PR #4004, opened 2015-10-01
Whole-post excerpts should match the post content

kevinoid:
When a post does not contain an `excerpt_separator`, meaning the excerpt includes the entire post, the excerpt should contain exactly the post content.  This is desirable both from a correctness standpoint, that the excerpt should not introduce any new content, and more practically to allow fast and easy detection of whole-post excerpts in Liquid templates using `post.excerpt == post.content`.  A common use-case is deciding whether to render "Read More" links to a post on a page containing post excerpts.

Currently whole-post excerpts include additional newlines at the end of the excerpt content, which makes it difficult to detect whether an excerpt includes the whole post (unless there is an easy method I am overlooking).  This PR fixes the issue and adds tests to ensure it stays fixed.

If there's anything I've overlooked or style changes that you'd like to see, let me know.

Thanks,
Kevin

### DIFF
--- a/features/post_excerpts.feature
+++ b/features/post_excerpts.feature
@@ -47,4 +47,4 @@ Feature: Post excerpts
     And the _site/2007/12/31 directory should exist
     And the "_site/2007/12/31/entry1.html" file should exist
     And I should see "<p>content for entry1.</p>" in "_site/index.html"
-    And I should see "<html><head></head><body><p>content for entry1.</p>\n\n</body></html>" in "_site/2007/12/31/entry1.html"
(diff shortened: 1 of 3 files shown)


parkr:
Seems logical to me. Any objections, @jekyll/core?

envygeeks:
I'm :+1: there was only once I wished this was the case but it was a wish none-the-less.

parkr:
Cool. :shipit:

mattr-:
I'm good with this.

kevinoid:
Thanks guys, much appreciated!

PR #4013, opened 2015-10-08
update plugins documentation

antonk52:
you have to install gem files for plugins for them in order to work

### DIFF
--- a/site/_docs/plugins.md
+++ b/site/_docs/plugins.md
@@ -28,19 +28,26 @@ You have 3 options for installing plugins:
 1. In your site source root, make a `_plugins` directory. Place your plugins
 here. Any file ending in `*.rb` inside this directory will be loaded before
 Jekyll generates your site.
-2. In your `_config.yml` file, add a new array with the key `gems` and the
+2. * In your `_config.yml` file, add a new array with the key `gems` and the
 values of the gem names of the plugins you'd like to use. An example:
 
+        ```rb
         gems: [jekyll-test-plugin, jekyll-jsonify, jekyll-assets]
         # This will require each of these gems automatically.
-3. Add the relevant plugins to a Bundler group in your `Gemfile`. An
+        ```
+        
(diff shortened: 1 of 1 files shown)


envygeeks:
I think that would be nicer as point 3, would you mind making it it's own point on the list rather than an addition onto point 2?

antonk52:
perhaps it may make more sense as a sub point to point 2, as it is not a separate way to use a plugin you want to install. To use a plugin you have to install a gem and specify the name in the config. Therefore, it it should be in one point

envygeeks:
Let me rephrase what I said: It's it's own step.

antonk52:
Maybe I did not phrase it write, as jekyll offers you 3 options how to install a plugin, installing a gem will be a part of option 2, isn't it? This is why it should not be a separate point

envygeeks:
I think we are both lost in translation here, I did not fully read that page and it seems we both missed point 3.  Since this applies to both point 2 and point 3 it should be a sub-note outside of all those notes (right outside of it.) I do apologize for not reading the entire page before commenting.

envygeeks:
Let me elaborate, instead of putting it on 2 or 3, put it under all of the points as a ***note*** so that people see it at the bottom since it would be pretty hard not to grep that when reading the page.

antonk52:
It is okay, communication can be difficult at times. I have updated my patch by taking the note outside the list of options.

parkr:
I think this block is confusing in general, because it says they are ways of _installing_ jekyll plugins, though none of them actually do that. They're more a way of enabling them in your site. In point 1, no `gem install` is required, point 2 required `gem install`, and point 3 requires `bundle install`.

antonk52:
I agree on the confusing part, I have edited my patch.

parkr:
Can you run `bundle exec rake site:preview` and fix the source until this looks right?

screen shot 2015-10-08 at 3 33 34 pm

antonk52:
The latest edit is what you meant, right?

PR #4018, opened 2015-10-09
added alternative jekyll gem installation instructions

mudpuddle:
I added the solution I used to install Jekyll after updating to El Capitan.  Brew isn't always my first choice...  This solution redirects the install to a folder that you do have access to after the new security updates in El Capitan.

gem install -n /usr/local/bin jekyll

I was not able to test this on a clean install of El Capitan.

### DIFF
--- a/site/_docs/troubleshooting.md
+++ b/site/_docs/troubleshooting.md
@@ -79,8 +79,15 @@ Line Tools.
 
 With the introduction of System Integrity Protection, several directories
 that were previously writable are now considered system locations and are no
-longer available. Given these changes, the simplest way to get up and running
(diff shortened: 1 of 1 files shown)


envygeeks:
I'm :+1:

parkr:
Do we want this to be the primary way we suggest things? This puts Homebrew as option 2, so most folks will not use Homebrew. I would be apt to think Homebrew is the solution less prone to error but because I don't think I'll be checking out El Capitan soon, I am willing to see what you think.

envygeeks:
I'm going to defer to you guys and gals on that because I'm a Linux user.

PR #4021, opened 2015-10-11
Avoid "mismatched indentations" warning for ruby -w

stomar:
For those who like to do tests with `warning = true` or `ruby -w`.


parkr:
:+1:

PR #4029, opened 2015-10-14
Fix pretty permalink example

MonsieurV:
Following what the documentation specifies above for the pretty permalink format (`/:categories/:year/:month/:day/:title/`), it should result for the example to `/2009/04/29/slap-chop/` and not `/2009/04/29/slap-chop/index.html`. Well, at least if I've understood correctly ;-)


envygeeks:
The file exists so technically it's right, but given your need to be highly technical even your example is wrong because it would be a directory and you removed the trailing slash.  I'm neutral (but :+1:) on this.

parkr:
The example points to the file, not to the permalink, so :+1:

Issue #4011, opened 2015-10-05
Jekyll 2.5.3 not building on Heroku


jcornelius:
**NOTE: This is a different but perhaps related to #3907**

Using Jekyll 2.5.3 on Heroku with Github deployment. Jekyll builds fine locally on OS X, but fails on Heroku. It appears there is an issue resolving the path to files, but honestly I'm not educated enough on these things to know for sure.

Repo here: https://github.com/ninelabs/ninelabs.com

[config.ru](https://github.com/ninelabs/ninelabs.com/blob/new-content-structure/config.ru)
```
require 'rack/jekyll'
require 'yaml'
run Rack::Jekyll.new
```
[Gemfile](https://github.com/ninelabs/ninelabs.com/blob/new-content-structure/Gemfile)
```
source 'https://rubygems.org'

gem 'bundler'
gem 'jekyll', '~> 2.5'
gem 'jekyll-redirect-from', '~> 0.8'
gem 'html-proofer'
gem 'rack-jekyll'
gem 'compass'
```
_Note the current Gemfile is specifying Jekyll 1.5 since that version builds fine on Heroku._

[app.json](https://github.com/ninelabs/ninelabs.com/blob/new-content-structure/app.json)
```
{
  "name":"ninelabs.com",
  "scripts":{},
  "env":{
    "LANG":{
      "required":true
    },
    "RACK_ENV":{
      "required":true
    }
  },
  "addons":[]
}
```
----
The log showing errors is below: Note the line 
`No such file or directory - /apple-icon-72x72.png (Errno::ENOENT)`

The file referenced on this line changes from time to time. It always seems to be one of the apple-icon files or sometimes it's the 'approach' folder.

Output of `heroku logs -t --app ninelabs-staging`
```
> 2015-10-04T16:41:32.741506+00:00 heroku[web.1]: Starting process with command `bundle exec rackup config.ru -p 13959`
> 2015-10-04T16:41:37.929983+00:00 app[web.1]: /app/vendor/bundle/ruby/2.0.0/gems/jekyll-2.5.3/lib/jekyll/utils.rb:102:in `initialize': No such file or directory - /apple-icon-72x72.png (Errno::ENOENT)
> 2015-10-04T16:41:37.930023+00:00 app[web.1]:  from /app/vendor/bundle/ruby/2.0.0/gems/jekyll-2.5.3/lib/jekyll/utils.rb:102:in `open'
[...]

PR #4032, opened 2015-10-16
Make a constant for the regex to find hidden files

mattr-:
A raw regular expression isn't very expressive, IMHO. Rather than having people who read this code parse the regular expression to figure out what it's for, let's give a name. This way, it becomes more obvious what exactly it is we're doing here.

This was something that I saw when I was working on #4011 that I didn't want to change in that PR but did want to come back to and change later.


parkr:
LGTM! :shipit:

PR #4033, opened 2015-10-16
Ignore `.bundle` when generating test coverage

mattr-:
Fix the code coverage reporting when using `.bundle` to store my gems in by having SimpleCov ignore that directory. Use of `.bundle` to store my gems consolidates things since since that directory also holds the bundler config file. It also keeps a `vendor` directory out of the project tree for non-Rails projects.  Simplecov was not ignoring that directory though, which meant that the code coverage numbers I were seeing locally were wrong (and very frightening). With this change, all is right with the world once again. :smiley:


envygeeks:
I forgot to :heart: thank you when I merged, silly merge script.

PR #4034, opened 2015-10-16
Update usage.md re: _config.yml not reloaded during regeneration

jaybe-jekyll:
_config.yml not reloaded during automatic regeneration

### DIFF
--- a/site/_docs/usage.md
+++ b/site/_docs/usage.md
@@ -22,6 +22,18 @@ $ jekyll build --watch
 #    watched for changes, and regenerated automatically.
 {% endhighlight %}
 
+<div class="note info">
+  <h5>Changes to _config.yml are not included during automatic regeneration.</h5>
+  <p>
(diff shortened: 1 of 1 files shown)


envygeeks:
Any reason for the brackets around `_config.yml`?

jaybe-jekyll:
Purpose was to call out as special; i.e. code, file, path, etc. I modeled styling off of block below; I see now the brackets are being used for variables and directives and not files.

To style or not to style; that is the question. Let me know your thoughts and I'll adjust.

envygeeks:
I'm not against it so I'll merge and if there are complaints we can go back and correct it.

jaybe-jekyll:
Looking at where and how `_config.yml` has been referenced throughout the documentation, it is referenced as:

```html
<code>_config.yml</code>
```

I would prefer to adjust and recommit with consistent style and approach.

envygeeks:
If you wish you can submit another pull and we can merge it without a history mark, otherwise I'll correct it later this afternoon after I get done working on some stuff.

Issue #2604, opened 2014-07-15
Liquid Exception: No header received back - Mac OS X


jhulst:
Trying to get started in Jekyll and ran

`jekyll new test`
and then
`jekyll serve`

as directed in the startup documentation results in

```
Configuration file: /Users/jhulst/test/_config.yml
            Source: /Users/jhulst/test
       Destination: /Users/jhulst/test/_site
      Generating...
  Liquid Exception: No header received back. in _posts/2014-07-15-welcome-to-jekyll.markdown
jekyll 2.1.1 | Error:  No header received back.
```

Ran a Gem system update and also updated all packages, same result. I'm just getting started in Jekyll, so I'm sure I'm doing something wrong.

Rafeh01:
Thank you for your comments @jhulst. That helped. I simply created an environment in my anaconda for python 2.7 and used that to host on jekyll using ``jekyll serve``.

gerrywastaken:
@parkr Shouldn't these issues be handled in a more graceful way? Such as catching the "Failed to get header" exception and then outputting a message that tells user that Python 3 isn't supported and that they should switch to rogue?

This issue had so many comments that the thread was locked:
https://github.com/jekyll/jekyll/issues/1181

This is only going to get worse as more people migrate to Python 3

DirtyF:
@gerrywastaken  or upgrade to Ubuntu 16.04 who comes with Python 3 installed by default.

parkr:
> @parkr Shouldn't these issues be handled in a more graceful way? Such as catching the "Failed to get header" exception and then outputting a message that tells user that Python 3 isn't supported and that they should switch to rouge?

@gerrywastaken If you'll submit a PR, I'll review it!

envygeeks:
@DirtyF that's wrong.  Ubuntu 16.04 comes with Python2 and Python3.

envygeeks:
Let me rephrase that, Ubuntu standard comes with Python2 and Python3 by default.  And Ubuntu server *still* comes with Python2, as our fresh installs of 16.04 still have Python2 on the manifest.  I know they slated it for being Python3 only but apparently that didn't work out.

```
docker@gcec281:~# dpkg --get-selections | grep python2
libpython2.7:amd64				install
libpython2.7-minimal:amd64			install
libpython2.7-stdlib:amd64			install
python2.7					install
python2.7-minimal				install
```

That server was recently redone with fresh Ubuntu since we ops that way.

[...]

PR #4036, opened 2015-10-17
Clarify Python version prerequisite

JeNeSuisPasDave:
Python 2.7 with the pypgments package is required. Python 3.* is
not currently compatible with Jekyll.

See issue https://github.com/jekyll/jekyll/issues/2604


parkr:
Python 2.7 is all that's required. The pygments package should be shipped as a part of pygments.rb, a gem.

JeNeSuisPasDave:
Withdrawing incorrect doc change.

I probably won't submit a second pull request, but it would be helpful if the docs say "python 2.7" is required instead of just "python".

parkr:
@JeNeSuisPasDave Ok! Sorry to hear you won't be submitting a second pull request. I admit, Jekyll is not straight-forward in many respects, and we're trying to fix that.

Issue #4041, opened 2015-10-17
jekyll docs fails in 2.5.3


JeNeSuisPasDave:
The online documentation says:

> If you’re interested in browsing these docs on-the-go, install the `jekyll-docs` gem and run `jekyll docs` in your terminal.

However, doing so results in a _file not found error_ ... it seems that jekyll-docs is looking for the generated site to be present in the jekyll gem installation, something that apparently stopped being generated or save with jekyll version 2.5. See [issue #3](https://github.com/jekyll/jekyll-docs/issues/3) in the jekyll-docs project.

This seems like some that a CI test could catch ...

envygeeks:
It should be caught but not by Jekyll, we don't provide that command, it's provided by: https://github.com/jekyll/jekyll-docs can you move this ticket there.

envygeeks:
Sorry, what I mean is (we) `jekyll/jekyll` don't provide that command, it's provided by `jekyll/jekyll-docs`

JeNeSuisPasDave:
OK, well there's already an issue posted there. So I'll close this.

PR #4043, opened 2015-10-17
Use jekyll-docs gem for 'jekyll docs' command

parkr:
/cc @jekyll/core 

alternatively, just add `site` back into the `.gem` package in the gemspec. thoughts?

/cc #4041


envygeeks:
Personally I would prefer the gem be stubbed to tell them to install `jekyll-docs` the same way that Pry does it, rather than including the entirety of docs inside of install.  Now I might be seeing stuff but according to what RubyGems is saying jekyll-docs 3.0.0.beta.1 is 500KB, this impacts our Docker images too.

envygeeks:
To further add, according to fileroller it's gonna expand to about 1.9MB (this could be off.)

mrmanc:
I think stubbing it with a hint sounds like a really intuitive way to hint to the developer that it is available and how to proceed, and trading lean gems off against having the docs there when you need them seems a fair choice to me as I imagine this is a minority use case.

Thanks so much for helping with this @parkr 😀

PR #4046, opened 2015-10-19
Apply the highlight class

chrisfinazzo:
See comments in #3940.

### DIFF
--- a/site/_docs/contributing.md
+++ b/site/_docs/contributing.md
@@ -45,20 +45,20 @@ Test Dependencies
 To run the test suite and build the gem you'll need to install Jekyll's
 dependencies. Simply run this command to get all setup:
 
-    $ script/bootstrap
+<figure class="highlight"><pre><code>$ script/bootstrap</code></pre></figure>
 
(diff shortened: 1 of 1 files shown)


envygeeks:
I don't know how hard this would be to regression proof.  That page is updated from our actual CONTRIBUTING.md in the repo and sometimes just copied right over, so it has a very high chance of regressing... and I don't remember if @parkr even still tests the site so we could prevent a regression there.

envygeeks:
OH: I was the one who requested that, well I'll stick to what I requested.

parkr:
What about backtick code blocks? If those work on the site, then they'll be cross-compatible.

chrisfinazzo:
Either way you guys would like is fine. The other style (say on the [Posts][] page) is visually distinct, but does seem "heavy" given that there are multiple lines where it would be applied. Examples are as follows: 

[Posts]: http://jekyllrb.com/docs/posts/

Backticks: 

screen shot 2015-10-18 at 11 22 54 pm

CSS w/ the "highlight" class: 

screen shot 2015-10-18 at 11 30 49 pm

PR #4049, opened 2015-10-20
Icon refactor

BigBlueHat:
While working on https://github.com/hypothesis/developers.hypothes.is I wanted to use the GitHub icons inline in some content. This is essentially the same refactoring I did there, but on the default Jekyll theme--in hopes that it helps others.

The footer has been refactored to use these new icon includes, and `about.md` shows an example of using them inline.

Example:
```
{% include icon-github.html username="bigbluehat" %}
```

These could also be handy on author pages and the like.

Hope it's useful!
:tophat:


envygeeks:
I am absolutely :+1: on this because it demonstrates to a further extend how some of our software works and it also makes it cleaner to read the source.  :heart:

lib/site_template/about.md
```diff
@@ -6,6 +6,10 @@ permalink: /about/
...
 
-You can find the source code for the Jekyll new theme at: [github.com/jglovier/jekyll-new](https://github.com/jglovier/jekyll-new)
+You can find the source code for the Jekyll new theme at:
+{% include icon-github.html username="jglovier" %} /
+[jekyll-new](https://github.com/jglovier/jekyll-new)
 
-You can find the source code for Jekyll at [github.com/jekyll/jekyll](https://github.com/jekyll/jekyll)
+You can find the source code for Jekyll at
+{% include icon-github.html username="jekyll" %} /
+[jekyll](https://github.com/jekyll/jekyll)
```
parkr: What is this addition about? Can you explain this?

parkr:
Where are the SVG's used?

lib/site_template/about.md
BigBlueHat: It's meant to be an example of in using the icons inline.

It looks like this:
---
2015-10-20 20_07_43-store

envygeeks:
I believe the only place we had them previously was in the footer? /cc @jglovier who would know more if I recall right yeah?

BigBlueHat:
@parkr @envygeeks previously they were only used in the footer. This allows them to also be used within page content--which can be pretty handy when generating author or code project pages.

See the screenshot here: https://github.com/jekyll/jekyll/pull/4049#discussion_r42570687

jglovier:
>I believe the only place we had them previously was in the footer?

Yup, that's correct.

Also, :+1: :+1: !

jglovier:
Sorry, I lied. There is also a hamburger menu bar which is used for collapsing the navigation on mobile.

gynter:
:+1: Why did You create separate SVG files for SVG content only? I'd probably recommend putting SVG data directly to the icon includes. One include less to make could mean smaller generation time for webs with hundreds of pages.

parkr:
Thanks!

envygeeks:
:heart:

BigBlueHat:
@gynter the reason for the separate `.svg` files is that they're actually `.svg` files now. :smile: They can be edited directly in [Inkscape](http://inkscape.org/) or whatever if needed. I like my files raw. :fork_and_knife: :wink: 

Thanks for the merge @parkr!

jglovier:
shaggy-guy-thumbsup

Issue #3921, opened 2015-08-18
Design updates for 3.0 for `jekyll new` theme


jglovier:
I'd like to make some design updates to the `jekyll new` theme. Mainly, I think we can take the typography to the next level, and add some more advanced styling for things like the next article section, etc.

## Typography

I'm thinking for typography to go for something more Medium-esqe and pair a nice bold sans-serif typeface for headings with a nice serif typeface for body copy. I've used Open Sans (headings) and Lora (body copy) on my own site for such typography, and it's very readable. 

One consideration is that this proposal would mean moving from default webfonts to using the Google Fonts API (still with the webfonts as a fallback). I don't have any performance concerns about this personally because a) this is intended to be a starter blog for the average person, which means high traffic performance concerns are negligible, and b) Google Fonts API is about as performant as you can get for a third party typography system (Open Sans was served [18.5 billion times](https://www.google.com/fonts/specimen/Open+Sans#stats) in just the last week, which means many uses are likely to have it cached anyway).

## Layout

- Add options for heading images

## Config

- Add an option to display article reading times

## New new stuff

- Add ability to pass the `--boilerplate` flag to get a design-agnostic, bareboness Jekyll setup (basically just basic file and directory structure and markup without any styling).

## etc

Thoughts? Other ideas?

alfredxing:
I think improving the typography and making the theme a bit cleaner is a good idea.

For a sans-serif heading, I'd recommend Roboto, Open Sans, and [League Spartan](https://www.theleagueofmoveabletype.com/league-spartan). Raleway is good too.
For a readable body font, Libre Caslon Text, Libre Baskerville, and Source Serif Pro.

I'm kind of iffy on adding things like header images and reading times, since those aren't necessarily core features that all of our bloggers would want and use (better left for plugins/themes?)

I'm also a teensy bit wary because you bring up Medium in particular. I think its design is great, but I think the default Jekyll theme needs to maintain its own identity instead of adopting the nowadays quite generic (and I think you phrased this very well) Medium-esque style. It's always nice running into a blog using the current default theme and being able to recognize it immediately.

Just some of my thoughts! ¯\_(ツ)_/¯

envygeeks:
I was cool up until you suggesting using Google fonts URL by default.  Then I jumped off the ship.

alfredxing:
@envygeeks Anything in particular against Google Fonts?

envygeeks:
[...]

PR #4050, opened 2015-10-22
Change default font weight to 400 to fix bold/strong text issues

### LINKED ISSUE #4042: Jekyll's default font-weight of 300 causes bold/strong text to be indistinguishable from regular text on IE11 and Firefox
Specs:

- OS: Windows 8.1 64-bit
- Jekyll version: 2.5.3 ([the current main.scss](https://github.com/jekyll/jekyll/blob/master/lib/site_template/css/main.scss#L11) is also using the same font-weight of 300).
- Browsers affected: IE11, Firefox 41, Palemoon 25

By "default", I mean the CSS used by the site when you run `jekyll new` 

The issue is that the default browser stylesheet in IE11, Firefox 41, Palemoon 25 (probably other versions too but these are the ones I have) defines the `font-weight` property of `b` and `strong` as `bolder`. So if the inherited font-weight is 300, that means `b` and `strong` will have a font-weight of 400. (c.f. ["Meaning of relative weights"](https://developer.mozilla.org/en/docs/Web/CSS/font-weight#Meaning_of_relative_weights)). Fonts like Arial doesn't show any visual difference between font-weights 300 and 400.

If inherited font-weight is 400, then `bolder` would be 700.

Chrome is unaffected because its default stylesheet defines `b` and `strong` with `font-weight: bold`.

I think there are 2 approaches to resolve this issue

- change default font-weight to 400 in [main.scss](https://github.com/jekyll/jekyll/blob/master/lib/site_template/css/main.scss#L11)
- define the rule `font-weight: bold` for `b` and `strong` in [_base.scss](https://github.com/jekyll/jekyll/blob/master/lib/site_template/_sass/_base.scss)

alfredxing:
Change default font weight in site template to 400 (normal), except for the site title, which remains at 300.
Should fix #4042.

Not sure if we're still going for #3921 though.

/cc @jekyll/core

### DIFF
--- a/lib/site_template/_sass/_layout.scss
+++ b/lib/site_template/_sass/_layout.scss
@@ -12,6 +12,7 @@
 
 .site-title {
     font-size: 26px;
+    font-weight: 300;
     line-height: 56px;
     letter-spacing: -1px;
     margin-bottom: 0;
--- a/lib/site_template/css/main.scss
+++ b/lib/site_template/css/main.scss
@@ -8,7 +8,7 @@
 // Our variables
 $base-font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
 $base-font-size:   16px;
-$base-font-weight: 300;
+$base-font-weight: 400;
 $small-font-size:  $base-font-size * 0.875;
 $base-line-height: 1.5;
 


envygeeks:
I'm :+1: if we just remove the font-weight entirely.  At the point of setting it to 400 by default we can just remove it since 400 is the default across all so it will just be normal, no need to be explicit about normal when normal is normal because it's normally normal.

alfredxing:
I thought that as well, but I left it in there in case users want to change/customize theirs (perhaps back to 300 if they're not worried about the bug).

envygeeks:
That's makes sense, and in light of that sensibility I say :shipit:

envygeeks:
:hear:

envygeeks:
I can't spell heart today, and it's breaking my heart.

parkr:
Haha. :)

Issue #4030, opened 2015-10-15
If a config file is empty Jekyll stops processing later configs


sparanoid:
For example:

```shell
$ jekyll build --config _config.yml,_config.user.yml,_config.local.yml
```

_config.user.yml:

```yaml
#
# Jekyll user configurations
#

# Define your variables here
```

if `_config.user.yml` is empty, Jekyll will stop processing `_config.local.yml`.

parkr:
I'd probably change the following code in Jekyll to accept `false`: https://github.com/jekyll/jekyll/blob/8add5aec4b8490aa129a01cee4d37c215f6b6dc8/lib/jekyll/configuration.rb#L155-L158

`false` is returned when `SafeYAML.load` gets no content

PR #4052, opened 2015-10-24
Handle empty config files

Crunch09:
This is my first PR for jekyll and i tried to follow @parkr 's suggestion given in #4030.
`SafeYAML.load_file` returns `false` when processing empty files so we convert this into an empty hash for further processing. I wasn't sure if we should log a warning on empty config files, so i left it out for now.

### DIFF
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -118,7 +118,7 @@ module Jekyll
         Jekyll::External.require_with_graceful_fail('toml') unless defined?(TOML)
         TOML.load_file(filename)
       when /\.ya?ml/i
-        SafeYAML.load_file(filename)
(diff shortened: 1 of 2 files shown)


parkr:
Thanks for your pull request!

As an alternative, you could use `|| {}` in the YAML reader in `safe_load_file`. I approach you have here is a little more defensive than I think we need to be, considering YAML and TOML have fairly stable API's. What do you think about that approach?

```ruby
SafeYAML.load_file(filename) || {}
```

Crunch09:
Thanks, your approach sounds good to me :+1: i'll fix the failing tests later today

envygeeks:
Marking this for 3.1 because 3.0 is due today AFAIW.

PR #4053, opened 2015-10-25
Add missing GFM syntax highlighting background

sparanoid:


parkr:
Can you show a before and after example, please? Curious how this changes things.

sparanoid:
@parkr sorry for the lack of reproduction steps:

```sh
~ $ mkdir jekyll-rouge && cd jekyll-rouge
~/jekyll-rouge $ jekyll new
```

Add the following config in `_config.yml`:

```yaml
highlighter: rouge
kramdown:
  input: GFM
  syntax_highlighter: rouge
```

Add the following snippets in `_posts/2015-10-26-welcome-to-jekyll.markdown`

    ```ruby
    def print_hi(name)
      puts "Hi, #{name}"
    end
    print_hi('Tom')
    ```
    
    {% highlight ruby %}
    def print_hi(name)
      puts "Hi, #{name}"
    end
    print_hi('Tom')
    {% endhighlight %}

Then:

```sh
~/jekyll-rouge $ jekyll s
```

Result:

image

I've re-apply background color for `.highlight` in `.highlighter-rouge`, this should fix this issue.

envygeeks:
I'm gonna be honest, there should be no disparity between Rogue and Pygments :-1:

sparanoid:
@envygeeks just rechecked the config after I found [this issue](https://github.com/jekyll/jekyll/pull/3323), you don't need to change the config since rouge is the default highlighter, so this PR should be “Add missing GFM syntax highlighting background”

envygeeks:
That's not relevant, it wasn't there for Pygments (apparently -- according to the CSS -- and btw, Pygments is still available, the default is in no way the defacto -- at least in this case.) That means there will be a disparity.

AndorChen:
@envygeeks As far as I know, there are no much differences between Pygments and Rouge, the problem issued here is caused by kramdown, as you can see in [the source code where the `.highlighter-rouge` class added](https://github.com/gettalong/kramdown/blob/master/lib/kramdown/converter/html.rb#L98).

parkr:
Yeah, it looks like Kramdown does this. https://github.com/gettalong/kramdown/blob/7a7bd675b9d2593ad40c26fc4c77bf8407b70b42/lib/kramdown/converter/html.rb#L365-L368

I don't see a huge issue with doing this in the CSS. I wouldn't want to change our Ruby for it, but adding this class seems pretty harmless to me.

envygeeks:
@parkr, it's not about adding the class, it's that the class is only added to Rogue, unless this affects pygments too, it's a disparity and an oddity to say the least.  We should keep it uniform across both... but I didn't look at the CSS fully to even see if Pygments is supported by default in that CSS anyways.

parkr:
> but I didn't look at the CSS fully to even see if Pygments is supported by default in that CSS anyways.

Yeah, the highlighting CSS was written to the output of Pygments some time ago.

PR #4055, opened 2015-10-26
collections: posts as collection

parkr:
I guess I need to release Jekyll 3 tomorrow...

This is the last piece we were considering doing.

In an effort to not break sites, this is all backwards compatible liquid-wise. The only big difference is that `site.posts` is a `Collection` instead of an `Array`. But I feel more comfortable breaking plugins than I do breaking people's Liquid. Especially because Liquid doesn't have a means of erroring out on a bad dereference (see https://github.com/Shopify/liquid/issues/490 for more).

The cukes all pass. I'm going to work on the presentation now and leave this. But here it is.

/cc @benbalter who was looking to see this in 3.


parkr:
cc @jekyll/core for thoughts. will merge now – please leave thoughts or submit follow-up PR's in the 3.0 milestone to fix things. it's not perfect, but at least it's something

lib/jekyll/configuration.rb
```diff
@@ -275,6 +275,21 @@ def fix_common_issues
...
+
+      if config['collections'].is_a?(Array)
+        config['collections'] = Hash[config['collections'].map{|c| [c, {}]}]
+      end
+      config['collections']['posts'] ||= {}
+      config['collections']['posts']['output'] = true
+      config['collections']['posts']['permalink'] = style_to_permalink(config['permalink'])
+
+      config
+    end
```
benbalter: Nicely implemented.

lib/jekyll/document.rb
```diff
@@ -242,19 +293,54 @@ def read(opts = {})
       end
     end
 
+    def post_read
```
benbalter: Does this mean that all collections can have a date now?

lib/jekyll/document.rb
```diff
@@ -242,19 +293,54 @@ def read(opts = {})
...
           "content"       => content,
           "relative_path" => relative_path,
           "path"          => relative_path,
           "url"           => url,
-          "collection"    => collection.label
-        }
+          "collection"    => collection.label,
+          "next"          => next_doc,
+          "previous"      => previous_doc,
+          "id"            => id,
```
benbalter: All documents having IDs is going to fix so many terrible workarounds I've created. :smile:

benbalter:
Helz yes

envygeeks:
Since @benbalter posted the Wonder Years guy is this gonna be Jekyll's Wonder Years?

benbalter:
Technically, that was from FOX's new hit show [the Grinder](http://www.imdb.com/title/tt4481322/), so I say we have a 50/50 chance. :smile:

lib/jekyll/document.rb
parkr: Yes!

parkr:
Thanks for taking the time to :eyes: this, @benbalter!

envygeeks:
He will forever be Kevin Arnold to me. Always. :100:

lib/jekyll/document.rb
benbalter: Huh. :metal:

parkr: `_decisions/2015-06-29-glossip-v-gross.md`

alfredxing:
Woohoo! :beers: 

I'll take a good look at the code a bit later!

PR #4058, opened 2015-10-26
Add Document#method_missing and Collection#method_missing

parkr:
Proxies calls to either #data or #docs, respectively. Deprecation warning is printed.

screen shot 2015-10-26 at 1 37 17 pm

_Don't break, deprecate._


envygeeks:
It does the job, I like it!  Well, at least until it's ready to be put to pasture.

parkr:
Failure is jruby. Merging.

Issue #4057, opened 2015-10-26
Documentation for incremental regeneration


alfredxing:
I still need to write some documentation for incremental regeneration specifying:

- What it does and a brief overview of how it works
- Warnings, shortfalls, and where it doesn't work
- Configuration options

It's midterm season (:cry:) so this will probably have to wait for a few days until after my exams this week.

envygeeks:
I think it was agreed on (informally) to disable it until 3.1 so we can work out the kinks and bugs because you were busy, @parkr would have to backtrack and confirm but I don't assume it's important to write them until it goes on by default.

alfredxing:
That's what I was thinking as well (making it an "experimental" feature for now).

It's definitely not critical as you mentioned, but It'll still be good to have docs for users who want to try it out or enable it though.

parkr:
> I think it was agreed on (informally) to disable it until 3.1 so we can work out the kinks and bugs

Oh, it's totally on by default right now and there's no flag to turn it on. When did we decide to turn it off by default?

envygeeks:
Quite a while back (probably like 4 months ago now) when I was struggling to get it to work with some stuff and filed that big ticket with a bunch of stuff coupled into a single ticket.

envygeeks:
Oh, it looks like it was about 6 months ago now: https://github.com/jekyll/jekyll/issues/3747

envygeeks:
If I'm honest though, I'm on the fence and will digress to @alfredxing, since that ticket was filed we've found ways to work around the lot of the stuff on sites we build on our cluster, except for pages not being rendered on update of another page and sometimes pages not being rendered I think others have had that problem too as there are a few tickets about it.

alfredxing:
I'll need to re-familiarize myself with exactly how the regenerator behaves. I know that basically the only thing that isn't recorded as a dependency right now is any reference to rendered post/page content outside of that post/page, since this usually involves references to `post.content` which is not directly checkable by the regenerator.

I guess one thing (that I just thought of!) is that we can check _in the post.content accessor method_ whether or not the post is rendered, then render it on-the-fly if it isn't...

alfredxing:
But: my call is that we disable it by default for now.

This should be a simple switch in `configuration.rb` of the `full_rebuild` option&mdash;maybe rename this too?&mdash;and fixing broken Regenerator tests by re-enabling that flag for those tests.

alfredxing:
:arrow_up: PR to disable incremental by default!

[...]

PR #4059, opened 2015-10-26
Disable incremental regeneration by default in Jekyll 3.0

alfredxing:
Disable the feature as it's still not 100% working 100% of the time. Incremental regeneration can be re-enabled by specifying `full_rebuild: false` in the configuration.

(As per discussion in #4057)


parkr:
Need to update `--full-rebuild` tag to something like `--incremental` or `--incr`.

parkr:
Is there _any_ documentation on this feature? Perhaps it can say it's off by default.

alfredxing:
On it!

There's barely any documentation, though it prints out whether it's enabled when building:

```
Configuration file: /Users/alfred/git/jekyll/test/_config.yml
            Source: /Users/alfred/git/jekyll/test
       Destination: /Users/alfred/git/jekyll/test/_site
 Incremental build: disabled
      Generating...
```

parkr:
> Incremental build: disabled

Maybe we should change this to say `disabled. Enable with --incremental.`?

jaybe-jekyll:
I prefer a switch/command/trigger to enable regeneration in general regardless of current state.

Building, Serving, Regenerating... all specific and chosen capabilities.

I would expect `jekyll build` to build once and return successful.

I would expect to `jekyll build ---watch` or `jekyll build -w` to enable/turn on regeneration.

$.02 💸

envygeeks:
@jaybe-jekyll that very well could be the case when it's enabled by default but with it being disabled by default as an experimental feature now, that won't work.

parkr:
@jaybe-jekyll It's still very experimental, but that's a fantastic idea for the future! :)

alfredxing:
Agreed! Commits coming as soon as the tests pass. It'll be `incremental` in configuration and `--incremental` or `-ir` as command-line flags.

jaybe-jekyll:
I need to correct myself; apologies --

I prefer a switch/command/trigger required to enable regeneration,

Building, Serving, Regenerating... all specific and chosen capabilities.

I would expect `jekyll build` to build once and end, returning as successful.

I would expect to `jekyll build --regnerate`to enable/turn on regeneration, with or without --watch, as --regeneration would presume --watch.

Pfew.

$.02 💸

parkr:
I think `-ir` might get confusing. Maybe `--ir` or use just one character for a single dash.

alfredxing:
@parkr: Would you prefer `-i` or `-r` for the shorthand flag?

@jaybe-jekyll Not sure about the "flag presumes `--watch`" part, since incremental regen would offer a performance improvement even without watching (so users might want to enable it without enabling watch)

jaybe-jekyll:
Another opinion; I think the term, "incremental", in "incremental regeneration" (i.e. --ir) applies more to a technical audience and isn't necessarily necessary or appropriate for general use(rs).

*Regenerate.*

Agreed; single character enablement; perhaps `-R` (for, **REGENERATION POWERS ACTIVATED**). And, for the verbose and visually-minded, `--regenerate` and or `--regeneration`.

parkr:
> @parkr: Would you prefer -i or -r for the shorthand flag?

Historically, I have used `-i` for "case-insensitive", and `-r` for "recursive." I would go with the first character of the double-dashed one or an alternative if it isn't there.

jaybe-jekyll:
@alfredxing I don't see how generation could occur within a system without --watch; does it not imply watching for changes (`--watch`) then acting on those changes (`--regenerate`)?

envygeeks:
@parkr @alfredxing: @jaybe-jekyll brings up a fine point, why not call the flag `--cache`? That's literally what it's doing, but we call it incremental regeneration as a technical term, when Occam is standing over there pointing to the term: caching.

parkr:
Does the `sass` compiler have a flag like this? Our Jekyllers tend to be front end devs so maybe Sass's CLI is known.

parkr:
@jaybe-jekyll Yeah, it's more about caching. `--watch` says "hey, regenerate every time you detect a changed file!" and `--regenerate` or `--increment` or `--cache` is "hey, when you do regenerate, only regenerate the changed files!".

envygeeks:
Sass by default caches so they have a `--force` flag which forces it to update the cache even if it's cached.

parkr:
@alfredxing just waiting on this PR. once it's merged, i'll release an RC and then do any final bug fixes before I release around 8pm PST tonight.

alfredxing:
Sure! Right now I have it as `-I` and `--incremental`. If we don't think `-I` is appropriate, I can remove it!

envygeeks:
`-I` is what I instantly thought of when you said you wanted `--incremental` so I'm :+1:

parkr:
Looks great! Thanks so much, @alfredxing. I'll merge once it's :green_heart: .

alfredxing:
:+1:

Good luck (& congrats!) on the launch tonight!

parkr:
> Good luck (& congrats!) on the launch tonight!

@alfredxing Congrats to you, too! It's been a huge team effort. Thanks for all your time and effort thus far – it's meant a lot!

PR #4077, opened 2015-10-29
Fix #4066: Move Convertible#render_liquid to using render!

### LINKED ISSUE #4066: Including a symlinked file in safe mode outputs the error to the rendered document, exits 0
## Steps to reproduce

1. Create an `index.html` file with the content:
  
    ```
    ---
    ---
    
    {% include foo.html %}
    ```

2. Make `_includes/foo.html` a symlink to a file outside your source directory

3. `jekyll build --safe`

## Expected

* Exit status > 0
* Red output to STDOUT with the text `Liquid error: The included file ‘_includes/foo.html’ should exist and should not be a symlink`
* `_site/index.html` should be blank

## Actual

* Exit status of 0 
* No error output
* `cat _site/index.html` returns:

    ```
   <p>Liquid error: The included file ‘/Users/benbalter/projects/jekyll-test/_includes/foo.html’ should exist and should not be a symlink</p>
   ```

(Note the full path, which we don't use elsewhere AFAIK).

I believe this to be a regression from 2.x to 3.x.

To note, the symlink file is properly *not* included.

envygeeks:
/cc @benbalter @fw42 @jekyll/core 
https://travis-ci.org/envygeeks/jekyll-symlink-test/builds/88182863

### DIFF
--- a/features/rendering.feature
+++ b/features/rendering.feature
@@ -5,6 +5,13 @@ Feature: Rendering
   But I want to make it as simply as possible
   So render with Liquid and place in Layouts
 
+  Scenario: When receiving bad Liquid
+    Given I have a "index.html" page with layout "simple" that contains "{% include invalid.html %}"
+    And   I have a simple layout that contains "{{ content }}"
+    When  I run jekyll build
+    Then  I should get a non-zero exit-status
+    And   I should see "Liquid Exception" in the build output
+
   Scenario: Render Liquid and place in layout
     Given I have a "index.html" page with layout "simple" that contains "Hi there, Jekyll {{ jekyll.environment }}!"
     And I have a simple layout that contains "{{ content }}Ahoy, indeed!"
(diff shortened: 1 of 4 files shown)


fw42:
Thanks! :+1:

fw42:
Should probably add a regression test for this

parkr:
@fw42 great idea.

I found the errors in the templates helpful, too... But I guess terminal errors would be better because they'd break the build which is what we want when something goes wrong.

parkr:
LGTM pending the regression test.

envygeeks:
I'll add the test after I finish up with my other pull requests.

benbalter:
Cowabunga

envygeeks:
It took a while to write that test, it was crazy because I couldn't figure out why it wasn't getting output and it turned out there might be a slight bug in POSIX::Spawn::Child somewhere.

fw42:
I would have just tested that `render_liquid` doesn't swallow exceptions, but I guess this works too

envygeeks:
I considered that, but then I decided that we want the errors to bubble at the UX so tested it in Cucumber.

parkr:
:shipit:

PR #4081, opened 2015-10-30
Remove the temp variables getting `existing_files`

mattr-:
This is basically a `Replace Temp with Query` refactor with some memoization added so that we don't suffer the performance hit of calling methods over and over again.

I appreciated the speedup that @fw42 put in but the extra variables really bothered me, so I just figured I'd go back and clean up the temp variables but keep the speedup.

### DIFF
--- a/lib/jekyll/cleaner.rb
+++ b/lib/jekyll/cleaner.rb
@@ -37,11 +37,9 @@ module Jekyll
     # Returns a Set with the file paths
     def existing_files
       files = Set.new
-      regex = keep_file_regex
-      dirs = keep_dirs
 
       Dir.glob(site.in_dest_dir("**", "*"), File::FNM_DOTMATCH) do |file|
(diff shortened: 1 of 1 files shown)


fw42:
Hmm I'm not sure if this is equivalent.. What if you run this with incremental regeneration and then add a new file? Doesn't that mean that `keep_dirs` and `keep_file_regex` will now use an outdated value (with my code, that wouldn't be an issue).

fw42:
I guess what I'm saying is: My code is caching those values for the duration of the `existing_files` method. Your new code is caching those values indefinitely (for the entire lifetime of that object). That might lead to weird behaviour...

mattr-:
hmmm, good point. All the tests passed, so I didn't think anything of it. I'll take a closer look.

mattr-:
yeah, this isn't going to work. Closing.

PR #4086, opened 2015-10-30
Update Font Awesome to v4.4.0.

XhmikosR:
Mostly for .woff2 support.


parkr:
:+1:

envygeeks:
I'm hesitant on this one, double check the SHA's before merging the binary files.

XhmikosR:
Come on, guys, I'm not new on this.

parkr:
We could have a SHA check built-into the rake task if we wanted.

envygeeks:
@XhmikosR it's not that we think you are new, it's that we have to provide our own integrity so we need to verify the validity of things like that before we update them (at least that's my 0.02c) I don't know if @jekyll/core agrees.

XhmikosR:
You can verify it if you really want, but seriously, you are barking at the wrong tree here.

XhmikosR:
Just check who updated it last time too.

envygeeks:
I'm not asking you to verify, I will happily verify when I have time unless somebody else does it before me.  There is no need to be defensive, nobody is saying you are going to do anything but I am saying we need to verify that everything is in order because fonts are a vector too.

alfredxing:
@envygeeks Just verified the SHA1 hashes, matches fonts downloaded from FontAwesome.

envygeeks:
:shipit:

PR #4087, opened 2015-10-30
Trim trailing whitespace.

XhmikosR:


envygeeks:
Does this need an entry? I don't think this needs an entry? /cc @parkr

parkr:
Nope, we're good as-is.

PR #4088, opened 2015-10-30
Ability to pass `raw` as argument to `highlight`

### LINKED ISSUE #3855: Feature request: raw flag to highlight tag
The highlight tag should take a raw tag. (IMHO it should be on by default, but that's another matter.) It's annoying to wrap code block, like this:

```
{% highlight %}{% raw %}
code
{% endraw %}{% endhighlight %}
```

and it'd be nice if we could do this:

```
{% highlight raw %}
code with templates and junk in it
{% endhighlight %}
```

Maybe there's a better way to accomplish this. I mean, I'd prefer to use github-markdown's ````lang` construct

Crunch09:
It will create an implicit `raw` tag.
Example:
```
{% highlight raw %}
Use {{ post.date }} to print the date of this post
{% endhighlight %}
```

Fixes #3855

### DIFF
--- a/lib/jekyll/tags/highlight.rb
+++ b/lib/jekyll/tags/highlight.rb
@@ -9,6 +9,7 @@ module Jekyll
       #
       # <quoted list> is a space-separated list of numbers
       SYNTAX = /^([a-zA-Z0-9.+#-]+)((\s+\w+(=(\w+|"([0-9]+\s)*[0-9]+"))?)*)$/
+      END_HIGHLIGHT = /\A(.*){%\s*endhighlight\s*(.*)?%}\z/
 
       def initialize(tag_name, markup, tokens)
         super
@@ -37,6 +38,17 @@ Syntax Error in tag 'highlight' while parsing the following markup:
 Valid syntax: highlight <lang> [linenos]
 eos
         end
+      end
+
+      def parse(tokens)
+        if @lang == 'raw'
+          end_pos = tokens.rindex{|x| x =~ END_HIGHLIGHT }
+          unless end_pos.nil?
+            tokens.insert(0, "{% raw %}")
+            tokens.insert(end_pos+1, "{% endraw %}")
+          end
+        end
+        super
       end
 
       def render(context)
@@ -79,7 +91,7 @@ eos
 
         highlighted_code = Pygments.highlight(
           code,
-          :lexer   => @lang,
+          :lexer   => highlight_lang,
           :options => sanitized_opts(@options, is_safe)
         )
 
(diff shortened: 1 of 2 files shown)


parkr:
Something we have wanted for a while.

@fw42, is this a kosher way of enforcing the content of a block is not rendered?

envygeeks:
I think there are more efficient ways to do this, I can't recall right now and @fw42 would have to confirm but I used to remember that Liquid itself has a way to get the raw context without the need for all that and if it doesn't you can always walk the tree and reconstruct it again without modifying their source.

Crunch09:
@envygeeks Oh ok, this can very well be the case. I haven't done this kind of thing before.

envygeeks:
@Crunch09 the code is great and if there isn't a way to get at it easier I'm for sure :+1: on this.  So on that let me say thanks for the pull and welcome!

Crunch09:
@envygeeks No worries, thank you for maintaining jekyll!

envygeeks:
Okay, I think we should hold this until tomorrow, I've merged jekyll-assets tag parser into Jekyll finally and added the ability for all tags to get the raw source of a block without needing to do anything special on their end, we will just automatically do it for them.

parkr:
> I think there are more efficient ways to do this, I can't recall right now and @fw42 would have to confirm but I used to remember that Liquid itself has a way to get the raw context without the need for all that and if it doesn't you can always walk the tree and reconstruct it again without modifying their source.

@pushrax or @dylanahsmith might also know about this.

dylanahsmith:
I'm not familiar with this tag, but it looks like you are making it conditionally have very different semantics.  Why not have a separate `highlight_raw` tag, so you can use Liquid::Raw as a superclass, then you can wrap the render method to do the highlighting on the result of the superclass.

lib/jekyll/tags/highlight.rb
```diff
@@ -39,6 +40,17 @@ def initialize(tag_name, markup, tokens)
         end
       end
 
+      def parse(tokens)
+        if @lang == 'raw'
+          end_pos = tokens.rindex{|x| x =~ END_HIGHLIGHT }
```
dylanahsmith: This isn't going to work in liquid 4, since it passes a Liquid::Tokenizer instead of an array as the argument to `parse`.  Does jekyll do something like this elsewhere?  Normally all that is needed is to call `tokens.shift`

envygeeks: Why would you shift to get an index?  That said this is a detail that needs to be worried about when Liquid 4 is released because we just entered into a stable period so we probably won't be jumping on Liquid 4 bandwagon for a while even so and there is plenty more that will probably need to change.

dylanahsmith: > Why would you shift to get an index?

I mean normally we don't need to scan the list of tokens, we just get and consume the next token.

parkr:
I like this, but I don't think this is the best way to go about it programmatically. Perhaps @dylanahsmith is right, and we should have a `highlight` tag and a `highlight_raw` tag. Thanks!

Crunch09:
@parkr Are you still interested in a `highlight_raw` tag? If yes, i'd be happy to try that one.

parkr:
@Crunch09 Yes!

PR #4090, opened 2015-11-01
Fix #3371 - kramdown:syntax_highlighter should take value of highlighter

### LINKED ISSUE #3371: kramdown:syntax_highlighter should automatically take value of highlighter
Currently, the `highlighter` config value only applies to code wrapped within `{% highlight %}`; fenced code blocks are ignored. You can change this behaviour by providing a value to `kramdown:syntax_highlighter`, but this seems like an unnecessary (and fairly opaque) step. Ideally, rather than this:

```yaml
highlighter: rouge
kramdown:
  syntax_highlighter: rouge
```

we could write this:

```yaml
highlighter: rouge
```

Assuming Rouge is the default, **we wouldn't need to provide a config value at all**. Then using either `~~~ ruby` or `{% highlighter ruby %}` would automatically highlight text. Which would be _lovely_.

paulrobertlloyd:
A fix for https://github.com/jekyll/jekyll/issues/3371

I’ve added a test that checks to see that code fenced syntax is rendered using Rouge. However, this test (and fix?) assumes you have Rouge as your preferred highlighter. I tried enabling Coderay highlighting, but this appeared to have no effect, either on syntax within `{% highlight %}` or `~~~` code fences. Is this a bug, or intended behaviour?

Also, as kramdown only accepts Rouge and Coderay as syntax highlighters, maybe we should only allow those options to be passed to the `kramdown: syntax_highlighter` config?

### DIFF
--- a/lib/jekyll/converters/markdown/kramdown_parser.rb
+++ b/lib/jekyll/converters/markdown/kramdown_parser.rb
@@ -5,7 +5,11 @@ module Jekyll
         def initialize(config)
           require 'kramdown'
           @config = config
-          @config['kramdown']['syntax_highlighter'] ||= @config['highlighter']
+          # If kramdown supported highlighter enabled, use that
+          highlighter = @config['highlighter']
+          if highlighter == 'rouge' || highlighter == 'coderay'
+            @config['kramdown']['syntax_highlighter'] ||= highlighter
+          end
         rescue LoadError
           STDERR.puts 'You are missing a library required for Markdown. Please run:'
           STDERR.puts '  $ [sudo] gem install kramdown'
--- a/lib/jekyll/converters/markdown/kramdown_parser.rb
+++ b/lib/jekyll/converters/markdown/kramdown_parser.rb
@@ -5,6 +5,7 @@ module Jekyll
         def initialize(config)
           require 'kramdown'
           @config = config
+          @config['kramdown']['syntax_highlighter'] ||= @config['highlighter']
         rescue LoadError
           STDERR.puts 'You are missing a library required for Markdown. Please run:'
           STDERR.puts '  $ [sudo] gem install kramdown'
--- a/features/collections.feature
+++ b/features/collections.feature
@@ -9,7 +9,7 @@ Feature: Collections
     And I have a configuration file with "collections" set to "['methods']"
     When I run jekyll build
     Then the _site directory should exist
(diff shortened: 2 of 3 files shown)


paulrobertlloyd:
Hmm, not sure why this is failing in Travis, but not when I run the tests locally… 🤔

envygeeks:
Hulk: ++prepare+patch,merge+and+trace report:ONLYME trace:true stack:true track:callers+with+source

parkr:
Hulk looks pretty cool.

@paulrobertlloyd Are you testing on Ruby 2.2? It succeeds for all the others, just fails on `script/test` in Ruby 2.2. I just kicked the build, let's see if it fails again.

parkr:
> I tried enabling Coderay highlighting, but this appeared to have no effect, either on syntax within {% highlight %} or ~~~ code fences. Is this a bug, or intended behaviour?

I'm not sure if Coderay is supported. I think @envygeeks knows more about this.

> maybe we should only allow those options to be passed to the kramdown: syntax_highlighter config?

Yes, sanitization is good!

envygeeks:
> I'm not sure if Coderay is supported. I think @envygeeks knows more about this.

@parkr In Jekyll coderay isn't supported because we don't accept the `syntax_highlighter` option: https://github.com/paulrobertlloyd/jekyll/blob/9d1641f163e6b20507436872336414cf6df931c7/lib/jekyll/converters/markdown/kramdown_parser.rb#L18  I vote we enable it but don't let it block this pull, I have the tags pull coming today so I'll send up a second pull to enable syntax_highlighter if ya'll want unless you want to do it @paulrobertlloyd

envygeeks:
Actually re-reading that code entirely after drinking some coffee points out that we need to readjust that method to explicitly enable all that on the latest Kramdown.  So once this is merged I'll ship that too.

lib/jekyll/converters/markdown/kramdown_parser.rb
```diff
@@ -5,6 +5,11 @@ class KramdownParser
         def initialize(config)
           require 'kramdown'
           @config = config
+          # If kramdown supported highlighter enabled, use that
+          highlighter = @config['highlighter']
+          if highlighter == 'rouge' || highlighter == 'coderay'
```
paulrobertlloyd: @parkr I’ve added this extra check so that we only set `['kramdown']['syntax_highlighter']` if rouge or coderay are set as the global highlighter. I’m a Ruby newbie, so please tell me if this works, or can be optimised! Thanks.

paulrobertlloyd:
@envygeeks I’ll let you make the changes you need regarding correct coderay support. Much quicker/less error-prone that way I bet!

Issue #4072, opened 2015-10-29
Consider adding a new permalink style with original title name (no downcase)


rebornix:
After upgrading my website from 2.5 to 3.0, `:title` in permalinks will be converted to lower case. All my website page links change slightly, for instance, from `/2015/10/27/NewPost/` to `/2015/10/27/newpost/`. It's a minor change, but all the links I used in other websites are broken totally.

As all permalink style are using the lower case `:title`, there is no backwards compability or any flexibility. Can we have another permalink style to use the original `title` from file name?

parkr:
Seems like a regression to me. Does this affect only posts for you, or for all documents/pages/posts?

rebornix:
It only affects posts, documents and pages are okay. I went through the code base roughly, it maybe not an easy fix. The code is using `Utils.slugify` to generate post file's metadata `permalink` or `title` in pre-process phase

```
  string.
        # Strip according to the mode
        gsub(re, '-').
        # Remove leading/trailing hyphen
        gsub(/^\-|\-$/i, '').
        # Downcase
        downcase 
```

When we finally generate the permalink per Permalink Style configuration
```
def add_permalink_suffix(template, permalink_style)
      case permalink_style
      when :pretty
        template << "/"
      when :date, :ordinal, :none
        template << ":output_ext"
      else
        template << "/" if permalink_style.to_s.end_with?("/")
        template << ":output_ext" if permalink_style.to_s.end_with?(":output_ext")
      end
      template
    end
```

The permalink is already finalized. We can't easily pass any optional parameter to determine whether to lowercase the title.

I'm not a ruby guy, correct me if I'm wrong.

rebornix:
BTW, there is a workaround for this issue: add permalink "/2015/10/31/UpperCase.html" in post's yaml front matter. But I have to update all my posts ...

parkr:
In 3.0.1, `:title` will be what it used to be.

PR #4100, opened 2015-11-03
Add a new property `slug` to document url_placeholders.

rebornix:
The only difference between `slug` and `title` is that uppercase letters are not converted to their lowercase counterparts in `slug`. After adding this property, the final permalink of `_posts/2015-10-27-Jekyll.md` with permalink style `/:year/:month/:day/:slug` will be `/2015/10/27/Jekyll/`.

This pr is related to issue #4072 and it's my first time writing ruby, correct me if I made anything wrong.

### DIFF
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -186,7 +186,9 @@ module Jekyll
         path:        cleaned_relative_path,
         output_ext:  output_ext,
         name:        Utils.slugify(basename_without_ext),
-        title:       Utils.slugify(data['slug']) || Utils.slugify(basename_without_ext),
+        title:       Utils.slugify(data['slug'], mode: "pretty", cased: true) || Utils
+                       .slugify(basename_without_ext, mode: "pretty", cased: true),
+        slug:        Utils.slugify(data['slug']) || Utils.slugify(basename_without_ext),
         year:        date.strftime("%Y"),
         month:       date.strftime("%m"),
         day:         date.strftime("%d"),
--- a/features/permalinks.feature
+++ b/features/permalinks.feature
@@ -96,3 +96,27 @@ Feature: Fancy permalinks
     Then the _site directory should exist
     And the _site/custom/posts directory should exist
     And I should see "bla bla" in "_site/custom/posts/some.html"
+
+  Scenario: Use pretty permalink schema with cased file name
+    Given I have a _posts directory
+    And I have an "_posts/2009-03-27-Pretty-Permalink-Schema.md" page that contains "Totally wordpress"
+    And I have a configuration file with "permalink" set to "pretty"
+    When I run jekyll build
+    Then the _site directory should exist
+    And I should see "Totally wordpress." in "_site/2009/03/27/Pretty-Permalink-Schema/index.html"
+
+  Scenario: Use custom permalink schema with cased file name
+    Given I have a _posts directory
+    And I have an "_posts/2009-03-27-Custom-Schema.md" page with title "Custom Schema" that contains "Totally awesome"
+    And I have a configuration file with "permalink" set to "/:year/:month/:day/:slug/"
+    When I run jekyll build
+    Then the _site directory should exist
+    And I should see "Totally awesome" in "_site/2009/03/27/custom-schema/index.html"
+
+  Scenario: Use pretty permalink schema with title containing underscore
+    Given I have a _posts directory
+    And I have an "_posts/2009-03-27-Custom_Schema.md" page with title "Custom Schema" that contains "Totally awesome"
+    And I have a configuration file with "permalink" set to "pretty"
+    When I run jekyll build
+    Then the _site directory should exist
+    And I should see "Totally awesome" in "_site/2009/03/27/Custom_Schema/index.html"
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -45,7 +45,7 @@ module Jekyll
     # Returns the given filename or title as a lowercase URL String.
     # See Utils.slugify for more detail.
     def slugify(input, mode=nil)
-      Utils.slugify(input, mode)
+      Utils.slugify(input, mode: mode)
     end
 
     # Format a date in short format e.g. "27 Jan 2011".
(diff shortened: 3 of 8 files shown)


envygeeks:
I'm on the fence with this, I don't know if the rest of @jekyll/core will be but surely to me a slug is the one that is normalized, so this is the flip of that.  I would much  rather slug to be something like "cased_title" or something to that effect.

rebornix:
Actually the word `slug` confuses me all the time, when we are trying to extract the title from file name (see code [here](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/document.rb#L298)), 

```
m, cats, date, slug, ext = *relative_path.match(DATE_FILENAME_MATCHER)
```
we are naming the variable as `slug` and it's not normalized. 

@envygeeks  your suggestion is great, `cased_title` is more intuitive and accurate. I'll take your advice if others are Okay with this name.

parkr:
What is most backwards-compatible? I thought `:title` was Cased and `:slug` was uncased.

envygeeks:
In theory if I were programming an app (a blog) that was in Rails or Lotus slug would be the one that downcases, normalizes and transforms and title would be the one that keeps case and does the minimal of normalizing to ensure that a path doesn't break.

parkr:
What did Jekyll do before, though? That's what I'm asking. It sounds like 3.0 changed this behaviour, and I'm not sure it did it for any good reason other than my own laziness or bad testing.

rebornix:
`title` was Cased before 3.0, but after upgrade to 3.0, it became lowercase now. As I mentioned in #4072, any blog which contains posts like "2015-10-27-Cased.md" will break. It's a breaking change.

parkr:
Ok so let's make `:title` cased and `:slug` uncased.

rebornix:
In old version Jekyll, we haven't moved `url_placeholders` from `post.rb` to `document.rb`, we assigned `slug` in `post.rb` to `title`

```
def url_placeholders
      {
        :year        => date.strftime("%Y"),
        :month       => date.strftime("%m"),
        :day         => date.strftime("%d"),
        :hour        => date.strftime("%H"),
        :minute      => date.strftime("%M"),
        :second      => date.strftime("%S"),
        :title       => slug,
```

while in `post.rb`, `slug` is just the title we extract from filename.

```
def process(name)
      m, cats, date, slug, ext = *name.match(MATCHER)
      self.date = Utils.parse_date(date, "Post '#{relative_path}' does not have a valid date in the filename.")
      self.slug = slug
      self.ext = ext
    end
```

But at that time, we don't even have `slugify`. So I suppose this issue is introduced by `utils.slugify`. 

I'll update my pr ASAP.

rebornix:
I've updated my pr to make `:title` cased and add uncased `:slug` for user customization, pls help review :smile:

envygeeks:
That's a pretty complete pull request.  I'm :+1: but just one request, can you squash it?

rebornix:
@envygeeks it's done. Can't wait to see this commit gets merged.

rebornix:
@parkr does this PR look good to you?

Sneezry:
Hey @envygeeks @parkr 

I have the same issue when I build my site with the latest version Jekyll 3.0, and all my links are broken, please merge this pr as soon as possible to solve my problem, thx.

Zhe Li

lib/jekyll/document.rb
```diff
@@ -186,7 +186,8 @@ def url_placeholders
         path:        cleaned_relative_path,
         output_ext:  output_ext,
         name:        Utils.slugify(basename_without_ext),
-        title:       Utils.slugify(data['slug']) || Utils.slugify(basename_without_ext),
+        title:       Utils.slugify_cased(data['slug']) || Utils.slugify_cased(basename_without_ext),
```
parkr: should this be `slugify(data['slug'], cased: true)` perchance?

rebornix: @parkr named parameter is a better solution, thanks for advice! 

BTW, you may want to take a look at issue #4135 , users are complaining about their underscore are replaced by hyphen, maybe someday users would say they want their `~` back. Should we use `Pretty` mode here for `:title`, seems Pretty mode `[^[:alnum:]._~!$&'()+,;=@]+` is designed to handle this kind of issue.

parkr: Yeah that sounds good.

parkr:
> @parkr does this PR look good to you?

@rebornix Cucumber tests fail :(

rebornix:
@parkr  such a silly mistake :( . Fixed now.

Stargator:
:+1: @rebornix For a first go at Ruby you did great! I'm learning Ruby as well.

parkr:
^^ agreed!

yous:
Hi, I used my `permalink` setting as `/:year/:month/:day/:title/`. Then after upgrading Jekyll from 3.0.0 to 3.0.1, the links are changed, for example, from `/2013/02/18/ios-6-1-music-album-shuffle/` to `/2013/02/18/ios-6.1-music-album-shuffle/`. Note the `6-1` and `6.1`. For the consistency, I need to change the `permalink` setting to `/:year/:month/:day/:slug/`.

I agree to the direction of change, making `title` pretty. But I think this might be considered as a breaking change, since it can break some sites silently as URLs are delicate. I see this pull request as adding `:slug` and changing `:title`, not fixing it.

rebornix:
@yous Actually if you are upgrading Jekyll from 2.x to 3.0.1, you will not notice any breaking change. From my perspective, 3.0.1 is like a quick turnaround for some 3.0.0 unplanned breaking change. 

BTW, instread of just making title pretty, I add slug for anyone who wants to keep up with Jekyll slugify trend. You got the chance to make everything correct, right?

People in community are trying to make slugify [customizable](https://talk.jekyllrb.com/t/is-there-a-desirability-for-configurable-slugify-character-in-core/1558/7), you might be interested.

yous:
@rebornix Oops, I hadn't noticed the change when I upgrade Jekyll 2.x to 3.0.0.pre.x. Well... it seems I need more tests. Thanks for pointing it out!

envygeeks:
If it's not noted in our upgrading docs that @parkr just released please file a ticket so we can update those docs to mention that.

PR #4101, opened 2015-11-03
Adds a note about installing the jekyll-gist gem to make gist tag work

ursooperduper:
I've been using the docs as I develop my own site and ran into errors trying to embed gists. I discovered you have to add the `jekyll-gist` gem to your site in order to have the `gist` tag work, so I've updated the templates page docs accordingly.


envygeeks:
:shipit:

envygeeks:
Thanks @ursooperduper

ursooperduper:
My pleasure! I'm looking for ways to get more involved.

PR #4104, opened 2015-11-03
align hooks documentation and implementation

### LINKED ISSUE #4102: missing post_init hook for documents
http://jekyllrb.com/docs/plugins/#hooks lists a "post_init" hook for "documents", however it is conspicuously missing from [hooks.rb](https://github.com/jekyll/jekyll/blob/a5b46821ad94666a6fb49cb3c2236b6a9a178328/lib/jekyll/hooks.rb#L32-L36).  This results in an error:

> jekyll 3.0.0 | Error:  Invalid hook. documents supports only the following hooks [:pre_render, :post_render, :post_write]

I'm happy to provide a patch, but wanted to first check on whether it's the documentation or the hooks that should be fixed.  Is there supposed to be a documents post_init hook?

willnorris:
- add site post_render hook, which was documented but wasn't being
  called
- define documents post_init hook, which was documented but caused an
  error when called (fixes #4102)
- add docs for site post_read hook, which was being called but wasn't
  documented
- fix container name in example: s/post/posts/

### DIFF
--- a/lib/jekyll/hooks.rb
+++ b/lib/jekyll/hooks.rb
@@ -15,6 +15,7 @@ module Jekyll
         after_reset: [],
         post_read: [],
         pre_render: [],
+        post_render: [],
         post_write: [],
       },
       :pages => {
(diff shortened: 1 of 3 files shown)


parkr:
Thank you so much, Will! Sorry for the brokenness. 

:shipit:

PR #4109, opened 2015-11-04
Modernize Kramdown for Markdown converter.

envygeeks:


parkr:
@envygeeks when you comment on the diff for a PR, it's hard to see what you're commenting on.

envygeeks:
@parkr what do you mean? That's standard behavior that is always done.

parkr:
> @parkr what do you mean? That's standard behavior that is always done.

I have never done this before, I always comment on the PR diff. The comments on a PR diff are hidden once you update the PR, whereas comments on a commit are never hidden. I can see you have two kinds of comments: (1) a pre-emptive "i'm going to fix this, don't worry" comment, and (2) a "for your information" comment. I think comment 1 is better on the PR diff, whereas 2 works great on the commit.

envygeeks:
The assumption here is that it's not a mitigation before another commit.  The problem is most of those changes should be obvious but after the fact I mitigated potential question before it happened because there is no plan to update this pull for a few days because there are more important things in the backlog than this, this was just fullfilling a comment on the problems with the processor and mitigating once the pull was made until the PR can be updated with the changes.

Both of those fall into category 1 in this circumstance.

envygeeks:
It's natural for me to mitigate questions if I do not plan work on something before those questions come, that way I don't have to backtrack and answer those questions I foresaw.  It's natural behavior for me.

parkr:
Ok. Let me know when this is ready!

lib/jekyll/converters/markdown/kramdown_parser.rb
```diff
@@ -1,33 +1,95 @@
+# Frozen-string-literal: true
+# Encoding: utf-8
+
 module Jekyll
   module Converters
     class Markdown
       class KramdownParser
+        CoderayDefaults = {
```
parkr: We usually do `SCREAMING_SNAKE_CASE` for constants.

lib/jekyll/converters/markdown/kramdown_parser.rb
```diff
@@ -1,33 +1,95 @@
...
-          # Check for use of coderay
-          if @config['kramdown']['enable_coderay']
-            %w[wrap line_numbers line_numbers_start tab_width bold_every css default_lang].each do |opt|
-              key = "coderay_#{opt}"
-              @config['kramdown'][key] = @config['kramdown']['coderay'][key] unless @config['kramdown'].key?(key)
+          Kramdown::Document.new(content, @config).to_html
+        end
+
+        private
+        def load_kramdown
```
parkr: this can be replaced by a call to `Jekyll::External.gracefully_require 'kramdown'` which has an error message and such built into it

parkr:
@envygeeks I'm confused by all the stuff you're doing with configuration in this converter. Maybe you can explain that to me?

lib/jekyll/converters/markdown/kramdown_parser.rb
```diff
@@ -1,33 +1,95 @@
...
-
-          Kramdown::Document.new(content, Utils.symbolize_hash_keys(@config['kramdown'])).to_html
         end
 
+        private
+        def strip_coderay(hash)
+          hash.inject({}) do |hash_, (key, val)|
+            cleaned_key = key.gsub(/\Acoderay_/, "")
+            if hash.has_key?(key)
+              Jekyll.logger.warn "You are using an old CodeRay key: '#{key}'." \
```
parkr: Is this considered a deprecation?

envygeeks:
It normalizes and modernizes everything that was legacy and supported in old Jekyll until 4.0 were all that will be stripped out and only "syntax_highlighter_opts", "syntax_highlighter" and "highlighter" will be supported. 

Right now Kramdown uses `syntax_highlighter_opts`: http://kramdown.gettalong.org/options.html#option-syntax-highlighter-opts and `syntax_highlighter`

So what we do is 1. Search for "enable_coderay" and tell them *NO* use "syntax_highlighter" on Kramdown, enable_coderay is old.. but we still let it pass and kick on coderay for them and 2. "highlighter" which is what Jekyll allows so we say nothing and we alias it over to what Kramdown prefers.

Then after we've done that, we then go through and normalize the configuration.  First we check to see if you have the modern version of Kramdown "syntax_hilighter_opts", then we search to see if you have "coderay" and if you do we merge them and then lastly since "coderay_" is old, we then go onto normalize that as non "coderay_" on "syntax_highlighter_opts" and ship a warning that "coderay_" is old, modernize your configuration.

In code:

```ruby
{
"kramdown": {
  "enable_coderay": true
  "coderay_opt1": value
  "coderay": {
    "opt2": value
  }
}
```

becomes:

```ruby
{
"kramdown": {
  "syntax_highlighter": :coderay,
  "syntax_highlighter_opts": {
    "opt1": value,
    "opt2": value
}
```

lib/jekyll/converters/markdown/kramdown_parser.rb
envygeeks: Yessum.

lib/jekyll/converters/markdown/kramdown_parser.rb
```diff
@@ -1,33 +1,84 @@
...
+            else
+              @main_fallback_highlighter
             end
           end
+        end
+
+        private
+        def strip_coderay_prefix(hash)
+          hash.inject({}) do |hash_, (key, val)|
+            cleaned_key = key.gsub(/\Acoderay_/, "")
```
parkr: Could we do a `sub` call here? To make things faster?

lib/jekyll/converters/markdown/kramdown_parser.rb
```diff
@@ -1,33 +1,84 @@
...
+
+        # config[kramdown][syntax_higlighter] > config[kramdown][enable_coderay] > config[highlighter]
+        # Where `enable_coderay` is now deprecated because Kramdown
+        # supports Rouge now too.
+
+        private
+        def highlighter
+          @highlighter ||= begin
+            if out = @config["syntax_highlighter"] then out
+            elsif @config.has_key?("enable_coderay") && @config["enable_coderay"]
```
parkr: I believe `has_key?` has been deprecated in favour of `key?`. Would you mind updating here and below?

parkr:
:tada:

parkr:
@envygeeks Know what the failure on Ruby 2.2 is about?

envygeeks:
@parkr I don't but I'll be getting back to this and updating it tomorrow anyways so I'll look at it then.  Now that the new Docker stuff for EnvyGeeks is out and released I have some free time.

lib/jekyll/converters/markdown/kramdown_parser.rb
envygeeks: I've moved this to using the fastest of all, `Hash#each_with_object`

lib/jekyll/converters/markdown/kramdown_parser.rb
envygeeks: Yes, I always forget about that, it's been moved to `Hash#key?` good catch!

lib/jekyll/converters/markdown/kramdown_parser.rb
```diff
@@ -1,33 +1,86 @@
...
           end
+        end
+
+        private
+        def strip_coderay_prefix(hash)
+          hash.inject({}) do |hash_, (key, val)|
+            cleaned_key = key.gsub(/\Acoderay_/, "")
+            if hash.has_key?(key)
+              Jekyll.logger.warn "You are using an old CodeRay key: '#{key}'." \
+                "It is being normalized to #{cleaned_key}."
```
envygeeks: I shortened the message to make it clearer to people who don't know about that trick.

envygeeks:
This is ready /cc @jekyll/core

PR #4113, opened 2015-11-05
Convert the doctype to lowercase

woneob:
HTML5 doctype declaration is case-insensitive but here's some reason.
Please check this issue  https://github.com/h5bp/html5-boilerplate/issues/1522

And some popular HTML boilerplate also use a lowercase doctype.
- html5-boilerplate: https://github.com/h5bp/html5-boilerplate/blob/master/src/index.html#L1
- Web Starter Kit: https://github.com/google/web-starter-kit/blob/master/app/index.html#L1

and google.com too.

### DIFF
--- a/site/_includes/top.html
+++ b/site/_includes/top.html
@@ -1,4 +1,4 @@
-<!DOCTYPE HTML>
+<!doctype html>
 <html lang="en-US">
 <head>
   <meta charset="UTF-8">


envygeeks:
Is there any non-superficial reason why this matters?  The Google bandwagon won't work here.

envygeeks:
Non-superficial also includes "compression" in the face of simple context benching.

woneob:
Perhaps, uppercase and lowercase doctypes are almost the same.
But, I found a page that compares the compression efficiency.

> `<!doctype html>` good (saves 2 bytes once compressed)

[pseudo thermal view of Gzip/Deflate compression efficiency](http://encode.ru/threads/1889-gzthermal-pseudo-thermal-view-of-Gzip-Deflate-compression-efficiency)

What do you think about this?

envygeeks:
That's what I pushed back against in comment number 2.  In the grandscheme of everything you download from every site 2bytes matters not, and even in the most optimized site when a connection is already established and the TTFB is where all the problem is, 2bytes really doesn't matter.

That's a microoptimization and subjective bench that neglects where the real problem is: the TTFB on connect w/ roundtrip.  2bytes isn't gonna fix that when even the most simple pages extend 1000x that.

parkr:
Thanks for the pull request, @woneob! I agree that the compression isn't a big enough win here. From looking at the HTML5BP issue you linked, it looks like there are issues with IDE's. Many Jekyll users are non-technical so messing with anyone's IDE to save 2 bytes doesn't seem all that beneficial.

Thanks again for the PR! Sorry that this one won't make it through. I appreciate your willingness to provide lots of details about your reasoning. :)

PR #4114, opened 2015-11-05
Fix the warning

chrisfinazzo:
Think this should be `site.posts.docs` instead. Does it need a test? - works with jekyllrb.com and `script/cibuild` passed locally.

### DIFF
--- a/lib/jekyll/commands/doctor.rb
+++ b/lib/jekyll/commands/doctor.rb
@@ -49,7 +49,7 @@ module Jekyll
           conflicting_urls = false
           urls = {}
           urls = collect_urls(urls, site.pages, site.dest)
-          urls = collect_urls(urls, site.posts, site.dest)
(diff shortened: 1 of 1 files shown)


envygeeks:
It doesn't need a test but it does need a narrow commit message please!

sondr3:
I guess this is the fix for this warning or?

```
Deprecation: Collection#each should be called on the #docs array directly.
             Called by /usr/local/lib/ruby/gems/2.2.0/gems/jekyll-3.0.0/lib/jekyll/commands/doctor.rb:81:in `collect_urls'.
```

chrisfinazzo:
@sondr3 Correct.

parkr:
@chrisfinazzo Please rephrase the commit message. Something like "Fix the deprecation warning in the doctor command" would work just fine. Thanks!

PR #4121, opened 2015-11-09
Change GitHub code block to highlight tag to avoid it overlaps parent div

rebornix:
image

Some code blocks are still using GitHub's code block syntax **```**, which lacks proper x-overflow css. Here I change it to `highlight` tag.

### DIFF
--- a/site/_docs/deployment-methods.md
+++ b/site/_docs/deployment-methods.md
@@ -116,9 +116,9 @@ That's why rrsync wrapper shall be installed. If it is not already installed by 
 
 [This process is described in a lot of places in the net](https://wiki.gentoo.org/wiki/SSH#Passwordless_Authentication). We will not cover it here. What is different from usual approach is to put the restriction to certificate-based authorization in ```~/.ssh/authorized_keys```). We will launch ```rrsync``` utility and supply it with the folder it shall have read-write access to:
 
-```
+{% highlight bash %}
 command="$HOME/bin/rrsync <folder>",no-agent-forwarding,no-port-forwarding,no-pty,no-user-rc,no-X11-forwarding ssh-rsa <cert>
-```
+{% endhighlight %}
 
 ```<folder>``` is the path to your site. E.g., ```~/public_html/you.org/blog-html/```.
 
--- a/site/_sass/_style.scss
+++ b/site/_sass/_style.scss
@@ -643,10 +643,14 @@ p > pre,
 p > code,
 p > nobr > code,
 li > code,
+li> pre,
 h5 > code,
 .note > code {
   background-color: #2b2b2b;
   color: #fff;
+  max-width: 100%;
+  overflow-x: auto;
+  vertical-align: middle;
   @include border-radius(5px);
   @include box-shadow(inset 0 1px 10px rgba(0,0,0,.3),
     0 1px 0 rgba(255,255,255,.1),


rebornix:
Actually, single ` may run into the same situation in small window, like below

image

So I update css for `li>pre` and `p>code`, make its `max-width` as `100%` and `overflow` as `auto`.

image
 
BTW, personally I suggest we use `highlight` block anytime we try to add a block of codes, instead of using GitHub multiline code block syntax.

mattr-:
This looks better, but the user still has to scroll sideways. The text should wrap within the column. Can you make that update please?

envygeeks:
Yeah definitely needs to be split up with \ and a line break.

rebornix:
@mattr- word wrap is good but the catch is, users would have no idea whether the code block is a single line command any more, like below

image

@envygeeks mentioned using `\` to split the long code snippet, it's like a content change but it's hard to tell where to break the line, users' window size vastly changes.

Actually this problem only happens when we put single line code block in paragraphs, we can't break the code block like other words. If you guys think a soft word-break like the image shown above is Okay, I'll give it a quick fix :hammer:

chrisfinazzo:
@mattr- Is there a reason adding the highlight [class][] wouldn't work? Interested to know why scrolling sideways is non-optimal in this case. If the plugin install is meant to be a single command (as the docs seem to imply), then scrolling makes sense.

[class]: https://github.com/jekyll/jekyll/blob/master/lib/jekyll/tags/highlight.rb#L116

rebornix:
@chrisfinazzo The reason I use `text` is that this line of code is likely config instead of a shell command, but `bash` can give users a better highlighting, I'm Okay with the change. Before I update the pr, let's wait for others' input.

envygeeks:
Personally I'm fine either way the choice goes because changing it shouldn't affect anything and if it does then we either have a UI problem on the site or a processing problem in Jekyll with our upstream providers.  The shell syntax provides no benefit because we aren't using any builtins for sh/bash.

On the "not knowing the screen sizes" (paraphrased) part, I assume that because we have a fixed layout that adjusts the font size as it gets smaller, in theory if whoever designed the site did it right, there should be no scrolling once the screens get smaller because the responsive side should shrink the font to a size that allows that to happen.  I've never tested it though.

rebornix:
I agree, the font-size or layout width should be taken care of carefully in mobile view. 

The root cause is we are heavily using single-line code block in paragraphs. When we resize our window to 414*736 ( iPhone 6 plus), 6 out of 30 documentation pages are broken, like in [collection](http://jekyllrb.com/docs/collections/)

image

It just breaks slightly and we can't say `<dest>/my_collection/some_subdir/some_doc.html` is too long, right? 

For now, `overflow:auto` is an acceptable solution for me or @chrisfinazzo . If Jekyll team has a full plan for responsive font-size shrinking, it can't be better.

envygeeks:
I'll leave it to @mattr-, @alfredxing and @parkr to decide but I'm gonna be honest, I wouldn't mind if we told CSS it's okay to break in the middle of the word and fold the text to a new line on inline code since it doesn't break anything like paste.  I don't know if users will notice it's a single word though.

alfredxing:
My vote would be to limit the width of the code block, then make the text inside scroll (using `max-width: 100%; overflow-x: auto`).

Example: http://jsfiddle.net/sn3Lu2zr/

chrisfinazzo:
👍🏻 on @alfredxing's change.

rebornix:
Rebasing the pr and restrict overflow to x-axis as @alfredxing suggested.

envygeeks:
That should be `sh`, `shell` or `bash`, not text.

PR #4130, opened 2015-11-11
Update pagination.md

nicolewhite:
elsif -> elif


mattr-:
Thanks for your contribution! :heart:

envygeeks:
Why did this need to change to `elif`? Liquid uses Rubyism's and in Ruby `elsif` is it, and even the documentation for Liquid itself states that: https://docs.shopify.com/themes/liquid-documentation/tags/control-flow-tags#elsif-else

nicolewhite:
Apologies, I thought it was a typo since all the examples I've seen use `elif`.

envygeeks:
So elif works? If elif works then I have no complaints and carry on.

Issue #4129, opened 2015-11-11
CoffeeScript compilation broken with Jekyll 3.0


perlun:
Unfortunately, I can't come to any other conclusion.

If I create a file named `foo.coffee` in the root folder of my project, and give it the following content:

```coffee
---
---

a = () -> 'bar'
```

...and then run `jekyll build`, the result is not the expected one. The file is copied *more-or-less* verbatim into the `_site` folder. But the front matter is removed, i.e. the first lines with three dashes (which is the [suggested approach](http://jekyllrb.com/docs/assets/) in the documentation), and the result is this:

`a = () -> 'bar'`

Que? :smile: 

My local environment:

```
$ jekyll --version
jekyll 3.0.0
$ uname -a
Darwin ECVAAWPLUN5.local 14.4.0 Darwin Kernel Version 14.4.0: Thu May 28 11:35:04 PDT 2015; root:xnu-2782.30.5~1/RELEASE_X86_64 x86_64
$ ruby -v
ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-darwin14]
$ coffee -v
CoffeeScript version 1.10.0
```

envygeeks:
Did you install `jekyll-coffeescript` which is not included by default in Jekyll 3.0 anymore.

lennyjpg:
gems: [jekyll-coffeescript]

at the end of _config.yml fixed this for me.

perlun:
Adding `jekyll-coffeescript` to my `Gemfile` and running `bundle install` didn't help. *However*, doing that and also adding it to the list of `gems in the `_config.yml` did the. Thanks!

I guess we should add this to the docs then? Right now, [they](http://jekyllrb.com/docs/assets/) are actually rather misleading:

> Jekyll provides built-in support for Sass and CoffeeScript. In order to use them, create a file with the proper extension name (one of .sass, .scss, or .coffee) and start the file with two lines of triple dashes, like this:

It doesn't say anything whatsoever about extra gems being needed. So we should definitely fix them if this is the expected behavior.

Lewiscowles1986:
@perlun check out https://github.com/Lewiscowles1986/jekyll/commit/c984d8c531c4bf8196576bd73875b052f3d3bd97 let me know if that makes it clearer?

PR #4137, opened 2015-11-14
Updated to reflect feedback in #4129

Lewiscowles1986:
Included Jekyll 3 specific update to mention that jekyll-coffeescript gem must be installed and the gem added to the `_config.yml` file


envygeeks:
:heart:

PR #4140, opened 2015-11-14
correcting typo: elif to elsif

krockgardin:


mattr-:
_sigh_

So which one is it, exactly?

envygeeks:
@mattr- the actually documentation for Liquid uses {% elsif %}

PR #4144, opened 2015-11-15
add "-o" option to serve command which opens server URL

friedenberg:
Hi Jekyll maintainers,

We were setting up a new Jekyll blog to play around and thought it would be helpful to add a "-o" option to the serve command. This option would open the URL of the server on the device's default browser. We used the "system" ruby method, but we're not sure if that's your preferred approach to opening URL's. It's also unclear if this would work on Windows, as we only tested this on OS X.

We're happy to modify this request to open the URL in a different way, if there are issues with the current implementation. All the unit tests passed when we ran them locally.

cc @ckearns1210

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -19,6 +19,7 @@ module Jekyll
             c.option 'host', '-H', '--host [HOST]', 'Host to bind to'
             c.option 'baseurl', '-b', '--baseurl [URL]', 'Base URL'
             c.option 'skip_initial_build', '--skip-initial-build', 'Skips the initial site build which occurs before the server is started.'
+            c.option 'open_url', '-o', '--open-url', 'Opens the local URL in your default browser'
 
             c.action do |args, options|
               options["serving"] = true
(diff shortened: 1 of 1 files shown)


envygeeks:
This is broken on Linux (and I don't think it will work on Windows) either, `open` does not do what you think it does on Ubuntu and Debian where it's actually `openvt`

friedenberg:
aha, that's what worried me. Is there any safe way to do this with multiple platform support, or is this a lost cause?

friedenberg:
would it be crazy to do something like [this](http://sweetme.at/2013/09/02/how-to-open-a-url-in-the-default-browser-across-os-platforms/)?
`python -m webbrowser #{server_address}`

It might not be safe to assume that python is installed.

envygeeks:
* `Windows`: `start url`
* `Linux`: `xdg-open url`
* `OS X`: As is.

friedenberg:
alright, updated to include different OS commands.

alfredxing:
:+1:

I know in our Rakefile, [we use Launchy](https://github.com/jekyll/jekyll/blob/master/Rakefile#L149) to open the built Jekyll site for the `preview` task.

lib/jekyll/commands/serve.rb
```diff
@@ -45,7 +46,24 @@ def process(options)
             file_handler_options
           )
 
-          Jekyll.logger.info "Server address:", server_address(s, options)
+          server_address = server_address(s, options)
```
parkr: this reassignment of `server_address` is confusing – could you rename the variable on the left here, please?

friedenberg:
ok, fixed the naming of the variable. I'll keep in mind Launchy for stuff like this in the future, too.

Issue #4067, opened 2015-10-28
Collections break if "[" in path.


Yarn:
### Steps to reproduce
jekyll b or jekyll serve a project that uses collections located at a path containing a left square bracket "[".
### Expected behaviour
Jekyll will function as it does with projects located at paths not containing a "[".
### Actual behaviour
No files or data corresponding to files in collections are generated.
### Notes
```{{ collections[0] }}``` results in something like
```{"output"=>true, "label"=>"things", "docs"=>[], "files"=>[], "directory"=>"/path/aa[/_things", "relative_directory"=>"_things"}``` even if files are present in ```/path/aa[/_things```
Tested on arch linux and windows with same results.

parkr:
Tangential question: when is a `[` used in a directory? Never seen that before, actually.

envygeeks:
I think this is a case for needing to pass around Pathname everywhere.  This is exactly why every one of my apps use Pathname for everything.

parkr:
How does Pathname fix this issue?

envygeeks:
Without seeing more data and testing it myself the assumption here is that there is a hiccup in the read and write because there is no escaping. Not that it should matter.  Pathname will escape everything for you and make it literal.

Yarn:
I had one at the start of the directory name for sorting, it's not something that normally happens. It took a good while to figure out why it wasn't working though.
Also of note, ], =, -, %, @ and # don't seem to cause issues.

BlockWorld:
That's very strange that "[" would cause issues but not "]"...

envygeeks:
I thought that too but it's a bug none-the-less if we can reproduce it.

envygeeks:
I am unable to reproduce this on my system (Ubuntu 14.04), can anybody else reproduce this?

ducktyper:
I found that using absolute path in Dir.glob was the problem.
Dir.glob identifies '[' as character set matching and I tested that it returns an empty array without ']'.
Having '{', '?', '*' in the absolute path also causes the similar issue.

Also tested that the results mentioned in Notes happen if entries are an empty set.

I fixed the issue here #4150 and also fixed one more similar case in cleaner.rb
I don't know how to add tests for this case so I didn't do it.

Let me know if there is any feedback to this commit.

PR #4150, opened 2015-11-17
Avoid using Dir.glob with absolute path

ducktyper:
the absolute path including '[', '{', '?', or '*'
could change the outcome

### DIFF
--- a/lib/jekyll/collection.rb
+++ b/lib/jekyll/collection.rb
@@ -74,11 +74,8 @@ module Jekyll
     def entries
       return Array.new unless exists?
       @entries ||=
-        Dir.chdir(collection_dir) do
-          Dir.glob("**/*.*").map do |f|
-            entry = collection_dir(f)
-            entry["#{collection_dir}/"] = ''; entry
-          end
+        Utils.safe_glob(collection_dir, "**/*.*").map do |entry|
+          entry["#{collection_dir}/"] = ''; entry
         end
     end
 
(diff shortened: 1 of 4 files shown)


ducktyper:
#4067

ducktyper:
That's a good idea. now Utils.safe_glob(dir, pattern, flags=0) is created and used.
I created 'test/safe_glob_test[/find_me.txt' for testing but I'm worrying that the folder name is too noticeable in test folder. Let me know if there is a better place to put it.

Also I found out that there are few more places have the same issue using Dir[] with absolute path. If you change the project folder name from 'jekyll' to 'jekyll[' then many tests fail.
I can fix them to this pull request if it needs to be fixed.

lib/jekyll/collection.rb
```diff
@@ -74,7 +74,7 @@ def read
     def entries
       return Array.new unless exists?
       @entries ||=
-        Dir.glob(collection_dir("**", "*.*")).map do |entry|
+        Utils.safe_glob(collection_dir, "**/*.*").map do |entry|
```
parkr: If `**/*.*` platform-independent?

parkr:
Looking good! Just a question above about Windows support.

> Also I found out that there are few more places have the same issue using Dir[] with absolute path.

In other parts of Jekyll? Or just the tests? If you want to update those, too, I would love that. Maybe in a different PR?

ducktyper:
Yes Windows support, I totally didn't see it. Thanks.
Now safe_glob also accepts array of patterns to support Windows.
I still leave linux style path in the test. Let me know if they need to be changed too.

The other problem I mentioned is in other parts of Jekyll. I will make another pull request if I fix them.

parkr:
LGTM!

Issue #1890, opened 2013-12-29
Add random select


zlatanvasovic:
It would be nice to have random select option, e.g. for data items, or posts.

BrendanBerkley:
Been searching around for a random feature. My use case is that I'm prototyping a product grid, and I'd like to be able to pull from an array of images to populate the grid without defining specific numbers. The grid would be randomly generated on build, and that'd be perfect - the grid now simulates a diverse set of products without me having to set anything manually other than a list of images.

I'll probably just go with a plugin, but I wanted to chime in to vocalize a use case.

parkr:
@BrendanBerkley I'd look at a pr that adds a `random` filter to https://github.com/jekyll/jekyll/blob/master/lib/jekyll/filters.rb

parkr:
Use the `sample` filter, coming your way in Jekyll 3.1.0: #4223.

PR #4156, opened 2015-11-18
Add random filter

BrendanBerkley:
Solution for #1890.

### DIFF
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -281,6 +281,16 @@ module Jekyll
       new_ary
     end
 
+    # Get a random value from an array
+    #
+    # array - the array to select from
+    #
+    # Returns a single item in the array
+    def random(array)
+      return array unless array.is_a?(Array)
+      array.sample(1)
+    end
+
     # Convert an object into its String representation for debugging
     #
     # input - The Object to be converted
--- a/site/_docs/templates.md
+++ b/site/_docs/templates.md
@@ -246,6 +246,17 @@ common tasks easier.
         </p>
       </td>
     </tr>
+    <tr>
+      <td>
+        <p class="name"><strong>Random</strong></p>
+        <p>Get a random item from an array.</p>
+      </td>
+      <td class="align-center">
+        <p>
+         <code class="filter">{% raw %}{{ site.members | random }}{% endraw %}</code>
+        </p>
+      </td>
+    </tr>
   </tbody>
 </table>
 </div>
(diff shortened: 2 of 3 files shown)


parkr:
From a UX perspective, I would expect

```liquid 
{% my_array | sample %}
```

to return one random item (not an array). In contrast, I'd expect

```liquid
{{ random }}
```

to return a random float in the range [0,1), and

```liquid
{{ random:5 }}
```

to return a random integer in the range [0,5).

Does that seem reasonable? Don't want `random` to confuse folks.

BrendanBerkley:
Seems reasonable. I'll work it out.

penibelst:
I’m glad randomness is not evil anymore.

The proposed implementation draws exactly one ball from the [urn](https://en.wikipedia.org/wiki/Urn_problem) an then returns it back. Next time the same ball can be drawn again. It’s impossible to draw more than one ball at a time.

My rejected proposal #1892 works for any number of balls.

## 1 ball

```
{{ balls | shuffle | first }}
```

## *n* balls

```
{% assign shuffled_balls = balls | shuffle %}
{% for ball in shuffled_balls limit:n %}
  {{ ball }}
{% endfor %}
```

What do you think about drawing more than one ball?

parkr:
@penibelst Why not `sample:n`?

```liquid
{{ balls | sample:12 | first }}
```

penibelst:
@parkr I like the `sample:n` notation too.

## 1 ball

```
{{ balls | sample: 12 | first }}
```

## *n* balls

```
{% assign shuffled_balls = balls | sample: n %}
{% for ball in shuffled_balls %}
  {{ ball }}
{% endfor %}
```

envygeeks:
Personally I feel like sample is where it's at.  It just feels right.

BrendanBerkley:
I'm with @envygeeks here - I like the simpler approach. Though I suppose it would be possible to do `{{ thing | sample }}` as well as `{{ thing | sample:12 }}`, right?

This is my first time doing a pull request where people don't agree on what to do, so I'm not sure how to move forward! I don't want to keep working on something that people won't want to use.

Also, @parkr - for the random filter idea, you used filter notation, but that'd have to be a custom tag right? I got something working locally as a tag but not as a filter, and I didn't see a precedent in the existing filters for something like you suggested.

parkr:
> for the random filter idea, you used filter notation, but that'd have to be a custom tag right? I got something working locally as a tag but not as a filter, and I didn't see a precedent in the existing filters for something like you suggested.

Can you clarify with liquid examples?

parkr:
Merged the `sample` filter. #4223

PR #4160, opened 2015-11-18
Fix #4082: Allow users to use .htm and .xhtml (XHTML5.)

### LINKED ISSUE #4082: files named .htm end up as index.html.htm (3.0)
I have a page, the file name is order-now.htm

In the front matter I have the permalink as: permalink: /order-now/

When I build it it creates a folder named order-now (correct) but inside the file is named index.html.htm instead of just index.html.

It worked fine in jekyll 2.x, apparently I named some of my files .html and others .htm. Anything that I named .htm is broken.

This seems like the same thing as this persons shtml problem: https://talk.jekyllrb.com/t/shtml-weirdness/1367/6

This is on Win 10.

envygeeks:


parkr:
Should this also be applied to documents?

parkr:
> Should this also be applied to documents?

@envygeeks ^^

envygeeks:
@parkr I tried but I couldn't figure out where (but that's probably because I was dealing with a storm of bugs in our stack yesterday so I wasn't paying attention.) I'll play with it today and make sure documents work as expected and update and ping if I need to!

parkr:
Thanks :)

parkr:
Document code: https://github.com/jekyll/jekyll/blob/3417a33e7a1a903a170565f910388f6ba3831ea1/lib/jekyll/document.rb#L235-L241

envygeeks:
On it, I forgot about this when I was doing stuff.  I'll get it fixed up.

PR #4167, opened 2015-11-19
Fix Jekyll::URL.sanitize_url to handle removal of triple slashes

vwochnik:
In it's current form, `sanitize_url` is only able to clear double slashes but for my use case it is necessary to sanitize triple slashes as well.

Here is an example of the broken `url_sanitize`:

```
Template:/:language/:path/:basename:output_ext
Permalink: 
Placeholders: {:path=>"/", :basename=>"404", :output_ext=>".html", :language=>"de"}
Unsanitized URL: '/de///404.html' -> Sanitized: '/de//404.html'
```

After the fix, this is the result:
```
Template:/:language/:path/:basename:output_ext
Permalink: 
Placeholders: {:path=>"/", :basename=>"404", :output_ext=>".html", :language=>"de"}
Unsanitized URL: '/de///404.html' -> Sanitized: '/de/404.html'
```

I hope for this to be merged since otherwise I will have to patch `url_sanitize` for my plugin and I don't like monkey patching that much.

### DIFF
--- a/lib/jekyll/url.rb
+++ b/lib/jekyll/url.rb
@@ -73,10 +73,8 @@ module Jekyll
     # Returns a sanitized String URL
     def sanitize_url(in_url)
       url = in_url \
-        # Remove all double slashes
-        .gsub(/\/\//, '/') \
-        # Remove every URL segment that consists solely of dots
-        .split('/').reject{ |part| part =~ /^\.+$/ }.join('/') \
+        # Remove empty URL segments and every URL segment that consists solely of dots
+        .split('/').reject{ |s| s.empty? || s =~ /^\.+$/ }.join('/') \
         # Always add a leading slash
         .gsub(/\A([^\/])/, '/\1')
 


envygeeks:
I think that's a bit unoptimized, you can solve that by just changing it to `gsub(/\/{2,}/, "/")` or even just `gsub(/\/+/, "/")` that way it doesn't come up for a billion, trillion or any number past what we care about and what you care about.

envygeeks:
Actually I think that whole method is unoptimized tbh.

vwochnik:
Well, this way the whole `.gsub()` part is removed and replaced by the `.empty?` check. I find that a much better and cleaner solution since we are already filtering out dots so why not filter out empty segments at once?

The only unoptimized part IMO is the last check preventing that the method can be one chain.

parkr:
> why not filter out empty segments at once?

:+1:

envygeeks:
Which is entirely inefficient as I said.  This isn't putting the blame on you it's not your fault, it's us who created that mess and it was just exacerbated when I seen that code just now:

```
you: /hello/world
 me: /hello/world
Comparison:
 me: ../../../../../../hello/../../../../../world:    86921.4 i/s
you: ../../../../../../hello/../../../../../world:    46718.2 i/s - 1.86x slower


you: /hello/world
 me: /hello/world
Comparison:
 me: ../hello/////world:   102811.0 i/s
you: ../hello/////world:    73789.5 i/s - 1.39x slower


you: /hello/world
 me: /hello/world
Comparison:
  me: ../hello/world:   140798.9 i/s
 you: ../hello/world:    82044.7 i/s - 1.72x slower


you: /hello/world
 me: /hello/world
Comparison:
    me: /hello/world:   136214.5 i/s
   you: /hello/world:    91659.3 i/s - 1.49x slower


you: /hello/world
 me: /hello/world
Comparison:
     me: hello/world:   322206.3 i/s
    you: hello/world:    94040.9 i/s - 3.43x slower
```

```ruby
"/" + str.gsub(/\/{2,}/, "/").gsub(%r!\.+\/|\A/+!, "")
```

vwochnik:
@envygeeks You have a point. But what if you have `/../`? That is a valid path component. I would only replace `/./` as in the original version. As for the speed, you are probably right since the reject{} is calling the block for each url segment. I wouldn't consider that a problem since URLs are fairly short but you call the shot.

envygeeks:
My code behaves exactly likes yours:

```ruby
[6] pry(main)> \
STDOUT.puts your_way "/../hello"; \
STDOUT.puts my_way "/../hello"; \
STDOUT.puts your_way "/hello/../"; \
STDOUT.puts my_way "/hello/../";
/hello
/hello
/hello/
/hello/
```

So what are you talking about?

envygeeks:
```ruby
[7] pry(main)> \
STDOUT.puts your_way "/../"; \
STDOUT.puts my_way "/../";
/
/
```

parkr:
I see no code. What's happening?

envygeeks:
@parkr my code is at the bottom of this post: https://github.com/jekyll/jekyll/pull/4167#issuecomment-158239083 his code is taken directly from here: https://github.com/jekyll/jekyll/commit/f81325cb1041a6c55c9602231fe65f349ebed38f it was a direct copy and paste with only comments stripped and changed to str, so you have a fair comparison:

```ruby
require "benchmark/ips"
def your_way(str)
  url = str. \
    split('/').reject { |s| s.empty? || s =~ /^\.+$/ }.join('/'). \
      gsub(/\A([^\/])/, '/\1')
  url << "/" if str.end_with?("/")
  url
end

def my_way(str)
  "/" + str.gsub(/\/{2,}/, "/").gsub(%r!\.+\/|\A/+!, "")
end

STDOUT.puts "you: #{your_way "../../../../../../hello/../../../../../world"}"
STDOUT.puts " me: #{  my_way "../../../../../../hello/../../../../../world"}"
Benchmark.ips do |x|
  x.report("you: ../../../../../../hello/../../../../../world") { your_way "../../../../../../hello/../../../../../world" }
  x.report(" me: ../../../../../../hello/../../../../../world") {   my_way "../../../../../../hello/../../../../../world" }
  x.compare!
end

STDOUT.puts "\n\n\n\n\n\n"
STDOUT.puts "you: #{your_way "../hello/////world"}"
STDOUT.puts " me: #{  my_way "../hello/////world"}"
Benchmark.ips do |x|
  x.report("you: ../hello/////world") { your_way "../hello/////world" }
  x.report(" me: ../hello/////world") {   my_way "../hello/////world" }
  x.compare!
end

STDOUT.puts "\n\n\n\n\n\n"
STDOUT.puts "you: #{your_way "../hello/world"}"
STDOUT.puts " me: #{  my_way "../hello/world"}"
Benchmark.ips do |x|
  x.report("you: ../hello/world") { your_way "../hello/world" }
  x.report(" me: ../hello/world") {   my_way "../hello/world" }
  x.compare!
end

STDOUT.puts "\n\n\n\n\n\n"
STDOUT.puts "you: #{your_way "/hello/world"}"
STDOUT.puts " me: #{  my_way "/hello/world"}"
Benchmark.ips do |x|
  x.report("you: /hello/world") { your_way "/hello/world" }
  x.report(" me: /hello/world") {   my_way "/hello/world" }
  x.compare!
end

STDOUT.puts "\n\n\n\n\n\n"
STDOUT.puts "you: #{your_way "hello/world"}"
STDOUT.puts " me: #{  my_way "hello/world"}"
Benchmark.ips do |x|
  x.report("you: hello/world") { your_way "hello/world" }
  x.report(" me: hello/world") {   my_way "hello/world" }
  x.compare!
end
```

parkr:
@envygeeks Want to submit a PR for your way? Seems like a nice way to clean up that method.

vwochnik:
@envygeeks I am sorry I forgot that the pull-request I have made also matches for `\.+` which I don't know why you guys have decided to strip `..` away since that is a valid path. Only single dots need to be removed.

What about `"/" + in_url.split('/').reject{ |s| s.empty? || s == '.' }.join('/')` plus the last condition? That should be a lot faster since the match in the reject block is gone.

envygeeks:
We strip "../" for security, well basic security, because Github needs us to.  They do not want you leaving the path they designate for you, and we are happy to do that for them.

envygeeks:
Yes, they run your instance in Docker too, but this is just an addition to that.

envygeeks:
Also, technically "." is a valid path too, but we don't worry about that because it won't go anywhere.

vwochnik:
Can be closed then. Your PR is better.

parkr:
Thanks, @vwochnik, for identifying the issue with this method! :)

envygeeks:
Yes, thank you so much!

PR #4169, opened 2015-11-20
Update plugins.md

vwochnik:
I have written a liquid filter to make email protection for static websites easy. This plugin contains the `protect_email` filter which encodes email addresses similar to the way GitHub does for the user profiles.


parkr:
Kewl. :shipit:

Issue #4028, opened 2015-10-14
Using more than one variable in {% include %} is not possible


omnidan:
I tried the following, which did not work:

```
{% include {{page.lang}}/features/{{feature.name}}.md %}
```

However, this works:

```
{% include {{page.lang}}/features/test.md %}
```

And so does this workaround:

```
{% capture filename %}{{page.lang}}/features/{{feature.name}}.md{% endcapture %}
{% include {{filename}} %}
```

parkr:
Code for this is here in case you want to contribute: https://github.com/jekyll/jekyll/blob/master/lib/jekyll/tags/include.rb

omnidan:
@parkr thanks for the answer - although I'm not really a ruby guy so I don't think I can do much there :grin:

parkr:
I wasn't a Ruby guy until I started writing Jekyll sites either! Once I decided I wanted to write a plugin, I said I would learn it. I don't regret that decision one bit!

omnidan:
I don't think I'll have time for it anytime soon, sorry :smile: but at least when someone else runs into this issue, (hopefully) they will find this thread and use the workaround (not get frustrated for a long time like me :sweat_smile:)

envygeeks:
I'm going to take this ticket, this is part of a larger problem with our tags I'm working on for any milestone > 3.0, we should be using Liquid to process the context to parse the variables.

envygeeks:
Shifting this to 3.2 as other things are more important.

PR #4183, opened 2015-11-25
Allow users to input multiple variables in include tag

rebornix:
I'm playing around Ruby regex these days and happen to see #4028 , this PR gives it a quick fix. 

@envygeeks might have a big refactor of Jekyll own tags referring to what he mentioned in #4028 , so it's ok if this doesn't get merged.

### DIFF
--- a/features/include_tag.feature
+++ b/features/include_tag.feature
@@ -89,3 +89,11 @@ Feature: Include tags
     Then I have an "_includes/one.html" file that contains "include content changed"
     When I run jekyll build
     Then I should see "include content changed" in "_site/index.html"
+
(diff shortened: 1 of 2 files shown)


parkr:
I'm OK with this change.

envygeeks:
Yeah, because the as is already documented we still have some work to do on our Liquid processing which is slightly more powerful but has a few quirks so the pull for my larger refactor of our tags to make every tag in (including user tags) easier won't land until jekyll-assets stabilizes it's liquid processing on tag arguments.

envygeeks:
I vote :shipit: because it could be 3.2 3.3 before that refactor even lands.

Issue #4068, opened 2015-10-28
Date for collections in front matter defaults is not working


iloveip:
Hello,

I upgraded to v3.0.0.pre.rc1. In my front matter defaults I have a date set for documents in a collection like this:

````
defaults:
  -
    scope:
      path: "" # empty string for all files
      type: "kb"
    values:
      layout: "post"
      date: "2015-09-15"
````

But it's not working. When I generate the site I get the following error:
````
jekyll 3.0.0.pre.rc1 | Error:  undefined method `strftime' for "2015-09-15":String
````

envygeeks:
Remove the quotes around the date.  You told YAML to create a string literal and YAML in Ruby will respect that, if you do not have quotes then YAML will attempt to convert it to a Date, DateTime or Time for you and pass it to us as that.

parkr:
We also process the date in `document.rb` entirely manually. You'll see in `#merge_data!` that if the date is there, it's parsed. That doesn't apply to the default proc.

iloveip:
@envygeeks If I remove the quotes around the date in my `_config.yml` file, it's still not working. The site is generating, but I can't see any of my posts.

parkr:
What's the output with `--trace`?

iloveip:
Hello @parkr, I've checked it once again, the date without quotes is working now. It's not working only if I put a future date. Thank you very much for your help and sorry for the trouble!

BlockWorld:
That is very weird that it would not work if it's a future date...

ducktyper:
Same as posts I found that collections requiring a conversion (with output: true) and having a future date will not be generated unless adding --future flag (jekyll build --future).
Is it correct behaviour or do collections with a future date need to be created by default?

envygeeks:
@ducktyper that's always been the behavior of Jekyll since I can remember, it might have changed a while back but certainly ever since I've been using it you either need to set `future: true` in your _config.yml or kick it on with the command line.

ducktyper:
@envygeeks Thanks for the quick answer.
I thought this issue was reopened because collections are not working (thought not generating) with future date.
If that is the correct behaviour then either I misunderstood the reason the issue was reopened or the meaning of not working.

envygeeks:
This issue was reopened because of the date quoting.

ducktyper:
Now I got it. So we need to allow quoted dates.
I want to solve this issue if I can.

envygeeks:
[...]

PR #4184, opened 2015-11-25
Allow quoted date in front matter defaults

ducktyper:
Fix to #4068

### DIFF
--- a/lib/jekyll/frontmatter_defaults.rb
+++ b/lib/jekyll/frontmatter_defaults.rb
@@ -30,6 +30,13 @@ module Jekyll
       set
     end
 
+    def parse_quoted_date(set)
+      return set unless set.key?('values') && set['values'].key?('date')
+      return set if set['values']['date'].is_a?(Time)
+      set['values']['date'] = Utils.parse_date(set['values']['date'], "Front matter defaults does not have a valid date.")
+      set
+    end
+
     # Finds a default value for a given setting, filtered by path and type
     #
     # path - the path (relative to the source) of the page, post or :draft the default is used in
@@ -159,7 +166,7 @@ module Jekyll
 
       sets.map do |set|
         if valid?(set)
-          update_deprecated_types(set)
+          parse_quoted_date(update_deprecated_types(set))
         else
           Jekyll.logger.warn "Defaults:", "An invalid front-matter default set was found:"
           Jekyll.logger.warn "#{set}"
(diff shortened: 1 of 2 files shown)


envygeeks:
:heart: for doing this @ducktyper

ducktyper:
Yes! that is the better name. Thanks for the quick feedback.

parkr:
This seems like a lack of flexibility on the users' parts. What if a user wants a string there and now we force a date? YAML parses dates just fine: just don't quote it.

envygeeks:
Personally if this were my site I would expect anything labled "date" to be turned into an Object and enforced as so, and if I want a string I would not use an internal front-matter variable that is expected to be a date.

ducktyper:
If I understand the code correctly, in my opinion, I believe the value of "date" from default front-matter and front-matter (of each post or collection file) need to be handled in the same way, because they both can be the value of data["date"] in a Document object which is used as a Time object.

Since Jekyll converts the value of "date" in front-matter to a Time object already in the same way, it is logical to do the same to the default front-matter unless that is not the desirable behaviour.

lib/jekyll/frontmatter_defaults.rb
```diff
@@ -30,6 +30,13 @@ def update_deprecated_types(set)
       set
     end
 
+    def ensure_time!(set)
+      return set unless set.key?('values') && set['values'].key?('date')
+      return set if set['values']['date'].is_a?(Time)
+      set['values']['date'] = Utils.parse_date(set['values']['date'], "Front matter defaults does not have a valid date.")
```
parkr: Can you indicate something more helpful in this error message? If you have a 200-line config.yml, this is going to be an annoying error.

ducktyper:
Now the error message shows which front-matter default set to look for by showing the set.
Also found a grammar mistake. Sorry for my unprofessional writing skill.
The invalid date is shown from Utils.parse_date method so I did not need to re-display it.

Here is an example error message that could be generated.

                    Invalid date '2000-01-0': An invalid date format was found in a front-matter default set: {"scope"=>{"path"=>""}, "values"=>{"layout"=>"post", "date"=>"2000-01-0"}}

Let me know if I missed something else or need more explanation.
Thanks for your feedback.

ducktyper:
I want to fix the bad test but could you enlighten me with little more detail?
I reviewed the tests and came with some ideas but I'm guessing I missed  something else.
1. Using .any? is more readable than .find with "assert" method because  .any? returns true/false while .find returns page/nil
2. Tests "not raise error" and "parse date" can be merged to reduce testing time since @site.process method takes some time to run.

PR #4191, opened 2015-11-27
Bug fix: extract title from filename when filename is dateless.

### LINKED ISSUE #4188: Title not working as expected / documented in collections not using dates in their filenames
If I create a collection named "portfolio" and use dates in my filenames like `2015-11-26-my-first-post.md` but do not set a title variable in the file YAML front-matter.

A `title` variable is available in Liquid. For example when doing this in Liquid, everything works as expected.

```liquid
{% for item in site.portfolio %}
  {% if forloop.first %}<ul>{% endif %}
    <li>{{ item.date }}  {{ item.title }}</li>
  {% if forloop.last %}</ul>{% endif %}
{% endfor %}
```

output

```html
<ul>
    <li>2015-11-26 00:00:00 +0100  My First Post</li>
</ul>
```

However, if I remove the date from the filename like so `my-first-post.md`, the `title` variable outputs nothing, although [the doc for collections](http://jekyllrb.com/docs/collections/) states that 

> If the document does not define a title in its front matter, this is equivalent to name.

Is this expected behaviour ?

rebornix:
Fixes #4188 . We forgot to extract slug/title from filename when it's dateless, if there is no title defined in FM as well, the title of this document will be empty.

Please notice, I add a new test folder called `_thanksgiving` instead of adding files to `_methods`. This is because our existing cases are heavily use For Loop `Collections: {% for method in site.methods %}{{ method.relative_path }} {% endfor %}"`. Changing `_methods` folder structure will ruin all existing cases.

Lastly, Happy Thanksgiving.


envygeeks:
Thanks!  I think this method needs to be reworked, the bug you are solving has pointed out smells in our source that need to be fixed, such as the code that doesn't speak for it self, the duplication that you added (which is actually doubling a smell,) and a few other things,  but it definitely needs to be reworked before, after or during this pull.

parkr:
> it definitely needs to be reworked before, after or during this pull.

@envygeeks Agreed, maybe merge this branch into one and clean it up if you have time?

PR #4194, opened 2015-11-28
Page 1 link sample code change

codykonior:
The code which shows a special link for page 1 doesn't appear to work, it always seems to point at the previous page. This is obvious when looking at page 8 and the "1" link points to page 7.

I've put something else which works for me locally. I'm not sure if there's a better way to do it.


envygeeks:
Please rebase for review, thanks!

PR #4198, opened 2015-11-29
Sample code for paginator page 1 link

codykonior:
The original code was showing previous page urls in the page 1 link; if you were on page 8 it would show the page 1 url as page 7. This code fix substitutes for the base url instead (plus a / in case you're running it locally with no baseurl).

### DIFF
--- a/site/_docs/pagination.md
+++ b/site/_docs/pagination.md
@@ -207,7 +207,7 @@ page with links to all but the current page.
     {% if page == paginator.page %}
       <em>{{ page }}</em>
     {% elsif page == 1 %}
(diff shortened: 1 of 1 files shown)


parkr:
The base of the pagination might not be `/`. It could be `/blog` or `/articles` and so on. So I'm not sure this works :/

codykonior:
It works if a site.baseurl is defined, and works in most cases where it's not defined. This is an improvement over not working at all. Do you have some other suggestion?

parkr:
Please see https://github.com/jekyll/jekyll/pull/4242#issuecomment-163791036

If a site.baseurl isn't defined, it should just prepend `""`.

PR #4210, opened 2015-12-02
Update windows.md

DavidBurela:
Added a very quick guide on getting Jekyll up and running on Windows in the fastest way possible with just a few command prompt commands.

### DIFF
--- a/site/_docs/windows.md
+++ b/site/_docs/windows.md
@@ -15,6 +15,8 @@ Julian Thilo has written up instructions to get
 The instructions were written for Ruby 2.0.0, but should work for later versions
 [prior to 2.2][hitimes-issue].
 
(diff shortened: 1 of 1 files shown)


parkr:
Great!

kaatt:
@DavidBurela Why is nodejs required?

parkr:
> @DavidBurela I think nodejs isn't required anymore. I don't have it installed and jekyll works for me.

@kaatt That's correct, NodeJS is only required for Jekyll 2.x, which depends on the CoffeeScript gem, which requires NodeJS. :)

DavidBurela:
That would make sense. I was following what documentation I could find for Windows which was for 2.x at the time and listed Node. Didn't realise it had been removed for 3.x. I'll update the instructions to bring it down to just 2 command prompt entries :grin:

PR #4233, opened 2015-12-07
Switch PluginManager to use require_with_graceful_fail

RochesterinNYC:


envygeeks:
I'm not against this I'm against the wording, how does this improve the error message that Ruby by default uses? And if it does that's a bug in Ruby not in Jekyll.

envygeeks:
Maybe you also want to just switch it to `Jekyll::External.require_with_graceful_fail`?

RochesterinNYC:
Okay, you're right, require_with_graceful_require looks like exactly the right error message we'd want. Thanks for the point in the right direction!

envygeeks:
@RochesterinNYC I would remove the "require gem" entirely and just just have that one single call, the graceful require does all that automatically, it's it's job.

RochesterinNYC:
Ah, gotcha. `require_with_graceful_fail` handles the require itself. So it was just an artifact/missed refactor that `require_gems` wasn't using this method? Updated the PR.

lib/jekyll/plugin_manager.rb
```diff
@@ -27,7 +27,7 @@ def require_gems
       site.gems.each do |gem|
         if plugin_allowed?(gem)
           Jekyll.logger.debug("PluginManager:", "Requiring #{gem}")
-          require gem
+          Jekyll::External.require_with_graceful_fail([gem])
         end
```
parkr: Could also improve this to be

```ruby
Jekyll::External.require_with_graceful_fail(site.gems.select { |g| plugin_allowed?(g) })
```

What do you think?

RochesterinNYC: That looks like a cleaner change. If we want to one line this, is there still a point to having the require_gems method? It's only used on line 19 (PluginManager#conscientious_require) so we could just use this one-liner there? How do we feel about that? Something like:
```
    def conscientious_require
      require_plugin_files
      Jekyll::External.require_with_graceful_fail(site.gems.select { |g| plugin_allowed?(g) })
      deprecation_checks
    end
```

RochesterinNYC: I also notice that require_plugin_files is still using `require` without the graceful_fail method.

envygeeks: All this debate is a small, that file needs to be reviewed.

parkr: > is there still a point to having the require_gems method?

Absolutely! Code clarity. This whole thing is for GitHub after all, for their security team to ensure that non-whitelisted gems are not required. So if we can encapsulate this as its own method, it's easier to audit (and to understand that line's purpose).

> I also notice that require_plugin_files is still using require without the graceful_fail method.

Good point. You can fix that up too if you would like.

RochesterinNYC: Okay, valid point, makes sense. I do notice that this would remove the Jekyll.logger.debug if we wanted to one-line it. Would this break backwards compatibility if anyone is asserting or relying on these debug messages?

parkr: We make no guarantees, so they should not rely on it. Is there a debug message in require_with_graceful_fallback?

RochesterinNYC: Nope, only a logger.error message about the Dependency Error in case of `LoadError` in `require_with_graceful_fail`. 

Think it would be valuable to include a Jekyll.logger.debug message about which gem is being required in `require_with_graceful_feedback`, provided that we remove any redundant logger.debug messages in places that use `require_with_graceful_fail`. And looking through the codebase, there don't seem to be any.

parkr: Make it so!

RochesterinNYC:
Thanks for all the feedback. Updated the PR to use `require_with_graceful_fail` in both require methods in `PluginManager` and added a helpful debug log message to `require_with_graceful_fail` stating which plugin is being required.

lib/jekyll/plugin_manager.rb
```diff
@@ -70,10 +65,9 @@ def whitelist
     # Returns nothing.
     def require_plugin_files
       unless site.safe
-        plugins_path.each do |plugins|
-          Dir[File.join(plugins, "**", "*.rb")].sort.each do |f|
-            require f
-          end
+        plugins_path.each do |plugin_search_path|
+          plugin_files = Dir[File.join(plugin_search_path, "**", "*.rb")]
```
parkr: Let's do `Utils.safe_glob(plugin_search_path, "**/*.rb")` here instead :)

parkr:
@RochesterinNYC Thank you for being cool with all the back-and-forth. :heart:

lib/jekyll/plugin_manager.rb
RochesterinNYC: Sure, that sounds good. Is there still a need for using `File.join` if we're passing in explicit parameters for the call? (i.e. it'll always just return `"**/*.rb"`

parkr: @RochesterinNYC Yes, the `File.join` returns `"**/*.rb"` on Linux and Mac, but will return `"**\*.rb"` on Windows. :)

RochesterinNYC:
Updated the PR to use `Utils#safe_glob`. Thanks for all the helpful feedback!

parkr:
Well done, @RochesterinNYC! :tada: Thanks for your contribution.

RochesterinNYC:
Thanks for the guidance and review!

PR #4234, opened 2015-12-07
Rewrite the warning to be clearer. Fixes #4199

### LINKED ISSUE #4199: Unclear warning: Deprecation: Collection#each should be called on the #docs array directly
My plugin contains the following loop:

        site.posts.each do |post|
          # do something
        end

I get the following warning on that line:

`Deprecation: Collection#each should be called on the #docs array directly.`

I have no idea what it wants from me, I'm not using any `#docs` anywhere...

Is this a bug? Or is the warning just not clear at all?

chrisfinazzo:
See the issue for discussion on this PR. Perhaps there should be a change to the [Upgrade][] guide to reflect this?

/cc @jekyll/core

[Upgrade]: http://jekyllrb.com/docs/upgrading/2-to-3/

### DIFF
--- a/lib/jekyll/collection.rb
+++ b/lib/jekyll/collection.rb
@@ -32,7 +32,7 @@ module Jekyll
     # Override of method_missing to check in @data for the key.
     def method_missing(method, *args, &blck)
       if docs.respond_to?(method.to_sym)
-        Jekyll.logger.warn "Deprecation:", "Collection##{method} should be called on the #docs array directly."
+        Jekyll.logger.warn "Deprecation:", "Collection##{method} should be called on posts#docs instead of on #posts."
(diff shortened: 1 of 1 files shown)


envygeeks:
I agree on both points.

parkr:
We could update the docs, I think. I am happy to word this better, but it may not _just_ apply to posts so hardcoding `posts` in there is not good. Maybe to `#{label}##{method}` if anything.

envygeeks:
We should probably also tap `caller_locations` to print out at least the last caller to them so that it makes the message even more relevant if they don't understand.  They could be using that method in dozens of places and we can help them narrow it down for them and point them to the problem spot.

PR #4242, opened 2015-12-09
Update pagination.md

babaleo:
The current example dealing with implementing a 'Page 1' link has an issue.

The link created by line 210 does not lead to page 1, instead it leads to the previous page.

I have modified the href value to point to the root URL (which is page 1 in most instances) to correct this issue.

### DIFF
--- a/site/_docs/pagination.md
+++ b/site/_docs/pagination.md
@@ -207,7 +207,7 @@ page with links to all but the current page.
     {% if page == paginator.page %}
       <em>{{ page }}</em>
     {% elsif page == 1 %}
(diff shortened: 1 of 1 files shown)


parkr:
Hey, @babaleo. Thanks for your pull request. We actually want this to be like this – the jekyll-paginate library handles this automatically here: https://github.com/jekyll/jekyll-paginate/blob/a2f213532b0f92420f24aad0ea7249bbe6d5feb8/lib/jekyll-paginate/pagination.rb#L56-L68

Thanks, though! If there is a bug in jekyll-paginate, let's fix it there.

PR #4243, opened 2015-12-09
Link to less generic FormKeep page

r00k:
This link used to point to a generic landing page. Now it links to a guide written specifically for Jekyll users.

PR #4244, opened 2015-12-09
Add utm params to link in docs

r00k:
Just a small thing. Feel free to close if it's a bother.

#4243 updated this link to a jekyll-specific page. This commit adds utm params to the link so we can tell that users came to us from the Jekyll documentation. Among other things, this will help us provide better support to Jekyll users who sign up, since we'll know what site generator they're using.


parkr:
I'm cool with it. :+1: Thanks for the neat service, @r00k.

PR #4254, opened 2015-12-11
For blessed gems, shim their commands so users know how to use them.

parkr:
Basically doing #4253 but for Jekyll 3.

### DIFF
--- a/bin/jekyll
+++ b/bin/jekyll
@@ -40,8 +40,11 @@ Mercenary.program(:jekyll) do |p|
       puts p
       abort
     else
-      unless p.has_command?(args.first)
-        Jekyll.logger.abort_with "Invalid command. Use --help for more information"
+      subcommand = args.first
+      unless p.has_command? subcommand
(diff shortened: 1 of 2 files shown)


envygeeks:
What if we did this for all gems instead of through commands? If Jane adds `gem: jekyll-this-gem` then we throw with the message: "unable to require jekyll-this-gem" have you installed `"jekyll-this-gem"? Most of the time we can infer what the gem name is reliably from the require.

parkr:
> If Jane adds gem: jekyll-this-gem then we throw with the message: "unable to require jekyll-this-gem" have you installed `"jekyll-this-gem"?

Yes, we do! Anywhere we use [`Jekyll::External.require_with_graceful_fail`](https://github.com/jekyll/jekyll/blob/3effae59d8a8ab0b78e3dba19f4d13aa9eea03d6/lib/jekyll/external.rb#L32-L56), we are doing that. This PR is for subcommands.

PR #4255, opened 2015-12-12
Upgrading 2-3: Added missing links to Pygments.rb and Rouge

ofalvai:
I think @parkr wanted to link to the projects, just forgot to add the URLs.


parkr:
Thanks!

PR #4267, opened 2015-12-16
Proper markup for coffeescript requirements

txels:
Replaced `*` with `-`

### DIFF
--- a/site/_docs/assets.md
+++ b/site/_docs/assets.md
@@ -85,8 +85,8 @@ here, too.
 
 To enable Coffeescript in Jekyll 3.0 and up you must
 
- * Install the `jekyll-coffeescript` gem
- * Ensure that your `_config.yml` is up-to-date and includes the following
+ - Install the `jekyll-coffeescript` gem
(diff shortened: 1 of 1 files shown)


parkr:
Hey @txels, thanks for the contribution! We are using `*` to denote lists in Markdown in most of our other markdown files (if not all of them), so we're going to stick with that. Thanks, though!

pathawks:
Using `*` is fine, but there is _something_ wrong with the way this list is rendered. Which is to say, it is **not** rendered as a list.

coffeescript

pathawks:
The issue seems to be the space preceding the `*`

If that is removed, it will render as a list.

parkr:
Excellent observation, @pathawks! Publishing now.

pathawks:
Everything is looking good now :+1:

Your @vscobot intrigues me

coffeescript

parkr:
> Your @vscobot intrigues me

Haha yeah, I ran something at [work](http://vsco.co) and it set my global git user!

PR #4269, opened 2015-12-17
WIP: Incremental regen improvements

alfredxing:
Done my exams! :sweat_smile:

The main change here is the addition of a `content` tag, which is equivalent to using something like `{{ post.content }}`, except it plays better with incremental regeneration.

Basically, when used, it adds the referenced document/page as a dependency in the regen metadata, and manually renders the document/page if it's not already rendered.

This is still very much a work in progress (except for the first commit, which is just additional tests).

**_Definite shortcoming**_: I don't think this handles the case of adding posts; we'll probably need a whole separate system for that.

A couple of questions, too:
- Should this be a filter, not a tag?
- Would Liquid Drops make it possible to put this in accesses to `post.content` directly instead? (@envygeeks?)


alfredxing:
Usage would be the same as the current uses for `post.content`:

```
{% for post in site.posts %}
  {% content post %}
{% endfor %}
```

Here, with incremental regen on, if a post is modified, the page with the above code would be regenerated as well.

***However,*** I don't think this handles the case of adding posts; we'll probably need a whole separate system for that.

envygeeks:
I'm a bit resistant on a hefty logic change like this in 3.x because we already ask a lot of users to move to 3.x, now we are asking even more in 3.1 or 3.2 but my resistance will be ended if we do move to drops because that will solve the core of this problem without asking anything of the user other than to upgrade.

parkr:
Moving to drops will break the merge_payload hook unless we convert to hash then create a new drop from that hash.

envygeeks:
@parkr nothing stops us from adding basic hash methods onto Drop and passing that drop instance as the "hash" since we would need to store a basic internal hash of static stuff anyways.

rebornix:
Changing the payload to Drop can allow us inject code into accessor ( we did this while using DotLiquid ), and this is the cleanest way for users, they don't need to make any content change. I vote for this one instead of creating another tag/filter.

alfredxing:
@envygeeks The rationale I have for this right now is that:
1. Incremental regeneration is disabled by default, but this improves the experience for users who have it enabled
2. No existing functionality is broken; this won't affect any existing site in any way (unless they happen to have specified `type` in their frontmatter, which is quite minor anyway).

So, as far as I can tell, this is a Pareto improvement in economic/game theory terms (at least one person is better off and nobody is worse off).

That said, if using Drops will make things easier, I would vote for going that route as well. We'll see once we get at least basic Drop functionality working. I'm definitely not going to merge this until then.

alfredxing:
_Re. the tests_: This seems kind of on the border to me. I'm fine either way. Putting the tests in unit allows us to be a bit more specific about things (for example, checking to see if the dependency was added explicitly by checking to see if it's in the array), but putting it as a feature test might be a bit more idiomatic...

parkr:
> So, as far as I can tell, this is a Pareto improvement in economic/game theory terms (at least one person is better off and nobody is worse off).

Ha! You have most certainly just emerged from the depths of finals. Hope they went well! :smiley: 

Let's go the Drop route. I'll write up something over Christmas and see if I can come up with an acceptable solution. I will add the usual `#[]` and `#[]=` methods as well for easy accessing of non-method keys. Anything else you'd recommend?

alfredxing:
Sounds good!

There's not much else... Maybe moving pages to Renderer if that's possible?

parkr:
> Maybe moving pages to Renderer if that's possible?

That would be excellent!

parkr:
@alfredxing what's the status of this?

alfredxing:
I haven't have time yet to look into this again since Drops. I'll see if I can take advantage of anything there.

alfredxing:
@parkr I moved all the logic into DocumentDrop, reflecting our shift to Drops. Everything is a lot simpler!

Using Drops has the benefit of a lot more context&mdash;the Drop knows what document is being requested, in addition to having access to the Liquid context (thus also knowing what page/document is requesting it).

However, I haven't touched any of the Page stuff yet (so usages of `{{ page.content }}` won't add a dependency yet). Do we want to move pages into a collection as well, or was that plan dropped? I'm mostly in favour of it since it would likely simplify the rendering process by quite a bit.

I was thinking in order to achieve this logically, we can make a new `PageCollection < Collection` class that contains all of the defaults needed to make everything compatible and output to the right place... anyway, this is probably for another discussion :wink:

parkr:
@alfredxing Thanks so much for doing this! :sparkles: Glad to hear the move to Drops has helped this endeavour.

> Do we want to move pages into a collection as well, or was that plan dropped

Yes, let's do that for 3.2. It's relatively simple. Once that's done, `Layout` is the last to be migrated. Ideally everything would be a document and belong to a collection based on its purpose. If everything is a document, then everything can go through the renderer. 

> we can make a new PageCollection < Collection class that contains all of the defaults needed to make everything compatible and output to the right place

We achieve this for posts [in the configuration](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/configuration.rb#L278-L280). Could that work?

> I plan on adding another front-matter option so users can specify if a document depends on another manually.

Yes. Please. Huzzah!!!!!! :tada:

parkr:
@alfredxing What is the status of this pull request? Would you like me to take another look?

alfredxing:
@parkr I was going to wait until #4412 was done to have things simpler for the incremental regen improvements (including cache). But if you want I can continue work on this without waiting!

parkr:
@alfredxing OK. We'll put this on hold then until #4412 is finished. Thanks!

jhabdas:
Now depends on #4822 and not #4412. Still a WIP. Should these changes still be dependent on the other, or should first man win?

envygeeks:
@jhabdas #4822 is independent of that IMO.

alfredxing:
@envygeeks With the way I'm refactoring and enhancing Incremental, it would be really nice not to deal with Pages and Documents being completely different, especially when it comes to things like Drops. So it's not directly dependent but something that would be great to have in before this.

parkr:
Hey @alfredxing! I know you're probably gearing up for finals here soon but would you have a bit of time to split this out into smaller pieces? People _love_ the incremental regenerator and I'd love to pull these updates in but I dropped the ball here and let this get stale... Let me know!

alfredxing:
@parkr Just finished my finals 😅  I'll see if I can work on this a bit over the break.

I think there are basically only two changes here:
- Linking a document/drop as a dependency when its `content` or `output` is referenced
- Adding some more tests for incremental regen

So this should solve the issue where a page (usually the blog index page) includes post content.

pathawks:
@alfredxing What's the status on this? Is this something that we can merge into Jekyll 4.0?

alfredxing:
😭 Oof, this is an old one. I haven't been keeping up to date on the project, so I've lost most of the context I would say. From what I remember this would still _ideally_ be blocked by #4822 since it would simply & unify how documents and Pages are handled. That one is a breaking change though since it breaks Page subclassing, but we could probably consider it for a major version bump.

pathawks:
@alfredxing I would like to take a stab at getting this up-to-date. Could you please allow edits from maintainers on this PR and #4822?

alfredxing:
@pathawks Thanks for taking this on! I've checked the box 🙂

lib/jekyll/drops/drop.rb
```diff
@@ -193,6 +193,32 @@ def merge!(other)
...
       end
 
+      # Adds the resource behind this Drop as a dependency to whatever resource
+      # is rendering it. Paths are taken from the Liquid context available to
+      # the drop
+      #
+      # Returns nothing.
+      def link_dependency
+        site = @context.registers[:site]
+        requirer = @context.registers[:page]
```
pathawks: I don't see where `@context` is set. 😕

alfredxing: IIRC `@context` is set by Liquid when it's rendering the Drop! https://github.com/Shopify/liquid/blob/v3.0.6/lib/liquid/template.rb#L176

Issue #3824, opened 2015-06-26
Docs for "Creating Pages" Should Mention Markdown Support


alphang:
I'm new to Jekyll. While reading http://jekyllrb.com/docs/pages/, I found myself thinking that only static HTML pages are supported, because of the way it's is worded.

Can those docs be expanded to mention _layouts/page.html, as well as dynamic content (Markdown etc) support?

Thanks.

gynter:
Technically markdown is not a dynamic content, it's a markup language which is easier to write and read also as plain text. All content (posts, pages and collections) are run trough markdown parser. More information is available at the [Welcome](http://jekyllrb.com/docs/home/) page. Also, layouts and includes are not parsed with markdown parser.

alphang:
Thanks. I didn't get that "All content (posts, pages, and collections) are run through the markdown parser." in the documentation. It's not clear on the Welcome page, either.

plaindocs:
I can pick this up if nobody else wants to add docs for it.

parkr:
@plaindocs I don't think anyone is working on it. Please feel free! Even just mentioning `index.md` or something after some carefully-chosen `index.html`s could be helpful enough here.

pathawks:
Looks like this was fixed via #4275 :+1:

parkr:
Thanks @pathawks :)

PR #4275, opened 2015-12-21
Adding markdown examples to Pages

plaindocs:
Fix for #3824

### DIFF
--- a/site/_docs/pages.md
+++ b/site/_docs/pages.md
@@ -81,7 +81,7 @@ might look like:
 ├── contact/
 |   └── index.html  # => http://example.com/contact/
 |── other/
-|   └── index.md    # => http://example.com/other.html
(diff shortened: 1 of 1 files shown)


site/_docs/pages.md
```diff
@@ -77,6 +80,8 @@ might look like:
 |   └── index.html  # => http://example.com/about/
 ├── contact/
 |   └── index.html  # => http://example.com/contact/
+|── other/
+|   └── index.md    # => http://example.com/other.html
```
parkr: I don't think this URL is correct. I believe it should be `http://example.com/other/`, no?

plaindocs: Dammit, you're right. Fix on the way.

envygeeks:
:shipit:

PR #4277, opened 2015-12-22
Use Liquid::Drops instead of Hashes in #to_liquid

### LINKED ISSUE #3224: Use Liquid::Drop to liberate and speed up Jekyll.
This is probably something for the Jekyll rewrite but recently while working on an an app we had performance problems even with user caching because of the way we were processing the liquid and I discovered `Liquid::Drop` (https://github.com/Shopify/liquid/wiki/Introduction-to-Drops) and while a lot of people use it for security we actually used it to our advantage because of the way it was designed.  We used it to defer processing of data by routing in and routing back out to the object that got routed in which increased our sites performance exponentially (you're talking going from 500ms to the average load of 20ms)

That basically means that a Page has Posts, the Page gets to_html ran with :vars => { :posts => [] } with the array being a bunch of Post drops and those post drops are just references to the objects we want to allow on Post (say title, blah, blah, blah) but there is one difference, normally we would pre-render content of those posts, with a drop we create a content method that allows us to defer the processing until the user wants it processed and when the user sends a message to the content method on the drop we then finally route `#content` back to `Post#content_as_markdown`.

Here let me show you in code because I'm bad at explaining: https://gist.github.com/envygeeks/52aadf5ad9166eec11dd

The above is a working example of how fast it could be, it creates 1000 posts and defers the content until the user wants it (rendering and reading!), but writes 12 of those posts, tell me it isn't fast.  To test it against Jekyll in actuality I would have to rewrite the parts of current Jekyll to use Drops (which wouldn't be that hard to do but would have to wait until I have time over the next few days.)

Advantages:

1. #3116 by @alfredxing can get exponentially faster and/or more efficient in regeneration.
2. We have a single place for people to look at when people want to see what data goes into Jekyll.
3. We have an entry point to start allowing people to insert data into Jekyll easier.

Disadvantages:

It doesn't really provide a boost to Jekyll as a whole other than making it "feel more responsive" because it will load tons faster but it could provide some benefits if mixed with with the way @alfredxing wants it to work and with the way that Jekyll is currently designed because incremental generation in both Jekyll and life incremental generation become significantly faster because of all the deferring, we are only reading what has actually changed and not reading everything upfront.

---

What I mean by all the above is that in total it probably wont help us much unless we mix it with incremental generation, and if we do, things just got tons faster because of the deferring alone, we could do that without drops, but we could do it cleaner and in a more encapsulated extensible way with drops because then we add data points to the drops, not to Jekyll's core.

parkr:
This requires test updates but it compiles all the (rudimentary) sites on my machine.

Fixes #3224.

@jekyll/core, please review.

### DIFF
--- a/features/collections.feature
+++ b/features/collections.feature
@@ -13,7 +13,7 @@ Feature: Collections
     And the "_site/methods/configuration.html" file should not exist
 
   Scenario: Rendered collection
-    Given I have an "index.html" page that contains "Collections: {{ site.collections }}"
+    Given I have an "index.html" page that contains "Collections: output => {{ site.collections[0].output }} label => {{ site.collections[0].label }}"
     And I have an "collection_metadata.html" page that contains "Methods metadata: {{ site.collections[0].foo }} {{ site.collections[0] }}"
     And I have fixture collections
     And I have a "_config.yml" file with content:
@@ -25,8 +25,8 @@ Feature: Collections
     """
     When I run jekyll build
     Then the _site directory should exist
-    And I should see "Collections: {\"output\"=>true" in "_site/index.html"
-    And I should see "\"label\"=>\"methods\"," in "_site/index.html"
+    And I should see "Collections: output => true" in "_site/index.html"
+    And I should see "label => methods" in "_site/index.html"
     And I should see "Methods metadata: bar" in "_site/collection_metadata.html"
     And I should see "<p>Whatever: foo.bar</p>" in "_site/methods/configuration.html"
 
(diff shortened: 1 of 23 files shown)


parkr:
@envygeeks How do I run this new code through Rubocop?

envygeeks:
You're on OS X so you'll need to install Docker: https://www.docker.com/docker-toolbox (the preferred Method for OS X.) If that doesn't work I'll create a bot that does this all automatically and posts it to the pull request and flags it as needs-manual-review.

alfredxing:
:+1: Looks good to me so far!

parkr:
Down to 3 failures & 6 errors on the test spec. Cucumber specs run and all pass, though there's a strange issue in the reporter on travis. Maybe a cached version or something that is causing issues.

parkr:
@jekyll/core I need to finish writing tests for these drops but otherwise we should be good to go :+1: 

Please review this PR.

alfredxing:
:+1: :tada:

PR #4281, opened 2015-12-26
drops: provide #to_h to allow for hash introspection

parkr:
Follow-up to #4277


parkr:
The duplication in this bothers me.

Issue #4218, opened 2015-12-04
Split up the Rakefile


envygeeks:
The Rakefile as it stands is hard to mantain, hard to understand; this is a bug we should split it up the way large projects normally do; Rails gets one thing right, splitting up the Rakefiles so they are easier to manage and maintain.

parkr:
How would you propose this? I think the rakefile is easy to understand.

envygeeks:
The same way Rails does it, create a rake folder in lib and fill it with `{namespace,task}.rake`  The reason I find it confusing is because it contains lots of logic that isn't grouped, it's pretty large and hard to spot tasks, splitting it up maintains that logic in the file it belongs to making it easier to spot stuff.

envygeeks:
This isn't finished yet, there is still work to be done.

envygeeks:
Sorry, sometimes my bot is mean, this is finished as far as I'm concerned.

PR #4282, opened 2015-12-26
moved namespaced rake tasks to separate .rake files under lib/tasks

leethomas:
Hi, I saw https://github.com/jekyll/jekyll/issues/4218 where @envygeeks wanted to split the Rakefile so I did so based on namespaces. I named all of the .rake files by their root namespace (i.e. docs.rake for namespace :docs). Not sure if that's exactly how you guys wanted it, but let me know if not and I'll make the changes.

No tests with this PR since it's just a refactor. All the existing tests passed.

Also merry Christmas, if you celebrate it. :santa: :gift: :christmas_tree:


envygeeks:
It this exactly what i was on about, just one minor comment, I would split up tasks without a namespace the same way I split up namespaces since the current form implies analysis is part of documentation, but it's actually it's own task that runs CodeClimate coverage and Rubocop.

lib/tasks/docs.rake
```diff
@@ -0,0 +1,60 @@
...
+    end
+
+    while data = err.gets
+      file.write data
+      $stderr.print ansi.red(data)
+    end
+  end
+
+  file.close
+end
```
parkr: let's put this in its own `.rake` file

leethomas: Hah, I was just about to push a commit which moved the analysis task out of docs.rake. Should I open another PR or just push it here?

PR #4287, opened 2015-12-26
Failing test: Utils.deep_merge_hashes

pathawks:
I believe this to be related to jekyll/jekyll-archives#54

This test passes on **3.0.1**, but fails on `master`.

Please tell me if there is a better place to put this, or a better way to organize this. I have no idea what I am doing. :rage3:

### DIFF
--- a/test/test_utils.rb
+++ b/test/test_utils.rb
@@ -1,6 +1,21 @@
 require 'helper'
 
 class TestUtils < JekyllUnitTest
+  context "The \`Utils.deep_merge_hashes\` method" do
+    setup do
+      clear_dest
+      config = Jekyll::Configuration::DEFAULTS.merge({'source' => source_dir, 'destination' => dest_dir})
+
+      @site = fixture_site(config)
+      @site.process
+    end
+
+    should "merge a page into the site" do
+      data = {"page" => {}}
+      assert Utils.deep_merge_hashes(data, @site.site_payload)
+    end
+  end
+
   context "hash" do
 
     context "pluralized_array" do


test/test_utils.rb
```diff
@@ -1,6 +1,21 @@
 require 'helper'
 
 class TestUtils < JekyllUnitTest
+  context "The \`Utils.deep_merge_hashes\` method" do
```
envygeeks: Test Smell - If using `rspec` do `describe "#method"` if using MiniTest do `context "#method"`.

pathawks: I do not understand. I was following the convention used elsewhere in this file.

* [`context "The Utils.parse_date method" do`](https://github.com/jekyll/jekyll/blob/master/test/test_utils.rb#L67)
* [`context "The Utils.slugify method" do`](https://github.com/jekyll/jekyll/blob/master/test/test_utils.rb#L100)
* [`context "The Utils.add_permalink_suffix method" do`](https://github.com/jekyll/jekyll/blob/master/test/test_utils.rb#L184)
* [`context "The Utils.safe_glob method" do`](https://github.com/jekyll/jekyll/blob/master/test/test_utils.rb#L199)
* etc.

Can you explain more precisely what you are looking for?

envygeeks: I'll digress that test to @jekyll/core, but the convention is always (if you test properly):

```ruby
describe ClassName do
  context "#method" do
    # Tests
  end
end
```

However, we do not fully utilize the convention of what we are using, we go between Minitest Spec and Minitest Unit in ways that prevent inferring context properly, which should change here soon... hopefully.

This test however is enforcing a context driven testing environment.

envygeeks: Let me add, I don't quite remember if Minitest has descriptions in descriptions.

envygeeks: It looks like it does: https://github.com/seattlerb/minitest#specs so the latter example is implied.

parkr:
This is a result of the movement to Drops.

pathawks:
> This is a result of the movement to Drops.

I figured. Is this an unintentional temporary regression, or will **jekyll-archives** need to do things differently?

envygeeks:
> I figured. Is this an unintentional temporary regression, or will jekyll-archives need to do things differently?

This is a minor release so it *shouldn't* have to, it should be able to remain exactly as is if it worked on 3.0 otherwise we failed at our jobs.

PR #4289, opened 2015-12-27
utils/drops: update Drop to support Utils.deep_merge_hashes

parkr:
Fixes #4287.

Thanks for finding this bug, @pathawks!


pathawks:
**Thank you!**

PR #4301, opened 2015-12-30
Trying to appease Rubocop

pathawks:
Am going through [this list of style issues](https://codeclimate.com/github/jekyll/jekyll/issues/categories/style) and tackling the low hanging fruit. If this is productive, I may continue.

This is mostly for my own learning.

### DIFF
--- a/lib/jekyll/commands/build.rb
+++ b/lib/jekyll/commands/build.rb
@@ -11,7 +11,7 @@ module Jekyll
 
             add_build_options(c)
 
-            c.action do |args, options|
+            c.action do |_, options|
               options["serving"] = false
               Jekyll::Commands::Build.process(options)
             end
--- a/lib/jekyll/commands/build.rb
+++ b/lib/jekyll/commands/build.rb
@@ -1,9 +1,7 @@
 module Jekyll
   module Commands
     class Build < Command
-
       class << self
-
         # Create the Mercenary command for the Jekyll CLI for this Command
         def init_with_program(prog)
           prog.command(:build) do |c|
@@ -71,9 +69,7 @@ module Jekyll
           External.require_with_graceful_fail 'jekyll-watch'
           Jekyll::Watcher.watch(options)
         end
-
       end # end of class << self
-
     end
   end
 end
--- a/lib/jekyll/commands/build.rb
+++ b/lib/jekyll/commands/build.rb
@@ -65,7 +65,7 @@ module Jekyll
         # options - A Hash of options passed to the command
         #
         # Returns nothing.
-        def watch(site, options)
+        def watch(_site, options)
           External.require_with_graceful_fail 'jekyll-watch'
           Jekyll::Watcher.watch(options)
         end
--- a/lib/jekyll/commands/doctor.rb
+++ b/lib/jekyll/commands/doctor.rb
@@ -50,7 +50,8 @@ module Jekyll
           urls = {}
           urls = collect_urls(urls, site.pages, site.dest)
           urls = collect_urls(urls, site.posts.docs, site.dest)
-          urls.select { |_, p| p.size > 1 }.each do |url, paths|
+          urls.each do |url, paths|
+            next unless paths.size > 1
             conflicting_urls = true
             Jekyll.logger.warn "Conflict:", "The URL '#{url}' is the destination" \
               " for the following pages: #{paths.join(", ")}"
@@ -77,7 +78,8 @@ module Jekyll
         def urls_only_differ_by_case(site)
           urls_only_differ_by_case = false
           urls = case_insensitive_urls(site.pages + site.docs_to_write, site.dest)
-          urls.select { |_, p| p.size > 1 }.each do |_, real_urls|
+          urls.each do |case_insensitive_url, real_urls|
+            next unless real_urls.uniq.size > 1
             urls_only_differ_by_case = true
             Jekyll.logger.warn "Warning:", "The following URLs only differ" \
               " by case. On a case-insensitive file system one of the URLs" \
--- a/lib/jekyll/commands/doctor.rb
+++ b/lib/jekyll/commands/doctor.rb
@@ -105,7 +105,7 @@ module Jekyll
         end
 
         def case_insensitive_urls(things, destination)
-          things.inject(Hash.new) do |memo, thing|
+          things.inject({}) do |memo, thing|
             dest = thing.destination(destination)
             (memo[dest.downcase] ||= []) << dest
             memo
--- a/lib/jekyll/commands/doctor.rb
+++ b/lib/jekyll/commands/doctor.rb
@@ -2,7 +2,6 @@ module Jekyll
   module Commands
     class Doctor < Command
       class << self
-
         def init_with_program(prog)
           prog.command(:doctor) do |c|
             c.syntax 'doctor'
@@ -108,7 +107,6 @@ module Jekyll
           end
         end
       end
-
     end
   end
 end
--- a/lib/jekyll/commands/doctor.rb
+++ b/lib/jekyll/commands/doctor.rb
@@ -58,7 +58,7 @@ module Jekyll
           conflicting_urls
         end
 
-        def fsnotify_buggy?(site)
+        def fsnotify_buggy?(_site)
           return true unless Utils::Platforms.osx?
           if Dir.pwd != `pwd`.strip
             Jekyll.logger.error "  " + <<-STR.strip.gsub(/\n\s+/, "\n  ")
--- a/lib/jekyll/commands/doctor.rb
+++ b/lib/jekyll/commands/doctor.rb
@@ -10,7 +10,7 @@ module Jekyll
 
             c.option '--config CONFIG_FILE[,CONFIG_FILE2,...]', Array, 'Custom configuration file'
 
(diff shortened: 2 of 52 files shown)


envygeeks:
Do you plan on doing more @pathawks because as far as I'm concerned as it stands right now this is ready to be merged and perfect!

envygeeks:
Well... asked too late since as I was asking you did more lol.

pathawks:
> Well... asked too late since as I was asking you did more lol.

lol
Yes, I plan to work on this a bit more over the next couple of days. There are some failing tests that I need to investigate :rage4:

pathawks:
Failing tests seem unrelated. Drops have no method `key?`, which caused a wide variety of problems... :goberserk:

pathawks:
What is the best way to do a big PR like this? Should I just keep committing and then squash+merge when I finish / get bored?
Or you could merge this batch and I’ll just open another PR for the next batch?

envygeeks:
Squash likes but keep others as their own commits so there is a clear history.  So for example you have two commits that have "key?" changes, those two should be squashed but everything else that isn't the same of another commit is easier to verify over multiple commits in this case.

lib/jekyll/frontmatter_defaults.rb
```diff
@@ -130,7 +130,7 @@ def valid?(set)
     # new_scope - the new scope hash
     #
     # Returns true if the new scope has precedence over the older
-    def has_precedence?(old_scope, new_scope)
+    def precedence?(old_scope, new_scope)
```
pathawks: `precedence?` is [`private`](https://github.com/jekyll/jekyll/pull/4301/files#diff-68dac361a7950a8a653b234e362ba78aR79), so renaming is not a breaking change.

envygeeks: I'm gonna caution on the side of it not being entirely clear to the user,  things like Rubocop tell you to only use `private` once but that's an ignorant approach considering I didn't even know it was private until I scrolled up quite a ways.  They also tell you to indent on that private, which just muddies the code view.

I'm not against the change though, just saying it might not be clear to a user.

pathawks: Do you have a suggestion for making it more clear?

envygeeks: For all my source, I disable Rubocops "worthless private" warning and add private above every private def no matter how many there are, so it's clear that method is private, the same way other languages enforce that kind of explicitness.

```ruby
private
def method1
  #
end

private
def method2
  #
end
```

envygeeks: Actually, I think I already disabled that in Jekyll too.

pathawks: Ok, I see that this convention is already implemented in most other parts of Jekyll.

pathawks:
Before:
> ```text
> 119 files inspected, 1853 offenses detected
> ```

After:
> ```text
> 119 files inspected, 1066 offenses detected
> ```

pathawks:
After b7d5a26:

> ```
> 69 files inspected, 606 offenses detected
> ```

I will weed some of these commits out when I get the chance.

envygeeks:
It's gone down slightly more after further refinement, now it sits at 533 offenses as it stands.

pathawks:
Down to 321, and I'm not sure why the test failed (only on 2.2)

> ```text
> Failure:
TestKramdown#test_: kramdown should render fenced code blocks with syntax highlighting.  [/home/travis/build/jekyll/jekyll/test/test_kramdown.rb:59]
Minitest::Assertion: Failed refutation, no message given
553 tests, 926 assertions, 1 failures, 0 errors, 0 skips
> ```

envygeeks:
It randomly fails on 2.2 because of bugs in 2.2 that were addressed in the latest 2.2 but Travis+RVM is absolutely totally broken so 2.2 points to the first release of 2.2.  To fix it we have to point it to the full version of 2.2.* which is annoying to maintain.

parkr:
Keep everything as individual commits, I think. I have encountered too many general "fix the style" commits for my liking.

lib/jekyll/commands/build.rb
```diff
@@ -67,13 +65,11 @@ def build(site, options)
         # options - A Hash of options passed to the command
         #
         # Returns nothing.
-        def watch(site, options)
+        def watch(_site, options)
```
parkr: you can make this just `_`, I think?

lib/jekyll/commands/doctor.rb
```diff
@@ -52,17 +51,16 @@ def conflicting_urls(site)
...
+            conflicting_urls = true
+            Jekyll.logger.warn "Conflict:", "The URL '#{url}' is the destination" \
+              " for the following pages: #{paths.join(", ")}"
           end
           conflicting_urls
         end
 
-        def fsnotify_buggy?(site)
-          return true if !Utils::Platforms.osx?
+        def fsnotify_buggy?(_site)
```
parkr: why pass the site if it's not used?

lib/jekyll/configuration.rb
```diff
@@ -19,7 +18,7 @@ class Configuration < Hash
       'safe'          => false,
       'include'       => ['.htaccess'],
       'exclude'       => [],
-      'keep_files'    => ['.git','.svn'],
+      'keep_files'    => ['.git', '.svn'],
```
parkr: this could be the `%w` syntax

lib/jekyll/drops/url_drop.rb
```diff
@@ -35,17 +35,49 @@ def categories
...
-      def month;       @obj.date.strftime("%m"); end
-      def day;         @obj.date.strftime("%d"); end
-      def hour;        @obj.date.strftime("%H"); end
-      def minute;      @obj.date.strftime("%M"); end
-      def second;      @obj.date.strftime("%S"); end
-      def i_day;       @obj.date.strftime("%-d"); end
-      def i_month;     @obj.date.strftime("%-m"); end
-      def short_month; @obj.date.strftime("%b"); end
-      def short_year;  @obj.date.strftime("%y"); end
-      def y_day;       @obj.date.strftime("%j"); end
```
parkr: i really prefer this shorter syntax – what's the rule it violates?

lib/jekyll/url.rb
```diff
@@ -93,7 +92,7 @@ def generate_url_from_drop(template)
     # as well as the beginning "/" so we can enforce and ensure it.
 
     def sanitize_url(str)
-      "/" + str.gsub(/\/{2,}/, "/").gsub(%r!\.+\/|\A/+!, "")
+      "/" + str.gsub(/\/{2,}/, "/").gsub(/\.+\/|\A\/+/, "")
```
parkr: is `%r` not allowed?

parkr:
Thanks, @pathawks!

lib/jekyll/commands/build.rb
envygeeks: You can if you wish but he probably left it _site so the API is clear.

lib/jekyll/url.rb
envygeeks: Both `%r!!` and `//` are allowed.

lib/jekyll/drops/url_drop.rb
envygeeks: Using a semicolon to terminate a line.

```
def year() @obj.date.strftime("%Y") end
```

lib/jekyll/commands/build.rb
pathawks: > he probably left it _site so the API is clear.

Yup

lib/jekyll/drops/url_drop.rb
pathawks: [**Avoid single-line method definitions**](https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/style/single_line_methods.rb)

parkr:
Down to 204 offenses. Thanks so much, @pathawks!

lib/jekyll/url.rb
pathawks: [`.rubocop.yml` prefers slashes](https://github.com/jekyll/jekyll/blob/master/.rubocop.yml#L11)

pathawks:
Thank you both for helping me through this. I learned a lot along the way :+1:

lib/jekyll/url.rb
envygeeks: If we kick out `AllowInnerSlashes: true` then it will go on and require you to do %r!! for clarity, sound fine?

lib/jekyll/drops/url_drop.rb
envygeeks: I'll disable this rule, I personally liked the way @parkr grouped the methods.

lib/jekyll/commands/doctor.rb
pathawks: Didn't want to do anything that would appear to change the API, since this is a public method.

I suppose it could be changed to `def fsnotify_buggy?(_ = nil)` and then not pass `site`?

lib/jekyll/url.rb
parkr: :+1:

PR #4307, opened 2016-01-03
Escape html from site.title and page.title

fulldecent:


envygeeks:
I'm not against this but I would like to know what end this serves? I don't think it should be our job to encapsulate and correct the mistakes of users.

fulldecent:
In the Front Matter Variables documentation https://jekyllrb.com/docs/variables/, the title is defined as 

 > The title of the Page.

which does not mention HTML-escaping the string. This implies that escaping/encoding would be the responsibility of downstream. 

This is also consistent with the XML outputting: https://github.com/jekyll/jekyll/blob/2807b8a012ead8b8fe7ed30f1a8ad1f6f9de7ba4/lib/site_template/feed.xml

    <title>{{ site.title | xml_escape }}</title>

This PR updates Jekyll HTML output to be consistent with documentation and other code and also enforcing a separation of content and presentation.

envygeeks:
I don't agree with the applied logic here but I'm neutral so I'll leave it to others to decide.

Issue #4246, opened 2015-12-09
YAML Front Matter not attached to `page` in layout


dalanmiller:
Also posted in #2836 

If I set YAML Front Matter in a layouts file `_layouts/post.html` like so:

```
--- 
layout: default
body_class: blog
---
```

I should be able to access `body_class` via `page.body_class` in the `_layouts/default.html` file, correct? This isn't happening currently in Jekyll 3.0.1 and the value results in `nil` and renders as empty space. This code was also working as expected in Jekyll 2.2. 

Interestingly, normal usage of `page.body_class` is working, such as in an `{% if .. %}` statement and other places. However, in this specific case (within a string?) it is failing:

```
<div class="{{page.body_class}}">
```

However, `<div class="{{ layout.body_class }}">` is working. Can someone explain why this is?

budparr:
I think this is what you're looking for: http://jekyllrb.com/docs/upgrading/2-to-3/#layout-metadata

parkr:
Thanks, @budparr! You're right. @dalanmiller We changed this in 3.0 to be `layout.body_class` because no one could decide whether page data should overwrite layout data, or the other way around. So we just separated it.

dalanmiller:
Thank you for the quick reply guys @budparr & @parkr! Is there a reason why some variables are still attached to `page`? Like in this instance, `page.body_class` works in pretty much every other instance except the one I showed above.

dalanmiller:
I also that if I change that specific line to `layout.body_class` it breaks my template in `./blog/index.html` whose header looks like this:

```
---
layout: default
title: Blog
active: blog
body_class: blog
---
```
This file works with `page.body_class` not `layout.body_class`. So there's something different happening when the inheriting template is another layout or a non-layout. Is there maybe something wrong with the way I have my directory setup? 

```
_layouts
├── default.html
└── post.html
blog
├── index.html
```

dalanmiller:
I think I now understand that the value is under `layout` when the layout inherits another layout, but under `page` when a page inherits from layout `_layouts/default.html`.

envygeeks:
That just broke my brain.

parkr:
@dalanmiller I believe this was fixed by #4205. It was inconsistent, but will be fixed in 3.1.

dalanmiller:
As a stopgap/jerry-rig, could I do something like this?

```
{% if layout.body_class.nil? || layout.body_class == "" %}
[...]

PR #4312, opened 2016-01-04
Drops should use content setters if they exist

parkr:
@evgenyneu, this fixes your issue as defined here: https://github.com/jekyll/jekyll/issues/4246#issuecomment-168367510

### DIFF
--- a/features/layout_data.feature
+++ b/features/layout_data.feature
@@ -1,0 +1,18 @@
+Feature: Layout data
+  As a hacker who likes to avoid repetition
+  I want to be able to embed data into my layouts
+  In order to make the layouts slightly dynamic
+
+  Scenario: Use custom layout data
+    Given I have a _layouts directory
+    And I have a "_layouts/custom.html" file with content:
+      """
+      ---
+      foo: my custom data
+      ---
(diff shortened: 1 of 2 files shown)


evgenyneu:
Thank you, I have added this change, but it does not fix [my issue](https://github.com/jekyll/jekyll/issues/4246#issuecomment-168367510).

features/layout_data.feature
```diff
@@ -0,0 +1,18 @@
...
+  In order to make the layouts slightly dynamic
+
+  Scenario: Use custom layout data
+    Given I have a _layouts directory
+    And I have a "_layouts/custom.html" file with content:
+      """
+      ---
+      foo: my custom data
+      ---
+      {{ content }} foo: {{ layout.foo }}
```
evgenyneu: In my case I was doing `{{ layout.foo }}` from parent layout file `index.html`, not from `custom.html`.

parkr: See a9b80b2 for this. It works on master according to my scenario.

evgenyneu:
Yes, you are right, @parkr, thank you. I needed to install jekyll from master to make it work.

PR #4314, opened 2016-01-05
Allow custom file extensions if defined in `permalink` YAML front matter

### LINKED ISSUE #4182: Unable to output .php files with 3.x
https://github.com/mpc-hc/mpc-hc.org/pull/58#issuecomment-159450803

All the info are on that PR and repo; let me know if something else is needed.

parkr:
Still a failing test. Fixes #4182 once it's passing.

What I'm trying to do is have `output_ext` return `File.extname(doc.permalink)` in the case that the doc has a permalink. This means that checking if the path ends with the output extension will always return true, thus skipping the addition.

I'm running into an issue where `permalink = /slides/example-slide-7.php`, but `output_ext = .html`, so it's being appended. If I print the document, the data hasn't been read, which is awfully strange. `permalink = nil`, actually, when we ask for the `output_ext` for the first time. Super weird.

@envygeeks, have a second to look at this? Super weird – why would `@site.process` in the tests not yield a fully-read document?

### DIFF
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -298,11 +298,11 @@ module Jekyll
       when :pretty
         "/:categories/:year/:month/:day/:title/"
       when :none
-        "/:categories/:title.html"
+        "/:categories/:title:output_ext"
       when :date
-        "/:categories/:year/:month/:day/:title.html"
+        "/:categories/:year/:month/:day/:title:output_ext"
       when :ordinal
-        "/:categories/:year/:y_day/:title.html"
+        "/:categories/:year/:y_day/:title:output_ext"
       else
         permalink_style.to_s
       end
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -4,7 +4,7 @@ module Jekyll
   class Document
     include Comparable
 
-    attr_reader :path, :site, :extname, :output_ext, :collection
+    attr_reader :path, :site, :extname, :collection
     attr_accessor :content, :output
 
     YAML_FRONT_MATTER_REGEXP = /\A(---\s*\n.*?\n?)^((---|\.\.\.)\s*$\n?)/m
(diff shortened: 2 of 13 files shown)


envygeeks:
Your problem surfaces here: https://github.com/jekyll/jekyll/blob/master/lib/jekyll/cleaner.rb#L56.  Cleaner messages `Document#destination` before `Document#read` is done, resulting in `Document#permalink` returning `nil` which leads to `converters.first.output_ext` being messaged with `document.extname` instead of `File.extname` with `document.permalink` in `Renderer#output_ext` which leads to `Document@output_ext` being cached with ".html"

envygeeks:
To add, the `Document` is actually being read and is fully populated.   `Document@output_ext` just ended up being cached early leading to this order problem.  Your source works, it's higher up the chain an order problem surfaces a bug.

parkr:
> Cleaner messages Document#destination before Document#read is done, resulting in Document#permalink returning nil which leads to converters.first.output_ext being messaged with document.extname instead of File.extname with document.permalink in Renderer#output_ext which leads to Document@output_ext being cached with ".html"

I think you're right, but I'm not sure why quite yet.

Consider `Site#process`:

```
    def process
      reset
      read
      generate
      render
      cleanup
      write
      print_stats
    end
```

`cleanup` is called after `read`, and `read` has to complete before `cleanup` can be called. `cleanup` calls `Cleaner#cleanup!` which calls `Cleaner#obsolete_files` which calls `Cleaner#new_dirs` which calls `Cleaner#new_files`, which calls, for each site document, `Document#destination`. So I'm not really sure what's up. Especially because the failing tests use `Site#process`...

parkr:
Only test failures are the ones currently on master:

```text
Failure:
TestKramdown#test_: kramdown when a custom highlighter is chosen should support legacy enable_coderay... for now.  [/home/travis/build/jekyll/jekyll/test/test_kramdown.rb:93]
Minitest::Assertion: Failed refutation, no message given
Failure:
TestKramdown#test_: kramdown when a custom highlighter is chosen should use the chosen highlighter if it's available.  [/home/travis/build/jekyll/jekyll/test/test_kramdown.rb:73]
Minitest::Assertion: Failed refutation, no message given
```

parkr:
@jekyllbot: merge +bug

PR #4332, opened 2016-01-09
WIP: Make jekyll awesome

envygeeks:
This is a pre-finished pull to start the discussion on everything but the specs.


envygeeks:
Notes:
* There was a pretty hefty bug with the ObjectSpace dumping, in that it didn't calculate for Jekyll::Site happening multiple times as the singleton was created, which means it never calculated for the singleton at all... this ultimately leads to people not being able to do Jekyll::Site twice, even if they only have one Jekyll object to begin with.  Meaning doing a = Jekyll::Site, b = Jekyll::Site results in an error.  That bug was discovered when building the Jekyll Mocker for RSpec.  I was kind of surprised it didn't happen with the tests as they were, but regardless, I switched it to something far less hefty... especially for Jekyll where we have lots of Objects.

* The move to /bin/sh is arbitrary but it will help us test Jekyll on our Docker image.
* The removal of the ruby profiling stuff was because there is one better library for memory (the one that Discourse built) and I plan to rewrite the Jekll profiling as a wrap on Jekyll inside of the commands so that anybody can profile Jekyll with their site and give us better information.  I thought that would be awesome.

* The rewrite of @benbalter script/vendor-mimes happened because I got confused when I first opened it, I figured if I got confused then somebody else trying to contribute might be even more confused.  Sorry @benbalter, I'm not trying to tread on your territory, just trying to help.

* The cleanup of the Rake tasks was because it was a bit cluttered and like some other stuff, I got confused, so I split a few bits off, condensed a few bits and added a few more files.  To be honest though, the current Rake tasks are kind of insane to manage, some of those tasks are extra-large and would do well with some encapsulation, but it's not my place to decide that.

This ultimately will kick off the great move to RSpec, I finally got fed up when I couldn't get those tests to pass again so I started doing all this, over the next few days I'll finish it out.

script/stackprof
```diff
@@ -1,22 +0,0 @@
...
-TEST_SCRIPT="Jekyll::Commands::Build.process({'source' => 'site'})"
-PROF_OUTPUT_FILE=tmp/stackprof-${STACKPROF_MODE}-$(date +%Y%m%d%H%M).dump
-
-echo Stackprof Mode: $STACKPROF_MODE
-test -f "$PROF_OUTPUT_FILE" || {
-  bundle exec ruby -r./lib/jekyll -rstackprof \
-    -e "StackProf.run(mode: :${STACKPROF_MODE}, interval: 100, out: '${PROF_OUTPUT_FILE}') { ${TEST_SCRIPT} }"
-}
-
-bundle exec stackprof $PROF_OUTPUT_FILE $@
```
parkr: Please add this back until there is a replacement.

script/rubyprof
```diff
@@ -1,19 +0,0 @@
...
-  printer = RubyProf::CallTreePrinter.new(result)
-  filename = "tmp/ruby_prof_#{rand 10000}"
-  puts "Writing profile to #{filename}"
-  file = File.open(filename, "w")
-  printer.print(file, {})
-  file.close
-RUBY
-)
-
-bundle exec ruby -r ./lib/jekyll -e "${RUBY}"
```
parkr: I use this too!

script/stackprof
envygeeks: This pull isn't finished, statements like this should wait.

script/rubyprof
envygeeks: Again, this pull isn't finished, wait! It will be put back in *this* pull, it was removed in this pull and will be rebuilt as part of the Jekyll command in this pull.

script/console
```diff
@@ -1,38 +0,0 @@
...
-  test_dir('source', *subdirs)
-end
-
-def test_dir(*subdirs)
-  File.join(TEST_DIR, *subdirs)
-end
-
-module Jekyll
-  binding.pry
-end
```
parkr: Is there an easy way to access these kinds of helpers through the method you think is better? They help with debugging.

envygeeks: There is, I plan on creating `pry-jekyll` on our Org which will add a lot of these as actual Pry commands with other cool stuff so you can even build your site and stuff directly through `pry-jekyll`

envygeeks:
Jekyll is already aweome.  It just needs more love.  I'm gonna split this up into multiple pull request.

PR #4323, opened 2016-01-07
Add smartify Liquid filter for SmartyPants

pathawks:
Is there a better place to put the `Kramdown::Parser::SmartyPants` class?

Am I on the right track with this?

PR #4333, opened 2016-01-10
Add documentation for smartify Liquid filter

pathawks:
Add documentation for jekyll/jekyll#4323

Sorry, this probably should have been part of that PR.

### DIFF
--- a/site/_docs/templates.md
+++ b/site/_docs/templates.md
@@ -188,6 +188,17 @@ common tasks easier.
     </tr>
     <tr>
       <td>
+        <p class="name"><strong>Smartify</strong></p>
+        <p>Convert "quotes" into &ldquo;smart quotes.&rdquo;</p>
+      </td>
+      <td class="align-center">
+        <p>
+         <code class="filter">{% raw %}{{ page.title | smartify }}{% endraw %}</code>
+        </p>
+      </td>
+    </tr>
+    <tr>
+      <td>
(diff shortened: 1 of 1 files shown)


parkr:
Ah, yes! Thank you. :)

parkr:
@jekyllbot: merge +site

jhabdas:
@pathawks et. al, would it be useful to also indicate in the documentation `smartify` will handle converting `--` into an ndash? Also, I noticed the filter works with `page.title` but not `content`. If that's expected, is it also worth mentioning?

pathawks:
> Also, I noticed the filter works with page.title but not content. If that's expected, is it also worth mentioning?

Can you provide a bit more detail?

jhabdas:
> Can you provide a bit more detail?

`{{ content | smartify }}` results in `--` output to page where content contains `--`. I tried this at both the `_default.md` layout with and without using the [`_compress` layout](http://jch.penibelst.de/).

jhabdas:
Ah, that makes good sense. Thanks @pathawks

pathawks:
@jhabdas You could open a PR if you think this is worth mentioning in the docs :+1:

jhabdas:
Sounds good to me. Need to grok our discussion in https://github.com/gettalong/kramdown/pull/299 a little more before I do. Thanks for staying on top of these issues and all of your helpful feedback, @pathawks.

PR #4318, opened 2016-01-05
Reorganize and cleanup the Gemfile, shorten required depends.

jekyllbot:
PR automatically created for @envygeeks.

```text
Reorganize and cleanup the Gemfile, shorten required depends.
```

PR #4335, opened 2016-01-10
Revert "Reorganize and cleanup the Gemfile, shorten required depends."

envygeeks:
Reverts jekyll/jekyll#4318

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-bundler_args: --without doc:util:site:benchmark:development
 language: ruby
 cache: bundler
 sudo: false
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,4 +1,3 @@
-bundler_args: --without doc:util:site:benchmark:development
```
parkr: Could be this line - maybe a dependency is needed that is included here.

envygeeks: I thought maybe that too, I'm gonna play with it in the morning and see how it broke it in the Travis-CI image.

parkr: Cool. 

Anyway, thanks for fixing this!!

PR #4338, opened 2016-01-10
Add script/travis so all people can play with Travis-CI images.

envygeeks:


script/travis
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
```
parkr: Would you mind adding usage info up here?

envygeeks: I'm gonna be honest, this is for programmers so if they can't get usage info from the simple 2 if's they shouldn't be using it.  I'm on the fence about whether I wish to comply with this because the script is so simple.

envygeeks: Let me rephrase that, did you understand how to use it @parkr?

parkr: I think it's simple enough, but usage examples save time. It took me a second to understand what $1 and $2 were supposed to do because I had to hunt. I sometimes have to get usage help for things because I use so many tools every day that I can mix them up sometimes.

envygeeks: That's fair then, after I finish up the ObjectSpace removal I'll add a usage info.

envygeeks:
Done.

parkr:
You rock! :tada:

parkr:
@jekyllbot: merge +dev

PR #4341, opened 2016-01-10
Remove script/rebund.

envygeeks:
It is my understanding that `script/redund` is a legacy script from before the days of Travis bundle caching, but now Travis has inline caching and we already take advantage of that, we should remove this legacy script before somebody tries to use it and gets confused.

You served well `script/rebund`, it's time to retire.


parkr:
@jekyllbot: merge +dev

PR #4342, opened 2016-01-10
Remove ObectSpace dumping and start using inherited, it's faster.

envygeeks:


envygeeks:
Hulk: benchmark +old-v-new : Prove to me my theory is right.

envygeeks:
I have done as you requested:

```
                   A   204.000  i/100ms
                   B     6.063k i/100ms
-------------------------------------------------
                   A      2.019k (±10.6%) i/s -      9.996k
                   B     70.913k (± 1.0%) i/s -    357.717k

Comparison:
                   B:    70913.0 i/s
                   A:     2018.6 i/s - 35.13x slower
```

Here is my source because I know you bang on about it constantly being important:

```ruby
require "benchmark/ips"
require "set"

class A
  def self.descendants
    descendants = []
    ObjectSpace.each_object(singleton_class) do |k|
      descendants.unshift k unless k == self
    end
    descendants
  end
end

class B
  def self.inherited(const)
    return catch_inheritance(const) do |const_|
      catch_inheritance(const_)
    end
  end

  #

  def self.catch_inheritance(const)
    const.define_singleton_method :inherited do |const_|
      (@children ||= Set.new).add const_
      if block_given?
        yield const_
      end
    end
  end

  #

  def self.descendants
    @children ||= Set.new
    out = @children.map(&:descendants)
    out << self unless superclass == B
    Set.new(out).flatten
  end
end

Object.const_set(:A1, Class.new(A))
Object.const_set(:A2, Class.new(A))
Object.const_set(:A3, Class.new(A))
Object.const_set(:A4, Class.new(A))
Object.const_set(:A5, Class.new(A))
Object.const_set(:A6, Class.new(A))
Object.const_set(:A7, Class.new(A))
Object.const_set(:B1, Class.new(B))
Object.const_set(:B2, Class.new(B))
Object.const_set(:B3, Class.new(B))
Object.const_set(:B4, Class.new(B))
Object.const_set(:B5, Class.new(B))
Object.const_set(:B6, Class.new(B))
Object.const_set(:B7, Class.new(B))

Benchmark.ips do |x|
  x.report("A") { A.descendants }
  x.report("B") { B.descendants }
  x.compare!
end
```

You started the race slow but came out ahead so far I'm pretty sure you rigged the race car some how. You won though until I can prove otherwise.

lib/jekyll/site.rb
```diff
@@ -263,25 +263,21 @@ def site_payload
...
-    #
-    # klass - The Class of the Converter to fetch.
-    #
     # Returns the Converter instance implementing the given Converter.
+    # klass - The Class of the Converter to fetch.
+
     def find_converter_instance(klass)
-      converters.find { |c| c.class == klass } || proc { raise "No converter for #{klass}" }.call
+      converters.find { |klass_| klass_.instance_of?(klass) } || \
+        raise("No Converters found for #{klass}")
```
parkr: We should raise some Jekyll-specific error here.

lib/jekyll/generator.rb
```diff
@@ -1,4 +1,5 @@
 module Jekyll
   class Generator < Plugin
+    #
```
parkr: Please remove

parkr:
Way faster! Thanks! :tada:

parkr:
@jekyllbot: merge +dev

PR #4347, opened 2016-01-11
Rearrange Cucumber and add some flair.

envygeeks:
* Move step_definitions/jekyll.rb to just step_definitions.rb
* Rename the formatter to Jekyll::Cucumber::Formatter, it's Jekyll's.
* Add some flair; switch to checks!
* Rename env.rb to helpers.rb

---

* The renaming of `env.rb` to `helpers.rb` facilitates what it actually does and hints at people contributing as to what it does and where to place stuff if they are inside of Cucumber.
* The moving of `step_definitions` is because it removes a single file folder.
* The flair is just because.


parkr:
@jekyllbot: merge +dev

mattr-:
I'm surprised cucumber doesn't break without a `step_definitions` folder.

parkr:
@mattr- _shrug_

envygeeks:
@mattr- I was surprised too, I was adding Cucumber to an app I was building and said "man this single folder file is ugly" so I tried it and it worked so I was like YAY for organization.

PR #4348, opened 2016-01-12
Add default Liquid filter for default values

pathawks:
Instead of

```html
{% if page.title %}
   <title>{{ page.title }}</title>
{% else %}
   <title>Default Title</title>
{% endif %}
```

This filter will allow:
```html
<title>{{ page.title | default: "Default Title" }}</title>
```

This filter is [included in the Liquid 4 release candidate](https://github.com/Shopify/liquid/blob/v4.0.0.rc1/lib/liquid/standardfilters.rb#L344-L347), but I would suspect that it will be a while before Jekyll can move to a new major version of Liquid.

- [x] Add Filter
- [x] Add Tests
- [x] Add Documentation

### DIFF
--- a/test/test_filters.rb
+++ b/test/test_filters.rb
@@ -430,5 +430,19 @@ class TestFilters < JekyllUnitTest
       end
     end
 
+    context "defualt" do
+      should "be overridable" do
+        assert_equal "foo", @filter.default("foo", "bar")
+      end
+
+      should "not be overridden by empty values" do
+        assert_equal "bar", @filter.default(nil, "bar")
+        assert_equal "bar", @filter.default("", "bar")
+        assert_equal "bar", @filter.default(false, "bar")
+        assert_equal "bar", @filter.default([], "bar")
+        assert_equal "bar", @filter.default({}, "bar")
+      end
+    end
+
   end
 end
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -311,6 +311,19 @@ module Jekyll
       CGI.escapeHTML(input.inspect)
     end
 
+    # Returns a input if it is not blank, default_value otherwise
+    #
+    # This filter is included in Liquid 3.0
+    # https://github.com/Shopify/liquid/blob/v3.0.0/lib/liquid/standardfilters.rb#L286-L289
+    def default(input, default = EMPTY_STR)
+      return default unless input
+      if input.respond_to?(:empty?) && input.empty?
+        default
+      else
+        input
+      end
+    end
+
     private
     def time(input)
       case input
--- a/site/_docs/templates.md
+++ b/site/_docs/templates.md
@@ -79,6 +79,17 @@ common tasks easier.
     </tr>
     <tr>
       <td>
+        <p class="name"><strong>Default</strong></p>
+        <p>If the input is blank, the output will be the given string. Otherwise, the input is passed through.</p>
+      </td>
+      <td class="align-center">
+        <p>
+         <code class="filter">{% raw %}{{ post.author | default:"Guest Writer" }}{% endraw %}</code>
+        </p>
+      </td>
+    </tr>
+    <tr>
+      <td>
         <p class="name"><strong>Where</strong></p>
         <p>Select all the objects in an array where the key has the given value.</p>
       </td>


parkr:
> I would suspect that it will be a while before Jekyll can move to a new major version of Liquid.

The major release of Liquid may not affect us. We may be able to bump in a 3.2 instead of in a 4.0

test/test_filters.rb
```diff
@@ -430,5 +430,19 @@ def to_liquid
       end
     end
 
+    context "defualt filter" do
```
parkr: `default`

parkr:
@pathawks I'm fine adding this to v3.1.0 if you have time today/tomorrow to add documentation. There's a page for the filters.

pathawks:
Yes, I will add docs today/tomorrow. :+1:

pathawks:
Hold on, this looks like it is already [in Liquid 3](https://github.com/Shopify/liquid/blob/v3.0.0/lib/liquid/standardfilters.rb#L286-L289).

Why does it only work if we include it again in Jekyll? Shouldn't it come with [Liquid 3](https://travis-ci.org/jekyll/jekyll/jobs/104242121#L267)

pathawks:
I feel like this is already included in Liquid and we shouldn't have to duplicate it here, but since the [tests do not pass if we do not include our own `default` filter](https://travis-ci.org/jekyll/jekyll/builds/104242120), I must be mistaken.

Either way, docs have been added and this PR is ready for review by @jekyll/core

envygeeks:
That code irked me because it was so slow for something so simple (and I know you didn't write it @pathawks) but it's much faster as:

```ruby
def default(str, default = ''.freeze)
  if !str || str == ''.freeze
    return default
  end

  str
end
```

---

#### Ruby 2.3

```
Comparison:
        _new:nil,nil:  3813840.2 i/s
        _old:nil,nil:  2651507.6 i/s - 1.44x slower

Comparison:
      _new:nil,hello:  3173889.5 i/s
      _old:nil,hello:  2301255.8 i/s - 1.38x slower

Comparison:
       _new:'',hello:  2262343.6 i/s
       _old:'',hello:  2119516.2 i/s - 1.07x slower
```

#### Ruby 2.2

```
Comparison:
        _new:nil,nil:  3503265.1 i/s
        _old:nil,nil:  2105375.9 i/s - 1.66x slower

Comparison:
      _new:nil,hello:  2254230.7 i/s
      _old:nil,hello:  1586608.5 i/s - 1.42x slower

Comparison:
       _new:'',hello:  1580059.5 i/s
       _old:'',hello:  1499492.3 i/s - 1.05x slower
```

And you can bump the last two up by another 0.15x if you move `"".freeze` to a constant.

envygeeks:
I see why they are using empty now, let me adjust my code.

envygeeks:
Shiny object distracted me, here is a faster version the way it currently is:

```ruby
def _new(obj, default = EMPTY_STR)
  return default unless obj
  if obj.respond_to?(:empty?) && obj.empty?
    default else obj
  end
end
```

envygeeks:
And the benches if you wish to see them:

```
Comparison:
        _new:nil,nil:  4154265.8 i/s
        _old:nil,nil:  2660080.4 i/s - 1.56x slower

Comparison:
      _new:nil,hello:  3062159.8 i/s
      _old:nil,hello:  2261820.6 i/s - 1.35x slower

Comparison:
       _new:'',hello:  2120534.5 i/s
       _old:'',hello:  2047534.6 i/s - 1.04x slower

Comparison:
       _new:[],hello:  2059903.1 i/s
       _old:[],hello:  2044665.0 i/s - 1.01x slower

Comparison:
       _new:{},hello:  2133132.5 i/s
       _old:{},hello:  2040050.0 i/s - 1.05x slower
```

pathawks:
Thanks @envygeeks!

parkr:
> I feel like this is already included in Liquid and we shouldn't have to duplicate it here, but since the tests do not pass if we do not include our own default filter, I must be mistaken.

Decided to give this a shot myself and it turns out that it does work! The reason it doesn't work for the tests is because `@filter` refers only to the `JekyllFilters` module, not to all Liquid filters.

```bash
$ echo -e "---\n---\n{{ page.title | default:'Default title' }}" > hi.html
$ jekyll serve
$ open http://localhost:4000/hi.html
# You should see "Default title"
```

I just did this and it works!

screen shot 2016-01-23 at 5 30 18 pm

Thanks for putting all the hard work in, @pathawks, and thanks @envygeeks for finding that performance improvement! We'll let the Liquid folks handle it from here.

pathawks:
> turns out that it does work!

It has been a long week.

Thanks for the sanity check :p

PR #4350, opened 2016-01-12
Clean up the Gemfile (and keep all the necessary dependencies)

envygeeks:
The problem last time was that we removed Pry and Pry brings in CodeRay, we were
testing legacy stuff and didn't have CodeRay in our dependencies, which resulted
in those tests failing.

This also quietly announces the intention to move to RSpec by moving the old
test dependencies to ":test_legacy" and is slightly less agressive in it's
organization than before.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
-cache: bundler
-script: script/cibuild
+bundler_args: --without benchmark:site:development
 before_script: bundle update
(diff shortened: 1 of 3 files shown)


parkr:
Great! Thank you for this. I'll add `pr/` to the list of accepted prefixes for autopull, too!

@jekyllbot: merge +dev

PR #4362, opened 2016-01-16
Upgrade to Liquid v4

parkr:
RC1 for now.

@fw42 @dylanahsmith, know what's going on here? `.parse` isn't working like before.

``` text
Error:
TestTags#test_: highlight tag in unsafe mode should recognize the hl_linenos option and its value. :
NoMethodError: undefined method `line_number' for {}:Hash
    /Users/parkr/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/liquid-4.0.0.rc1/lib/liquid/tag.rb:21:in `initialize'
    /Users/parkr/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/liquid-4.0.0.rc1/lib/liquid/block.rb:4:in `initialize'
    /Users/parkr/jekyll/jekyll/lib/jekyll/tags/highlight.rb:14:in `initialize'
    /Users/parkr/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/liquid-4.0.0.rc1/lib/liquid/tag.rb:9:in `new'
    /Users/parkr/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/liquid-4.0.0.rc1/lib/liquid/tag.rb:9:in `parse'
    test/test_tags.rb:46:in `highlight_block_with_opts'
    test/test_tags.rb:92:in `block (2 levels) in <class:TestTags>'
    /Users/parkr/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/shoulda-context-1.2.1/lib/shoulda/context/context.rb:413:in `instance_exec'
    /Users/parkr/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/shoulda-context-1.2.1/lib/shoulda/context/context.rb:413:in `block in create_test_from_should_hash'
```

### DIFF
--- a/History.markdown
+++ b/History.markdown
@@ -36,6 +36,7 @@
   * Update Core team list in the README file (#5643)
   * Improve Permalinks documentation. (#5653)
   * Fix typo in Variables doc page (#5657)
+  * Fix a couple of typos in the docs (#5658)
 
 ## 3.3.1 / 2016-11-14
 
--- a/History.markdown
+++ b/History.markdown
@@ -35,6 +35,7 @@
   * use backticks for Gemfile for consistency since in the next sentence … (#5641)
   * Update Core team list in the README file (#5643)
   * Improve Permalinks documentation. (#5653)
+  * Fix typo in Variables doc page (#5657)
 
 ## 3.3.1 / 2016-11-14
 
--- a/History.markdown
+++ b/History.markdown
@@ -34,6 +34,7 @@
   * Fixed typo (#5632)
   * use backticks for Gemfile for consistency since in the next sentence … (#5641)
   * Update Core team list in the README file (#5643)
+  * Improve Permalinks documentation. (#5653)
 
 ## 3.3.1 / 2016-11-14
 
--- a/jekyll.gemspec
+++ b/jekyll.gemspec
@@ -29,7 +29,7 @@ Gem::Specification.new do |s|
   s.rdoc_options = ['--charset=UTF-8']
   s.extra_rdoc_files = %w[README.markdown LICENSE]
 
-  s.add_runtime_dependency('liquid',    '~> 3.0')
+  s.add_runtime_dependency('liquid',    '4.0.0.rc1')
   s.add_runtime_dependency('kramdown',  '~> 1.3')
   s.add_runtime_dependency('mercenary', '~> 0.3.3')
   s.add_runtime_dependency('safe_yaml', '~> 1.0')
--- a/jekyll.gemspec
+++ b/jekyll.gemspec
@@ -29,7 +29,7 @@ Gem::Specification.new do |s|
   s.rdoc_options = ['--charset=UTF-8']
   s.extra_rdoc_files = %w[README.markdown LICENSE]
 
-  s.add_runtime_dependency('liquid',    '4.0.0.rc1')
+  s.add_runtime_dependency('liquid',    '4.0.0.rc2')
   s.add_runtime_dependency('kramdown',  '~> 1.3')
   s.add_runtime_dependency('mercenary', '~> 0.3.3')
   s.add_runtime_dependency('safe_yaml', '~> 1.0')
--- a/jekyll.gemspec
+++ b/jekyll.gemspec
@@ -29,7 +29,7 @@ Gem::Specification.new do |s|
   s.rdoc_options = ['--charset=UTF-8']
   s.extra_rdoc_files = %w[README.markdown LICENSE]
 
-  s.add_runtime_dependency('liquid',    '4.0.0.rc2')
+  s.add_runtime_dependency('liquid',    '4.0.0.rc3')
   s.add_runtime_dependency('kramdown',  '~> 1.3')
   s.add_runtime_dependency('mercenary', '~> 0.3.3')
   s.add_runtime_dependency('safe_yaml', '~> 1.0')
--- a/jekyll.gemspec
+++ b/jekyll.gemspec
@@ -29,7 +29,7 @@ Gem::Specification.new do |s|
   s.rdoc_options = ['--charset=UTF-8']
   s.extra_rdoc_files = %w[README.markdown LICENSE]
 
-  s.add_runtime_dependency('liquid',    '4.0.0.rc3')
+  s.add_runtime_dependency('liquid',    '~> 4.0')
   s.add_runtime_dependency('kramdown',  '~> 1.3')
   s.add_runtime_dependency('mercenary', '~> 0.3.3')
   s.add_runtime_dependency('safe_yaml', '~> 1.0')
--- a/lib/jekyll/drops/site_drop.rb
+++ b/lib/jekyll/drops/site_drop.rb
@@ -19,6 +19,10 @@ module Jekyll
         end
       end
 
+      def key?(key)
+        (@obj.collections.key?(key) && key != "posts") || super
+      end
+
       def posts
         @site_posts ||= @obj.posts.docs.sort { |a, b| b <=> a }
       end
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -402,9 +402,11 @@ module Jekyll
       operator = parser.consume?(:comparison)
       condition =
         if operator
-          Liquid::Condition.new(left_expr, operator, parser.expression)
+          Liquid::Condition.new(Liquid::Expression.parse(left_expr),
+                                operator,
+                                Liquid::Expression.parse(parser.expression))
         else
-          Liquid::Condition.new(left_expr)
+          Liquid::Condition.new(Liquid::Expression.parse(left_expr))
         end
       parser.consume(:end_of_string)
 
--- a/lib/jekyll/filters/grouping_filters.rb
+++ b/lib/jekyll/filters/grouping_filters.rb
@@ -40,7 +40,7 @@ module Jekyll
 
       private
       def parse_expression(str)
-        Liquid::Variable.new(str, {})
+        Liquid::Variable.new(str, Liquid::ParseContext.new)
       end
 
       private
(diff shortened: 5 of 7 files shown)


dylanahsmith:
https://github.com/jekyll/jekyll/blob/281f75182e63998d1fb823aee4f9562aa1f6fded/test/test_tags.rb#L46 passes in a list of tokens into a Tag, but we now use a tokenizer which also keeps track of the current line number during parsing.

This could be worked around by using `Liquid::Tokenizer.new(["test", "{% endhighlight %}", "\n"])`, but the tests would still be brittle to internal liquid changes, so it is recommended to parse a complete template using `Liquid::Template` rather than constructing individual tags with mocked out data.

envygeeks:
Would this affect normal everyday plugins @dylanahsmith ?

parkr:
> it is recommended to parse a complete template using Liquid::Template rather than constructing individual tags with mocked out data.

:+1: 

> Would this affect normal everyday plugins

This looks like it's only in our tests where this is going to be an issue. Most plugins don't use this `Tag.parse` method directly, I think only our tests do that.

parkr:
@dylanahsmith Our tests depend on having access to the instance created. I gave it a shot with `Liquid::Template.parse` but gave up after about 30 minutes of looking through the Liquid code.

It looks like v4 works (CI works fine for me locally). Would you consider releasing a new version soon and we can upgrade?

parkr:
@dylanahsmith @fw42 @pushrax Hi! Would it be possible to reconsider a 4.0 release? Especially keen to get the whitespace change out to our users. Thanks!

fw42:
I'm ok with releasing a new version of Liquid. As far as I know, there isn't any pending non-backwards compatible changes that we want to get in before a new release. @pushrax @dylanahsmith, thoughts?

dylanahsmith:
version ⬆️  :+1:

pushrax:
Also 🆗 from me

DirtyF:
🎉  https://github.com/Shopify/liquid/releases/tag/v4.0.0.rc3

fw42:
Yea I just released 4.0rc versions of both Liquid and Liquid-C. @parkr, can you see if everything works on your end? If so, I can release a real non-rc release.

parkr:
Interesting –– all of our `where_exp` tests fail.

parkr:
Looks like things are working here. I'd be interested in getting this into a v3.4, though I have concerns about what incompatibilities for plugin authors.

Crunch09:
@parkr I think the only public API thing we changed was to add `Jekyll::Drop::SiteDrop#key?`. Do you know of any liquid-related plugins so we could check them out before?

Also if you think we're ok, we could give @fw42 a go to release 4.0 non-rc 🎆

parkr:
@fw42 If this CI passes, then feel free to release v4.0 of Liquid. ❤️

fw42:
We have a few last minute breaking changes that we want to get in, but I will try to release v4 this week if possible.

parkr:
@fw42 More Breaking changes may mean we can't upgrade :frowning: Would it be possible to release a 5.0 with those changes? We may be stuck on 3.x if we can't guarantee compatibility at the high level for plugins. Or, perhaps we can run our test suite against the individual pull requests?

dylanahsmith:
@parkr https://github.com/Shopify/liquid/pull/835 is the change that I would like to get into liquid v4, since it provides a more secure default for rendering errors, so that sensitive information doesn't get leaked through the error message.  I tried running the tests on this jekyll branch against that liquid branch and everything passed.  It looks like jekyll uses Liquid::Template#render!, which isn't affected by that pull request, since it will continue to re-raise the exception as it did before.

parkr:
@dylanahsmith Perfect, thank you!

envygeeks:
Yay for Liquid 4.

ashmaroli:
Liquid 4 is now [officially](https://rubygems.org/gems/liquid/versions/4.0.0) here :tada: 
[`ad00998`](https://github.com/Shopify/liquid/commit/ad00998ef8a6722065e1e17476ac26aa715b84ed)

pathawks:
I don't think I'll have time to put Liquid 4 through it's paces for another week or so, but I am super excited for some of the new features it brings!

ashmaroli:
I wonder why tests suddenly fail when they had passed with `v4.0.0.rc3`

fw42:
Test failures look related to https://github.com/Shopify/liquid/pull/835 or https://github.com/Shopify/liquid/pull/837.

fw42:
Probably the latter since @dylanahsmith said that he tested the former with this branch.

parkr:
> Probably the latter since @dylanahsmith said that he tested the former with this branch.
> 
Can you confirm that? I think I have seen this error before - it's where a Liquid::Context instance is expected but instead we get a Hash.

ashmaroli:
@parkr and @fw42 I think the failures were caused by something else.. and that's not related to Liquid 4
Travis is [all green](https://travis-ci.org/ashmaroli/jekyll/builds/184469318) with the run on [my fork](https://github.com/ashmaroli/jekyll/tree/liquid-4)

dylanahsmith:
I'm not seeing those failure locally when testing locally.  It doesn't look related to to Shopify/liquid#835 or Shopify/liquid#837, but it does look like a hash is being passed to Liquid::Variable instead of a Liquid::ParseContext, which is something that is new in liquid v4.

dylanahsmith:
The backtrace for the test failure has lib/jekyll/filters/grouping_filters.rb but the commit doesn't even have that file in [lib/jekyll/filters](https://github.com/jekyll/jekyll/tree/3bf169d6cb0f92677c8ea39599dd49c2c47f4740/lib/jekyll/filters).  For some reason the travis build for 3bf169d6cb0f92677c8ea39599dd49c2c47f4740 was actually testing using commit d5d2ac3da99ad16ab9b7597c72b7e223278ac809

dylanahsmith:
That's really cool that travis is run against a merge of the branch into the latest master.  So it turns out the problem is with https://github.com/jekyll/jekyll/pull/5513/files#diff-4bd19cf62920ce863d0114cd23fc3cdcR43 creating a Liquid::Variable manually with `Liquid::Variable.new(str, {})` where liquid expects to be the one to create Liquid::Variable and now passes a Liquid::ParseContext.

The following patch fixes the tests with liquid v4

```diff
diff --git a/lib/jekyll/filters/grouping_filters.rb b/lib/jekyll/filters/grouping_filters.rb
index a16901d..f2c0346 100644
--- a/lib/jekyll/filters/grouping_filters.rb
+++ b/lib/jekyll/filters/grouping_filters.rb
@@ -40,7 +40,7 @@ module Jekyll

       private
       def parse_expression(str)
-        Liquid::Variable.new(str, {})
+        Liquid::Variable.new(str, Liquid::ParseContext.new)
       end

       private
```

although ideally jekyll wouldn't create a Liquid::Variable manually

ashmaroli:
:smiley:  Patch fix confirmed [on my fork](https://travis-ci.org/ashmaroli/jekyll/builds/184575280)

Crunch09:
@dylanahsmith So ideally we would just run the expression through `Liquid::Parser` then ?

dylanahsmith:
Normally we just expect a whole liquid template to be parsed rather than a single variable, so I'm not really sure what would be the ideal way of implementing that feature.

pushrax:
`Expression.parse` is the correct way at the moment.

dylanahsmith:
@pushrax the filter added by https://github.com/jekyll/jekyll/pull/5513 supports filtering nested inside of a string, e.g. `{% data | group_by_exp: "item", "item.version | split: '.' | first" %}` so `Expression.parse` wouldn't be enough to support parsing and evaluating that.

Here are my concerns with using `Liquid::Variable.new(str, ParseContext.new).render(@context)`:
* It means the parsing is being done at render time with different parsing options
* Liquid::Variable#render sounds like it should return a string for the rendered output, but actually returns an object.  Liquid::Variable#evaluate would be a more appropriate name for its behaviour
* Having a Liquid::Variable.parse class method would make it more consistent with Liquid::Template and tags classes

pushrax:
Oh god, working around Liquid's lack of expressions with strings 😭 

We (@gmalette, @sgnr) have been talking a bit about how to add proper expression support to Liquid. I think we can do it in 2017.

parkr:
> Having a Liquid::Variable.parse class method would make it more consistent with Liquid::Template and tags classes

@dylanahsmith I like that idea. I think we're fairly comfortable breaking the rules a bit for now since there aren't tons of releases of Liquid and y'all are very kind about helping us test our code when you do want to make updates. Having `Liquid::Variable.parse` that offers a stable external API would be preferable, of course, but we're aware that our use of Liquid is not really your primary development concern.

Is there a way for us to make our current usage line up better with how it *should* work, e.g. by passing the proper `Liquid::ParseContext` from the parent template?

parkr:
Well, I have certainly sullied up this branch. Let's see if I can fix that git history...

JamesDrummond:
When will this PR be merged? Patiently waiting for liquid 4.0 features ;) . Also I was trying to find where you list when the next release and ship date is planned for.

parkr:
I think we'll try to ship a 3.4 this week, with a 3.5 a bit after that.

parkr:
Fixes #5772.

keslert:
Any update on this? Waiting not quite as patiently as @JamesDrummond ;).

Crunch09:
@keslert It is planned to be released in jekyll 3.5. Jekyll 3.4 should be out [this week](https://github.com/jekyll/jekyll/pull/5814), so it probably will be merged in afterwards.

parkr:
@jekyllbot: merge +minor

PR #4368, opened 2016-01-18
WIP: Move to Bundler + a Gemfile

chrisfinazzo:


site/_docs/drafts.md
```diff
@@ -14,7 +14,7 @@ first draft:
 |   |-- a-draft-post.md
 {% endhighlight %}
 
-To preview your site with drafts, simply run `jekyll serve` or `jekyll build`
-with the `--drafts` switch. Each will be assigned the value modification time
-of the draft file for its date, and thus you will see currently edited drafts
-as the latest posts.
+To preview your site with drafts, simply run `bundle exec jekyll serve` or
+`bundle exec jekyll build` with the `--drafts` switch. Each will be assigned
+the value modification timeof the draft file for its date, and thus you will
```
Strangehill: typo: timeof -> time of

chrisfinazzo:
There is more to do on this, but as I was playing around with it yesterday, a thought occurred to me...

`jekyll new` will fail if the directory in which it's being installed in either a) Already exists or b) Is not empty. As the version of Jekyll is specified in the `Gemfile` and the purpose of this change is so that all that needs to be done is `bundle install`, what should the workaround be?

I was able to get the install to succeed by moving `Gemfile` and `Gemfile.lock` out of my test directory temporarily, but that's a kludge. I seem to remember a `--force` flag somewhere, but I would like to avoid that if at all possible. An example of this is show below - not correct for the reason I mentioned, but in spirit I think this is where things will eventually end up.

/cc @jekyll/core

screen shot 2016-01-21 at 12 19 45 pm

envygeeks:
@chrisfinazzo we can probably address that easily, mind filing a separate ticket suggesting that?

chrisfinazzo:
Huh. This error looks familiar (See `test_static_file.rb:100`). As this failure is on 2.0 - which may be dropped in #4381, can it be removed from Travis?

chrisfinazzo:
Updated w/ change from #4542.

parkr:
This requires further consideration. I don't think we'll go this way soon. If anything, we will likely choose `jekyll _3.0.3_ build` or something similar, where you specify the version during execution. It's not fully ideal but should get the job done without as many steps. Thanks!

chrisfinazzo:
Makes sense. In  that case, we can probably repurpose the [instructions][] and procedures that apply to prereleases with a few small changes. The other use cases that stand out to me are:

[instructions]: https://jekyllrb.com/docs/installation/#pre-releases

- GitHub Pages (Older versions of `gh-pages`)
- Windows (Not sure about the current status of this as it relates to v3)
- Docker (This is really @envygeeks domain)

envygeeks:
We can probably make the Docker image a bit more intelligent.

PR #4374, opened 2016-01-19
Update & prune the short list of example sites

fusion809:
I am making this pull request as I think that this page could do with an indicator of when each of these sites were last updated as this gives one an idea as to whether or not these sites are still valid examples to follow (as it is possible that their code has become outdated and so has their layout, due to new releases of Jekyll).

### DIFF
--- a/site/_docs/sites.md
+++ b/site/_docs/sites.md
@@ -9,17 +9,13 @@ with. Below are some Jekyll-powered blogs which were hand-picked for
 learning purposes.
 
 - [Tom Preston-Werner](http://tom.preston-werner.com/)
-    ([source](https://github.com/mojombo/mojombo.github.io)) (last commit [87f9f24](https://github.com/mojombo/mojombo.github.io/commit/87f9f24) at 20 June 2015)
-- [Nick Quaranto](http://quaran.to/)
-    ([source](https://github.com/qrush/qrush.github.com)) (last commit [c569be1](https://github.com/qrush/qrush.github.com/commit/c569be159c4c1075e8bf65b1995e43b262c86899) at 2 December 2014)
+    ([source](https://github.com/mojombo/mojombo.github.io))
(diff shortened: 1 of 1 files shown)


envygeeks:
This is a hostile commit IMO.  It's a burden to maintain such a thing, would rarely be updated and if the user suddenly updates their source, misleads anybody browsing our site.  It's really not that hard to click the link and see on the Source page exactly when it was updated last.

fusion809:
True, but I was hoping this might help one find up-to-date sites faster. I will accept this commit being trumped but two of those sites listed haven't been updated in over two years, do yas (the maintainers of this repo) still think they're valid examples for the currently-supported releases of Jekyll? Oh and I just realized that GitHub training source link should be to https://github.com/github/training-kit, as the repo linked has moved.

fusion809:
Look at my [second commit](https://github.com/jekyll/jekyll/commit/8507cda6ddfaf59fc25837df111b77a6651258fe), is this more amenable?

parkr:
I'll take a look at this in a bit. This list is very old and there might be some political elements to adding or removing. Thanks!

fusion809:
@pakr If you want to add another site to this list feel free to add mine "The Hornery" at https://fusion809.github.io, source: https://github.com/fusion809/fusion809.github.io. I think it would be helpful for instructional purposes as look at my README: https://github.com/fusion809/fusion809.github.io/blob/master/README.md. It gives Jekyll-related instructions for Linux users in particular.

envygeeks:
Jekyll doesn't run on Debian Ruby it only runs on Ubuntu Ruby if you are above 15.04 (which is EOL.)

envygeeks:
I take that back, they finally updated it, so nevermind, it runs on Debian Ruby, carry on.

parkr:
@jekyllbot: merge +site

PR #4389, opened 2016-01-22
Refactor Page#permalink method

zsyed91:
- Remove extra OR condition since a missing hash key will return a nil anyway and made it a 1 liner
- Added a test to catch this nil condition since it was missing to begin with
- Reduced line length in test_page.rb

Hope the change makes sense. I noticed there wasn't a test case to check for the sad path for `.permalink`. Does where I added this test make sense?

### DIFF
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -63,8 +63,7 @@ module Jekyll
     #
     # Returns the String permalink or nil if none has been set.
     def permalink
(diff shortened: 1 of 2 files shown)


parkr:
This LGTM! Thanks for being so thorough, @zsyed91. :smile:

parkr:
@jekyllbot: merge +minor

Issue #4372, opened 2016-01-18
Support Liquid `strict` parsing mode


parkr:
Some users are seeing invalid queries/commands being swallowed by Liquid. We should test strict mode, which is available in Liquid v3 to more strictly parse Liquid templates.

Enabling it is as simple as: `Liquid::Template.error_mode = :strict`. Ideally, this would be turned on via:

```yaml
liquid:
  mode: strict
```

or similar, like `strict: true`.

Ref: https://github.com/jekyll/jekyll/issues/4371

pathawks:
Is there some way that `jekyll doctor` could try to generate the site with Liquid strict mode?

This way, if a user just encounters a one time problem, they could run one command to diagnose, rather than modifying the site config.

parkr:
@pathawks That's a good idea! Using `jekyll doctor` is excellent. Want to add that line to the Doctor command?

PR #4393, opened 2016-01-23
Doctor should check for Liquid errors

pathawks:
`jekyll doctor` Should try to generate the site using Liquid `:strict` mode and report any errors.

#4372

### DIFF
--- a/test/test_doctor_command.rb
+++ b/test/test_doctor_command.rb
@@ -33,4 +33,22 @@ class TestDoctorCommand < JekyllUnitTest
       assert_includes output, "Warning: The following URLs only differ by case. On a case-insensitive file system one of the URLs will be overwritten by the other: #{dest_dir}/about/index.html, #{dest_dir}/About/index.html"
     end
   end
+
+  context 'Liquid errors' do
+    setup do
+      clear_dest
+    end
+
+    should 'return error for invalid Liquid' do
+      @site = Site.new(Jekyll.configuration({
+        "source" => File.join(source_dir, '/_liquid_errors'),
+        "destination" => dest_dir
+      }))
+      @site.process
+      output = capture_stderr do
+         ret = Jekyll::Commands::Doctor.liquid_errors(@site)
+         assert_equal true, ret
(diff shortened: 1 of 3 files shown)


pathawks:
Unfortunately, it's not working. This isn't going to be as simple as I had expected. :rage1:

test/test_doctor_command.rb
```diff
@@ -33,4 +33,22 @@ class TestDoctorCommand < JekyllUnitTest
...
+    should 'return error for invalid Liquid' do
+      @site = Site.new(Jekyll.configuration({
+        "source" => File.join(source_dir, '/_liquid_errors'),
+        "destination" => dest_dir
+      }))
+      @site.process
+      output = capture_stderr do
+         ret = Jekyll::Commands::Doctor.liquid_errors(@site)
+         assert_equal true, ret
+      end
```
parkr: Could you assert a particular output? We want to make sure these errors are surfaced.

parkr:
:tada:

test/test_doctor_command.rb
```diff
@@ -33,4 +33,22 @@ class TestDoctorCommand < JekyllUnitTest
...
   end
+
+  context 'Liquid errors' do
+    setup do
+      clear_dest
+    end
+
+    should 'return error for invalid Liquid' do
+      @site = Site.new(Jekyll.configuration({
+        "source" => File.join(source_dir, '/_liquid_errors'),
```
parkr: Does `test/source/_liquid_errors` exist?

pathawks: Added in this PR, yes.

test/test_doctor_command.rb
pathawks: I *could*, but right now there is no output. It is not throwing any errors or anything.

parkr: That's not good! If there's no output, then the user has no idea why the build failed.

pathawks:
I'm in too deep here. :sweat:

PR #4401, opened 2016-01-25
Renderer#output_ext: honor folders when looking for ext

### LINKED ISSUE #4397: Jekyll 3.1.0 Issue: Permalinks with dots are broken (e.g., version numbers)
Since Jekyll 3.1.0, our permalinks for news about new releases are broken. The `index.html` file that is produced does not have the right name; it's called `index.html.1` (or another digit). Links the to the page are "correct" (end with `/`) but then their destination is a 404.

Example:
permalink: `/news/2016/01/25/announcing-scalajs-0.6.6/`
file produced: `/news/2016/01/25/announcing-scalajs-0.6.6/index.html.6`
expected: `/news/2016/01/25/announcing-scalajs-0.6.6/index.html`

The actual file (and the rest of the website's repo) can be found at https://github.com/scala-js/scala-js-website/blob/master/_posts/news/2016-01-25-announcing-scalajs-0.6.6.md
Currently I downgraded to Jekyll 3.0.1.

I suspect this comes from #4373.

parkr:
Previously, even if the document permalink was a folder, it would look for
an extension on that.  For example, if I have:

```text
permalink: "/new-version-jekyll-v3.0.0/"
```

the output_ext would be `".0"`.  Now, the output_ext honors the trailing
slash and will report based on the converters instead.

Fixes #4397.

### DIFF
--- a/lib/jekyll/renderer.rb
+++ b/lib/jekyll/renderer.rb
@@ -168,7 +168,7 @@ module Jekyll
     private
 
     def permalink_ext
-      if document.permalink
+      if document.permalink && !document.permalink.end_with?("/")
         permalink_ext = File.extname(document.permalink)
         permalink_ext unless permalink_ext.empty?
       end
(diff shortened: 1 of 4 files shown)


parkr:
@jekyllbot: merge +dev

PR #4408, opened 2016-01-27
Minor refactor to convertible.published?

zsyed91:
Flipped the logic from the `!` case to the `true` scenario. This makes it slightly easier to read and maintain. Added test cases for this method as well.

### DIFF
--- a/lib/jekyll/convertible.rb
+++ b/lib/jekyll/convertible.rb
@@ -25,7 +25,7 @@ module Jekyll
 
     # Whether the file is published or not, as indicated in YAML front-matter
     def published?
-      !(data.key?('published') && data['published'] == false)
+      data.key?('published') && data['published']
     end
 
     # Returns merged option hash for File.read of self.site (if exists)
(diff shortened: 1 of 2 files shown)


zsyed91:
Yeah that is true. I was going back and forth on that. As in what happens if its a value that is not true (but not false). But since we were explicitly comparing false, I will remove that `=== true`.

zsyed91:
@envygeeks Alright removed the `== true`. Ready for review.

parkr:
Hey, @zsyed91! Thanks for the PR. While this is easier to maintain and read, it flips the logic. 

The intended logic is: "an item is published if (1) there is no `published` frontmatter item, or (2) the value of that item is not `false`". It's a blacklist model – only if you say it's _not_ published do we not publish it. 

Your change introduces a whitelist model – only if you say it's published do we publish it.

Thanks for your PR and feel free to suggest any other changes, even if it's just better comments for this method :smiley:

zsyed91:
@parkr Sure, that makes sense. I will create another branch with some comment changes and those tests from this one :smiley:

PR #4409, opened 2016-01-27
Suppress stdout in liquid profiling test

alfredxing:
The test was spewing out some whitespace ¯\_(ツ)_/¯
This just suppresses output to STDOUT for this test only.

Previously:

screen shot 2016-01-27 at 1 44 09 pm

### DIFF
--- a/test/test_site.rb
+++ b/test/test_site.rb
@@ -465,6 +465,14 @@ class TestSite < JekyllUnitTest
         @site = Site.new(site_configuration('profile' => true))
       end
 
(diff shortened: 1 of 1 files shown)


alfredxing:
Fixed, thanks for the suggestion :)

parkr:
@jekyllbot: merge +dev

parkr:
Thanks so much, @alfredxing! :sparkles:  I was wondering where that newline came from but didn't have time to dig into it. :smile:

PR #4412, opened 2016-01-28
Pages as a collection

alfredxing:
***Still a work in progress!***

This PR moves pages to be a collection (like the recent move of posts). When looking through the code it'll probably make more sense to look at the diffs commit-by-commit, but the overall diff isn't too large either.

Here is a summary of the changes:
- Removed the old Page class, replacing it with a subclass of Document
- Updated the read and render flows to reflect this change
- Updated Page class, Drop classes, and unit tests to try and make everything backward compatible with what we had before (i.e. all of the accessors, methods)

The only issues I'm running into right now in the unit tests are:
- The documents are not sorted (should be an easy fix)
- For some reason the CoffeeScript test is failing; Liquid isn't being rendered in.


lib/jekyll/page.rb
```diff
@@ -18,37 +9,36 @@ class Page
       name
       path
       url
-    )
+    ).freeze
```
parkr: These attributes should be specified in the `Drop`. Let's use `DocumentDrop` if possible.

lib/jekyll/site.rb
```diff
@@ -208,6 +200,10 @@ def posts
       collections['posts'] ||= Collection.new(self, 'posts')
     end
 
+    def pages
+      collections['pages'] || Collection.new(self, 'pages')
```
parkr: You might have missed an `=` here: `||= Collection...`

parkr:
Looking great so far, @alfredxing! Thanks for taking this on. Ping me when you're ready for a full review.

If there's any way to make pages work without needing the `Page` subclass, that'd be ideal. Things like excerpts and the like should all fall into the scope of `Document` configuration, I think. Some people want Page excerpts. :smile:

alfredxing:
@parkr Thanks for taking a look! My main reason for using the `Page` class was because I wanted to make sure everything was backwards-compatible and that all of the existing tests pass. That means I needed to override some of the default Document behaviour, even if it wasn't too sane (for example, one thing that really annoys me is the discrepancy in `path`, `name`, etc. between Pages and Documents).

alfredxing:
@parkr Finally had a bit of time to work on this! Basically I fixed the last few tests that I could. There are still a few failing:

- Pagination (not sure what the state of `jekyll-paginate` is?)
- Jekyll test plugin gem
- One inconsistency between Document and Page: in Document, `{{ page.path }}` is not overwritable, while in Page, it is? Not sure which way to go.

parkr:
@alfredxing Excellent! Thanks for all your work on this.

> One inconsistency between Document and Page: in Document, {{ page.path }} is not overwritable, while in Page, it is? Not sure which way we want to go, but I'm leaning toward immutable.

I think we have to leave it as overridable for the sake of not breaking things.

> - Pagination (not sure what the state of jekyll-paginate is?)
> - Jekyll test plugin gem

They should both work, again because we can't break things.

/cc @jekyll/build

alfredxing:
Got it. I'll look into those some more. I think most of these stem from the change in the constructor.
Another issue was the output extension for `.xhtml` and `.htm` files, I'll fix this as well.

parkr:
@alfredxing Pagination is working like normal for me. Note that the pagination generator creates an instance of `Jekyll::Page`, and adds it to `site.pages` for rendering with the `pager` instance var set. As long as Render can see `document.pager`, it should work.

alfredxing:
Closing in favour of #4822

PR #4415, opened 2016-01-28
Correct ruby2.0-dev package name

hultberg:
On ubuntu trusty, the package is named `ruby2.0-dev`.

Searching on packages.debian.org I found that there is no ruby2.0-dev package as jessie provides ruby version 2.1 (ref https://packages.debian.org/jessie/ruby).


hultberg:
I close this as after installing `ruby2.0-dev` on my trusty server I still got errors while installing jekyll. Opening an issue on this.

PR #4417, opened 2016-01-28
Add space and Rus-lang chars support in filenames

figliohulio:
I needed this feature in my project, as forced to use Russian-language filenames in some cases. I think it's useful not only for me and I think that in the future should enter at least limited support for UTF filenames for other languages. If the project team will agree with this, I can even prepare the necessary regular expression for a part of national alphabets.


parkr:
Should we just be using `\w`?

This will need tests.

parkr:
Closing as abandoned.

PR #4410, opened 2016-01-28
Docs: Quickstart - added documentation about the `--force` option

michaellee:
Thought this would be helpful in scenarios where folks want to create a new Jekyll project in an existing folder that is non-empty.

Issue #4391, opened 2016-01-22
`jekyll new .` fails under Bundler


chrisfinazzo:
It is currently not possible to run `jekyll new` in an existing directory if Bundler is being used for gem management. The `jekyll new` command obviously requires that the gem be installed, but Bundler is used to carry out this step of the process. As a result, when you run `jekyll new .`, the script will complain that the directory you are using already exists or is not empty.

Related discussion can be found in PR #4368

envygeeks:
If @parkr is intent with us moving to suggesting `bundle` we can filter out key files on `new` like Gemfile, `bundle` `vendor/` (or otherwise by parsing the `.bundle/config`) and then if nothing is left after that, go ahead and generate the site, otherwise complain about the first file we can pull.

knightotoldcode:
This would be a nice addition. I've been using ```jekyll new . --force``` in the past as a way to get around this when I want to pre-initialize a new directory with a Gemfile (nice for syncing up an initial build with the "github-pages" gem, for example).

chrisfinazzo:
@knightotoldcode My thinking was more along the lines of a refactoring to get this done. As you said, `--force` will do the job, but (IMO) as part of the first-run experience, it's pretty crappy.

envygeeks:
`--force` should be reserved for crazier actions.

knightotoldcode:
I wasn't advocating for ```--force``` to be the blessed mechanism. It feels like an unfortunate hack. And fully supporting bundler is a *much* more pleasant experience.

envygeeks:
I was agreeing with you two :wink:

parkr:
If we go file-by-file and fail if any of them exist, then it'd be easier for this to be run in an existing directory without `--force`.

PR #4420, opened 2016-01-28
Add a warning about using force

chrisfinazzo:
Related: #4391 and #4410.

/cc @michaellee @envygeeks


parkr:
I think we should just fix the behaviour of this for 3.2.

PR #4428, opened 2016-01-30
Fix #4427: Make our @config hash symbol accessible.

### LINKED ISSUE #4427: GFM fenced code blocks regression in Jekyll 3
I could be missing something obvious, but I don't seem to be able to get fenced code blocks to work with Jekyll 3, using a config which worked fine in 2.5.3.

Tested with Jekyll 3.1.0.

To reproduce:

```
~ $ mkdir jekyll-rouge && cd jekyll-rouge
~/jekyll-rouge $ jekyll new
```

Add the following snippets in _posts/2016-XX-XX-welcome-to-jekyll.markdown:

    ```ruby
    def print_hi(name)
      puts "Hi, #{name}"
    end
    print_hi('Tom')
    ```
    
    {% highlight ruby %}
    def print_hi(name)
      puts "Hi, #{name}"
    end
    print_hi('Tom')
    {% endhighlight %}

And to the config (for compatibility with 2.5.3) add:

```
markdown: kramdown

kramdown:
  input: GFM
  hard_wrap: false
```

Then:

```bash
# For 2.5.3
~/jekyll-rouge $ jekyll _2.5.3_ s

# For 3.1.0
~/jekyll-rouge $ jekyll _3.1.0_ s
```

These are the same steps as https://github.com/jekyll/jekyll/pull/4053#issuecomment-151048257. Jekyll 2.5.3 generates a syntax-highlighted code block as expected, however the same code run with 3.1.0 generates:

welcome_to_jekyll_

I believe from https://github.com/jekyll/jekyll/issues/4202 that the additional GFM config I was using for 2.5.3 is the default in 3.1.0. Removing those lines and using the stock config doesn't affect the result.

I've just rolled back to 2.5.3, which is doing just fine, so this isn't urgent for me.

envygeeks:


lib/jekyll/converters/markdown/kramdown_parser.rb
```diff
@@ -32,22 +32,44 @@ def setup
...
+          make_accessible
         end
 
         def convert(content)
           Kramdown::Document.new(content, @config).to_html
         end
 
+        private
+        def make_accessible(hash = @config)
+          proc_ = proc { |hash_, key| hash_[key.to_s] if key.is_a?(Symbol) }
```
parkr: When do we access with symbols?

envygeeks: We don't, it's because Kramdown wants symbols and we want strings.  So we give the best of both worlds so we don't disrupt our behavior and they can access it how they wish to without affecting us and us blindly accepting symbols.

parkr:
@jekyllbot: merge +bug

PR #4424, opened 2016-01-30
Rename CONDUCT.md to CONDUCT.markdown

SuriyaaKudoIsc:
@mojombo @parkr @mattr- 

----

Hi guys, :smile: 

I renamed the **`CONDUCT.md`** file to **`CONDUCT.markdown`** because all other Markdown files like *README*, *CONTRIBUTING* and the *HISTORY* files got a *`.markdown`* file ending instead of *`.md`*!

*Yours,*
[**Suriyaa Kudo**](https://github.com/SuriyaaKudoIsc) :octocat:

PR #4434, opened 2016-02-01
fix broken link to CONDUCT.markdown in README.markdown

chris72205:


parkr:
Changed per #4424. Great catch, thanks for the fix.

PR #4445, opened 2016-02-02
rogue ~> rouge

parkr:
Thanks to @tombell [who reported this](https://twitter.com/thetombell/status/694410112576913409) to the Twitter ether and @daviddarnes for alerting me to it.

/cc @envygeeks

### DIFF
--- a/lib/jekyll/converters/markdown/kramdown_parser.rb
+++ b/lib/jekyll/converters/markdown/kramdown_parser.rb
@@ -16,7 +16,7 @@ module Jekyll
 
         def initialize(config)
           Jekyll::External.require_with_graceful_fail "kramdown"
-          @main_fallback_highlighter = config["highlighter"] || "rogue"
+          @main_fallback_highlighter = config["highlighter"] || "rouge"
           @config = config["kramdown"] || {}
           setup
         end


envygeeks:
That fix is already merged into the other fixes for that class so I don't see the point of this?

parkr:
> That fix is already merged into the other fixes for that class so I don't see the point of this?

Where?

daviddarnes:
:+1: thanks y'all.

tombell:
:+1: I was actually trying to work out if it was genuinely misspelt or not before saying something. :smile: :

envygeeks:
@tombell it wasn't intentional, when I see rouge I think rogue instead of "this is a valid word for red." so subconsciously corrected it without even realizing since as a gamer rogue is muscled in.

daviddarnes:
@tombell always good to do a sanity check. I saw the tweet after I was having issues with switching a site from pygments to rouge.

tombell:
@envygeeks half of me switching over to rouge was correcting rogue typos (pun intended).

parkr:
Fixed by 65a1fc4120bf5b39e0474ac6ac5c3726bd821b60 in #4428. Thanks, @envygeeks!

PR #4449, opened 2016-02-02
Backport Github Flavored Markdown.

envygeeks:
Github Flavored Markdown was etched for 3.1 with the Kramdown converter rewrite but people are assuming it's enabled in 3.0, this backports that option into 3.0 so it can be hot-released onto Github pages.

### DIFF
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -64,6 +64,7 @@ module Jekyll
       },
 
       'kramdown' => {
+        'input'          => 'GFM',
         'auto_ids'       => true,
         'footnote_nr'    => 1,
         'entity_output'  => 'as_char',


parkr:
@benbalter This is set on the GHP side, right?

parkr:
But yeah we can backport it anyway for good measure to ensure things are mirrored.

benbalter:
> This is set on the GHP side, right?

Yes, and hoping to move this to the GitHub Pages gem for transparency/consistency :soon:.

envygeeks:
@benbalter should it be on the github-pages gem or on Jekyll itself? I know more people who build without `github-pages` and just use Jekyll assuming it's the same than I know who don't, and I did for the first year or so too, actually even our Docker image used vanilla Jekyll and just grepped the gems you wanted before I decided to just use Github-Pages with Docker-Template.

benbalter:
> should it be on the github-pages gem or on Jekyll itself?

Didn't mean to confuse the issue. Yes, Jekyll itself. We'd pull the config out then, if it were. There are a few more defaults / overrides we set, that we'd like to expose, kramdown config being one of them.

parkr:
I would like to enforce this in the `github-pages` gem. We'll tackle that here: https://github.com/github/pages-gem/pull/209 :smile: Thanks!

envygeeks:
@parkr that's not necessary.  We already enforce GFM in Jekyll by default, it's been in there for months.

PR #4464, opened 2016-02-03
Remove old flag which breaks htmlproof

chrisfinazzo:
The `script/proof` command fails under 3.1.1.

screen shot 2016-02-03 at 11 53 45 am

Related discussion: #4059.

Can the new flag just get swapped in? 

/cc @gjtorikian

### DIFF
--- a/script/proof
+++ b/script/proof
@@ -27,7 +27,7 @@ bundle install -j8 > /dev/null || bundle install > /dev/null
 
 # 2.
 msg "Building..."
-bundle exec jekyll build -s $SOURCE -d $DESTINATION --full-rebuild --trace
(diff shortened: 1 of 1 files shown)


gjtorikian:
I am not sure what this has to do with `htmlproof`; can you explain?

parkr:
@gjtorikian It's only that the purpose of this script is to run html-proofer against our built site. The problem is an invalid flag in the build portion of the script; html-proofer is working as expected. :smile_cat:

parkr:
Thanks, @chrisfinazzo!

chrisfinazzo:
@gjtorikian My concern was about passing arbitrary options around (e.g, adding `--incremental` as an option for the script) and this would break things - again.

Carry on...

PR #4468, opened 2016-02-03
Jekyll.sanitized_path: escape tildes before sanitizing a questionable path

### LINKED ISSUE #4457: Files that start with a tilde (~) break builds
The way [we use `File.expand_path`](https://github.com/jekyll/jekyll/blob/7df8f502487f3b6476b5ce0f296be28f70df59cb/lib/jekyll.rb#L156), we expand paths of every file. We should escape a number of different values in this call. [StackOverflow has a great idea of what to escape](http://stackoverflow.com/questions/6159355/how-do-i-escape-parentheses-and-brackets-in-ruby).

```irb
>> File.expand_path("~hi.txt")
ArgumentError: user hi.txt doesn't exist
	from (irb):3:in `expand_path'
	from (irb):3
	from /Users/parkr/.rbenv/versions/2.3.0/bin/irb:11:in `<main>'
>> File.expand_path('\~hi.txt')
=> "/Users/parkr/jekyll/jekyll/\\~hi.txt"
```

Should be patched on 3.0.2 and 3.1.2.

/cc @benbalter

parkr:
Fixes #4457.

/cc @benbalter

### DIFF
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -153,7 +153,7 @@ module Jekyll
     def sanitized_path(base_directory, questionable_path)
       return base_directory if base_directory.eql?(questionable_path)
 
-      questionable_path.insert(0, '/') unless questionable_path.start_with?('/')
+      questionable_path.insert(0, '/') if questionable_path.start_with?('~')
       clean_path = File.expand_path(questionable_path, "/")
       clean_path.sub!(/\A\w\:\//, '/')
 
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -153,8 +153,9 @@ module Jekyll
     def sanitized_path(base_directory, questionable_path)
       return base_directory if base_directory.eql?(questionable_path)
 
+      questionable_path.insert(0, '/') unless questionable_path.start_with?('/')
       clean_path = File.expand_path(questionable_path, "/")
-      clean_path = clean_path.sub(/\A\w\:\//, '/')
+      clean_path.sub!(/\A\w\:\//, '/')
 
       unless clean_path.start_with?(base_directory.sub(/\A\w\:\//, '/'))
         File.join(base_directory, clean_path)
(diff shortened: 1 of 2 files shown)


lib/jekyll.rb
```diff
@@ -153,8 +153,9 @@ def sites
     def sanitized_path(base_directory, questionable_path)
       return base_directory if base_directory.eql?(questionable_path)
 
+      questionable_path.insert(0, '/') if questionable_path.start_with?('~')
```
parkr: @mastahyeti Hello! Could I borrow your :eyes: for a quick moment? We're seeing issues with questionable paths which start with `~`, such as `~$microsoftSwp.docx`. This is an attempt to fix the `ArgumentError`. Is it safe? :lock: 

```irb
>> File.expand_path('~$microsoftSwp.docx', '/')
ArgumentError: user $microsoftSwp.docx doesn't exist
	from (irb):1:in `expand_path'
	from (irb):1
	from /Users/parkr/.rbenv/versions/2.3.0/bin/irb:11:in `<main>'
>> File.expand_path('/~$microsoftSwp.docx', '/')
=> "/~$microsoftSwp.docx"
```

Figured I'd ask you before merging in Jekyll so we can get some extra reassurances. :heart:

parkr:
@benbalter We can't in a PATCH release, but maybe for Jekyll v3.2, [we ignore files that start with `~`](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/entry_filter.rb#L3)?

benbalter:
> @benbalter We can't in a PATCH release, but maybe for Jekyll v3.2, we ignore files that start with ~?

Is there a legitimate use case for publishing files starting for a `~`?

parkr:
> Is there a legitimate use case for publishing files starting for a `~`?

Considering I didn't catch this bug beforehand, I would say "no". If you have `https://example.com/~benbalter/` that usually maps specially for your server to some `/home/benbalter/public_html` folder or something. That's what we did [at McGill CS](http://cs.mcgill.ca/~pmoore6/) way back in the day. So no, generally `~` is considered "special."

lib/jekyll.rb
mastahyeti: Seems fine. Thanks for asking.

benbalter:
:+1:

Issue #3926, opened 2015-08-21
{% post_url %} not working with subdirectory


katelovescode:
I'm getting this error when presumably using the correct format for the tag.

```
Liquid Exception: Could not find post "/recipes/2015-07-08-overnight-oats-recipe" in tag 'post_url'. Make sure the post exists and the name is correct. in blogs/_posts/2015-07-01-coconut-milk.md
jekyll 2.5.3 | Error:  Could not find post "/recipes/2015-07-08-overnight-oats-recipe" in tag 'post_url'.

Make sure the post exists and the name is correct.
```

Here's my code:
```
I put it in smoothies, [overnight oats]({% post_url /recipes/2015-07-08-overnight-oats-recipe %})
```

Here's my directory structure:
```
- blogs/
-- _posts/
--- 2015-07-01-coconut-milk.md
- recipes/
-- _posts/
--- 2015-07-08-overnight-oats-recipe.md
```

Is it possible that because I'm referencing from within one subdirectory, it won't find the other?  It works if I don't include the subdirectory, but that precludes me having a blog post and recipe on the same day with the same name (which is how this issue originally arose - wanting to post a blog & recipe on the same day with the same topic, and linking using the subdirectory).

parkr:
Hi Kate! This looks like an issue to me with the [PostComparer](https://github.com/jekyll/jekyll/blob/cd4829d28d3323bdf74dbb6808be99be8f1c1d41/lib/jekyll/tags/post_url.rb#L8-L18). Are you willing to take a stab at it?

yihangho:
I am interested to give this a try. However, I am unable to reproduce the exact error, got this instead:

```
Deprecation: A call to '{{ post_url jekyll::tags::posturl }}' did not match a post using the new matching method of checking name (path-date-slug) equality. Please make sure that you change this tag to match the post's name exactly.
```

parkr:
@yihangho That means the arguments to the `post_url` is different than the slug on the filesystem. That error message could be improved...

yihangho:
I see. In that case, how should I proceed from here? Should I modify PostComparer to handle Kate's directory structure or simply improve the error message?

parkr:
I believe I fixed the error in Jekyll 3 stable. PostComparer should be improved, however, to allow this. Mind taking a look?

yihangho:
Not at all. I think we need to figure out how to deal with ambiguity though. Consider this directory structure:

```
Project Root
+-- _posts
[...]

PR #4470, opened 2016-02-04
typo in post_url.rb

yegor256:
This seems to be a typo.

### DIFF
--- a/lib/jekyll/tags/post_url.rb
+++ b/lib/jekyll/tags/post_url.rb
@@ -18,7 +18,7 @@ module Jekyll
       end
 
       def deprecated_equality(other)
-        date = Utils.parse_date(name, "'#{name}' does not contain valid date and/or title.")
+        date = Utils.parse_date(@date, "'#{@date}' does not contain valid date.")
         slug == post_slug(other) &&
           date.year  == other.date.year &&
           date.month == other.date.month &&


parkr:
@yegor256 Are you getting an error message from the previous?

yegor256:
@parkr how can it work, if `name` is not defined in the method? :) it's definitely a typo

yegor256:
@parkr and yes, I'm getting an error

envygeeks:
@yegor256 because uhm: https://github.com/yegor256/jekyll/blob/e428dad5bb00edd5d49764b4a3c7576b886c2ce3/lib/jekyll/tags/post_url.rb#L6

yegor256:
@envygeeks make sense, but this still have to be `date`, not `name`. `name` contains full URL, while `date` is the exact date.

envygeeks:
@yegor256 what is the error you are getting? Can you provide the trace?

yegor256:
@envygeeks I already fixed it by implementing this post_url myself. yours doesn't work anyway, because of this: https://github.com/jekyll/jekyll/issues/3926

envygeeks:
I'm against this unless you can post the error you are getting to confirm what you want to change makes sense.  Blindly accepting a pull request without an error or even a better explanation of how and what exactly it is fixing (that aligns properly) is bad for business.

applqpak:
I agree.

lib/jekyll/tags/post_url.rb
```diff
@@ -18,7 +18,7 @@ def ==(other)
       end
 
       def deprecated_equality(other)
-        date = Utils.parse_date(name, "'#{name}' does not contain valid date and/or title.")
+        date = Utils.parse_date(@date, "'#{@date}' does not contain valid date.")
```
parkr: I don't even know why we're doing this because `date` is an `attr_reader` for `@date` anyway so why we're shadowing it instead of just parsing it in the initializer?

parkr:
I have implemented a fix in #4670 as this has fallen stale. I added a test case in my example and changed it around to be satisfactory to all parties, I hope. Thanks for the contribution, and sorry it wasn't resolved more quickly.

PR #4472, opened 2016-02-05
3.0 stable

pandasec:


parkr:
3.0-stable is an old branch we are not bringing into `master`. Thanks, though!

Issue #4385, opened 2016-01-21
Feature request: add where_contains filter


drumanagh:
This would be a counterpart to the `where` filter, allowing selection by one item in an array in a variable rather than by the full contents of a variable.

For an example:

In `file-1.md`, I have
```
---
title: File 1
permalink: /file-1/
array_a:
  - dogs
  - fish
  - cats
---
```

In `file-2.md`, I have
```
---
title: File 2
permalink: /file-2/
array_a:
  - giraffes
  - lions
  - fish
---
```

In `file-3.md`, I have
```
---
title: File 3
permalink: /file-3/
---

{% assign post_list = site.posts | where_contains: 'array_a', 'giraffes' %}
{% for post in post_list %}
- {{ post.title }}
{% endfor %}
```

And the output would be:
```
- File 2
```

At the moment, the only way to do this kind of thing seems to be to do a `for` loop with an `if x.y contains z` checking over each item, which isn't a huge problem, but is less than elegant and is probably rather slower for big collections than it'd be with more built-in handling.

davidbiehl:
How would you expect non-array values to be handled? I'm thinking it could fall-back to `where`. For example, if you had a `file-4.md` like this:

```
---
title: File 4
array_a: giraffes
---
```
Then `- File 4` would be included in the output using your example above. What do you think?

drumanagh:
That sounds like it'd be good. It'd cover some edge cases and I can't really think of a case where you specifically wouldn't want it to behave like that.

edgemaster:
I've implemented this with a generic where_exp filter of the form:
```liquid
{% assign post_list = site.posts | where_exp: "array_a contains 'giraffes'" %}
```
I'll reference this issue in a PR.

yalov:
Is there a way to combine `where_exp` with `unless`? Something like:
 
`{% assign posts_noncats = site.posts | where_exp: "item", unless item.category contains 'cats'" %}`

PR #4478, opened 2016-02-05
Add a where_exp filter for filtering by expression

edgemaster:
This commit introduces a where_exp filter, which can be used as follows:
  `{{ array | where_exp: "item", "item == 10" }}`
  `{{ array | where_exp: "item", "item.field > 10" }}`
  `{{ site.posts | where_exp: "post", "post.keys contains 'field'" }}`
  `{{ site.posts | where_exp: "post", "post.array contains 'giraffes'" }}`

This permits a variety of use cases, such as reported in: jekyll#4467,
jekyll#4385, jekyll#2787.

### DIFF
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -1,6 +1,7 @@
 require 'uri'
 require 'json'
 require 'date'
+require 'liquid'
 
 module Jekyll
   module Filters
@@ -225,6 +226,26 @@ module Jekyll
       input.select { |object| Array(item_property(object, property)).map(&:to_s).include?(value.to_s) }
     end
 
+    # Filters an array of objects against an expression
+    #
+    # input - the object array
+    # variable - the variable to assign each item to in the expression
+    # expression - a Liquid comparison expression passed in as a string
+    #
+    # Returns the filtered array of objects
+    def where_exp(input, variable, expression)
+      return input unless input.is_a?(Enumerable)
+      input = input.values if input.is_a?(Hash)
+
+      c = parse_condition(expression)
+      @context.stack do
+        input.select do |object|
+          @context[variable] = object
+          c.evaluate(@context)
+        end
+      end
+    end
+
     # Sort an array of objects
     #
     # input - the object array
@@ -362,6 +383,21 @@ module Jekyll
           item
         end
       end
+    end
+
+    # Parse a string to a Liquid Condition
+    def parse_condition(exp)
+      parser = Liquid::Parser.new(exp)
+      a = parser.expression
+      if op = parser.consume?(:comparison)
+        b = parser.expression
+        condition = Liquid::Condition.new(a, op, b)
+      else
+        condition = Liquid::Condition.new(a)
+      end
+      parser.consume(:end_of_string)
+
+      condition
     end
   end
 end
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -389,8 +389,9 @@ module Jekyll
     def parse_condition(exp)
       parser = Liquid::Parser.new(exp)
       left_expr = parser.expression
+      operator = parser.consume?(:comparison)
       condition =
-        if operator = parser.consume?(:comparison)
+        if operator
           Liquid::Condition.new(left_expr, operator, parser.expression)
         else
           Liquid::Condition.new(left_expr)
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -235,13 +235,13 @@ module Jekyll
     # Returns the filtered array of objects
     def where_exp(input, variable, expression)
       return input unless input.is_a?(Enumerable)
-      input = input.values if input.is_a?(Hash)
+      input = input.values if input.is_a?(Hash) # FIXME
 
-      c = parse_condition(expression)
+      condition = parse_condition(expression)
       @context.stack do
         input.select do |object|
           @context[variable] = object
-          c.evaluate(@context)
+          condition.evaluate(@context)
         end
       end
     end
@@ -388,13 +388,13 @@ module Jekyll
     # Parse a string to a Liquid Condition
     def parse_condition(exp)
       parser = Liquid::Parser.new(exp)
-      a = parser.expression
-      if op = parser.consume?(:comparison)
-        b = parser.expression
-        condition = Liquid::Condition.new(a, op, b)
-      else
-        condition = Liquid::Condition.new(a)
-      end
+      left_expr = parser.expression
+      condition =
+        if operator = parser.consume?(:comparison)
+          Liquid::Condition.new(left_expr, operator, parser.expression)
+        else
+          Liquid::Condition.new(left_expr)
+        end
       parser.consume(:end_of_string)
 
       condition
--- a/site/_docs/templates.md
+++ b/site/_docs/templates.md
@@ -100,7 +100,7 @@ common tasks easier.
          <code class="filter">{% raw %}{{ site.members | where_exp:"item",
 "item.graduation_year < 2014" }}{% endraw %}</code>
          <code class="filter">{% raw %}{{ site.members | where_exp:"item",
-"item.projects includes 'foo'" }}{% endraw %}</code>
+"item.projects contains 'foo'" }}{% endraw %}</code>
         </p>
       </td>
     </tr>
--- a/site/_docs/templates.md
+++ b/site/_docs/templates.md
@@ -90,6 +90,22 @@ common tasks easier.
     </tr>
     <tr>
       <td>
+        <p class="name"><strong>Where Expression</strong></p>
+        <p>Select all the objects in an array where the expression is true.</p>
+      </td>
+      <td class="align-center">
+        <p>
+         <code class="filter">{% raw %}{{ site.members | where_exp:"item",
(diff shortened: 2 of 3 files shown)


parkr:
I'm 👍 on this as a feature. I think this adds power to Jekyll without making it too complicated, as it's extending current functionality. I'd like to rename `where_exp` to something more user-friendly. Also wondering if `where` could do this:

```liquid
{{ array | where: "item", "item == 10" }}
{{ array | where: "field", "10", ">" }}
{{ site.posts | where: "keys", "field", "contains" }}
{{ site.posts | where: "array", "giraffes", "contains" }}
```

Postfix is hardly ever used, though, so I think I like yours better.

/cc @jekyll/documentation for documentation review.
/cc @jekyll/stability to help with the name for this filter. Perhaps `matches`?

benbalter:
A meta-question, that shouldn't derail the discussion here, but should we be adding non-Jekyll-specific liquid filters to Jekyll? Shouldn't we instead try to get those adopted upstream in Liquid itself (and its fancy new documentation)? Would they accept something like this? If not, why not?

site/_docs/templates.md
```diff
@@ -90,6 +90,22 @@ common tasks easier.
...
+        <p>Select all the objects in an array where the expression is true.</p>
+      </td>
+      <td class="align-center">
+        <p>
+         <code class="filter">{% raw %}{{ site.members | where_exp:"item",
+"item.graduation_year == 2014" }}{% endraw %}</code>
+         <code class="filter">{% raw %}{{ site.members | where_exp:"item",
+"item.graduation_year < 2014" }}{% endraw %}</code>
+         <code class="filter">{% raw %}{{ site.members | where_exp:"item",
+"item.projects includes 'foo'" }}{% endraw %}</code>
```
mneumegen: Should this be `"item.projects contains 'foo'`?

mneumegen:
Love this feature. I'm not a big fan of the postfix notation, I find it much harder to read.

site/_docs/templates.md
edgemaster: Good catch. I have a feeling I need to write more test cases for the `contains` case also as I tripped myself up using it in production.

edgemaster:
@benbalter Since this relies on liquid internals(?), the best place probably is upstream. However, this was in direct response to a feature request here (#4385) though, so I thought it could be of immediate use to the original reporter.

parkr:
> should we be adding non-Jekyll-specific liquid filters to Jekyll? Shouldn't we instead try to get those adopted upstream in Liquid itself

@benbalter We have a much stronger case to make later if we can say "Look how this is used in Jekyll – wouldn't it be super cool if you could use them everywhere there is Liquid?" The bar for adding new features to Liquid core is fairly high as it impacts all themes on Shopify.com. It's worth submitting, but as Liquid is on v4 already ( #4362 ), any new releases there would still require a Jekyll bump. That, combined with the fact that this is using Liquid 3 internals may cause problems.

@edgemaster I wonder if it's possible to unify on one variable name, like `item`. It's a bit confusing for new users, though, so that's why I'm OK with it as-is (verbosity can sometimes be really helpful). Would you please add more test cases for `contains` as noted and fix up the docs? It looks solid to me but any added assurance is good by me.

lib/jekyll/filters.rb
```diff
@@ -225,6 +226,26 @@ def where(input, property, value)
...
+    # Filters an array of objects against an expression
+    #
+    # input - the object array
+    # variable - the variable to assign each item to in the expression
+    # expression - a Liquid comparison expression passed in as a string
+    #
+    # Returns the filtered array of objects
+    def where_exp(input, variable, expression)
+      return input unless input.is_a?(Enumerable)
+      input = input.values if input.is_a?(Hash)
```
envygeeks: This will cache many off-guard.  I disagree with this.

lib/jekyll/filters.rb
```diff
@@ -363,5 +384,20 @@ def as_liquid(item)
         end
       end
     end
+
+    # Parse a string to a Liquid Condition
+    def parse_condition(exp)
+      parser = Liquid::Parser.new(exp)
+      a = parser.expression
```
envygeeks: Please make these variable names mean something.

lib/jekyll/filters.rb
```diff
@@ -363,5 +384,20 @@ def as_liquid(item)
...
+      if op = parser.consume?(:comparison)
+        b = parser.expression
```
envygeeks: ""

lib/jekyll/filters.rb
```diff
@@ -363,5 +384,20 @@ def as_liquid(item)
...
+        condition = Liquid::Condition.new(a, op, b)
```
envygeeks: Please do:

```
var = \
if condition
  val1
else
  val2
end
```

Or for bonus points and extreme readability:

```
var = begin
  if condition
    val1
  else
    val2
  end
end
```

parkr: Hmm, is that in our Rubocop setup? I don't like either of those personally – I find them both harder to read than the current version. Indentation in the first example doesn't help me know that the `if` statement affects `var` and the second one is noisy. I think with better variable names (not just `a`, `b`, and `op`), we can make this code more readable.

envygeeks: Our Rubocop setup is setup to accept any style as long as it's not multiple defines.  So if you like to indent on the var like:

```ruby
var = \
  if blah
    var
  else
    var
  end
```

It'll accept it.  Those two are just recommendations, the former can be added.

lib/jekyll/filters.rb
edgemaster: My motivation was to follow the existing semantics of `where` over hashes as in https://github.com/jekyll/jekyll/blob/b06af5a44f47ec6cf5f203e8eb318868eac0ae86/lib/jekyll/filters.rb#L224

edgemaster:
Made the minor adjustments as suggested, except for https://github.com/jekyll/jekyll/pull/4478#discussion_r57447663, I have chosen to filter over the values of the hash to match how the existing `where` filter behaves.

parkr:
Ooooh, people are going to love this feature. 👍 ❤️ 

@jekyllbot: merge +minor

edgemaster:
Oh, wow. I wasn't expecting this to be landed without the more major changes (eg, to support a 3-argument where filter) suggested by @parkr

pathawks:
This doesn't seem to work with `site.posts` at all, unless I am missing something&hellip;?

parkr:
@pathawks Can you provide an example of something you're unable to accomplish? Perhaps add a test and submit a PR?

budparr:
Is this in 3.1x? I see the 3.2 milestone, but it's in the docs already, so I assume it's available now. https://jekyllrb.com/docs/templates/#filters

edgemaster:
It doesn't appear to have landed in v3.1.3.

edgemaster:
With regards to it not working with `site.posts`, would I be correct in thinking that the appropriate place to add a test for this would be in `features/embed_filters.feature`?

envygeeks:
@edgemaster no, it's a unit problem, not a feature problem.  Your feature should define a high level set of behaviors, your units should mutate those behaviors to ensure your code is up to snuff.

ryrych:
Hi, is there any alternative to `where_exp` till the feature is released / supported on Github pages? I'd like to filter out posts which contain *protip* path. I can add another property to a post like `protip: true` and iterate over the collection: `{% assign posts = site.posts | where: "protip", "false" %}`. The problem is that it doesn't go well with pagination.

Yeah, sorry if the question was better asked for on StackOverflow.

parkr:
> Hi, is there any alternative to where_exp till the feature is released / supported on Github pages?

@ryrych Not directly – what this filter does is collapse:

```liquid
{% assign things = "" | split: "|" %}
{% for item in my_array %}
{% if <EXPR> %}
  {% assign things = things | push: item %}
{% endif %}
{% endfor %}
```

into 

```liquid
{% assign things = my_array | where_exp: "item", "<EXPR>" %} 
```

This will have to wait until GitHub Pages upgrades to 3.2 to use on GitHub Pages.

TWiStErRob:
What is the language of the expression? I guess it's Liquid and not Ruby.

su-narthur:
This is a pretty cool feature! The thing it doesn't address, though are expressions that reference other variables:

```
my_page.permalink contains section.permalink
```

I guess that's probably too much complexity to cram into a filter.

mryellow:
https://github.com/jekyll/jekyll/pull/4478#issuecomment-209038370

> (In fact, filtering over the keys of the input hash probably wouldn't be the intended operation for the end-user, anyway).

Unfortunately that is my use-case and this seems to come up lacking. I have a hash with epochs as keys which needs to be filtered for begin/finish dates.

Filtering with `where_exp:"item", "item.key > 0"` heads in the right direction but results in only the value being output rather than the key and value.

Morikko:
```
{% assign guides = site.pages
    | where_exp: "item" , "item.title  contains 'Data' or item.title  contains 'Groups'"
 %}
```
I got : `Error: Liquid syntax error (.../guide.html line 12): Expected end_of_string but found id included`

Are the operators `or` and `and` not supported in where_exp ?

Note: actually `and` can be done by chaining many where_exp but not `or`.

pathawks:
@Morikko This PR is pretty old. You may want to open a fresh issue.

PR #4479, opened 2016-02-05
EntryFilter: only include 'excluded' log on excluded files

parkr:
this is a 3.1 change that i'd love to have in 3.0. will help clear up verbose logs which get mighty verbose when every file jekyll looks at is printed :/

/cc @jekyll/gh-pages


benbalter:
Sounds good to me.

parkr:
`You have to install development tools first.`

Travis is off the hook today. Just killin' it.

PR #4491, opened 2016-02-08
EntryFilter#special?: ignore filenames which begin with '~'

parkr:
Per https://github.com/jekyll/jekyll/pull/4468#issuecomment-179954048

In Jekyll 3.2, we will ignore these by default. I don't believe this is a breaking change
given my reasoning in the link, but feel free to disagree.

/cc @benbalter @jekyll/core @jekyll/gh-pages


parkr:
@jekyllbot: merge +bug

PR #4492, opened 2016-02-08
Jekyll.sanitized_path: sanitizing a questionable path should handle tildes

### LINKED ISSUE #4457: Files that start with a tilde (~) break builds

parkr:
Fixes #4457 (and duplicates #4468) on `master`, targeted for v3.1.2.

### DIFF
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -153,8 +153,9 @@ module Jekyll
     def sanitized_path(base_directory, questionable_path)
       return base_directory if base_directory.eql?(questionable_path)
 
+      questionable_path.insert(0, '/') if questionable_path.start_with?('~')
       clean_path = File.expand_path(questionable_path, "/")
-      clean_path = clean_path.sub(/\A\w\:\//, '/')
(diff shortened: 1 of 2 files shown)


parkr:
@jekyllbot: merge +bug

PR #4497, opened 2016-02-08
Allow users to set `highlighter`/`kramdown.syntax_highlighter` to `nil`

### LINKED ISSUE #4494: unable to overwrite existing hash value with null
A recent change in `Utils::deep_merge_hashes!` (this happened in 5d79c55b2cee0aa19ca57ad331333edba9db47b7) broke the ability to overwrite an existing value in a hash with nil.  This is because of the obvious:

```ruby
if new_val.nil?
  old_val
```

An example of how this previous behavior is useful can be seen in my jekyll config files.  In my [main _config.yml](https://github.com/willnorris/willnorris.com/blob/99f27a1b2b903532a75ded22e8ccc3c9f2bee535/_config.yml#L5), I set a `imageproxy_baseurl` key, but in [my _debug.yml](https://github.com/willnorris/willnorris.com/blob/99f27a1b2b903532a75ded22e8ccc3c9f2bee535/_debug.yml#L2) I set this back to `null`.  My [startup script for local development](https://github.com/willnorris/willnorris.com/blob/99f27a1b2b903532a75ded22e8ccc3c9f2bee535/tools/run) loads these two files in order, and expects there to be no value for this key when run in development.

I can certainly work around this on my side (and I recall there being a proposal for loading different config files based on the value of `JEKYLL_ENV` or something... not sure if that ever went anywhere, maybe I should be using that. *update: it's #3173*).  But I wanted to bring it up because this is such a widely used utility function, and I suspect there may be other places where people intentionally override an existing value with null.

/cc @parkr

envygeeks:
Fixes #4494.

### DIFF
--- a/test/test_kramdown.rb
+++ b/test/test_kramdown.rb
@@ -31,6 +31,24 @@ class TestKramdown < JekyllUnitTest
       assert_equal "<h1>Some Header</h1>", @markdown.convert('# Some Header #').strip
     end
 
+    context "when asked to disable the highlighter" do
+      should "accept nil" do
+        override = {
+          "kramdown" => {
+            "syntax_highlighter" => nil
+          }
+        }
+
+        markdown = Converters::Markdown.new(Utils.deep_merge_hashes(
+          @config, override
+        ))
+
+        result = nokogiri_fragment(markdown.convert "```ruby\nhello\n```")
+        select = "pre/code[contains(@class, 'language-ruby')][contains(text(),'hello')]"
+        refute result.xpath(select).empty?
+      end
+    end
+
     context "when asked to convert smart quotes" do
       should "convert" do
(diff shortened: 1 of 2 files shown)


test/test_kramdown.rb
```diff
@@ -30,6 +30,24 @@ class TestKramdown < JekyllUnitTest
...
+          }
+        }
+
+        markdown = Converters::Markdown.new(Utils.deep_merge_hashes(
+          @config, override
+        ))
+
+        result = nokogiri_fragment(markdown.convert "```ruby\nhello\n```")
+        select = "pre/code[contains(@class, 'language-ruby')][contains(text(),'hello')]"
+        refute result.xpath(select).empty?
```
parkr: Doesn't look like this refutation is passing: https://travis-ci.org/jekyll/jekyll/jobs/107765459#L345

parkr:
@envygeeks Do you know why that might be failing?

PR #4503, opened 2016-02-09
Add user directory_index option for serve command

hgoodman:
For development I find it useful to always have WEBrick show a directory listing, plus I figured that others might want to provide a custom list of directory index files, so I added that option for the serve command.

This commit adds the `--no-directory-index` command line option as well as the ability to provide a comma-separated list of index files with `--directory-index [...]`. Putting `directory_index` in the config file works too.

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -10,6 +10,8 @@ module Jekyll
           "ssl_key"  => ["--ssl-key [KEY]", "X.509 (SSL) Private Key."],
           "port"     => ["-P", "--port [PORT]", "Port to listen on"],
           "baseurl"  => ["-b", "--baseurl [URL]", "Base URL"],
+          "directory_index" => ["--[no-]directory-index [FILE1,[FILE2,...]]", Array,
+            "Directory index files"],
           "skip_initial_build" => ["skip_initial_build", "--skip-initial-build",
             "Skips the initial site build which occurs before the server is started."]
         }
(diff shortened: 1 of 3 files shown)


parkr:
Hey, @hgoodman! Thanks for the pull request (with tests!!)!

> For development I find it useful to always have WEBrick show a directory listing

What do you mean by this? Every time you go to a directory, regardless of its contents, it shows the files it contains instead of some `index.html`? I can see this being useful if you don't want to `tree _site` or just want to browse your files like the old FTP days.

> the ability to provide a comma-separated list of index files

What's the motivation for this? We try to keep parity with [GitHub Pages](https://pages.github.com/) and this would break parity. Just want to understand its value before pulling the trigger, so to speak.

Thanks again for the great pull request! :smile_cat:

hgoodman:
What I meant was that it's easier to click through an old-style listing to get to a particular file (not all my pages are linked) than to type in the file name or copy and paste it into my browser address bar.

As for the feature of allowing custom index files, I personally don't need it, but I thought it would make sense to bundle it along with the ```--no-directory-index``` option since it's the same WEBrick option that's being modified. I can take it out if you like, but it seems harmless since the defaults haven't changed.

hgoodman:
@parkr, do you have a documentation link to the list of directory index files that are accepted by GitHub Pages? It might be good to add that in a comment next to the defaults.

Also, did you need me to remove the option to override the defaults?

parkr:
> Also, did you need me to remove the option to override the defaults?

If you don't mind, yes. They should always match the GitHub Pages defaults. I don't have a link unfortunately but from some testing, I found: `index.html`, `index.htm`, `index.xml`, and `index.xhtml`. `.cgi` and `.rhtml` are not presently supported. @benbalter, could we document this somewhere in help.github.com?

hgoodman:
This PR is superseded by jekyll/jekyll/pull/4533. I figured it would make the commit history cleaner by committing to a more recent version, and that it would make more sense since I'm essentially reversing and rewriting the update. Unfortunately, Travis CI isn't happy for some reason, but that's a separate issue.

PR #4505, opened 2016-02-10
add consistency to the deprecation message

lonnen:
The deprecation message that presents when you use switches says jekyll now uses subcommands. Then it prompts you to learn more by running jekyll with a switch for `--help`. Conveniently Jekyll supports both, but the message seems self-contradictory.

This proposes to make the message internally consistent without otherwise changing behavior.


parkr:
This looks good to me, thank you!

parkr:
@jekyllbot: merge +minor

PR #4508, opened 2016-02-11
Add a format filter

JesseTG:
Thought it might be useful.  However, I know very little Ruby.

### DIFF
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -14,6 +14,29 @@ module Jekyll
       converter = site.find_converter_instance(Jekyll::Converters::Markdown)
       converter.convert(input)
     end
+    
+    # Format a string with Kernel::format
+    #
+    # input - The format string
+    # args - The arguments to input
+    #
+    # Returns the formatted string
+    def format(input, *args)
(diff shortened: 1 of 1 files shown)


JesseTG:
Ah, right, I should probably add tests.  I'll do so tomorrow, gotta get some sleep.

parkr:
Thanks, @JesseTG! Let me know if you need help on those tests.

What is your use-case for this? Why can't this be done another way?

JesseTG:
Honestly?  I can't think of one.  I wrote this for a project last summer 
but never wound up using it (switched technology).  I just didn't want 
to see it go to waste; I'm sure there's /some/ edge case that Jekyll's 
existing tags and filters don't cover, and I suppose this could be good 
for that.  So, to answer your question, "just in case."

parkr:
@JesseTG Haha, ok. This seems general purpose enough to be good for Liquid's [standard filters](https://github.com/Shopify/liquid/blob/master/lib/liquid/standardfilters.rb) library if you can find a solid use case. We haven't needed it so far so I think I'll pass.

I appreciate the PR, though! :heart:

PR #4522, opened 2016-02-15
Update structure.md

bojanland:
Two grammatical corrections. :)

### DIFF
--- a/site/_docs/structure.md
+++ b/site/_docs/structure.md
@@ -7,9 +7,9 @@ permalink: /docs/structure/
 Jekyll is, at its core, a text transformation engine. The concept behind the
 system is this: you give it text written in your favorite markup language, be
 that Markdown, Textile, or just plain HTML, and it churns that through a layout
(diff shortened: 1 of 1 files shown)


site/_docs/structure.md
```diff
@@ -7,9 +7,9 @@ permalink: /docs/structure/
 Jekyll is, at its core, a text transformation engine. The concept behind the
 system is this: you give it text written in your favorite markup language, be
 that Markdown, Textile, or just plain HTML, and it churns that through a layout
-or series of layout files. Throughout that process you can tweak how you want
+or a series of layout files. Throughout that process you can tweak how you want
```
parkr: This is interesting. `series of layout files` does work by itself as it is technically acceptable to start the list _after_ the `a`:

```text
churns through a ['layout, 'series of layout files']
```

I think it's not as common to see that split, however, but an interesting linguistic oddity nevertheless!

parkr:
@jekyllbot: merge +site

PR #4523, opened 2016-02-15
Fix pagination example

irarayjenkins:
The previous version produced an incorrect href for page "1" on all pages after page "2". This version depends on site.paginate_path. By removing the paging information, the href becomes the "base" url of the index.html file that begins the pagination, that is, page "1". A site-wide variable might be convenient, but would eliminate multiple paginations beginning at different locations. An alternative may be to add a "baseurl" to paginator?

### DIFF
--- a/site/_docs/pagination.md
+++ b/site/_docs/pagination.md
@@ -207,7 +207,7 @@ page with links to all but the current page.
     {% if page == paginator.page %}
       <em>{{ page }}</em>
     {% elsif page == 1 %}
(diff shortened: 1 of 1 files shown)


parkr:
Hey @irarayjenkins, `paginator.previous_page_path` is the correct way to do this as prescribed. If this is returning `/page/1` for you, then it's a bug in jekyll-paginate. [Here is the prescription](https://github.com/jekyll/jekyll-paginate/blob/a2f213532b0f92420f24aad0ea7249bbe6d5feb8/lib/jekyll-paginate/pager.rb#L65). Thanks!

PR #4525, opened 2016-02-15
Fix titleize so already capitalized words are not dropped

atomicules:
Previously `titleize` used `capitalize!` which has the side effect of
returning `nil` for anything already starting with a capital letter. This
commit changes it to just `capitalize`.

Example, before:

A file "2016-01-01-This-is-a-title-with-Capitals.markdown" would return "Is A
Title With" for `post.title`

Example, after:

A file "2016-01-01-This-is-a-title-with-Capitals.markdown" will return "This Is A
Title With Capitals" for `post.title`

Note: this is based off tag v3.1.1 as at the time of cloning this master
was causing me problems with markdown parsing on one of my posts and I
only ended up looking into this because I was trying to write a blog
post after updating Jekyll to 3.1.1.

Not sure if tests are in the right place. Seems to be correct place as
have added tests for differently capitalised filenames with and without
dates.

Fix problem introduced in 67f842546efa980146778c85fb613e6c9b53dcb4

### DIFF
--- a/lib/jekyll/utils.rb
+++ b/lib/jekyll/utils.rb
@@ -20,7 +20,7 @@ module Jekyll
 
     def titleize_slug(slug)
       slug.split("-").map! do |val|
-        val.capitalize!
+        val.capitalize
       end.join(" ")
     end
 
--- a/test/test_utils.rb
+++ b/test/test_utils.rb
@@ -206,6 +206,14 @@ class TestUtils < JekyllUnitTest
     end
   end
 
+  context "The \`Utils.titleize_slug\` method" do
+    should "capitalize all words and not drop any words" do
+      assert_equal "This Is A Long Title With Mixed Capitalization", Utils.titleize_slug("This-is-a-Long-title-with-Mixed-capitalization")
+      assert_equal "This Is A Title With Just The Initial Word Capitalized", Utils.titleize_slug("This-is-a-title-with-just-the-initial-word-capitalized")
+      assert_equal "This Is A Title With No Capitalization", Utils.titleize_slug("this-is-a-title-with-no-capitalization")
+    end
+  end
+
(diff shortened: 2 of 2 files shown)


envygeeks:
This is a complicated fix to a simple problem.  There is no need for an entire scenario, a test case inside of tests sure, and without the extra fixtures as you only need to test to ensure that the method returns all the proper words.

atomicules:
Thanks for taking a look.

On 15-Feb-2016 08:02:59, Jordon Bedwell wrote:
>This is a complicated fix to a simple problem.

Well, the fix is simple, but the tests I'll give you :)

>There is no need for an entire scenario, a test case inside of tests 
>sure, and without the extra fixtures as you only need to test to 
>ensure that the method returns all the proper words.

If you could give me guidance as to where I should fit the tests in 
that would speed up me being able to implement it, but I'll start taking a 
look anyway to simplify them.

My reason for wanting to test filenames with and without dates is 
because I wasn't sure to start with if `titleize` was using 
`capitalize!` intentionally to remove the date from the start of the 
filename.

atomicules:
I'll look at adding something to test_utils.rb and update here when 
I've got it.

envygeeks:
test_utils.rb is the perfect place for it! And yes, the fix is super simple and thanks for catching it.  I didn't even think about that when we switched to `#capitalize!`

atomicules:
Ok, added tests under test_utils.rb. You may still want me to reduce 
the number of tests, let me know. I have reverted all the other tests.

I'm guessing you'll want this cleaned up / squashed before merging?

envygeeks:
Yes please!

atomicules:
Ok, squashed those. Let me know if there is anything else that needs 
doing or changing.

parkr:
This looks great to me. My gratitude!

@jekyllbot: merge +bug

Issue #4458, opened 2016-02-03
Scope values not applied


borisyankov:
I am using the approach to localize pages, described here:
https://www.sylvaindurand.org/making-jekyll-multilingual/

It was working perfectly on 2.x, but now both on GH Pages, and locally (3.1.1) it does not work.

When there is lang: en for root, all subfolders do have lang: en (though they are set otherwise)
If the lang: en for root is removed, the subfolders also have no value set.
It seems the variables don't work at all for the subfolders.

1nsunym:
I've got the exact same problem.

borisyankov:
@1nsunym unfortunately not.
I think it is genuine bug, and not some behaviour that changed.
And not aware of a workaround. But looking urgently for a fix/workaround.

1nsunym:
@borisyankov Mines only a small blog, so my fix-around was setting 'en' default for lang variables and setting lang variable for other language post/pages explicitly on frontmatters.

sylvaindurand:
I'm working on it. 
Defaults doesn't seem to work anymore, I'm trying to understand why!

borisyankov:
@sylvaindurand Great that you will update your article. I used it as a guideline an year ago.
Seems to be part of the 'distributed documentation' of Jekyll. ;)

Crunch09:
@borisyankov @sylvaindurand 
Hi, i tried to come up with a testcase to reproduce this bug but i wasn't able to get it to fail both on 3.1 and master. Could you maybe take a [look at it](https://gist.github.com/Crunch09/b41156779c998e9a3c6b) an see if i missed something?

parkr:
@Crunch09 Were you able to test this on v3.0.3? Check out the `v3.0.3` tag in your fork (I do this via `git checkout v3.0.3` in the shell) and add this test and run it. It may have been fixed in 3.1.

Crunch09:
@parkr Thanks! You were right, the test is failing on `v3.0.3`. Should i submit a PR with this test to prevent further regression?

parkr:
> @parkr Thanks! You were right, the test is failing on v3.0.3. Should i submit a PR with this test to prevent further regression?

Yes, please! Submit a patch based on the `3.0-stable` branch and pointed to that branch instead of `master`.

PR #4530, opened 2016-02-16
add failing test for scope values in subdirs

Crunch09:
ref #4458


parkr:
Thank you!

parkr:
A patch for this fundamentally changes the way `Cleaner` works. We will have to check that the full path of the file (which we don't necessarily have) matches any part of the include.

parkr:
Fixed in #4806.

PR #4531, opened 2016-02-16
Add a redirect for Pull Requests

chrisfinazzo:
Unless I misunderstood the intent [here], it seems like this would be useful to have. The last 3 are probably not used as frequently and can be left as-is.

[here]: https://github.com/jekyll/jekyll/commit/d38d90f2782e630660d3db55ebb3a60f810b0b1d

### DIFF
--- a/site/redirects/pull-requests.html
+++ b/site/redirects/pull-requests.html
@@ -1,0 +1,4 @@
+---
+permalink: /pull-requests.html
+redirect_to: https://github.com/jekyll/jekyll/pulls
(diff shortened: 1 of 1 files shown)


parkr:
@chrisfinazzo I added those so I could future-proof URL's to issues and to our repo. We aren't really in the business of redirecting people to the list of pulls unless they are a maintainer and in that case they're probably already on the page with the list of pulls, like me. :smile: Thanks though!

PR #4533, opened 2016-02-17
Add show_dir_listing option for serve command and fix index file names

hgoodman:
This is a replacement for PR jekyll/jekyll/pull/4503, which I will close momentarily.

This PR adds the ```--show-dir-listing``` CLI option, which can also be included in the config as ```show_dir_listing```, to show the WEBrick default directory listing instead of loading an index file (e.g. index.html). When testing, I find it much more convenient to be able to navigate to a specific file by clicking through a directory list instead of typing in the file name (which may not be linked from index.html).

This PR also fixes the default directory index file list to maintain parity with the file names supported by GitHub Pages (I tested to verify that the order is correct).

I changed the negative option ```--no-directory-index``` from PR jekyll/jekyll/pull/4503 to the affirmative ```--show-dir-listing``` for the sake of clarity and also because I noticed that GitHub Pages gives its own 404 page when an index file is not present.

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -10,6 +10,8 @@ module Jekyll
           "ssl_key"  => ["--ssl-key [KEY]", "X.509 (SSL) Private Key."],
           "port"     => ["-P", "--port [PORT]", "Port to listen on"],
           "baseurl"  => ["-b", "--baseurl [URL]", "Base URL"],
+          "show_dir_listing" => ["--show-dir-listing",
+            "Show a directory listing instead of loading your index file."],
           "skip_initial_build" => ["skip_initial_build", "--skip-initial-build",
             "Skips the initial site build which occurs before the server is started."]
         }
@@ -90,6 +92,8 @@ module Jekyll
               index.xml
             )
           }
+
+          opts[:DirectoryIndex] = [] if opts[:JekyllOptions]['show_dir_listing']
 
           enable_ssl(opts)
           enable_logging(opts)
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -44,6 +44,7 @@ module Jekyll
       'port'          => '4000',
       'host'          => '127.0.0.1',
       'baseurl'       => '',
+      'show_dir_listing' => false,
 
       # Output Configuration
       'permalink'     => 'date',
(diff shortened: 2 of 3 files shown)


hgoodman:
Well, this is awkward (tests failing, I mean). I'm pretty sure it's caused by the execution order since changing the random seed makes it pass and setting it to the value that Travis used makes it fail. My first look at the problem was unfruitful but I'll give it another shot.

hgoodman:
The latest commit emulates GitHub Pages by returning an error when trying to list directory contents in WEBrick unless it's made clear that the user intends to override that behavior in a dev environment with the ```--show-dir-listing``` option.

There's that, plus I kinda wanted to take another roll of the dice with Travis CI. :wink:

lib/jekyll/commands/serve.rb
```diff
@@ -99,9 +103,9 @@ def webrick_opts(opts)
         # Recreate NondisclosureName under utf-8 circumstance
 
         private
-        def file_handler_opts
+        def file_handler_opts(opts)
           WEBrick::Config::FileHandler.merge({
-            :FancyIndexing     => true,
+            :FancyIndexing     => opts['show_dir_listing'] ? true : false,
```
parkr: Shouldn't this be flipped? Only do fancy indexing if `show_dir_listing` is false?

hgoodman: No @parkr, a "fancy" index in WEBrick-speak turns out to mean showing the contents of the directory when no index is present, which is what I'm trying to do. This is affirmed by the tests.

Anyway, @envygeeks made a good point about not changing behavior for a point release so I was planning to revert it anyway, so as not to hold up the merge. I thought of it as more of a bug fix based on what you were saying about parity with GHP. You decide.

envygeeks: @hgoodman FancyIndexing by default was a conscious disparity between production and development to enforce WEBrick only being dev quality (as used by us) and to discourage the many users we (or maybe just me?) using it in production.  People dislike directory indexing in production unless it's for a download folder which helps us let them know this is not meant for production by any means.

hgoodman: Right on, @envygeeks, I don't like it in production either and as I said, I'm planning on reverting it later.

hgoodman:
I rolled back the index file changes and rebased the branch. Let me know if there's anything else I can do to improve index file handling. Thank you for the thorough review!

envygeeks:
:+1:

parkr:
LGTM. Thanks for the PR and thanks for your thorough work! Will merge once we have 3.1.2 out.

parkr:
@jekyllbot: merge +minor

Issue #4476, opened 2016-02-05
Fix warnings


parkr:
Was running some specs for github-metadata and noticed some annoying warnings in the latest version. We should fix these. :smile: 

- [ ] `lib/jekyll/regenerator.rb:147: warning: instance variable @disabled not initialized`
- [ ] `lib/jekyll/frontmatter_defaults.rb:97: warning: shadowing outer local variable - path`
- [ ] `lib/jekyll/url.rb:119:in 'escape_path': warning: URI.escape is obsolete`
- [ ] `lib/jekyll/url.rb:133:in 'unescape_path': warning: URI.unescape is obsolete`
- [ ] `lib/jekyll/converter.rb:12: warning: instance variable @highlighter_prefix not initialized`
- [ ] `lib/jekyll/converter.rb:24: warning: instance variable @highlighter_suffix not initialized`

From other libs in case you want to submit a PR to them:

- `liquid-3.0.6/lib/liquid/drop.rb:65: warning: instance variable @invokable_methods not initialized`
- `liquid-3.0.6/lib/liquid/template.rb:86: warning: instance variable @error_mode not initialized`
- `liquid-3.0.6/lib/liquid/template.rb:90: warning: instance variable @taint_mode not initialized`
- `liquid-3.0.6/lib/liquid/drop.rb:65: warning: instance variable @invokable_methods not initialized`

DmitryKa:
How can I reproduce it?

Crunch09:
@DmitryKa You need to pass the `-w` option to ruby to show the warnings.
@parkr How should we replace `URI.escape` ? I tried different solutions and ended up with the `addressable` gem according to this [StackOverflow answer](http://stackoverflow.com/a/13059657) but it caused a regression on [this test case](https://github.com/jekyll/jekyll/blob/4c8c59dfcb8b80c8dd3d8df562a9d664be10348c/test/test_page.rb#L38-L41):

```
Expected: "/+/%25%23%20+.html"
Actual: "/+/%25#%20+.html"
```
I am not sure how to proceed here?

envygeeks:
@Crunch09 `adressable` is a sledgehammer.  Use `CGI.escape`, it is what you are looking for.

Crunch09:
@envygeeks Thanks, will do!

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least
one month.

The resources of the Jekyll team are limited, and so we are asking for your help.

[...]

PR #4537, opened 2016-02-17
Fix many warnings

Crunch09:
Hi,
you can see all the warnings this PR removes in the commit message. This commit does not remove the following warnings:
- `lib/jekyll/url.rb:119:in 'escape_path': warning: URI.escape is obsolete`
- ` lib/jekyll/url.rb:133:in 'unescape_path': warning: URI.unescape is obsolete`

which were mentioned in #4476 . I tried to replace the calls with `CGI.escape` /  `CGI.unescape` but i couldn't get them to return the same results as the current implementation and i thought it would be too risky to replace it with a different behavior just to get rid of those warnings.

### DIFF
--- a/lib/jekyll/converter.rb
+++ b/lib/jekyll/converter.rb
@@ -8,7 +8,9 @@ module Jekyll
     #
     # Returns the String prefix.
     def self.highlighter_prefix(highlighter_prefix = nil)
-      @highlighter_prefix = highlighter_prefix if highlighter_prefix
+      if !defined?(@highlighter_prefix) || !highlighter_prefix.nil?
+        @highlighter_prefix = highlighter_prefix
+      end
       @highlighter_prefix
     end
 
@@ -20,7 +22,9 @@ module Jekyll
     #
     # Returns the String suffix.
     def self.highlighter_suffix(highlighter_suffix = nil)
-      @highlighter_suffix = highlighter_suffix if highlighter_suffix
+      if !defined?(@highlighter_suffix) || !highlighter_suffix.nil?
+        @highlighter_suffix = highlighter_suffix
+      end
       @highlighter_suffix
     end
 
--- a/lib/jekyll/converters/markdown/kramdown_parser.rb
+++ b/lib/jekyll/converters/markdown/kramdown_parser.rb
@@ -18,6 +18,7 @@ module Jekyll
           Jekyll::External.require_with_graceful_fail "kramdown"
           @main_fallback_highlighter = config["highlighter"] || "rouge"
           @config = config["kramdown"] || {}
+          @highlighter = nil
           setup
         end
 
--- a/lib/jekyll/plugin.rb
+++ b/lib/jekyll/plugin.rb
@@ -60,7 +60,7 @@ module Jekyll
     #
     # Returns the safety Boolean.
     def self.safe(safe = nil)
-      if safe
+      if !defined?(@safe) || !safe.nil?
         @safe = safe
       end
       @safe || false
--- a/lib/jekyll/regenerator.rb
+++ b/lib/jekyll/regenerator.rb
@@ -1,6 +1,8 @@
 module Jekyll
   class Regenerator
     attr_reader :site, :metadata, :cache
+    attr_accessor :disabled
+    private :disabled, :disabled=
 
     def initialize(site)
       @site = site
@@ -115,7 +117,7 @@ module Jekyll
     #
     # Returns nothing.
     def add_dependency(path, dependency)
-      return if metadata[path].nil? || @disabled
+      return if metadata[path].nil? || disabled
 
       unless metadata[path]["deps"].include? dependency
         metadata[path]["deps"] << dependency
@@ -144,8 +146,8 @@ module Jekyll
     #
     # Returns a Boolean (true for disabled, false for enabled).
     def disabled?
-      @disabled = !site.incremental? if @disabled.nil?
-      @disabled
+      self.disabled = !site.incremental? if disabled.nil?
+      disabled
     end
 
     private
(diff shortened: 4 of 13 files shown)


envygeeks:
Awesome, thanks for the submission, just a few comments!

Crunch09:
@envygeeks @parkr Thanks for your review, i tried to address your concerns and updated the code.

parkr:
Hey @Crunch09! There are some problems with this PR, specifically around changes of behaviour. I will address them for you now.

lib/jekyll/converter.rb
```diff
@@ -8,7 +8,9 @@ class Converter < Plugin
     #
     # Returns the String prefix.
     def self.highlighter_prefix(highlighter_prefix = nil)
-      @highlighter_prefix = highlighter_prefix if highlighter_prefix
+      if !defined?(@highlighter_prefix) || !highlighter_prefix.nil?
```
parkr: This changes the behaviour slightly. Before: If `highlighter_prefix` is truthy, set its value to `@highlighter_prefix`. After: if `highlighter_prefix` is not nil or `@highlighter_prefix` isn't defined, then it sets `@highlighter_prefix` to `highlighter_prefix`. 

Why add the `!defined?` check?

lib/jekyll/converter.rb
```diff
@@ -20,7 +22,9 @@ def self.highlighter_prefix(highlighter_prefix = nil)
     #
     # Returns the String suffix.
     def self.highlighter_suffix(highlighter_suffix = nil)
-      @highlighter_suffix = highlighter_suffix if highlighter_suffix
+      if !defined?(@highlighter_suffix) || !highlighter_suffix.nil?
+        @highlighter_suffix = highlighter_suffix
```
parkr: Same concern as I had above with `_prefix`.

lib/jekyll/converters/markdown/kramdown_parser.rb
```diff
@@ -57,7 +57,7 @@ def make_accessible(hash = @config)
 
         private
         def highlighter
-          return @highlighter if @highlighter
+          return @highlighter if @highlighter ||= nil
```
parkr: This is redundant.

lib/jekyll/plugin.rb
```diff
@@ -60,10 +60,7 @@ def self.priority(priority = nil)
     #
     # Returns the safety Boolean.
     def self.safe(safe = nil)
-      if safe
-        @safe = safe
-      end
-      @safe || false
+      @safe ||= !!safe
```
parkr: This changes the behaviour. Before: If `safe` is truthy, set its value to `@safe`. After: set `@safe` to boolean equivalent of truthiness of `safe` only if `@safe` is not set.

This is a breaking change.

lib/jekyll/regenerator.rb
```diff
@@ -4,6 +4,7 @@ class Regenerator
 
     def initialize(site)
       @site = site
+      @disabled = nil
```
parkr: Can we create a `private` `attr_accessor`? You may have to do this via:

```ruby
attr_accessor :disabled
private :disabled, :disabled=
```

and change all references of `@disabled` to `disabled` and `@disabled =` to `self.disabled =`

test/test_site.rb
```diff
@@ -519,7 +519,7 @@ def convert(*args)
         contacts_html = @site.pages.find { |p| p.name == "contacts.html" }
         @site.process
 
-        source = @site.in_source_dir(contacts_html.path)
+        @site.in_source_dir(contacts_html.path)
```
parkr: You can just remove this if the return value isn't used.

lib/jekyll/plugin.rb
mattr-: Maybe this needs a comment? IIRC, I think I tried to do something similar awhile back.

lib/jekyll/converter.rb
Crunch09: The description of the method says:

> When an argument is specified, the prefix will be set. If no argument is specified, the current prefix will be returned.

That's why i check for `nil` (the default argument value) explicitly instead of just a `truthy` value. It seemed more accurate according to the comment to me.
I added the `!defined?` check to set the `@higlighter_prefix` in this specific case even if the argument is `nil` to remove the warning and initialize the instance variable with `nil`.

lib/jekyll/converter.rb
Crunch09: same reason as above :)

lib/jekyll/converters/markdown/kramdown_parser.rb
Crunch09: Maybe
```ruby
return @highlighter if defined?(@highlighter) && @highlighter
```

but i'm not too happy with that either.

lib/jekyll/plugin.rb
Crunch09: You're right i tried to change the code now similar to the `_suffix` / `_prefix` stuff because i think it is closest to the method description in this case as well.

Crunch09:
@parkr Thank you for the review! I tried to address all of your points and thought it would be best to create another commit. I'll squash them once everybody is okay with the changes.

lib/jekyll/converters/markdown/kramdown_parser.rb
parkr: Can we initialize `@highlighter` in `initialize`?

lib/jekyll/converter.rb
parkr: @Crunch09 Absolutely! The change in behaviour is slight, but it is a change nevertheless that could cause problems. I think we'll go with it and revert in 3.2.1 if things go sour.

parkr:
@Crunch09 Thanks! One comment above in response to your `kramdown_parser.rb` reply and this should be good to go.

Crunch09:
@parkr Hey, i added the suggested initialization in `kramdown_parser.rb` and squashed the commits

parkr:
@jekyllbot: merge +dev

PR #4542, opened 2016-02-19
Site Template: write a Gemfile which is educational to the new site

### LINKED ISSUE #4539: Add version to _config.yml and warn if a different version is used on build
If we could add a version field to _config.yml this would save debugging time when using multiple versions of Jekyll. An example _config.yml would be:

```
name: Version Test
author: george@cloudcannon.com
version: 3.0.3
```

If you ran `jekyll serve` and it ran anything other than 3.0.3 it would output:

```bash
$ jekyll serve
Running Jekyll XXX but site expected 3.0.3 see http://jekyllrb.com/docs/upgrading/2-to-3/ for potential changes
```

Running the correct version would print nothing

```bash
$ jekyll _3.0.3_ serve
...
```

This matches a feature on CloudCannon at http://docs.cloudcannon.com/building/jekyll/.

parkr:
Fixes #4539. Kind of.

- [x] Create Gemfile template
- [x] Test it

### DIFF
--- a/lib/jekyll/commands/new.rb
+++ b/lib/jekyll/commands/new.rb
@@ -35,6 +35,10 @@ module Jekyll
             File.open(File.expand_path(initialized_post_name, new_blog_path), "w") do |f|
               f.write(scaffold_post_content)
             end
+            
+            File.open(File.expand_path("Gemfile", new_blog_path), "w") do |f|
+              f.write(gemfile_contents)
+            end
           end
 
           Jekyll.logger.info "New jekyll site installed in #{new_blog_path}."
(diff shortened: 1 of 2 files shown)


lib/jekyll/commands/new.rb
```diff
@@ -59,6 +63,27 @@ def initialized_post_name
         end
 
         private
+        
+        def gemfile_contents
+          <<-RUBY
+source 'https://rubygems.org'
+  
+# Hello! This is where you manage which Jekyll version is used to run.
+# When you wwant to use a different version, change it below, save the
```
benbalter: `wwant` -> `want`

benbalter:
I'm a big fan of this. Was kinda surprised the default site didn't come with a Gemfile. Think it'd go a long way to help ease non-Ruby users into the Jekyll world, and if a user doesn't use Bundler, no harm done.

parkr:
@jekyll/stability, look good to you?

parkr:
@jekyllbot: merge +minor

PR #4544, opened 2016-02-19
Allow collections to have documents that have no file extension

parkr:


parkr:
Not sure if we really need this, but :shrug:

/cc @benbalter

parkr:
This will be a 3.2 feature.

Issue #4534, opened 2016-02-17
Version 3 creating wrong permalinks for collection items


probins:
I raised this as a question at http://stackoverflow.com/questions/35382768/converting-collection-permalinks-jekyll-2-3 (and also contacted Github support), as I wasn't sure whether this was a bug or whether I had something misconfigured or there was some other problem. As the conclusion was that this 'seems to be a bug', I'll raise it here. This is a regression, as what was working fine in v2 no longer works since Github Pages was updated to v3.

The original site is very large with literally 000s of GeoJSON files, so I've created a small test repo to illustrate the issues https://github.com/pilgrimdb/testdb. This contains 4 files in the _features collection. The feature layout includes the geometry for each section (in `_includes`), adds the properties (in `_data/features`), and outputs a GeoJSON feature.

I think there are actually 3 separate issues:
1. the items in _features are named `*.f` and the permalink in _config specifies /features/:path.geojson. In v2, this was output as `/features/*.geojson`, which is what I want. But in v3, it now puts these in `/features/*.geojson.f`, which is unusable, as mapping programs need them as `*.geojson`.
2. I had 2 similar files which were 'pages' and not in any collection (these are not in testdb). These were originally called `*.geo` and also output as `*.geojson` in v2. In v3, these were also output as `*.geojson.geo`, but I found that by renaming them without an extension, v3 would output them ok as `*.geojson`. However, if I try renaming the files in _features so they no longer have an extension, then v3 ignores them: no features are output at all.
3. the docs aren't very helpful here. They talk about 'path' without specifying what that means here. I would normally think of path as including the extension (as in url path), but that doesn't seem to be the case here. If it's `:path.geojson` it should surely be `*.f.geojson`; `*.geojson.f` makes no sense. OTOH, if path doesn't include extension, how does it differ from 'name', and what's 'output_ext' mean?

parkr:
Hi, @probins! Which version of `github-pages` are you using?

> I think there are actually 3 separate issues:

In the future, each GitHub Issue should correlate to one problem you're having with Jekyll. Makes it easier to split up the work and keep track of each.

> 1. the items in _features are named *.f and the permalink in _config specifies /features/:path.geojson. In v2, this was output as /features/*.geojson, which is what I want. But in v3, it now puts these in /features/*.geojson.f, which is unusable, as mapping programs need them as *.geojson.

[...]

PR #4545, opened 2016-02-19
Allow collections to have documents that have no file extension

parkr:
Duplicates #4544 for `master`. Fixes one item from #4534.

/cc @jekyll/core


parkr:
@jekyllbot: merge +minor

PR #4546, opened 2016-02-19
Permalinks which end in a slash should always output HTML

parkr:
Duplicates #4493 for 3.1.1. Ports a fix we did for 3.0.3 to `master`.

/cc @jekyll/core


parkr:
@jekyllbot: merge +bug

PR #4551, opened 2016-02-20
Prevent permalink_ext return extension with colon

vwochnik:
Consider a permalink like this: `/path/to/file.notmyextension:output_ext`. In the case of File.extname, `.notmyextension:output_ext` is returned which causes a bad output file name.

I have replaced `File.extname` with a regular expression. Normally, `File.extname` is accurate but here we are not dealing with file names but with permalinks containing `:`s.

### DIFF
--- a/lib/jekyll/renderer.rb
+++ b/lib/jekyll/renderer.rb
@@ -169,8 +169,8 @@ module Jekyll
 
     def permalink_ext
       if document.permalink && !document.permalink.end_with?("/")
-        permalink_ext = File.extname(document.permalink)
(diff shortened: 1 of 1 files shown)


parkr:
I see. Would you please add a test for your case above?

vwochnik:
Unfortunately, I am not very familiar with the test system used by Jekyll and my time management is very tight.

parkr:
If a user has a permalink of `.html:ifonly`, then the Principle of Least Magic says we should do their bidding and output a file ending in `.html:ifonly`.

/cc @benbalter

PR #4553, opened 2016-02-20
Add jekyll-seo-tag, jekyll-avatar, and jekyll-sitemap to the site

benbalter:
This pull request adds three plugins to the `jekyllrb.com` site with a few targeted improvements:

### Jekyll Avatar

This plugin improves the contributor avatars used in the news feed, specifically:

* Avatars will load directly from the GitHub avatar server, avoiding the need for a 301 redirect from `/username.png`.
* Avatars load in parallel across 3 different avatar servers, using the same hashing function used within GitHub, improving the likelihood that they're already cached on the visitor's machine
* Avatars upgrade to Retina-size images, where supported

To use Jekyll Avatar, there's a small template change, but no change in behavior/use.

### Jekyll Sitemap

This adds a sitemaps.org compliant sitemap at `/sitemap.xml`. That's it. No configuration or fuss. It'll help Google and other search engines better index some of the deeper (and newer) docs and posts.

### Jekyll SEO Tag

This plugin makes a handful of improvements not only in how the site is index by search engines, but also how pages look like when they're shared on social media. Specifically:

* Page titles are appended with the site title, meaning the title actually has Jekyll in it, and not just, e.g., "Assets"
* Adds canonical and next and previous (for pages and docs) URLs to the site head 
* Adds JSON-LD metadata for richer indexing, including the logo
* Adds Open Graph metadata for sharing on LInkedIn, Facebook, Twitter, etc.
* Adds Twitter summary card metadata, including author information

Again, no change in behavior/use. We can begin to add human-curated descriptions to posts, pages, and docs if we wanted, but that's about it.

Thoughts?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -81,4 +81,7 @@ group :site do
     gem "html-proofer", "~> 2.0"
   end
   gem "jemoji"
+  gem "jekyll-sitemap"
+  gem "jekyll-seo-tag"
+  gem "jekyll-avatar"
 end
--- a/site/_config.yml
+++ b/site/_config.yml
@@ -19,7 +19,15 @@ name: Jekyll • Simple, blog-aware, static sites
 description: Transform your plain text into static websites and blogs
 url: http://jekyllrb.com
 
+twitter:
+  username: jekyllrb
+
+logo: img/logo-2x.png
+
 gems:
   - jekyll-feed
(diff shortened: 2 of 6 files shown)


parkr:
LGTM.

parkr:
@jekyllbot: merge +site

PR #4554, opened 2016-02-20
Add Google site verification meta

benbalter:
To allow us to register the site with Google webmaster tools to register the sitemap added via https://github.com/jekyll/jekyll/pull/4553.

### DIFF
--- a/site/_config.yml
+++ b/site/_config.yml
@@ -31,3 +31,5 @@ gems:
   - jekyll-sitemap
   - jekyll-seo-tag
   - jekyll-avatar
+
+google_site_verification: onQcXpAvtHBrUI5LlroHNE_FP0b2qvFyPq7VZw36iEY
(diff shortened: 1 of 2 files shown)


site/_config.yml
```diff
@@ -31,3 +31,5 @@ gems:
   - jekyll-sitemap
   - jekyll-seo-tag
   - jekyll-avatar
+
+google_site_verification: onQcXpAvtHBrUI5LlroHNE_FP0b2qvFyPq7VZw36iEY
```
parkr: Perhaps this should go above with `google_analytics_id`?

parkr:
One thought, otherwise :+1:/

site/_config.yml
benbalter: Good call. b6a0b9e

envygeeks:
:+1:

PR #4555, opened 2016-02-21
Add array support to `where` filter

### LINKED ISSUE #4385: Feature request: add where_contains filter

timwis:
In addition to `where` supporting basic string matches (ie. `category == "foo"`) this allows the `where` filter to work on array values, ie. `category contains "foo"`. So whether the page has
```yaml
---
category: Foo
---
```
or
```yaml
---
category:
  - Foo
---
```
The `where` filter will still work (`{{ site.pages | where:"category", "foo" }}`)

Note that this should not have an effect on any existing (string-match) use of `where` (non-breaking change)

Resolves #4385 (feature request)

### DIFF
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -222,7 +222,7 @@ module Jekyll
     def where(input, property, value)
       return input unless input.is_a?(Enumerable)
       input = input.values if input.is_a?(Hash)
-      input.select { |object| item_property(object, property).to_s == value.to_s }
+      input.select { |object| item_property(object, property).to_s == value.to_s or item_property(object, property).include? value.to_s }
     end
 
     # Sort an array of objects
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -222,7 +222,7 @@ module Jekyll
     def where(input, property, value)
       return input unless input.is_a?(Enumerable)
       input = input.values if input.is_a?(Hash)
-      input.select { |object| item_property(object, property).to_s == value.to_s or item_property(object, property).include? value.to_s }
+      input.select { |object| Array(item_property(object, property)).map(&:to_s).include?(value.to_s) }
     end
 
     # Sort an array of objects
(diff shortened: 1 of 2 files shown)


parkr:
This changes the behaviour of this filter in possibly nefarious ways because it'd be grabbing things that previously it didn't. @benbalter, what say you? If we can't come up with an example where it'd be a problem, then I'm :+1:.

benbalter:
I can't think of an example where this would break existing behavior (also :wave: @timwis).

timwis:
Hey @benbalter! Thanks for the review guys.

timwis:
Okay, figured out why the tests were failing and rewrote the PR to account for all property types. I also added 2 tests. I'm new to ruby so please let me know if I've made any convention mistakes or anything.

parkr:
What happens if you have `category: bear` and you search for `where: "category", "ear"`? From my reading of the code, you'd get back that post with `category: bear` even though you weren't really interested in that.

timwis:
Hey @parkr, perhaps I'm misunderstanding, but I don't think that's the case. I've added a test for it to demonstrate. Let me know if I'm off.

parkr:
This LGTM. Thanks!

@jekyllbot: merge +minor

parkr:
@jekyllbot: merge +minor

parkr:
@jekyllbot: merge +minor

parkr:
@jekyllbot: merge +minor

Issue #4507, opened 2016-02-10
Verbose build does not explain why future posts are skipped


zandaleph:
I'm a new Jekyll user, just getting started with my first jekyll blog.  I attempted to create a future-dated post about my experiences with Jekyll, and then was very confused when it was not published with either `jekyll serve` (which I was running in the background) or `jekyll build`.  A bit of poking around led me to try `jekyll build --verbose`, which generated the output below.  It was only a few minutes later that I realized what was probably happening, moved the post date back a couple days, and then tried again.  It would have been very useful for the background server and the verbose build to emit a message explaining why the file was skipped.

> $ jekyll build --verbose
  Logging at level: debug
Configuration file: /Users/zspencer/Code/testblog/_config.yml
         Requiring: kramdown
            Source: /Users/zspencer/Code/testblog
       Destination: /Users/zspencer/Code/testblog/_site
 Incremental build: disabled. Enable with --incremental
      Generating...
           Reading: _posts/2016-02-10-welcome-to-jekyll.markdown
           Reading: _posts/2016-02-11-using-jekyll-as-a-layout-engine.markdown
         Rendering: _posts/2016-02-10-welcome-to-jekyll.markdown
  Pre-Render Hooks: _posts/2016-02-10-welcome-to-jekyll.markdown
  Rendering Liquid: _posts/2016-02-10-welcome-to-jekyll.markdown
         Requiring: rouge
  Rendering Markup: _posts/2016-02-10-welcome-to-jekyll.markdown
         Requiring: kramdown
  Rendering Layout: _posts/2016-02-10-welcome-to-jekyll.markdown
         Rendering: _posts/2016-02-10-welcome-to-jekyll.markdown
  Pre-Render Hooks: _posts/2016-02-10-welcome-to-jekyll.markdown
  Rendering Liquid: _posts/2016-02-10-welcome-to-jekyll.markdown
  Rendering Markup: _posts/2016-02-10-welcome-to-jekyll.markdown

hubgit:
Same here - it took a while to work out that I needed to add `future: true` to `_config.yml` for future-dated posts to show up. Perhaps `--serve` could output a warning message like "3 future-dated posts were skipped", and a note about adding `--future` to render them, and needing the setting in `_config.yml` for GitHub Pages?

envygeeks:
Shouldn't it be obvious that today `11/02/2016` is not `14/02/2016` and anything I write for the 14th won't be published?

hubgit:
[...]

PR #4558, opened 2016-02-21
Adding a debug log statment for skipped future documents.

zandaleph:
For https://github.com/jekyll/jekyll/issues/4507

All tests pass, but didn't add any new tests for the log line.  There's only one test that looks at log output at this point, and I didn't know enough to write a good test of this change.

One thing I wasn't sure was okay - this change places a requirement on all things to be published to have a `#relative_path` method.  I'm not sure how to validate if this is a good assumption or not.  Guidance here would be appreciated.

Sample output of `jekyll build --verbose`:

```text
   Logging at level: debug
 Configuration file: /Users/zspencer/Code/testblog/_config.yml
          Requiring: kramdown
             Source: /Users/zspencer/Code/testblog
        Destination: /Users/zspencer/Code/testblog/_site
       Generating...
            Reading: _posts/2016-02-10-welcome-to-jekyll.markdown
            Reading: _posts/2016-02-29-using-jekyll-as-a-layout-engine.markdown
           Skipping: _posts/2016-02-29-using-jekyll-as-a-layout-engine.markdown because it is in the future.
          Rendering: _posts/2016-02-10-welcome-to-jekyll.markdown
   Pre-Render Hooks: _posts/2016-02-10-welcome-to-jekyll.markdown
   Rendering Liquid: _posts/2016-02-10-welcome-to-jekyll.markdown
```

### DIFF
--- a/lib/jekyll/readers/post_reader.rb
+++ b/lib/jekyll/readers/post_reader.rb
@@ -35,7 +35,11 @@ module Jekyll
       read_content(dir, magic_dir, matcher).tap do |docs|
         docs.each(&:read)
       end.select do |doc|
-        site.publisher.publish?(doc)
+        site.publisher.publish?(doc).tap do |will_publish|
+          if !will_publish && site.publisher.hidden_in_the_future?(doc)
+            Jekyll.logger.debug "Skipping:", "#{doc.relative_path} has a future date"
+          end
+        end
       end
     end
 
--- a/lib/jekyll/publisher.rb
+++ b/lib/jekyll/publisher.rb
@@ -8,14 +8,14 @@ module Jekyll
       can_be_published?(thing) && !hidden_in_the_future?(thing)
     end
 
+    def hidden_in_the_future?(thing)
+      thing.respond_to?(:date) && !@site.future && thing.date.to_i > @site.time.to_i
+    end
+
     private
 
     def can_be_published?(thing)
       thing.data.fetch('published', true) || @site.unpublished
-    end
-
-    def hidden_in_the_future?(thing)
-      thing.respond_to?(:date) && !@site.future && thing.date.to_i > @site.time.to_i
     end
   end
 end


envygeeks:
I like the premise, I don't like the spot it's place in though, boolean methods shouldn't be logging like that because they can be called dozens of times to validate something which can lead to plenty of duplicate logging.  Is there no better place to put it?

/cc @jekyll/core -- @parkr maybe you can point out a better place?

zandaleph:
I'm not a huge fan of the location, but as the code is currently factored there's nowhere else this statement can go without either duplicating the logic in `Publisher#hidden_in_the_future?` or refactoring (which I am not familiar enough with the jekyll project to do without guidance).  The method is only called as part of the `Publisher#publish?` method, and since that method considers factors other than the post being in the future, there's no way to tell from the result of that method why the thing should not be published.

I think it is a reasonable location because while it is possible that it could be called dozens of times on the same object, currenly `Publisher#publish?` is only used in an `Enumerable#select` context, which means that once an object is found to be unpublishable, it is not considered again.  I think it is reasonable to expect that this will continue to be the only way `Publisher#publish?` will be used until a refactoring is necessary for other reasons.

envygeeks:
Well it's not a matter of whether we use it only once, it's a matter of who else uses it more than once since that's a public API.  If there is no better place to put it and it's really only used once then there is certainly room for refactoring but I'll digress on when that happens to other @jekyll/core.

zandaleph:
Agh, of course! I hadn't considered other consumers of Publisher outside of the core Jekyll code.  I feel a bit silly now.

Alright, since this proposed solution is even worse than I had realized, I'd argue that the only way to know why a document wasn't published is to have an alternative method on `Publisher` (say, `#check_publish`) which returns a hash of `{:should_publish => boolean, :reason => string}`.  Then `#publish?` can just delegate to `#check_publish` and return the first item, and integrators which need more visibility into why an item couldn't be published can call `#check_publish` instead.  Does this sound like a reasonable way forward?  If so I think I can hack something together like this in my spare time over the next couple of days.

parkr:
@zandaleph I don't see why that is necessary. Why not add it [where the publisher.publish? call is actually made?](https://github.com/jekyll/jekyll/blob/4048b8e2a735565de2f515f55036c859415a06ae/lib/jekyll/readers/post_reader.rb#L38)

zandaleph:
@parkr Okay, I've done the change as you suggested - only doing the check where the `Publisher.publish?` call is actually made.  There are two things I don't like about this change:

1. I've had to expose another public method on `Publisher`.
2. I've had to duplicate logic from `publish?` in order to disambiguate why the post was skipped.

If this is still the preferred approach, there it is.

I've left alone the logic from `PageReader` and `Collection` where `publish?` is called, the former because pages aren't dated (I assume), and the latter because it already has logging on skipping, albeit without explaining why.

zandaleph:
@jekyll/core Any further work that needs done on this change? I'm still happy to prototype a more comprehensive solution for signaling why `publish?` decided to skip something if needed.

zandaleph:
@parkr @envygeeks Not sure what the etiquette is on bumping pull requests, so this is the last time I'll update this issue unless I get some feedback.  I believe this change captures what I was looking for in #4507 .  Please let me know if there's anything I can do.

lib/jekyll/readers/post_reader.rb
```diff
@@ -35,7 +35,11 @@ def read_publishable(dir, magic_dir, matcher)
       read_content(dir, magic_dir, matcher).tap do |docs|
         docs.each(&:read)
       end.select do |doc|
-        site.publisher.publish?(doc)
+        publish = site.publisher.publish?(doc)
+        if !publish && site.publisher.hidden_in_the_future?(doc)
+          Jekyll.logger.debug "Skipping:", "#{doc.relative_path} because it is in the future."
+        end
+        publish
```
parkr: I like it! Let's change up the style though, using `#tap`:

```ruby
site.publisher.publish?(doc).tap do |will_publish|
  if !will_publish && site.publisher.hidden_in_the_future?(doc)
    Jekyll.logger.debug "Skipping:", "#{doc.relative_path} has a future date"
  end
end
```

zandaleph: I had no idea you could `#tap` arbitrary values - I've only ever used it on `Enumerable`s before.  I like it, and I've updated the change to use your suggestion.  Thanks!

envygeeks:
I'm :+1: now, seems like this also addresses your concerns too @parkr?

parkr:
> I'm :+1: now, seems like this also addresses your concerns too @parkr?

Indeed, it does!

Thanks for your patience, @zandaleph! And thanks for the pull request. :smile: 

@jekyllbot: merge +minor

zandaleph:
My pleasure, thank you for the excellent feedback and guidance.

PR #4560, opened 2016-02-22
Set kramdown gem version from 1.3 to 1.9 in jekyll.gemspec

SuriyaaKudoIsc:
@parkr @mojombo 

New kramdown version: https://rubygems.org/gems/kramdown!

### DIFF
--- a/jekyll.gemspec
+++ b/jekyll.gemspec
@@ -29,7 +29,7 @@ Gem::Specification.new do |s|
   s.extra_rdoc_files = %w[README.markdown LICENSE]
 
   s.add_runtime_dependency('liquid',    '~> 3.0')
-  s.add_runtime_dependency('kramdown',  '~> 1.3')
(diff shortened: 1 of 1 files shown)


pathawks:
Why do you need to specifically disallow older versions of Kramdown?

SuriyaaKudoIsc:
I would like to set Kramdown to the newest version.

pathawks:
> I would like to set Kramdown to the newest version.

Right. By default the newest version of Kramdown will be installed. But some users may want to use an older version of Kramdown.

Your change does nothing to allow the latest version of Kramdown, it only blocks users from using a version older than 1.9.

parkr:
I think we will remain more flexible about this for now. Thanks, though! You can update your version with `gem update kramdown` or `bundle update kramdown`. :smile

SuriyaaKudoIsc:
:+1:

PR #4562, opened 2016-02-22
Set url in _config.yml to https://

SuriyaaKudoIsc:
@parkr @mojombo 

HTTPS is better! :smile:


parkr:
Everything seems to load fine for me. We had a font issue previously but now :+1:

PR #4570, opened 2016-02-23
Explain how site variables work (Fixes #4556)

### LINKED ISSUE #4556: Default _config.yml contains site variables that aren't explained in the docs (github_username and twitter_username)
Jekyll version: 3.1.2
OS: Ubuntu 14.04.1 LTS

A default install of a Jekyll site (via `jekyll new whatever`) produces a site with a `_config.yml` containing:

```
twitter_username: jekyllrb
github_username: jekyllrb
```

Naturally I change these to `detly` and then wonder... what other links are available? How do these variables work?

Since I'm editing this in the default `_config.yml`, I go to the [Jekyll documentation for configuration](http://jekyllrb.com/docs/configuration/). There is no mention of these variables, nor is there any mention of general branded 3rd party site links at all.

I think that if Jekyll's out-of-the-box config file is going to contain something, it should probably be mentioned in the basic documentation. I incorrectly guessed that this was some sort of plugin, and spent ages searching for what made them work, before realising that they were a kind of global variable and worked via the `_includes/icon-github.html` template.

I don't see this kind of mechanism mentioned in the config docs. It goes into YAML front matter defaults, but it doesn't say you can just straight-up type random variables in here and have them be available to every template via the `site` namespace. (I <kbd>ctrl</kbd>+<kbd>f</kbd>ed for `site.` on that page and didn't see anything; sorry if it actually is there and I skipped over it.)

> The configuration file also allows you to set site-wide variables (eg. `my_variable`) that can be accessed as `site.my_variable`. The default `_config.yml` already contains a couple of these: `github_username` and `twitter_username`. You'll notice in the default site layout, these produce branded links to the corresponding Twitter and Github account. The template `_include/footer.html` and `_include/icon-github.html` show you how to use these: just reference the `site.github_username` to determine what to include and how to construct the appropriate URLs.

chrisfinazzo:


lib/site_template/_config.yml
```diff
@@ -17,5 +17,12 @@ url: "http://yourdomain.com" # the base hostname & protocol for your site
 twitter_username: jekyllrb
 github_username:  jekyll
 
+# The configuration file also allows you to set site-wide variables that can be
+# accessed as `site.my_variable`. You'll notice in the default site layout,
+# these produce branded links to the corresponding Twitter and Github account.
```
parkr: Please capitalize the `H` in `GitHub` :smile:

lib/site_template/_config.yml
```diff
@@ -17,5 +17,12 @@ url: "http://yourdomain.com" # the base hostname & protocol for your site
...
+# The template `_include/footer.html` and `_include/icon-github.html` show you
+# how to use these: just reference the site.github_username to determine what
```
parkr: Not sure what you mean by this and the following line. Would you please explain?

parkr:
Thanks for the PR, @chrisfinazzo. I think we can go about this more simply.

lib/site_template/_config.yml
chrisfinazzo: I'll send a small patch for this, it was included in the original branch.

PR #4580, opened 2016-02-25
Add lots of timing metrics around the render process.

parkr:
I was just helping a user get some timing data from his site and noticed that we don't really offer timing data. Silly us.

This PR is not necesarily the final iteration but is a minimal implementation of producing timing data for each step in the rendering process.

I often wonder what's holding up the regeneration of my site. I think this should eventually be merged into `--profile` (which is currently Liquid-only). For now, it's simply useful as just `--verbose` output.

What do you think? @jekyll/core @benbalter

### DIFF
--- a/lib/jekyll/external.rb
+++ b/lib/jekyll/external.rb
@@ -1,6 +1,15 @@
 module Jekyll
   module External
     class << self
+      def required_gems
+        @required_gems ||= Set.new
+      end
+
+      def require_gem(name)
+        require name
+        required_gems << name
+      end
+
       #
       # Gems that, if installed, should be loaded.
       # Usually contain subcommands.
@@ -19,8 +28,10 @@ module Jekyll
       #
       def require_if_present(names, &block)
         Array(names).each do |name|
+          next if required_gems.include? name
           begin
             require name
+            required_gems << name
           rescue LoadError
             Jekyll.logger.debug "Couldn't load #{name}. Skipping."
             block.call(name) if block
@@ -38,9 +49,10 @@ module Jekyll
       #
       def require_with_graceful_fail(names)
         Array(names).each do |name|
+          next if required_gems.include? name
           begin
             Jekyll.logger.debug "Requiring:", "#{name}"
-            require name
+            require_gem name
           rescue LoadError => e
             Jekyll.logger.error "Dependency Error:", <<-MSG
 Yikes! It looks like you don't have #{name} or one of its dependencies installed.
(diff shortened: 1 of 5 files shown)


mattr-:
What if we re-arrange where the hooks fire so that we could use the hook system to gather the timing data or is that not-feasible?

mattr-:
definitely :+1: for timing data

benbalter:
Good call. How would the output differ if we just timed between hooks (potentially adding missing ones)?

parkr:
> What if we rearrange where the hooks fire so that we could use the hook system to gather the timing data or is that not-feasible?

What do you mean? Do you mean we render via a hook and time all hook calls?

mattr-:
> >What if we rearrange where the hooks fire so that we could use the hook system to gather the timing data or is that not-feasible?

> What do you mean? Do you mean we render via a hook and time all hook calls?

Sorry, I wasn't clear. I was wondering if we could write something that starts timing with the first call to a hook, and provides timing info based on time between hooks. IOW, rather than adding timing to Jekyll core by mixing the concerns, write hooks that would be used to gather and report timing info.

parkr:
> write hooks that would be used to gather and report timing info

@mattr- We might have to introduce a few more hooks... Think that would be worthwhile?

mattr-:
> We might have to introduce a few more hooks... Think that would be worthwhile?

I think so. I looked at the existing hooks awhile back so see what I could hook into and came away feeling like there were more that could be added to make things go better. I'm trying to get back to my benchmarking project but it's slow going as other more important and urgent things keep getting in the way. :disappointed:

envygeeks:
@mattr- @parkr yeah when they were implemented just the basics were done as there were places where things needed to change to implement more hooks (like ocnfiguration was convoluted so we couldn't allow users to hook into that easily.) And then nobody ever suggested more hook points. I do think we have a ticket with suggestions though, no?

parkr:
So we'd start adding in `pre` and `post` hooks for each action we'd like to time, right? Then keep some sort of global hash in a singleton instance or something which holds the start time and can be used to determine the duration later in the `post` hook?

parkr:
That would require a number of hooks if we wanted very granular timing info:

```text
         Rendering: collections.html
  Pre-Render Hooks: collections.html in 0.00000600s
  Rendering Liquid: collections.html in 0.00083600s
  Rendering Markup: collections.html through Jekyll::Converters::Identity in 0.00000100s
  Rendering Markup: collections.html in 0.00001800s
  Rendering Layout: collections.html in 0.00002100s
          Rendered: collections.html in 0.00107600s
```

What is useful to time?

parkr:
This output isn't that helpful for even moderately sized sites – 7+ lines of output _per file_? 682 lines of output total on the code for `byparker.com`. Maybe we should just build this into `--profile`.

parkr:
I'm going to go about this completely differently.

PR #4585, opened 2016-02-26
Correct content in collections.feature

SuriyaaKudoIsc:
@parkr @mojombo


parkr:
This causes the test to fail. I see that it's more grammatical, but you'd then have to change the test input as well as the check. I think it's fine to leave as-is, as grammar isn't a requirement for these tests. :smile: Thanks for the PR!

SuriyaaKudoIsc:
:+1:

PR #4586, opened 2016-02-26
Add comment header in Rakefile

SuriyaaKudoIsc:
@parkr @mojombo


parkr:
Hey! Thanks for the PR. This header is not really necessary and only explains the `require` calls. I don't think we need a comment for these lines. One for the file might make sense, but it should explain the entire file. If you come up with a short and simple way of describing the whole file, feel free to send another PR! :smile:

SuriyaaKudoIsc:
:+1:

PR #4588, opened 2016-02-26
Add footer comment in default.html

SuriyaaKudoIsc:
@parkr @mojombo

### DIFF
--- a/site/_layouts/default.html
+++ b/site/_layouts/default.html
@@ -11,3 +11,4 @@
 
 </body>
 </html>
+<!-- Build the <> with <3 by the Jekyll Community ~ Generated {{ site.time }} -->


parkr:
@SuriyaaKudoIsc Please do not ping `@mojombo` – he is no longer actively working on this project. Please use the description to write a reason for your change. We have a process for triaging issues, so don't worry about us not responding.

Why do you think we should add this comment?

SuriyaaKudoIsc:
It's better to use the generated site time tag in a comment because some developers want to now how fast the site was build by Jekyll.

I use it also on my website https://bit.ly/Suriyaa, too!

envygeeks:
@SuriyaaKudoIsc then they should look at the logs, the generated sites source is not the place to look for that information, it's out of the way and unintuitive.

SuriyaaKudoIsc:
Ok.

PR #4589, opened 2016-02-26
Add Google search query to /docs/help/

SuriyaaKudoIsc:
@parkr @mojombo


parkr:
I like the www but am a little hesitant about prefilling the query. Hopefully folks know to add their query to the end. For now, it's fine. :+1: 

@jekyllbot: merge +site

Issue #4582, opened 2016-02-26
Post timezones overwritten


tonyg:
Jekyll 3.0.1 on Debian (3.0.1+dfsg-1).

After upgrading to Jekyll 3, my blog has been generating one particular post at /2004/09/14/stoj/index.html, but giving it the permalink and URL of /2005/09/15/stoj/.

The issue is that the `data['date']` is for the 15th, at two minutes past midnight in the BST (+01:00) timezone.

At some point during processing, Jekyll is destructively mutating the date object so that its timezone is the rendering machine's local timezone, -04:00. This happens after the permalink is constructed, but before the post is written to file—hence the mismatch between the URL and the filename.

There is a call to `.localtime` in filters.rb which seems to be at fault. If I modify `Jekyll::Filters::time`, replacing `input` with `input.clone` for the `Time` branch of the case, the problem goes away and the timezone appears to be preserved. For now, I am monkey-patching using a plugin (pasted below).

Perhaps this is related to issue #4096?

Monkey-patch to avoid using `.localtime` destructively on `data['date']`:

```ruby
module Jekyll
  module Filters
    private
    def time_with_clone(input)
      case input
      when Time
        time_without_clone(input.clone)
      else
        time_without_clone(input)
      end
    end
    alias_method :time_without_clone, :time
    alias_method :time, :time_with_clone
  end
end
```

tonyg:
(To be clear, it seems to me that this is the line at fault: https://github.com/jekyll/jekyll/blob/acb2263f51af818ed74734b8843ddb0a18ac4532/lib/jekyll/filters.rb#L318 )

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least
three months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If you can still reproduce this error on the <code>3.1-stable</code> or <code>master</code> branch,
please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider building it first as a plugin. Jekyll 3 introduced
[hooks](http://jekyllrb.com/docs/plugins/#hooks) which provide convenient access points throughout
the Jekyll build pipeline whereby most needs can be fulfilled. If this is something that cannot be
built as a plugin, then please provide more information about why in order to keep this issue open.

Thank you for all your contributions.

PR #4590, opened 2016-02-26
Filters#time: clone an input Time so as to be non-destructive

tonyg:


parkr:
Fixes #4582.

@willnorris, does this happen to also fix your issue in the Jekyll 3 upgrade around `#localtime` changing when your links are generated?

parkr:
@tonyg Thanks so much! :smile: 

@envygeeks tried `.dup` and it broke the test. We have to full-out clone to make this work.

@jekyllbot: merge +bug

parkr:
@willnorris Ah, bummer. Thanks for letting me know. Let me know if you find anything!

Issue #4510, opened 2016-02-11
Gem-based themes


benbalter:
Similar to what we did with Gem-based plugins, we should introduce the concept of Gem-based themes, making it easier for users to style their content, and hopefully creating a more robust theme ecosystem.

### How themes work today

Today, while there are lots of great Jekyll themes out there, the installation process generally involves either forking a repo or copying files into the destination site. The problem with this approach, beyond ease of use and the ability to quickly switch themes, is that there's no good way to downstream subsequent updates (not to mention, from a DRY perspective, the theme's CSS and templates now exist in hundreds or thousands of repos, each with their own versions and edits).

### How Gem-based themes would work for users

Find a theme you like? Simply add `theme: awesome-theme` in your site's `_config.yml` file. That'll instruct Jekyll to load the Gem-based theme, just like it would any plugin. The theme would be able to register layouts, and its static assets (css, javascript, images, etc.) would be automatically included in the built site. Heck, maybe there's even an `{{ asset_include_tag }}`, or similar, to include the theme's assets in your HTML. Lets say the theme has a `main` layout. If you don't like it, simply create a layout in your standard `_layouts` folder with the same name. At render time Jekyll will first look to the user's layouts and includes, and if one doesn't exist, it'll then look to the theme's. As an added bonus, if you're just using the theme defaults, we've allowed you to better separate content from presentation, as your repo is now almost entirely your own content, with less Jekyll noise getting in your way.

### How Gem-based themes would work for theme creators

Themes are just Gems with a defined folder structure, not unlike the Jekyll site itself. The only required file would be the Gemspec itself. Beyond that, you can introduce as many layouts you'd like, in a `_layouts` folder, the same with `_includes`. For assets, the `_sass` folder can automatically be included in sass's include path (meaning the user could just do `@import 'awesome-theme'` to include your css in there own). We can also define some sort of `assets` (`_assets`?) folder, which'll be automatically included as a whole, in the built site. Heck, maybe themes can even be dependencies of other themes (e.g., a vanilla Bootstrap theme might be a dependency of Hyde). We'll probably also want to include a screenshot in a predictable path.

### A note on security

[...]

PR #4595, opened 2016-02-26
Gem-based themes

benbalter:
This is a **very** early attempt to try to figure out what would be involved to support https://github.com/jekyll/jekyll/issues/4510.

Right now, adding `theme: foo` to your config file, will init the `foo` theme, and the theme can find itself on the filesystem, but that's about it.


spudowiar:
Ooh, just noticed this when I'd already started working on this! My method turned `layouts_dir` into an array or a string, the array case obviously using multiple layout directories. Rinse and repeat for other types of directory.

benbalter:
@spudowiar that's awesome. Ignoring the change to `lib/jekyll/readers/layout_reader.rb` (which didn't work), it sounds like we could combine our efforts, by doing something like `layouts_dir << site.theme.layouts_path` (and rinse and repeat for other directories)?

spudowiar:
That's what I'm hoping. Patch should be complete soon. AFAICT the only other directories are `includes_dir` and `sass_dir` in `jekyll-sass-converter`

benbalter:
Pushed up passing tests. Believe this should solidly create a `Theme` class, that can safely determine its own paths. Along with @spudowiar work over in https://github.com/jekyll/jekyll/pull/4610 to support multiple paths, it looks like the basic implementation of https://github.com/jekyll/jekyll/issues/4510 is coming together.

benbalter:
Pushed up two more rounds of changes:

1. Added documentation for where I'm hoping things will land.

2. Moved theme assets from `/_assets` to `/assets`. It makes the code *slightly* more complex, but would allow theme developers to preview their theme by simply using `jekyll build` and `jekyll serve` within the theme folder, without having to do any `/_assets` -> `/assets` rewriting magic.

envygeeks:
@benbalter I don't see why that is necessary, do Jekyll collections not work with static files?

envygeeks:
@benbalter @parkr Actually, that might be a good idea, create a default collection and have it output, instead of doing all that, then Jekyll-Assets can come in and remove that collection if the user chooses to use us and process independent of that.  This makes it so that asset authors can get the benefit of all worlds no?

benbalter:
>  This makes it so that asset authors can get the benefit of all worlds no?

@envygeeks I'm not sure I follow. How would you collections to implement what's described in https://github.com/jekyll/jekyll/issues/4510? The idea would be to move presentation assets out of the repository, and someplace where they can be commonly distributed, versioned, and benefit from upstream changes. 

The motivation here is less so "I want a better way to organize the assets that are specific to my site" and more so "I want a better way to use an open source theme somebody else created".

History.markdown
```diff
@@ -1,3 +1,5 @@
+# History
+
```
parkr: Please remove :smile:

lib/jekyll/theme.rb
```diff
@@ -0,0 +1,48 @@
+module Jekyll
+  class Theme
+    attr_reader :name
+
+    def initialize(name)
+      @name = name.downcase.strip
+      raise Jekyll::Errors::MissingDependencyException unless gemspec
```
parkr: Would you please add a message to this raise as a second argument?

```ruby
unless gemspec
  raise Jekyll::Errors::MissingDependencyException, "#{name} could not be found on your system."
end
```

Also, what motivated your choice to place this logic here? Just an early fail?

lib/jekyll/theme.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    def root
+      @root ||= gemspec.full_gem_path
+    end
+
+    def version
+      gemspec.version
```
parkr: Please use a `def_delegator` (via Forwardable) for this.

site/_config.yml
```diff
@@ -2,6 +2,9 @@ markdown: kramdown
 highlighter: pygments
 permalink: /news/:year/:month/:day/:title/
 excerpt_separator: ""
+kramdown:
+  input: GFM
+  hard_wrap: false
```
parkr: I believe this has deleterious effects on the markup of the site. Please submit this in another pull request.

parkr:
This is great work, @benbalter! Thank you for this pull request. Some concerns above.

How do you plan on integrating the theme into the render process?

benbalter:
@parkr believe I addressed your feedback.

> How do you plan on integrating the theme into the render process?

Once @spudowiar's https://github.com/jekyll/jekyll/pull/4610 lands, includes, layouts and sass should be as simple as:

```ruby
site.includes_dir << theme.includes_path
site.layouts_dir" << theme.layouts_path
site.sass.source << theme.sass_path
```

I'm waiting to see what happens in https://github.com/jekyll/jekyll/pull/4610 to implement that.

The last remaining step would be assets. Any thoughts on the best way to do that?

site/_docs/history.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: History
+title:
```
parkr: Ruh roh

site/_docs/themes.md
```diff
@@ -0,0 +1,99 @@
...
+## Overriding theme defaults
+
+Jekyll themes set default layouts, includes, stylesheets and static assets, that can be overridden by your site's content. For example, if your selected theme has a `page` layout, you can override the theme's layout by creating your own `page` layout in the `_layouts` folder (e.g., `_layouts/page.html`).
+
+Jekyll will look first to your site's content, before looking to the theme's defaults, for any requested file in the following folders:
+
+* `/_layouts`
+* `/_includes`
+* `/_sass`
+* `/assets/`
```
parkr: does this need the trailing slash?

site/_docs/themes.md
```diff
@@ -0,0 +1,99 @@
...
+Jekyll themes are distributed as Ruby gems. The only required file is the [Ruby Gemspec](http://guides.rubygems.org/specification-reference/). Here's an example of a minimal Gemspec for the `my-awesome-jekyll-theme` theme, saved as `/my-awsome-jekyll-theme.gemspec`:
+
+{% highlight ruby %}
+Gem::Specification.new do |s|
+  s.name     = 'My Awesome theme'
+  s.version  = '0.1.0'
+  s.license  = 'MIT'
+  s.summary  = 'This is an awesome Jekyll theme!'
+  s.author   = 'Dr. Jekyll'
+  s.email    = 'doc@jekyllrb.com'
```
parkr: we should probably make this blank or invalid so folks don't release with these values (inevitably someone will)

benbalter:
Chatting with @parkr offline, he suggested we remove asset support for a 0.1, since layouts, includes, and sass will all gain core support once https://github.com/jekyll/jekyll/issues/4610 lands, and assets would require custom logic. The intent is follow up with a 0.2 that includes support for assets, but we wanted to get the basic functionality in core.

Once https://github.com/jekyll/jekyll/issues/4610 lands,  we'll just have to add something like:

```ruby
site.includes_dir << theme.includes_path
site.layouts_dir  << theme.layouts_path
site.sass.source  << theme.sass_path
```

And we should be :sparkles:.

parkr:
@benbalter it might be best to just not merge master into feature branches in the future unless you need something that master has, in which case a rebase is in order.

spudowiar:
@parkr You've got the possibility of changes to `master` breaking WIP features, however unlikely

benbalter:
>  unless you need something that master has

There was a merge conflict, now resolved. :smile_cat:

envygeeks:
I don't know, without asset support this feature looses about 40% of it's value to me.  We'll never be able to use it here without asset's, preferably Jekyll-Assets being drop-in too.  We have at-scale sites using Jekyll that only benefit from this if assets can be merged into themes because they all have assets of some kind and js, images, and scss/less are all part of that bundle.

Whether or not it lands is irrelevant to us as at the end of the day because we have our own solution to address this problem across our sites, but with assets we can stop maintaining a disparity and rely on Jekyll to do the job which would make upstream tracking much cleaner.

spudowiar:
I'm looking at these commits and I'm struggling to understand why a) the tests suddenly broke?, b) they were working in the first place?

benbalter:
> without asset support this feature loses about 40% of it's value to me. 

Agreed. I'm hoping that (A) Jekyll assets can begin to use this, and that (B) it *will* gain asset support, but I wanted to keep the PR small, ship what we know we can, and then iterate by adding functionality.

spudowiar:
Is it possible to Pull Request a Pull Request?

test/test_theme.rb
```diff
@@ -0,0 +1,55 @@
...
+      end
+    end
+
+    should "generate folder paths" do
+      expected = File.expand_path("./_sass", @expected_root)
+      assert_equal expected, @theme.send(:path_for, :sass)
+    end
+
+    should "not allow paths outside of the theme root" do
+      assert_equal nil, @theme.send(:path_for, "../../source")
```
parkr: Doesn't this create `theme_dir/_../../source` because it's prefixed with the `_`? Maybe there's a way for us to get around that by doing `path_for :_layouts` and not adding the `_` when we resolve the dir. #security

spudowiar: @parkr how about resolving the path relative to the site source (and throwing errors for anything outside the site source). But that begs the question, should `path_for "folder/folder"` give `_folder/folder` or `folder/_folder`? Maybe we shouldn't allow subfolders?

parkr: > how about resolving the path relative to the site source

Do you mean relative to the theme directory?

> should `path_for "folder/folder"` give `_folder/folder` or `folder/_folder`?

That's why I'm interested in changing the method. `path_for "folder/folder"` should resolve the path `theme_dir/folder/folder`. The argument to `path_for` should be responsible for placing the `_` so we don't have to worry about this.

spudowiar: @parkr Yeah, theme directory :)

But if it just concatenates the theme directory, shouldn't it follow a similar naming convention to `Jekyll::Site`? But personally, I think `Jekyll::Site` and `Jekyll::Theme` should both extend a common class, many of their methods are similar and more similarities may be added.

parkr: > But personally, I think Jekyll::Site and Jekyll::Theme should both extend a common class

For now, I think it's fine as is. If there's a lot of cross-over, we can patch it up later. I considered writing a filesystem handler that would basically enforce that a path was in a certain base, but a simple call to `Jekyll.sanitized_path` works just fine for that.

Anyway, the Site should be the orchestrator, the Theme is just a certain way of supplying content.

benbalter:
Based on the discussion over in https://github.com/jekyll/jekyll/pull/4610, here's how I'm hoping to tackle this PR going forward:

1. Cherry @spudowiar's work from https://github.com/jekyll/jekyll/pull/4610 to abstract directories into arrays internally (with the intention of making it easy for a plugin to read a config key and inject a path into one of those arrays).

2. Assuming everything works, merge this as a bare-minimum 0.1 to get this out into the wild and see how it's used, then follow up with a few iterations:

  * 0.2 - Support for an `assets` folder
  * 0.3 - Developer improvements (e.g., `jekyll new-theme [PATH]`, passing a theme as a path, etc.)

parkr:
> Developer improvements (e.g., jekyll new-theme [PATH], passing a theme as a path, etc.)

@benbalter I'd like to ship `jekyll new-theme PATH` in Jekyll v3.2. Happy to help.

Release plan:

1. Merge this PR
2. Write a PR which generates a new theme
3. Release our current `site_template`'s layouts & includes as a theme & update site template to include this new theme
4. Release a beta of Jekyll v3.2
5. Write a blog post for our beta release soliciting feedback on the new feature.

Sound ok?

lib/jekyll/theme.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    private
+
+    def path_for(folder)
+      path = Jekyll.sanitized_path root, "_#{folder}"
+      path if Dir.exists?(path)
+    end
+
+    def gemspec
+      @gemspec ||= Gem::Specification.find_by_name(name)
```
parkr: @benbalter I think it would help discovery if we asked everyone to release themes under `jekyll-X-theme`. Is that totally infeasible, you think? Right now, I could set my theme to `activesupport` and it'd use a totally unrelated gem.

spudowiar:
@benbalter would you really want to allow `include`s from the main site, `include` files from a theme?

lib/jekyll/theme.rb
benbalter: I see what you're getting at, but not sure if it solves the problem. I could also set my theme as `jekyll-paginate` and we'd be in the same boat. The gem wouldn't have `_layouts`, `_includes`, or `_sass` directories, so it'd be a noop, no?

benbalter:
> would you really want to allow includes from the main site, include files from a theme?

I think so. Imagine you have a theme with a `default` template, with `header` and `footer` includes. If I just want to change the footer, I should be able to add an `_includes/footer.html` to my site, and Jekyll should use that instead.

lib/jekyll/theme.rb
parkr: @benbalter What if we enforced a certain [metadata field](http://guides.rubygems.org/specification-reference/#metadata) in the gemspec, like `s.metadata = { "theme" => true }`?

envygeeks: @benbalter @parkr I love abusing enumerators and Gem::Specification provides a few:

```ruby
theme = Gem::Specification.find_all do |g|
  (g.name == name || \
   g.name =~ /\Ajekyll\-theme\-#{Regexp.escape(name)}\Z/)) && \
   g.files.grep(/\A_theme\//) 
end
```

Expand it how you see fit to enforce any kind of basic assurance a theme is a theme.

benbalter: @parkr I like it. What if we made a standardize Jekyll key, (or `jekyll_type`, or similar) with values of `plugin` or `theme`? Just thinking of grabbing a namespace that may help with an eventual plugin/theme directory, or other automated disambiguation down the line.

parkr:
@benbalter I added all the necessary code –– if you could go ahead and add some tests for the theme stuff, that'd be swell. ❤️

benbalter:
@parkr Looking at my week, I'm not going to have the chance to right tests for this in the foreseeable future (but still very much would like to see this :ship:). Unassigning myself in case you or anyone else would like to jump in to unblock 3.2.

envygeeks:
I don't see why this should block 3.2? A feature should never block. only bugs.

parkr:
@benbalter It's working! 🎉

parkr:
@benbalter I updated it to resolve the `realpath` of the folder so people couldn't get around our symlink checks by shipping a theme with symlinks in them.

Well done, all!

@jekyllbot: merge +minor

spudowiar:
@parkr it's a minor enhancement? :/

AndorChen:
But what about plugins? After a quick test, I found Jekyll cannot load plugins in `_plugins` folder from a theme.

envygeeks:
> @parkr it's a minor enhancement? :/

A flag on the bot is no indication of what is actually happening.  It just happens to be the flag that the bot merges with at this time.  I'm sure it will be updated later and I'm sure @parkr realizes that.

> But what about plugins? After a quick test, I found Jekyll cannot load plugins in _plugins folder from a theme.

And it should remain so... for now.  If you wish to have themes quickly available on Github at your will, and whitelisted, your best option at this point is to disallow plugins from themes, as they post an risk to Github, and therefore require a more careful eye which impedes the goal of this pull request.  I'm sure in the future your request can be considered (and perhaps you could file a ticket to spark a discussion on it) but for now, the goal here is to give users a large array of themes they can quickly and readily use.

parkr:
> @parkr it's a minor enhancement? :/

@spudowiar Yep! Minor enhancements indicate a minor bump, major enhancements indicate a major bump. If it were labeled a major enhancement, we'd be on the road to Jekyll 4!

SirRawlins:
Hey folks, this is a great enhancement.

What's the state of play with _assets in a theme, are they supported yet? are they on a roadmap?

@envygeeks you mention you are running a workaround for this, may I ask about your setup? Happy to chat offline.

spudowiar:
I'm currently just calling Jekyll Assets on the folder from a plugin (on the Gem's `_assets` folder)

SirRawlins:
@spudowiar thanks for the reply. So your theme is utilizing the Jekyll-Assets gem? and that works out the box? You install the theme and the _assets folder becomes available to you?

spudowiar:
@SirRawlins it requires a file to be dropped into `_plugins` and the `jekyll-assets` Gem

SirRawlins:
@sparanoid Thanks. any chance of a copy of that plugin? Or a gist of what it's doing? Playing around with ideas myself here, I assume is determines the path of the theme and then passes that to the jekyll-assets sources config?

SirRawlins:
@spudowiar @envygeeks moved this into Jekyll talk to save polluting the PR. https://talk.jekyllrb.com/t/jekyll-assets-gem-based-theme/2667

PR #4597, opened 2016-02-26
Upgrading, documentation

mlinksva:
"Upgrading Documentation" reads to me like "upgrading the documentation" rather than "documentation for upgrading".

There's a link in the site navigation to documentation, but seems worth a mention here, even though the Google option will often bring one to it. :smile:


parkr:
@mlinksva You're absolutely right! Thank you for this PR! :smile: 

@jekyllbot: merge +site

PR #4596, opened 2016-02-26
Refresh the contributing file

benbalter:
This pull request contains a rewrite of the `CONTRIBUTING.markdown` file. Reading through the file, I felt that it was written for maintainers, not for potential contributors, that it tended to police contributions, more than encourage them.

[**Rendered**](https://github.com/jekyll/jekyll/blob/contributing-refresh/CONTRIBUTING.markdown)

To that end, this pull request does a few things:

* Explicitly encourage contributions, and make things a bit less intimidating for first-time contributors

* Create a new "where to get help or report a problem" section right up front, to help disambiguate the various Jekyll forums and resources

* Describe different ways to contribute, both technical and non-technical

* Explain the web-based pull request flow

* Generally try to make the file flow from least-technical to most technical, putting how before "you must" or specific implementation details

* Symlinked `/CONTRIBUTING.markdown` to `site/_docs/contributing.md` (we could theoretically do this in reverse as well) to keep the two in sync.

PR #4601, opened 2016-02-27
Unify method for copying special files from repo to site

benbalter:
In support of https://github.com/jekyll/jekyll/pull/4596, this pull request introduces a `siteify` method to the rake tasks, to DRY up copying markdown files from the root to the site (CONDUCT, CONTRIBUTING, and HISTORY).


envygeeks:
Can we add a bit of splitting to that method please? Perhaps use Pathname or Pathutil? Anything that merges the logic of all that `File#` stuff, including the replacement of extensions.

benbalter:
> Can we add a bit of splitting to that method please? Perhaps use Pathname or Pathutil? Anything that merges the logic of all that File# stuff, including the replacement of extensions.

@envygeeks I'm afraid I don't follow. Can you elaborate?

envygeeks:
@benbalter Pathutil can encapsulate everything you did more more cleanly.

```ruby
file = Pathutil.new(file)
data = file.read
contents = "stuff"

file.sub_ext(".md").write(
  contents
)
```

As stated before you'll have to manually add `pathutils` gem to the Jekyll dependencies until my pull to convert all our stuff to Pathutil gets in there but for now we can at least start doing bits and pieces more organized like.

benbalter:
@envygeeks started to implement things, and think a refactor like that is better suited for a subsequent pull request to keep this one targeted. The abstraction isn't as clean because we're actually working with two different files.

History.markdown
```diff
@@ -1,3 +1,5 @@
+# History
+
```
parkr: I would prefer to leave this out. I believe our changelog generator will nix it anyway.

Rakefile
```diff
@@ -87,6 +87,24 @@ def converted_history(markdown)
...
+  output_file = file.sub(/\.markdown\z/, ".md").downcase
+  slug = File.basename(output_file, ".md")
+  front_matter = front_matter.merge({
+    "title"     => title,
+    "layout"    => "docs",
+    "permalink" => "/docs/#{slug}/",
+    "note"      => "This file is autogenerated. Edit /#{file} instead."
+  })
+  contents.gsub!(/\A# #{title}\n\n?/, "")
+  contents = converted_history(contents) if output_file == "history.md"
```
parkr: This is a special case I'd rather avoid.

benbalter: > I would prefer to leave this out... This is a special case I'd rather avoid.
 
These two statements feel contradictory to me. Do you have any suggestions on how to convert the history otherwise? As a pre-process step? Leave the history processing as it is?

parkr: I didn't mean for them to be contradictory – I meant I'd rather avoid the `if` if possible. Could we pass in the content? Or have a method which fetches the content of a page and have a `case` in that method? `when /\Ahistory/`, then convert, otherwise read file and return?

benbalter: @parkr 7e21d2a what you had in mind?

benbalter:
> I think it's called History.markdown

Good call. Fixed.

parkr:
Thanks, Ben!

@jekyllbot: merge +dev

PR #4602, opened 2016-02-27
Add 'view source' entry

mlinksva:
Sometimes the best help is seeing how others' have done.

Just an idea...

I notice there's also https://talk.jekyllrb.com/t/showcase-sites-made-using-jekyll/18 and https://github.com/showcases/github-pages-examples but this Help page seems to use links exclusively in headings and second link probably includes non-Jekyll sites.


parkr:
Yes! @mlinksva, thank you for this. This will be very helpful for newcomers. I know I learned from others' sites when I was first starting. :+1: 

@jekyllbot: merge +site

PR #4606, opened 2016-02-28
Site template: Escape title and description where it is used in HTML

yous:
Follows after #4307. `site.title`, `page.title`, and `site.description` are usually written in plaintext, so escaping is needed.


parkr:
Agreed! :+1: 

@jekyllbot: merge +bug

yous:
Thanks!

Issue #4250, opened 2015-12-10
Fix flaky test


parkr:
Part of #4109 was an update to our tests for them. One seems to be flaking out somewhat frequently in Ruby 2.2:

```text
Finished tests in 114.908768s, 4.7255 tests/s, 7.9193 assertions/s.
Failure:
TestKramdown#test_: kramdown should render fenced code blocks with syntax highlighting.  [/home/travis/build/jekyll/jekyll/test/test_kramdown.rb:59]
Minitest::Assertion: Failed refutation, no message given
543 tests, 910 assertions, 1 failures, 0 errors, 0 skips
```

@envygeeks, know why that would sometimes not work?

envygeeks:
Without seeing the error I have no idea, It's not the test that's flaky more than likely, it's the stuff around it that is, we've always had fragile testing (which is why the rewrite is in the works,) but something could be exacerbating it with a mutation that isn't corrected somewhere because we don't properly mock.  I'll take a look.

parkr:
@envygeeks: It seems related to what we were discussing yesterday:

```ruby
    should "render fenced code blocks with syntax highlighting" do
      result = nokogiri_fragment(@markdown.convert(Utils.strip_heredoc <<-MARKDOWN))
        ~~~ruby
        puts "Hello World"
        ~~~
      MARKDOWN

      selector = "div.highlighter-rouge>pre.highlight>code"
      refute result.css(selector).empty?
    end
```

spudowiar:
+1, but annoyingly I cannot reproduce it on my computer to do debugging. I've only ever seen it on Travis CI.

parkr:
> @parkr Could I have some context on your system setup?

@spudowiar I am running Mac OS X 10.11.3. I use rbenv to setup my rubies, and probably ran this test with either 2.1.7-github (branch of github/ruby) or 2.3.0. I have libxml installed via Homebrew, though I think nokogiri uses its own version when I install.

spudowiar:
It's state leakage, you can reproduce it by running `rake TESTOPTS="--seed=31954" test` with any version of Ruby. This means some of the tests are modifying something that then breaks this test.

spudowiar:
It should be noted (although it might be my fault) that that seed breaks the OpenSSL test (because it's relying on it being `require`d elsewhere?)

spudowiar:
[...]

PR #4610, opened 2016-02-29
Allow multiple directories

spudowiar:
Work in progress and currently multiple `layouts_dir`s and `includes_dir`s has been implemented.

This is necessary for #4510 and is more generic than the current implementation in #4595 in that it allows you to list multiple `layout_dirs`, whereas #4595 manually deals with a single case.

### DIFF
--- a/lib/jekyll/readers/layout_reader.rb
+++ b/lib/jekyll/readers/layout_reader.rb
@@ -7,23 +7,30 @@ module Jekyll
     end
 
     def read
-      layout_entries.each do |f|
-        @layouts[layout_name(f)] = Layout.new(site, layout_directory, f)
+      layout_entries.each do |d, entries|
+        entries.each do |f|
+          @layouts[layout_name(f)] = Layout.new(site, d, f)
+        end
       end
 
       @layouts
     end
 
-    def layout_directory
-      @layout_directory ||= (layout_directory_in_cwd || layout_directory_inside_source)
+    def layout_directories
+      @layout_directories ||= Array(site.config['layouts_dir']).map do |directory|
+        layout_directory_in_cwd(directory) || layout_directory_inside_source(directory)
+      end
     end
 
     private
 
     def layout_entries
-      entries = []
-      within(layout_directory) do
-        entries = EntryFilter.new(site).filter(Dir['**/*.*'])
+      entries = {}
+      layout_directories.each do |directory|
+        entries[directory] = []
+        within(directory) do
+          entries[directory].concat EntryFilter.new(site).filter(Dir['**/*.*'])
+        end
       end
       entries
     end
@@ -37,12 +44,12 @@ module Jekyll
       Dir.chdir(directory) { yield }
     end
 
-    def layout_directory_inside_source
-      site.in_source_dir(site.config['layouts_dir'])
+    def layout_directory_inside_source(directory)
+      site.in_source_dir(directory)
     end
 
-    def layout_directory_in_cwd
-      dir = Jekyll.sanitized_path(Dir.pwd, site.config['layouts_dir'])
+    def layout_directory_in_cwd(directory)
+      dir = Jekyll.sanitized_path(Dir.pwd, directory)
       if File.directory?(dir) && !site.safe
         dir
       else
--- a/lib/jekyll/readers/layout_reader.rb
+++ b/lib/jekyll/readers/layout_reader.rb
@@ -7,9 +7,9 @@ module Jekyll
     end
 
     def read
-      layout_entries.each do |d, entries|
-        entries.each do |f|
-          @layouts[layout_name(f)] = Layout.new(site, d, f)
+      layout_entries.each do |directory, entries|
+        entries.each do |name|
+          @layouts[layout_name(name)] = Layout.new(site, directory, name)
         end
       end
 
--- a/lib/jekyll/tags/include.rb
+++ b/lib/jekyll/tags/include.rb
@@ -104,14 +104,32 @@ eos
 
       def render(context)
         site = context.registers[:site]
-        @includes_dir = tag_includes_dir(context)
-        dir = resolved_includes_dir(context)
+        @includes_dir = Array(tag_includes_dir(context))
+
+        includes_dir = @includes_dir.map do |dir|
+          resolved_includes_dir(context, dir)
+        end
 
         file = render_variable(context) || @file
         validate_file_name(file)
 
+        dir = includes_dir.reverse[0..-2].detect do |dir|
+          begin
+            path = File.join(dir, file)
+            validate_path(path, dir, site.safe, site.source)
+            true
+          rescue IOError => e
+            false
+          end
+        end
+ 
+        unless dir
+          dir = includes_dir[0]
+          path = File.join(dir, file)
+          validate_path(path, dir, site.safe, site.source)
+        end
+
         path = File.join(dir, file)
-        validate_path(path, dir, site.safe)
 
         # Add include to dependency tree
         if context.registers[:page] && context.registers[:page].key?("path")
@@ -129,7 +147,7 @@ eos
             partial.render!(context)
           end
         rescue => e
-          raise IncludeTagError.new e.message, File.join(@includes_dir, @file)
+          raise IncludeTagError.new e.message, path
         end
       end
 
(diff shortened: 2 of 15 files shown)


parkr:
@spudowiar Please ping me directly when this is finished. Thanks!

spudowiar:
Lovely! Hit by #4250!

spudowiar:
@benbalter @parkr Other than `_layouts` and `_includes`, which other directories need to be implemented?

benbalter:
> Other than _layouts and _includes, which other directories need to be implemented?

I believe just those two. For https://github.com/jekyll/jekyll/pull/4595, [sass has it's own mechanism](https://github.com/twbs/bootstrap-sass/blob/master/lib/bootstrap-sass.rb#L62) which we can use, and `_assets` or other static files will need to be added to static files separately.

spudowiar:
I was looking at SASS but it's mechanism only works when `!site.safe` so it will still need implementing.

benbalter:
> I was looking at `jekyll-sass-converter` but its mechanism only works when `!site.safe` so it will still need implementing.

It should work in safe mode (I use it for my site on Pages). I believe *some* configuration options are disabled via YAML, but we can override them via Ruby.

spudowiar:
In `jekyll-sass-converter` the only way to have multiple SASS directories is disabled when `site.safe` (see https://github.com/spudowiar/jekyll-sass-converter/blob/master/lib/jekyll/converters/scss.rb#L81)

Overriding it in Ruby is not a good idea as I intend for the same functionality to be achieved using `sass_dir` as an array and we would have to sanitise the path anyway so why not just do it there.

spudowiar:
@benbalter @parkr ETA on this being merged?

envygeeks:
hulk: smash +security +performance

envygeeks:
I have done what you requested, good luck. /cc @envygeeks

envygeeks:
Other then my bot needing some consistency fixes I see no major problems.  I'm :+1:

test/source/_layouts2/default2.html
```diff
@@ -0,0 +1,25 @@
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en-us">
+<head>
+   <meta http-equiv="content-type" content="text/html; charset=utf-8" />
+   <title>{{ page.title }}</title>
+   <meta name="author" content="<%= @page.author %>" />
```
pathawks: What's with the ERB tag?

spudowiar: It's from the existing `default.html`, I just duplicated it

parkr: Oh weird, it shouldn't have any ERB?

lib/jekyll/readers/layout_reader.rb
```diff
@@ -7,23 +7,30 @@ def initialize(site)
...
+      end
     end
 
     private
 
     def layout_entries
-      entries = []
-      within(layout_directory) do
-        entries = EntryFilter.new(site).filter(Dir['**/*.*'])
+      entries = {}
```
parkr: I suppose it is a bit less efficient, but using `reduce` or `inject` would be a bit more elegant, e.g.

```ruby
def layout_entries
  layout_directories.inject({}) do |layouts, directory|
    within(directory) do
      layouts[directory] = \
        EntryFilter.new(site).filter(Utils.safe_glob("*", ["**", "*.*"]))
    end 
    layouts
  end
end
```

lib/jekyll/readers/layout_reader.rb
```diff
@@ -7,23 +7,30 @@ def initialize(site)
...
+        end
       end
 
       @layouts
     end
 
-    def layout_directory
-      @layout_directory ||= (layout_directory_in_cwd || layout_directory_inside_source)
+    def layout_directories
+      @layout_directories ||= Array(site.config['layouts_dir']).map do |directory|
```
parkr: I am not convinced this should be configurable as an array. I think it's useful for themes, but the `Site` can ask the `Theme` for its directory. I'd rather see this as:

```ruby
def layout_directories
  @layout_directories ||= Array(site.config['layout_dir']) + Array(theme.layout_dir)
end
```

Before #4595, just the first piece will do.

lib/jekyll/tags/include.rb
```diff
@@ -104,14 +104,32 @@ def tag_includes_dir(context)
 
       def render(context)
         site = context.registers[:site]
-        @includes_dir = tag_includes_dir(context)
-        dir = resolved_includes_dir(context)
+        @includes_dir = Array(tag_includes_dir(context))
+
+        includes_dir = @includes_dir.map do |dir|
+          resolved_includes_dir(context, dir)
+        end
```
parkr: We already have `include_relative`, why do we need this?

parkr:
@spudowiar Ok, I have taken a look at this. So far, I am not convinced that this is required for gem-based plugins. The problem you're trying to solve (if I'm understanding correctly) is what @benbalter is trying to solve in his PR as well: being able to bring in resources from external sources. In terms of security and ease-of-use, I think his route will serve the community better. If a firm (such as 18F, GitHub, or even Jekyll) wishes to ship layouts or includes, they will be able to do so in a versioned gem. It solves the same problem very similarly but with some added perks.

Maybe @jekyll/stability and @jekyll/ecosystem have some thoughts about this?

lib/jekyll/readers/layout_reader.rb
envygeeks: Use each with object, it's faster than inject.

spudowiar:
There are many uses for this as a configuration option for where something is not complex enough to bundle as a theme

pathawks:
> There are many uses for this as a configuration option for where something is not complex enough to bundle as a theme

What are some use cases you have in mind?

spudowiar:
>What are some use cases you have in mind?

If you have, for example, two sites which share layouts, it's far easier to simply have a `_shared` directory that is a symlink or a Git submodule and adjust `layouts_dir` in `config.yml` than create a Gem containing these files and configuring that (also, what if you already have a theme set?)

pathawks:
It seems like Git submodules is exactly tge type of problem that Themes are aiming to solve.

spudowiar:
Themes are a great idea, someone creates a theme to be used by lots of people, lots of people use it, no version fragmentation, etc.

They are far too heavy for simple projects though. The problem with Git submodules is that people don't keep them up to date, they modify it and there are merge conflicts and messes etc.

But if you have a simple project, you don't need all this overhead, you can just have a simple shared folder, far simpler. It's a lot easier for most users, who won't know or want to know how to package a Gem and allows for quicker changes. (For a Gem, `modify file -> package gem -> install gem -> rebuild site`, for a shared folder it's `modify file -> rebuild site (optional, --watch also works)`

pathawks:
But a "Gem" can also just be a local folder; its not something you have to package & install if you’re only using it yourself.

spudowiar:
>But a "Gem" can also just be a local folder; its not something you have to package & install if you’re only using it yourself.

I don't understand, this uses RubyGems, how would you use a local folder? But still, this is a far simpler method

pathawks:
In your Gemfile:

```Gemfile
gem 'MyJekyllTheme', :path => '../MyJekyllTheme/'
```

spudowiar:
But still, you have to create a `Gemfile` or `.gemspec` for this folder

parkr:
Ok, after talking for a while with @benbalter, I'm happy with the decision to move forward with Gem-based themes *only*. The advantages to gem-based themes are significant:

1. Less prone to errors when sharing either your site or your theme ( 🌟 🌟 🌟 🌟 🌟 )
2. We will ship a command, `jekyll new-theme path_to_theme`, which makes the creation of the files and directories a one-step process. It will include instructions as well as a scaffold.
3. Private, ad-hoc themes are still supported through github.com (private repo) or local (`:path` in Gemfile) distribution.
4. A little added overhead for the theme creator makes it _much_ simpler for users. Versioning (including, crucially, the ability to roll back to a previous version) is paramount here.

This decision took me a couple weeks of thought which is much longer than I usually take for these sorts of things. I read all the comments above and talked it out at length with Ben, arguing that it should be configurable. Ultimately, he convinced me that the need for a wonderful user experience for the consumer of a theme (not the producer) over many months and years is what is important. Cloning a repository to some random location on your machine and pointing your config there is not a sustainable way to solve the problem of shared layouts, includes, and styles.

As an implementation detail, we will have to setup load paths for layouts & includes anyway, so a plugin could easily inject more paths if it so chose.

With that, I'm going to close this PR. @spudowiar, feel free to add support you added here for multiple directories to #4595, leaving out the configuration via `_config.yml`.

Thanks for the great discussion!

spudowiar:
@parkr Provided the internal implementation remains stable, I'm perfectly happy with this decision

parkr:
> @parkr Provided the internal implementation remains stable, I'm perfectly happy with this decision

@spudowiar It might take a few iterations to make it fully stable but we'll try to make migration to the new API really obvious and painless for plugin authors.

spudowiar:
@parkr I just want to be able to expose this functionality through a plugin that won't break randomly

parkr:
@spudowiar Absolutely. We try to do that but sometimes bugs are introduced. Did you join any of our Affinity Teams? https://teams.jekyllrb.com/

spudowiar:
@parkr Yeah, I'm in build and ecosystem

parkr:
@spudowiar Great! Keep an eye on those notifications. We'll be sure to notify if there is anything drastic happening – I don't suspect so but I'll use those teams as a means of ensuring this.

spudowiar:
@parkr Don't worry, been trying to help out on all of the issues and pull requests that come through those Affinity Teams as much as I can :)

PR #4620, opened 2016-03-01
Ensures related_posts are only set for a post

jeffkole:
Prior to this change, the related posts for the most recently rendered post
stayed set on the `site` object. This could result in pages that showed related
posts even when the page represented an entire collection of posts, such as on
an index page.  This change restores the functionality from Jekyll V2.


parkr:
I agree with this! Thank you for sending the patch with a test. :tada: 

Once Travis is done doing its thing, I'll merge if all looks good. :+1:

parkr:
@jekyllbot: merge +bug

PR #4622, opened 2016-03-02
codebeat badge

korzonek:
Is it fine to add codebeat badge to README? 

codebeat is automated code review tool for Swift, Ruby & Go that helps get instant feedback on code quality. 

"Quick wins" suggested by codebeat could be a nice candidate for a pull request and help other developers become contributors.

FYI. To be fully open and honest. I'm co-founder of codebeat.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -6,6 +6,7 @@
 [![Code Climate](https://codeclimate.com/github/jekyll/jekyll/badges/gpa.svg)][codeclimate]
 [![Dependency Status](https://gemnasium.com/jekyll/jekyll.svg)][gemnasium]
 [![Security](https://hakiri.io/github/jekyll/jekyll/master.svg)][hakiri]
+[![codebeat](https://codebeat.co/badges/b6b1292a-fbdc-4054-ba17-18b41df6d5ae)](https://codebeat.co/projects/github-com-jekyll-jekyll)
 
 [ruby-gems]: https://rubygems.org/gems/jekyll
 [gemnasium]: https://gemnasium.com/jekyll/jekyll


parkr:
Hey! Thanks for suggesting this.

Looking at Codebeat, it looks like the functionality is already covered by Code Climate, which we already have. Is there anything I'm missing?

korzonek:
There are a few things that are not available in code climate and you can find in codebeat.

* _Multi-Language support for **mobile & web** applications_. We implemented all algorithms on our own. It means relatively quickly we can support more languages both for mobile and web. At this stage we analyse **Swift**, Ruby, Python and **Go**.
* _Quick Wins_. After analysing your project we will suggest you issues in code with the biggest overall impact on your project's health. Try to refactor these hot spots first and see how your GPA improves.
* _Compare Changes_. In the stream of Recent Changes you will be able see what exactly caused your classes to improve or degrade.

We are also about to release nice support for pull requests in codebeat.

envygeeks:
> Multi-Language support for mobile & web applications. We implemented all algorithms on our own. It means relatively quickly we can support more languages both for mobile and web. At this stage we analyse Swift, Ruby, Python and Go.

I don't know if that's an upside, with CodeClimate I can fix problems by sending a bug report to Rubocop, which I have done a dozen times since we started using it on Jekyll.  We also aren't stuck on a platform with CodeClimate, because it uses open source tools, any one of those tools we can (and actually we do in some circumstances) decouple from the website entirely and allow any kind of workflow.

We don't expect people to use an app or a website and actually I consider that a downside for anybody but maintainers, we allow them to do `rubocop` (which is what we want them to do) and that same result is displayed on CodeClimate (for us maintainers) exactly as it's reported in `rubocop`.  Custom algorithms and tools prevent that and create vendor lockin.

I like the product (even if I found a few of the phrases used rather confusing -- such as "namespace too long") but I don't see what it provides that CodeClimate doesn't, and vendor lockin isn't such a good thing to aim for, the reason I suggested CodeClimate finally was because it didn't create vendor lockin, it was out of the users way (which is why we don't integrate with Pull requests) and users can use it from the terminal now.

mattr-:
>Multi-Language support for mobile & web applications. We implemented all algorithms on our own. It means relatively quickly we can support more languages both for mobile and web. At this stage we analyse Swift, Ruby, Python and Go.

We're not really a mobile or a web application though and we don't need the cross-platform nature of what Codebeat provides.

> Quick Wins. After analysing your project we will suggest you issues in code with the biggest overall impact on your project's health. Try to refactor these hot spots first and see how your GPA improves.

IMHO, we get these things from our integration with CodeClimate.

> Compare Changes. In the stream of Recent Changes you will be able see what exactly caused your classes to improve or degrade.

Also something we get with CodeClimate.

In short, I don't see a reason to add this. That's not because your tool is bad, but because we already get enough functionality out of our existing toolset and there's not a compelling reason to switch.

Issue #2252, opened 2014-04-21
Add an equivalent of post_url for collections


mscharley:
eg, given that I have a collection called `my_collection` and `_my_collection/test.md` exists, I should be able to do one of the following to access the link

```
{% post_url test.md %}
{% document_url test.md %}
```

Not sure on the naming of `document_url`. Is there something else perhaps? `collection_url` came to mind, but that may be misleading as we aren't linking to the collection but an item in the collection.

karlmdavis:
Isn't this feasible, now that all collection items have an ID property?

jeffkole:
The way I implemented this in [jekyll-page-collections](https://github.com/jeffkole/jekyll-page-collections#page_url-tag) was like so:

```
{% page_url projects 2014-03-05-jekyll-page-collections %}
```

Where the first argument to the tag is the name of the collection, which would need to match the name specified in the `_config.yml` file.  Would that be an acceptable solution?

PR #4624, opened 2016-03-02
Adds `link` Liquid tag to make generation of URL's easier

jeffkole:
This tag mirrors the post_tag functionality but for collections instead of just posts.


jeffkole:
This is a possible solution to #2252.

If the design looks good and the code is good, I can also add docs to describe how to use it.

mattr-:
What's the benefit to linking to a collection? I'm not seeing one at the moment since collections refer to a group of things and a post is a single entity.

jeffkole:
Perhaps this is just misnamed.  The tag links to a specific item in a collection, not to the collection itself.

If you have a collection that generates pages (ie, `output: true`), then you might want to link directly to those pages.  For instance, if you have a collection of blog authors and a page with bio is generated for each one, you may want to link to that page for a specific author from the byline of a post page.

budparr:
Not commenting on the PR, necessarily, but the use-case you bring up, @jeffkole, can be accomplished in 3.1 with something like this:

```
{% assign author = site.authors | where: 'slug', page.author %}

<a href="{{ author[0].url }}"> {{ author[0].title }}</a> 
```

jeffkole:
@budparr Let's say that I have another collection that is of books I have read and written summaries of, and that I like to link to those summary pages in posts that I write.  Don't you think it would be nicer for the user to write

```
In [The Hard Thing About Hard Things]({% collection_url books hard-thing-about-hard-things %}), Ben Horowitz describes...
```

than

```
In [The Hard Thing About Hard Things]({% assign book = site.books | where: 'slug', 'hard-thing-about-hard-things' %}{{ book[0].url }}), Ben Horowitz describes...
```

The first strikes me as a nice clean interface to get a specific job done.  It can be documented and tested.  The second is obviously more flexible, but it requires knowing more about the internals of Jekyll to use it.

budparr:
@jeffkole - Neither one seems all that easy (I think if I were doing that I'd grab the url directly).

But, I had/have no intention to argue the elegance of the solution. I only brought it up because being able to access a collection item's `slug` is new as of 3.1, and makes accessing an item much easier than before, so I wanted to make sure you were aware of it. 

Cheers!

parkr:
@jeffkole I'm interested in a multi-purpose tag which can work for more than just collection documents.

What do you think about `{% link _books/hard-things-about-hard-things.md %}`? It'd check against all documents in `site.docs_to_write` for that `relative_path` and output the `url` for that document, or error otherwise.

jeffkole:
@parkr Updated to be a `link` tag.  Let me know what you think.  Probably could stand a rebase, and then I can add documentation, too.

envygeeks:
IMO, rebasing is dead now.  Github just made it easy to review changes without rebasing.

test/test_tags.rb
```diff
@@ -14,6 +14,9 @@ def create_post(content, override = {}, converter_class = Jekyll::Converters::Ma
     if override['read_posts']
       site.posts.docs.concat(PostReader.new(site).read_posts(''))
     end
+    if override['read_collections']
```
envygeeks: It would be preferable to make both these if's single line and stack them above the other.

lib/jekyll/tags/link.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      def render(context)
+        site = context.registers[:site]
+
+        site.docs_to_write.each do |document|
+          return document.url if document.relative_path == @relative_path
+        end
+
+        raise ArgumentError.new <<-eos
+Could not find document "#{@relative_path}" in tag '#{TagName}'.
```
envygeeks: * Use ArgumentError, "string".

Also:

```ruby
raise ArgumentError, "This is my long string\n" \
  "This is the second part."
```

jeffkole: I was following the convention set in `post_url.rb`.  Has the style changed?

test/test_tags.rb
jeffkole: Like this?

```
site.posts.docs.concat(PostReader.new(site).read_posts('')) if override['read_posts']
CollectionReader.new(site).read if override['read_collections']
```

lib/jekyll/tags/link.rb
envygeeks: For removing the "new" yes.  For the former, it's up for debate @jekyll/core will have to decide that though as of late it seems a lot of errors are shifting that way because it's cleaner to read.

test/test_tags.rb
envygeeks: Yes, that looks much cleaner!

parkr: Isn't that over 90 chars?

envygeeks: @parkr the bot (and I believe Rubocop too) doesn't track line width on Jekyll/Jekyll for now because we have a bunch of lines way over that and it was impacting the score for something so trivial that we could fix over time and didn't impact the actual code or it's readability for the most part.

test/test_tags.rb
```diff
@@ -476,6 +475,69 @@ def highlight_block_with_opts(options_string)
...
+    should "not cause an error" do
+      refute_match /markdown\-html\-error/, @result
+    end
+
+    should "have the url to the \"sanitized_path\" item" do
+      assert_match %r{1\s/methods/sanitized_path}, @result
+    end
+
+    should "have the url to the \"site/generate\" item" do
+      assert_match %r{2\s/methods/site/generate}, @result
```
parkr: Should these have a file extension?

test/test_tags.rb
```diff
@@ -476,6 +475,69 @@ def highlight_block_with_opts(options_string)
...
+    should "cause an error" do
+      content = <<CONTENT
+---
+title: Invalid linking
+---
+
+{% link non-existent-collection-item %}
+CONTENT
+
+      assert_raises ArgumentError do
```
parkr: Note to self: change this in https://github.com/jekyll/jekyll/pull/4674.

parkr:
Thanks, @jeffkole. One comment above about the tests then I'm 👍 to 🚢.

jeffkole:
Thanks for the review @parkr and @envygeeks.  I think this should be good to go now.

parkr:
@jekyllbot: merge +minor

parkr:
Thanks a lot, @jeffkole!

kainjow:
Should this tag be documented on the website? It's mentioned in the release notes for 3.2.

parkr:
@kainjow Huh. Yeah, it should be. It should be documented like [`post_url` is](https://jekyllrb.com/docs/templates/#post-url), probably at the top of that list.

DirtyF:
@jeffkole A few questions regarding the current implementation : 

1. Should we be able to link to a page with `link? 
2. Would it be possible to keep the same behavior as `post_url` and omit the file extension in the relative path?

jeffkole:
@DirtyF You should be able to link to any document that Jekyll is generating, including pages. Is that not working?

As for the file extension, it could certainly be done.  I'll leave it to @parkr for guidance, since that kind of stylistic decision is more up to him than me.  If he likes it, I can make the changes.

parkr:
> Could we keep the same behavior as post_url and omit the file extension in the relative path?

@DirtyF What's the reason for this?

DirtyF:
@jeffkole I didn't manage to link to pages when testing for #5182 

@parkr just for cohesiveness between `post_url` and `link`tags. IMHO as a user it's not intuitive to have difference between both syntax.

jeffkole:
Opened #5199 to link to pages and static files.

PR #4628, opened 2016-03-03
Update continuous-integration docs with sudo: false information

mejackreed:
Updates the continuous-integration docs to provide information about the `sudo: false` behavior which can route builds to the faster container-based infrastructure. https://docs.travis-ci.com/user/workers/container-based-infrastructure/#Routing-your-build-to-container-based-infrastructure

### DIFF
--- a/site/_docs/continuous-integration.md
+++ b/site/_docs/continuous-integration.md
@@ -112,6 +112,8 @@ branches:
 env:
   global:
   - NOKOGIRI_USE_SYSTEM_LIBRARIES=true # speeds up installation of html-proofer
+
+sudo: false # route your build to the container-based infrastructure for a faster build
 {% endhighlight %}
 
 Ok, now for an explanation of each line:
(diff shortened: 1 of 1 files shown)


mattr-:
I like this idea, but I think it makes more sense to just have `sudo: false` be the default instead of pointing it out as an option.

Could you switch the language in the docs around to default to `sudo: false` and then explain how to enable `sudo` if something fails?

parkr:
> Could you switch the language in the docs around to default to sudo: false and then explain how to enable sudo if something fails?

:point_up: Yes please!

Also, @mattr-, you can use @jekyllbot to merge when you're ready. :smile: I just added better auth so that anyone who has push access to a repo can use jekyllbot to merge.

mejackreed:
Thanks for your feedback! I've modified the language in the following way:

 - removed "optional" from the example `travis.yml` configuration
 - Added "By default you should supply the `sudo: false` command to Travis."
 - Added "If you have any trouble with your build, or if your build does need `sudo` access, remove the `sudo:false` line."

I amended the commit and pushed it up. Let me know if you would like anything else changed.

site/_docs/continuous-integration.md
```diff
@@ -202,6 +204,16 @@ environment variable `NOKOGIRI_USE_SYSTEM_LIBRARIES` to `true`.
 exclude: [vendor]
 {% endhighlight %}
 
+By default you should supply the `sudo: false` command to Travis. This command 
+explicitly tells Travis to run your build on Travis's [container-based
+ infrastructure](https://docs.travis-ci.com/user/workers/container-based-infrastructure/#Routing-your-build-to-container-based-infrastructure). Running on the container-based infrastructure can often times
+speed up your build. If you have any trouble with your build, or if your build
+does need `sudo` access, remove the `sudo:false` line.
```
parkr: I believe you have to actually specify `sudo: required` in this case. Travis caches your previous sudo value, so it would still use `false`.

mejackreed: Thanks @parkr I didn't realize that. Modified to use your suggestion.

parkr:
:heart:

@jekyllbot: merge +site

PR #4630, opened 2016-03-03
Adding Aerobatic to list of deployment options

dvonlehman:
Aerobatic is a static host with built in CD Jekyll builds


parkr:
Thanks!

@jekyllbot: merge +site

PR #4631, opened 2016-03-04
Adding a test gate ensuring that the posts' content is YAML-valid

borisschapira:


mattr-:
I don't believe the CI guide is really meant to be a catch all for all the way that you might set up continuous integration for your Jekyll site, but I'm happy to be told I'm wrong.

//cc @jekyll/core would love some input here. :smiley:

parkr:
Hey @borisschapira! There's specialized code for reading posts that you'd want to plug into. If you did that, though, you may as well be running `jekyll build` or `jekyll doctor`. I like the idea but we should improve our error handling in Jekyll in these situations so errors are surfaced. Would you be willing to do that instead?

borisschapira:
Hi @parkr! In fact I'm not aware of the behaviour of `jekyll doctor` and I find very difficult to gather information about it as many search end with an article on the "Strange Case of Dr Jekyll and Mr Hyde" book.

[Reading the code](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/commands/doctor.rb), I understand it provides some kind of tests to run on content so yes, I think I will try to propose something there instead of in an external component.

parkr:
@borisschapira We could definitely add it to the top of https://jekyllrb.com/help/!

PR #4632, opened 2016-03-04
DOC: Describe --allow-hash-href arg for htmlproof

scottclowe:
It is very confusing to new users (it was to me at least) when you initialise a fresh jekyll site with `jekyll new PATH`, add the tests to `.travis.yml`, and then find that they don't pass.

It took a while for me to discover that this is because the initial site includes links to `#`, and there is the `--allow-hash-href` argument which can be supplied to `htmlproof` to allow this behaviour.

This commit adds a description of the `--allow-hash-href` argument and the need to use it for a fresh initialisation of jekyll to the documentation.

If you aren't going to include `--allow-hash-href` as standard in the bash script `./script/cibuild` (i.e. Line 46), it seems to me that you should mention it as an option.
Alternatively, one could stop linking to `#` in the jekyll site created with `jekyll new` so this problem does not arise.

### DIFF
--- a/site/_docs/continuous-integration.md
+++ b/site/_docs/continuous-integration.md
@@ -56,6 +56,21 @@ For example to avoid testing external sites, use this command:
 $ bundle exec htmlproof ./_site --disable-external
 {% endhighlight %}
 
+and to allow internal links which point to `#` (and empty anchor on the
+current page), use this command
+
+{% highlight bash %}
+$ bundle exec htmlproof ./_site --allow-hash-href
+{% endhighlight %}
+
+which is necessary for an initial jekyll site created with
+
+{% highlight bash %}
+jekyll new PATH
+{% endhighlight %}
+
+to pass.
+
 ### The HTML Proofer Library
 
 You can also invoke `html-proofer` in Ruby scripts (e.g. in a Rakefile):


envygeeks:
I'm kind of against this as it stands, I don't think it belongs there, can we just create a page dedicated to using htmlproofer with Jekyll instead? As that page is about continuous integration and this option has little to do with that in the end.

parkr:
@scottclowe Interesting. From https://github.com/gjtorikian/html-proofer/pull/264, the author of the PR writes:

> This option, `--allow-hash-href`, is equivalent to `--href-ignore '#'`.

We shouldn't be ignoring the hashed hrefs, we should be testing them! What error output are you seeing?

> can we just create a page dedicated to using htmlproofer with Jekyll instead

@envygeeks We made this page to do that. It talks very briefly about just doing `jekyll build`, but that's not really that useful. To CI a Jekyll site, the idea is that you want valid HTML that links to real things, etc. Is the page getting too long?

scottclowe:
@envygeeks If you would not like the CI tutorial to get any longer, we could cut the `--disable-external', replacing it with the more pertinent `--allow-hash-href` example.

@parkr It does not ignore all hashed hrefs, only the href which is a hash on its own with no anchor on the page to point to.
```html
<a href="#">link</a>
```

Steps to reproduce:

```bash
gem install jekyll --user-install
gem install html-proofer --user-install
jekyll new my-awesome-site
cd my-awesome-site
jekyll build
htmlproof ./_site
```

Error message:
```
Running ["ScriptCheck", "LinkCheck", "ImageCheck"] on ./_site on *.html... 


Checking 12 external links...
Ran on 3 files!


- ./_site/about/index.html
  *  linking to internal hash # that does not exist (line 28)
- ./_site/index.html
  *  linking to internal hash # that does not exist (line 28)
- ./_site/jekyll/update/2016/03/09/welcome-to-jekyll.html
  *  linking to internal hash # that does not exist (line 27)
htmlproof 2.6.4 | Error:  HTML-Proofer found 3 failures!
```

An alternative solution would be to not use `"#"` as a link in the content generated with `jekyll new`.

Would it be better if I made a regular issue instead of a PR so the issue can be discussed fully with this option also considered?

parkr:
@scottclowe Interesting. Where is that link? Should we just remove it from the site_template? Looks like it's in some partial.

scottclowe:
It's [here](
https://github.com/jekyll/jekyll/blob/37cc45239db392d368cfe2b0acfe15a9a0793949/lib/site_template/_includes/header.html#L8).

pathawks:
```html
    <nav class="site-nav">
      <a href="#" class="menu-icon">
        <svg viewBox="0 0 18 15">
```

It comes from [`_includes/header.html`](https://github.com/jekyll/jekyll/blob/09b9db259c56236cbe49fb1d0a35e455dab0c90c/lib/site_template/_includes/header.html#L8)


**edit:** @scottclowe Beat me to it :+1:

parkr:
Great! Is there any downside to removing that `href`? It's not needed in HTML5 anyway, AFAICT.

pathawks:
Correct, [rot required!](https://www.w3.org/TR/2012/WD-html5-20121025/links.html#attr-hyperlink-href)

Let’s tackle the problem at the source, rather than relaxing our linters.

scottclowe:
I'll close this PR then.

pathawks:
@scottclowe Mind opening another PR to remove that `href="#"`?

envygeeks:
I've not read any of this but chiming in to throw my distaste for `href="#"` because an accidental click is a lost spot in the page, blank href's are better and should be entirely valid.  I don't know why people see the need to fill them in.

```html
<a href>This works and is valid.</a>
```

scottclowe:
@pathawks I'm not sure `href="#"` should be replaced with.

@envygeeks Unassigned or empty href attributes are treated as errors by htmlproof, so that does not solve this problem.

```
$ htmlproof ./_site/
Running ["ScriptCheck", "LinkCheck", "ImageCheck"] on ./_site/ on *.html... 


Checking 12 external links...
Ran on 3 files!


- ./_site/about/index.html
  *  anchor has no href attribute (line 28)
- ./_site/index.html
  *  anchor has no href attribute (line 28)
- ./_site/jekyll/update/2016/03/09/welcome-to-jekyll.html
  *  anchor has no href attribute (line 27)
htmlproof 2.6.4 | Error:  HTML-Proofer found 3 failures!
```

envygeeks:
That's a bug in HTMLProofer then and should be taken upstream at that point.

PR #4633, opened 2016-03-04
Added a default lang attribute

HugoGiraudel:
Omitting the `lang` attribute is [an accessibility concern](https://twitter.com/stevefaulkner/status/705767128163606528), and given how popular Jekyll is I thought it would be a good idea to set a default `lang` attribute to the `html` element. I went with `en` as I figured out it would be a meaningful default.

### DIFF
--- a/lib/site_template/_layouts/default.html
+++ b/lib/site_template/_layouts/default.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
 
   {% include head.html %}
 


envygeeks:
@jekyllbot merge +site

envygeeks:
ah boo he doesn't listen to me yet.

envygeeks:
:heart:

HugoGiraudel:
Wow, is it that easy to update the default Jekyll template? I’ll come back. :D

parkr:
@envygeeks you need a colon after @jekyllbot's name. kinda strict on purpose :wink:

parkr:
@HugoGiraudel This is not necessarily the case, I don't think this should be a part of the default. If anything, it should be `{{ site.lang }}` and add `lang` to the `_config.yml`. A lot of people use this template to write in all sorts of languages, so hard-coding this would be lying to the interwebs!

HugoGiraudel:
> If anything, it should be {{ site.lang }} and add lang to the _config.yml.

That is a very good idea. A very good idea indeed. Sorry I did not think of it in the first place. Let’s do that instead.

PR #4634, opened 2016-03-04
Added role="contentinfo" to <footer>

HugoGiraudel:
Cheers. :)

### DIFF
--- a/lib/site_template/_includes/footer.html
+++ b/lib/site_template/_includes/footer.html
@@ -1,4 +1,4 @@
-<footer class="site-footer">
+<footer class="site-footer" role="contentinfo">
 
   <div class="wrapper">
 


envygeeks:
I don't know if that's the proper aria role for that footer.

mattr-:
There really isn't a good ARIA role for footers. `contentinfo` is probably the closest role for something like a footer, but i'm not sure it's close enough that we should add it to the site template.

In this case, I'd rather do nothing than put something in wrong. /cc @jekyll/core thoughts on just closing this out for now?

HugoGiraudel:
Maybe we could ping @heydon and @ffoodd as a11y experts?

Heydon:
contentinfo is correct for document/page footers.

ffoodd:
My two cents: [the WAI-ARIA spec](https://www.w3.org/TR/wai-aria/roles#contentinfo).

This footer is included in every page in the default site template, am I right? If so:

Pro: as a default file and according to its current content, I think it's accurate.

Con: `<footer>`'s content may be modified to something that shouldn't be in  `[role="contentinfo"]`, and user may not know he should remove the role.

I'd go with the role :)

parkr:
I'm sorry to say I don't know very much about ARIA. We have been hesitant to add too much complexity to this default template, and I'm afraid these types of attributes make it that much more difficult to pick up and use out of the box. I have been doing web development for about 5 years now – if I haven't learned what ARIA is and how to use it, how can we expect every user to know? 

I'm :-1: on ARIA roles. I'd be happy to go full-monty on accessibility once gem-based themes ( #4595 ) ships. Then, this complexity is mostly hidden from the user and therefore we can add it without any worry that they will misuse it.

Would that work?

ffoodd:
IMHO this sounds perfect to avoid mistakes, indeed :)

mattr-:
Sounds like we have an agreement to close this PR and I am doing so based on that agreement. We'll revisit a better placed role for accessibility again once #4595 ships. 

Thanks everybody for the great discussion! :heart:

PR #4647, opened 2016-03-10
Fixes typo on collections

gesa:
There was a line referring to the `render` key in `_config.yml` but the actual name of the key is `output`.

### DIFF
--- a/site/_docs/collections.md
+++ b/site/_docs/collections.md
@@ -303,11 +303,8 @@ file, each document has the following attributes:
       </td>
       <td>
         <p>
-          The URL of the rendered collection. The file is only written to the
-          destination when the name of the collection to which it belongs is
-          included in the <code>render</code> key in the site's configuration
-          file.
-        </p>
+          The URL of the rendered collection. The file is only written to the destination when the collection to which it belongs has <code>output: true</code> in the site's configuration.
+          </p>
       </td>
     </tr>
     <tr>


site/_docs/collections.md
```diff
@@ -305,7 +305,7 @@ file, each document has the following attributes:
         <p>
           The URL of the rendered collection. The file is only written to the
           destination when the name of the collection to which it belongs is
-          included in the <code>render</code> key in the site's configuration
+          included in the <code>output</code> key in the site's configuration
```
parkr: I guess this whole paragraph is wrong. What about:

> The URL of the rendered collection. The file is only written to the destination when the collection to which it belongs has `output: true` in the site's configuration.

gesa:
@parkr you're absolutely right. thanks for the note, updated now.

site/_docs/collections.md
```diff
@@ -303,11 +303,8 @@ file, each document has the following attributes:
       </td>
       <td>
         <p>
-          The URL of the rendered collection. The file is only written to the
-          destination when the name of the collection to which it belongs is
-          included in the <code>render</code> key in the site's configuration
-          file.
-        </p>
+          The URL of the rendered collection. The file is only written to the destination when the collection to which it belongs has `output: true` in the site's configuration.
```
parkr: Unfortunately the backticks won't work here. Could you use `<code>output: true</code>` instead?

gesa: shit, i'm sorry. habits die hard. updated now!

parkr:
Thanks so much! :heart: 

@jekyllbot: merge +site

PR #4657, opened 2016-03-12
Emboldened the "Jekyll" word

SuriyaaKudoIsc:
@parkr 

----

#### Why did I emboldened the "Jekyll" word?

By contrast, an **emboldened font** weight makes text darker than the surrounding text.

With this technique, the emphasized text strongly stands out from the rest; we should therefore be used to highlight certain keywords like "**`Jekyll`**" that are important to the subject of the text, for easy visual scanning of text.

----

Yours,
[**Suriyaa Kudo**](https://github.com/SuriyaaKudoIsc) :octocat:

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -14,7 +14,7 @@
 [hakiri]: https://hakiri.io/github/jekyll/jekyll/master
 [travis]: https://travis-ci.org/jekyll/jekyll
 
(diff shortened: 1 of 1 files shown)


alfredxing:
Thanks for your PR!

Personally I'm not too keen on this change. I don't think there's a need to emphasize this particular instance of "Jekyll", especially since it's right underneath the huge heading.

SuriyaaKudoIsc:
:+1: Ok.

SuriyaaKudoIsc:
Ok.

PR #4672, opened 2016-03-16
Add ability to jsonify Drop

### LINKED ISSUE #4668: "Liquid Exception: stack level too deep" on jsonify

jekyll 3.0.3

Darwin ***** 15.3.0 Darwin Kernel Version 15.3.0: Thu Dec 10 18:40:58 PST 2015; root:xnu-3248.30.4~1/RELEASE_X86_64 x86_64

###### What did you do?

```
$ jekyll build
```

in `_layouts/city.html` this line is present:

```
{{ page | jsonify }}
````

A complete build.

```
    Server address: http://127.0.0.1:4000//
      Regenerating: 1 file(s) changed at 2016-03-14 22:08:52 ...done in 1.435773 seconds.
      Regenerating: 1 file(s) changed at 2016-03-14 22:08:56
  Liquid Exception: stack level too deep in _layouts/city.html
```

`{{ page | jsonify }}` causes the exception. This used to work previously. I assume `jsonify`ing the `page` object breaks because something in the page object is of recursive nature or not "dumpable". `jsonify`ing individual keys in `page` (e.g. `{{ page.item | jsonify }}`) works.

### PR DESCRIPTION
**Pull request by stefanw**
Currently `{{ page | jsonify }}` returns a JSON string of the page content while a JSON representation of the pages data is more useful. This PR adds a `to_json` to the `Drop` class and dumps the YAML front matter data.

The Drop contains more than just a page's YAML front matter data. However, dumping the complete drop is apparently not possible due to the recursive nature of some objects (e.g. probably `previous` and `next`).

This PR would close #4668.

### DIFF
--- a/lib/jekyll/drops/drop.rb
+++ b/lib/jekyll/drops/drop.rb
@@ -144,7 +144,14 @@ module Jekyll
       #
       # Returns a JSON dump of the YAML front matter data.
       def to_json
-        fallback_data.to_json
+        results = keys.each_with_object({}) do |(key, _), result|
+          if %w{previous next}.include? key
+            result[key] = self[key].inspect
+          else
+            result[key] = self[key]
+          end
+        end
+        results.to_json
       end
 
       # Collects all the keys and passes each to the block in turn.
--- a/lib/jekyll/drops/drop.rb
+++ b/lib/jekyll/drops/drop.rb
@@ -137,7 +137,16 @@ module Jekyll
         require 'json'
         JSON.pretty_generate to_h
       end
-      alias_method :to_json, :inspect
+
+      # When a JSON representation of the drop is requested (e.g. through
+      # template filters) then only show the YAML front matter data.
+      # Other Jekyll objects might not be dumpable due to recursiveness.
+      #
+      # Returns a JSON dump of the YAML front matter data.
+      def to_json
+        require 'json'
+        JSON.generate fallback_data
+      end
 
       # Collects all the keys and passes each to the block in turn.
       #
--- a/lib/jekyll/drops/drop.rb
+++ b/lib/jekyll/drops/drop.rb
@@ -137,6 +137,7 @@ module Jekyll
         require 'json'
         JSON.pretty_generate to_h
       end
+      alias_method :to_json, :inspect
 
       # Collects all the keys and passes each to the block in turn.
       #
--- a/lib/jekyll/drops/drop.rb
+++ b/lib/jekyll/drops/drop.rb
@@ -144,8 +144,7 @@ module Jekyll
       #
       # Returns a JSON dump of the YAML front matter data.
       def to_json
-        require 'json'
-        JSON.generate fallback_data
+        fallback_data.to_json
       end
 
       # Collects all the keys and passes each to the block in turn.
--- a/lib/jekyll/drops/drop.rb
+++ b/lib/jekyll/drops/drop.rb
@@ -145,8 +145,8 @@ module Jekyll
       # Returns a JSON dump of the YAML front matter data.
       def to_json
         results = keys.each_with_object({}) do |(key, _), result|
-          if %w{previous next}.include? key
-            result[key] = self[key].inspect
+          if %w{previous next}.include? key && !self[key].nil?
+            result[key] = self[key].url
           else
             result[key] = self[key]
           end
--- a/test/test_document.rb
+++ b/test/test_document.rb
@@ -44,6 +44,16 @@ class TestDocument < JekyllUnitTest
       assert_equal "foo.bar", @document.data["whatever"]
     end
 
+    should "be able to jsonify its data" do
+      require 'json'
+
+      page_json = @document.to_liquid.to_json
+      parsed = JSON.parse(page_json)
+
+      assert_equal 'Jekyll.configuration', parsed['title']
+      assert_equal 'foo.bar', parsed['whatever']
+    end
+
     context "with YAML ending in three dots" do
 
       setup do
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by stefanw** [comment]
I updated the original PR to reflect that only YAML front matter as present in the Drop's `fallback_data` is JSON dumped.

File: lib/jekyll/drops/drop.rb [hunk 1]
```diff
@@ -137,6 +137,7 @@ def inspect
         require 'json'
         JSON.pretty_generate to_h
       end
+      alias_method :to_json, :inspect
```
envygeeks: * Use alias without symbolizations: `alias method other_method`
* Group `alias_method` at the bottom of the class rather than close to the method.
* Use alias instead of alias method.

test/test_document.rb
```diff
@@ -44,6 +44,16 @@ def assert_equal_value(key, one, other)
       assert_equal "foo.bar", @document.data["whatever"]
     end
 
+    should "be able to jsonify its data" do
+      require 'json'
```
envygeeks: * Do not require within a test.  This is bad practice, if the library is not available you're doing it wrong.

lib/jekyll/drops/drop.rb
```diff
@@ -137,7 +137,16 @@ def inspect
...
-      alias_method :to_json, :inspect
+
+      # When a JSON representation of the drop is requested (e.g. through
+      # template filters) then only show the YAML front matter data.
+      # Other Jekyll objects might not be dumpable due to recursiveness.
+      #
+      # Returns a JSON dump of the YAML front matter data.
+      def to_json
+        require 'json'
+        JSON.generate fallback_data
```
envygeeks: * Prefer to use `to_json` instead of `JSON.generate` unless you intend to use `JSON.pretty_generate`

envygeeks:
Just a few comments after the bot did it's review but I'm :+1: this seems to be solid.

test/test_document.rb
stefanw: How should I do it right then? I'm not a Ruby person, sorry!

lib/jekyll/drops/drop.rb
stefanw: I mimicked the method aliasing done for the `to_h` method. But this is not part of the PR now anyway ;)

test/test_document.rb
envygeeks: In this case I would just add it to Jekyll's main require file because it's part of StdLib but if you want to be local I would require it at the top of formatters file you edited because it has no performance impact and doesn't slow anything down because most of the time it's required anyways by one thing or another.

parkr: Sorry for the tone of the bot, @stefanw, in that first comment. I'm sure it didn't mean to tell you off quite so harshly.

You should be able to remove this `require`. From the looks of it, the method above `#to_json` uses `JSON` too, so it should already be required. Were you seeing it break before?

parkr: Ah, I misspoke. It's required in `#inspect`. So it looks like you'll have to manually require it either in `lib/jekyll.rb` or as the first line of the body of this method.

Why don't you just omit certain keys? Maybe something like:

```ruby
keys.each_with_object({}) do |(key, _), result|
  result[key] = self[key] unless %w{previous next}.include? key
end
```

lib/jekyll/drops/drop.rb
```diff
@@ -138,6 +138,15 @@ def inspect
         JSON.pretty_generate to_h
       end
 
+      # When a JSON representation of the drop is requested (e.g. through
+      # template filters) then only show the YAML front matter data.
+      # Other Jekyll objects might not be dumpable due to recursiveness.
+      #
+      # Returns a JSON dump of the YAML front matter data.
+      def to_json
+        fallback_data.to_json
```
parkr: This feels incomplete to me... We definitely need more than just the fallback data to be jsonified. Ideally we just call `.inspect` on the values in `next` and `previous`.

parkr:
@stefanw Still have time to work on this, or are you unable to complete this? I have a concern above I'd like to know your thoughts on.

stefanw:
@parkr I could whitelist or blacklist keys to include in the dump, but I guess the best way would be to have [all variables listed in the documentation](https://jekyllrb.com/docs/variables/) to be available in the dump.

In addition to `page`, `site` and `paginator` should also be dumpable, I think. The question remains how to avoid dumping without loops. E.g. I could see `page.previous` just becoming a URL path in the dump.

Is that what you have in mind?

parkr:
> The question remains how to avoid dumping without loops.

If we can identify `page.next` and `page.previous` as problematic, I think representing them via `inspect` is fine. We can start with those two properties and add more later if they come up.

stefanw:
Representing via `inspect` results in this though:

```json
{
"next": "#<Jekyll::Document _cities/rochester.md collection=cities>",
"previous":"#<Jekyll::Document _cities/rennes.md collection=cities>",
"path":"_cities/rhein-neckar.md",
"...": "..."
}
```

Is that really useful in a JavaScript context?

pathawks:
> Is that really useful in a JavaScript context?

Nope. A URL would be useful.

stefanw:
Now I'm calling `url` on the previous/next values when they're not `nil`.

parkr:
@stefanw @pathawks A URL is invalid, though. If you were using `jsonify` to try to understand what `page.next` was, they'd see a URL even though it's the Document. We should represent this some other way.

stefanw:
I used a URL in the spirit of HATEOAS. If you're interested in what the next document is, you can send a request to that URL which then contains that info.

parkr:
@stefanw That's fine in theory but not everyone is writing a REST-ful service so HATEOAS is not valid here. People often get confused if `page.next` is a document or a URL and this would further confuse them. It really must be a document itself. It can be limited, but it must be a document.

parkr:
#4945 fixed this.

Issue #4623, opened 2016-03-02
Slugify a string doesn't seem to work on Unicode/Swedish letters


Saucy:
###### What version of Jekyll are you using (`jekyll -v`)?

v3.0.3 (using `github-pages` gem)
###### What operating system are you using?

Windows 7 Ultimate 64bit
###### What did you do?

Trying to `slugify` strings with Swedish letters in them. Not behaving like I would expect (to remove them or change the letters).
Example string: `{{ 'kvalité, då, äta, öl' | slugify }}`
###### What did you expect to see?

`kvalit-d-ta-l`
OR
`kvalite-da-ata-ol`
###### What did you see instead?

`kvalité-då-äta-öl`

---

My workaround would be to replace/remove the single letters before slugify'ing, but I'm wondering if there's anything else I can do.
Example replace: `{{ 'kvalité, då, äta, öl' | replace: 'é', 'e' | replace: 'å', 'a' | replace: 'ä', 'a' | replace: 'ö', 'o' | slugify }}`

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least
three months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If you can still reproduce this error on the <code>3.1-stable</code> or <code>master</code> branch,
please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider building it first as a plugin. Jekyll 3 introduced
[hooks](http://jekyllrb.com/docs/plugins/#hooks) which provide convenient access points throughout
the Jekyll build pipeline whereby most needs can be fulfilled. If this is something that cannot be
built as a plugin, then please provide more information about why in order to keep this issue open.

Thank you for all your contributions.

yaitloutou:
I confirm this issue. see this [stackoverflow question](http://stackoverflow.com/questions/41941320/jekyll-encoding-name-of-category-special-characters).
since I generate categories pages and folder using a personalized local generator, and I'm not using the default categories generated ids in the permalink, I can get a similar  behavior as the one wanted by the OP. 

and I'm willing to implement a fix for this issue.

pnn:
@yaitloutou that would be very much appreciated!

yaitloutou:
after reading #129 and #782  where @parkr explain concerns about breaking URLs created with the current system. 
[...]

PR #4673, opened 2016-03-17
Fixes jekyll/jekyll issue #4623

jussikinnula:
There's an issue:
https://github.com/jekyll/jekyll/issues/4623

The problem with slugify not removing non-ASCII characters is that many web browsers and servers cannot handle Unicode characters on filenames and directories. Many Jekyll plugins produce content, which generate files or directories from slugs generated from titles.

The best solution would be to do natural conversion (for example à/ä/å to a, etc). But an intermediate solution is to strip out the characters we don't want to have on files or directory names.

There should be a separate issue to implement such functionality on Jekyll, however, it's very imperative to get this fix in since there's people that have issues on this particular slugify problem.

### DIFF
--- a/jekyll.gemspec
+++ b/jekyll.gemspec
@@ -36,4 +36,5 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('rouge', '~> 1.7')
   s.add_runtime_dependency('jekyll-sass-converter', '~> 1.0')
   s.add_runtime_dependency('jekyll-watch', '~> 1.1')
+  s.add_runtime_dependency('i18n', '~> 0.7')
 end
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -30,6 +30,7 @@ require 'liquid'
 require 'kramdown'
 require 'colorator'
 require 'i18n'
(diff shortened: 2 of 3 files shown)


jussikinnula:
Updated version, which does transliteration properly (reverts the intermediate hack). However, there's now dependency to i18n GEM, and also I needed to manually add available locales + select `:en` as the selected locale. It should be checked if this is how we want to do things, but this is now tested in practice that the slugs are as we want them to be.

envygeeks:
I'm :-1: on this.  This is broken as i18n is broken.  ä is not a and yes, all modern browsers support UTF-8 characters via idna and all operating systems support UTF-8 as a standard (actually it's the default for Linux, and I believe OS X too), we need to address this properly.

jussikinnula:
I'm closing the pull request, actually I found myself out that i18n is quite broken - and I'm also thinking that well, there's a definite need to have an URL safe method of slugify (which would work on web servers, which don't support Unicode characters properly, for example like Google's Firebase, Amazon's S3 buckets, etc). But it could be an alternative method for slugifying.

PR #4674, opened 2016-03-17
WIP: Move to raising custom errors

parkr:
GitHub Pages currently does some regexp to pluck out errors, which is tedious
and essentially means writing custom errors is a chess match of upgrades and specs.
By giving each error type its own error class, it should be much easier to produce
better errors down the road. Clients of Jekyll aren't just the executable!

### DIFF
--- a/script/spec
+++ b/script/spec
@@ -1,0 +1,2 @@
+#!/bin/sh
+bundle exec rspec --color --require spec_helper $@
--- a/spec/jekyll/configuration_spec.rb
+++ b/spec/jekyll/configuration_spec.rb
@@ -1,0 +1,37 @@
+RSpec.describe(Jekyll::Configuration) do
+  let(:config) { described_class.from({}) }
+
+  describe "#safe_load_file" do
+    it "throws an InvalidFileFormat on a bad extension" do
+      filename = "my_config.json"
+      expect(-> { config.safe_load_file(filename) }).to raise_error(
+        Jekyll::Errors::InvalidFileFormatError,
+        "No parser for '#{filename}' is available. Use a .toml or .y(a)ml file instead."
+      )
+    end
+  end
+
+  describe "#read_config_file" do
+    it "throws a FileNotFoundError if the file doesn't exist" do
+      filename = "_config.yml"
+      expect(-> { config.read_config_file(filename) }).to raise_error(
+        Jekyll::Errors::FileNotFoundError,
+        "The configuration file '#{filename}' could not be found."
+      )
+    end
+  end
+
+  describe "#check_config_is_hash!" do
+    it "does not throw if the input is a hash" do
+      expect(config.send(:check_config_is_hash!, {}, "_config.yml")).to eq(nil)
+    end
+
+    it "throws an InvalidConfigurationError if the input is not a hash" do
+      filename = "_config.yml"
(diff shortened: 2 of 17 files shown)


script/spec
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+bundle exec rspec --color --require spec_helper $@
```
envygeeks: .rspec -> '--color --require spec_helper`

spec/jekyll/configuration_spec.rb
```diff
@@ -0,0 +1,37 @@
+RSpec.describe(Jekyll::Configuration) do
+  let(:config) { described_class.from({}) }
+
+  describe "#safe_load_file" do
+    it "throws an InvalidFileFormat on a bad extension" do
+      filename = "my_config.json"
+      expect(-> { config.safe_load_file(filename) }).to raise_error(
```
envygeeks: `expec {  }`

spec/jekyll/configuration_spec.rb
```diff
@@ -0,0 +1,37 @@
+RSpec.describe(Jekyll::Configuration) do
+  let(:config) { described_class.from({}) }
+
+  describe "#safe_load_file" do
+    it "throws an InvalidFileFormat on a bad extension" do
```
envygeeks:
```ruby
context "when given a bad extension" do
  it "should throw InvalidFileFormat"
    # Do Work
  end
end
```

So that if the circumstance changes the test can resurface without effort.

spec/jekyll/configuration_spec.rb
```diff
@@ -0,0 +1,37 @@
+RSpec.describe(Jekyll::Configuration) do
+  let(:config) { described_class.from({}) }
```
envygeeks: This should be the subject.

spec/jekyll/errors_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'spec_helper'
+
+RSpec.describe(Jekyll::Errors) do
+  it "defines the base exception, FatalException" do
+    expect(described_class).to be_const_defined(:FatalException)
```
envygeeks: https://github.com/envygeeks/docker-template/blob/491decb0241d5277960d60be34cba58060772746/spec/support/matchers/have_const.rb

spec/spec_helper.rb
```diff
@@ -0,0 +1,66 @@
+ROOT = Pathname.new(File.expand_path("../", __dir__))
+
+require ROOT.join("lib", "jekyll.rb")
```
envygeeks: `require_relative "../lib/jekyll"`, however "jekyll" should already be in the path.

spec/jekyll/configuration_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  describe "#safe_load_file" do
+    it "throws an InvalidFileFormat on a bad extension" do
+      filename = "my_config.json"
+      expect(-> { config.safe_load_file(filename) }).to raise_error(
+        Jekyll::Errors::InvalidFileFormatError,
+        "No parser for '#{filename}' is available. Use a .toml or .y(a)ml file instead."
```
envygeeks: Never test the message unless you are sure it will remain static, this leads to coupled testing that is refactor-unfriendly and leads us down the slippery slope that is currently our coupled testing... with another library.

PR #4680, opened 2016-03-18
Add "ascii" mode to Utils.slugify

jussikinnula:
After careful thinking to solve problem when using web-servers which don't support Unicode characters in directory and file names, I came up with solution. We can use a new "urlsafe" mode for slugification, which just allows characters A-Z, a-z and 0-9 on the slug. Everything else is changed to "-".

The accompanying patch to use "urlsafe" on our case is here:
https://github.com/jussikinnula/jekyll-contentful/commit/24d1c854d27701a4e70649fc6802e6585108da3b

If we can get this first part accepted, I'll create secondary pull request.

More specifically the web-server problem on our case is when using Google Firebase or Amazon S3 to host the Jekyll static site, produced with the content from Contentful CMS service.

This would also mitigate the problem described on "Slugify a string doesn't seem to work on Unicode/Swedish letters #4623" issue.

I'm also open on reworking the problem with another solution, if there could be one.

### DIFF
--- a/lib/jekyll/utils.rb
+++ b/lib/jekyll/utils.rb
@@ -10,6 +10,7 @@ module Jekyll
     SLUGIFY_RAW_REGEXP = Regexp.new('\\s+').freeze
     SLUGIFY_DEFAULT_REGEXP = Regexp.new('[^[:alnum:]]+').freeze
     SLUGIFY_PRETTY_REGEXP = Regexp.new("[^[:alnum:]._~!$&'()+,;=@]+").freeze
+    SLUGIFY_URLSAFE_REGEXP = Regexp.new("[^[A-Za-z0-9]]+").freeze
 
     # Takes an indented string and removes the preceding spaces on each line
 
@@ -190,6 +191,11 @@ module Jekyll
           # "._~!$&'()+,;=@" is human readable (not URI-escaped) in URL
           # and is allowed in both extN and NTFS.
           SLUGIFY_PRETTY_REGEXP
+        when 'urlsafe'
+          # For web servers not being able to handle Unicode, the safe
+          # method is to ditch anything else but latin letters and numeric
+          # digits.
+          SLUGIFY_URLSAFE_REGEXP
         end
 
       # Strip according to the mode
--- a/lib/jekyll/utils.rb
+++ b/lib/jekyll/utils.rb
@@ -6,7 +6,7 @@ module Jekyll
     autoload :Ansi, "jekyll/utils/ansi"
 
     # Constants for use in #slugify
-    SLUGIFY_MODES = %w(raw default pretty)
+    SLUGIFY_MODES = %w(raw default pretty urlsafe)
     SLUGIFY_RAW_REGEXP = Regexp.new('\\s+').freeze
     SLUGIFY_DEFAULT_REGEXP = Regexp.new('[^[:alnum:]]+').freeze
     SLUGIFY_PRETTY_REGEXP = Regexp.new("[^[:alnum:]._~!$&'()+,;=@]+").freeze
--- a/lib/jekyll/utils.rb
+++ b/lib/jekyll/utils.rb
@@ -197,11 +197,11 @@ module Jekyll
           # "._~!$&'()+,;=@" is human readable (not URI-escaped) in URL
           # and is allowed in both extN and NTFS.
           SLUGIFY_PRETTY_REGEXP
-        when 'urlsafe'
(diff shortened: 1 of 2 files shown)


parkr:
Thanks for the pull request! It needs tests & documentation in order to be accepted. Also, what do you think about renaming `urlsafe` to `ascii`, as it only allows ascii alphanumerical values? Or `latin`? `urlsafe` wouldn't be "true" in the case where browsers do support non `a-zA-Z0-9` characters.

I think it has real potential.

envygeeks:
I'm against this one as well.  This does not solve the problem and will only exacerbate it when it's time to come and fix the problem for reals.  If we really want all this to happen then please just add `addressable` and do `Addressable::URI.parse(url).normalize` and return that value as it will not only encode the fqdn but it will also encode the path-part making it entirely safe.

```ruby
Addressable::URI.parse("https://Ä.com/oh look, here").normalize.to_s
# => "https://xn--4ca.com/oh%20look,%20here"
```

jussikinnula:
The problem isn't on this case on the browser side, but it's on the server side. We are using Firebase and Amazon S3 buckets to deploy our Jekyll sites - and those are not supporting Unicode-characters (it's no matter if you would encode the strings, the server just doesn't work with them). In a perfect world it would be just as envygeeks commented, but we're not living in a perfect world.

I would have to say that with "regular" HTTP servers and browser that Addressable::URI would be a perfect option, but generally it would not be even needed - as the current default slugify works also on those (there's only IE8 and older which would potentially need urlencoding).

I would be happy about renaming the "urlsafe" to "ascii", as it represents the "thing" better. If you can give a pointer to documentation where I need to fill in, I'm happy to contribute. The tests obviously need to be added, I'll work on those.

envygeeks:
Seems like a problem in Firebase to me, and they need to deal with that problem.

jussikinnula:
So, we ask Google and Amazon to fix their static delivery hosting options? :-) ... Even if they would start fixing something like this, it would take months - and I guess there always will be web servers & hosting services which don't support Unicode properly.

I added documentation in place (to the utils.rb), changed the naming from "urlsafe" to "ascii", and implemented a test (to test_utils.rb).

Updated dependent fix for our trouble is at:
https://github.com/jussikinnula/jekyll-contentful

But for the sake of improving also Google and Amazon static hosting options (which are very important in my opinion also for using Jekyll to built sites). I created a test case for Firebase (and filed a bug to them), please see https://github.com/jussikinnula/firebase-unicode-url-bug. And I'm going to do similar bug report to Amazon (for the problem on S3 filenames and directories).

I would like still to have the "best of both worlds" approach. I'm actually thinking that jekyll-contentful plug-in should have the ASCII fallback as an configurable option (I'm going to update my fork later to include also that).

jussikinnula:
I created a test site on Amazon S3 bucket (the test case is on GitHub at https://github.com/jussikinnula/amazon-s3-unicode-url-bug):

http://jussikinnula.s3-website-us-west-2.amazonaws.com/

I could not file myself a bug about it, but I'll get somebody from my employer to do so (I have only free tier personally, and I don't have IAM role set on my corporate Amazon account to get technical support).

envygeeks:
> So, we ask Google and Amazon to fix their static delivery hosting options? :-) ... 

You need me to ask? I'm happy to make a call.  The answer directly though, is yes.  It is their platform, it is their problem.  I understand the want to fix it here, but unless you take the problem to them too, you are solving nothing.

envygeeks:
Let me reiterate, I'm not rejecting the fix (or saying it shouldn't be implemented) I am saying that unless they are aware this is problem so they can fix it, our fix is worthless in the grand-scheme because there are hundreds of other static file generates that will create this problem.

jussikinnula:
OK, I got it :-) ... So I have notified Firebase on the issue, and I'll inform (here) when I've got the Amazon report filed as well. I guess the test cases should explain the issue, so basically if I have a standard nginx or expressjs (with serveStatic) site everything works beautifully with both of the links (the first link is unencoded, and the second one is using the "percent" url encodings).

So are we happy now with not forgetting any of the options? :-)

parkr:
This just happened for a user: https://github.com/jekyll/jekyll/issues/5327

parkr:
LGTM.

parkr:
Accepted due to the change: `ascii` is more accurate and less intrusive for doing url-safe things. When and if the time comes, we can add a `urlsafe` mode that uses addressible.

PR #4693, opened 2016-03-21
Fix #4689: Use SSLEnable instead of EnableSSL and make URL HTTPS.

### LINKED ISSUE #4689: Incorrect WEBrick settings for SSL
Jekyll 3.1.1

Fedora 23

###### What did you do?
jekyll server --ssl-cert test.crt --ssl-key test.key

Jekyll serving over https

[2016-03-21 11:08:35] ERROR bad URI `�h�b�\x00\x00\x16�+�/�'.

Using openssl's s_client, you can see that the server is not even responding to the ClientHello message, so it's not "speaking" SSL/TLS at all.

The issue is in serve.rb's `enable_ssl` method.  WEBrick is being configured with ":EnableSSL" but the correct option is [:SSLEnable](http://docs.ruby-lang.org/en/2.2.0/WEBrick.html#module-WEBrick-label-HTTPS).

Also, there is an error in `server_address` where the protocol is hard-coded to http where it should instead take into account the presence of the ssl-cert and ssl-key options and use https if they are present.

envygeeks:


envygeeks:
This is ready /cc @jekyll/core - Travis doesn't test non-master but it's tested locally on 2.3

parkr:
👍

PR #4694, opened 2016-03-21
Use String#encode for xml_escape filter instead of CGI.escapeHTML

pathawks:
Is there a reason to use `CGI.escapeHTML` over `encode`?

### DIFF
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -117,7 +117,7 @@ module Jekyll
     #
     # Returns the escaped String.
     def xml_escape(input)
-      CGI.escapeHTML(input.to_s)
+      input.to_s.encode(:xml => :attr).gsub(/\A"|"\Z/, "")
     end
 
     # CGI escape a string for use in a URL. Replaces any special characters
@@ -308,7 +308,7 @@ module Jekyll
     #
     # Returns a String representation of the object.
     def inspect(input)
-      CGI.escapeHTML(input.inspect)
+      xml_escape(input.inspect)
     end
 
     private
--- a/test/test_filters.rb
+++ b/test/test_filters.rb
@@ -394,6 +394,10 @@ class TestFilters < JekyllUnitTest
       should "return a HTML-escaped string representation of an object" do
         assert_equal "{&quot;&lt;a&gt;&quot;=&gt;1}", @filter.inspect({ "<a>" => 1 })
       end
+
+      should "quote strings" do
+        assert_equal "&quot;string&quot;", @filter.inspect("string")
+      end
     end
 
     context "slugify filter" do


parkr:
> Is there a reason to use CGI.escapeHTML over encode?

Questions of compatibility come into play, as well as "do they actually do the same thing?" I have never used `encode` seriously so we'll see what the tests say!

envygeeks:
The only problem here is that `encode` will wrap the string inside of a string.  So: `"hello & world".encode(:xml => :attr)` will become `"\"hello &amp; world\""`  I think by default to maintain compatibility it need use Text and accept an extra attribute to enable attribute quoting (though lets be honest, we already quote most of the time so that would be far and few between.)

pathawks:
Currently `xml_escape` *will* escape quotes as `&quot;`.

`encode(:xml => :attr)` *will* wrap the string in quotes, but will also escape quotes as `&quot;`. That is why I am deleting `"` from the result. (So `'hello & "world"'.encode(:xml => :attr)` will become `hello &amp; &quot;world&quot;`)

`encode(:xml => :text)` does not wrap the string in quotes, but does not escape quotes either.

Tests pass.

envygeeks:
Remove the `delete` and use `gsub` (`gsub(/\A"|"\Z/, "")`) so it speaks more to it's intention and I'm :+1: on it as it's completely backwards, I just have a problem with the `delete` this point because the code doesn't speak well to it's intentions to non-Rubyist's.

pathawks:
Oh no, I accidentally pushed **master** to my PR branch and closed the pull request! Git is hard.

Is there any way for you to reopen this pull, or should I just open a new one?

lib/jekyll/filters.rb
```diff
@@ -117,7 +117,7 @@ def date_to_rfc822(date)
     #
     # Returns the escaped String.
     def xml_escape(input)
-      CGI.escapeHTML(input.to_s)
+      input.to_s.encode(:xml => :attr).gsub(/\A"|"\Z/, "")
```
pathawks: @envygeeks :+1:

parkr:
Do we have tests for inspect? I'd expect:

```liquid
{{ page.layout | inspect }}
```

to output

```html
&quot;default&quot;
```

and the user would see

```text
"default"
```

on their site so they know it's a string.

pathawks:
> Do we have tests for inspect? 

[Here's](https://github.com/jekyll/jekyll/blob/a93c3eadd135ea7c1ebd066bf0e3f976fb8ea5b6/test/test_filters.rb#L393-L397) what I was able to find:

```ruby
    context "inspect filter" do
      should "return a HTML-escaped string representation of an object" do
        assert_equal "{&quot;&lt;a&gt;&quot;=&gt;1}", @filter.inspect({ "<a>" => 1 })
      end
    end
```

If you like, I could add a test specifically for strings:

```ruby
      should "quote strings" do
        assert_equal "&quot;string&quot;", @filter.inspect("string")
      end
```

I just tested this locally and it passes. :white_check_mark:

parkr:
@pathaws Yes, please add a test specifically for strings. The `gsub` worried me :smile:

test/test_filters.rb
```diff
@@ -394,6 +394,10 @@ def to_liquid
       should "return a HTML-escaped string representation of an object" do
         assert_equal "{&quot;&lt;a&gt;&quot;=&gt;1}", @filter.inspect({ "<a>" => 1 })
       end
+
+      should "quote strings" do
+        assert_equal "&quot;string&quot;", @filter.inspect("string")
```
pathawks: [It passes](https://travis-ci.org/jekyll/jekyll/builds/117584923) :+1:

envygeeks:
gsub has no idea that &quot; is `"` so if you send a quote and it escapes it then it'll pass.

parkr:
Sah weeet.

Thanks, @pathawks!

@jekyllbot: merge +minor

PR #4703, opened 2016-03-23
Add `:after_init` hook & add `Site#config=` to make resetting config easy

parkr:
@benbalter and I were working through https://github.com/github/pages-gem/pull/209
and realized that it'd be way nicer to have a one-time hook which could run after initialization
but before generators/converters/etc are initialized for the site. It would also be nice to have
a "safe" way to set `@config` knowing that it'd fan out properly.

This PR adds `:after_init` to the list of `:site` hooks and adds `Site#config=` to make all that much easier.

/cc @jekyll/stability @jekyll/ecosystem

### DIFF
--- a/lib/jekyll/hooks.rb
+++ b/lib/jekyll/hooks.rb
@@ -12,6 +12,7 @@ module Jekyll
     # initial empty hooks
     @registry = {
       :site => {
+        :after_init => [],
         :after_reset => [],
         :post_read => [],
         :pre_render => [],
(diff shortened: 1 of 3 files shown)


envygeeks:
I :heart: this but I have to ask, what is the difference between this and `:after_reset` semantically since they're both technically the same as far as usage goes (all our current plugins everywhere tap into `:after_reset` for the same purpose.)

That might be a debate for another day though.
:shipit:

parkr:
> what is the difference between this and `:after_reset` semantically

This plugin happens **once** per Jekyll process, whereas `:after_reset` happens for every single regeneration. `:after_init` happens **before** the call to `setup` and `reset`, which mean it can affect the `setup` process (affecting list of generators and converters).

parkr:
I want this for #4595.

@jekyllbot: merge +minor

PR #4706, opened 2016-03-24
Bold-italicize note in assets documentation about needing yaml front matter

yordis:
Just because developer are lazy and tools like this is for move forward faster, normally we don't read (it's a fact) and because of that I missed this super important sentence. At least this should help.


parkr:
👍 

@jekyllbot: merge +site

PR #4717, opened 2016-03-28
Don't blindly assume the last system when determining "open" cmd

envygeeks:
As it was we assumed that any system that wasn't Windows or OS X must be Linux
but the reality of that can be very unlikely. BSD is popular in some places and
it's not Linux and this would cause an error there.  If we do not know the
launcher for a platform we should ship an error and have the user file
a bug if they feel it necessary and skip the launch otherwise.


parkr:
👍 

@jekyllbot: merge +dev

Issue #4719, opened 2016-03-28
Move away from Deprecated features.


envygeeks:
- [ ] URI.escape - Deprecated - _Use `CGI.escape`_: See Notes
- [ ] URI.unescape - Deprecated - _Use `CGI.unescape`._
## Notes

There are problems with this move, in that Jekyll does a lot of non-standard stuff that most people would not do on a normal basis (or they would find a library that does it better.)
- `URI.escape` with a second argument - See: http://ruby-doc.org/stdlib-2.1.1/libdoc/uri/rdoc/URI/Escape.html#method-i-escape -- This normally un-standard behavior.  We might need to use a Regexp to solve this problem.

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least
three months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If you can still reproduce this error on the <code>3.1-stable</code> or <code>master</code> branch,
please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider building it first as a plugin. Jekyll 3 introduced
[hooks](http://jekyllrb.com/docs/plugins/#hooks) which provide convenient access points throughout
the Jekyll build pipeline whereby most needs can be fulfilled. If this is something that cannot be
built as a plugin, then please provide more information about why in order to keep this issue open.

Thank you for all your contributions.

pathawks:
Now that we are requiring **Addressable**, maybe we can use that instead

pathawks:
Fixed in #5726

PR #4720, opened 2016-03-28
Update Rake & disable warnings when running tests

envygeeks:
Update Rake and disable verbosity when running the specs because we have some
deprecated usage (for now -- however, see: jekyll/jekyll#4719) and because
Rouge, and Liquid throw out thousands (probably hyperbolic) of warnigns when the
specs are being ran.  We need upstream to fix their problems while we fix ours
or we'll all have a bad day, not that we aren't already.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source "https://rubygems.org"
 gemspec :name => "jekyll"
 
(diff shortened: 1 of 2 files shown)


parkr:
@jekyllbot: merge +dev

PR #4733, opened 2016-03-30
docs: fix wrong snippet in pagination page

adam8157:
page1 should be index page but not the previous one, and some other small amendments.

### DIFF
--- a/site/_docs/pagination.md
+++ b/site/_docs/pagination.md
@@ -157,11 +157,12 @@ title: My Blog
 {% for post in paginator.posts %}
   <h1><a href="{{ post.url }}">{{ post.title }}</a></h1>
   <p class="author">
-    <span class="date">{{ post.date }}</span>
+    <span class="date">{{ post.date | date: "%b %-d, %Y" }}</span>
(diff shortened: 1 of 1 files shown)


parkr:
This same suggestion has come up multiple times. [jekyll-paginate should handle this](https://github.com/jekyll/jekyll-paginate/blob/b5f06145d75b0fa823ee9e0ff5f1d27aecb3b671/lib/jekyll-paginate/pager.rb#L65). If you're having troubles with jekyll-paginate, we can investigate a bug on that repo.

adam8157:
My point is that example in docs page is wrong, not jekyll-paginate.

That snippet calls "previous_page_path", not the "paginate_path" you were talking. For example, on page16, the link of page1 is actually page15... obviously wrong.

parkr:
@adam8157 `jekyll-paginate` should output `/` or `/blog/` or whatever is your first page. Doing `"/"` is blind. If your site is producing an incorrect result, then it's probably a bug in `jekyll-paginate` and should be addressed.

PR #4734, opened 2016-03-30
Added an explicit rerun note to configuration.md, defaults section

nsgonultas:
I've copied note structure and content from **usage.md**. I've just changed the title of the note. But it's ok if a native speaker changes it with a better one.

I've confused while reading docs and trying to change my defaults. So I thought a little message would be good.


parkr:
I like it! Thank you! 

@jekyllbot: merge +site

PR #4741, opened 2016-03-31
DocumentDrop: add `#<=>` which sorts by date (falling back to path)

### LINKED ISSUE #4727: What is the default sort param when none specified?
Tried on Jekyll 3.1.1 & 3.1.2
Tried on Ruby 2.0.0 & 2.1.8
Tried on Windows 10 and Ubuntu 14.04.4

When constructing an array of posts, was getting the following error until I added 'date' as the sort parameter

```
comparison of Jekyll::Drops::DocumentDrop with Jekyll::Drops::DocumentDrop failed
```

Here's the relevant code that produces the error:

```
  {%      assign sorted_list = site.empty_array
%}{%      for topic in site.categories.topics
%}{%        if topic.categories.last == page.categories.last
%}{%          assign sorted_list = sorted_list | push: topic
%}{%        endif
%}{%      endfor
%}{%      assign sorted_list = sorted_list | sort
%}
```
(`empty_array` assigned to `[]` in config)

Fixed it by changing the last line:

```
assign sorted_list = sorted_list | sort: 'date'
```

The [Templates topic](https://jekyllrb.com/docs/templates/) in the jekyll doc says that a parameter to the sort filter is optional. I guess I was mislead in thinking that 'date' was the default, because elsewhere, if I try something like this, it works fine:
```
{% assign foo = site.posts | sort %}
{{ foo | inspect }}
```
Results:
```
[#<Jekyll::Document _posts/1996-10-11-data-abstraction-1.md collection=posts>, ... ]
```
Can the doc be made more specific as to what the default is, when a parameter is required, etc.? Or, if there should be a default, something's broken in this context.
I'm guessing the `push` to an array creates a completely different data structure than a site.posts. So, it would also be really cool to know how to _properly_ create a similar data structure with what looks to be references rather than objects, if that's possible.

More details [on stack overflow](http://stackoverflow.com/questions/36217943/comparison-of-jekylldropsdocumentdrop-with-jekylldropsdocumentdrop-faile?noredirect=1#comment60099452_36217943) if you need 'em.

parkr:
Fixes #4727.

### DIFF
--- a/lib/jekyll/drops/document_drop.rb
+++ b/lib/jekyll/drops/document_drop.rb
@@ -20,6 +20,13 @@ module Jekyll
         fallback_data['excerpt'].to_s
       end
 
+      def <=>(other)
+        return nil unless other.is_a? DocumentDrop
+        cmp = self['date'] <=> other['date']
+        cmp = self['path'] <=> other['path'] if cmp.nil? || cmp == 0
+        cmp
+      end
+
       private
       def_delegator :@obj, :data, :fallback_data
     end


envygeeks:
:shipit:

parkr:
@jekyllbot: merge +minor

Issue #3041, opened 2014-10-30
Proposal: multiple formats per page using layouts


Floppy:
As far as I can tell, it's not possible to have two output formats for a single page/post at the moment.

We have a case where we'd like to generate an iCal file for each post, as well as the HTML version. It seems to me that a nice way to do this would be to allow multiple extensions for each layout, and generate a file for each extension in the specified layout.

For instance, if we had a post `2014-10-30-hello-world.md`, and two layouts, `_layouts/post.html` and `_layouts/post.ical`, you would end up with output in both `/2014/10/30/hello-world.html` and `/2014/10/30/hello-world.ical`.

I'll have a go at this when I find time, but I was wondering if anyone could highlight any fundamental reasons why this wouldn't work, or problems that might be encountered when writing it?

ericmwalsh:
Is there a PR for this yet?

kbrock:
(I withdraw my request to allow multiple html files from one document)

Anyone able to share the current proposal?

- How did it know to generate multiple pages?
- Where is the extension determined? The layout's extension or the front matter of the post?

ericmwalsh:
Some thoughts on the proposal:

 - You should be able to choose to generate multiple pages in the config file, the layout file, and the page file (with rules of precedence applying)
 - Layouts can be listed out... ex.   layout: page-article page-article.json
 - For backward compatibility layouts listed without an extension (*.whatever) should default to HTML (or whichever user default is already specified)

dominikwilkowski:
:+1:

saclark:
:+1: I wrote a plugin a while back to accomplish this (https://github.com/saclark/jekyll-multipost) but standard support for this would be ten times better.

nfreear:
:+1:  for this! I really need HTML and JSON outputs for a project I'm currently on.

parkr:
Hey @Floppy, been thinking about this a lot and we should implement this for v3.2 or v3.3. If you have time, I was thinking we'd go this route:

1. Each page/post would define the output formats it wants **in addition to the one assigned by its source filename** (e.g. `html` for `index.md`). Front matter defaults would also apply, so doing this for a larger number of files would be trivial.
2. Those formats would use the same layout as for the HTML page, but with the new ending, e.g. `_layouts/docs.json` when `layout: docs` in the document.
3. The layout would define the behaviour. Whether you want JSON, CSV, XML, or some other output, the layout would be in charge of manipulating the data into the desired output.
[...]

PR #4742, opened 2016-03-31
Multi-format layouts [WIP]

Floppy:
This is a work-in-progress experiment having a go at supporting multiple output formats (i.e. HTML & JSON) for a layout. See #3041 for discussion of the idea.

I'm exploring Jekyll as I go, so this might take a while to come together, but I thought I'd open early for discussion and see where we end up.

/cc @parkr @pikesley


Floppy:
So far, this just deals with the loading of multiple layouts into the site layouts hash. If multiple layouts are present, only the first one (order undefined) is used in rendering.

lib/jekyll/renderer.rb
```diff
@@ -131,7 +131,9 @@ def invalid_layout?(layout)
     # Returns the content placed in the Liquid-rendered layouts
     def place_in_layouts(content, payload, info)
       output = content.dup
-      layout = site.layouts[document.data["layout"]]
+      # Load first layout from list, for now
+      layouts = site.layouts[document.data["layout"]]
+      layout = layouts ? layouts.first : nil
```
Floppy: I'd love use the Ruby 2.3 `&.` operator, but we can't do that yet I guess because of support for older rubies. Need to check if activesupport is available for `#try`, but I've not done that yet.

lib/jekyll/layout.rb
```diff
@@ -43,7 +45,11 @@ def initialize(site, base, name)
     #
     # Returns nothing.
     def process(name)
-      self.ext = File.extname(name)
+      self.mime_type = MIME::Types.find{|t| ".#{t.preferred_extension}" == File.extname(name)}
```
Floppy: What happens here when two MIME types have the same preferred extension? Does that happen? I bet it does somewhere...

Floppy: Also, `.html` vs `.htm`, and so on.

timwis: Perhaps an example: I've seen `application/vnd.geo+json` in both `.json` and `.geojson` files

lib/jekyll/renderer.rb
kbrock: When I don't have `try` or `&.` (forgot about that one) I tend to use `&&`

```ruby
layout = layouts && layouts.first
```

lib/jekyll/layout.rb
kbrock: Does this need to use `mime_type`?
Will this work just having a bunch of extensions?

(the question started out being rhetorical, but to be honest, I don't know)

Floppy: Future proofing, really, and avoiding a partial solution when a better one exists. If `mime_types` was a new dependency I'd think twice, but as it's already in the bundle, no harm in using it.

lib/jekyll/renderer.rb
Floppy: Thanks, good tip - I never quite internalised that construction...

lib/jekyll/renderer.rb
```diff
@@ -155,7 +157,7 @@ def place_in_layouts(content, payload, info)
           site.in_source_dir(layout.path)
         ) if document.write?
 
-        if layout = site.layouts[layout.data["layout"]]
+        if layout = site.layouts[layout.data["layout"]] ? site.layouts[layout.data["layout"]].first : nil
```
kbrock: Having trouble reading this line.
Maybe pull it out before the `if` block?

(but you don't need me to be able to read it - you need the person who is clicking merge to be able to read it)

Floppy: Good god, yes, that's horrible. What was I doing there?

kbrock:
I think mime-type is an added feature - keep it separate.

This PR is quite simple. And that is a good thing.
Does it get us 90% of the way there?

Floppy:
Sadly I didn't get very far with this. I might get chance to resurrect it, but I wouldn't hold your breath... so much else going on :(

kbrock:
agreed. my jekyll blog died of bit rot.
trying to find time to just put something back together again.

PR #4753, opened 2016-04-04
Configuration: allow users to specify a collections.posts.permalink directly

parkr:
…instead of overwriting with permalink.

If you have a config file with

```yaml
permalink: date
collections:
  posts:
    output: true
    permalink: /blog/:title/
```

then Post permalinks will match `permalink` (i.e. `"/:categories/:year/:month/:day/:title:output_ext"`) instead of `collections.posts.permalink` (i.e. `"/blog/:title/"`)

I'd like to ship a 3.0.4 for this and push it out to GitHub Pages.

/cc @benbalter @jekyll/gh-pages @jekyll/core

### DIFF
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -96,7 +96,7 @@ module Jekyll
     #
     # Returns the final configuration Hash.
     def configuration(override = Hash.new)
-      config = Configuration[Marshal.load(Marshal.dump(Configuration::DEFAULTS))]
+      config = Configuration[Configuration::DEFAULTS]
       override = Configuration[override].stringify_keys
       unless override.delete('skip_config_files')
         config = config.read_config_files(config.config_files(override))
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -96,7 +96,7 @@ module Jekyll
     #
     # Returns the final configuration Hash.
     def configuration(override = Hash.new)
-      config = Configuration[Configuration::DEFAULTS]
+      config = Configuration[Marshal.load(Marshal.dump(Configuration::DEFAULTS))]
       override = Configuration[override].stringify_keys
       unless override.delete('skip_config_files')
         config = config.read_config_files(config.config_files(override))
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -96,14 +96,13 @@ module Jekyll
     #
     # Returns the final configuration Hash.
     def configuration(override = Hash.new)
-      config = Configuration[Configuration::DEFAULTS]
-      override = Configuration[override].stringify_keys
+      config = Configuration.new
       unless override.delete('skip_config_files')
         config = config.read_config_files(config.config_files(override))
       end
 
       # Merge DEFAULTS < _config.yml < override
-      config = Utils.deep_merge_hashes(config, override).stringify_keys
+      config = Configuration.from Utils.deep_merge_hashes(config, override).stringify_keys
       set_timezone(config['timezone']) if config['timezone']
 
       config
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -82,6 +82,23 @@ module Jekyll
       }
     }].freeze
 
+    class << self
+      # Static: Produce a Configuration ready for use in a Site.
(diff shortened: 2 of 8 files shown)


benbalter:
So this would be per-collection permalinks?

parkr:
> So this would be per-collection permalinks?

@benbalter Precisely! `permalink` is _very_ broad and I believe we should phase it out in favor of collection-specific permalinks (not sure what to do about pages, but that's not entirely relevant to this PR).

[Since v3.0, collections have offered a `permalink` key](https://github.com/jekyll/jekyll/blob/v3.0.3/lib/jekyll/collection.rb#L190-L194) as a means of setting up the URL template for its constituent documents. In 3.0.3, we **set** it outright to the value of `permalink`. This PR simply **defaults** it instead, allowing 3.0.4 users to be more specific in their `permalink` selection.

benbalter:
Makes sense to me. That how I thought it worked, TBH. :smile:

parkr:
Turns out our tests are overwriting `Jekyll::Configuration::DEFAULTS`, which isn't good! Specifically, sub-hashes aren't being refreshed.

Fixed using a hack and pushed for now. Would be nice to have [`#deep_dup`](https://github.com/rails/rails/blob/4d8f62dcfa0a5157b3facbd71f75fc6639636347/activesupport/lib/active_support/core_ext/object/deep_dup.rb) one day.

hg:
LGTM

lib/jekyll/configuration.rb
```diff
@@ -177,6 +177,7 @@ def read_config_files(files)
 
       begin
         files.each do |config_file|
+          next if config_file.nil? or config_file.empty?
```
envygeeks: This is overly defensive while being non-defensive at the same time.  What I mean is it's better to just do `next unless config_file` instead of the empty because what happens if they send true? Or a symbol, or otherwise.  It might all still be toppling down.

lib/jekyll.rb
```diff
@@ -96,7 +96,7 @@ def env
     #
     # Returns the final configuration Hash.
     def configuration(override = Hash.new)
-      config = Configuration[Configuration::DEFAULTS]
+      config = Configuration[Marshal.load(Marshal.dump(Configuration::DEFAULTS))]
```
envygeeks: What is this trying to do? Why not just freeze the hash?

test/helper.rb
```diff
@@ -42,8 +42,30 @@ def jruby?
...
+  def dest_dir(*subdirs)
+    test_dir('dest', *subdirs)
+  end
+
+  def source_dir(*subdirs)
+    test_dir('source', *subdirs)
+  end
+
+  def test_dir(*subdirs)
+    File.join(File.dirname(__FILE__), *subdirs)
```
envygeeks: `File.join(__dir__, *subdirs)`

envygeeks:
Recently I was told not to Marshal for security reasons that were wild and overly broad for doing the exact same thing happening here, what makes this instance different and okay?

lib/jekyll.rb
parkr: It makes a deep duplicate of the defaults. The PR body describes the problem of modifying the defaults in some situations.

parkr:
> Recently I was told not to Marshal for security reasons that were wild and overly broad for doing the exact same thing happening here, what makes this instance different and okay?

I removed it from the runtime usage and now only using it to make a deep duplicate in the tests.

lib/jekyll/configuration.rb
parkr: If you send `{ "config" => "" }`, we try to load the file. That's not such a great thing. What's a better way which preserves some of the defensiveness?

parkr:
@envygeeks Removing the deep duplication causes this PR to fail CI. Know how we can make `deep_merge_hashes` duplicate properly?

envygeeks:
@parkr if you wish to avoid Marshal you can do:

```ruby
Hash[{ :hello => :world }.frozen.to_a].frozen? # => false
```

parkr:
@envygeeks This seems to work much more nicely. `DEFAULTS['collections'].object_id` was the same as `config['collections'].object_id` in `add_default_collections` so when we added or changed anything there, the defaults were also changed.

PR #4757, opened 2016-04-05
Serve options

keegoid:
Add --profile command option for `jekyll help serve`.

When I started playing with Jekyll, I ran this help command but couldn't find any info about the --profile option, so I added it in. Seems to me like it should be listed there, right?

I ran all the tests including `script/test test/test_commands_serve.rb.


parkr:
Thanks for your pull request! This is actually a problem in the command library, Mercenary. These options do exist and function as you'd expect, but aren't properly documented with `jekyll serve --help`.

Opened https://github.com/jekyll/mercenary/pull/42 to fix this.

PR #4759, opened 2016-04-05
Fixed the example fork url in CONTRIBUTING

clarkwinkelmann:
The `<your-username>` was breaking the plain text. I switched it to pre text.

Also removed the ending `/` as GitHub don't use it at the end of URLs.


envygeeks:
:heart:

PR #4765, opened 2016-04-07
Add roadmap for Jekyll

parkr:
This is something @benbalter suggested we start doing. This PR spikes out an incomplete roadmap. Before I flesh out all of it, I wanted to get some feedback. Is this a good idea?

By adding more structure to our communication about changes in Jekyll, we offer more assurances to the community at large that things coming down the pipeline (a) are going to happen and (b) communicates _when_ they're going to happen. Writing a paragraph for each minor and major bump provides transparency to users and contributors and forces us to consider the purpose & focus for each release.

Looping everyone into this. What do you think?

/cc @jekyll/core @jekyll/stability @jekyll/ecosystem @jekyll/documentation @jekyll/performance @jekyll/gh-pages @jekyll/windows
/cc @bkeepers

### DIFF
--- a/site/roadmap.md
+++ b/site/roadmap.md
@@ -7,11 +7,14 @@ This is a high-level roadmap for Jekyll. It is a list of things we're
 planning for each release. While plans are subject to change, this should
 offer a bit of an overview for anyone interested in future features.
 
-- [v3.2](#v3.2)
-- [v3.3](#v3.3)
+If you'd like to discuss any of the items, visit the link next to each
+item and leave us a comment! We'd love to hear your thoughts.
+
+- [v3.2](#v32)
+- [v3.3](#v33)
 
 ## v3.2
 
 For v3.2, we're interested in shipping:
 
-1. **Gem-based themes.** This will allow users to ship layouts, includes, and sass (to start with) in a way that is more easily consumable and versionable.
+1. **Gem-based themes.** This will allow users to ship layouts, includes, and sass (to start with) in a way that is more easily consumable and versionable. ([discuss]({{ site.repository }}/issues/4510))
--- a/site/roadmap.md
+++ b/site/roadmap.md
@@ -1,0 +1,17 @@
+---
+layout: page
+title: Roadmap
+---
+
+This is a high-level roadmap for Jekyll. It is a list of things we're
+planning for each release. While plans are subject to change, this should
+offer a bit of an overview for anyone interested in future features.
+
+- [v3.2](#v3.2)
+- [v3.3](#v3.3)
+
+## v3.2
+
+For v3.2, we're interested in shipping:
+
+1. **Gem-based themes.** This will allow users to ship layouts, includes, and sass (to start with) in a way that is more easily consumable and versionable.


mattr-:
:+1:

envygeeks:
:+1:

protesilaos:
+1

bkeepers:
I think it's great!

What do you think about linking each item in the roadmap to an issue where discussion can happen around that feature?

SuriyaaKudoIsc:
:+1:

KingFelix:
Good idea!

Il gio 7 apr 2016, 17:50 Suriyaa Kudo <notifications@github.com> ha scritto:

DirtyF:
Since there is no roadmap published yet, I thought I could ask here @jekyll/core if they have a vision of what to expect from Jekyll for the next minor and major versions.

DirtyF:
@jekyll/core Can we help update the roadmap? This seems to be left out of the release workflow for now. Should this be automatically generated based on milestones and labels?

parkr:
@DirtyF I didn't add this because it was pretty sparse and that bugged me haha. But feel free to submit a PR or edit this branch directly with your thoughts for Jekyll 3.4. Liquid and Rouge upgrades are possible, though not definite.

DirtyF:
@parkr So maybe we should simply forget about this document for now, as we don't seem to have vision on Jekyll's future and it must be handly written. I'd rather list only the features that @jekyll/core think are the most important ones (could be priority or more than three stars label). It's not up to me to decide what to include or not in the roadmap.

Except from bug fixes and dependencies bumps in minor versions, what I expect from a product Roadmap is where are we going, what is most important to achieve in order to add value to users. Pages as collections, i18n, plug external templates engines, export to different formats have all been discussed in previous issues but are they what matter the most right now to users? What should be the focus of the core team, it's what I'm wondering right now.

mattr-:
Since this is out of date, I'm going to close it. If there's a newer roadmap available, then let's open a new issue for that. 

Or maybe we just go without a roadmap?

DirtyF:
@mattr- IMHO going roadmapless is not helpful to state what matter the most (whether be it for current maintainers or for all the users)

mattr-:
Doing the roadmap thing for open source projects has always confused me a bit because so much of what we do is in response to what's come in from the community. If we have tickets that are open already that we want to target for a particular release, why not just put them in a milestone?

DirtyF:
@mattr-  that's what we try to do I guess (https://github.com/jekyll/jekyll/milestones), but a high-level roadmap like @parkr was proposing is a more focused vision/version of where the product is heading.

I find it nice to just have to read something like https://hugodocs.info/about/roadmap/ to know what the current priorities are.

pathawks:
> Doing the roadmap thing for open source projects has always confused me a bit because so much of what we do is in response to what's come in from the community.

Is Jekyll designed with a purpose, or will we just add small features one at a time as requested until feature bloat takes over? Or maybe just declare Jekyll feature complete and only fix bugs as we discover them.

* * *

I don't think we need to say "This is what the next version will do and this is when it will be released," but a set of goals I think would go a long way toward stating what we would like to become.

parkr:
I found coming up with a roadmap challenging to create (we don't always have a big feature to work on... do we list bugs?) and often fluctuating. If we used an `accepted` label or similar and used milestones, then it would be fairly easy to see what was planned for each release.

We could start a process whereby someone has to write a PR to update the roadmap and if it's accepted then it's worked on and scheduled. I think it's a somewhat unique way to handle open source, and since none of us can really work on this full-time then that work may never happen. The best thing we can do is encourage contributions. In my experience all of those contributions are from users who want a given feature and decided to write the code themselves. They submit a PR and, if it's accepted, then it goes into master. Putting something in our roadmap doesn't *do* anything because the maintainers don't really write lots of code (mostly bug fixes and documentation fixes that we have time for, maybe a feature here and there). So while it'd be nice to have a roadmap, I'm not sure we can consider it more than a "wishlist."

Issue #2540, opened 2014-06-24
Link to a document inside collection via post_url


svetlanapuro:
Can I link to a document inside collection via `post_url`?

I try to do it like this:

```
[Name of Link]({% post_url /_collection/start/01 %})
```

But I get the following error:

```
Liquid Exception: Could not parse name of post "/_collection/start/01" in tag 'post_url'.
Make sure the post exists and the name is correct.
```

iloveip:
Hello,

Has this been implemented already? I would also like to track it somewhere.

parkr:
> Has this been implemented already?

Not presently, no. There are many unanswered questions about behavior. If you would like to open a pull request, I would be happy to take a look. The code is in `lib/jekyll/tags/post_url.rb`.

jonathanthomas83:
Any news on being able to link to a document in a collection yet? Many thanks

parkr:
> Any news on being able to link to a document in a collection yet? Many thanks

@jonathanthomas83 No, we have not created anything and I haven't seen a PR yet. If you have some time, I'd gladly look at any PR you're interested in submitting!

parkr:
The `link` tag has been added in v3.2 which is the next release we expect to make. See the pull request: #4624.

PR #4777, opened 2016-04-09
Add collection support to the "post_url" tag

Mizumi:
These changes add very simplistic support for collections to the post_url tag, making it easier to link to files that reside within collections (in response to Jekyll issue #2540). ;D

Functionally, these changes cause the post_url tag to split the tag input parameter on all spaces. If the array generated by the split operation is exactly two elements long, the post_url tag will use the first element as the name of the collection and the second element as the document to find in the collection. If the array is not exactly two elements long, post_url will act as normal.

This is my first time playing around with the Jekyll source, so I'd be happy to have some feedback; I'm sure there's a better way to parse out multiple tag arguments, but split(" ") was the best I could find right now. It might also be reasonable to instead parse the incoming URL to see if it starts with a known collection name, and go from there.

### DIFF
--- a/lib/jekyll/tags/post_url.rb
+++ b/lib/jekyll/tags/post_url.rb
@@ -52,7 +52,16 @@ module Jekyll
     class PostUrl < Liquid::Tag
       def initialize(tag_name, post, tokens)
         super
-        @orig_post = post.strip
+        
+        # Check to see if the user specified a collection name.
+        params = post.strip.split(" ")
+        if params.length == 2
+            @collection = params[0]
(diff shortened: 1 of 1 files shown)


envygeeks:
Please add tests as this is unacceptable without tests. Thanks.

Mizumi:
Ok, cool. I'll write up the tests when I get some time.

parkr:
@Mizumi You can look forward to the `link` tag in 3.2! #4624

Mizumi:
Awesome! Totally missed that one, thanks. ;D

PR #4780, opened 2016-04-10
Update collections.md

lzl124631x:
The example used `-` together with tab for list, which makes me confused. I replaced the `-` notations with tabs for freshman like me.

### DIFF
--- a/site/_docs/collections.md
+++ b/site/_docs/collections.md
@@ -19,7 +19,7 @@ with the name of your collection:
 
 {% highlight yaml %}
 collections:
-- my_collection
+  my_collection
 {% endhighlight %}
 
 You can optionally specify metadata for your collection in the configuration:
@@ -34,11 +34,11 @@ Default attributes can also be set for a collection:
 
 {% highlight yaml %}
 defaults:
-  - scope:
-      path: ""
-      type: my_collection
-    values:
-      layout: page
+  scope:
+    path: ""
+    type: my_collection
+  values:
+    layout: page
 {% endhighlight %}
 
 ### Step 2: Add your content


lzl124631x:
@parkr  I am still confused! I experimented on <http://yaml-online-parser.appspot.com/>

```
collections:
- my_collection
```
is compiled into 

```
{
  "collections": [
    "my_collection"
  ]
}
```

But, 
```
collections:
  my_collection:
    foo: bar
```

is compiled into

```
{
  "collections": {
    "my_collection": {
      "foo": "bar"
    }
  }
}
```

If you put these yaml together

```
collections:
- my_collection

collections:
  my_collection:
    foo: bar
```

The second one just overwrites the first one, so it's

```
{
  "collections": {
    "my_collection": {
      "foo": "bar"
    }
  }
}
```

`collections` is not an array/list!

Is the following the correct one?

```
collections:
- my_collection:
    foo: bar
```

which is compiled into 

```
{
  "collections": [
    {
      "my_collection": {
        "foo": "bar"
      }
    }
  ]
}
```

parkr:
`collections` is a separate story. If you want a collection without any metadata, you can just have it as an empty key:

```yaml
collections:
  my_collection:
    foo: bar
  other_collection:
```

The `defaults` key _must_ be a list, but `collections` can be either a list or a hash (but not both).

If it makes it easier, you can use straight JSON inside of `_config.yml` since JSON is a subset of YAML.

lzl124631x:
@parkr  Thanks for your explanation!

As for
> The `defaults` key *must* be a list

I can refer to <https://jekyllrb.com/docs/configuration/>

> The `defaults` key holds an array of scope/values pairs that define what defaults should be set for a particular file path, and optionally, a file type in that path.

But I can't find a clue in <https://jekyllrb.com/docs/collections/> that shows

> `collections` can be either a list or a hash (but not both).

I think this would confuse newcomers, and maybe it should be add to <https://jekyllrb.com/docs/collections/>

PR #4781, opened 2016-04-10
Explain data file format requirements more clearly in documentation

kevee:
The doc read “These files must be YAML files,” then lists a few extensions that are not YAML.


parkr:
Excellent, thank you, @kevee!

❤️ 

@jekyllbot: merge +site

PR #4786, opened 2016-04-14
convertible: use Document::YAML_FRONT_MATTER_REGEXP to parse transformable files

parkr:
Replaces #4736, which was my commit in a PR but with some git conflict.

/cc @jekyll/build

### DIFF
--- a/lib/jekyll/convertible.rb
+++ b/lib/jekyll/convertible.rb
@@ -41,7 +41,7 @@ module Jekyll
       begin
         self.content = File.read(site.in_source_dir(base, name),
                                  Utils.merged_file_read_opts(site, opts))
-        if content =~ /\A(---\s*\n.*?\n?)^((---|\.\.\.)\s*$\n?)/m
+        if content =~ Document::YAML_FRONT_MATTER_REGEXP
           self.content = $POSTMATCH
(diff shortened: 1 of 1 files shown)


lib/jekyll/convertible.rb
```diff
@@ -41,7 +41,7 @@ def read_yaml(base, name, opts = {})
       begin
         self.content = File.read(site.in_source_dir(base, name),
                                  Utils.merged_file_read_opts(site, opts))
-        if content =~ /\A(---\s*\n.*?\n?)^((---|\.\.\.)\s*$\n?)/m
+        if content =~ Document::YAML_FRONT_MATTER_REGEXP
```
envygeeks: This is only used in Convertible, why is on Document if it's only used here?

envygeeks: Take that back.

parkr: You're OK with it then?

envygeeks: I see no problem.

envygeeks:
:shipit:

parkr:
😄 

@jekyllbot: merge +bug

PR #4789, opened 2016-04-14
Update Rack-Jekyll Heroku deployment blog post url

duksis:
Failed to deploy to heroku by following old blog post,
but after stumbling up on https://github.com/adaoraul/rack-jekyll/issues/20
managed to do it and would like to avoid the trouble for others.


parkr:
Thanks! The other one is very old.
@jekyllbot: merge +site

PR #4792, opened 2016-04-15
Globalize Jekyll's Filters.

envygeeks:
As it stands Jekyll does not globalize it's filters.  So anybody wishing to go
into Jekyll's context to process their own Liquid (say in a plugin) may be taken
aback when they find out that Jekyll's filters are not available.
See: jekyll/jekyll-assets#252.

/cc @jekyll/core @jekyll/stability

### DIFF
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -45,7 +45,6 @@ module Jekyll
   autoload :Errors,              'jekyll/errors'
   autoload :Excerpt,             'jekyll/excerpt'
   autoload :External,            'jekyll/external'
-  autoload :Filters,             'jekyll/filters'
   autoload :FrontmatterDefaults, 'jekyll/frontmatter_defaults'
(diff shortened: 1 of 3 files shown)


parkr:
@envygeeks Great idea! This seems to have broken the cucumber tests though... know what's going on there?

envygeeks:
@parkr looking into it now.

envygeeks:
Should be ready to :ship:  now.  I forgot to make the filters explicit instead of implicit.

pathawks:
Is this going to affect plugins that have been manually registering Jekyll’s filters?

envygeeks:
@pathawks it won't.  They're explicit re-inclusion will just overwrite with the same methods locally.

2474:
@envygeeks Thanks for doing this work. I appreciate your help.

parkr:
@jekyllbot: merge +minor

PR #4793, opened 2016-04-15
Example in the site template should be IANA-approved example.com

sl4m:
yourdomain(dot)com is a deceptive website. If a jekyll user does not change this in their config, they will default to this website and it appears in places like the RSS feed. I've just experienced this with another jekyll user's blog.

### DIFF
--- a/lib/site_template/_config.yml
+++ b/lib/site_template/_config.yml
@@ -19,7 +19,7 @@ description: > # this means to ignore newlines until "baseurl:"
   line in _config.yml. It will appear in your document head meta (for
   Google search results) and in your feed.xml site description.
 baseurl: "" # the subpath of your site, e.g. /blog
-url: "http://yourdomain.com" # the base hostname & protocol for your site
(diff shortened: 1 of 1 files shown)


envygeeks:
/cc @benbalter @parkr 
This should be merged and maybe Github should do a check?

sl4m:
I'm sure GitHub staff can do a much better job searching, but here's results I pulled from GitHub search. Public repositories with references to the bad domain: https://github.com/search?utf8=%E2%9C%93&q=http%3A%2F%2Fyourdomain.com+extension%3Ayml&type=Code&ref=advsearch&l=&l=

parkr:
What do you think about commenting it out to prevent this?

envygeeks:
IMO example.com might be the route to go since it's the IANA domain for example domains.  It's owned by IANA, is reserved by IANA and it (along with a dozen other domains) suite to serve as url's for apps to use without being taken (like the current example demonstrates.)

parkr:
Sounds good to me!

@jekyllbot: merge +bug

PR #4796, opened 2016-04-15
Comment out the example domain

chrisfinazzo:
Follow up to #4793.


parkr:
I just merged #4793, but thanks ❤️

PR #4804, opened 2016-04-17
Fixed typo on Configuration page

henrywright:
Add an apostrophe so that "projects" becomes a possessive noun instead of a plural.


parkr:
Thanks!

@jekyllbot: merge +site

PR #4808, opened 2016-04-18
3.2.x/master: Fix defaults for Documents (posts/collection docs)

### LINKED ISSUE #4458: Scope values not applied

parkr:
Ports #4806 forwards to the 3.2.x/`master` line.

It was using "url" to determine the path, which isn't how it should work, per the `default_proc` in the initializer.

Fixes #4458.

/cc @jekyll/build @jekyll/stability


parkr:
@jekyllbot: merge +bug

PR #4813, opened 2016-04-19
Added missing single quote on rsync client side command

joshaidan:
The rsync command was missing a single quote around the --rsh='ssh -p2222' parameter.


envygeeks:
:shipit:

parkr:
Thanks!

@jekyllbot: merge +site

PR #4822, opened 2016-04-23
Pages as a collection: Again!

alfredxing:
(Revamp/redo of #4412)

So, I finally had the time to sit down and work on this 😅 
I ended up starting from scratch since the original PR broke a lot of things.

I had to copy a lot of original stuff from the old Page class for compatibility, but it doesn't depend on Convertible anymore. There are also a couple weird/hacky/workaround pieces of code. I've commented on them in the diffs; let me know if there's a better way to do it!

All the tests pass, except for the test plugin. This is then arguably a breaking change, though I don't think there's anything we can do to avoid it. The situation is that the initialization process for Page changed (to comply with how Document handles things). The `Page.new` interface & signature is still the same; however, for custom hooks and plugins that use Page subclasses _and don't call `super`_, things won't work since the collection is not specified, among other things like filepath handling.

If custom subclasses call `super`, everything should work. I've changed the one example in `hooks.feature` as an example (and to make it pass).

Let me know what you think!

### DIFF
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -256,7 +256,7 @@ module Jekyll
     # values
     #
     # Returns nothing.
-    def read(opts = {})
+    def read(opts = {}, path: self.path)
       Jekyll.logger.debug "Reading:", relative_path
 
       if yaml_file?
--- a/lib/jekyll/drops/document_drop.rb
+++ b/lib/jekyll/drops/document_drop.rb
@@ -13,7 +13,7 @@ module Jekyll
       def_delegators :@obj, :name, :name
 
       def path
-        if (@obj.is_a? Page)
+        if @obj.is_a?(Page)
           @obj.path
         else
           @obj.relative_path
--- a/lib/jekyll/drops/document_drop.rb
+++ b/lib/jekyll/drops/document_drop.rb
@@ -9,8 +9,16 @@ module Jekyll
 
       def_delegator :@obj, :next_doc, :next
       def_delegator :@obj, :previous_doc, :previous
-      def_delegator :@obj, :relative_path, :path
       def_delegators :@obj, :id, :output, :content, :to_s, :relative_path, :url
+      def_delegators :@obj, :name, :name
+
+      def path
+        if @obj.is_a? Page
+          @obj.path
+        else
+          @obj.relative_path
+        end
+      end
 
       def collection
         @obj.collection.label
--- a/lib/jekyll/drops/document_drop.rb
+++ b/lib/jekyll/drops/document_drop.rb
@@ -13,7 +13,7 @@ module Jekyll
       def_delegators :@obj, :name, :name
 
       def path
-        if @obj.is_a? Page
+        if (@obj.is_a? Page)
           @obj.path
         else
           @obj.relative_path
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -46,7 +46,7 @@ module Jekyll
       })
 
       process(@name)
-      read(path: full_path)
+      read(:path => full_path)
     end
 
     # NOTE: COMPATIBILITY
@@ -190,7 +190,7 @@ module Jekyll
     # Returns the String value or nil if the property isn't included.
     def [](property)
       if ATTRIBUTES_FOR_LIQUID.include?(property)
-        send(property)
+        public_send(property)
       else
         data[property]
       end
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -1,16 +1,16 @@
 module Jekyll
-  class Page
-    include Convertible
-
+  class Page < Document
     attr_writer :dir
-    attr_accessor :site, :pager
+    attr_accessor :pager
     attr_accessor :name, :ext, :basename
-    attr_accessor :data, :content, :output
+    attr_writer :data
 
     alias_method :extname, :ext
 
     FORWARD_SLASH = '/'.freeze
 
+    # Compatibility
+    #
     # Attributes for Liquid templates
     ATTRIBUTES_FOR_LIQUID = %w(
       content
@@ -22,7 +22,6 @@ module Jekyll
 
     # A set of extensions that are considered HTML or HTML-like so we
     # should not alter them,  this includes .xhtml through XHTM5.
-
     HTML_EXTENSIONS = %W(
       .html
       .xhtml
@@ -36,22 +35,22 @@ module Jekyll
     # dir  - The String path between the source and the file.
     # name - The String filename of the file.
     def initialize(site, base, dir, name)
-      @site = site
       @base = base
-      @dir  = dir
+      @dir = dir
       @name = name
-      @path = site.in_source_dir(base, dir, name)
 
-      process(name)
-      read_yaml(File.join(base, dir), name)
+      full_path = site.in_source_dir(base, dir, name)
+      super(full_path, {
+        :site => site,
+        :collection => site.pages
+      })
 
-      data.default_proc = proc do |_, key|
-        site.frontmatter_defaults.find(File.join(dir, name), type, key)
-      end
-
-      Jekyll::Hooks.trigger :pages, :post_init, self
+      process(@name)
+      read(path: full_path)
     end
 
+    # Compatibility
+    #
     # The generated directory into which the page will be placed
     # upon generation. This is derived from the permalink or, if
     # permalink is absent, we be '/'
@@ -66,18 +65,10 @@ module Jekyll
       end
     end
 
-    # The full path and filename of the post. Defined in the YAML of the post
-    # body.
-    #
-    # Returns the String permalink or nil if none has been set.
-    def permalink
-      data.nil? ? nil : data['permalink']
-    end
-
     # The template of the permalink.
     #
     # Returns the template String.
-    def template
+    def url_template
       if !html?
         "/:path/:basename:output_ext"
       elsif index?
@@ -85,17 +76,6 @@ module Jekyll
       else
         Utils.add_permalink_suffix("/:path/:basename", site.permalink_style)
       end
-    end
-
-    # The generated relative url of this page. e.g. /about.html.
-    #
-    # Returns the String url.
-    def url
-      @url ||= URL.new({
-        :template => template,
-        :placeholders => url_placeholders,
-        :permalink => permalink
-      }).to_s
     end
 
     # Returns a hash of URL placeholder names (as symbols) mapping to the
@@ -116,21 +96,21 @@ module Jekyll
     def process(name)
       self.ext = File.extname(name)
       self.basename = name[0..-ext.length - 1]
+
+      # Invalidate URL
+      @url = nil
+      url
     end
 
-    # Add any necessary layouts to this post
+    # Compatibility
     #
-    # layouts      - The Hash of {"name" => "layout"}.
-    # site_payload - The site payload Hash.
-    #
-    # Returns nothing.
+    # Render the page
     def render(layouts, site_payload)
-      site_payload["page"] = to_liquid
-      site_payload["paginator"] = pager.to_liquid
-
-      do_layout(site_payload, layouts)
+      self.output = Jekyll::Renderer.new(@site, self, site_payload).run
     end
 
+    # Compatibility
+    #
     # The path to the source file
     #
     # Returns the path to the source file
@@ -138,11 +118,15 @@ module Jekyll
       data.fetch('path') { relative_path.sub(/\A\//, '') }
     end
 
+    # Compatiblity
+    #
     # The path to the page source file, relative to the site source
     def relative_path
       File.join(*[@dir, @name].map(&:to_s).reject(&:empty?))
     end
 
+    # Compatibility
+    #
     # Obtain destination path.
     #
     # dest - The String path to the destination dir.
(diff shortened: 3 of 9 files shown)


lib/jekyll/document.rb
```diff
@@ -256,7 +256,7 @@ def published?
     # values
     #
     # Returns nothing.
-    def read(opts = {})
+    def read(opts = {}, path: self.path)
```
alfredxing: Hacky code 1: I needed to add this optional parameter since `Page.path` is actually the relative path, while `read` needs the full path. So when `read` is called in the Page constructor, I pass in the full path.

lib/jekyll/drops/document_drop.rb
```diff
@@ -9,8 +9,16 @@ class DocumentDrop < Drop
...
       def_delegators :@obj, :id, :output, :content, :to_s, :relative_path, :url
+      def_delegators :@obj, :name, :name
+
+      def path
+        if @obj.is_a? Page
+          @obj.path
+        else
+          @obj.relative_path
+        end
+      end
```
alfredxing: *Hacky code 2*: Pretty self-explanatory: `{{ page.path }}` should go to `path` for Pages, and `relative_path` for everything else. Weird thing is, `Page#path` is basically just the relative path, but overridable and without leading slashes.

lib/jekyll/page.rb
```diff
@@ -1,16 +1,16 @@
...
+    attr_accessor :pager
     attr_accessor :name, :ext, :basename
-    attr_accessor :data, :content, :output
+    attr_writer :data
 
     alias_method :extname, :ext
 
     FORWARD_SLASH = '/'.freeze
 
+    # Compatibility
```
alfredxing: I've added "Compatibility" flags to methods that are mainly there for compatibility reasons (i.e. removing them shouldn't break core features, only method calls, etc.)

lib/jekyll/page.rb
```diff
@@ -116,33 +96,37 @@ def url_placeholders
     def process(name)
       self.ext = File.extname(name)
       self.basename = name[0..-ext.length - 1]
+
+      # Invalidate URL
+      @url = nil
+      url
```
alfredxing: *Hacky code 3*: This one isn't as bad but since the URL depends on the name, we need to invalidate the url when it changes...

lib/jekyll/page.rb
```diff
@@ -174,8 +153,39 @@ def trigger_hooks(hook_name, *args)
...
+    # Always returns true.
+    def render_with_liquid?
       true
     end
+
+    # Don't generate a title
+    def post_read
+      original = data['title']
+      super
+      data['title'] = original
```
alfredxing: *Hacky code 4*: Yeah...

lib/jekyll/page.rb
```diff
@@ -116,33 +96,37 @@ def url_placeholders
     def process(name)
       self.ext = File.extname(name)
       self.basename = name[0..-ext.length - 1]
```
envygeeks: * Please use `File.basename`

lib/jekyll/page.rb
```diff
@@ -66,18 +65,10 @@ def dir
...
-    # Returns the String permalink or nil if none has been set.
-    def permalink
-      data.nil? ? nil : data['permalink']
-    end
-
     # The template of the permalink.
     #
     # Returns the template String.
-    def template
+    def url_template
```
envygeeks: * This is a breaking change, please consider reverting this name change.  
* This change will impact what milestone it can be used at, it breaks the API.

lib/jekyll/page.rb
```diff
@@ -36,22 +35,22 @@ class Page
...
-      end
+      full_path = site.in_source_dir(base, dir, name)
+      super(full_path, {
+        :site => site,
+        :collection => site.pages
+      })
 
-      Jekyll::Hooks.trigger :pages, :post_init, self
+      process(@name)
+      read(path: full_path)
```
envygeeks: * Please use hash rockets, `=>` this provides consistency.

lib/jekyll/page.rb
envygeeks: * Please use: `NOTE: COMPATIBILITY`, this makes it easy for people to find a list of what you consider "compatible"

lib/jekyll/drops/document_drop.rb
envygeeks: * Please use () in all `if` statements.  Never skip the () in an if statement as it can be ambiguous to a human.

lib/jekyll/page.rb
```diff
@@ -36,22 +35,22 @@ class Page
     # dir  - The String path between the source and the file.
     # name - The String filename of the file.
     def initialize(site, base, dir, name)
-      @site = site
       @base = base
-      @dir  = dir
+      @dir = dir
```
envygeeks: * Please align your setters or do a wave effect see Example(1) and Example(2)

```ruby
@var1        = 1
@longvar2 = 2
@var3        = 3
```

```ruby
@var1 = 1
@longvar2 = 2
@var3 = 3
```

lib/jekyll/page.rb
```diff
@@ -116,33 +96,37 @@ def url_placeholders
...
+      self.output = Jekyll::Renderer.new(@site, self, site_payload).run
     end
 
+    # Compatibility
+    #
     # The path to the source file
     #
     # Returns the path to the source file
     def path
       data.fetch('path') { relative_path.sub(/\A\//, '') }
```
envygeeks: * Please do not single line block single line methods.  Please see Example(1)

```ruby
def path
  data.fetch("path") do
    relative_path(/\A\//, "")
  end
end
```

lib/jekyll/page.rb
```diff
@@ -155,11 +139,6 @@ def destination(dest)
       path
     end
 
-    # Returns the object as a debug String.
-    def inspect
```
envygeeks: * WARNING: Jekyll holds very larges objects, removing a custom inspect can cause a large amount of data in REPL's.

lib/jekyll/page.rb
envygeeks: * I crashed, I'm sorry.

lib/jekyll/page.rb
```diff
@@ -174,8 +153,39 @@ def trigger_hooks(hook_name, *args)
...
+
+    # Compatibility
+    #
+    # Accessor for data properties by Liquid.
+    #
+    # property - The String name of the property to retrieve.
+    #
+    # Returns the String value or nil if the property isn't included.
+    def [](property)
+      if self.class::ATTRIBUTES_FOR_LIQUID.include?(property)
```
envygeeks: * `self.class` should be unnecessary.  If it is there might be a problem, please prefer cleaner pathing.

envygeeks:
Hi @alfredxing, thanks for submitting a pull request, it was great reading your source.  Just a few comments, however, don't let that get you down.

envygeeks:
The bot feels like it was taking the mick, it was too nice this time and it didn't do a full review.  At least it wasn't taking shots at you though!  I'm excited for this pull though!

alfredxing:
Haha, I'd really appreciate comments from you the human as well 😁 
I'll implement the changes the bot suggested tomorrow!

lib/jekyll/page.rb
alfredxing: I'm using keyword arguments here, so rockets don't work (those are for hashes I think?)

envygeeks: @alfredxing keywords are almost literally hash arguments in Ruby, so you can do either when sending a message to the method, but when you are creating the signature it needs to be `def method(kwd: :val)`, however, like mentioned, when sending a message it can be `method(:kwd => :other_val)`

lib/jekyll/drops/document_drop.rb
```diff
@@ -9,8 +9,16 @@ class DocumentDrop < Drop
 
       def_delegator :@obj, :next_doc, :next
       def_delegator :@obj, :previous_doc, :previous
-      def_delegator :@obj, :relative_path, :path
       def_delegators :@obj, :id, :output, :content, :to_s, :relative_path, :url
+      def_delegators :@obj, :name, :name
+
+      def path
+        if (@obj.is_a? Page)
```
envygeeks:
```
if @obj.is_a?(Page)
```

lib/jekyll/page.rb
```diff
@@ -36,22 +35,22 @@ class Page
...
-      data.default_proc = proc do |_, key|
-        site.frontmatter_defaults.find(File.join(dir, name), type, key)
-      end
+      full_path = site.in_source_dir(base, dir, name)
+      super(full_path, {
+        :site => site,
+        :collection => site.pages
+      })
 
-      Jekyll::Hooks.trigger :pages, :post_init, self
```
envygeeks: I see you removed the hook here, is it added back in elsewhere? The removal of this might cause it's milestone to shift later than you might want because removing this is a major change that will disrupt users.

lib/jekyll/page.rb
```diff
@@ -87,17 +78,6 @@ def template
...
-    #
-    # Returns the String url.
-    def url
-      @url ||= URL.new({
-        :template => template,
-        :placeholders => url_placeholders,
-        :permalink => permalink
-      }).to_s
-    end
-
```
envygeeks: Is there a way to avoid this API change? The removal or change of these methods will disrupt the stable API for 3.x.  It might not be possible and if it's not is there a way to encapsulate the old into the new with a deprecation message for 4.x?

lib/jekyll/page.rb
```diff
@@ -174,8 +162,39 @@ def trigger_hooks(hook_name, *args)
...
+    # NOTE: COMPATIBILITY
+    #
+    # Accessor for data properties by Liquid.
+    #
+    # property - The String name of the property to retrieve.
+    #
+    # Returns the String value or nil if the property isn't included.
+    def [](property)
+      if ATTRIBUTES_FOR_LIQUID.include?(property)
+        send(property)
```
envygeeks: Even though it matters little since you are already guarding, the double guard (`public_send`) would be great!

lib/jekyll/page.rb
alfredxing: Oh, didn't know that! Thanks :+1:
I had tried `method(kwd => :other_val)`.

lib/jekyll/page.rb
alfredxing: Yeah, Document calls `trigger_hooks(:post_init)`. I've overridden `trigger_hooks` for Page, which triggers it on `:pages`

lib/jekyll/page.rb
alfredxing: As far as I'm aware, Document (superclass) has the exact same `url` method (apart from renaming `template` to `url_template`), hence the removal here.

lib/jekyll/page.rb
alfredxing: :+1: I just blindly copied this over from the original `Page`, but now would be a good time to make these adjustments!

parkr:
Hmmmmmmmmm we need to avoid breaking changes until v4.0. What of this can be accomplished without breaking changes? I could see much of this working out fine. What, to you, @alfredxing, are the main sticking points that necessitate a breaking change?

alfredxing:
@parkr I'm actually just taking another look at the code right now!

From what I remember the main thing was that plugins or hooks that subclass Page would be incompatible since some of them override the constructor and thus don't set the required properties.

alfredxing:
@parkr Any thoughts on this? Is this something we can tolerate or do we need to wait for 4.0? 😛

parkr:
> plugins or hooks that subclass Page would be incompatible since some of them override the constructor and thus don't set the required properties.

@alfredxing Are we assuming they aren't using `super()`?

I think we should be able to get around the `path` stuff through some other twisting and bending of pre-existing parameters. For example, `Collection#relative_directory` could easily be modified to be simply `""` and allow the page path to stand on its own.

Perhaps what we need here, @alfredxing, is simply a `Collection` as a container (rather than `Array`) and `Page`s as `Document`s. The pages feature doesn't benefit from `Collection#read` at all, so that can be ignored and a `PageReader` can be put in its place to ensure pages are read properly, as now.

I'll re-read the code here. Ultimately, we can't break anything until a (potential) v4.0. If there is a way to take advantage of collections without causing massive amounts of breakage, then we should do it. If you think we will not be able to use even part of collections easily to unify some of the code, then we should maybe hop on a call and see if there is something more systemically wrong.

parkr:
> I don't like it either but I can't think of a way around this...

Any easy way around this is to move to helper methods for these properties. In the helper method, we could default to a value, or raise a useful error saying a plugin didn't set the right property or something. It seems like the quick fix would be to use `super` instead of copying the initializer.

This is why monkey-patching is the worst. 😄

alfredxing:
@parkr I'll have to take a look at the code again too, but as far as I remember the only breaking change was the plugin use-case, with that `super` call... so I was wondering if you knew of any way to fix this in Ruby?

parkr:
> so I was wondering if you knew of any way to fix this in Ruby?

@alfredxing This does not have a fix, sadly. The only way we could make this work is to take a variable number of arguments in `Document#initialize` and construct one or the other based on this. In terms of instance variables not set, we'd just say "you're supposed to use `super`." They are, after all, supposed to use `super`. Even though I didn't in our test gem 😦

alfredxing:
Some context from what I can remember: I wrote this PR with full compatibility in mind. There is some hacky code as a result of this. If we're willing to write more sematic code but break more API in 4.0, that is probably the better option.

The only soft dependency #4269 has on this is that it would be nice if Pages were Documents too so the incremental regen stuff could all just live in one place...

pathawks:
@alfredxing Thank you so much for all the work that you did on this.

I am going to focus my efforts on making #5211 mergable, only because bringing that PR up-to-date with `master` resulted in fewer failing tests than when I tried to do the same with this PR. At any rate, hopefully soon Pages will be Documents and everything will be right with the world.

PR #4839, opened 2016-04-25
Add a Gemfile to the site template.

envygeeks:
This adds a Gemfile to the site template and sets the `ruby` to the currently running Ruby version.  This will "cheat the system" in that if a user is using an older version of Ruby, it will install an older version of dependencies.  Which according to: https://github.com/bundler/bundler-features/issues/120#issuecomment-214514847 should in theory work.  Only @indirect can confirm or deny this will actually happen and perhaps let us know if this is currently available within stable.

### DIFF
--- a/lib/site_template/Gemfile
+++ b/lib/site_template/Gemfile
@@ -1,0 +1,3 @@
+source "https://rubygems.org"
+ruby RUBY_VERSION
+gem "jekyll"


e2:
Given a simple `Gemfile` on Ruby 1.9.3 :

```ruby
source 'https://rubygems.org'
gem 'guard'
```

And a prerelease of Bundler:

```
$ gem install bundler --pre                                                    
Successfully installed bundler-1.12.0.rc.4
1 gem installed
```


When I run:

```
$ bundle update
Fetching gem metadata from https://rubygems.org/
Fetching version metadata from https://rubygems.org/
Fetching dependency metadata from https://rubygems.org/
Resolving dependencies...
Using coderay 1.1.1
(...)
Installing listen 3.0.6
Installing guard 2.13.0
Bundle updated!
```

It correctly installs Listen 3.0.6 on Ruby 1.9.3 !!

parkr:
@envygeeks This is actually already done, but via [`new.rb`](https://github.com/jekyll/jekyll/blob/f535497d90a0441d392338b3cbde46ffc369f465/lib/jekyll/commands/new.rb#L67-L90). Want to add it there instead?

envygeeks:
@parkr, I will! Give me a few minutes.

pathawks:
Where is `RUBY_VERSION` typically defined?

When I `cd` into the directory, RVM tells me `Unknown ruby interpreter version (do not know how to handle): RUBY_VERSION.`

envygeeks:
`RUBY_VERSION` is a system variable, it's always defined, even on JRuby.  The lack of RVM being able to handle that is actually an RVM bug, because that trick is the easiest way to work around common problems on systems.  Even here we shim it out on build.

envygeeks:
And by "system" I mean Ruby.  I guess I should have said Ruby, not system.

PR #4848, opened 2016-04-27
Add 'jekyll new-theme' command to help users get up and running creating a theme

parkr:
This command allows users to create new themes _very_ easily:

```text
$ jekyll new-theme THEME_NAME
$ cd THEME_NAME
$ # add files
$ # commit files
$ bundle exec rake release
```

Just saving my place for now.

### DIFF
--- a/lib/jekyll/theme_builder.rb
+++ b/lib/jekyll/theme_builder.rb
@@ -72,7 +72,10 @@ class Jekyll::ThemeBuilder
     %w(example/_config.yml example/index.html example/style.scss).each do |filename|
       write_file(filename, template(filename))
     end
-    write_file("example/_posts/#{Time.now.strftime("%Y-%m-%d")}-my-example-post.md", template("example/_post.md"))
+    write_file(
+      "example/_posts/#{Time.now.strftime("%Y-%m-%d")}-my-example-post.md",
+      template("example/_post.md")
+    )
   end
 
   def initialize_git_repo
--- a/lib/jekyll/theme_builder.rb
+++ b/lib/jekyll/theme_builder.rb
@@ -1,12 +1,12 @@
 class Jekyll::ThemeBuilder
-  SCAFFOLD_DIRECTORIES = %w{
+  SCAFFOLD_DIRECTORIES = %w(
     _layouts _includes _sass
-  }.freeze
+  ).freeze
 
   attr_reader :name, :path
 
   def initialize(theme_name)
-    @name = theme_name.to_s.gsub(/ /, "_").gsub(/_+/, "_")
+    @name = theme_name.to_s.tr(" ", "_").gsub(/_+/, "_")
     @path = Pathname.new(File.expand_path(name, Dir.pwd))
   end
 
@@ -26,10 +26,8 @@ class Jekyll::ThemeBuilder
   def template_file(filename)
     [
       root.join("theme_template", "#{filename}.erb"),
-      root.join("theme_template", "#{filename}")
-    ].find do |pathname|
-      pathname.exist?
-    end
+      root.join("theme_template", filename.to_s)
+    ].find(&:exist?)
   end
 
   def template(filename)
@@ -43,14 +41,14 @@ class Jekyll::ThemeBuilder
   def mkdir_p(directories)
     Array(directories).each do |directory|
       full_path = path.join(directory)
-      Jekyll.logger.info "create", "#{full_path}"
+      Jekyll.logger.info "create", full_path.to_s
       FileUtils.mkdir_p(full_path)
     end
   end
 
   def write_file(filename, contents)
     full_path = path.join(filename)
-    Jekyll.logger.info "create", "#{full_path}"
+    Jekyll.logger.info "create", full_path.to_s
     File.write(full_path, contents)
   end
 
@@ -64,13 +62,13 @@ class Jekyll::ThemeBuilder
   end
 
   def create_accessories
-    %w{README.md CODE_OF_CONDUCT.md LICENSE.txt}.each do |filename|
+    %w(README.md CODE_OF_CONDUCT.md LICENSE.txt).each do |filename|
       write_file(filename, template(filename))
     end
   end
 
   def initialize_git_repo
-    Jekyll.logger.info "initialize", "#{path.join(".git")}"
+    Jekyll.logger.info "initialize", path.join(".git").to_s
     Dir.chdir(path.to_s) { `git init` }
   end
 
--- a/lib/jekyll/theme_builder.rb
+++ b/lib/jekyll/theme_builder.rb
@@ -1,6 +1,6 @@
 class Jekyll::ThemeBuilder
   SCAFFOLD_DIRECTORIES = %w(
-    _layouts _includes _sass
+    _layouts _includes _sass example example/_posts
   ).freeze
 
   attr_reader :name, :path
@@ -14,6 +14,7 @@ class Jekyll::ThemeBuilder
     create_directories
     create_gemspec
     create_accessories
+    create_example_site
     initialize_git_repo
   end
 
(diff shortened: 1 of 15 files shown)


benbalter:
This may be a terrible idea, but what about `jekyll new --theme`? That way, down the line, `jekyll new --theme [THEME]` can create a new site with that theme, and `jekyll new --theme` can create a template site.

Alternatively, borrowing from `rails g`, `jekyll new site` (with site being implied when omitted) and `jekyll new theme`?

parkr:
@benbalter the reason I don't want to nest it is that it makes it harder to find with just `jekyll --help`. In the spirit of `bundle gem <name>`, we could do `jekyll theme <name>`.

envygeeks:
I like @benbalter's `jekyll new --theme` too but I also agree with you @parkr, what about if `jekyll new --theme` just routed internally to `jekyll new-theme` because it would be intuitive for people to also try `jekyll new --theme` (I know I would try that first without even looking at the docs.)

DirtyF:
Do you think there could be more options related to theme management?
If so, you may want to consider a dedicated `theme` command?

```
jekyll theme --new THEME_NAME #create new theme
jekyll theme --list #list all installed themes
jekyll theme --install <url> #install a theme from a distant URL (gem or github repo)
jekyll theme --uninstall THEME_NAME
```

lib/jekyll/theme_builder.rb
```diff
@@ -0,0 +1,108 @@
...
+    create_directories
+    create_gemspec
+    create_accessories
+    initialize_git_repo
+  end
+
+  private
+
+  def root
+    @root ||= Pathname.new(File.expand_path("../", __dir__))
```
envygeeks: `@root ||= Pathutil.new("../").expand_path(__dir__)` you should be able to do the same thing with Pathname.

lib/jekyll/theme_builder.rb
```diff
@@ -0,0 +1,108 @@
+class Jekyll::ThemeBuilder
+  SCAFFOLD_DIRECTORIES = %w{
+    _layouts _includes _sass
+  }.freeze
+
+  attr_reader :name, :path
+
+  def initialize(theme_name)
+    @name = theme_name.to_s.gsub(/ /, "_").gsub(/_+/, "_")
+    @path = Pathname.new(File.expand_path(name, Dir.pwd))
```
envygeeks: `@path = Pathutil.pwd.join(name)`

lib/jekyll/theme_builder.rb
```diff
@@ -0,0 +1,108 @@
...
+
+  def erb
+    @erb ||= ERBRenderer.new(self)
+  end
+
+  def mkdir_p(directories)
+    Array(directories).each do |directory|
+      full_path = path.join(directory)
+      Jekyll.logger.info "create", "#{full_path}"
+      FileUtils.mkdir_p(full_path)
```
envygeeks: If you convert `@path` to Pathutil you can simply do `path.mkdir_p` and it will do the rest for you.

lib/jekyll/theme_builder.rb
```diff
@@ -0,0 +1,108 @@
...
+    end
+  end
+
+  def write_file(filename, contents)
+    full_path = path.join(filename)
+    Jekyll.logger.info "create", "#{full_path}"
+    File.write(full_path, contents)
```
envygeeks: This is a good candidate for `Pathutil#write` since it's Windows compatible and writes Windows compatible line endings on your behalf.  This way Windows users get what they use natively and Linux/Unix will too.

lib/jekyll/theme_builder.rb
```diff
@@ -0,0 +1,108 @@
...
+    mkdir_p(SCAFFOLD_DIRECTORIES)
+  end
+
+  def create_gemspec
+    write_file("Gemfile", template("Gemfile"))
+    write_file("#{name}.gemspec", template("theme.gemspec"))
+  end
+
+  def create_accessories
+    %w{README.md CODE_OF_CONDUCT.md LICENSE.txt}.each do |filename|
```
envygeeks: Why `CODE_OF_CONDUCT`? It's not our repository our our choice to decide if this should be in there.

lib/theme_template/README.md.erb
```diff
@@ -0,0 +1,40 @@
...
+
+TODO: Write usage instructions here. Describe your available layouts, includes, and/or sass.
+
+## Contributing
+
+Bug reports and pull requests are welcome on GitHub at https://github.com/[USERNAME]/hello. This project is intended to be a safe, welcoming space for collaboration, and contributors are expected to adhere to the [Contributor Covenant](http://contributor-covenant.org) code of conduct.
+
+## License
+
+The theme is available as open source under the terms of the [MIT License](http://opensource.org/licenses/MIT).
```
parkr: @benbalter Random thought: should we suggest Creative Commons here instead?

lib/theme_template/example/index.html
```diff
@@ -0,0 +1,14 @@
+---
+# Specify a layout from your theme!
+---
+
+Lorem ipsum dolor sit amet, quo id prima corrumpit pertinacia, id ius dolor dolores, an veri pertinax explicari mea. Agam solum et qui, his id ludus graeco adipiscing. Duis theophrastus nam in, at his vidisse atomorum. Tantas gloriatur scripserit ne eos. Est wisi tempor habemus at, ei graeco dissentiet eos. Ne usu aliquip sanctus conceptam, te vis ignota animal, modus latine contentiones ius te.
```
parkr: @benbalter When a user creates a new theme using this command, they get a `bundle exec rake preview` command for free, which shows them this `example` website. Should we use this page to explain what to do, or just leave it to the README?

parkr:
@jekyllbot: merge +minor

benbalter:
@parkr Nice work. Ran things locally, and everything worked as expected, but had some feedback on the experience and on the defaults:

> `spec.metadata["plugin_type"] = "theme"`

How is this used? I remember there was a discussion about this when the feature was originally proposed, but depending on its use, wondering if we should namespace things in a `jekyll` namespace. It looks like values must be strings, but maybe `jekll_plugin_type`?

> `  spec.bindir        = "exe"`

When would a theme have an executable? (From the default gemspec)

> `  spec.files         = `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(exe|_layouts|_includes|_sass|example|example/_posts)/}) }`

Unless I'm misreading, we want to include layouts/includes/sass, the gemspec and exclude everything else (including exe). (A) I don't think this line accomplishes what we want right now, and (B) I'm wondering if we should default to a whitelist to encourage stronger defaults.

> Rakefile... `example` dir

This feels *super* heavyweight to me, and given that we have no way to e.g., distribute bug fixes to themes, I think we're setting ourselves up for heartbreak and failure "To fix this issue, copy and paste this Gist into your theme's Rakefile" `/headdesk`. 

I'm not sure the `example` dir + `Rakefile` is the best approach.... it feels a bit heavy handed and adds a lot of clutter for theme maintainers (whom we should assume are new to Ruby).  Based on the [theme docs](
http://jekyllrb.com/docs/themes/#previewing-your-theme), they should be able to simple add `page.md`, `index.html`, or any other file that they want and just run jekyll serve. The theme should be a valid Jekyll site, so that we use Jekyll's existing tooling, rather than having each theme build their own.

benbalter:
One other thing to note, we may not want to add a code of conduct by default. Having a code of conduct that the author doesn't have any intention to support could send the wrong signal... that those codes of conduct are intended to protect should contribute to this project (thinking it's safe and that they'd be protected), when in fact, that's not true. 

I'd definitely be in favor of a `--code-of-conduct` flag, or something similar, especially for a first pass, but don't know the creating a situation where a might run a command and publish a repo without knowing that my project even has a code of conduct (or what the contributor covenant is), is the best idea.

PR #4849, opened 2016-04-27
Flesh out the issue template to be much more detailed

envygeeks:
This makes the issue template more robust, making check boxes so users can submit and check or check as they go.  It also starts using HTML comments so that directions for the users aren't always displayed to us (gotta love Markdown.) And provides the users with more hints on what we would like from them when filing a ticket.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,20 +1,64 @@
-###### What version of Jekyll are you using (`jekyll -v`)?
+<!--
+  Hi! Thanks for considering to file a bug with Jekyll.  Please take the
+  the basic questions (convert `[ ]` into `[x]` to check it (or submit and check.)
+  If are no need for certain fields like output and redirection, please delete
+  those headers before submitting.  We know not all tickets require those
+  steps.  Otherwise, please try to be as detailed as possible.
+  
+  If you are unsure this is a bug in Jekyll, or this is a bug caused
+  by a plugin that isn't directly related to Jekyll, or if this is just
+  a generic usage question, please consider asking your question at
+  https://talk.jekyllrb.com where non-bug questions go.
+  
+  Thanks!
+-->
+
+- [ ] I believe this to be a bug, not a question about using Jekyll.
+- [ ] I Read the CONTRIBUTION file at https://github.com/jekyll/jekyll/blob/master/.github/CONTRIBUTING.markdown
+- [ ] I Updated to the latest Jekyll (or) if on Github Pages to the latest `github-pages`
+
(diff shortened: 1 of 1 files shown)


envygeeks:
An example of the final result can be seen here: https://github.com/envygeeks/jekyll/issues/1

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,20 +1,64 @@
...
+  If you are unsure this is a bug in Jekyll, or this is a bug caused
+  by a plugin that isn't directly related to Jekyll, or if this is just
+  a generic usage question, please consider asking your question at
+  https://talk.jekyllrb.com where non-bug questions go.
+  
+  Thanks!
+-->
 
+- [ ] I believe this to be a bug, not a question about using Jekyll.
+- [ ] I Read the CONTRIBUTION file at https://github.com/jekyll/jekyll/blob/master/.github/CONTRIBUTING.markdown
```
parkr: This should be `I read the contributing documentation at https://jekyllrb.com/docs/contributing/`, right?

parkr:
I like this! I think we should add back the section about what they expected to get. Maybe `The Output I Wanted`.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,20 +1,64 @@
...
 
+---
 
-###### What operating system are you using?
+- [ ] I am on ***MacOS*** 10+
+- [ ] I am on ***Ubuntu*** GNU/Linux
+- [ ] I am on ***Fedora*** GNU/Linux
+- [ ] I am on ***Arch*** GNU/Linux
+- [ ] I am on ***Other*** GNU/Linux
+- [ ] I am on ***Windows*** 10+
```
DirtyF: Why only Windows 10+? Because of upcoming bash support?

envygeeks: @DirtyF Because Ruby already has shoddy Windows support, the last thing I want to do is exacerbate our support for it by supporting what we already have iffy support for by supporting a larger array of Windows versions.

envygeeks:
Ping: @parkr @benbalter @alfredxing @jekyll/documentation @jekyll/ecosystem 
This pull is completed, you can see an example: https://github.com/envygeeks/jekyll/issues/1
Please review for merge.

parkr:
@envygeeks No where in the example does it explain the bug. It just has "What I Wanted," not "What I Got Instead"

parkr:
A little concerned this will be too much and folks won't read it but let's give it a shot.

@jekyllbot: merge +dev

PR #4854, opened 2016-04-29
site: use liquid & reduce some whitespace noise

parkr:
/cc @jekyll/documentation

### DIFF
--- a/site/_docs/continuous-integration.md
+++ b/site/_docs/continuous-integration.md
@@ -38,7 +38,7 @@ Save the commands you want to run and succeed in a file: `./script/cibuild`
 
 ### The HTML Proofer Executable
 
-{% highlight bash %}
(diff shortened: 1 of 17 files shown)


parkr:
@jekyllbot: merge +site

parkr:
Modified compiled site: 598d902

site/_docs/continuous-integration.md
```diff
@@ -38,7 +38,7 @@ Save the commands you want to run and succeed in a file: `./script/cibuild`
 
 ### The HTML Proofer Executable
 
-{% highlight bash %}
```
envygeeks: I always wondered this, does Rogue and Pygments really treat Bash and Sh differently?  I mean one might have more tokens I guess but I had always figured the former would just inherit the latter.

PR #4857, opened 2016-04-29
Don't rescue LoadError or bundler load errors for Bundler.

### LINKED ISSUE #4856: Jekyll surfaces the wrong error.
Jekyll Version: 3.1.3
OS: Ubuntu

Jekyll will surface the wrong error given an error happens before it runs Liquid.  Returning no error at all and tricking the user into thinking the error is surfaced in another spot.  Take this example:

```
Configuration file: /home/jordon/development/envygeeks/envygeeks.io/_config.yml
            Source: /home/jordon/development/envygeeks/envygeeks.io
       Destination: /home/jordon/development/envygeeks/envygeeks.io/_site
      Generating... 
  Liquid Exception: Liquid syntax error (line 8): Unknown tag 'css' in _layouts/default.html
jekyll 3.1.3 | Error:  Liquid syntax error (line 8): Unknown tag 'css'
```

However, the truth of it is, that Jekyll got an error long before that when it tried to require a plugin, which had another error, the real error being:

```
cannot load such file -- /home/jordon/development/jekyll/assets/lib/jekyll/\
  assets/patches/sprockets/helpers (LoadError)
```

Jekyll entirely dropped the error from upstream which makes it hard to debug actual problems in Jekyll.  It's not surfacing errors at all.  ***NOTE: Even with `--trace` it will surface the same wrong error, with all the paths involved.***

envygeeks:
When it comes to bundler it's smart enough to know what to require, and in casees it's not, it's smart enough to accept :require.  In most cases when bundler has a LoadError (or otherwise) it's because there is a problem inside of the Gem itself and when this happens, Jekyll will happily let that error slip when it shouldn't, resulting in a badly placed error that is actually wrong. This corrects that so errors can surface properly.

### DIFF
--- a/lib/jekyll/plugin_manager.rb
+++ b/lib/jekyll/plugin_manager.rb
@@ -30,16 +30,16 @@ module Jekyll
     def self.require_from_bundler
       if !ENV["JEKYLL_NO_BUNDLER_REQUIRE"] && File.file?("Gemfile")
         require "bundler"
-        Bundler.setup # puts all groups on the load path
-        required_gems = Bundler.require(:jekyll_plugins) # requires the gems in this group only
+
+        Bundler.setup
+        required_gems = Bundler.require(:jekyll_plugins)
         Jekyll.logger.debug("PluginManager:", "Required #{required_gems.map(&:name).join(', ')}")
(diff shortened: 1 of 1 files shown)


parkr:
LGTM.

@jekyllbot: merge +bug

PR #4860, opened 2016-04-30
where_exp filter should filter posts

pathawks:
```liquid
{{ site.posts | where_exp: "posts", "posts.title == 'My Title'" }}
```

This just returns `site.posts`, it does not filter posts by title.

### DIFF
--- a/test/test_filters.rb
+++ b/test/test_filters.rb
@@ -15,6 +15,10 @@ class TestFilters < JekyllUnitTest
     end
   end
 
+  class SelectDummy
+    def select; end
+  end
+
   context "filters" do
     setup do
       @filter = JekyllFilter.new({
@@ -530,6 +534,11 @@ class TestFilters < JekyllUnitTest
         assert_equal 1, results.length
         assert_equal 4.7, results[0]["rating"]
       end
+
+      should "always return an array if the object responds to `select`" do
+        results = @filter.where(SelectDummy.new, "obj", "1 == 1")
+        assert_equal [], results
+      end
     end
 
     context "where_exp filter" do
@@ -601,6 +610,11 @@ class TestFilters < JekyllUnitTest
         results = @filter.where_exp(posts, "obj", "obj.title == 'Foo Bar'")
         assert_equal 1, results.length
         assert_equal site.posts.find { |p| p.title == "Foo Bar" }, results.first
+      end
+
+      should "always return an array if the object responds to `select`" do
+        results = @filter.where_exp(SelectDummy.new, "obj", "1 == 1")
+        assert_equal [], results
       end
     end
 
--- a/test/test_filters.rb
+++ b/test/test_filters.rb
@@ -594,6 +594,14 @@ class TestFilters < JekyllUnitTest
         assert_equal "b", results[1]["id"]
         assert_equal "d", results[2]["id"]
       end
+
+      should "filter posts" do
+        site = fixture_site.tap(&:read)
+        posts = site.site_payload["site"]["posts"]
+        results = @filter.where_exp(posts, "obj", "obj.title == 'Foo Bar'")
+        assert_equal 1, results.length
+        assert_equal site.posts.find { |p| p.title == "Foo Bar" }, results.first
+      end
     end
 
     context "sort filter" do
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -233,11 +233,11 @@ module Jekyll
     #
     # Returns the filtered array of objects
     def where(input, property, value)
-      return input unless input.is_a?(Enumerable)
+      return input unless input.respond_to?(:select)
       input = input.values if input.is_a?(Hash)
       input.select do |object|
         Array(item_property(object, property)).map(&:to_s).include?(value.to_s)
-      end
+      end || []
     end
 
     # Filters an array of objects against an expression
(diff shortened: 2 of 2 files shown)


parkr:
@pathawks If someone filed an issue for that here, we'd likely ask them to change it to `{{ site.posts | where: "title", "My Title" }}`. It's definitely a bug but also good to know that there's another way for this particular operation.

Is it all `==` operations? Could you `assert_equal EXPECTED_SITE, results.size` instead of comparing against `site.posts`? Is the result supposed to be empty?

pathawks:
Ultimately, my use case is to get a list of all posts from the past 12 months. When tge date stuff wasn’t working, I decided to see if I could get the `where_exp` filter to work _at all_, and I could not.

I know there is at least one post who’s title is `Foo Bar`. I will take another look tonight to see if I can make the test a bit tighter.

parkr:
Ok, thanks @pathawks 😄 

You could assert that (1) the result is the right size (i.e. 1) and (2) that the result contains the post you're looking for (find the document yourself using `site.posts.find {|p| p.title == "Foo Bar"}` and compare)

pathawks:
Changes have been made :+1:

Test still fails, which is the point. The filter is _not_ filtering, but just returning all posts.

jeromecoupe:
Same here with Jekyll 3.1.6

- This works fine: '{% assign potsByAuthor = site.posts | where: 'author', page.teamId %}`
- Whereas this just returns all the posts `{% assign potsByAuthor = site.posts | where_exp: 'item', 'item.author == page.teamId' %}` The size of the `potsByAuthor` is always the total number of posts
- Same if I try with a string: `{% assign potsByAuthor = site.posts | where_exp: 'item', 'item.author == "jeromecoupe"' %}`

parkr:
Hm, tests aren't being triggered. Would you mind rebasing on `jekyll:master` to try to re-trigger the build (and resolve conflicts)?

pathawks:
Not sure how the failing test could be related (and only with Ruby 2.3). Would you mind restarting that build to see if it was just a fluke?

pathawks:
Thanks! Looks like all tests pass now, and no conflicts with **master**

parkr:
@pathawks on mobile so can't double check, but do we actually test filtering `site.posts`? I bet drops behave strangely :stuck_out_tongue:

test/test_filters.rb
```diff
@@ -594,6 +603,18 @@ def to_liquid
         assert_equal "b", results[1]["id"]
         assert_equal "d", results[2]["id"]
       end
+
+      should "filter posts" do
+        site = fixture_site.tap(&:read)
+        results = @filter.where_exp(site.posts, "obj", "obj.title == 'Foo Bar'")
+        assert_equal 1, results.length
+        assert_equal site.posts.find { |p| p.title == "Foo Bar" }, results.first
```
pathawks: > do we actually test filtering `site.posts`?

:+1:

test/test_filters.rb
```diff
@@ -594,6 +603,18 @@ def to_liquid
         assert_equal "b", results[1]["id"]
         assert_equal "d", results[2]["id"]
       end
+
+      should "filter posts" do
+        site = fixture_site.tap(&:read)
+        results = @filter.where_exp(site.posts, "obj", "obj.title == 'Foo Bar'")
```
parkr: This should be `site.site_payload["site"]["posts"]` to mimic the Liquid context.

pathawks:
Travis has given up on me?

pathawks:
Failure is because [Travis cannot install activesupport](https://travis-ci.org/jekyll/jekyll/jobs/144208193#L465)

parkr:
Looks like we need to depend on activesupport v4.x for Ruby 2.1 and below?

Crunch09:
@parkr [Yes](https://github.com/rails/rails/blob/5-0-stable/activesupport/activesupport.gemspec#L10), would you be interested in a PR for this?

parkr:
@Crunch09, Yes!

parkr:
#5100 is merged, would you mind rebasing?

pathawks:
-[x] Rebased

test/test_filters.rb
```diff
@@ -594,6 +603,19 @@ def to_liquid
         assert_equal "b", results[1]["id"]
         assert_equal "d", results[2]["id"]
       end
+
+      should "filter posts" do
+        site = fixture_site.tap(&:read)
+        posts = site.site_payload["site"]["posts"]
+        results = @filter.where_exp(posts, "obj", "obj.title == 'Foo Bar'")
+        assert_equal 1, results.length
+        assert_equal site.posts.find { |p| p.title == "Foo Bar" }, results.first
```
parkr: Here's a curiosity: why isn't this comparing a Document with a DocumentDrop? Can you ensure `results.first` is, in fact, a DocumentDrop?

parkr:
@pathawks ❤️

test/test_filters.rb
pathawks: `results.first` is a `Document`, not a `DocumentDrop`.
So... what should I do about that?

parkr: Maybe the CollectionDrop should be recursively calling to_liquid on the documents it's accessing. When and how Liquid calls this method has always been a mystery.

parkr:
@jekyllbot: merge +bug

pathawks:
I will not `@mention` in commit messages.
I will not `@mention` in commit messages.
I will not `@mention` in commit messages.

parkr:
@pathawks ❤️

Issue #4882, opened 2016-05-12
the HTML snippet in docs/pagination cannot work well


WaTer9527:
###### What version of Jekyll are you using (`jekyll -v`)?

jekyll 3.0.1

###### What operating system are you using?

win10 PC

###### What did you do?
(Please include the content causing the issue, any relevant configuration settings, and the command you ran)

I used jekyll-paginate,there is bug when I click paginate button after used the HTML snippet in the end of [document](https://jekyllrb.com/docs/pagination/).

###### What did you expect to see?

when I click the button 2, the url change to 'localhost:4000/page2/'.when I click the button 3, the url change to 'localhost:4000/page3/'

###### What did you see instead?

when I click the button 2, the url change to 'localhost:4000/page2/'.when I click the button 3, the url change to 'localhost:4000/page2/page3/'

xHN35RQ:
Please link to the code that is causing problems for you.

WaTer9527:
the following HTML snippet for [pagination documen](https://jekyllrb.com/docs/pagination/)t:
`{% if paginator.total_pages > 1 %}
<div class="pagination">
  {% if paginator.previous_page %}
    <a href="{{ paginator.previous_page_path | prepend: site.baseurl | replace: '//', '/' }}">&laquo; Prev</a>
  {% else %}
    <span>&laquo; Prev</span>
  {% endif %}

  {% for page in (1..paginator.total_pages) %}
    {% if page == paginator.page %}
      <em>{{ page }}</em>
    {% elsif page == 1 %}
      <a href="{{ paginator.previous_page_path | prepend: site.baseurl | replace: '//', '/' }}">{{ page }}</a>
    {% else %}
      <a href="{{ site.paginate_path | prepend: site.baseurl | replace: '//', '/' | replace: ':num', page }}">{{ page }}</a>
    {% endif %}
  {% endfor %}

  {% if paginator.next_page %}
    <a href="{{ paginator.next_page_path | prepend: site.baseurl | replace: '//', '/' }}">Next &raquo;</a>
  {% else %}
    <span>Next &raquo;</span>
  {% endif %}
</div>
{% endif %}`

xHN35RQ:
Sorry, what I meant is can you link me to your jekyll repository so I can look at the code for your site?

WaTer9527:
I'm sorry for my poor english, In fact, I want to submit a pull request.

PR #4883, opened 2016-05-12
add '/' in pagination sample snippets

WaTer9527:
issue:#4882


pathawks:
From the docs:

```
paginate_path: "/blog/page:num/"
```

It looks like `paginate_path` is expected to have the slash in front already.

WaTer9527:
thanks

pathawks:
It might not be a bad idea to raise a warning if the slash is missing.

It should be easy and obvious how to do the right thing, and there seems to be room for improvement here.

pathawks:
I was going to work on adding this warning, but it looks as though **jekyll-paginate** is no longer being actively developed.

envygeeks:
See: https://talk.jekyllrb.com/t/announcement-retiring-jekyll-paginate/1004

Issue #4885, opened 2016-05-12
Make jekyll/jekyll code compliant with rubocop rules


parkr:
Rubocop can help us keep our code consistent in style. Consistency makes it easier to read the code and having Rubocop handle reporting offenses makes feedback to pull requests quick & less controversial.

Instructions for fixing a file:
1. Ensure there isn't an existing pull request fixing the files you're looking at updating
2. Remove the file from `.rubocop.yml` under `AllCops.Exclude`.
3. Run `bundle install` to ensure you have all the gems.
4. Run `script/fmt`.
5. Fix errors until `script/fmt` reports no errors.
6. Commit to a feature branch on your fork and create a pull request mentioning this issue.
7. Rejoice!

Here's a checklist of files **in-progress**:
- [x] Fix offenses in `lib/jekyll/convertible.rb` and enable in `.rubocop.yml` (#5031, #5308)
- [x] Fix offenses in `lib/jekyll/document.rb` and enable in `.rubocop.yml` (#5030, #5045)
- [x] Fix offenses in `lib/jekyll/collection.rb` and enable in `.rubocop.yml` (#5022)
- [x] Fix offenses in `lib/jekyll/configuration.rb` and enable in `.rubocop.yml` (#5024)
- [x] Fix offenses in `lib/jekyll/regenerator.rb` and enable in `.rubocop.yml` (#5025)
- [ ] Fix offenses in `lib/jekyll/renderer.rb` and enable in `.rubocop.yml` (#5052)
- [x] Fix offenses in `lib/jekyll/utils.rb` and enable in `.rubocop.yml` (#5026)

Here's a checklist of completed tasks:
- [x] Finalize Rubocop rules (#4886)
- [x] Enable Rubocop check for pull requests (#4886)
- [x] Fix offenses in `exe/jekyll` and enable in `.rubocop.yml` (#5017)
- [x] Fix offenses in `lib/jekyll/cleaner.rb` and enable in `.rubocop.yml` (#4892)
- [x] Fix offenses in `lib/jekyll/commands/build.rb` and enable in `.rubocop.yml` (#4888)
- [x] Fix offenses in `lib/jekyll/commands/clean.rb` and enable in `.rubocop.yml` (#4888)
- [x] Fix offenses in `lib/jekyll/commands/doctor.rb` and enable in `.rubocop.yml` (#4888)
- [x] Fix offenses in `lib/jekyll/commands/help.rb` and enable in `.rubocop.yml` (#4888)
[...]

PR #4886, opened 2016-05-12
Update Rubocop rules

parkr:
This merges the rule set from [github/pages-gem `.rubocop.yml`](https://github.com/github/pages-gem/blob/master/.rubocop.yml) with our set here, overriding what we have with what is present in `pages-gem`'s Rubocop rules.

I sorted the keys so they're grouped a bit more logically and excluded all the files so we can enable one-by-one. It'd be an easy pull request to fix the rubocop errors and enable the file. :smile: Tracking in #4885.

@jekyll/core Will you please take a look? There are likely things we should tweak here.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,80 +1,199 @@
-Metrics/MethodLength: { Max:  24 }
-Metrics/ClassLength:  { Max: 240 }
-Metrics/ModuleLength: { Max: 240 }
-Metrics/LineLength:   { Max: 112 }
-Metrics/CyclomaticComplexity: { Max: 8 }
-Metrics/PerceivedComplexity: { Max: 8 }
-Metrics/ParameterLists: { Max: 4 }
-Metrics/MethodLength: { Max: 24 }
-Metrics/AbcSize: { Max: 20 }
-
-Style/IndentHash: { EnforcedStyle: consistent }
-Style/HashSyntax: { EnforcedStyle: hash_rockets }
-Style/SignalException: { EnforcedStyle: only_raise }
-Style/AlignParameters: { EnforcedStyle: with_fixed_indentation }
-Style/StringLiteralsInInterpolation: { EnforcedStyle: double_quotes }
-Style/MultilineMethodCallIndentation: { EnforcedStyle: indented }
-Style/MultilineOperationIndentation: { EnforcedStyle: indented }
-Style/FirstParameterIndentation: { EnforcedStyle: consistent }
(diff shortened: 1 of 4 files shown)


.rubocop.yml
```diff
@@ -1,80 +1,199 @@
...
-  - Readme.{md,markdown,txt,textile}
-  - ReadMe.{md,markdown,txt,textile}
-  - COPYING
-  - LICENSE
-
-  - site/**/*
-  - test/**/*
-  - vendor/**/*
-  - features/**/*
+  - lib/jekyll/cleaner.rb
```
envygeeks: It seems this not only broadened our scope but possibly lessoned it no? This list would be mighty unmaintainable an seems refactor unfriendly.  We shouldn't expect users to maintain our Rubocop file, my vote is for sticking with the old list at least on this part.

parkr: @envygeeks In terms of `AllCops.Exclude`, I added each file individually so we can go through each and enable one-by-one. It's easier to review PR's for one or a handful of files and it means we can add the CodeClimate CI status _now_ instead of waiting.

.rubocop.yml
```diff
@@ -1,80 +1,199 @@
...
+  - lib/jekyll/utils.rb
+  - lib/jekyll/version.rb
+  - lib/jekyll.rb
+  - bin/**/*
   - script/**/*
-  - spec/**/*
+  - vendor/**/*
+Lint/AmbiguousRegexpLiteral:
+  Exclude:
+  - features/step_definitions.rb
```
envygeeks: You are actually working around a literal Ruby warning, so this error will still surface on other linters (like Atom Linter, Sublime Text Linter and others.) According to Cucumber themselves it seems the preferred way to fix this with them is to just use `%r!!` which works for all parties, it's less escaping!

.rubocop.yml
```diff
@@ -1,80 +1,199 @@
...
+Style/SignalException:
+  EnforcedStyle: only_raise
+Style/SingleLineMethods:
+  Enabled: false
+Style/SpaceAroundOperators:
+  Enabled: false
+Style/SpaceInsideBrackets:
+  Enabled: false
+Style/StringLiterals:
+  Enabled: false
```
envygeeks: I would love if we kicked this on, I like Github have a huge love of "" over '' because I find that I interpolate more than I don't but I'll digress that I don't even know if Github themselves enforces it in their own source.

parkr:
@jekyllbot: merge +dev

PR #4888, opened 2016-05-12
Update `jekyll/commands*` to pass rubocop rules

TheLucasMoore:
One done. Working on the rest of commands folder now.


lib/jekyll/commands/clean.rb
```diff
@@ -4,8 +4,9 @@ class Clean < Command
       class << self
         def init_with_program(prog)
           prog.command(:clean) do |c|
-            c.syntax 'clean [subcommand]'
-            c.description 'Clean the site (removes site output and metadata file) without building.'
+            c.syntax "clean [subcommand]"
+            c.description "Clean the site
+            (removes site output and metadata file) without building."
```
envygeeks: This will output literally and it's hard for static analysis to catch this.  Try:
```ruby
"string1 " \
  "string2"
```

envygeeks:
Thanks for your contribution! Just a comment, and thanks again!

lib/jekyll/commands/clean.rb
TheLucasMoore: Yes, I noticed that in later files. Fixed in this coming PR. Thanks @envygeeks.

envygeeks:
Accidental close?

TheLucasMoore:
Yeah, sorry. I see now that commits continually update here. Currently running tests.

parkr:
@TheLucasMoore 😍 Thank you so much for the pull request!

One easier way to see if the changes will work is to run `script/fmt` from your computer locally. It should work fine on Windows, because Rubocop is (I think?) pure Ruby. You should be able to see the failures there and can commit and push once they're all fixed. Two left!

```text
lib/jekyll/commands/new.rb:21:9: C: Metrics/AbcSize: Assignment Branch Condition size for process is too high. [23.22/20]
        ^^^
lib/jekyll/commands/serve.rb:174:9: C: Metrics/AbcSize: Assignment Branch Condition size for enable_ssl is too high. [29.7/20]
        def enable_ssl(opts)
        ^^^
```

lib/jekyll/commands/serve.rb
```diff
@@ -174,19 +174,22 @@ def enable_logging(opts)
         def enable_ssl(opts)
           return if !opts[:JekyllOptions]["ssl_cert"] && !opts[:JekyllOptions]["ssl_key"]
           if !opts[:JekyllOptions]["ssl_cert"] || !opts[:JekyllOptions]["ssl_key"]
-            raise RuntimeError, "--ssl-cert or --ssl-key missing."
+            raise RuntimeError
```
parkr: Can you add this message back? Otherwise it's confusing. 😄

lib/jekyll/commands/serve/servlet.rb
```diff
@@ -27,7 +27,7 @@ def search_file(req, res, basename)
 
         #
 
-        def do_GET(req, res)
+        def do_get(req, res)
```
parkr: I believe this is a special method so you can add a comment above the method to disable whatever it violates via `# rubocop:disable <copname>` e.g. `# rubocop:disable Metrics/LineLength`.

parkr:
@TheLucasMoore [The ABC Size cop is about complexity of the method](http://www.rubydoc.info/gems/rubocop/0.27.0/RuboCop/Cop/Metrics/AbcSize). The way to fix it is by breaking big methods into smaller methods that do the `if/else`'s and so on.

lib/jekyll/commands/serve/servlet.rb
envygeeks: It is, this is an upstream method that we override upon subclassing.  So in order for us to fix this upstream would have to fix it.

TheLucasMoore:
Alright, the commands subfolder should all be passing now. I added rubocop exceptions for the redundant error and fixed one of the ABC metric size errors. The **enable_ssl** private method in **serve.rb** is passing with a rubocop exception. I tried a few ways of breaking it apart, but it seems like an exception is the most efficient way to have it passing.

envygeeks:
I'm gonna be honest, looking at all these lines that have to shift down decreasing readability makes me think two things (which are exclusive:)

1. We need to rewrite those files to be cleaner to begin with, they shift a lot, this is hindering readability extraordinarily (not @TheLucasMoore's fault, it's our fault.)
2. We need to increase the length we allow for lines (I'm almost against this because sometimes I work on tiny screens and it would make it hard to audit a bug in that case.)

Ultimately though, reviewing, most of it, it LGTM.

lib/jekyll/commands/new.rb
```diff
@@ -28,19 +28,7 @@ def process(args, options = {})
...
-
-            File.open(File.expand_path(initialized_post_name, new_blog_path), "w") do |f|
-              f.write(scaffold_post_content)
-            end
-
-            File.open(File.expand_path("Gemfile", new_blog_path), "w") do |f|
-              f.write(gemfile_contents)
-            end
-          end
+          if_options_blank(new_blog_path, options)
```
parkr: Why don't we have the `if` still in this method and move just the larger part of the `else` into its own method:

```ruby
  # ...
  if options["blank"]
    create_blank_site new_blog_path
  else
    create_site new_blog_path
  end

  # ...
end

def create_site(new_blog_path)
  create_sample_files new_blog_path

  File.open(...) do |f|
    # ...
  end

  File.open(...) do |f|
    # ..
  end
end
```

Does that sound good to you? Then the method doesn't start with an `if`.

parkr:
> We need to increase the length we allow for lines (I'm almost against this

I'm ok bumping it 100 but 90 is working fine for me. The problem is really that we're so far indented that you only get 50 characters to work with.

envygeeks:
> I'm ok bumping it 100 but 90 is working fine for me. The problem is really that we're so far indented that you only get 50 characters to work with.

What do you think about maybe shifting to `class Jekyll::Blah::Blah` for def of classes which would give us way more room.  Since we autoload a lot and people normally load "jekyll" before anything I think that would be a good idea to give us more room since we have deep nesting in some places.

parkr:
> What do you think about maybe shifting to class Jekyll::Blah::Blah for def of classes which would give us way more room

Yeah, that's a good idea. I know that has slightly different semantics, though. Should we do

```ruby
module Jekyll::Commands
  class Build < Command
  end
end
```

i.e. just collapse to two levels of nesting?

envygeeks:
> i.e. just collapse to two levels of nesting?

:+1:

TheLucasMoore:
@envygeeks - Alright, so that change is quite a large scope which has a lot of structural implications. How should we proceed within the scope of this PR? I'm happy to continue to bring other folders into compliance with Rubocop.

parkr:
Thanks very much for this, @TheLucasMoore!

@jekyllbot: merge +dev

PR #4889, opened 2016-05-13
Update YAML for default collection attributes

mikewheaton:
The current syntax results in a compilation error. This update matches the syntax in the [original pull request](https://github.com/jekyll/jekyll/pull/2419) and compiles without errors.

### DIFF
--- a/site/_docs/collections.md
+++ b/site/_docs/collections.md
@@ -34,7 +34,8 @@ Default attributes can also be set for a collection:
 
 {% highlight yaml %}
 defaults:
-  - scope:
+  -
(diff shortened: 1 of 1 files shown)


pathawks:
The current syntax seems like it should be valid YAML. What is the compilation error? Is there a bug in **safe_yaml**?

mikewheaton:
That's strange, it's working fine for me now. I must have had an extra space in there or something. Sorry for the trouble!

pathawks:
Thanks for investigating! Glad everything is working properly 😅:+1:

PR #4892, opened 2016-05-13
Fixing rubocop offenses in lib/jekyll/cleaner.rb

derekgottlieb:


parkr:
Thank you so much! 😍 

@jekyllbot: merge +dev

derekgottlieb:
If I cleaned up some more of these, would you prefer a PR per file or a PR for several of them?

envygeeks:
A single PR with multiple and a commit for each file would be preferable for cleanliness.

derekgottlieb:
Can do!

PR #4904, opened 2016-05-17
Update Jekyll Styleguide, Fixes #4893

### LINKED ISSUE #4893: Contribution: Ruby Styleguide not present
@jekyll/documentation I was trying to find a style guide for Ruby to fix rubocop offenses and noticed that [styleguide/ruby](https://github.com/styleguide/ruby) repo does not exist anymore.

I think we should find another style guide that matches our current one. I don't know if there is already one but we should fix the link in [this](https://jekyllrb.com/docs/contributing/#code-contributions-generally) section.

mertkahyaoglu:
@parkr What do you think? Rubocop depends on this styleguide. Should I also mention it in the docs?

### DIFF
--- a/site/_docs/contributing.md
+++ b/site/_docs/contributing.md
@@ -95,7 +95,7 @@ If your contribution changes any Jekyll behavior, make sure to update the docume
 
 ### Code contributions generally
 
-* Jekyll follows the [GitHub Ruby Styleguide](https://github.com/styleguide/ruby).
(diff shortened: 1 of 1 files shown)


parkr:
This will work in the interim but I hear it's coming back: https://github.com/Homebrew/brew/issues/106#issuecomment-214010443

mertkahyaoglu:
I'm closing this since our styleguide is back :fireworks:

Issue #4906, opened 2016-05-17
page.excerpt isn't being properly cleared when rendering excerpts


parkr:
This applies to v3.1.3, where this was uncovered earlier today.

When rendering an excerpt, we use the parent document's data. When we used Hashes, [generating the `Document#to_liquid` without an excerpt present worked fine](https://github.com/jekyll/jekyll/blob/v3.1.3/lib/jekyll/excerpt.rb#L61-L66), but now that it's a `Liquid::Drop`, it's no longer working as expected.

We should create an `ExcerptDrop` that explicitly removes the ability to get the `excerpt` of an Excerpt. 😄 

To recreate the stack overflow that currently occurs, you can use `electron/electron.atom.io` at [revision 21b27f7a836c911204a337cd32e943029c068aff](https://github.com/electron/electron.atom.io/commit/21b27f7a836c911204a337cd32e943029c068aff) with Jekyll v3.1.3.

/cc @kevinsawicki

envygeeks:
@parkr all that data sits on a Drop no? Why not just generate that on the drop... when the user asks for it?

parkr:
> Why not just generate that on the drop... when the user asks for it?

The problem is when an excerpt references itself – it creates an infinite recursive loop.

envygeeks:
@parkr oh, I love bugs like that.

PR #4907, opened 2016-05-17
Add ExcerptDrop and remove excerpt's ability to refer to itself in Liquid

parkr:
This fixes the bug in v3.1.3 described in https://github.com/jekyll/jekyll/issues/4906.

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -21,7 +21,7 @@ require 'minitest/autorun'
 require 'minitest/reporters'
 require 'minitest/profile'
 require 'rspec/mocks'
-require 'jekyll'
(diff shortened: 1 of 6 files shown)


test/helper.rb
```diff
@@ -21,7 +21,7 @@ def jruby?
 require 'minitest/reporters'
 require 'minitest/profile'
 require 'rspec/mocks'
-require 'jekyll'
+require File.expand_path("../lib/jekyll.rb", __dir__)
```
envygeeks: `require_relative` it's been so long I can't remember but I think it's been there since 1.9 so you should be able to use it.

PR #4908, opened 2016-05-18
Add timings for each scenario in cucumber & print worst offenders

parkr:
This should help us understand where we can improve.

I think it takes around half a second just to boot the VM and load Jekyll :(


parkr:
@jekyllbot: merge +dev

Issue #4897, opened 2016-05-14
Layout data used on page with different layout


stomar:
I'm in the process of migrating from Jekyll 2 to Jekyll 3.1.3 (on Linux) with a site that uses nested layouts:

- default
  - page
  - special

`special` defines a property in the YAML front matter that is used in `default` if present; the property is not defined by `page`.

I have several pages (page_1, page_2, page_3, ...) where some use `page` and some use `special`.

After the first page that uses `special` and defines the property, the property seems to be defined for *all* pages, even the pages that use the "normal" `page` layout.

With Jekyll 2 this worked fine (with `page.property` instead of `layout.property`).

I believe this is a bug, or are the layout properties meant to be available globally?

(I can provide the source of a minimal example project that demonstrates the behavior if needed.)

agriffis:
This seems to be the same as what @kvz describes at https://github.com/jekyll/jekyll/issues/4433#issuecomment-177990626

stomar:
Seems so, yes. And I just noticed the very same problem is also described in #4884, and in #4442 (from February 2nd!), and possibly other issues.

It really would be nice to have this fixed soon; currently this is a blocker for migrating www.ruby-lang.org to Jekyll 3.

parkr:
@stomar Can you give `master` a try?

parkr:
@stomar I'm working on a 3.1.4. I'll add this to the list.

stomar:
Tested with master (952947c65efd9ca5d4636edd8b26f5d7ac2bfdbf), no difference. With which commit has this supposedly been fixed?

parkr:
@stomar I don't believe it has been fixed, but I wanted to make sure – we don't have this particular behaviour tested, sadly. 😦 Working on it now!

stomar:
3.1.5 works as expected :smile: :clap:

Issue #4433, opened 2016-02-01
Layout inheritance broken in Jekyll 3.1.1?


kvz:
Hi,

I have the following layout inheritance set up: `post.html > blog.html > default.html`.

Example of `post.html`:

```
---
layout: blog
bodyClasses: "blog-post"
---

<div>
```

How do I make a frontmatter variable called `bodyClasses`, set in `post.html`, available in `default.html`?

I have tried using `{{layout.bodyClasses}}` as well as `{{page.bodyClasses}}`, but in both cases, the value set in `blog.html` is used.

I would expect layouts that inherit from parent layouts to be able to override these variables, and I think this used to work in 2.0 releases too?

envygeeks:
The worry is if it worked in 3.0 and not 3.1, if it worked in 2.0 and not in 3.1 then that's a possible WONTFIX as feature changes are expected in major releases as SemVer implies.  Did it work in 3.0?

kvz:
I just tested it and it doesn't work in 3.0 either. I am aware of this breaking change: http://jekyllrb.com/docs/upgrading/2-to-3/#layout-metadata and it there said that going from `page` to `layout` was to help make it clearer what gets merged. It does not mention however how the merging works if you have several layouts extending each other. In my case, an ancestor layout wins and overrides any layout extending on top of it.

Would it not make sense if the 'highest' layout would overwrite variables that ancestors have already set?

I agree that major version bumps are a legit way of doing breaking changes, so I'm all for that. I'm just wondering if a bug (or at least unpredictable behavior) could be at play here. 

Possibly related #4203 #4246

envygeeks:
I do think it makes sense that a child should override the ancestor and it's a bit unintuitive for a child's value not to appear at all, however I'll defer the decision on if it's intended or not to @parkr as he works on that area.  I wonder if the work around isn't `{{ page.data.var | default: pay.layout.var }}` does trying the front-matter directly and then falling back to layout work?

envygeeks:
Oh sorry, I'm bad at reading, I didn't notice you had already tried that.  Ignore that trial and error.

kvz:
Yes this is zooming in just on `layout.var` now, where the problem is Jekyll seems unclear on how to merge child over ancestors.

[...]

PR #4909, opened 2016-05-18
3.1: Fix `{{ layout }}` oddities (proper inheritance & fixing overflow of old data)

parkr:
This pull request fixes two bugs:

1. #4433, which describes a bug where if you have a variable in a layout "special" that inherits from "default," you will see the data value from "default" instead of "special." The inheritance should mean that more specific layouts' data override the less specific layouts' data.
2. #4897, which describes a bug where the rendering of a page means its `{{ layout }}` variables exist for pages rendered later without the same data hierarchy (they "overflow" from one page to the next without being specified by the subsequent page)

This is slated for v3.1.4. 😄

### DIFF
--- a/lib/jekyll/renderer.rb
+++ b/lib/jekyll/renderer.rb
@@ -135,10 +135,13 @@ module Jekyll
 
       used   = Set.new([layout])
 
+      # Reset the payload layout data to ensure it starts fresh for each page.
+      payload["layout"] = nil
+
       while layout
         payload['content'] = output
         payload['page']    = document.to_liquid
-        payload['layout']  = Utils.deep_merge_hashes(payload['layout'] || {}, layout.data)
+        payload['layout']  = Utils.deep_merge_hashes(layout.data, payload["layout"] || {})
 
         output = render_liquid(
           layout.content,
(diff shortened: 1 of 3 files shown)


lib/jekyll/renderer.rb
```diff
@@ -135,10 +135,13 @@ def place_in_layouts(content, payload, info)
...
       used   = Set.new([layout])
 
+      # Reset the payload layout data to ensure it starts fresh for each page.
+      payload["layout"] = nil
+
       while layout
         payload['content'] = output
         payload['page']    = document.to_liquid
-        payload['layout']  = Utils.deep_merge_hashes(payload['layout'] || {}, layout.data)
+        payload['layout']  = Utils.deep_merge_hashes(layout.data, payload["layout"] || {})
```
parkr: @benbalter This is a breaking change that is considered a bug fix (it was a bug that the default layout, for example, overwrote the metadata from a more specific/child layout like "page" or "post). How can I best communicate this change?

benbalter: Chatted offline, but just to memorialize here:

1. Changelog
2. Jekyll blog post upon release
3. GitHub blog post upon upgrade

PR #4916, opened 2016-05-20
Rubocop cleanup for some utils and further test files

brint:
More cleanup for #4885

Files addressed:

lib/jekyll/utils/ansi.rb
lib/jekyll/utils/platforms.rb
lib/jekyll/version.rb
test/test_related_posts.rb
test/test_sass.rb
test/test_static_file.rb
test/test_theme.rb
test/test_url.rb

### DIFF
--- a/lib/jekyll/version.rb
+++ b/lib/jekyll/version.rb
@@ -1,3 +1,3 @@
 module Jekyll
-  VERSION = '3.2.0.pre.beta1'
+  VERSION = "3.2.0.pre.beta1".freeze
 end
--- a/test/test_static_file.rb
+++ b/test/test_static_file.rb
@@ -63,16 +63,24 @@ class TestStaticFile < JekyllUnitTest
     end
 
     should "have a destination relative directory with a collection" do
-      static_file = setup_static_file_with_collection("root", "_foo/dir/subdir",
-        "file.html", { "output" => true })
+      static_file = setup_static_file_with_collection(
+        "root",
+        "_foo/dir/subdir",
+        "file.html",
+        { "output" => true }
+      )
       assert_equal :foo, static_file.type
       assert_equal "/foo/dir/subdir/file.html", static_file.url
       assert_equal "/foo/dir/subdir", static_file.destination_rel_dir
     end
 
     should "use its collection's permalink template for destination relative directory" do
-      static_file = setup_static_file_with_collection("root", "_foo/dir/subdir",
-        "file.html", { "output" => true, "permalink" => "/:path/" })
+      static_file = setup_static_file_with_collection(
+        "root",
+        "_foo/dir/subdir",
+        "file.html",
+        { "output" => true, "permalink" => "/:path/" }
+      )
       assert_equal :foo, static_file.type
       assert_equal "/dir/subdir/file.html", static_file.url
       assert_equal "/dir/subdir", static_file.destination_rel_dir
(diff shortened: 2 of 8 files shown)


envygeeks:
You can skip everything inside of `ansi.rb` recenly Simple::Ansi and Colorator merged and everything in that file will land in Jekyll once we upgrade Colorator so it'll probably be dropped in favor of just using our dependency which has way more stuff and is managed independently.

lib/jekyll/version.rb
```diff
@@ -1,3 +1,3 @@
 module Jekyll
-  VERSION = '3.1.6'
+  VERSION = "3.1.6".freeze
```
envygeeks: Anything to do with freezing strings can be skipped, there are plans to add the magic comments soon.

envygeeks:
Awesome, thanks for helping out!

lib/jekyll/version.rb
brint: Not sure what to do here. Revert the change and wait?

brint:
Would you like the `ansi.rb` changes reverted? I can do that. It was in the list for #4885.

envygeeks:
@brint nah that would be a waste of your effort, that was more of a note for anybody else.

test/test_static_file.rb
```diff
@@ -62,17 +63,16 @@ def setup_static_file_with_defaults(base, dir, name, defaults)
     end
 
     should "have a destination relative directory with a collection" do
-      static_file = setup_static_file_with_collection(
-        "root", "_foo/dir/subdir", "file.html", "foo", {"output" => true})
+      static_file = setup_static_file_with_collection("root", "_foo/dir/subdir",
+        "file.html", { "output" => true })
```
parkr: @brint Let's put each of these arguments on its own line:

```ruby
static_file = setup_static_file_with_collection(
  "root",
  "_foo/dir/subdir",
  "file.html",
  { "output" => true }
)

File: test/test_static_file.rb [hunk 3]
```diff
@@ -62,17 +63,16 @@ def setup_static_file_with_defaults(base, dir, name, defaults)
...
       assert_equal "/foo/dir/subdir", static_file.destination_rel_dir
     end
 
-    should "use its collection's permalink template for the destination relative directory" do
-      static_file = setup_static_file_with_collection(
-        "root", "_foo/dir/subdir", "file.html", "foo",
-        {"output" => true, "permalink" => "/:path/"})
+    should "use its collection's permalink template for destination relative directory" do
+      static_file = setup_static_file_with_collection("root", "_foo/dir/subdir",
+        "file.html", { "output" => true, "permalink" => "/:path/" })
```
parkr: Multiple lines here too:

```ruby
static_file = setup_static_file_with_collection(
  "root",
  "_foo/dir/subdir",
  "file.html",
  { "output" => true, "permalink" => "/:path/" }
)

File: test/test_static_file.rb [hunk 4]
```diff
@@ -86,13 +86,13 @@ def setup_static_file_with_defaults(base, dir, name, defaults)
...
       defaults = [{
-        "scope" => {"path" => "private"},
-        "values" => {"published" => false}
+        "scope"  => { "path" => "private" },
+        "values" => { "published" => false }
       }]
-      static_file = setup_static_file_with_defaults(
-        "root", "private/dir/subdir", "file.html", defaults)
+      static_file = setup_static_file_with_defaults("root", "private/dir/subdir",
+                                                    "file.html", defaults)
```
parkr: And here:

```ruby
static_file = setup_static_file_with_collection(
  "root",
  "private/dir/subdir",
  "file.html",
  defaults
)

parkr:
This is great! Just a couple comments above. Thank you for doing this, @brint!

lib/jekyll/version.rb
parkr: Leave as-is. When we do the magic comment for 2.3.0 and we drop support for <= 2.3.0, we'll remove this :smiel

brint:
@parkr Thanks for the feedback. The args have been split onto their own lines.

parkr:
Thanks again, @brint! You're on a roll!

@jekyllbot: merge +dev

Issue #4877, opened 2016-05-09
Valid markdown corrupted in meta-description tag


amw:
###### What version of Jekyll are you using (`jekyll -v`)?
3.1.3

###### What operating system are you using?
OS X

###### What did you do?
I've added this paragraph as the first one in the "welcome-to-jekyll.markdown" post:

``` markdown
This is an example paragraph
that is broken into a few
lines. New lines render as
spaces in HTML.
```

###### What did you expect to see?
I expected valid meta-description content in the head section of the generated post.

###### What did you see instead?
The meta tag contained word run-ins on the original line boundaries: `This is an example paragraphthat is broken into a fewlines. New lines render asspaces in HTML.`

###### Suggested fix
Problem is caused by `strip_newlines` that removes the "\n" characters instead of replacing them with spaces. Here:
https://github.com/jekyll/jekyll/blob/master/lib/site_template/_includes/head.html#L7

The bug was introduced in PR #2982. I think the solution is to either keep the newlines inside the description meta tag or replace it with a custom liquid filter like this one:

``` ruby
  def newlines_to_spaces input
    input.to_s.gsub(/\r?\n/, ' '.freeze)
  end
```

XhmikosR:
The reason was that many parsers had issues. Generally newlines are ignored.

My original patch should have a replace for newlines with a single space.

I generally didn't have such a use case so I didn't notice the issue before.

I'll need to verify all this though with a parser to make sure everything
is OK.
On May 12, 2016 23:11, "Parker Moore" <notifications@github.com> wrote:

/cc @XhmikosR <https://github.com/XhmikosR> who might be able to provide
context into why this newline stripping behaviour is necessary.

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/jekyll/jekyll/issues/4877#issuecomment-218871374>

XhmikosR:
So currently, this is what I'm using

```
{% capture meta_description %}{% if page.description %}{{ page.description }}{% if paginator.page and paginator.page != 1 %} - Page {{ paginator.page }}{% endif%}{% elsif page.excerpt %}{{ page.excerpt | newline_to_br | replace: '<br />', ' ' | strip_html | strip_newlines | replace: '  ', ' ' | rstrip | truncate: 155 }}{% else %}{{ site.description }}{% endif %}{% endcapture %}
```

It can be improved, but for my use case it does the job.

pathawks:
[...]

PR #4917, opened 2016-05-20
Add normalize_whitepace filter

pathawks:
See #4877

### DIFF
--- a/site/_docs/templates.md
+++ b/site/_docs/templates.md
@@ -258,6 +258,17 @@ common tasks easier.
     </tr>
     <tr>
       <td>
+        <p class="name"><strong>Normalize Whitespace</strong></p>
+        <p>Replace any occurance of whitespace with a single space.</p>
+      </td>
+      <td class="align-center">
+        <p>
+         <code class="filter">{% raw %}{{ "a \n b" | normalize_whitepace }}{% endraw %}</code>
+        </p>
+      </td>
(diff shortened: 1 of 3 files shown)


envygeeks:
Maybe we should also add `strip_whitespace` that does `\s+|\n+` to " "?

site/_docs/templates.md
```diff
@@ -258,6 +258,17 @@ common tasks easier.
     </tr>
     <tr>
       <td>
+        <p class="name"><strong>Normalize Whitespace</strong></p>
+        <p>Replace any occurance of whitespace with a single space.</p>
+      </td>
+      <td class="align-center">
+        <p>
+         <code class="filter">{% raw %}{{ "a \n b" | normalize_whitepace }}{% endraw %}</code>
```
parkr: Is it possible to add the resulting output?

parkr:
This looks more like a `strip_newlines` kind of change.

pathawks:
The problem is that `strip_newlines` will improperly concatenate words.

```text
cat
dog
```

becomes

```text
catdog
```

I would not mind just changing `strip_newlines` to behave this way, but others prefer the current behavior of `strip_newlines` (it does just what it says).

parkr:
@pathawks Ok! It's a new filter so I'm fine with it.

@jekyllbot: merge +minor

Issue #4923, opened 2016-05-21
Unable to run jekyll 3.2 pre beta 1


djacquel:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I Updated to the latest Jekyll "3.2.0.pre.beta1"

---

- [x] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux

---

## My Reproduction Steps

**Gemfile**

    source 'https://rubygems.org'
    gem 'jekyll', '3.2.0.pre.beta1'

Install

    bundle install (bundle is up to date)
    bundle exec jekyll new .
    bundler: failed to load command: jekyll (~/.rbenv/versions/2.3.0/bin/jekyll)
    NameError: uninitialized constant Jekyll::Commands
      ~/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/jekyll-3.2.0.pre.beta1/lib/jekyll/commands/new_theme.rb:3:in `<top (required)>'
      ~/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/jekyll-3.2.0.pre.beta1/lib/jekyll.rb:11:in `require'
      ~/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/jekyll-3.2.0.pre.beta1/lib/jekyll.rb:11:in `block in require_all'
      ~/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/jekyll-3.2.0.pre.beta1/lib/jekyll.rb:10:in `each'
      ~/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/jekyll-3.2.0.pre.beta1/lib/jekyll.rb:10:in `require_all'
      ~/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/jekyll-3.2.0.pre.beta1/lib/jekyll.rb:176:in `<top (required)>'
      ~/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/jekyll-3.2.0.pre.beta1/bin/jekyll:6:in `require'
      ~/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/jekyll-3.2.0.pre.beta1/bin/jekyll:6:in `<top (required)>'
      ~/.rbenv/versions/2.3.0/bin/jekyll:23:in `load'
[...]

PR #4924, opened 2016-05-21
fixes error in commands/new_theme.rb and conditionnaly create Gemfile

djacquel:
- Fixed `NameError: uninitialized constant Jekyll::Commands` in new_theme.rb (#4923)
- `jekyll new` create Gemfile if it doesn't already exist (#4925)

### DIFF
--- a/lib/jekyll/commands/new.rb
+++ b/lib/jekyll/commands/new.rb
@@ -91,7 +91,7 @@ RUBY
           end
 
           gemfile_path = File.expand_path("Gemfile", new_blog_path)
-          if !File.exist?(gemfile_path)
+          unless File.exist?(gemfile_path)
             File.open(gemfile_path, "w") do |f|
               f.write(gemfile_contents)
             end
--- a/lib/jekyll/commands/new.rb
+++ b/lib/jekyll/commands/new.rb
@@ -90,8 +90,11 @@ RUBY
             f.write(scaffold_post_content)
           end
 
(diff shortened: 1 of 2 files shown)


envygeeks:
@djacquel please update your message to reflect what is actually happening.

lib/jekyll/commands/new.rb
```diff
@@ -90,8 +90,11 @@ def create_site(new_blog_path)
             f.write(scaffold_post_content)
           end
 
-          File.open(File.expand_path("Gemfile", new_blog_path), "w") do |f|
-            f.write(gemfile_contents)
+          gemfile_path = File.expand_path("Gemfile", new_blog_path)
+          unless File.exist?(gemfile_path)
```
parkr: If we're generating a new site, we'll want to write this always. Using --force should indicate some "peril," so I would like us to write no matter what. Fewer support tickets if we know the right gemfile was created :)

lib/jekyll/commands/new_theme.rb
```diff
@@ -1,33 +1,36 @@
 require "erb"
+module Jekyll
+  module Commands
+    class NewTheme < Command
```
parkr: I was trying to avoid all the indentation here. We're going to move to using autoload, which should take care of not having Command.

djacquel: @parkr Yes, but this introduced a bug #4923.

PR #4931, opened 2016-05-24
Rubocop: converters

pathawks:


lib/jekyll/converters/markdown/redcarpet_parser.rb
```diff
@@ -17,19 +18,20 @@ def block_code(code, lang)
             Jekyll::External.require_with_graceful_fail("pygments")
             lang = lang && lang.split.first || "text"
             add_code_tags(
-              Pygments.highlight(code, :lexer => lang, :options => { :encoding => 'utf-8' }),
+              Pygments.highlight(code, :lexer => lang, :options => \
```
envygeeks: I think this should just be split up into a multiline hash.

lib/jekyll/converters/markdown/redcarpet_parser.rb
```diff
@@ -91,11 +87,29 @@ def class_with_proper_highlighter(highlighter)
         end
 
         def convert(content)
-          @redcarpet_extensions[:fenced_code_blocks] = !@redcarpet_extensions[:no_fenced_code_blocks]
-          @renderer.send :include, Redcarpet::Render::SmartyPants if @redcarpet_extensions[:smart]
-          markdown = Redcarpet::Markdown.new(@renderer.new(@redcarpet_extensions), @redcarpet_extensions)
+          @redcarpet_extensions[:fenced_code_blocks] = \
+            !@redcarpet_extensions[:no_fenced_code_blocks]
```
envygeeks: What happens if we shorten this class like previously discussed and do something like:

```
module Parent::NameSpace::Blah
  class NameSpace
  end
end
```

It seems like something that short really shouldn't be split off.  That or we could increase the limit again.  Seems like this limit is constantly sabotaging readability.

envygeeks:
Thanks! Just a few comments.

pathawks:
Whoops, I got cocky and neglected to run tests locally. Looks like I broke something :rage3:

lib/jekyll/converters/markdown/redcarpet_parser.rb
pathawks: That only gets us down to 95 characters 😕

lib/jekyll/converters/markdown/redcarpet_parser.rb
```diff
@@ -1,102 +1,104 @@
...
 
-        def convert(content)
-          @redcarpet_extensions[:fenced_code_blocks] = !@redcarpet_extensions[:no_fenced_code_blocks]
-          @renderer.send :include, Redcarpet::Render::SmartyPants if @redcarpet_extensions[:smart]
-          markdown = Redcarpet::Markdown.new(@renderer.new(@redcarpet_extensions), @redcarpet_extensions)
-          markdown.render(content)
-        end
+    def convert(content)
+      @redcarpet_extensions[:fenced_code_blocks] = \
+        !@redcarpet_extensions[:no_fenced_code_blocks]
```
pathawks: Without breaking this up, the line is 97 characters instead of 90 😕

pathawks:
#4885

pathawks:
@envygeeks Is there a reason `<figure>` is used [on line 32](https://github.com/jekyll/jekyll/blob/v3.1.6/lib/jekyll/converters/markdown/redcarpet_parser.rb#L32) but `<div>` is used [on line 45](https://github.com/jekyll/jekyll/blob/v3.1.6/lib/jekyll/converters/markdown/redcarpet_parser.rb#L45)?

#3779

lib/jekyll/converters/markdown/redcarpet_parser.rb
```diff
@@ -1,101 +1,109 @@
...
+      end
 
-        def convert(content)
-          @redcarpet_extensions[:fenced_code_blocks] = !@redcarpet_extensions[:no_fenced_code_blocks]
-          @renderer.send :include, Redcarpet::Render::SmartyPants if @redcarpet_extensions[:smart]
-          markdown = Redcarpet::Markdown.new(@renderer.new(@redcarpet_extensions), @redcarpet_extensions)
-          markdown.render(content)
+      def convert(content)
+        @redcarpet_extensions[:fenced_code_blocks] = \
+          !@redcarpet_extensions[:no_fenced_code_blocks]
```
pathawks: This would be 99 characters if it was all on one line.

lib/jekyll/converters/markdown/kramdown_parser.rb
```diff
@@ -100,7 +103,8 @@ def strip_coderay_prefix(hash)
         private
         def modernize_coderay_config
           if highlighter == "coderay"
-            Jekyll::Deprecator.deprecation_message "You are using 'kramdown.coderay' in your configuration, " \
+            Jekyll::Deprecator.deprecation_message \
+              "You are using 'kramdown.coderay' in your configuration, " \
               "please use 'syntax_highlighter_opts' instead."
```
pathawks: This warning is 92 characters by itself 😳
Perhaps we could use a `\n` in there?

lib/jekyll/converters/markdown/kramdown_parser.rb
```diff
@@ -100,8 +105,10 @@ def strip_coderay_prefix(hash)
         private
         def modernize_coderay_config
           if highlighter == "coderay"
-            Jekyll::Deprecator.deprecation_message "You are using 'kramdown.coderay' in your configuration, " \
+            Jekyll::Deprecator.deprecation_message(
+              "You are using 'kramdown.coderay' in your configuration, " \
               "please use 'syntax_highlighter_opts' instead."
```
pathawks: This warning is 101 characters by itself 😳
Perhaps we could use a `\n` in there?

lib/jekyll/converters/markdown/kramdown_parser.rb
```diff
@@ -68,8 +71,10 @@ def highlighter
 
           @highlighter = begin
             if @config.key?("enable_coderay") && @config["enable_coderay"]
-              Jekyll::Deprecator.deprecation_message "You are using 'enable_coderay', " \
+              Jekyll::Deprecator.deprecation_message(
+                "You are using 'enable_coderay', " \
                 "use syntax_highlighter: coderay in your configuration file."
```
pathawks: This warning is 91 characters by itself 😳
Perhaps we could use a `\n` in there?

parkr: Hehe yeah! Or if you can think of a way to trim down the message.

lib/jekyll/converters/markdown/kramdown_parser.rb
parkr: Yeah, could we just say "kramdown.coderay is deprecated. Use kramdown.syntax_highlighter_opts instead.", or is that still really long?

lib/jekyll/converters/markdown/redcarpet_parser.rb
```diff
@@ -1,101 +1,109 @@
...
+      module WithPygments
+        include CommonMethods
+        def block_code(code, lang)
+          Jekyll::External.require_with_graceful_fail("pygments")
+          lang = lang && lang.split.first || "text"
+          add_code_tags(
+            Pygments.highlight(
+              code,
+                :lexer   => lang,
+                :options => { :encoding => "utf-8" }
```
parkr: Could we wrap this hash with `{ }`?

parkr:
Wheeeee another one!!! Thank you!!!

@jekyllbot: merge +dev

PR #4343, opened 2016-01-10
Move Cucumber to using RSpec-Expections and furthering JRuby support.

envygeeks:
* Removes posix-spawn in favor of Open3#popen3
* Encapsulates all the paths into a single easy class.
* Moves to %r{} to avoid ambiguious warnings per-Cucumber suggestion.
* Starts passing around Pathname to make some actions faster.
* Clean's up some methods to make them easier to read.
* AUTOMATIC: Add "#" between each method.

PR #4934, opened 2016-05-24
Rubocop: features

pathawks:
#4885

### DIFF
--- a/features/support/formatter.rb
+++ b/features/support/formatter.rb
@@ -1,7 +1,7 @@
-require 'fileutils'
-require 'colorator'
-require 'cucumber/formatter/console'
-require 'cucumber/formatter/io'
+require "fileutils"
+require "colorator"
+require "cucumber/formatter/console"
+require "cucumber/formatter/io"
 
 module Jekyll
   module Cucumber
@@ -17,7 +17,7 @@ module Jekyll
         :undefined => "\u2718".red,
         :passed    => "\u2714".green,
         :skipped   => "\u203D".blue
-      }
+      }.freeze
 
       #
 
@@ -34,7 +34,7 @@ module Jekyll
 
       #
 
-      def before_features(features)
+      def before_features(_features)
         print_profile_information
       end
 
@@ -47,7 +47,7 @@ module Jekyll
 
       #
 
-      def before_feature(feature)
+      def before_feature(_feature)
         @exceptions = []
         @indent = 0
       end
@@ -55,20 +55,23 @@ module Jekyll
       #
 
       def tag_name(tag_name); end
+
       def comment_line(comment_line); end
+
       def after_feature_element(feature_element); end
+
       def after_tags(tags); end
 
       #
 
-      def before_feature_element(feature_element)
+      def before_feature_element(_feature_element)
         @indent = 2
         @scenario_indent = 2
       end
 
       #
 
-      def before_background(background)
+      def before_background(_background)
         @scenario_indent = 2
         @in_background = true
         @indent = 2
@@ -76,7 +79,7 @@ module Jekyll
 
       #
 
-      def after_background(background)
+      def after_background(_background)
         @in_background = nil
       end
 
@@ -104,8 +107,9 @@ module Jekyll
 
       #
 
-      def before_step_result(keyword, step_match, multiline_arg, status, exception, \
-          source_indent, background, file_colon_line)
+      # rubocop:disable Metrics/ParameterLists
+      def before_step_result(_keyword, _step_match, _multiline_arg, status, exception, \
+              _source_indent, background, _file_colon_line)
 
         @hide_this_step = false
         if exception
@@ -127,10 +131,11 @@ module Jekyll
 
       #
 
-      def step_name(keyword, step_match, status, source_indent, background, file_colon_line)
+      def step_name(_keyword, _step_match, status, _source_indent, _background, _file_colon_line)
         @io.print CHARS[status]
         @io.print " "
       end
+      # rubocop:enable Metrics/ParameterLists
 
       #
 
@@ -153,7 +158,7 @@ module Jekyll
       #
 
       private
-      def print_feature_element_name(keyword, name, source_line, indent)
+      def print_feature_element_name(keyword, name, source_line, _indent)
         @io.puts
 
         names = name.empty? ? [name] : name.each_line.to_a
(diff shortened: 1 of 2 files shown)


pathawks:
> ```text
> features/support/formatter.rb:110:29: C: Avoid parameter lists longer than 4 parameters.
>       def before_step_result(_keyword, _step_match, _multiline_arg, status, exception, \ ...
> 
> features/support/formatter.rb:133:20: C: Avoid parameter lists longer than 4 parameters.
>       def step_name(_keyword, _step_match, status, _source_indent, _background, _file_colon_line)
> ```

Is there any reason for the unused parameters? I could pull them out, but I just want to make sure they're not there to maintain consistency with an other API or something.

> ```text
> features/support/helpers.rb:99:1: C: Assignment Branch Condition size for run_in_shell is too high. [22.14/20]
> def run_in_shell(*args)
> ```

I am reluctant to modify [this function](https://github.com/jekyll/jekyll/blob/v3.1.6/features/support/helpers.rb#L94-L112) because it seems pretty concise. I am not sure if anything I could modify would make any improvement 😕

features/support/helpers.rb
```diff
@@ -1,3 +1,5 @@
+# rubocop:disable Metrics/AbcSize
```
pathawks: ⚠️ I am cheating

features/support/formatter.rb
```diff
@@ -1,7 +1,9 @@
-require 'fileutils'
-require 'colorator'
-require 'cucumber/formatter/console'
-require 'cucumber/formatter/io'
+# rubocop:disable Metrics/ParameterLists
```
pathawks: ⚠️ I am cheating

features/support/formatter.rb
```diff
@@ -107,6 +107,7 @@ def before_step(step)
 
       #
 
+      # rubocop:disable Metrics/ParameterLists
```
pathawks: ⚠️ I am cheating

features/support/helpers.rb
```diff
@@ -96,6 +96,7 @@ def run_jekyll(args)
 
 #
 
+# rubocop:disable Metrics/AbcSize
```
pathawks: ⚠️ I am cheating

features/support/helpers.rb
```diff
@@ -110,26 +114,27 @@ def run_in_shell(*args)
...
 
 def slug(title = nil)
   if !title
     then Time.now.strftime("%s%9N") # nanoseconds since the Epoch
-    else title.downcase.gsub(/[^\w]/, " ").strip.gsub(/\s+/, '-')
+  else title.downcase.gsub(/[^\w]/, " ").strip.gsub(/\s+/, "-")
   end
 end
 
 #
```
pathawks: @envygeeks What is the purpose of the `#` separating some methods?
Seems to have been added in #4343 #4347

envygeeks: I audit code so much in my daily life that normally add comments between methods and where there are no comments I either leave "# --" or "#" because it helps to spot a coming method quickly (lets admit it, we all have trouble finding a def at some point when looking through a large file or object.)

@parkr doesn't like it so when he mentions it I remove them (as my editor -- Atom -- has a plugin I built to automatically do that.) Some may argue that's why we have symbol finders, but I argue that when auditing code, I'm hardly interested in that and at a normal basis I am too because I feel that a file that cannot be browsed without a symbol browser and without syntax highlighting is no code that should be unleashed upon the world.

Long explanation, sorry.

features/support/formatter.rb
envygeeks: :100: Increase the size in Rubocop.  6 is a good number if you ask me, if it's over size I don't know what we should increase it to.  I vote increase it mostly because Rubocop does count kwd's, so since it's already impacting us without them, it'll impact us more with them if we ever do use them (which we do in a few places.)

features/support/formatter.rb
```diff
@@ -104,8 +107,9 @@ def before_step(step)
 
       #
 
-      def before_step_result(keyword, step_match, multiline_arg, status, exception, \
-          source_indent, background, file_colon_line)
+      # rubocop:disable Metrics/ParameterLists
```
pathawks: ⚠️ I am cheating

features/support/formatter.rb
```diff
@@ -104,8 +107,9 @@ def before_step(step)
...
+      def before_step_result(_keyword, _step_match, _multiline_arg, status, exception, \
+          _source_indent, background, _file_colon_line)
```
envygeeks: Indent to the first word and then indent twice past that:

```ruby
def before_step_result(args1,
        arg2)

  # Stuff
end
```

features/support/formatter.rb
```diff
@@ -34,7 +34,7 @@ def initialize(runtime, path_or_io, options)
 
       #
 
-      def before_features(features)
+      def before_features(_features)
```
envygeeks: Since this is an implementation on a reference I vote just do `(_)` instead of the full name.  People can always look at the reference class for the implementation that they need and what arguments it accepts, they just need know we don't care about them here.

envygeeks: I take that back, it looks like we aren't referencing their implementation as a sub-class.  Carry on.

envygeeks:
Thanks! Just a few comments.

features/support/formatter.rb
pathawks: ✅

pathawks:
Should be good to go :+1:

features/support/helpers.rb
```diff
@@ -149,13 +154,13 @@ def seconds_agnostic_datetime(datetime = Time.now)
...
-  .join("\\ ")
+    .join("\\ ")
 end
 
 #
 
 def seconds_agnostic_time(time)
   time = time.strftime("%H:%M:%S") if time.is_a?(Time)
-  hour, minutes, _ = time.split(":")
+  hour, minutes, = time.split(":")
```
parkr: huuuuh? this seems like a syntax error?

parkr:
Hm, it all works. Thanks again, @pathawks!!!

@jekyllbot: merge +dev

features/support/helpers.rb
pathawks: The other way seemed clearer to me too, but Rubocop wants what Rubocop wants.

Maybe when things settle down, I will dig into this a bit more to figure out _why_ RC wanted this changes.

pathawks: It is a violation of `Style/TrailingUnderscoreVariable`

I guess a "better" way to do this would be

```ruby
hour, minutes = time.split(":")
```

TWiStErRob: I don't know about RC, but some tools allow named exceptions to a rule, something like:
```ruby
hour, minutes, ignore = time.split(":")
hour, minutes, ignore_seconds = time.split(":")
```

Also http://stackoverflow.com/a/33006942/253468

PR #4936, opened 2016-05-24
Rubocop: Low hanging fruit

pathawks:
This PR removes files from the exclude list that are either already passing or can be automatically fixed by Rubocop

#4885


lib/jekyll/external.rb
```diff
@@ -23,7 +23,7 @@ def require_if_present(names, &block)
             require name
           rescue LoadError
             Jekyll.logger.debug "Couldn't load #{name}. Skipping."
-            block.call(name) if block
+            yield(name) if block
```
parkr: I think for this you have to remove the `&block` argument?

parkr:
@pathawks, you're my hero! ❤️ 

@jekyllbot: merge +dev

PR #4938, opened 2016-05-24
Rubocop: tags

ayastreb:
#4885

### DIFF
--- a/lib/jekyll/tags/highlight.rb
+++ b/lib/jekyll/tags/highlight.rb
@@ -65,7 +65,7 @@ eos
 
       def parse_options(input)
         options = {}
-        if defined?(input) && input != ""
+        unless input.empty?
           # Split along 3 possible forms -- key="<quoted list>", key=value, or key
           input.scan(/(?:\w="[^"]*"|\w=\w|\w)+/) do |opt|
             key, value = opt.split("=")
--- a/lib/jekyll/tags/highlight.rb
+++ b/lib/jekyll/tags/highlight.rb
@@ -112,7 +112,7 @@ eos
         Jekyll::External.require_with_graceful_fail("rouge")
         formatter = Rouge::Formatters::HTML.new(
           :line_numbers => @highlight_options[:linenos],
-          :wrap => false
+          :wrap         => false
         )
         lexer = Rouge::Lexer.find_fancy(@lang, code) || Rouge::Lexers::PlainText
         formatter.format(lexer.lex(code))
--- a/lib/jekyll/tags/highlight.rb
+++ b/lib/jekyll/tags/highlight.rb
@@ -93,7 +93,7 @@ eos
         )
 
         if highlighted_code.nil?
-          Jekyll.logger.error <<-eos
+          Jekyll.logger.error <<eos
 There was an error highlighting your code:
 
 #{code}
@@ -111,7 +111,8 @@ eos
       def render_rouge(code)
         Jekyll::External.require_with_graceful_fail("rouge")
         formatter = Rouge::Formatters::HTML.new(
-          :line_numbers => @highlight_options[:linenos], :wrap => false
+          :line_numbers => @highlight_options[:linenos],
+          :wrap => false
         )
         lexer = Rouge::Lexer.find_fancy(@lang, code) || Rouge::Lexers::PlainText
         formatter.format(lexer.lex(code))
--- a/lib/jekyll/tags/highlight.rb
+++ b/lib/jekyll/tags/highlight.rb
@@ -14,22 +14,9 @@ module Jekyll
         super
         if markup.strip =~ SYNTAX
           @lang = Regexp.last_match(1).downcase
-          @highlight_options = {}
-          if defined?(Regexp.last_match(2)) && Regexp.last_match(2) != ''
-            # Split along 3 possible forms -- key="<quoted list>", key=value, or key
-            Regexp.last_match(2).scan(/(?:\w="[^"]*"|\w=\w|\w)+/) do |opt|
-              key, value = opt.split('=')
-              # If a quoted list, convert to array
-              if value && value.include?("\"")
-                value.delete!('"')
-                  value = value.split
-              end
-              @highlight_options[key.to_sym] = value || true
-            end
-          end
-          @highlight_options[:linenos] = "inline" if @highlight_options.key?(:linenos) && @highlight_options[:linenos] == true
+          @highlight_options = parse_options(Regexp.last_match(2))
         else
-          raise SyntaxError.new <<-eos
+          raise SyntaxError, <<-eos
 Syntax Error in tag 'highlight' while parsing the following markup:
 
   #{markup}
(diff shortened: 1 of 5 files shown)


ayastreb:
Finished fixing all tags.
@pathawks , @envygeeks what do you think?

lib/jekyll/tags/highlight.rb
```diff
@@ -84,22 +93,26 @@ def render_pygments(code, is_safe)
...
         if highlighted_code.nil?
-          Jekyll.logger.error "There was an error highlighting your code:"
-          puts
-          Jekyll.logger.error code
-          puts
-          Jekyll.logger.error "While attempting to convert the above code, Pygments.rb" \
-            " returned an unacceptable value."
-          Jekyll.logger.error "This is usually a timeout problem solved by running `jekyll build` again."
-          raise ArgumentError.new("Pygments.rb returned an unacceptable value when attempting to highlight some code.")
+          Jekyll.logger.error <<-eos
```
pathawks: Oh wow, this is *much* improved! 😳:+1:

lib/jekyll/tags/highlight.rb
```diff
@@ -84,22 +93,26 @@ def render_pygments(code, is_safe)
...
-        highlighted_code.sub('<div class="highlight"><pre>', '').sub('</pre></div>', '')
+        highlighted_code.sub('<div class="highlight"><pre>', "").sub("</pre></div>", "")
       end
 
       def render_rouge(code)
-        Jekyll::External.require_with_graceful_fail('rouge')
-        formatter = Rouge::Formatters::HTML.new(:line_numbers => @highlight_options[:linenos], :wrap => false)
+        Jekyll::External.require_with_graceful_fail("rouge")
+        formatter = Rouge::Formatters::HTML.new(
+          :line_numbers => @highlight_options[:linenos], :wrap => false
```
pathawks: Is it better to have each on it's own line?

```ruby
formatter = Rouge::Formatters::HTML.new(
  :line_numbers => @highlight_options[:linenos],
  :wrap => false
)
```

lib/jekyll/tags/link.rb
```diff
@@ -16,11 +16,15 @@ def render(context)
           return document.url if document.relative_path == @relative_path
         end
 
-        raise ArgumentError, "Could not find document '#{@relative_path}' in tag '#{TagName}'.\n\n" \
-          "Make sure the document exists and the path is correct."
+        raise ArgumentError, <<eos
+Could not find document '#{@relative_path}' in tag '#{TAG_NAME}'.
+
+
```
pathawks: Do we want *two* blank lines here, or just one?

It looks like the first `\n` terminated the first line, and the second `\n` inserted one blank line, but I might be reading it wrong.

pathawks:
Thank you *so much* for all your hard work on this! Some of those changes make a **huge** improvement in readability! 🍻

lib/jekyll/tags/highlight.rb
ayastreb: Yes, I think each argument on it's own line would be better 👍

lib/jekyll/tags/link.rb
ayastreb: Yes, right! I'll remove the extra line 😄

ayastreb:
Hm, Travis build fails with rubocop error in file drops/drop.rb but I didn't change anything there 😕 
Was it changed in master branch?

lib/jekyll/tags/highlight.rb
```diff
@@ -14,22 +14,9 @@ def initialize(tag_name, markup, tokens)
...
-                  value = value.split
-              end
-              @highlight_options[key.to_sym] = value || true
-            end
-          end
-          @highlight_options[:linenos] = "inline" if @highlight_options.key?(:linenos) && @highlight_options[:linenos] == true
+          @highlight_options = parse_options(Regexp.last_match(2))
         else
-          raise SyntaxError.new <<-eos
+          raise SyntaxError, <<-eos
```
envygeeks: I think we provide a heredoc stripper in the Utils class so that you can indent.  I digress though on it's usage, I like to use it but if you prefer not to I'm okay with that too!

lib/jekyll/tags/highlight.rb
```diff
@@ -66,16 +53,38 @@ def sanitized_opts(opts, is_safe)
...
         else
           opts
         end
       end
 
+      private
+
+      def parse_options(input)
+        options = {}
+        if defined?(input) && input != ""
```
envygeeks: `input` will always be defined because it's in the signature.

lib/jekyll/tags/highlight.rb
```diff
@@ -110,13 +124,14 @@ def render_codehighlighter(code)
 
       def add_code_tag(code)
         code_attributes = [
-          "class=\"language-#{@lang.to_s.tr('+', '-')}\"",
+          "class=\"language-#{@lang.to_s.tr("+", "-")}\"",
```
envygeeks: It's preferred to use `%Q{}` here, did it not warn about this? I don't know if Rubocop can but it it didn't we should probably check and see if it can, I'll admit though that this might be a limitation in static analysis (it should be though IMO.)

lib/jekyll/tags/highlight.rb
```diff
@@ -110,13 +124,14 @@ def render_codehighlighter(code)
...
           "data-lang=\"#{@lang}\""
         ].join(" ")
-        "<figure class=\"highlight\"><pre><code #{code_attributes}>#{code.chomp}</code></pre></figure>"
+        "<figure class=\"highlight\"><pre><code #{code_attributes}>"\
```
envygeeks: ""

lib/jekyll/tags/include.rb
```diff
@@ -12,17 +12,23 @@ def initialize(msg, path)
...
         super
         matched = markup.strip.match(VARIABLE_SYNTAX)
         if matched
-          @file = matched['variable'].strip
-          @params = matched['params'].strip
+          @file = matched["variable"].strip
+          @params = matched["params"].strip
         else
-          @file, @params = markup.strip.split(' ', 2)
+          @file, @params = markup.strip.split(" ", 2)
```
envygeeks: This might be a good time to fix that bad split.  Users are prone to alignment because it's prettier, if you don't mind can you make this `markup.strip.split(/\s+/, 2)` that way users can infinity space and align any way they see fit.  I don't know if it's worth it though considering it's not been a problem before?

envygeeks:
Thanks! Just a few comments!

lib/jekyll/tags/highlight.rb
ayastreb: Do we need that check at all? Maybe it would be enough to do it like this:
`
unless input == ""
`

pathawks:
```ruby
unless input.empty?
```

?

lib/jekyll/tags/highlight.rb
ayastreb: No, it didn't warn, Rubocop only noticed single quotes here.
Sorry, I'm not too deep in Ruby yet, trying to learn it here 😄 
Do you suggest to create this code tag with single interpolation string?
Like
`
%Q{<figure class="highlight"><pre><code class="language-#{@lang.to_s.tr("+", "-")}" data-lang="#{@lang}">#{code.chomp}</code></pre></figure>}
` 
? It would be too long then. And if I change it to multiline string with heredoc then a lot of tests are failing because they expect a single line string.
Should we change it to multiline here and update the tests?

Issue #4911, opened 2016-05-18
Port changes in v3.1.4 & v3.1.5 & v3.1.6 forward


parkr:
We improved v3.1.3 for GitHub Pages and released them as v3.1.{4,5,6}. We should forward-port all those changes. From the changelog:

  * [x] DONE in #4941 – Add `ExcerptDrop` and remove excerpt's ability to refer to itself in Liquid (#4907)
  * [x] DONE IN #4942 – Configuration permalink fix where `collections.posts.permalink` inherits properly from `permalink` only when it doesn't exist (#4910)
  * [x] DONE IN #4942 – Add `Configuration.from` to make it easier to build configs from just a hash
  * [x] DONE IN #4942 – Sorting `site.collections` in Liquid by label (#4910)
  * [x] DONE IN #4943 – Fix bug where `layout` in Liquid would inherit from previously-rendered layouts' metadatas (#4909)
  * [x] DONE IN #4943 – Fix bug where `layout` in Liquid would override in the wrong direction (more-specific layouts' data were overwritten by their parent layouts' data; this has now been reversed) (#4909)
  * [x] DONE IN #4944 – Sort the results of the `require_all` glob (affects Linux only). (#4912)
  * [x] DONE IN #4945 – 3.1: Add ability to render drops as JSON (#4914)

parkr:
Many thanks to @pathawks for taking on all of these all in one day! WOWZA. 🍻 

![](http://gifs.joelglovier.com/excited/leo-toast.gif)

PR #4942, opened 2016-05-24
Forward-port: Configuration permalink fix and addition of Configuration.from and sorting `site.collections` by label

pathawks:
Merge #4910 into **master**

#4911


parkr:
Yay!!! Thanks so much, @pathawks! ❤️ 

@jekyllbot: merge +port

PR #4953, opened 2016-05-26
`Serve.process` should receive same config as `Build.process`

### LINKED ISSUE #4850: Second configuration file is being ignored with --config _config.yml,_second_config.yml

`3.0.5`

OSX

###### What did you do?

I have two configuration files. `_config.yml` and `_development.yml`. `_development.yml` includes the configuration `port: 1337`.

I'm running the command `jekyll serve --config _config.yml,_development.yml`

I expected the command `jekyll serve --config _config.yml,_development.yml` to serve the site from `http://localhost:1337`

The `port` configuration is ignored, and the site is served from the default `http://localhost:4000`

###### Notes

I noticed this after running `bundle update github-pages`. Previously I was on Jekyll `3.0.3` which worked as expected. On `3.0.5`, the `port` configuration is still honored from yaml if it is in the first configuration file, but not in the second.

Crunch09:
**Before**:
bildschirmfoto 2016-05-27 um 00 01 43

**After**:
bildschirmfoto 2016-05-27 um 00 02 36

The config option is removed in the [build process](https://github.com/jekyll/jekyll/blob/632f3fd8de0c63813f46240216cb12905d829968/lib/jekyll/configuration.rb#L146) so it was not available for the serve command anymore. I tried to replace the `delete` in this line with a `fetch` at first but this was causing other tests to fail. 

fixes #4850

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -33,7 +33,9 @@ module Jekyll
             cmd.action do |_, opts|
               opts["serving"] = true
               opts["watch"  ] = true unless opts.key?("watch")
+              config = opts["config"]
               Build.process(opts)
+              opts["config"] = config
               Serve.process(opts)
             end
           end
--- a/test/test_commands_serve.rb
+++ b/test/test_commands_serve.rb
@@ -74,6 +74,20 @@ class TestCommandsServe < JekyllUnitTest
         assert custom_opts(opts)[:DirectoryIndex].empty?
       end
 
+      should "keep config between build and serve" do
+        custom_options = {
+          "config"  => %w(_config.yml _development.yml),
+          "serving" => true,
+          "watch"   => false # for not having guard output when running the tests
+        }
+        allow(SafeYAML).to receive(:load_file).and_return({})
+        allow(Jekyll::Commands::Build).to receive(:build).and_return("")
+
+        expect(Jekyll::Commands::Serve).to receive(:process).with(custom_options)
+        @merc.execute(:serve, { "config" => %w(_config.yml _development.yml),
+                                "watch"  => false })
+      end
+
       context "verbose" do
         should "debug when verbose" do
           assert_equal custom_opts({ "verbose" => true })[:Logger].level, 5


lib/jekyll/commands/serve.rb
```diff
@@ -33,7 +33,9 @@ def init_with_program(prog)
             cmd.action do |_, opts|
               opts["serving"] = true
               opts["watch"  ] = true unless opts.key?("watch")
+              cfg = opts["config"]
```
envygeeks: The short we normally use in Ruby is `config`.

lib/jekyll/commands/serve.rb
```diff
@@ -33,7 +33,9 @@ def init_with_program(prog)
...
               Build.process(opts)
+              opts["config"] = cfg
```
envygeeks: Why does this need to happen? This is not only highly disoptimized by I don't understand why it has to happen and if it does I think there is a much bigger problem that needs to be addressed that either should be addressed here or this blocked on that being fixed.

Crunch09: Thanks for the review! As i said in my initial comment the config option gets lost in the build process on this specific [line](https://github.com/jekyll/jekyll/blob/632f3fd8de0c63813f46240216cb12905d829968/lib/jekyll/configuration.rb#L146) and the serve calls, which is started after the build process has finished has no information about the given configs anymore. I tried to replace the `delete` call with `fetch` but this was causing other tests to fail so i took another approach.

parkr:
Your test has a formatting error! Would you mind fixing that up, please?

Crunch09:
@parkr Of course! Sorry, i should've run the formatting tests locally first.

test/test_commands_serve.rb
```diff
@@ -76,15 +76,16 @@ def custom_opts(what)
 
       should "keep config between build and serve" do
         custom_options = {
-          'config'  => ['_config.yml', '_development.yml'],
-          'serving' => true,
-          'watch'   => false # for not having guard output when running the tests
+          "config"  => ["_config.yml", "_development.yml"],
```
envygeeks: This should be `%w(_config.yml _development.yml)` This is a bug in our `.rubocop.yml`

Crunch09: Thanks!

envygeeks:
Even though I don't like we have to reset the options, this is GTG.  :shipit:

tfe:
I'm not familiar with the Jekyll release process, but I've got some stuff dependent on this fix being merged. When do you think it'll make it into a release?

parkr:
@tfe: Merging now. Won't be in a release until v3.2.0.

@jekyllbot: merge +bug

TWiStErRob:
For anyone finding this issue/fix: I tested a lot of different versions and this broke between [3.0.3->3.0.4](https://github.com/jekyll/jekyll/compare/v3.0.3...v3.0.4) and between [3.1.3->3.1.4](https://github.com/jekyll/jekyll/compare/v3.1.3...v3.1.4). I discovered and researched this with `baseurl` not being picked up.

This means that the following versions cannot be used for local development properly:
Jekyll 3.0.4, 3.0.5 (pages v76-78), 3.1.4, 3.1.5 (pages v79), 3.1.6 (pages v80+)

@envygeeks if you don't like resetting options, maybe there's a different fix clue hidden in those diffs.

PR #4956, opened 2016-05-27
rubocop: features/step_definitions.rb

ayastreb:
Fixes in features/step_definitions.rb
#4885

### DIFF
--- a/features/step_definitions.rb
+++ b/features/step_definitions.rb
@@ -93,11 +93,12 @@ end
 #
 
 Given(/^I have a configuration file with "(.*)" set to "(.*)"$/) do |key, value|
-  config = if source_dir.join("_config.yml").exist?
-             SafeYAML.load_file(source_dir.join("_config.yml"))
-           else
-             {}
-           end
+  config = \
+    if source_dir.join("_config.yml").exist?
+      SafeYAML.load_file(source_dir.join("_config.yml"))
+    else
+      {}
+    end
   config[key] = YAML.load(value)
   File.write("_config.yml", YAML.dump(config))
 end
--- a/features/step_definitions.rb
+++ b/features/step_definitions.rb
@@ -14,24 +14,24 @@ end
 
 #
 
-Given %r{^I have a blank site in "(.*)"$} do |path|
-  if !File.exist?(path)
+Given(/^I have a blank site in "(.*)"$/) do |path|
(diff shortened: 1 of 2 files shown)


features/step_definitions.rb
```diff
@@ -14,24 +14,24 @@
 
 #
 
-Given %r{^I have a blank site in "(.*)"$} do |path|
-  if !File.exist?(path)
+Given(/^I have a blank site in "(.*)"$/) do |path|
```
envygeeks: This should still fail, we use `%r!!` over `//` or `%r{}`.

features/step_definitions.rb
```diff
@@ -92,27 +92,27 @@
...
 #
 
-Given %r{^I have a configuration file with "(.*)" set to "(.*)"$} do |key, value|
+Given(/^I have a configuration file with "(.*)" set to "(.*)"$/) do |key, value|
   config = if source_dir.join("_config.yml").exist?
-    SafeYAML.load_file(source_dir.join("_config.yml"))
-  else
-    {}
-  end
+             SafeYAML.load_file(source_dir.join("_config.yml"))
```
envygeeks: That's over tabbed and breaks readability, no matter what Rubocop tries to tell you and because it's not customizable more in this aspect we do:

```ruby
val = \
  if blah
    work(
      here
    )
  end
```

envygeeks:
Thanks! Just a few comments.

features/step_definitions.rb
ayastreb: Hm, actually it failed with %r!!, I guess because we have following in .rubocop.yml
```yaml
Style/RegexpLiteral:
  EnforcedStyle: slashes
```
Should I change it to `percent_r` ?

envygeeks: I'll digress to @parkr but we've been moving to `%r!!` all over the place.

pathawks: Only when there is a literal `/` in the regex

pathawks:
LGTM :shipit:

features/step_definitions.rb
pathawks: Which is to say, `/test/` is fine, but `/test\//` is ugly so in that case we use `%r!test/!` instead.

parkr:
Thank you so much!

@jekyllbot: merge +dev

PR #4958, opened 2016-05-28
markdownify and smartify should convert input to string before conversion

pathawks:
If I have a page with the following front matter:

```yaml
layout: default
title: 404
```

and a layout called `_layouts/default.html` with the following contents:

```text
{{ page.title | markdownify }}
```

Jekyll will fail to build the site, throwing the error:

```text
NoMethodError: undefined method `gsub' for 404:Fixnum
```

### DIFF
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -13,7 +13,7 @@ module Jekyll
     def markdownify(input)
       site = @context.registers[:site]
       converter = site.find_converter_instance(Jekyll::Converters::Markdown)
-      converter.convert(input)
+      converter.convert(input.to_s)
     end
 
     # Convert quotes into smart quotes.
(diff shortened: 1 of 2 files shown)


pathawks:
Thank you, @jekyllbot, for removing a label that didn't exist every time I force push 😜

envygeeks:
Hulk here,

It looks like @jekyllbot doesn't feel like responding to you.  So let me say "you're welcome" on it's behalf.

envygeeks:
Pretty sure you just got trolled by my bot accidentally.  Can't tell.

pathawks:
Even though there is a green checkmark next to my "Failing test" I can assure you that it does indeed fail. I forgot that I need to let Travis run all the tests on the failing commit before I push the fix.

envygeeks:
@jekyllbot: merge +dev

envygeeks:
I think I did that wrong? /cc @parkr

parkr:
@envygeeks That's right! Use +major, +minor, +bug, +site, or +dev. This would be a `+bug` from my perspective.

PR #4959, opened 2016-05-28
Rubocop theme and url jekyll libs

derekgottlieb:
Quick pass at trying to get lib/jekyll/theme and lib/jekyll/url passing rubocop for #4885

### DIFF
--- a/lib/jekyll/theme.rb
+++ b/lib/jekyll/theme.rb
@@ -27,7 +27,7 @@ module Jekyll
 
     def configure_sass
       return unless sass_path
-      require 'sass'
+      require "sass"
       Sass.load_paths << sass_path
     end
 
@@ -38,7 +38,7 @@ module Jekyll
       return unless resolved_dir
 
       path = Jekyll.sanitized_path(root, resolved_dir)
-      path if Dir.exists?(path)
+      path if Dir.exist?(path)
     end
 
     def realpath_for(folder)
@@ -50,7 +50,8 @@ module Jekyll
     def gemspec
       @gemspec ||= Gem::Specification.find_by_name(name)
     rescue Gem::LoadError
-      raise Jekyll::Errors::MissingDependencyException, "The #{name} theme could not be found."
+      raise Jekyll::Errors::MissingDependencyException,
+        "The #{name} theme could not be found."
     end
   end
 end
--- a/lib/jekyll/theme.rb
+++ b/lib/jekyll/theme.rb
@@ -38,7 +38,7 @@ module Jekyll
       return unless resolved_dir
 
       path = Jekyll.sanitized_path(root, resolved_dir)
-      path if Dir.exist?(path)
+      path if File.directory?(path)
     end
 
     def realpath_for(folder)
--- a/lib/jekyll/url.rb
+++ b/lib/jekyll/url.rb
@@ -92,7 +92,7 @@ module Jekyll
     # as well as the beginning "/" so we can enforce and ensure it.
 
     def sanitize_url(str)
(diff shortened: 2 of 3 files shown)


lib/jekyll/theme.rb
```diff
@@ -38,7 +38,7 @@ def path_for(folder)
       return unless resolved_dir
 
       path = Jekyll.sanitized_path(root, resolved_dir)
-      path if Dir.exists?(path)
+      path if Dir.exist?(path)
```
envygeeks: Use `File.directory?` it's straight to the point.  `Dir.exist?` actually points to `File.directory?`

lib/jekyll/url.rb
```diff
@@ -67,32 +67,32 @@ def generate_url(template)
 
     def generate_url_from_hash(template)
       @placeholders.inject(template) do |result, token|
-        break result if result.index(':').nil?
+        break result if result.index(":").nil?
         if token.last.nil?
-          # Remove leading '/' to avoid generating urls with `//`
-          result.gsub(/\/:#{token.first}/, '')
+          # Remove leading "/" to avoid generating urls with `//`
+          result.gsub(%r!/:#{token.first}!, "")
```
envygeeks: Why isn't this escaped? /cc @jekyll/core.

lib/jekyll/url.rb
```diff
@@ -67,32 +67,32 @@ def generate_url(template)
...
           result.gsub(/:#{token.first}/, self.class.escape_path(token.last))
         end
       end
     end
 
     def generate_url_from_drop(template)
-      template.gsub(/:([a-z_]+)/.freeze) do |match|
-        replacement = @placeholders.public_send(match.sub(':'.freeze, ''.freeze))
+      template.gsub(/:([a-z_]+)/) do |match|
+        replacement = @placeholders.public_send(match.sub(":".freeze, "".freeze))
```
envygeeks: If we are gonna do `"".freeze` everywhere, we need do like Liquid does and create EMPTY_STRING constant, because this is nasty and overwhelming. /cc @jekyll/core

lib/jekyll/url.rb
```diff
@@ -67,32 +67,32 @@ def generate_url(template)
...
-      end.gsub(/\/\//.freeze, '/'.freeze)
+      end.gsub(%r!//!, "/".freeze)
     end
 
     # Returns a sanitized String URL, stripping "../../" and multiples of "/",
     # as well as the beginning "/" so we can enforce and ensure it.
 
     def sanitize_url(str)
-      "/" + str.gsub(/\/{2,}/, "/").gsub(/\.+\/|\A\/+/, "")
+      "/" + str.gsub(%r!/{2,}!, "/").gsub(%r!\.+\/|\A\/+!, "")
```
envygeeks: `%r!\.+/|\A/+!`

envygeeks:
Thanks! Just a few comments, some for you and the others for @jekyll/core.

lib/jekyll/url.rb
derekgottlieb: Good catch. Would it be worth adding a test for this method since nothing caught this?

pathawks: `%r!\.+/|\A/+!` Now that we’re not using `/` to terminate the regex :+1:

lib/jekyll/url.rb
parkr: @envygeeks Why isn't `token.first` escaped? Not sure.

envygeeks:
:shipit:

lib/jekyll/url.rb
parkr: @derekgottlieb A test would be wonderful, yes!

parkr:
Thanks, Derek!

@jekyllbot: merge +dev

PR #4960, opened 2016-05-29
rubocop: lib/jekyll/document.rb

ayastreb:
#4885

### DIFF
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -225,8 +225,8 @@ module Jekyll
 
     # Whether the file is published or not, as indicated in YAML front-matter
     #
-    # Returns true if the 'published' key is
-    # specified in the YAML front-matter and not `false`.
+    # Returns 'false' if the 'published' key is specified in the
+    # YAML front-matter and is 'false'. Otherwise returns 'true'.
     def published?
       !(data.key?("published") && data["published"] == false)
     end
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -1,7 +1,6 @@
 # encoding: UTF-8
 
 module Jekyll
-  # rubocop:disable ClassLength
   class Document
     include Comparable
 
@@ -48,30 +47,6 @@ module Jekyll
       @data ||= {}
     end
 
-    # Merge some data in with this document's data.
-    #
-    # Returns the merged data.
-    def merge_data!(other, source: "YAML front matter")
-      merge_categories!(other)
-      Utils.deep_merge_hashes!(data, other)
-      if data.key?("date") && !data["date"].is_a?(Time)
-        data["date"] = Utils.parse_date(
-          data["date"].to_s,
-          "Document '#{relative_path}' does not have a valid date in the #{source}."
-        )
-      end
-      data
-    end
-
-    def merge_categories!(other)
-      if other.key?("categories") && !other["categories"].nil?
-        if other["categories"].is_a?(String)
-          other["categories"] = other["categories"].split(" ").map(&:strip)
-        end
-        other["categories"] = (data["categories"] || []) | other["categories"]
-      end
-    end
-
     def date
       data["date"] ||= (draft? ? source_file_mtime : site.time)
     end
@@ -111,7 +86,7 @@ module Jekyll
     #
     # Returns the basename without the file extname.
     def basename_without_ext
-      @basename_without_ext ||= File.basename(path, ".*")
+      @basename_without_ext ||= File.basename(path, File.extname(path))
     end
 
     # The base filename of the document.
@@ -164,7 +139,7 @@ module Jekyll
     #
     # Returns true if extname == .coffee, false otherwise.
     def coffeescript_file?
-      ".coffee".eql?(extname)
+      ".coffee" == extname
     end
 
     # Determine whether the file should be rendered with Liquid.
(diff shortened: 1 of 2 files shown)


lib/jekyll/document.rb
```diff
@@ -1,15 +1,16 @@
 # encoding: UTF-8
 
 module Jekyll
+  # rubocop:disable ClassLength
```
ayastreb: I couldn't think of a way to reduce the size of this class (it's 275 lines now, max is 240) 😞 
Any ideas?)

envygeeks: @ayastreb increase the max class size to `300` in `.rubocop.yml` that was an arbitrary number I just randomly set and @parkr decided to keep, I see no problem with raising it.

lib/jekyll/document.rb
```diff
@@ -51,24 +52,28 @@ def data
...
+      if data.key?("date") && !data["date"].is_a?(Time)
+        data["date"] = Utils.parse_date(
+          data["date"].to_s,
           "Document '#{relative_path}' does not have a valid date in the #{source}."
         )
       end
       data
     end
 
+    def merge_categories!(other)
```
envygeeks: `private`, because this belongs to `merge_data!`

lib/jekyll/document.rb
```diff
@@ -51,24 +52,28 @@ def data
...
+      if other.key?("categories") && !other["categories"].nil?
+        if other["categories"].is_a?(String)
+          other["categories"] = other["categories"].split(" ").map(&:strip)
```
envygeeks: Now would be a good time to fix that bad split to `\s+`

lib/jekyll/document.rb
```diff
@@ -81,15 +86,18 @@ def source_file_mtime
...
     end
 
     # The path to the document, relative to the site source.
     #
     # Returns a String path which represents the relative path
     #   from the site source to this document
     def relative_path
-      @relative_path ||= Pathname.new(path).relative_path_from(Pathname.new(site.source)).to_s
+      @relative_path ||= Pathname.new(path)
+        .relative_path_from(Pathname.new(site.source))
```
envygeeks: `Pathutil.new(path).relative_path_from(site.source).to_s` // Note to self.

lib/jekyll/document.rb
```diff
@@ -103,7 +111,7 @@ def output_ext
     #
     # Returns the basename without the file extname.
     def basename_without_ext
-      @basename_without_ext ||= File.basename(path, '.*')
+      @basename_without_ext ||= File.basename(path, ".*")
```
envygeeks: `File.basename(path, File.extname(path))` is much more clear to non-programmers.

lib/jekyll/document.rb
```diff
@@ -156,7 +164,7 @@ def sass_file?
     #
     # Returns true if extname == .coffee, false otherwise.
     def coffeescript_file?
-      '.coffee'.eql?(extname)
+      ".coffee".eql?(extname)
```
envygeeks: `==`

lib/jekyll/document.rb
```diff
@@ -237,7 +245,7 @@ def destination(base_directory)
     def write(dest)
       path = destination(dest)
       FileUtils.mkdir_p(File.dirname(path))
-      File.open(path, 'wb') do |f|
+      File.open(path, "wb") do |f|
```
envygeeks: `File.write(output)` is the same thing without the extra lines.

lib/jekyll/document.rb
```diff
@@ -263,47 +272,61 @@ def read(opts = {})
...
-            raise e
-          end
+        rescue StandardError => e
+          raise e if e.is_a? Jekyll::Errors::FatalException
           Jekyll.logger.error "Error:", "could not read file #{path}: #{e.message}"
         end
       end
     end
 
+    def merge_defaults
```
envygeeks: `private`

lib/jekyll/document.rb
```diff
@@ -263,47 +272,61 @@ def read(opts = {})
...
+          merge_defaults
+          read_content(opts)
           post_read
         rescue SyntaxError => e
           Jekyll.logger.error "Error:", "YAML Exception reading #{path}: #{e.message}"
-        rescue Exception => e
-          if e.is_a? Jekyll::Errors::FatalException
-            raise e
-          end
+        rescue StandardError => e
```
envygeeks: Please revert this and send a separate pull request for this, this has been brought up by me before but there is a larger surface that needs to be tested and thought about, @parkr has his reasons for this and this will block your pull because it's a feature change.

lib/jekyll/document.rb
```diff
@@ -263,47 +272,61 @@ def read(opts = {})
...
 
+    def merge_defaults
+      defaults = @site.frontmatter_defaults.all(
+        relative_path,
+        collection.label.to_sym
+      )
+      merge_data!(defaults, :source => "front matter defaults") unless defaults.empty?
+    end
+
+    def read_content(opts)
```
envygeeks: `private`

lib/jekyll/document.rb
```diff
@@ -263,47 +272,61 @@ def read(opts = {})
...
+    end
+
     def post_read
+      populate_title
+      populate_categories
+      populate_tags
+      generate_excerpt
+    end
+
+    def populate_title
```
envygeeks: `private`

lib/jekyll/document.rb
```diff
@@ -263,47 +272,61 @@ def read(opts = {})
...
-      data['slug'] ||= slug
-      data['ext']  ||= ext
+      data["slug"] ||= slug
+      data["ext"]  ||= ext
+    end
 
-      populate_categories
-      populate_tags
-      generate_excerpt
+    def modify_date(date)
```
envygeeks: `private`

lib/jekyll/document.rb
```diff
@@ -312,16 +335,22 @@ def post_read
     #
     # Returns nothing.
     def categories_from_path(special_dir)
-      superdirs = relative_path.sub(/#{special_dir}(.*)/, '').split(File::SEPARATOR).reject do |c|
+      superdirs = relative_path.sub(/#{special_dir}(.*)/, "")
+        .split(File::SEPARATOR)
+        .reject do |c|
         c.empty? || c.eql?(special_dir) || c.eql?(basename)
```
envygeeks: `==`

lib/jekyll/document.rb
```diff
@@ -394,17 +423,13 @@ def next_doc
       pos = collection.docs.index { |post| post.equal?(self) }
```
envygeeks: `is_a?`

envygeeks:
Thanks! Just a few comments.

lib/jekyll/document.rb
ayastreb: Ok, I will revert this and add Rubocop ignore for this

lib/jekyll/document.rb
ayastreb: Should I move all the private methods to the bottom of the class below `private`, or do we prefer to leave methods where they are and defined them private, e.g.
```
def merge_data!(other, source: "YAML front matter")
...
end

def merge_categories!(other)
...
end 
private :merge_categories
```
?

envygeeks: At the bottom of the class and each one gets an explicit private.  Rubocop doesn't like that because they consider it useless but I consider it good form to be explicit because one could easily miss a private 20 methods away when quickly viewing a file.

ayastreb: Sorry, when you say each one gets an explicit private, you mean something like this:
```
private
def merge_categories!(other)
...
end

private
def merge_defaults
...
end
```
?
I've checked other files and it seems to be only one `private` keyword per file everywhere.
Just want to make sure I understand you correctly 😄

envygeeks: That is what I mean, see: https://github.com/jekyll/jekyll/blob/master/lib/jekyll/converters/markdown/kramdown_parser.rb, https://github.com/jekyll/jekyll/blob/master/lib/jekyll/commands/serve.rb

lib/jekyll/document.rb
```diff
@@ -237,18 +218,17 @@ def destination(base_directory)
...
     end
 
     # Whether the file is published or not, as indicated in YAML front-matter
     #
-    # Returns true if the 'published' key is specified in the YAML front-matter and not `false`.
+    # Returns true if the 'published' key is
+    # specified in the YAML front-matter and not `false`.
     def published?
-      !(data.key?('published') && data['published'] == false)
+      !(data.key?("published") && data["published"] == false)
```
pathawks: I realize you didn’t change any functionality here, but this line seems to conflict with the description:

> Returns true if the 'published' key is specified in the YAML front-matter and not 'false'.

But it also returns `true` if `published` is _not_ specified.

lib/jekyll/document.rb
```diff
@@ -263,75 +243,19 @@ def read(opts = {})
...
-            merge_data!(data_file, source: "YAML front matter") if data_file
-          end
-
-          post_read
+          merge_defaults
+          read_content(opts)
+          read_data
         rescue SyntaxError => e
           Jekyll.logger.error "Error:", "YAML Exception reading #{path}: #{e.message}"
+        # rubocop: disable RescueException
```
pathawks: Please be sure to enable this again immediately after this block.

lib/jekyll/document.rb
ayastreb: Hmm.. should I fix the comment or the method? 😄 
This logic is also duplicated in `Jekyll::Convertible.published?`, it has a different comment there though)

lib/jekyll/document.rb
ayastreb: Thx, I forgot to enable it back here 😊

lib/jekyll/document.rb
pathawks: The comment, I think.

```ruby
# Returns 'false' if the 'published' key is specified in the YAML front-matter and is 'false'. Otherwise returns 'true'.
```

¯\\\_(ツ)_/¯

ayastreb:
Merged these changes in PR #4961

PR #4961, opened 2016-05-29
rubocop: document.rb, convertible.rb and renderer.rb

ayastreb:
#4885 

Ok, now I think I'm finished with this 😄 
Main goal of this PR is to remove code duplication in `document.rb`, `renderer.rb` and `convertible.rb` and make them all Rubocop compliant.

I included `Convertible` into `Document` and implemented `render` method the same way it is implemented in `Page`
I removed all the logic from `Renderer` class and just call this `render` method on the document instance.

I did some refactoring in `convertible.rb`, mainly:
- move private methods to the bottom of the class
- move payload from methods argument to class attribute

Feedback is welcome! 😄

### DIFF
--- a/lib/jekyll/convertible.rb
+++ b/lib/jekyll/convertible.rb
@@ -1,6 +1,6 @@
 # encoding: UTF-8
 
-require 'set'
+require "set"
 
 # Convertible provides methods for converting a pagelike item
 # from a certain type of markup into actual content
@@ -20,12 +20,15 @@ module Jekyll
   module Convertible
     # Returns the contents as a String.
     def to_s
-      content || ''
+      content || ""
     end
 
     # Whether the file is published or not, as indicated in YAML front-matter
+    #
+    # Returns 'false' if the 'published' key is specified in the
+    # YAML front-matter and is 'false'. Otherwise returns 'true'.
     def published?
-      !(data.key?('published') && data['published'] == false)
+      !(data.key?("published") && data["published"] == false)
     end
 
     # Read the YAML frontmatter.
@@ -34,56 +37,15 @@ module Jekyll
     # name - The String filename of the file.
     # opts - optional parameter to File.read, default at site configs
     #
-    # Returns nothing.
+    # Returns String data read from file.
     def read_yaml(base, name, opts = {})
       filename = File.join(base, name)
 
-      begin
-        self.content = File.read(@path || site.in_source_dir(base, name),
-                                 Utils.merged_file_read_opts(site, opts))
-        if content =~ Document::YAML_FRONT_MATTER_REGEXP
-          self.content = $POSTMATCH
-          self.data = SafeYAML.load(Regexp.last_match(1))
-        end
-      rescue SyntaxError => e
-        Jekyll.logger.warn "YAML Exception reading #{filename}: #{e.message}"
-      rescue Exception => e
-        Jekyll.logger.warn "Error reading file #{filename}: #{e.message}"
-      end
-
-      self.data ||= {}
-
+      read_data filename, opts
       validate_data! filename
       validate_permalink! filename
 
       self.data
-    end
-
-    def validate_data!(filename)
-      unless self.data.is_a?(Hash)
-        raise Errors::InvalidYAMLFrontMatterError, "Invalid YAML front matter in #{filename}"
-      end
-    end
-
-    def validate_permalink!(filename)
-      if self.data['permalink'] && self.data['permalink'].size == 0
-        raise Errors::InvalidPermalinkError, "Invalid permalink in #{filename}"
-      end
-    end
-
-    # Transform the contents based on the content type.
-    #
-    # Returns the transformed contents.
-    def transform
-      converters.reduce(content) do |output, converter|
-        begin
-          converter.convert output
-        rescue => e
-          Jekyll.logger.error "Conversion error:", "#{converter.class} encountered an error while converting '#{path}':"
-          Jekyll.logger.error("", e.to_s)
-          raise e
-        end
-      end
     end
 
     # Determine the extension depending on content_type.
@@ -94,55 +56,11 @@ module Jekyll
       Jekyll::Renderer.new(site, self).output_ext
     end
 
-    # Determine which converter to use based on this convertible's
-    # extension.
-    #
-    # Returns the Converter instance.
-    def converters
-      @converters ||= site.converters.select { |c| c.matches(ext) }.sort
-    end
-
-    # Render Liquid in the content
-    #
-    # content - the raw Liquid content to render
-    # payload - the payload for Liquid
-    # info    - the info for Liquid
-    #
-    # Returns the converted content
-    def render_liquid(content, payload, info, path)
-      site.liquid_renderer.file(path).parse(content).render!(payload, info)
-    rescue Tags::IncludeTagError => e
-      Jekyll.logger.error "Liquid Exception:", "#{e.message} in #{e.path}, included in #{path || self.path}"
-      raise e
-    rescue Exception => e
-      Jekyll.logger.error "Liquid Exception:", "#{e.message} in #{path || self.path}"
-      raise e
-    end
-
-    # Convert this Convertible's data to a Hash suitable for use by Liquid.
-    #
-    # Returns the Hash representation of this Convertible.
-    def to_liquid(attrs = nil)
-      further_data = Hash[(attrs || self.class::ATTRIBUTES_FOR_LIQUID).map do |attribute|
-        [attribute, send(attribute)]
-      end]
-
-      defaults = site.frontmatter_defaults.all(relative_path, type)
-      Utils.deep_merge_hashes defaults, Utils.deep_merge_hashes(data, further_data)
-    end
-
     # The type of a document,
     #   i.e., its classname downcase'd and to_sym'd.
     #
     # Returns the type of self.
     def type
-      if is_a?(Page)
-        :pages
-      end
-    end
-
-    # returns the owner symbol for hook triggering
-    def hook_owner
       if is_a?(Page)
         :pages
       end
@@ -168,7 +86,7 @@ module Jekyll
     #
     # Returns true if extname == .coffee, false otherwise.
     def coffeescript_file?
-      '.coffee'.eql?(ext)
+      ".coffee" == ext
     end
 
     # Determine whether the file should be rendered with Liquid.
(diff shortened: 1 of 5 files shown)


lib/jekyll/convertible.rb
```diff
@@ -34,58 +36,17 @@ def published?
     # name - The String filename of the file.
     # opts - optional parameter to File.read, default at site configs
     #
-    # Returns nothing.
+    # Returns data read from file.
```
envygeeks: `Returns [Type] description`

ayastreb:
I think it would be better to include Convertible in Document, and then just call render on document instance in Renderer.
I will try to do that and merge this PR with another one for Document

envygeeks:
hey @ayastreb can you clean up the commit history so it's easier to read? Thanks!

ayastreb:
Hi @envygeeks I've squashed the commits and rebased, but it looks like that same `test_site.rb` failed again 😄 
Could you please re-start the tests?

envygeeks:
Hulk here,

I've restarted the failing tests for you.
Thanks.

parkr:
> I included Convertible into Document and implemented render method the same way it is implemented in Page
> I removed all the logic from Renderer class and just call this render method on the document instance.

@ayastreb This is the opposite of our intent on creating the `Render` class. The idea was that we'd use a new instance of the `Render` class to render documents, and any document passed in would have to respond to a given interface.

Ultimately, we want to _remove_ the `Convertible` module in favor of the `Render` class. So much of your refactoring here is not acceptable in its current form. Would you please rework it?

ayastreb:
@parkr, I'we reworked this PR and moved back the rendering logic to `Renderer` class.
I removed all the rendering logic from `Convertible` class and now it seems to be doing other things: reading yaml files, writing files and checking file types. Some methods (`published?`, `write`, `sass_file?`, etc) are still duplicated in `Document` class. Maybe we should split or rename `Convertible` somehow?

In `Page` class I've changed `render` method to use `Renderer` instead of using methods from `Convertible`. I also changed `Page.render` signature - we don't need `layouts` variable there.
Do you think it's ok? Could it be that other people use this method somehow? I only found this method used in tests, so not sure 😕

ayastreb:
Hi @parkr, just pinging here 😄 
Is there anything I could do more in this PR?

parkr:
@ayastreb This PR now has a number of merge conflicts. Would you mind rebasing on the latest `origin/master`?

ayastreb:
Hi @parkr, I rebased my branch but looks like that test failed again.
It fails from time to time randomly :(
Could you restar the build, please?

lib/jekyll/convertible.rb
```diff
@@ -34,58 +37,17 @@ def published?
...
-        end
-      rescue SyntaxError => e
-        Jekyll.logger.warn "YAML Exception reading #{filename}: #{e.message}"
-      rescue Exception => e
-        Jekyll.logger.warn "Error reading file #{filename}: #{e.message}"
-      end
-
-      self.data ||= {}
-
+      read_data filename, opts
```
parkr: Let's call this method `read_and_parse_contents` rather than `read_data`.

lib/jekyll/document.rb
```diff
@@ -7,9 +7,9 @@ class Document
     attr_reader :path, :site, :extname, :collection
     attr_accessor :content, :output
 
-    YAML_FRONT_MATTER_REGEXP = /\A(---\s*\n.*?\n?)^((---|\.\.\.)\s*$\n?)/m
-    DATELESS_FILENAME_MATCHER = /^(.+\/)*(.*)(\.[^.]+)$/
-    DATE_FILENAME_MATCHER = /^(.+\/)*(\d+-\d+-\d+)-(.*)(\.[^.]+)$/
+    YAML_FRONT_MATTER_REGEXP = %r!\A(---\s*\n.*?\n?)^((---|\.\.\.)\s*$\n?)!m
+    DATELESS_FILENAME_MATCHER = %r!^(?:.+\/)*(.*)(\.[^.]+)$!
+    DATE_FILENAME_MATCHER = %r!^(?:.+\/)*(\d+-\d+-\d+)-(.*)(\.[^.]+)$!
```
parkr: Why did you add the `?:`? That doesn't seem necessary? You trying to pull a fast one on me? 😂

parkr:
@ayastreb For the record, there are definitely some substantive changes in here! Code clean-up is easy to review but larger changes, e.g. switching Pathname for Pathutil, need to be done very carefully and in an isolated manner so we don't break anything.

lib/jekyll/convertible.rb
```diff
@@ -278,10 +111,8 @@ def do_layout(payload, layouts)
     def write(dest)
       path = destination(dest)
       FileUtils.mkdir_p(File.dirname(path))
-      File.open(path, 'wb') do |f|
-        f.write(output)
-      end
-      Jekyll::Hooks.trigger hook_owner, :post_write, self
+      File.write(path, output)
```
parkr: The `wb` mode helped with content that was non-ASCII. We need to make sure we keep that mode.

lib/jekyll/convertible.rb
```diff
@@ -296,5 +127,42 @@ def [](property)
...
+        self.content = File.read(@path || site.in_source_dir(filename),
+                                 Utils.merged_file_read_opts(site, opts))
+        if content =~ Document::YAML_FRONT_MATTER_REGEXP
+          self.content = $POSTMATCH
+          self.data = SafeYAML.load(Regexp.last_match(1))
+        end
+      rescue SyntaxError => e
+        Jekyll.logger.warn "YAML Exception reading #{filename}: #{e.message}"
+      # rubocop:disable RescueException
+      rescue Exception => e
```
parkr: You can probably change this to `rescue => e`.

lib/jekyll/convertible.rb
```diff
@@ -296,5 +127,42 @@ def [](property)
...
+      self.data ||= {}
+    end
+
+    private
+    def validate_data!(filename)
+      unless self.data.is_a?(Hash)
+        raise(
+          Errors::InvalidYAMLFrontMatterError,
+          "Invalid YAML front matter in #{filename}"
+        )
```
parkr: I'd rather not use parentheses. How about:

```ruby
raise Errors::InvalidYAMLFrontMatterError,
  "Invalid YAML front matter in #{filename}"
```

lib/jekyll/document.rb
```diff
@@ -103,7 +84,7 @@ def output_ext
     #
     # Returns the basename without the file extname.
     def basename_without_ext
-      @basename_without_ext ||= File.basename(path, '.*')
+      @basename_without_ext ||= File.basename(path, File.extname(path))
```
parkr: This may change behaviour. Please revert.

lib/jekyll/document.rb
```diff
@@ -237,18 +218,16 @@ def destination(base_directory)
     def write(dest)
       path = destination(dest)
       FileUtils.mkdir_p(File.dirname(path))
-      File.open(path, 'wb') do |f|
-        f.write(output)
-      end
-
+      File.write(path, output)
```
parkr: Please ensure we are using the `wb` mode.

lib/jekyll/document.rb
```diff
@@ -263,75 +242,20 @@ def read(opts = {})
...
-          end
-
-          post_read
+          merge_defaults
+          read_content(opts)
+          read_post_data
         rescue SyntaxError => e
           Jekyll.logger.error "Error:", "YAML Exception reading #{path}: #{e.message}"
+        # rubocop: disable RescueException
         rescue Exception => e
```
parkr: You can remove `Exception` – just do `rescue => e` here.

lib/jekyll/document.rb
ayastreb: I added it to make the first group not captured, so that I can use all the captures later like this:
```ruby
if relative_path =~ DATE_FILENAME_MATCHER
  date, slug, ext = Regexp.last_match.captures
```
instead of
```ruby
if relative_path =~ DATE_FILENAME_MATCHER
   date, slug, ext = $2, $3, $4
```
So I changed the regexp to only capture date, slug and ext so that I don't need to use $2, $3, $4 (Rubocop doesn't like Pearl-style captures)

ayastreb:
Hi @parkr, switching Pathname for Pathutil was proposed to me by @envygeeks in one of the previous reviews (#4960), so I thought it is safe :)
Should I revert it and open a separate PR for that?

lib/jekyll/document.rb
```diff
@@ -433,15 +353,129 @@ def respond_to?(method, include_private = false)
     # Override of method_missing to check in @data for the key.
     def method_missing(method, *args, &blck)
       if data.key?(method.to_s)
-        Jekyll.logger.warn "Deprecation:", "Document##{method} is now a key in the #data hash."
+        Jekyll.logger.warn(
+          "Deprecation:",
+          "Document##{method} is now a key in the #data hash."
+        )
```
parkr: Let's do `Jekyll::Deprecator. deprecation_message "Document##{method} is now a key in the #data hash."`

lib/jekyll/document.rb
```diff
@@ -433,15 +353,129 @@ def respond_to?(method, include_private = false)
...
+    private
+    def merge_data!(other, source: "YAML front matter")
+      merge_categories!(other)
+      Utils.deep_merge_hashes!(data, other)
+      if data.key?("date") && !data["date"].is_a?(Time)
+        data["date"] = Utils.parse_date(
+          data["date"].to_s,
+          "Document '#{relative_path}' does not have a valid date in the #{source}."
+        )
+      end
```
parkr: Why don't we do a `merge_date!` here, like `merge_categories!`.

parkr:
> Should I revert it and open a separate PR for that?

It appears to be working so we can leave it. But keeping PR's small makes them more manageable for us maintainers 😄

lib/jekyll/document.rb
```diff
@@ -433,15 +353,129 @@ def respond_to?(method, include_private = false)
...
+
+    private
+    def populate_categories
+      merge_data!({
+        "categories" => (
+        Array(data["categories"]) + Utils.pluralized_array_from_hash(
+          data,
+          "category",
+          "categories"
+        )
```
parkr: This looks like the wrong indentation. Would you push `Array` two spaces right and align the following lines?

```ruby
def populate_categories
  merge_data!({
    "categories" =>
      Array(data["categories"]) +
      Utils. pluralized_array_from_hash(data, "category", "categories")
  })
end
```

Does that work?

lib/jekyll/renderer.rb
```diff
@@ -185,10 +219,21 @@ def converter_output_ext
...
+
+    private
+    def path
+      document.path
+    end
+
+    private
+    def relative_path
+      document.relative_path
+    end
```
parkr: These two methods are superfluous – there is no reason to make a method here, and `Renderer#relative_path` doesn't make sense because the Renderer doesn't have any concept of path.

test/test_page.rb
```diff
@@ -11,10 +11,10 @@ def setup_page(*args)
   end
 
   def do_render(page)
-    layouts = {
+    @site.layouts = {
       "default" => Layout.new(@site, source_dir("_layouts"), "simple.html")
     }
-    page.render(layouts, @site.site_payload)
+    page.render(@site.site_payload)
```
parkr: We cannot change this method signature. Please revert.

lib/jekyll/renderer.rb
```diff
@@ -13,170 +13,204 @@ def initialize(site, document, site_payload = nil)
...
+      document.output = converters.reduce(document.output) do |output, converter|
         begin
           converter.convert output
         rescue => e
-          Jekyll.logger.error "Conversion error:", "#{converter.class} encountered an error while converting '#{document.relative_path}':"
+          Jekyll.logger.error(
+            "Conversion error:",
+            "#{converter.class} encountered an error "\
+            "while converting '#{relative_path}':"
+          )
```
parkr: Let's do an alternative route here:

```ruby
Jekyll.logger.error "Conversion error:",
  "#{converter.class} encountered an error while converting '#{document.relative_path}':"
```

lib/jekyll/renderer.rb
```diff
@@ -13,170 +13,204 @@ def initialize(site, document, site_payload = nil)
...
+    private
+    def render_document
       if document.render_with_liquid?
-        Jekyll.logger.debug "Rendering Liquid:", document.relative_path
-        output = render_liquid(output, payload, info, document.path)
+        Jekyll.logger.debug "Rendering Liquid:", relative_path
+        document.output = render_liquid(document.content, path)
       end
 
-      Jekyll.logger.debug "Rendering Markup:", document.relative_path
```
parkr: What happened to this log message?

parkr:
@ayastreb I think at this time I would like you to split these out into multiple PR's. It's just too much all at once. Could you please take each file, and only do style-required refactors? Then refactors which change things more substantively could come later (e.g. Pathutil and changing functions around).

lib/jekyll/document.rb
envygeeks: Personally I'd format it as:

```ruby
def populate_categories
  merge_data!({
    "categories" =>
      Array(data["categories"]) + Utils. pluralized_array_from_hash(
        data, "category", "categories"
      )
  })
end

ayastreb:
@parkr thanks for the feedback!
I will try to do separate PRs today.

PR #4964, opened 2016-05-30
make lib/jekyll.rb rubocop compliant 

floydpraveen:
fix for one of the file listed in #4885

### DIFF
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -6,79 +6,79 @@ $LOAD_PATH.unshift File.dirname(__FILE__) # For use/testing when no gem is insta
 #
 # Returns nothing.
 def require_all(path)
-  glob = File.join(File.dirname(__FILE__), path, '*.rb')
+  glob = File.join(File.dirname(__FILE__), path, "*.rb")
   Dir[glob].sort.each do |f|
     require f
   end
 end
 
 # rubygems
-require 'rubygems'
+require "rubygems"
 
 # stdlib
 require "pathutil"
-require 'forwardable'
-require 'fileutils'
-require 'time'
-require 'English'
-require 'pathname'
-require 'logger'
-require 'set'
+require "forwardable"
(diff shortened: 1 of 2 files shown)


lib/jekyll.rb
```diff
@@ -115,8 +116,8 @@ def configuration(override = Hash.new)
     # timezone - the IANA Time Zone
     #
     # Returns nothing
-    def set_timezone(timezone)
-      ENV['TZ'] = timezone
+    def add_timezone(timezone)
```
envygeeks: This is unacceptable, it's an API change, please change this back to what it was. Thanks.

lib/jekyll.rb
```diff
@@ -154,29 +155,29 @@ def sites
     def sanitized_path(base_directory, questionable_path)
       return base_directory if base_directory.eql?(questionable_path)
 
-      questionable_path.insert(0, '/') if questionable_path.start_with?('~')
+      questionable_path.insert(0, "/") if questionable_path.start_with?("~")
       clean_path = File.expand_path(questionable_path, "/")
-      clean_path.sub!(/\A\w\:\//, '/')
+      clean_path.sub!(%r!/\A\w\:\//!, "/")
```
envygeeks: The `//` at the beginning and end didn't get removed, this will break the current API.

lib/jekyll.rb
```diff
@@ -154,29 +155,29 @@ def sites
...
 
-      if clean_path.start_with?(base_directory.sub(/\A\w\:\//, '/'))
+      if clean_path.start_with?(base_directory.sub(%r!/\A\w\:\//!, "/"))
```
envygeeks: The `//` at the beginning and end didn't get removed, this will break the current API.

envygeeks:
Thanks! Just a few comments.

floydpraveen:
Will send different PR after correcting these

PR #4966, opened 2016-05-30
Rubocop jekyll.rb

floydpraveen:
fixes one file from #4964 
#### work in progress
yet to fix below offence reported by rubocop
```
lib/jekyll.rb:119:9: C: Do not prefix writer method names with set_.
    def set_timezone(timezone)
        ^^^^^^^^^^^^
```

### DIFF
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -6,79 +6,79 @@ $LOAD_PATH.unshift File.dirname(__FILE__) # For use/testing when no gem is insta
 #
 # Returns nothing.
 def require_all(path)
-  glob = File.join(File.dirname(__FILE__), path, '*.rb')
+  glob = File.join(File.dirname(__FILE__), path, "*.rb")
   Dir[glob].sort.each do |f|
     require f
   end
 end
 
 # rubygems
-require 'rubygems'
+require "rubygems"
 
 # stdlib
 require "pathutil"
-require 'forwardable'
-require 'fileutils'
-require 'time'
-require 'English'
-require 'pathname'
-require 'logger'
-require 'set'
+require "forwardable"
+require "fileutils"
+require "time"
+require "English"
+require "pathname"
+require "logger"
+require "set"
 
 # 3rd party
-require 'safe_yaml/load'
-require 'liquid'
-require 'kramdown'
-require 'colorator'
+require "safe_yaml/load"
+require "liquid"
+require "kramdown"
+require "colorator"
 
 SafeYAML::OPTIONS[:suppress_warnings] = true
 
 module Jekyll
   # internal requires
-  autoload :Cleaner,             'jekyll/cleaner'
-  autoload :Collection,          'jekyll/collection'
-  autoload :Configuration,       'jekyll/configuration'
-  autoload :Convertible,         'jekyll/convertible'
-  autoload :Deprecator,          'jekyll/deprecator'
-  autoload :Document,            'jekyll/document'
-  autoload :Draft,               'jekyll/draft'
-  autoload :EntryFilter,         'jekyll/entry_filter'
-  autoload :Errors,              'jekyll/errors'
-  autoload :Excerpt,             'jekyll/excerpt'
-  autoload :External,            'jekyll/external'
-  autoload :FrontmatterDefaults, 'jekyll/frontmatter_defaults'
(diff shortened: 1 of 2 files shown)


envygeeks:
hey @floydpraveen do you mind adding `# rubocop:disable Style/AccessorMethodName` above `set_timezone` and then below it and above the next method name do `# rubocop:enable Style/AccessorMethodName`

This is :shipit: after that minor change!
Thanks!

floydpraveen:
@envygeeks done, thanks for reviewing it

floydpraveen:
@envygeeks let me reword the commit messages

lib/jekyll.rb
```diff
@@ -115,9 +116,11 @@ def configuration(override = Hash.new)
     # timezone - the IANA Time Zone
     #
     # Returns nothing
+    # rubocop:disable Style/AccessorMethodName
     def set_timezone(timezone)
-      ENV['TZ'] = timezone
+      ENV["TZ"] = timezone
     end
+    # rubocop:enable Style/AccessorMethodName
```
parkr: I don't think this is necessary 😄  It's just applies to the method when it's right above the method like that, I'm pretty sure.

lib/jekyll.rb
```diff
@@ -154,29 +157,29 @@ def sites
     def sanitized_path(base_directory, questionable_path)
       return base_directory if base_directory.eql?(questionable_path)
 
-      questionable_path.insert(0, '/') if questionable_path.start_with?('~')
+      questionable_path.insert(0, "/") if questionable_path.start_with?("~")
       clean_path = File.expand_path(questionable_path, "/")
-      clean_path.sub!(/\A\w\:\//, '/')
+      clean_path.sub!(%r!\A\w\:\/!, "/")
```
parkr: Correct me if I'm wrong, @envygeeks, but I think this is `%r!\A\w\:/!`.

lib/jekyll.rb
```diff
@@ -154,29 +157,29 @@ def sites
...
 
-      if clean_path.start_with?(base_directory.sub(/\A\w\:\//, '/'))
+      if clean_path.start_with?(base_directory.sub(%r!\A\w\:\/!, "/"))
```
parkr: Same regexp here as above, I guess? Maybe we can pull that out into a constant.

parkr:
Looking pretty good! Just needs a rebase and I left a couple comments above. ❤️

lib/jekyll.rb
envygeeks: @parkr it continues for the entire file, though you are right that it isn't necessary.

lib/jekyll.rb
envygeeks: `%r!\A\w:/!`

lib/jekyll.rb
floydpraveen: @parkr @envygeeks since it continues for the entire file, let it be there ? so new method added would be checked by rubocop

floydpraveen:
@parkr @envygeeks made the changes, thanks for comments

lib/jekyll.rb
```diff
@@ -154,29 +157,31 @@ def sites
...
       return base_directory if base_directory.eql?(questionable_path)
 
-      questionable_path.insert(0, '/') if questionable_path.start_with?('~')
+      questionable_path.insert(0, "/") if questionable_path.start_with?("~")
       clean_path = File.expand_path(questionable_path, "/")
-      clean_path.sub!(/\A\w\:\//, '/')
+      clean_path.sub!(%r!\A\w:/!, "/")
 
-      if clean_path.start_with?(base_directory.sub(/\A\w\:\//, '/'))
+
```
envygeeks: This is causing your formatting specs to fail: you have two new lines instead of one.

lib/jekyll.rb
envygeeks: @floydpraveen yessum

lib/jekyll.rb
floydpraveen: removed it

lib/jekyll.rb
pathawks: @floydpraveen :+1:

envygeeks:
@jekyllbot: merge +dev

PR #4967, opened 2016-05-30
Suggested a more elegant way to handle the lang attribute

HugoGiraudel:
A few months ago I added the `lang` attribute to the `<html>` element of the default site because it causes an accessibility issue not to have it: https://github.com/jekyll/jekyll/pull/4633. I think setting it to `en` is a bit of a quick assumption though.

I suggest we set it to `page.lang` if it is set, or `en` as a default value. This change will play well with the regular pattern of setting a `lang` option in the YAML front matter of a page to specify its language, and the `en` default value seems reasonable enough since it is, well, a default. :)

[Someone suggested we use `site.lang`](https://github.com/jekyll/jekyll/pull/4633#issuecomment-194942403) and add it in `_config.yml`. I first thought it would be a good idea, but I now disagree on this point since it would not make sense anymore in a multi-lingual Jekyll website. I think relying on `page` is simply safer.

### DIFF
--- a/lib/site_template/_layouts/default.html
+++ b/lib/site_template/_layouts/default.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="{{ page.lang | default: 'en' }}">
 
   {% include head.html %}
 


ayastreb:
+1 to rely on `page` - on a multi-lingual site there could be pages with different languages on the same site. Depends of course how this "multi-lingual" is implemented on concrete website 😄 
Maybe it could fallback to `site` if there is no `page.lang`?

envygeeks:
That someone is @parkr and he's the project lead, and I happen to agree with him on `site.lang`.

HugoGiraudel:
@envygeeks I think `site.lang` itself is not enough. If I have a project running both English and Spanish, there is no reason to have a static language set in the configuration. Each page will have its own `lang` attribute based on its content.

At this point, I guess the best solution would be to rely on `page.lang`, then `site.lang` then `en` as a default, in this specific order.

Thoughts?

pathawks:
```
{% assign lang = page.lang | default: site.lang %}
<html{% if lang %} lang="{{ lang }}"{% endif %}>
```

pathawks:
Why use `en` as a default?
Garbage data is _worse_ than absnse of data. If no language is specified, the browser can try to guess based on context.

envygeeks:
Your question is a trap. There is no right answer because no matter what it can be construed negatively.

envygeeks:
And the browser will *not* attempt to infer a language from context, it will use *your* default language.

envygeeks:
@HugoGiraudel I may agree with @parkr on the variable name, but if I'm truly honest, I'll agree with anything that happens as far as fixing the problem, we can both agree that it being non-configurable easily is a problem and as far as I'm concerned, that's all that matters to me, whether it's on site or page matters little to me at the end of the day so whatever prevails as the common vote I'm good with.

pathawks:
I do not see any information from [W3C](http://www.w3.org/TR/html5/dom.html#attr-lang) that suggests the default language is that of the user agent.

Anecdotally, I have visited sites in Google Chrome that did not seem to have a `lang` set that Chrome offered to translate into English for me.

I think that setting `site.lang` to `en` in the default `_config.yml` is a fine idea. I don’t see why that wouldn’t be Good Enough.

envygeeks:
@pathawks 

> I do not see any information from W3C that suggests the default language is that of the user agent.

That's because it's the user's agent doing it for them.  Not the W3C dictating it.  The W3C gave us a standard on how to configure it and your browser is there to make it easy.  Your browser sends Accept-Lang header and most of the time if no language is configured by the HTML, it will make that the language is uses when working with the content.

> Anecdotally, I have visited sites in Google Chrome that did not seem to have a lang set that Chrome offered to translate into English for me.

That does happen, if you have it enabled, Chrome will try to infer the language at that point when it can detect a language (and I forgot all about that as I disable it,) actually it'll even do it when the language is set and it's wrong (last I remember) and even if the language is set and right.  Heck, one time it even tried to do it for English on my mums browser when she was browsing Twitter, in the early days of web Emoji.

DirtyF:
@HugoGiraudel Event if it's not a jekyll core feature, there's a nice gem by @sylvaindurand for that: https://rubygems.org/gems/jekyll-multilingual

HugoGiraudel:
@DirtyF Thanks for the tip. :)

caspervonb:
I'd fallback to site.lang if page.lang is not available.

```jekyll
<html lang="{{ page.lang | default: site.lang | default: "en" }}">
```

Clearer to default to en than relying on an implementation detail IMHO? Just from a documentation standpoint.

parkr:
@HugoGiraudel @caspervonb I'm fine with:

```text
<html lang="{{ page.lang | default: site.lang | default: "en" }}">
```

We're about to move this to https://github.com/jekyll/minima – would you please submit a PR to change that code, too?

PR #4975, opened 2016-06-02
Instructions on how to install github-gem on Windows

jwillmer:
I added a line to a blog post I wrote because it took me 3 days to figure it out.

### DIFF
--- a/site/_docs/github-pages.md
+++ b/site/_docs/github-pages.md
@@ -76,6 +76,8 @@ gem 'github-pages'
 {% endhighlight %}
 
     And be sure to run <code>bundle update</code> often.
+    
+    If you like to install <code>pages-gem</code> on Windows you can find instructions by Jens Willmer on <a href="http://jwillmer.de/blog/tutorial/how-to-install-jekyll-and-pages-gem-on-windows-10-x46">how to install github-pages gem on Windows (x64)</a>.
   </p>
 </div>
 
--- a/site/_docs/github-pages.md
+++ b/site/_docs/github-pages.md
@@ -77,7 +77,7 @@ gem 'github-pages'
 
     And be sure to run <code>bundle update</code> often.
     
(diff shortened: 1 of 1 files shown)


envygeeks:
I don't know what the @jekyll/ecosystem and @jekyll/core wish for this kind of thing but I would much rather prefer the directions be embedded directly in the page with a link back to your site as the credit, this reduces the friction to solving a problem IMO.

jwillmer:
Sure, I can add it direct to the docs I only did it this way because I saw that in the [windows installation instructions](https://jekyllrb.com/docs/windows/#installation) the instructions are also only linked.

spudowiar:
@jwillmer maybe we should create an issue for that to be integrated and cited rather than linked

parkr:
Let's wait on that – the instructions there have a hard dependency on Chocolatey which I haven't heard of. Linking out is fine for now. 😄 

@jekyllbot: merge +site

spudowiar:
@parkr I agree, there should *not* be a hard requirement on Chocolatey or anything non essential to the process

envygeeks:
@spudowiar by that logic then we should throw out all directions mentioning boxen, and the other mac packages since they are non-essential.

@parkr @spudowiar as I'm sure @jwillmer already knows, Microsoft has pretty much endorsed Chocolately and even has mentioned it in their own docs, also, as far as I'm aware Microsofts own CLI package manager even has compatibility with it.

Not saying we shouldn't include it it, just saying that Chocolatey is to Windows what Boxen/Homebrew is to OS X except in this case it's actually somewhat integrated into the system from the operating itself.  As an example, this can be done on an out-of-box Windows install:

```
get-packageprovider -name chocolatey
install-package vlc
```

Maybe the post should be updated to switch to native Windows cmdlets on Windows 10 (with directions on how to install OneGet on Windows 7/8 -- it's open source and available from Microsoft for both?)

spudowiar:
@envygeeks I was not aware of this, I don't dabble on OS X or Windows :)

I still believe, not to bloat the documention however, we should also include instructions using the minimal amount of unnecessary software.

envygeeks:
The source for all this is from @shanselman at http://www.hanselman.com/blog/AptGetForWindowsOneGetAndChocolateyOnWindows10.aspx

PR #4976, opened 2016-06-02
Fix inaccurate HTTP response header field name

galopin:
Custom WEBrick headers default settings documentation is incorrect.

“We only provide one default and that’s a _Content-Type_ header that disables caching in development […]”

should read:

“We only provide one default and that’s a _Cache-Control_ header that disables caching in development […]”

> Cf. https://github.com/jekyll/jekyll/blob/master/lib/jekyll/commands/serve/servlet.rb#L8-9

### DIFF
--- a/site/_docs/configuration.md
+++ b/site/_docs/configuration.md
@@ -383,7 +383,7 @@ before your site is served.
   </p>
 </div>
 
-## Custom WEBRick Headers
+## Custom WEBrick Headers
 
 You can provide custom headers for your site by adding them to `_config.yml`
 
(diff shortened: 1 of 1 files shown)


envygeeks:
@galopin actually that is correct to an extent, it's correct in that we do provide Content-Type, but it's not right that we *only* provide it, we actually provide by default Content-Type ***and*** Cache-Control, one dynamic, the other static: https://github.com/jekyll/jekyll/blob/master/lib/jekyll/commands/serve/servlet.rb#L40

Do you mind updating the pull request to reflect that?
Thanks!

galopin:
@envygeeks thank you for the clarification :bulb: I updated this pull request accordingly.

envygeeks:
:heart:

@jekyllbot: merge +site

PR #4986, opened 2016-06-04
Add a basic template for pull requests

chrisfinazzo:
Credit goes to @jh3y for his work on [kody][], which I lifted this from. Anything else specific that people would like to see?

cc/ @jekyll/core 

[kody]: https://github.com/jh3y/kody

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,8 @@
+Fixes # || Adds new feature X
+===
+
+Changes include:
+
+* a
+* b
+* c


envygeeks:
@chrisfinazzo I'm not against this but I'm intrigued to know how this will affect pull request messages from the Git body. Do you know how this will affect this? I always fill out the pull request with information from the body of the Git message body (the more area.).  If this keeps that then I see no reason not to have it.

/cc @parkr @jekyll/core

parkr:
Cool! I'm going to turn down this proposal because I don't really feel the template you added gives us much value. It is very sparse and basically mimics the GitHub UI directly. Thanks for the suggestion, but we should come up with something that helps us evaluate pull requests more effectively.

chrisfinazzo:
OK, nbd.

Similar to the existing template for issues, it might be better to determine a few of the common categories for PRs and build a template based on that.

Off the top of my head, this might include: 

* Fixes (although this goes to your reasoning for not accepting this request)
* New features
* Website/Docs/Minima updates

Jordan's comment makes more sense to me in that light. Duplicating or messing with existing workflows would make it a non starter.

The use case for it isn't as strong as the one for issues, but anything that helps organize ongoing development is worth a look.

parkr:
For @envygeeks's comment, the web UI appends the commit comment to the bottom of the PR body template. So it's not lost.

I agree, coming up with a few categories would be nice. I think I'd like to come up with something easy that directs users to provide as much context as possible. "Here's why I wrote this feature/fix, and here's why it's applicable to all users." The hard part is making it a generally-applicable template. If you can come up with something easy but structured which encourages lots of context, I'd greatly appreciate it!

jh3y:
Just thought I'd pipe in as I got a mention :wink: Thanks for the kind words @chrisfinazzo 👍 

Personally, I see any guidance as better than none. Particularly when regarding the desired content of a PR/issue message. It is important that proposed changes to a project are of course communicated effectively. The same would be said of commit messages.

Good guidance for contribution can of course reduce the risk of comment threads being polluted with questions like "What does this do?", "Why is X implemented like this?", "Could you please tell us the browser used to recreate this?" etc.

Fully agree that it's tricky and not so trivial to find a template that provides a good balance between generic, informative whilst also being appropriate.

However, another good use of the PR template could be to ensure certain criteria is being met with PRs that are submitted. This can be done with the use of checkboxes in a checklist for contributors opening PRs. This communicates a clear list of criteria for a contributor to meet/assess in order to issue a PR whether it be coding standards, that things are done in a certain way or to ensure that all things required when implementing a change are considered/present. 

Examples could include;
```markdown
[x] Includes newly built `dist` files
[ ] Tests pass
[x] Updates `README`(where appropriate)
[ ] Linter throws no errors/warning
```

Hope you don't mind me chiming in :smile:

@jh3y

PR #4993, opened 2016-06-07
rubocop: jekyll/lib/filters.rb

ayastreb:
#4885

### DIFF
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -1,7 +1,7 @@
-require 'uri'
-require 'json'
-require 'date'
-require 'liquid'
+require "uri"
+require "json"
+require "date"
+require "liquid"
 
 module Jekyll
   module Filters
@@ -56,7 +56,7 @@ module Jekyll
     #
     # Returns the given filename or title as a lowercase URL String.
     # See Utils.slugify for more detail.
-    def slugify(input, mode=nil)
+    def slugify(input, mode = nil)
       Utils.slugify(input, :mode => mode)
     end
 
@@ -118,7 +118,7 @@ module Jekyll
     #
     # Returns the escaped String.
     def xml_escape(input)
-      input.to_s.encode(:xml => :attr).gsub(/\A"|"\Z/, "")
+      input.to_s.encode(:xml => :attr).gsub(%r!\A"|"\Z!, "")
     end
 
     # CGI escape a string for use in a URL. Replaces any special characters
@@ -133,7 +133,7 @@ module Jekyll
     #
     # Returns the escaped String.
     def cgi_escape(input)
-      CGI::escape(input)
+      CGI.escape(input)
     end
 
     # URI escape a string.
@@ -180,7 +180,7 @@ module Jekyll
       when 2
         "#{array[0]} #{connector} #{array[1]}"
       else
-        "#{array[0...-1].join(', ')}, #{connector} #{array[-1]}"
+        "#{array[0...-1].join(", ")}, #{connector} #{array[-1]}"
       end
     end
 
@@ -203,11 +203,14 @@ module Jekyll
     #   "items" => [...] } # all the items where `property` == "larry"
     def group_by(input, property)
       if groupable?(input)
-        input.group_by do |item|
-          item_property(item, property).to_s
-        end.inject([]) do |memo, i|
-          memo << { "name" => i.first, "items" => i.last, "size" => i.last.size }
-        end
+        input.group_by { |item| item_property(item, property).to_s }
+          .each_with_object([]) do |item, array|
+            array << {
+              "name"  => item.first,
+              "items" => item.last,
+              "size"  => item.last.size
+            }
+          end
       else
         input
       end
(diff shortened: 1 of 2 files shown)


lib/jekyll/filters.rb
```diff
@@ -203,9 +203,10 @@ def jsonify(input)
     #   "items" => [...] } # all the items where `property` == "larry"
     def group_by(input, property)
       if groupable?(input)
-        input.group_by do |item|
+        grouped = input.group_by do |item|
           item_property(item, property).to_s
-        end.inject([]) do |memo, i|
+        end
+        grouped.inject([]) do |memo, i|
```
envygeeks:
```ruby
input.group_by { |v| item_property(v, property).to_s }.
  each_with_object([]) do |item, ary|
    array << { 
      "name" => item.first, 
      "size" => item.last.size 
      "items" => item.last, 
    }
  end
```

lib/jekyll/filters.rb
```diff
@@ -223,7 +224,9 @@ def group_by(input, property)
     def where(input, property, value)
       return input unless input.is_a?(Enumerable)
       input = input.values if input.is_a?(Hash)
-      input.select { |object| Array(item_property(object, property)).map(&:to_s).include?(value.to_s) }
+      input.select do |object|
+        Array(item_property(object, property)).map(&:to_s).include?(value.to_s)
```
envygeeks: `[item_property(object, property)]`

lib/jekyll/filters.rb
```diff
@@ -333,6 +324,22 @@ def inspect(input)
     end
 
     private
+    def sort_input(input, property, order)
+      input.sort do |apple, orange|
```
envygeeks: This should simply do "a, b", I like the apple and orange though!

lib/jekyll/filters.rb
```diff
@@ -363,6 +372,7 @@ def item_property(item, property)
       end
     end
 
+    private
```
envygeeks: AFAIW as liquid should be public, /cc @jekyll/core is this assumption right?

envygeeks:
Thanks! Just a few comments, some for you, some for @jekyll/core.

lib/jekyll/filters.rb
ayastreb: Sorry, didn't get it 😕 
You suggest to replace `Array(item_property(object, property)).map(&:to_s).include?(value.to_s)` line with `[item_property(object, property)].map(&:to_s).include?(value.to_s)` to make it shorter?

envygeeks: Only Array() should be replaced with []

ayastreb: I tried that, so the method looks like this:
```ruby
def where(input, property, value)
  return input unless input.is_a?(Enumerable)
  input = input.values if input.is_a?(Hash)
  input.select do |object|
    [item_property(object, property)].map(&:to_s).include?(value.to_s)
  end
end
```
but then 2 tests fail:
```
Failure:
TestFilters#test_: filters where filter should filter array properties appropriately.  [test/test_filters.rb:472]
Minitest::Assertion: Expected: 2
  Actual: 0

Failure:
TestFilters#test_: filters where filter should filter array properties alongside string properties.  [test/test_filters.rb:481]
Minitest::Assertion: Expected: 2
  Actual: 1
```

envygeeks: Can you post a copy of the test?

ayastreb: Sure, it's this one:
```ruby
should "filter array properties appropriately" do
  hash = {
    "a" => { "tags"=>%w(x y) },
    "b" => { "tags"=>["x"] },
    "c" => { "tags"=>%w(y z) }
  }
  assert_equal 2, @filter.where(hash, "tags", "x").length
end
```
and also this one:
```ruby
should "filter array properties alongside string properties" do
  hash = {
    "a" => { "tags"=>%w(x y) },
    "b" => { "tags"=>"x" },
    "c" => { "tags"=>%w(y z) }
  }
  assert_equal 2, @filter.where(hash, "tags", "x").length
end
```

envygeeks: Sorry for the late reply @ayastreb, I didn't get an email for some reason and just randomly checked to see if this had feedback.  That said, I guess keep it as `Array`, I didn't think they were doing a cheap flatten, but it seems to me now that they are.  I don't quite agree since ([blah].flatten(1) is more explicit to a non-programmer but I digress.

Lets get some context on this from @parkr or anybody else on @jekyll/core who might have more info.

lib/jekyll/filters.rb
pathawks: These are all just helper methods that [were already private](https://github.com/jekyll/jekyll/blob/v3.1.6/lib/jekyll/filters.rb#L314) :+1:

parkr:
Hey @ayastreb! Thank you for doing this. ❤️ This looks pretty good to me so far. One question: why is `private` specified so many times? Is there a way we could specify it once? Or was it done to be more explicit? We don't do that anywhere else in the code so we should be consistent one way or the other. Thanks!

envygeeks:
@parkr it was done at my request because it makes the code more explicit and audit-able by quick glance.

parkr:
@envygeeks Ok, sounds great!

@jekyllbot: merge +dev

PR #5006, opened 2016-06-13
Windows docs: Command needs to be called from blog path

jbilinski:
unless targeting the blog local repository root, command will throw: 

> Could not locate Gemfile or .bundle/ directory


parkr:
Great! 

@jekyllbot: merge +site

PR #5009, opened 2016-06-14
Prevent reset of page in Liquid payload right before rendering layouts

### LINKED ISSUE #4414: Jekyll 3.1 broke page payload in layout
I have a plugin running on every page and adding a property to the payload:

```ruby
Jekyll::Hooks.register :pages, :pre_render do |page, payload|
## do stuff
    payload['page']['breadcrumbs'] = breadcrumbs
    #binding.pry

  end
```

In my layout file I have an include:
`{% include breadcrumb.html %}`

My `breadcrumb.html` uses the variable `{{page.breadcrumbs}}`.

With 3.0 this worked fine and I had access to this variable. Since upgrading to 3.1 this variable is now empty.

If I try to print the same variable in the page itself then it works. It does not work in the layout files or includes within the layout.

nasht00:
Note that this prevents the workaround from working anymore.

### DIFF
--- a/features/hooks.feature
+++ b/features/hooks.feature
@@ -333,3 +333,28 @@ Feature: Hooks
     Then I should get a zero exit status
     And the _site directory should exist
     And I should see "Wrote document 0" in "_site/document-build.log"
+
+  Scenario: Set a custom payload['page'] property
+    Given I have a _plugins directory
+    And I have a "_plugins/ext.rb" file with content:
+    """
+    Jekyll::Hooks.register :pages, :pre_render do |page, payload|
+        payload['page']['foo'] = "hello world"
+    end
+    """
+    And I have a _layouts directory
+    And I have a "_layouts/custom.html" file with content:
+      """
+      ---
+      ---
+      {{ content }} {% include foo.html %}
+      """
+    And I have a _includes directory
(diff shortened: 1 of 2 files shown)


nasht00:
Checks are apparently failing...

envygeeks:
Checks are fine, restarted the test for you.  I'm curious if this will have any problems down the line but it's LGTM.  :shipit:

parkr:
Perfect! @nasht00, would you please add a test for this? Create a hook like you did in #4414 and place it in a `.feature` file. You may create a new scenario in `features/hooks.feature` if you like. 😄 I don't want to break this again!

nasht00:
I'm not familiar with this .feature syntax, this reads like plain English. The test will execute this as-is??

nasht00:
Added a scenario. I would love some code review as this is a first for me.

parkr:
Looks good to me! Failing Travis check is a formatting error unrelated to your change. Thank you 😄 

@jekyllbot: merge +bug

parkr:
> I'm not familiar with this .feature syntax, this reads like plain English. The test will execute this as-is??

Yes it does! 😄 https://travis-ci.org/jekyll/jekyll/jobs/137730352#L466

PR #5014, opened 2016-06-15
Move bin/jekyll to exe/jekyll to prevent collision with binstubs

parkr:
If you run `bundle install --binstubs` in this repo, lots of executables are created for the dependencies of this project. The directory is not ignored because we have historically kept our executable (i.e. the `jekyll` CLI command) in this directory.

This PR moves it to `exe/jekyll` and ignores the `bin/` directory.


SuriyaaKudoIsc:
:+1:

envygeeks:
:shipit:

.gitignore
```diff
@@ -13,6 +13,7 @@
 /vendor
 Gemfile.lock
 _site/
+bin/
```
envygeeks: You should probably add !bin/jekyll because we need that to be tracked.

parkr: The PR body explains why that is not necessary.

parkr:
> Looks like some of our stuff is tightly coupled to bin?

Yep, in `features`. Fixing.

.gitignore
envygeeks: @parkr yeah I noticed that after I commented :blush:

parkr:
@jekyllbot: merge +minor

PR #5016, opened 2016-06-15
Rubocop : exe/jekyll

DirtyF:
#4885


DirtyF:
way too many commits in here, sorry I didn't mean too

PR #5019, opened 2016-06-16
rubocop: lib/jekyll/static_file.rb

ayastreb:
#4885

### DIFF
--- a/lib/jekyll/static_file.rb
+++ b/lib/jekyll/static_file.rb
@@ -2,6 +2,17 @@ module Jekyll
   class StaticFile
     attr_reader :relative_path, :extname
 
+    class << self
+      # The cache of last modification times [path] -> mtime.
+      def mtimes
+        @mtimes ||= {}
+      end
+
+      def reset_cache
+        @mtimes = nil
+      end
+    end
+
     # Initialize a new StaticFile.
     #
     # site - The Site.
@@ -17,8 +28,6 @@ module Jekyll
       @collection = collection
       @relative_path = File.join(*[@dir, @name].compact)
       @extname = File.extname(@name)
-      # The cache of last modification times [path] -> mtime.
-      @mtimes = {}
     end
     # rubocop: enable ParameterLists
 
@@ -57,7 +66,7 @@ module Jekyll
     #
     # Returns true if modified since last write.
     def modified?
-      @mtimes[path] != mtime
+      self.class.mtimes[path] != mtime
     end
 
     # Whether to write the file to the filesystem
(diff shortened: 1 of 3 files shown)


lib/jekyll/static_file.rb
```diff
@@ -11,6 +8,7 @@ class StaticFile
     # base - The String path to the <source>.
     # dir  - The String path between <source> and the file.
     # name - The String filename of the file.
+    # rubocop: disable ParameterLists
```
ayastreb: I'm not sure about this disabling...
It seems like second argument - base - is always taken from site.source variable:
https://github.com/jekyll/jekyll/blob/master/lib/jekyll/readers/static_file_reader.rb#L18
https://github.com/jekyll/jekyll/blob/master/lib/jekyll/collection.rb#L70

Do you think we should change the signature and usage of this class? Or just ignore it?

lib/jekyll/static_file.rb
```diff
@@ -19,7 +17,10 @@ def initialize(site, base, dir, name, collection = nil)
       @collection = collection
       @relative_path = File.join(*[@dir, @name].compact)
       @extname = File.extname(@name)
+      # The cache of last modification times [path] -> mtime.
+      @mtimes = {}
```
parkr: We definitely want this to be a class variable not an instance variable so that all instances share the same Hash. There are alternative ways of implementing a class variable than what we have done. What error prompted this change?

lib/jekyll/static_file.rb
parkr: Due to SemVer, we are not able to change it until Jekyll 4.

lib/jekyll/static_file.rb
ayastreb: It was this error:
```
lib/jekyll/static_file.rb:4:5: C: Style/ClassVars: Replace class var @@mtimes with a class instance var.
    @@mtimes = {}
    ^^^^^^^^
```

What is the alternative way? Sorry, I'm a newbie in Ruby 😄

parkr: One way to do it is:

```ruby
class StaticFile
  class << self
    def mtimes
      @mtimes ||= {}
    end

    def reset_cache
      @mtimes = nil
    end
  end

  # Now access that with self.class.mtimes, like self.class.mtimes[path] = mtime
  # You can access it outside the class (like in the tests) like before with StaticFile.mtimes or StaticFile.reset_cache
end

File: test/test_site.rb [hunk 3]
```diff
@@ -130,7 +129,6 @@ def generate(site)
 
     should "write static files if not modified but missing in destination" do
       clear_dest
-      StaticFile.reset_cache
```
parkr: Would you mind putting these back? I believe they were important.

parkr:
This is looking great, thank you! Just one more comment :smile:

parkr:
Thank you!!

@jekyllbot: merge +dev

PR #5025, opened 2016-06-19
rubocop: lib/jekyll/regenerator.rb

ayastreb:
#4885

### DIFF
--- a/lib/jekyll/regenerator.rb
+++ b/lib/jekyll/regenerator.rb
@@ -90,7 +90,14 @@ module Jekyll
         return cache[path]
       end
 
-      check_path_exists(path)
+      if metadata[path]
+        # If we have seen this file before,
+        # check if it or one of its dependencies has been modified
+        existing_file_modified?(path)
+      else
+        # If we have not seen this file before, add it to the metadata and regenerate it
+        add(path)
+      end
     end
 
     # Add a dependency of a path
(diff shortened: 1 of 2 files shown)


lib/jekyll/regenerator.rb
```diff
@@ -20,18 +20,14 @@ def initialize(site)
...
-            document.path, document.destination(@site.dest)
-          )
+        regenerate_document?(document)
       else
-        source_path = document.respond_to?(:path)        ? document.path                    : nil
-        dest_path   = document.respond_to?(:destination) ? document.destination(@site.dest) : nil
+        source_path = document.respond_to?(:path) ? document.path : nil
+        dest_path = if document.respond_to?(:destination)
+                      document.destination(@site.dest)
+                    end
```
parkr: What was wrong with the previous two lines? Just too long?

It seems like your solution for `source_path` and for `dest_path` are two ways of doing the same thing.

lib/jekyll/regenerator.rb
```diff
@@ -173,5 +153,42 @@ def read_metadata
...
+            return cache[dependency] = cache[path] = true
+          end
+        end
+        if File.exist?(path) && data["mtime"].eql?(File.mtime(path))
+          return cache[path] = false
+        end
+      end
+
+      # Path does not exist in metadata, add it
+      add(path)
```
parkr: `check_path_exists` should not modify anything.

lib/jekyll/regenerator.rb
```diff
@@ -94,23 +90,7 @@ def modified?(path)
...
-        if File.exist?(path) && data["mtime"].eql?(File.mtime(path))
-          return cache[path] = false
-        else
-          return add(path)
-        end
-      end
-
-      # Path does not exist in metadata, add it
-      return add(path)
+      check_path_exists(path)
```
parkr: The comment on the old line 97 only referenced the old line 98. 99 thru 110 determined if the file was modified, so it has nothing to do with the comment. This is three operations:

1. If we have seen this file before and one of its dependencies have been modified, set the regeneration bit for both the dependency and the file to `true`
2. If we have seen this file before and it has not been modified, set the regeneration bit to `false`. If it has been modified, set it to `true`.
3. If we have not seen this file before, add it to the metadata and regenerate it.

lib/jekyll/regenerator.rb
ayastreb: Yes, `dest_path` line was 98 chars long.
Is there another way to make it shorter?)

ayastreb:
Hi @parkr,
I've reworked `modified?` method a bit, hope now it's more clear 😄

parkr:
This looks GREAT! Well done, thank you ❤️ 

@jekyllbot: merge +dev

PR #5026, opened 2016-06-20
rubocop: lib/jekyll/utils.rb

ayastreb:
#4885

### DIFF
--- a/lib/jekyll/utils.rb
+++ b/lib/jekyll/utils.rb
@@ -2,27 +2,25 @@
 module Jekyll
   module Utils
     extend self
-    autoload :Platforms, 'jekyll/utils/platforms'
+    autoload :Platforms, "jekyll/utils/platforms"
     autoload :Ansi, "jekyll/utils/ansi"
 
     # Constants for use in #slugify
-    SLUGIFY_MODES = %w(raw default pretty)
+    SLUGIFY_MODES = %w(raw default pretty).freeze
     SLUGIFY_RAW_REGEXP = Regexp.new('\\s+').freeze
-    SLUGIFY_DEFAULT_REGEXP = Regexp.new('[^[:alnum:]]+').freeze
+    SLUGIFY_DEFAULT_REGEXP = Regexp.new("[^[:alnum:]]+").freeze
     SLUGIFY_PRETTY_REGEXP = Regexp.new("[^[:alnum:]._~!$&'()+,;=@]+").freeze
 
     # Takes an indented string and removes the preceding spaces on each line
 
     def strip_heredoc(str)
-      str.gsub(/^[ \t]{#{(str.scan(/^[ \t]*(?=\S)/).min || "").size}}/, "")
+      str.gsub(%r!^[ \t]{#{(str.scan(%r!^[ \t]*(?=\S)!).min || "").size}}!, "")
     end
 
     # Takes a slug and turns it into a simple title.
 
     def titleize_slug(slug)
-      slug.split("-").map! do |val|
-        val.capitalize
-      end.join(" ")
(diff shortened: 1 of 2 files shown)


lib/jekyll/utils.rb
```diff
@@ -146,11 +134,13 @@ def parse_date(input, msg = "Input could not be parsed.")
     # Determines whether a given file has
     #
     # Returns true if the YAML front matter is present.
+    # rubocop: disable PredicateName
```
ayastreb: Rubocop suggest to rename `has_yaml_header?` to `yaml_header?` but can we do that?

parkr: We cannot rename public methods, no. Unless it was marked `private` before, it cannot be renamed.

lib/jekyll/utils.rb
```diff
@@ -297,5 +287,36 @@ def merged_file_read_opts(site, opts)
...
+
+    private
+    def merge_default_proc(target, overwrite)
+      if default_proc?(target) && default_proc?(overwrite) && target.default_proc.nil?
+        target.default_proc = overwrite.default_proc
+      end
+    end
+
+    private
+    def default_proc?(object)
```
parkr: This method seems like it might be checking whether the object _is_ a default proc. Which is a bit confusing. Perhaps we could just check if these are `Hash`es?

lib/jekyll/utils.rb
```diff
@@ -297,5 +287,36 @@ def merged_file_read_opts(site, opts)
...
 
+    private
+    def merge_values(target, overwrite)
+      target.merge!(overwrite) do |_key, old_val, new_val|
+        if new_val.nil?
+          old_val
+        elsif mergable?(old_val) && mergable?(new_val)
+          deep_merge_hashes(old_val, new_val)
+        else
+          new_val
```
parkr: We could probably get away with doing a `new_val.frozen? ? new_val.dup : new_val` here, or using a method like `unfreeze(new_val)` to do the same for both this and `old_val` on line 294.

parkr:
Thanks, @ayastreb! Just two thoughts above.

lib/jekyll/utils.rb
ayastreb: The original code was:
```ruby
if target.respond_to?(:default_proc) && overwrite.respond_to?(:default_proc) && target.default_proc.nil?
  target.default_proc = overwrite.default_proc
end
```
I couldn't find the way to make `if` statement line shorter, so I created this `default_proc?` method.
Do you suggest to change the `if` condition to check if both `target` and `overwrite` are `Hash`?
Like this:
```ruby
if target.is_a?(Hash) && overwrite.is_a?(Hash) && target.default_proc.nil?
  target.default_proc = overwrite.default_proc
end
```

lib/jekyll/utils.rb
ayastreb: I just tried and a lot of tests fail with:
`TypeError: can't dup FalseClass`
looks like we need to check if the value is `duplicable?`.
But it was not duplicating values in the original code here, do we need to do it?

lib/jekyll/utils.rb
parkr: > Do you suggest to change the if condition to check if both target and overwrite are Hash?

That is one solution, yes. I don't think `Drop`s support `default_proc`, so there is no need to try to merge a default proc for anything but a Hash.

parkr:
@jekyllbot: merge +dev

parkr:
Thanks! I will take care of my nit-picks on `master`. 😄

ayastreb:
Great! Thanks!

PR #5036, opened 2016-06-26
It's called macOS now

chrisfinazzo:


README.markdown
```diff
@@ -20,9 +20,9 @@ Jekyll is a simple, blog-aware, static site generator perfect for personal, proj
 
 Jekyll does what you tell it to do — no more, no less. It doesn't try to outsmart users by making bold assumptions, nor does it burden them with needless complexity and configuration. Put simply, Jekyll gets out of your way and allows you to concentrate on what truly matters: your content.
 
-## Having trouble with OS X El Capitan?
+## Having trouble with macOS 10.11 El Capitan?
```
pathawks: Has _El Capitan_ been retroactively renamed, or is it just future versions that will be known as macOS going forward?

chrisfinazzo: To be honest, I think this is still in flux. Dev documentation on APFS still references "OS X". These things get updated somewhat regularly, especially for something that's new so I wouldn't be surprised if they renamed things at some point.

While I'm here, what's the reason for the `fmt` failure? From what I can see, it just runs the rubocop rules with a certain level of verbosity. For a change that's just a find & replace in a few places, how did this get tripped up?

pathawks: El Cap is still referred to as **OS X** on [Apple’s website](http://www.apple.com/osx/), I hesitate to deviate from that.

sondr3:
It is not in flux, all the old versions of OS X will still be OS X, it's only from macOS Sierra and onwards that will be called macOS.

PR #5038, opened 2016-06-27
Change the name to macOS and clean up one instruction

chrisfinazzo:
Second try at #5036.


parkr:
It seems like all instances of `macOS` should come with a caveat of `OS X El Capitan or later` because `macOS` hasn't been released with any version yet.

chrisfinazzo:
Hmmm. I'll keep a local copy for when this gets close to release.

Is there something going on with `fmt`? Both times I've sent builds to Travis, they fail, but errors are in files I haven't touched. `script/fmt` runs without any errors on my machine.

parkr:
@chrisfinazzo Ok sounds good. Fmt issues are likely due to upgrade in Rubocop version. Check out which Travis is using and which you're using.

PR #5042, opened 2016-06-29
Fix more theme things

parkr:
Based on @benbalter's feedback at https://github.com/jekyll/jekyll/pull/4848#issuecomment-220862806

> > `spec.files = %x{git ls-files -z}.split("\x0").reject { |f| f.match(%r{^(exe|_layouts|_includes|_sass|example|example/_posts)/}) }`
>
> Unless I'm misreading, we want to include layouts/includes/sass, the gemspec and exclude everything else (including exe). (A) I don't think this line accomplishes what we want right now, and (B) I'm wondering if we should default to a whitelist to encourage stronger defaults.

@benbalter Yeah, we just want `_layouts`, `_includes`, and `_sass` to be shipped with the gem. The purpose of the line is to package up the theme for release.

> > Rakefile... `example` dir
>
> This feels super heavyweight to me, and given that we have no way to e.g., distribute bug fixes to themes, I think we're setting ourselves up for heartbreak and failure "To fix this issue, copy and paste this Gist into your theme's Rakefile" `/headdesk`.

Yeah, that's totally a concern here. I just want an easy way to say "run this, open your browser, and it'll all work". Maybe it's a matter of saying: "run `jekyll new-theme your-theme`, run `jekyll new testing-your-theme`, then modify `testing-your-theme/Gemfile` and `_config.yml` to set theme to `your-theme` and..." man this is already feeling too complicated. I liked the idea of saying "run `jekyll new-theme my-theme`, `cd my-theme`, `bundle install && bundle exec rake preview` and go!" Maybe We need to create another Jekyll command...

> we may not want to add a code of conduct by default

Great idea. I'll move this to a flag like `--coc` or `--code-of-conduct`.

### DIFF
--- a/features/theme.feature
+++ b/features/theme.feature
@@ -8,6 +8,12 @@ Feature: Writing themes
     Then I should get a zero exit status
     And the my-cool-theme directory should exist
 
+  Scenario: Generating a new theme scaffold with a code of conduct
+    When I run jekyll new-theme my-cool-theme --code-of-conduct
+    Then I should get a zero exit status
+    And the my-cool-theme directory should exist
+    And the "my-cool-theme/CODE_OF_CONDUCT.md" file should exist
+
   Scenario: A theme with SCSS
     Given I have a configuration file with "theme" set to "test-theme"
     And I have a css directory
--- a/lib/jekyll/commands/new_theme.rb
+++ b/lib/jekyll/commands/new_theme.rb
@@ -6,20 +6,23 @@ class Jekyll::Commands::NewTheme < Jekyll::Command
       prog.command(:"new-theme") do |c|
         c.syntax "new-theme NAME"
         c.description "Creates a new Jekyll theme scaffold"
+        c.option "code_of_conduct", \
+          "-c", "--code-of-conduct", \
+          "Include a Code of Conduct. (defaults to false)"
 
-        c.action do |args, _|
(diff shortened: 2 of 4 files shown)


parkr:
@benbalter How do you think we should structure the docs around "testing your theme"?

benbalter:
> We need to create another Jekyll command...

I don't follow. Say a theme is a folder with `_layouts`, `_includes`, and `_sass` folders. Those folders are distributed with the Gem due to the gemfile. I can have any other files (e.g., `page.md` or `test/fixtures/page.md` with `permalink: /page.html`) that won't be distributed. Running `bundle exec jekyll build` from the root of the theme would produce a working Jekyll site, because the theme *is* a working Jekyll site (of which we distribute a subset).

parkr:
@benbalter That makes sense – my thought was to make the addition of content as seamless and uniform as possible. They'd run `jekyll serve-theme` which would essentially do `jekyll serve` with pre-filled content. That is more work for us, but simplifies the flow for them.

My only concern with saying "it's a jekyll site" is that people will be really confused about what _exactly_ is being distributed as a gem and maybe even folks will be confused about how to install the theme and will just fork it... Theme creators are confused about the scope of what makes a theme and theme users are confused about how to install. Those are my two worries 😄

parkr:
LGTM, too.

Issue #884, opened 2013-03-25
Use hard links


torrentkino:
Hello,

I have an enhancement request:

Please use "hard links" while creating _site whenever possible. That is fast and does not need additional disk space. Tons of images and HTML5 videos could be "copied" to _site in no time.

Kind regards,
Aiko Barz

ke352802081770314:
I have deployed jekyll recently, and very soon I was wondering why hardlinks are not used for assets, primarily due to disk space, but not having to lug many megabytes of data around would certainly speed up the build process too.
Incremental regeneration addresses build time, but the waste of space by systematically keeping all assets double on disk remains. Also incremental builds are not always a solution, if a site contains tag clouds, most recent posts and whatever in a sidebar of every page as is common in blogs, only a full generate will result in a consistent and appealing web presence.

Anyway, unfortunately I can't contribute any code as I don't speak ruby, so I can only make suggestions.
I verified by searching that ruby offers a "link" method for files which generates the hardlink: http://ruby-doc.org/core-1.9.3/File.html#method-c-link
It says it is not implemented on all platforms, which leaves it unclear whether it was implemented for Windows or not (yet). 
Both on Unix and Windows, hard links only work within the same file system, so depending on the setting of the _config.yml item "destination", they may or may not work. So, the approach has to be:

if (file is an asset i.e. is to be placed in destination unprocessed)
try File.link(sourcefile, destinationfile)
if that didn't work, copy sourcefile to destinationfile (as before)

Needless to say, I would appreciate if implementing this could be reconsidered. Many thanks.

PR #5049, opened 2016-07-02
Use hard links for static files.

stevecheckoway:
Speed up builds in nonproduction by using hard links for static files, if
possible. If the hard link fails, fall back to copying.

This is a significant speedup for sites with large static files.

### DIFF
--- a/lib/jekyll/static_file.rb
+++ b/lib/jekyll/static_file.rb
@@ -189,14 +189,16 @@ module Jekyll
     private
 
     def copy_file(dest_path)
-      if @site.safe || Jekyll.env == "production"
+      if @site.safe || Jekyll.env == "production" || !@site.config["hardlinks"]
         FileUtils.cp(path, dest_path)
       else
-        FileUtils.copy_entry(path, dest_path)
+        # Link to the real file, not a symlink.
+        FileUtils.ln(File.realpath(path), dest_path)
       end
 
       unless File.symlink?(dest_path)
-        File.utime(self.class.mtimes[path], self.class.mtimes[path], dest_path)
+        times = self.class.mtimes[path]
+        File.utime(times, times, dest_path)
       end
     end
   end
--- a/docs/_docs/configuration/default.md
+++ b/docs/_docs/configuration/default.md
@@ -59,6 +59,7 @@ show_dir_listing    : false
 permalink           : date
 paginate_path       : /page:num
 timezone            : null
+hardlinks           : false
 
 quiet               : false
 verbose             : false
--- a/docs/_docs/configuration/options.md
+++ b/docs/_docs/configuration/options.md
@@ -108,6 +108,21 @@ class="flag">flags</code> (specified on the command-line) that control them.
       </td>
     </tr>
     <tr class="setting">
+        <p class="name"><strong>Hardlink static files</strong></p>
+        <p class="description">
+            Enable the hardlink feature. Creates hard links to static files in
+            the site destination directory rather than copying them. This
+            results in a significant performance improvement for large files,
+            but will not work if the site source and site destination are
+            on different files system nor if the file system does not support
+            hard links (notably FAT).
+        </p>
+      </td>
+      <td class="align-center">
+        <p><code class="option">hardlinks: BOOL</code></p>
+      </td>
+    </tr>
+    <tr class="setting">
       <td>
         <p class="name"><strong>Encoding</strong></p>
         <p class="description">
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -53,6 +53,7 @@ module Jekyll
       "permalink"           => "date",
       "paginate_path"       => "/page:num",
       "timezone"            => nil, # use the local timezone
+      "hardlinks"           => false,
 
       "quiet"               => false,
       "verbose"             => false,


stevecheckoway:
I should have noted that this pull request is related to #884.

lib/jekyll/static_file.rb
```diff
@@ -144,7 +144,11 @@ def copy_file(dest_path)
       if @site.safe || Jekyll.env == "production"
         FileUtils.cp(path, dest_path)
       else
-        FileUtils.copy_entry(path, dest_path)
+        begin
+          FileUtils.ln(path, dest_path)
+        rescue
```
envygeeks: Begin and Rescue should not to be used as flow control.  This is a blocker as this is highly disoptimized.

stevecheckoway: I don't actually know Ruby so suggestions appreciated.

As near as I can tell, `FileUtils.ln` simply calls `File.link` which, I believe, is implemented [here](http://rxr.whitequark.org/mri/source/file.c#2436). One of three things happens: (1) `File.link` isn't implemented in which case it raises an exception; (2) it is implemented and `link(2)` fails in which case it raises an exception; or (3) it is implemented and `link(2)` succeeds so it returns 0.

Maybe one could check if `File.link` is implemented ahead of time (I'm not sure if there's a good way or not), but that still doesn't handle `link(2)` being unable to perform the link.

envygeeks: Let the error surface and throw the application. It's the user and the file systems job to deal with those kinds of problems past making sure the old exists and the new doesn't exist.  We don't validate permissions, too many links, unaddressed links, memory and other stuff that might surface through link into File#link into Ruby's native errors.

As far as checking if link exists, it should exist on all Unix and Unix like machines in one form or another through it's C library, except for Windows... I would simply just check if it's Windows, but at that same point, I'm hesistant to accept a feature that doesn't work across all platforms as we need stop treating Windows like it's second-class.

stevecheckoway: I don't think Windows is the issue. It appears to define [`HAVE_LINK`](https://github.com/ruby/ruby/blob/trunk/win32/Makefile.sub#L730).

The real issue is that `link` failing need not be a user error. It could simply be that `_site` points to a different file system.

envygeeks: HAVE_LINK only decides whether they throw an unavailable error.  It doesn't make it easier to test if symlinks are explicitly supported (that is unless you create a helper that does that out side of this context and tests for that failure directly.)

If your worry is x-links then that's a problem that needs to be solved with a configuration option, not an implicit option that drags down performance without public cause.  I still consider that a user problem but you are right that is a case that need be addressed but to me, the point of least resistance in acceptance is configuring hard links in explicitly and old behavior implicitly.  This way nobody is caught off guard, and we aren't to explain why it's not working when we claim it will.

stevecheckoway:
Per the discussion, I added a configuration option `hardlinks` with a default of `false` in the latest version of this PR.

parkr:
I am generally 👎 on these kinds of changes. This feels like a filesystem issue to me. I have never used hardlinks on Mac or Linux before so I don't know how they'd behave using various filesystems.

stevecheckoway:
I'm not sure I understand your statement or your opposition. Currently, Jekyll is copying each static file. With large files, this is slow.  With the patch, enabling hardlinks causes Jekyll to link the files instead. This is significantly faster as it essentially only has to create a directory entry and increment a link count.

As for OS or file system support, every modern OS supports link. Certainly Windows supports it and it's part of [POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/). As for file system support, it's supported on EXT2/3/4, JFS, ReiserFS, btrfs, FFS, UFS, XFS, HFS+, ZFS,and NTFS (among, I'm sure, many others). The file systems that I know don't support it FAT12/16/32, exFAT, and Plan 9.

I don't understand what you mean about not knowing how they behave on various file systems? "The `link()` function shall atomically create a new link for the existing file and the link count of the file shall be incremented by one."  Do you mean the subtle point in the POSIX specification about how it behaves when linking to a symlink being implementation dependent? If so, that seems easy to canonicalize the behavior by calling `File.realpath` on the source first. Maybe I'm misunderstood your point.

As a single data point, building the small site I'm working on that contains about 4 GB of static files takes about 11 seconds without hard links and about 1.3 seconds with. This is a pretty big win, particularly when running `jekyll serve`.

envygeeks:
>  If so, that seems easy to canonicalize the behavior by calling File.realpath on the source first. Maybe I'm misunderstood your point.

This is certainly something that needs to be taken into consideration, though I don't quite remember if we reject earlier or not, if we do then we are fine, if we don't then there is a chance for them to escape early, which is something we do not wish to happen.  We should probably test this behavior explicitly if this s to get a another vote in favor.

stevecheckoway:
I added documentation for the option.

I'm not sure exactly what more would help this PR land, if anything, but I'm happy to do more as necessary.

stevecheckoway:
Yikes, sorry about that. I forgot that this branch was attached to a pull request still. I was just trying to rebase on top of the latest.

pathawks:
Oh! Sorry, I assumed this was opened by mistake.

ashmaroli:
@stevecheckoway Please fix the branch so that it show `250+` commits from our `master`..
If you don't know how to do it, please enable `maintainer-access` to this branch and I'll push the fix for you..

stevecheckoway:
@ashmaroli Yeah, I really screwed this up. I believe I pulled and rebased my master from upstream/master, then rebased this branch off of that. Clearly, that was the wrong sequence of steps. What should I do in the future to avoid this?

I turned on allow edits from the maintainers.

ashmaroli:
You don't have to `rebase` to get your PR in sync to the latest `master`.. Simply adding a merge-commit (`git pull upstream master && git push origin [branch-name]` would do fine if there are no conflicts. The same thing can be done easily by clicking the `Update Branch` on the GitHub interface.

If there are conflicts, then you can easily resolve them using Github interface as well. On the comman line. first `git pull upstream master`, then resolve conflicts, then commit and push the changes. Rebasing can be done too.. but I won't go into the details of that.. this is not the place.. :wink:

ashmaroli:
@stevecheckoway  Jekyll has changed considerably since the time this PR was opened. Most notably, Jekyll now supports using gem-based themes (which may contain static assets) to fashion a site.
Long story short, if a user has `<source>/assets/images/hero.jpg` and a theme-gem they're using has `<gem>/assets/images/hero.jpg`, the one at the `source` would be *copied* over to the `destination`, along with any another gem assets not overridden (e.g. `<gem>/assets/images/bg.png`).

Now, my question is,
  - How are the hard-links going to be fashioned in the above scenario?
  - A Windows-based user has the `source` on `C:\` and his Ruby environment and therefore his `GEM_HOME` is on `D:\` partition. Would hard-links work as expected in this case and in the case where the user simply *builds* his site (via `jekyll build`) and then pushes the contents of `_site` to his remote host?

pathawks:
I’m pretty 👎🏼 on this idea conceptually. I think this should be handled by copy-on-write at the file system level (like APFS).
I’m not sure that copying static files is going to be the bottleneck for typical sites.

stevecheckoway:
@ashmaroli Ideally, hard links would be made for files on the same file system and files would be copied for files on different file systems. As it's currently implemented, it doesn't check for that and hard links would simply not work in the scenario where the destination of the copy is not the same file system as the source.

@pathawks then I think the PR should just be closed since it seems like Jekyll developers are opposed. I admit, I don't really understand the opposition. Being able to create hard links is a pretty massive speed up for my site which has 5 files that together are more than 7 GB. (They are VM images for my students.)

I also don't really understand the contention that this should be handled by the file system itself. Many file systems support hard links but most don't handle the sort of deduping necessary to make this process fast.

Incremental builds aren't really an option either since they don't properly track changes in data files (or correctly handle generated content in some cases). I spent a little time trying to wrap an external dependency tracking around an incremental Jekyll build, but it was error-prone

I agree that this is unlikely to be a bottleneck for most sites (and I believe the default was to have hard links disabled). But it's a pretty minimal change with nice performance benefits.

parkr:
I'm totally fine adding hardlinks support. There are some very clear benefits. What remains unclear is when to use them instead of copying:

1. Opt-in using a `--hard-links` CLI option or `static_files_hard_links: true` option in the configuration?
2. Always use soft links in `JEKYLL_ENV=development`
3. Always use hard links in `JEKYLL_ENV=development` as long as the source and destination are on the same disk (requires some intimate knowledge of the system, but theoretically doable – does the system throw an error when making a hard link to a file that doesn't exist?)

This indecision ☝️ is what is causing hesitation in my opinion. I think hard links are a wonderful idea, but there are some complications. 🐉

psifertex:
WTB the aforementioned Option 1. 

I've got a few large static files managed in jekyll websites and not only is there the generation cost, but also the on-disk waste of space. 

You could even make it more granular -- I'd be happy even with the ability to specify which files should be hard-linked on a per-file basis. I expect that would solve @stevecheckoway's issue as well since he has a few large files that are the majority of the problem. 

That might also help some of the concern about complexity. If it allows the user to opt-in specifically where they need it, presumably it would only be done for large files. Heck, call the option `large-file-link`. That should prevent any confusion with gem theme overrides.

stevecheckoway:
I rebased on top of master, but made no other changes.

I'm not sure what that travis failure is. I got that locally on master once. It seems like that failure might be nondeterministic.

I note that a "needs-tests" label was added. I'm happy to add tests but would appreciate a suggestion of where the test(s) should be added.

stevecheckoway:
It seems clear this functionality isn't desired. No worries, I can continue to maintain my own fork but having this PR open is kind of a hassle because pushes to my branch of this fork update the PR and annoy everybody.

psifertex:
Unfortunate--at least six of us are going to miss it based on the thumbs. :-)

morfikov:
It's really weird to wait 2-5min for a site to be regenerated when you just fix one misspelled word. :]    

And not 6, only 10... but in reality much more.

PR #5053, opened 2016-07-03
Enable strict (or lax) liquid parsing via a config variable.

stevecheckoway:
Insert

```
liquid:
  error_mode: strict # or lax or warn
```
in `_config.yml` to change the error mode. #4372

### DIFF
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -55,6 +55,10 @@ module Jekyll
       "verbose"           => false,
       "defaults"          => [],
 
+      "liquid"            => {
+        "error_mode" => "warn"
+      },
+
       "rdiscount"         => {
         "extensions" => []
       },
--- a/lib/jekyll/liquid_renderer.rb
+++ b/lib/jekyll/liquid_renderer.rb
@@ -5,6 +5,7 @@ module Jekyll
   class LiquidRenderer
     def initialize(site)
       @site = site
+      Liquid::Template.error_mode = @site.config["liquid"]["error_mode"].to_sym
       reset
     end
 


parkr:
Hey @stevecheckoway, thanks for the pull request. We try not to change configuration settings too often as they cause problems for folks unnecessarily. Can you help explain why you submitted this pull request?

stevecheckoway:
I find having strict liquid parsing helpes diagnose errors. In general, liquid warnings aren't shown to the user (and I failed to fix that). This makes it easy to see those earnings (as errors) without changing the default behavior.

envygeeks:
I've personally had a need for this before, I thought we enabled strict by default now? Or maybe it was removed because there were problems,  can't quite remember.

stevecheckoway:
`:warn` is the default. It uses the same strict parsing but records the first error in the Template object and then attempts `:lax` parsing. The warnings are ignored and I haven't had the opportunity look into why in detail.

Mailing strict parsing the default would certainly work for me.

stevecheckoway:
Actually, I was incorrect. I believed the Liquid documentation:

``` ruby
      # Sets how strict the parser should be.
      # :lax acts like liquid 2.5 and silently ignores malformed tags in most cases.
      # :warn is the default and will give deprecation warnings when invalid syntax is used.
      # :strict will enforce correct syntax.
      attr_writer :error_mode
```
but in reality

``` ruby
      def error_mode
        @error_mode || :lax
      end
```

It's not so hard to log the warnings by setting `:warn` and then plumbing the warnings through `Jekyll::LiquidRender::File`. I created a [branch](https://github.com/jekyll/jekyll/compare/master...stevecheckoway:display-liquid-warnings) that does just that. I can create a PR for that, if desired.

Regardless of displaying the warnings or not, I still think it's useful to be able to make the errors fatal with `:strict`.

parkr:
I am OK with this, then, as long as it doesn't cause trouble for current users. What does a warning message look like?

stevecheckoway:
I simply copied the formatting of the error messages:

```
    Liquid Warning: Liquid syntax error (line 13): Expected end_of_string but found pipe in "a in b | c" in index.md
    Liquid Warning: Liquid syntax error (line 70): Expected end_of_string but found pipe in "a in b | c" in _layouts/index.html
```

The line numbers are off by the number of lines in the YAML front matter, but I haven't had a chance to investigate fixing that. If that number is available somewhere, then it'd be a simple matter to check if the error message is a `Liquid::Error` and increment the line number accordingly. I haven't had a chance to look into that yet, but it's probably a simple matter to modify document.rb to store the number of front matter lines.

Just as an aside, is `Jekyll::Tags::IncludeTagError` used anywhere? If not, it could be removed and the error formatting code simplified.

parkr:
LGTM.

@jekyll/core?

envygeeks:
LGTM.

parkr:
@jekyllbot: merge +minor

PR #5054, opened 2016-07-04
Add Formingo to the list of Jekyll form SaaS

andrewmunsell:
Formingo (https://www.formingo.co) is a new service I'm working on that handles form processing for static websites, such as those generated through Jekyll. Adding it to the list of form services because I believe it'd be a helpful resource and alternative to some of the other listed services.


parkr:
Cool!

@jekyllbot: merge +site

PR #5056, opened 2016-07-05
Add fetch method to Drops

### LINKED ISSUE #4615: New Drops::DocumentDrop page object doesn't support .fetch() method
3.1.2

CentOS 7.2

###### What did you do?
Executing a plugin function that looks up dotted names and returns jekyll hash values, i.e., "site.baseurl", "page.url", etc.  Recursively does this function, i.e.,:

 def lookup(context, name)                                                                                                    
 31      lookup = context                                                                                                           
 32      name.split(".").each { |value| lookup = lookup[value] }                                                                    
 33      lookup                                                                                                                     
 34    end           

The value of context.registers[:page]['url']

NoMethodError: undefined method `fetch' for #<Jekyll::Drops::DocumentDrop:0x00000003390410>

The issue is that there seems to be a new object type, Jekyll::Drops:DocumentDrop introduced in commit 82c3ee365f5a8b948ce2ea853d75a954fa709939, full text:

   Initial work on using Liquid::Drops instead of Hashes.  …
   parkr committed on Dec 21, 2015
   The properties of Liquid::Drops are only evaluated when they're asked for
   and therefore save computation time. This prevents a lot of GC time cleaning
   up objects that are not needed, because they're not created unless requested.
   Additionally, this saves time for actual computation of those values because
   they can be computed only if needed.
   It's funny how much it helps when you only do what is needed. Far less overhead.

The new object doesn't support the standard '.fetch()' method on hashes.  As such, it behaves differently than other non-page objects ... 

i.e:

ls lookup
Liquid::Drop#methods: before_method  context=  has_key?  invoke_drop  to_liquid
Jekyll::Drops::Drop#methods: []  []=  content_methods  each_key  inspect  key?  keys  merge  merge!  to_h  to_hash
Jekyll::Drops::DocumentDrop#methods: 
  collection  content  excerpt  fallback_data  id  next  output  path  previous  relative_path  to_s  url
instance variables: @content_methods  @context  @mutations  @obj

and a standard hash:
ls {}
Enumerable#methods: 
  all?            count       each_cons         entries     flat_map  map     minmax     reduce        take      
  any?            cycle       each_entry        find        grep      max     minmax_by  reverse_each  take_while
  chunk           detect      each_slice        find_all    group_by  max_by  none?      slice_before  to_set    
  collect         drop        each_with_index   find_index  inject    min     one?       sort          zip       
  collect_concat  drop_while  each_with_object  first       lazy      min_by  partition  sort_by     
JSON::Ext::Generator::GeneratorMethods::Hash#methods: to_json
Hash#methods: 
  ==                    default        each_key    has_key?    keep_if  merge!              replace  to_h       values_at
  []                    default=       each_pair   has_value?  key      pretty_print        select   to_hash  
  []=                   default_proc   each_value  hash        key?     pretty_print_cycle  select!  to_liquid
  assoc                 default_proc=  empty?      include?    keys     rassoc              shift    to_s     
  clear                 delete         eql?        index       length   rehash              size     update   
  compare_by_identity   delete_if      fetch       inspect     member?  reject              store    value?   
  compare_by_identity?  each           flatten     invert      merge    reject!             to_a     values   

This breaks a *bunch* of code ... sure, I can work around it, but it seems that having to special case each object type complicates plugin code when in fact you are looking to greatly reduce the complications.  

If there are better, less fragile ways to do the lookup please let me know.
Thanks for any feedback ... good stuff.

-Fred Cassirer

ayastreb:
Fixes #4615 

As pointed out by @parkr - added `fetch` method to `Drops::Drop` class.

### DIFF
--- a/lib/jekyll/drops/drop.rb
+++ b/lib/jekyll/drops/drop.rb
@@ -199,6 +199,10 @@ module Jekyll
           end
         end
       end
+
+      def fetch(*args, &block)
+        to_h.fetch(*args, &block)
+      end
     end
   end
 end
--- a/lib/jekyll/drops/drop.rb
+++ b/lib/jekyll/drops/drop.rb
@@ -200,8 +200,15 @@ module Jekyll
         end
       end
 
-      def fetch(*args, &block)
-        to_h.fetch(*args, &block)
+      # Imitate Hash.fetch method in Drop
+      #
+      # Returns value if key is present in Drop, otherwise returns default value
+      # KeyError is raised if key is not present and no default value given
+      def fetch(key, default = nil, &block)
+        return self[key] if key?(key)
+        raise KeyError, %(key not found: "#{key}") if default.nil? && block.nil?
+        return yield(key) unless block.nil?
+        return default unless default.nil?
       end
     end
   end
--- a/test/test_drop.rb
+++ b/test/test_drop.rb
@@ -27,8 +27,16 @@ class TestDrop < JekyllUnitTest
       assert_equal "default", @drop.fetch("not_existing_key", "default")
     end
 
+    should "fetch default boolean value correctly" do
+      assert_equal false, @drop.fetch("bar", false)
+    end
+
     should "fetch default value from block if key is not found" do
       assert_equal "default bar", @drop.fetch("bar") { |el| "default #{el}" }
+    end
+
+    should "fetch default value from block first if both argument and block given" do
+      assert_equal "baz", @drop.fetch("bar", "default") { "baz" }
     end
   end
 end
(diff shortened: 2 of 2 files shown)


lib/jekyll/drops/drop.rb
```diff
@@ -199,6 +199,10 @@ def merge!(other)
           end
         end
       end
+
+      def fetch(*args, &block)
+        to_h.fetch(*args, &block)
```
parkr: `#to_h` is an expensive operation. Could we mimic the behaviour without having to actually create a hash?

parkr:
This LGTM. @jekyll/core?

envygeeks:
:shipit:

parkr:
Thanks @envygeeks :)

@jekyllbot: merge +minor

PR #5057, opened 2016-07-06
Added JekyllConf page

mneumegen:
Lists all the JekyllConf talks. Also features one talk every build.

### DIFF
--- a/site/_data/docs.yml
+++ b/site/_data/docs.yml
@@ -40,6 +40,7 @@
   - troubleshooting
   - sites
   - resources
+  - jekyllconf
   - upgrading/0-to-2
   - upgrading/2-to-3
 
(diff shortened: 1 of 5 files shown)


envygeeks:
I'm :shipit: but this should be two commits not one.  There are other changes here.

parkr:
I'm thinking this should be just at `jekyllrb.com/jekyllconf/` as it's not a docs page. But other than that, LGTM!

mneumegen:
@envygeeks Yep sorry forgot I fixed that typo.

@parkr Sweet! Agreed but I'm not sure how to do this. The way the navigation is built it assumes the links are prefixed with /docs/:

`{% assign item_url = item | prepend:"/docs/" | append:"/" %}`

parkr:
@mneumegen It'd be a stand-alone page, I think. I was thinking about adding a "Community" section that'd behave like `/docs/` or `/news/`, but have community-related resources.

PR #5060, opened 2016-07-06
Stringify configuration overrides before first use

stomar:
This fixes a regression with configuration override options where the keys are not stringified before the first use (namely `config` and `skip_initial_build`).
The regression was introduced with version 3.1.4, commit d01f7943debf08dee645990ddb0e00d59cf846bb.

See the example below (with 3.1.5, since requiring of 3.1.4 fails for me):

```
$ cat _my_config.yml
exclude:
  - file

$ cat test-3.1.3.rb
gem "jekyll", "3.1.3"
require "jekyll"

p Jekyll.configuration(:config  => "_my_config.yml")["exclude"]
p Jekyll.configuration("config" => "_my_config.yml")["exclude"]

$ cat test-3.1.5.rb
gem "jekyll", "3.1.5"
require "jekyll"

p Jekyll.configuration(:config  => "_my_config.yml")["exclude"]
p Jekyll.configuration("config" => "_my_config.yml")["exclude"]
```

Running the example with 3.1.5 produces:

```
Configuration file: none
[]
Configuration file: _my_config.yml
["file"]
```

while for 3.1.3 all is fine:

```
Configuration file: _my_config.yml
["file"]
Configuration file: _my_config.yml
["file"]
```

### DIFF
--- a/test/test_configuration.rb
+++ b/test/test_configuration.rb
@@ -334,6 +334,18 @@ class TestConfiguration < JekyllUnitTest
         Jekyll.configuration(test_config.merge({ "config" => @paths[:other] }))
     end
 
+    should "load different config if specified with symbol key" do
+      allow(SafeYAML).to receive(:load_file).with(@paths[:default]).and_return({})
+      allow(SafeYAML)
+        .to receive(:load_file)
+        .with(@paths[:other])
+        .and_return({ "baseurl" => "http://example.com" })
+      allow($stdout).to receive(:puts).with("Configuration file: #{@paths[:other]}")
+      assert_equal \
+        site_configuration({ "baseurl" => "http://example.com" }),
+        Jekyll.configuration(test_config.merge({ :config => @paths[:other] }))
+    end
+
     should "load default config if path passed is empty" do
       allow(SafeYAML).to receive(:load_file).with(@paths[:default]).and_return({})
       allow($stdout).to receive(:puts).with("Configuration file: #{@paths[:default]}")
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -101,6 +101,7 @@ module Jekyll
     # Returns the final configuration Hash.
     def configuration(override = {})
       config = Configuration.new
+      override = Configuration[override].stringify_keys
       unless override.delete("skip_config_files")
         config = config.read_config_files(config.config_files(override))
       end


parkr:
Cool. Would you please add a test? Thanks!

envygeeks:
I'm :shipit: after @parkr's requests are met.

test/test_configuration.rb
```diff
@@ -335,6 +335,18 @@ class TestConfiguration < JekyllUnitTest
         Jekyll.configuration(test_config.merge({ "config" => @paths[:other] }))
     end
 
+    should "load different config if specified with symbol key" do
+      allow(SafeYAML).to receive(:load_file).with(@paths[:default]).and_return({})
```
stomar: (in the case when the config option is not recognited properly, this line produces a regular test failure, instead of an exception due to wrongly set up stub)

stomar:
Test case added.

I'm not very familiar with the code base and even less with the test suite, though. Hope it's ok.

stomar:
Note: it seems there sometimes occur random test failures on Travis. My original commit had two unrelated failures, after pushing again all was fine.

parkr:
This LGTM. @envygeeks want to take one more look?

test/test_configuration.rb
```diff
@@ -335,6 +335,18 @@ class TestConfiguration < JekyllUnitTest
...
+      allow(SafeYAML)
+        .to receive(:load_file)
+        .with(@paths[:other])
+        .and_return({ "baseurl" => "http://wahoo.dev" })
+      allow($stdout).to receive(:puts).with("Configuration file: #{@paths[:other]}")
+      assert_equal \
+        site_configuration({ "baseurl" => "http://wahoo.dev" }),
```
envygeeks: `.dev` is no longer a private extension, lets just make this example.com which is the IANA example domain that should be save (and is encouraged) as the test example.

stomar: wahoo.dev is used in various places; should I change the new test case now or will you change all occurrences after merging?

stomar:
@envygeeks Never mind the previous comment; applied your suggestion.

See also PR #5068 for other occurrences of `wahoo.dev`.

parkr:
LGTM.

parkr:
@jekyllbot: merge +bug

PR #5065, opened 2016-07-07
Remove useless statement from Configuration test

stomar:
IMO the removed line doesn't do anything useful.


parkr:
It doesn't, you're right. Good catch!

@jekyllbot: merge +dev

PR #5067, opened 2016-07-07
Add missing fields to ExcerptDrop

parkr:
Fix! I noticed that if you tried inspecting an `ExcerptDrop`, it threw an error because of missing fields, e.g. `previous` and `next`'s dependencies `previous_doc` and `next_doc`.

This PR adds the missing elements of the Excerpt so the ExcerptDrop functions properly.


parkr:
@jekyllbot: merge +bug

PR #5068, opened 2016-07-08
Change baseurl to example.com for some test cases

stomar:


parkr:
Thanks :)

@jekyll: merge +dev

PR #5098, opened 2016-07-12
Updates html-proofer code

esasse:
As in html-proofer docs: https://github.com/gjtorikian/html-proofer


parkr:
@jekyllbot: merge +site

PR #5100, opened 2016-07-12
use activesupport < 5 if we are on a Ruby < 2.2.2

Crunch09:
ActiveSupport 5.0.x depends on [Ruby >= 2.2.2](https://github.com/rails/rails/blob/48512d790680a4fbfc20026f3c976f8fa759b2e5/activesupport/activesupport.gemspec#L10).

I only added it to the `test_legacy` and `site` group because activesupport is an indirect dependency of `shoulda` (via `shoulda-matchers`) and an indirect dependency of `jekyll-mentions` and `jemoji` (both via `html-pipeline`.)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,9 @@ source "https://rubygems.org"
 gemspec :name => "jekyll"
 
 gem "rake", "~> 11.0"
+
+gem "activesupport", "~> 4.2", :groups => [:test_legacy, :site] if RUBY_ENGINE == 'ruby' && RUBY_VERSION < '2.2.2'
+
 group :development do
   gem "launchy", "~> 2.3"
   gem "pry"


Crunch09:
One test failed for 2.2.4 and jRuby but it seems unrelated?

Gemfile
```diff
@@ -2,6 +2,11 @@ source "https://rubygems.org"
 gemspec :name => "jekyll"
 
 gem "rake", "~> 11.0"
+
+if RUBY_ENGINE == 'ruby' && RUBY_VERSION < '2.2.2'
+  gem "activesupport", "< 5", :groups => [:test_legacy, :site]
```
parkr: What from `:site` depends on this gem?

I'd love to see this only take up one line.

Probably best to use `~> 4.2` instead of `< 5`?

parkr:
A couple things in my comment above. :smile: Thanks!

Gemfile
Crunch09: Thanks for the review! It's used in [html-pipeline](https://rubygems.org/gems/html-pipeline) which is a dependency of both [jemoji](https://rubygems.org/gems/jemoji) and [jekyll-mentions](https://rubygems.org/gems/jekyll-mentions). Will try to get it into one line and change the version requirement.

Crunch09:
Done :)

Gemfile
```diff
@@ -2,6 +2,9 @@ source "https://rubygems.org"
 gemspec :name => "jekyll"
 
 gem "rake", "~> 11.0"
+
+gem "activesupport", "~> 4.2", :groups => [:test_legacy, :site] if RUBY_ENGINE == 'ruby' && RUBY_VERSION < '2.2.2'
```
parkr: Is `RUBY_ENGINE` necessary here?

parkr:
Thank you! One more question, thanks for sticking with me!

envygeeks:
I thought we dropped activesupport, why did it make a come back?

envygeeks:
To also add, this isn't really necessary, or shouldn't be.  AFAIW both Gem and Bundler now do this automatically.  In that if we want Rails and we do `~>4` and it finds 5.0 and it's not supported on 2.1 then it will install the latest version of "4" that is.  Otherwise it will install the latest version of 5.x.  I've only heard rumors of this and never actually tested.

pathawks:
Travis was [choking](https://travis-ci.org/jekyll/jekyll/jobs/144208193#L465) earlier.

parkr:
@envygeeks We're using jekyll-mentions to automatically mentionify `@mention`s in jekyllrb.com. activesupport is a dependency of html-pipeline, which is used by jekyll-mentions.

LGTM.

@jekyllbot: merge +dev

Gemfile
Crunch09: @parkr Good question, i used Rails' [ruby_version_check.rb](https://github.com/rails/rails/blob/52ce6ece8c8f74064bb64e0a0b1ddd83092718e1/railties/lib/rails/ruby_version_check.rb) as an example but could be possible to leave it out i guess.

Issue #5071, opened 2016-07-09
_data file keys and values get assigned as integers, should be strings


pakamore:
EDITED: My bug report was wrong. It's a feature request.

parkr:
Pass both `page.category` and `mycategory` through the `inspect` filter – does `mycategory` have a newline at the end?

pakamore:
The issue was how {% assign %} treats integers.
```
{% capture category_id %}2484881380235357006{% endcapture %}
{{ category_id | inspect }} – &quot;2484881380235357006&quot;
{% assign category_idd = 2484881380235357006 %}
{{ category_idd | inspect }} – 2484881380235357006
```
These values return `false` when comparing with `==` operator. In such case it`s probably my mistake. I should use:
```
{% assign category_idd = "2484881380235357006" %}
```
However, my problem was when these IDs are assigned from a `_data` YML file. There I had:
```
2484881380235357006:
    id: 2484881380235357006
    link: long-permalink
    title: Nice title
```
The key and `id` value were not escaped as strings, and other values were. So I was not able to compare them with other strings from Jekyll (site.*, page.*, etc. – they are all inside quotation marks). When I changed my `_data` YML file to such, everything works fine now:
```
"2484881380235357006":
    id: "2484881380235357006"
    link: long-permalink
    title: Nice title
```
I believe it's a bug. In Liquid docs I was not able to find any methods for converting strings to integers and vice versa. So all values should be strings in Jekyll, no matter whether assigned with `"` or without.
This is especially important in `_data` files. For me it's more common to see keys without `"`. Jekyll `_config.yml` does not put keys inside `"` either.

parkr:
> So all values should be strings in Jekyll, no matter whether assigned with " or without.

Of course, this would break countless sites so we can't do this. What we can do is add filters.

@fw42 👋 Have you ever considered adding type-conversion filters to Liquid? We'd like to be able to run `{{ some_val | to_integer }}` or something similar to convert a value to an integer.

crispgm:
@parkr Is there any way to register Liquid filter in Jekyll?
The ```to_integer``` filter can be easy, but I failed to register the filter unless we add the implementing code in Liquid lib.

parkr:
@crispgm Yes. See the bottom of filters.rb

pakamore:
@parkr, assign format is under the user`s control, so I agree there is no need to change it. But I'm more concerned about `_data` files, where the user has no control how values get assigned.

[...]

PR #5101, opened 2016-07-13
Add filter: to_integer

crispgm:
Fix or support #5071


crispgm:
Sorry for broke the checks.
I succeeded in passing all checks but failed on travis.

```
lib/jekyll/filters.rb:7:3: C: Metrics/ModuleLength: Module has too many lines. [242/240]
```
Can I get some help for fix error?

lib/jekyll/filters.rb
```diff
@@ -260,6 +260,22 @@ def where_exp(input, variable, expression)
...
+    #
+    # Returns the integer value
+    def to_integer(input)
+      return input if input.is_a?(Integer)
+      return 1 if input == true
+      return 0 if input == false
+      begin
+        input.to_i
+      rescue
+        raise ArgumentError, "Invalid input object type."
```
parkr: This error should be more descriptive:

> Object '#{input.inspect}' could not be converted into an integer.

parkr:
Above line 7 where it says `module Filters`, add `# rubocop:disable Metrics/ModuleLength`. :shrug:

lib/jekyll/filters.rb
```diff
@@ -273,7 +273,7 @@ def to_integer(input)
       begin
         input.to_i
       rescue
-        raise ArgumentError, "Invalid input object type."
+        raise ArgumentError, "Object '#{input.inspect}' could not be converted into an integer."
```
parkr: To fix the `fmt` error, move this to two lines:

```ruby
raise ArgumentError, 
  "Object '#{input.inspect}' could not be converted into an integer."
```

crispgm: Thanks. It should be solved.
I learned the lesson that ```script/fmt``` should be run before commits. 😄

parkr:
LGTM.

parkr:
@jekyll/core: would someone else mind taking a look? Just need another LGTM.

lib/jekyll/filters.rb
```diff
@@ -260,6 +261,23 @@ def where_exp(input, variable, expression)
...
+    #
+    # Returns the integer value
+    def to_integer(input)
+      return input if input.is_a?(Integer)
+      return 1 if input == true
+      return 0 if input == false
+      begin
+        input.to_i
+      rescue
+        raise ArgumentError,
```
envygeeks: IMO, this can just be removed, let Ruby's native error surface.

lib/jekyll/filters.rb
```diff
@@ -260,6 +261,23 @@ def where_exp(input, variable, expression)
       end
     end
 
+    # Convert the input into integer
+    #
+    # input - the object string
+    #
+    # Returns the integer value
+    def to_integer(input)
+      return input if input.is_a?(Integer)
```
envygeeks: This is uncessary, Ruby is already optimized for this and just slows down to_i already.

envygeeks:
Just a few comments! Thanks.

lib/jekyll/filters.rb
crispgm: It's okay but if this was removed, users would get a ```undefined method `to_i'``` instead.
Will this error message mislead users?

envygeeks: There is nothing ambiguous about that error.  Our only job is to do what we are told to do, we can either be defensive or we can throw, if we choose to throw then the native error will suffice, they told us to run to_i, and to_i doesn't exist, therefore the native error is the most literally right, we need not rescue that as it was exactly what happened, if we choose to be defensive then return 0 like Ruby does with unpareseable strings.

crispgm: Thanks for your review and advice. The modification has been done. 😸

envygeeks:
LGTM after a quick rebase!

parkr:
LGTM, thank you.

parkr:
LGTM.

parkr:
2 LGTM's were given above. 👍 

@jekyllbot: merge +minor

PR #5112, opened 2016-07-18
Slightly, improve PNG compression.

XhmikosR:
Before: 101 KB (103.742 bytes)
After:  100 KB (102.817 bytes)


envygeeks:
Seeing as there are several vulnerabilities surrounding PNG, provide your compression method so we can verify it before accepting this.  I'm hesitant to accept fonts and images without verification.

XhmikosR:
No, I'm trying to inject malware specific to jekyllrb.com.

Come on, not this crap again.

envygeeks:
It is what it is,  whether your like it or not.  I'm not good at my job because I'm lax.

XhmikosR:
I don't want to flame this anymore, but try to distinguish the people.

```
zopflipng.exe -m --lossy_transparent -y --iterations=1000
```

parkr:
@envygeeks If we could add that to CI, that would be great. Do we already have security vulnerability scanning? If not, then we may not be able to reliably catch things like a potential malware from our known contributors. I certainly wouldn't suspect @XhmikosR or any of our other previous contributors of placing malware in an asset like this. Maybe the system could ensure we'd catch it and do so without having to ask tough and sometimes interrogative questions like this ourselves.

@XhmikosR Haha less than 1,000 bytes! But I'll take it. Thanks!

@jekyllbot: merge +site

envygeeks:
@parkr I can have Hulk scan it behind the scenes like we do our own assets if you want.

parkr:
@envygeeks That would work.

PR #5115, opened 2016-07-18
Add AppVeyor support.

XhmikosR:
This is the second attempt after a long time. Now AppVeyor is 10x faster so there shouldn't be a problem there.

Someone from the repo owners (/CC @parkr) should enable AppVeyor for the whole repo when the errors are fixed.

I tried to mimic the Travis config as close as possible.

BTW, if you want to see what is currently failing, check https://ci.appveyor.com/project/XhmikosR/jekyll until someone enables AppVeyor for this repo.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -1,10 +1,11 @@
 # [Jekyll](https://jekyllrb.com/)
 
 [![Gem Version](https://img.shields.io/gem/v/jekyll.svg)][ruby-gems]
-[![Build Status](https://travis-ci.org/jekyll/jekyll.svg?branch=master)][travis]
-[![Test Coverage](https://codeclimate.com/github/jekyll/jekyll/badges/coverage.svg)][coverage]
-[![Code Climate](https://codeclimate.com/github/jekyll/jekyll/badges/gpa.svg)][codeclimate]
-[![Dependency Status](https://gemnasium.com/jekyll/jekyll.svg)][gemnasium]
+[![Linux Build Status](https://img.shields.io/travis/jekyll/jekyll/master.svg?label=Linux%20build)][travis]
+[![Windows Build status](https://img.shields.io/appveyor/ci/jekyll/jekyll/master.svg?label=Windows%20build)][appveyor]
+[![Test Coverage](https://img.shields.io/codeclimate/coverage/github/jekyll/jekyll.svg)][coverage]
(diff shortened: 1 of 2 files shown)


XhmikosR:
BTW, if you want to see what is currently failing, check https://ci.appveyor.com/project/XhmikosR/jekyll until someone enables AppVeyor for this repo.

parkr:
Thanks, @XhmikosR! Is there any configuration required on the AppVeyor side that I have to do in order to get this working?

parkr:
LGTM. @jekyll/core, you down to add this? The tests will fail for a while but all the more reason to fix them. We have historically not supported Windows because we don't have development machines which run Windows but we could get closer to supporting it with this CI strategy.

/cc @jekyll/windows

XhmikosR:
@parkr: only enable AppVeyor for the repo and perhaps enable "skip branches without appveyor.yml". Everything else should be set up already.

XhmikosR:
@parkr:

>... and enable "skip branches without appveyor.yml"

XhmikosR:
One thing that's different compared to Travis is

1. No jruby
2. fast_finish

For the first we can't do anything, but should I remove `fast_finish`?

envygeeks:
@XhmikosR 

> No jruby

This is fine, our JRuby support is waiting for me (or really anybody) to get the time to finish cleaning it up so that @jekyll/core can make it official.  Right now it's lowdown supported simply because we use JRuby here where I am.

XhmikosR:
@envygeeks: How about `fast_finish`?

envygeeks:
@XhmikosR if we slim down what Rubies we support and can go parallel on Cucumber and Units I don't think that'll really be a big problem.  We use fast_finish on Travis because we've so many versions to test on. If @jekyll/core is cool I think for AppVeyor we only test on the most common version of Windows Ruby for now.

envygeeks:
Let me rephrase this: "if we slim down what Rubies we support"
It should really read: "If we slim down what Rubies we test on."

parkr:
> If @jekyll/core is cool I think for AppVeyor we only test on the most common version of Windows Ruby for now.

:+1: on this. No need for `fast_finish` or jruby on Windows.

parkr:
LGTM.

envygeeks:
LGTM.

parkr:
@jekyllbot: merge +dev

PR #5118, opened 2016-07-19
site: fix validation errors.

XhmikosR:
Only issue left is

>Error: A document must not include both a meta element with an http-equiv attribute whose value is content-type, and a meta element with a charset attribute.

Does anyone know where `http-equiv` comes from?

### DIFF
--- a/site/_includes/footer.html
+++ b/site/_includes/footer.html
@@ -1,4 +1,4 @@
-<footer role="contentinfo">
+<footer>
   <div class="grid">
     <div class="unit one-third center-on-mobiles">
       <p>The contents of this website are <br />&copy;&nbsp;{{ site.time | date: '%Y' }} under the terms of the <a href="{{ site.repository }}/blob/master/LICENSE">MIT&nbsp;License</a>.</p>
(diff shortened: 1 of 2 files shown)


site/_includes/footer.html
```diff
@@ -1,4 +1,4 @@
-<footer role="contentinfo">
+<footer>
```
parkr: You added these in https://github.com/jekyll/jekyll/pull/3256 – should we modify the validation to accept them, or are they no longer needed?

pathawks:
> Does anyone know where `http-equiv` comes from?

sparklemotion/nokogiri#1008

site/_includes/footer.html
XhmikosR: It was a bug in the validator back then.

XhmikosR:
@pathawks: thanks. I commented there, we'll see how it goes.

site/_includes/footer.html
parkr: Are you saying then that this is not a valid attribute?

XhmikosR: No, it's redundant. You can always run the validator against the live site
to see the warning.

Ideally you should have the vnu checker but I'm not sure there is a gem nor
can I help integrating...

parkr:
LGTM.

parkr:
@jekyllbot: merge +site

PR #5127, opened 2016-07-23
filters: raise error if no input given to date filters

### LINKED ISSUE #5107: Filters#time should NOT exit; it should raise an error
If you have the Liquid `{{ "" | date_to_xml_string }}`, it will explode with the error:

```text
 Invalid Date: '' is not a valid datetime.
  Liquid Exception: exit in sitemap.xml
```

We should not use `exit(1)` or `abort` in any of our liquid processing code. This should be replaced with a custom error.

@ayastreb, @pathawks, either one of you want to fix this? 😄  Should be fairly straight-forward.

ayastreb:
Fixes #5107

### DIFF
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -373,8 +373,8 @@ module Jekyll
       when Numeric
         Time.at(input)
       else
-        Jekyll.logger.error "Invalid Date:", "'#{input}' is not a valid datetime."
-        exit(1)
+        raise Errors::InvalidDateError,
+          "Invalid Date: '#{input.inspect}' is not a valid datetime."
       end.localtime
     end
 
--- a/test/test_filters.rb
+++ b/test/test_filters.rb
@@ -268,6 +268,15 @@ class TestFilters < JekyllUnitTest
           )
         end
       end
+
+      context "without input" do
+        should "raise an error if input is nil" do
+          err = assert_raises Jekyll::Errors::InvalidDateError do
+            @filter.date_to_xmlschema(nil)
+          end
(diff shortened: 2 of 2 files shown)


lib/jekyll/filters.rb
```diff
@@ -373,8 +373,8 @@ def time(input)
       when Numeric
         Time.at(input)
       else
-        Jekyll.logger.error "Invalid Date:", "'#{input}' is not a valid datetime."
-        exit(1)
+        raise Errors::InvalidDateError,
+          "Invalid Date: '#{input}' is not a valid datetime."
```
ayastreb: I'm not sure about error message here.
Theoretically it should only get here if input is not given, so maybe it should be something like:
`No input given to filter #{filter_name} in #{document_path}` ?
Can we determine somehow the path of the document where the error is?

parkr: @ayastreb When an error is raised in Liquid, the renderer catches that and prints the document for which the error occurred so that's all take care of. What's important is to ensure the input is displayed properly. How about `"Invalid Date: \"#{input.inspect}\" is not a valid time."` ?

parkr: Adding the `.inspect` call means instead of `Invalid Date: "" is not a valid datetime.`, you'd see `Invalid Date: "nil" is not a valid datetime.`

parkr:
LGTM!

ayastreb:
@parkr, yes that sounds good! I've chnaged the error message to include `.inspect`.

parkr:
LGTM.

parkr:
@jekyllbot: merge +bug

PR #5128, opened 2016-07-24
Add a baseurl option to the build subcommand, resolves #4919

### LINKED ISSUE #4919: Add --baseurl option to build subcommand

---

- [x] I am on (or have tested on) ***Mac OS*** 10+

---

Please see #2874 for previous discussion. Thanks to @envygeeks and @parkr for suggesting opening a new issue.

## Background
I currently develop and deploy multiple prototypes of sites that I am working on to show experimental implementations and modifications to colleagues. 
I have a web server which, using apache, serves up a particular directory e.g. /var/www/proto to example.com/proto
I deploy these prototypes in subdirectories, e.g. /var/www/proto/feature-x becomes example.com/proto/feature-x

## Current process
To automate this deployment I use fabric (http://www.fabfile.org/) to run commands locally and on the server to build the static site and deploy it.

To deploy a new feature to a new subdirecory I have to generate a new config file and either update or create a new fabfile to use the new config and directory.

## Improvement / Feature request
If I could supply `--baseurl` to the `build` subcommand the process would be much simpler as I could just pass this value to fabric, it could use it to create the correct directory on the server and use it on the jekyll `build` subcommand to have _site generated correctly.

## Alternative / work-around
Techincally I could pass `--baseurl` to fabric and have it start and stop the Jekyll `serve` subcommand with `--baseurl` instead. This has the same outcome for the generated _site as being able to pass baseurl to `build`, but this seems a bit heavy handed. 

This is my first feature request on anything. Any improvements or clarification I can make please make suggestions etc.
Cheers.

chrisfinazzo:


lib/jekyll/command.rb
```diff
@@ -55,6 +55,7 @@ def add_build_options(c)
         c.option "limit_posts", "--limit_posts MAX_POSTS", Integer,
           "Limits the number of posts to parse and publish"
         c.option "watch", "-w", "--[no-]watch", "Watch for changes and rebuild"
+        c.option "baseurl", "-b", "--baseurl", "Serve the website from the given base URL"
```
parkr: This option takes an argument so the third argument here must be `--baseurl URL` or something so OptionsParser does the Right Thing. Without this, it will only return `true` (when present) or `false` (when omitted).

parkr:
Fixes #4919. One comment above.

chrisfinazzo:
Yeah, I should have seen this earlier. The key-value pair for source_dir has a similar pattern. Will update.

chrisfinazzo:
why u no work

`script/ci` and `script/fmt` check out locally.

parkr:
@chrisfinazzo If you rebase, you should be good.

PR #5129, opened 2016-07-24
Show liquid warnings.

stevecheckoway:
Liquid warnings are not shown (unless liquid's `error_mode` is set to `strict` in which case they are errors). This change causes the warnings to be printed as warnings, but does not stop the build.

For example, given the erroneous code

    {% for a in b | c %}{% endfor %}

in `index.md`, the following warning message is displayed

<pre>    Liquid Warning: Liquid syntax error (line 13): Expected end_of_string but found pipe in "a in b | c" in index.md</pre>

This is identical to the `strict` case except it says "Liquid Warning" rather than "Liquid Exception".

The line numbers displayed do not count any front matter lines, but this is already true of the exceptions.

### DIFF
--- a/lib/jekyll/convertible.rb
+++ b/lib/jekyll/convertible.rb
@@ -114,16 +114,16 @@ module Jekyll
     #
     # Returns the converted content
     def render_liquid(content, payload, info, path)
-      site.liquid_renderer.file(path).parse(content).render!(payload, info)
-    rescue Tags::IncludeTagError => e
-      Jekyll.logger.error(
-        "Liquid Exception:",
-        "#{e.message} in #{e.path}, included in #{path || self.path}"
-      )
-      raise e
+      template = site.liquid_renderer.file(path).parse(content)
+      template.warnings.each do |e|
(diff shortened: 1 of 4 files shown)


stevecheckoway:
I fixed the overly-long line error. I'm not sure what the other errors are, but I can't reproduce them locally and they don't seem related to this change.

lib/jekyll/convertible.rb
```diff
@@ -114,16 +114,16 @@ def converters
...
-      )
-      raise e
+      template = site.liquid_renderer.file(path).parse(content)
+      template.warnings.each do |e|
+        Jekyll.logger.warn "Liquid Warning:",
+          LiquidRenderer.format_error(e, path || self.path)
+      end
+      template.render!(payload, info)
     # rubocop: disable RescueException
     rescue Exception => e
```
parkr: We should probably change this to just `rescue => e`.

parkr:
LGTM.

parkr:
@jekyllbot: merge +minor

PR #5130, opened 2016-07-24
add `plugins` config key as replacement for `gems`

### LINKED ISSUE #4509: Rename the "gems" config key to "plugins"
What do you think about renaming the `gems` config directive to `plugins`, with full backwards compatibility?

The idea being, if you're not from the Ruby world, "gems" isn't a very intuitive term (even though that's what they are). Further, users shouldn't have to care about (or see) the behind-the-scenes plumbing, in terms of how things are distributed.

Instead, let's call the key "plugins", a term which has a broader understanding outside the Ruby ecosystem, and one that should be familiar to any user coming from a more traditional CMS like WordPress or Drupal.

In terms of practicality, it'd just be a matter of combining the `gems` and `plugins` arrays, if they exist when we load the config, and updating the docs in the next minor release.

Crunch09:
`gems` won't be deprecated. This fixes #4509 .

### DIFF
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -229,9 +229,10 @@ module Jekyll
       # Provide backwards-compatibility
       check_auto(config)
       check_server(config)
+      check_plugins(config)
 
       renamed_key "server_port", "port", config
-      renamed_key "plugins", "plugins_dir", config
+      renamed_key "plugins", "gems", config, false
       renamed_key "layouts", "layouts_dir", config
       renamed_key "data_source", "data_dir", config
 
@@ -280,11 +281,11 @@ module Jekyll
       config
     end
 
-    def renamed_key(old, new, config, _ = nil)
+    def renamed_key(old, new, config, with_deprecation_warning = true)
       if config.key?(old)
         Jekyll::Deprecator.deprecation_message "The '#{old}' configuration" \
           " option has been renamed to '#{new}'. Please update your config" \
-          " file accordingly."
+          " file accordingly." if with_deprecation_warning
         config[new] = config.delete(old)
       end
     end
@@ -383,6 +384,23 @@ module Jekyll
           "We recommend you switch to Kramdown. To do this, replace " \
           "`markdown: maruku` with `markdown: kramdown` in your " \
           "`_config.yml` file."
+      end
+    end
+
+    # Private: Checks if the `plugins` config is a String
+    #
+    # config - the config hash
+    #
+    # Raises a Jekyll::Errors::InvalidConfigurationError if the config `plugins`
+    # is a string
+    private
+    def check_plugins(config)
+      if config.key?("plugins") && config["plugins"].is_a?(String)
+        Jekyll.logger.error "Configuration Error:", "You specified the" \
+          " `plugins` config in your configuration file as a String, please" \
+          " use an Array instead. If you wanted to set the directory of your" \
+          " plugins, use the config key `plugins_dir` instead."
+        raise Jekyll::Errors::InvalidConfigurationError, "plugins: #{config["plugins"]}"
       end
     end
   end
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -401,7 +401,7 @@ module Jekyll
           " use an array instead. If you wanted to set the directory of your" \
           " plugins, use the config key `plugins_dir` instead."
         raise Jekyll::Errors::InvalidConfigurationError,
-          "'plugins' should not be a string, but was: "\
+          "'plugins' should not be a string, but was: " \
           "#{config["plugins"].inspect}. Use 'plugins_dir' instead."
       end
     end
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -232,7 +232,7 @@ module Jekyll
       check_plugins(config)
 
       renamed_key "server_port", "port", config
-      renamed_key "gems", "plugins", config, false
+      renamed_key "gems", "plugins", config
       renamed_key "layouts", "layouts_dir", config
       renamed_key "data_source", "data_dir", config
 
@@ -281,11 +281,11 @@ module Jekyll
       config
     end
 
-    def renamed_key(old, new, config, with_deprecation_warning = true)
+    def renamed_key(old, new, config, _ = nil)
       if config.key?(old)
         Jekyll::Deprecator.deprecation_message "The '#{old}' configuration" \
           " option has been renamed to '#{new}'. Please update your config" \
-          " file accordingly." if with_deprecation_warning
+          " file accordingly."
         config[new] = config.delete(old)
       end
     end
@@ -397,11 +397,12 @@ module Jekyll
     def check_plugins(config)
       if config.key?("plugins") && config["plugins"].is_a?(String)
         Jekyll.logger.error "Configuration Error:", "You specified the" \
-          " `plugins` config in your configuration file as a String, please" \
-          " use an Array instead. If you wanted to set the directory of your" \
+          " `plugins` config in your configuration file as a string, please" \
+          " use an array instead. If you wanted to set the directory of your" \
           " plugins, use the config key `plugins_dir` instead."
         raise Jekyll::Errors::InvalidConfigurationError,
-          "plugins: #{config["plugins"].inspect}"
+          "'plugins' should not be a string, but was: "\
+          "#{config["plugins"].inspect}. Use 'plugins_dir' instead."
       end
     end
   end
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -400,7 +400,8 @@ module Jekyll
           " `plugins` config in your configuration file as a String, please" \
           " use an Array instead. If you wanted to set the directory of your" \
           " plugins, use the config key `plugins_dir` instead."
-        raise Jekyll::Errors::InvalidConfigurationError, "plugins: #{config["plugins"]}"
+        raise Jekyll::Errors::InvalidConfigurationError,
+          "plugins: #{config["plugins"].inspect}"
       end
     end
   end
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -33,7 +33,7 @@ module Jekyll
 
       # Plugins
       "whitelist"         => [],
-      "gems"              => [],
+      "plugins"           => [],
 
       # Conversion
       "markdown"          => "kramdown",
@@ -232,7 +232,7 @@ module Jekyll
       check_plugins(config)
 
       renamed_key "server_port", "port", config
-      renamed_key "plugins", "gems", config, false
+      renamed_key "gems", "plugins", config, false
       renamed_key "layouts", "layouts_dir", config
       renamed_key "data_source", "data_dir", config
 
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -6,9 +6,9 @@ module Jekyll
     attr_reader   :source, :dest, :config
     attr_accessor :layouts, :pages, :static_files, :drafts,
                   :exclude, :include, :lsi, :highlighter, :permalink_style,
-                  :time, :future, :unpublished, :safe, :plugins, :limit_posts,
+                  :time, :future, :unpublished, :safe, :plugins_dir, :limit_posts,
                   :show_drafts, :keep_files, :baseurl, :data, :file_read_opts,
-                  :gems, :plugin_manager, :theme
+                  :plugins, :plugin_manager, :theme
 
     attr_accessor :converters, :generators, :reader
     attr_reader   :regenerator, :liquid_renderer, :includes_load_paths
@@ -45,7 +45,7 @@ module Jekyll
       @config = config.clone
 
       %w(safe lsi highlighter baseurl exclude include future unpublished
-        show_drafts limit_posts keep_files gems).each do |opt|
+        show_drafts limit_posts keep_files plugins).each do |opt|
         self.send("#{opt}=", config[opt])
       end
 
@@ -414,7 +414,7 @@ module Jekyll
     private
     def configure_plugins
       self.plugin_manager = Jekyll::PluginManager.new(self)
-      self.plugins        = plugin_manager.plugins_path
+      self.plugins_dir    = plugin_manager.plugins_path
     end
 
     private
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -45,9 +45,11 @@ module Jekyll
       @config = config.clone
 
       %w(safe lsi highlighter baseurl exclude include future unpublished
-        show_drafts limit_posts keep_files gems).each do |opt|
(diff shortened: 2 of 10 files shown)


lib/jekyll/configuration.rb
```diff
@@ -226,9 +226,10 @@ def backwards_compatibilize
       # Provide backwards-compatibility
       check_auto(config)
       check_server(config)
+      check_plugins(config)
 
       renamed_key "server_port", "port", config
-      renamed_key "plugins", "plugins_dir", config
+      renamed_key "plugins", "gems", config, false
```
parkr: I think the order of the arguments is wrong here. The first argument is the old config, in this case `"gems"`. The second is the new config name, in this case `"plugins"`. 😄

lib/jekyll/configuration.rb
```diff
@@ -277,11 +278,11 @@ def add_default_collections
       config
     end
 
-    def renamed_key(old, new, config, _ = nil)
+    def renamed_key(old, new, config, with_deprecation_warning = true)
       if config.key?(old)
         Jekyll::Deprecator.deprecation_message "The '#{old}' configuration" \
           " option has been renamed to '#{new}'. Please update your config" \
-          " file accordingly."
+          " file accordingly." if with_deprecation_warning
```
parkr: I think we should keep the deprecation warning.

lib/jekyll/configuration.rb
```diff
@@ -382,5 +383,22 @@ def check_maruku(config)
...
+    # Raises a Jekyll::Errors::InvalidConfigurationError if the config `plugins`
+    # is a string
+    private
+    def check_plugins(config)
+      if config.key?("plugins") && config["plugins"].is_a?(String)
+        Jekyll.logger.error "Configuration Error:", "You specified the" \
+          " `plugins` config in your configuration file as a String, please" \
+          " use an Array instead. If you wanted to set the directory of your" \
+          " plugins, use the config key `plugins_dir` instead."
+        raise Jekyll::Errors::InvalidConfigurationError, "plugins: #{config["plugins"]}"
```
parkr: This should use `config["plugins"].inspect` so we see a nice stringified version of the contents, e.g. `plugins: "./plugins"` instead of `plugins: ./plugins`.

parkr:
Just two quick comments.

@benbalter, anything further here?

benbalter:
> won't be deprecated.

✨ . This is exactly what I was proposing. Nice work. A quick win for usability.

lib/jekyll/configuration.rb
Crunch09: Thanks! Will change this.

lib/jekyll/configuration.rb
Crunch09: It was intentional 😊  This way we can still use `config.gems` internally but i can change this of course

Crunch09:
@parkr Thanks for the review! I tried to address your comments and also changed `Site#gems` and `Site#plugins` to `Site#plugins` and `Site#plugins_dir` to match their `Configuration` counterparts.
I didn't change the display of a deprecation warning so far because of @benbalter 's comment.

lib/jekyll/configuration.rb
```diff
@@ -226,9 +226,10 @@ def backwards_compatibilize
       # Provide backwards-compatibility
       check_auto(config)
       check_server(config)
+      check_plugins(config)
 
       renamed_key "server_port", "port", config
-      renamed_key "plugins", "plugins_dir", config
+      renamed_key "gems", "plugins", config, false
```
parkr: Based on the presence of `false`, this could just be `config["gems"] = config["plugins"] if config.key?("plugins")`

parkr: That said, I still think we should be deprecating `gems`. 😄

lib/jekyll/configuration.rb
```diff
@@ -382,5 +383,23 @@ def check_maruku(config)
...
+    # is a string
+    private
+    def check_plugins(config)
+      if config.key?("plugins") && config["plugins"].is_a?(String)
+        Jekyll.logger.error "Configuration Error:", "You specified the" \
+          " `plugins` config in your configuration file as a String, please" \
+          " use an Array instead. If you wanted to set the directory of your" \
+          " plugins, use the config key `plugins_dir` instead."
+        raise Jekyll::Errors::InvalidConfigurationError,
+          "plugins: #{config["plugins"].inspect}"
```
parkr: Let's update this to `'plugins' should not be a string, but was: #{config["plugins"].inspect}. Use 'plugins_dir' instead.`

lib/jekyll/site.rb
```diff
@@ -6,9 +6,9 @@ class Site
     attr_reader   :source, :dest, :config
     attr_accessor :layouts, :pages, :static_files, :drafts,
                   :exclude, :include, :lsi, :highlighter, :permalink_style,
-                  :time, :future, :unpublished, :safe, :plugins, :limit_posts,
+                  :time, :future, :unpublished, :safe, :plugins_dir, :limit_posts,
                   :show_drafts, :keep_files, :baseurl, :data, :file_read_opts,
-                  :gems, :plugin_manager, :theme
+                  :plugins, :plugin_manager, :theme
```
parkr: Hm, I'm concerned this might be considered a breaking change...

envygeeks: I've the same concern as @parkr, we should probably alias the old values over until 4.0?

lib/jekyll/site.rb
```diff
@@ -45,7 +45,7 @@ def config=(config)
       @config = config.clone
 
       %w(safe lsi highlighter baseurl exclude include future unpublished
-        show_drafts limit_posts keep_files gems).each do |opt|
+        show_drafts limit_posts keep_files plugins).each do |opt|
```
envygeeks: ""

lib/jekyll/site.rb
```diff
@@ -415,7 +415,7 @@ def site_cleaner
     private
     def configure_plugins
       self.plugin_manager = Jekyll::PluginManager.new(self)
-      self.plugins        = plugin_manager.plugins_path
+      self.plugins_dir    = plugin_manager.plugins_path
```
envygeeks: ""

envygeeks:
This is great, just a few comments!

lib/jekyll/configuration.rb
Crunch09: Good catch, this solution didn't come to my mind at all 😆 . I'll update it though with the deprecation of `gems`.

Crunch09:
Changed it back so `Site#plugins` and `Site#gems` doesn't change their behavior.

Also, `Config#gems` IS now deprecated.

DirtyF:
@Crunch09 sorry for not merging this sooner, could you please fix the conflicts before @jekyll/stability can review this PR?

Crunch09:
@DirtyF Thank you for taking a look at this! I just rebased against master and hope the tests still run 😄  If i should squash the commits as well just let me know

pnn:
@Crunch09 i believe the tests are failing because of this line: https://github.com/Crunch09/jekyll/blob/965421600ba7e7cbe90f33a9c34587e5f010b37a/test/test_plugin_manager.rb#L135

fmt failures should be related to #5987 and therefore irrelevant

Crunch09:
@pnn Thank you! The Travis UI unfortunately doesn't seem to load for me but i'll investigate and fix later today.

pnn:
@Crunch09 same here 😭 i just looked at appveyor instead

Crunch09:
@pnn Ha, good idea! :) Travis seems to be working again now and i fixed the failing test

lib/jekyll/configuration.rb
```diff
@@ -385,5 +386,24 @@ def check_maruku(config)
...
+    #
+    # Raises a Jekyll::Errors::InvalidConfigurationError if the config `plugins`
+    # is a string
+    private
+    def check_plugins(config)
+      if config.key?("plugins") && config["plugins"].is_a?(String)
+        Jekyll.logger.error "Configuration Error:", "You specified the" \
+          " `plugins` config in your configuration file as a string, please" \
+          " use an array instead. If you wanted to set the directory of your" \
+          " plugins, use the config key `plugins_dir` instead."
```
benbalter: Would [squigly heredoc](https://infinum.co/the-capsized-eight/multiline-strings-ruby-2-3-0-the-squiggly-heredoc) be easier here?

lib/jekyll/site.rb
```diff
@@ -45,10 +45,12 @@ def config=(config)
       @config = config.clone
 
       %w(safe lsi highlighter baseurl exclude include future unpublished
-        show_drafts limit_posts keep_files gems).each do |opt|
+        show_drafts limit_posts keep_files).each do |opt|
         self.send("#{opt}=", config[opt])
       end
 
+      self.gems = config["plugins"]
```
benbalter: I suspect this is here because we want to maintain the Ruby API even if the user-facing configuration key is changing? If so, I'd add a quick line comment here so a well-meaning developer doesn't "optimize" this in the future.

lib/jekyll/configuration.rb
DirtyF: @benbalter sure but heredoc only works on Ruby 2.3, isn't that a concern if we want to merge this before v4?

benbalter: It looks like squiggly heredoc came in 2.3, but heredoc worked prior... either way, not worth over-optimizing something so minor. 👍

envygeeks: `<<-TEXT` would be a deoptimization as you would have to strip the heredoc and multiple lines to get it to display nicely (so that it can fold.) This avoids that issue while making it readable (albeit more annoying for the developer.)

lib/jekyll/site.rb
Crunch09: Good point, thank you! I'll add it.

lib/jekyll/configuration.rb
Crunch09: For the mentioned reasons, i'm gonna keep it that way but thank you for mentioning the squigly heredoc, i didn't know about this!

Crunch09:
Addressed the latest feedback, let me know if anything else is missing 🎉

pnn:
are we waiting on something with this or is it good to merge?

DirtyF:
@pnn you're right, let's merge this!

@jekyllbot: merge +minor

sourcejedi:
> gems won't be deprecated.

>  	deprecate `gems` 

Ok.  But aren't you missing an update to `lib/site_template/_config.yml` ?

(Feel free to poke if I'm being too lazy by commenting on a merged PR, but this is still relatively recent).

    $ jekyll --version
    jekyll 3.5.0
    $ jekyll new .
    ...
    $ bundle exec jekyll serve
    Configuration file: /home/sysnote/publish/_config.yml
           Deprecation: The 'gems' configuration option has been renamed to 'plugins'. Please update your config file accordingly.

DirtyF:
@sourcejedi exact. Watch #6045, waiting for GitHub Pages to upgrade to Jekyll 3.5+

ashmaroli:
Can we remove the `Deprecation:` error message in `v3.5.0` till we start development on `v4`? Its kinda annoying seeing that message, and doesn't seem right having everyone switch to new `plugins` key in the middle of `minor-release-cycle`.

/cc @parkr, @Crunch09, @pathawks

envygeeks:
That makes no sense, that's the entire point...

ashmaroli:
> that's the entire point

okay, can it be toned down to an `info` level message at least?

Crunch09:
I agree with @envygeeks that we need to keep the deprecation message so we can remove the old functionality in Jekyll 4.
I'm afraid `info` is too easy to be overseen. Maybe `warn` [instead of `error`](https://github.com/jekyll/jekyll/blob/57fd5f887da1189a16bdfbb982d75f725c38d725/lib/jekyll/deprecator.rb#L41) would be the right choice for deprecation messages instead?

envygeeks:
`WARN` is fitting, that's how I was informed of changes in Ruby 2.4 going into the next major version and only on the CI which explicitly enables warnings.

ashmaroli:
> so we can remove the old functionality in Jekyll 4

Yes, that's true.. But then, we're nowhere near even starting development on a `v4`, right? (Or are we?) So, i thought we'll display that message when we're at around `v3.8.0`..

Crunch09:
I don't know what the release plan will be until Jekyll 4 but removing the deprecation now and bringing it back later would be very confusing i think.

pathawks:
> So, i thought we'll display that message when we're at around v3.8.0..

You are planning to release 3.8.0?? News to me.

Many users have experienced pain when something unexpectedly changes and their site won't build all of a sudden. This is why we offer deprecation messages: to give them plenty of time to address these things.

We *know* we will change this when we have an opportunity; why would we *not* communicate that to users? Why would we want to give users **less** warning to update their site?

pathawks:
> Its kinda annoying seeing that message

You could always make it go away by, you know, following its advice.

parkr:
I don't see a path for Jekyll 4 yet, as we don't have anything absolutely crucial that we can't do in a backwards-compatible way. I would like to keep our tradition of shipping backwards-compatible changes instead of breaking things.

This deprecation is annoying for users who want to use this version but are shipping out to GitHub Pages where the plugins key is not yet supported. Pages will upgrade in due time; for now, I recommend downgrading to Jekyll v3.4.5 for all Pages users' day-to-day needs.

parkr:
If the message isn't clear or isn't clearly actionable, then I'm open to a PR that rewords the message for clarity.

ashmaroli:
Looks like I stirred up a hornet's nest of sorts..
I surrender :white_flag:

:fire:

parkr:
All good! Just wanted to be clear here. :smiley:

envygeeks:
> I would like to keep our tradition of shipping backwards-compatible changes instead of breaking things.

Might as well throw that out the window @parkr.  Jekyll 3.5 was a total disaster for some plugins because of breaking core changes.  Namely in context, and I'm sure there are hundreds of plugins that haven't even noticed yet, but we felt that pain on Jekyll-Assets when we had to wrap around the breaking class to support everything BEFORE 3.5 and everything AFTER 3.5.  You forced us to make a split on logic based on a breaking change already :wink:

parkr:
I'm sorry to hear it was a total disaster. Most plugins I use still worked great. I know jekyll-assets uses some APIs that we don't consider public, but the principle is the same. I'm sorry for shipping breaking changes. We'll try to do better next time. When we upgrade in a minor release, we'll make sure it's simply adding support for the new version rather than removing support for the old version as we did with Liquid.

We want to upgrade to Rouge 2. I'll be sure to ship an rc first and give two weeks before we ship the finished product so you we can test things out. I'll open an issue on jekyll-assets when the rc for 3.6 ships.

deanilvincent:
screen shot 2017-09-12 at 8 09 31 pm

I got this error while trying to run my github page website locally. Any help?

sourcejedi:
@deanilvincent  The Deprecation message is not the error.  The only error is the last line.  There is no reason here to believe it is related to this PR.

deanilvincent:
Thank you @sourcejedi :)

Issue #5064, opened 2016-07-07
Why are _posts and _data special collections?


frickler01:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I Updated to the latest Jekyll (or) if on Github Pages to the latest `github-pages`
- [x] I Read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [x] This is a feature request. Well, sort of.

This is an architectural queston, so the stuff above does not really apply.

Users are currently forced to have a `_posts` collection, because it is hardwired into Jekyll. It is not possible to delete it from the data structure at all. If you want to iterate over all collections and do not use `_posts`, you have to filter it out explicitly.

Also `_posts` documents are treated differently than documents in other collections. For example posts don't need a YAML front matter to be converted, while other documents do. And you cannot change the sort order of posts if you treat them as posts, only if you treat them as a collection `posts`. So `_posts` is and _is not_ a collection. This is inconsistent and was enormously confusing to me as a newcomer.

There is at least another person having a [related issue](https://talk.jekyllrb.com/t/suppress-generation-of-pages-for-posts/2423).

I also do not see the purpose of a special `_data` collection, since it does not add any functionality.
The use case described in the docs

```
{% for foo in site.data.bar %}
{{ foo.key }}
{% endfor %}
```

can be covered by any collection, since YAML data is parsed in for each file anyway. One would only need to make sure that other structured files like CSV or JSON are treated equivalently.

I had a brief look at the code and there are a bunch of special cases implemented to create this behavior. Removing them would make the whole thing more generic and homogeneous, shortening the code without losing functionality. One would need to add the `posts` collection to the default `_config.yml` to give the user something to start with, though.

For both issues, the current transformation rules (based on filename prefix and YAML front matter) may be in the way, but I have seen at least [one proposal to restructure it](https://github.com/jekyll/jekyll/issues/290#issuecomment-1264028) (conversion based on file extension).

I see how such a change might affect backwards compatibility and that it involves quite some refactoring.

Though my question is: Other than for historical reasons, why has it ever been done this way? It seems redundant.

jake-low:
A proposal (from a novice in Jekyll's architecture):

What if the `_posts` and `_data` collections could be disabled under the `collections` key in `_config.yml`, just as other collections can be enabled? Effectively they'd be just like any other custom collection, but they default to `output: true` instead of `output: false`.

Then, you could disable posts like this:

```yaml
collections:
  posts:
    output: false
[...]

PR #5131, opened 2016-07-25
Update the configuration docs to match the code

stevecheckoway:
The current documentation of the default options is wrong. I've tried to make them agree. The major things

- coderay is no longer in the defaults
- kramdown's `input: GFM` is already the default so I removed the section talking about adding it
- I added documentation for liquid's `error_mode`

### DIFF
--- a/site/_docs/configuration.md
+++ b/site/_docs/configuration.md
@@ -571,11 +571,13 @@ file or on the command-line.
 # Where things are
 source:       .
 destination:  ./_site
-plugins_dir:  ./_plugins
-layouts_dir:  ./_layouts
-data_dir:     ./_data
-includes_dir: ./_includes
-collections:  null
+plugins_dir:  _plugins
+layouts_dir:  _layouts
+data_dir:     _data
+includes_dir: _includes
+collections:
+  posts:
+    output:   true
 
 # Handling Reading
 safe:         false
@@ -607,6 +609,7 @@ detach:  false
 port:    4000
 host:    127.0.0.1
 baseurl: "" # does not include hostname
+show_dir_listing: false
 
 # Outputting
 permalink:     date
@@ -614,7 +617,11 @@ paginate_path: /page:num
 timezone:      null
 
 quiet:    false
+verbose:  false
 defaults: []
+
+liquid:
+  error_mode: warn
 
 # Markdown Processors
 rdiscount:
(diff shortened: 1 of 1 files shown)


site/_docs/configuration.md
```diff
@@ -571,11 +571,11 @@ file or on the command-line.
...
 destination:  ./_site
-plugins_dir:  ./_plugins
-layouts_dir:  ./_layouts
-data_dir:     ./_data
-includes_dir: ./_includes
-collections:  null
+plugins_dir:  _plugins
+layouts_dir:  _layouts
+data_dir:     _data
+includes_dir: _includes
```
parkr: Why change these but not modify `./_site`?

parkr:
Other than my two comments above, LGTM.

site/_docs/configuration.md
```diff
@@ -571,11 +571,11 @@ file or on the command-line.
...
+collections:  {}
```
stevecheckoway: It doesn't [appear](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/configuration.rb#L15) to be.

site/_docs/configuration.md
stevecheckoway: The definition for `destination` seems correct. It's not relative to `source` the way the others are.
```
    DEFAULTS = Configuration[{
      # Where things are
      "source"            => Dir.pwd,
      "destination"       => File.join(Dir.pwd, "_site"),
      "plugins_dir"       => "_plugins",
      "layouts_dir"       => "_layouts",
      "data_dir"          => "_data",
      "includes_dir"      => "_includes",
      "collections"       => {},
```

site/_docs/configuration.md
stevecheckoway: I see now. Is there a reason that the DEFAULTS hash does not actually contain that? As far as I can tell, it only forces `posts: output` to be true even if explicitly set to false. Seems odd for `posts` to be special in this way.

frickler01: @stevecheckoway `posts` is a very special flower in many ways, see #5064.

stevecheckoway:
New version addresses the comment about `posts`.

site/_docs/configuration.md
parkr: They're resolved using `Jekyll.sanitized_path(site.source, the_dir)` so it shakes out to be the same. :shrug:

parkr:
LGTM.

@jekyllbot: merge +site

PR #5135, opened 2016-07-25
Add `--baseurl` to `build` subcommand

chrisfinazzo:
I had this rebased and ready to go yesterday, but ran into [this][] issue - which explains the Travis failures I had been seeing. I've since updated and I think this is ready to go.

[this]: https://github.com/jekyll/jekyll/commit/83a72606b3a0bb89696ce69bcf6f6de971137fb4#commitcomment-18372995

PR #5136, opened 2016-07-25
Run script/fmt before CI

chrisfinazzo:
Per this [comment][], do you think it is worthwhile to add `script/fmt` to the CI script to prevent simple formatting errors? When doing work on what eventually became #5135, I was unsure about why the error kept appearing simply because I didn't know the bug existed.

Tl;dr Stupid mistakes are stupid, let's make fewer of them 😀.

[comment]: https://github.com/jekyll/jekyll/pull/5128#issuecomment-234817525

### DIFF
--- a/script/cibuild
+++ b/script/cibuild
@@ -4,6 +4,8 @@ script/branding
 
 set -e
 
+script/fmt
+
 if [[ -z "$TEST_SUITE" ]]
 then
   script/test ci


pathawks:
Wouldn’t it slow down CI builds to do this _every time_? CI is already running `script/fmt` once.

chrisfinazzo:
I guess it could be conditional depending on what's been modified, perhaps only within `/lib` or `/features` where Rubocop is going to walk the directory tree? Moreover, I feel bad sending a request that I know will fail because of simple mistakes that could be caught earlier.

While this does run on Travis' end, I didn't notice a big slowdown - might be different for larger changes 🤔

cc\ @jekyll/core @jekyll/performance

crispgm:
@chrisfinazzo I agree with you. I got into the same situation in my first PR.

XhmikosR:
AppVeyor just runs the same script, `scripts/ci`. One would just need to remove the `fmt` from the matrix` if this goes in.

parkr:
We added this as a separate CI target on purpose. It only needs to be run once – this is its own test suite in a sense. I would accept a PR that included it in the `if [ -z "$TEST_SUITE" ]` section, but it should not be run before every cucumber or minitest run as it would given your patch above.

DirtyF:
Maybe, it's worth mentionning to check your code locally with `script/fmt` before submitting a PR in  https://jekyllrb.com/docs/contributing/#code-contributions

parkr:
> Maybe, it's worth mentionning to check your code locally with script/fmt before submitting a PR in https://jekyllrb.com/docs/contributing/#code-contributions

:+1:

chrisfinazzo:
> Maybe, it's worth mentionning to check your code locally with script/fmt before submitting a PR in https://jekyllrb.com/docs/contributing/#code-contributions

> 👍

@DirtyF In retrospect, I think that's the way to go - keep it simple.

PR #5138, opened 2016-07-25
Add a note about script/fmt

chrisfinazzo:


parkr:
LGTM. Thanks!

envygeeks:
Rubocop should be under the Github Styleguide because we most importantly use the Github Styleguide to setup Rubocop in a similar format.  We do not use many of the rules they define by default.

parkr:
@envygeeks Are you saying you don't like the wording of Chris's addition?

parkr:
@jekyllbot: merge +site

Issue #4644, opened 2016-03-10
Idea: Jekyll live-reload


benbalter:
This may be scope/feature creep, and may be better suited for a plugin, but when Jekyll is in watch mode locally, if I have a file open in my browser, and I edit that file, when Jekyll's finished rebuilding the site, it should refresh the page for me. 

I don't know about you, but at least for me, for now when I make a change to a blog post, I then switch windows and compulsively hit refresh for 5-10 seconds while the site rebuilds. This feels like an easy way to delight users.

This could be done, for example, through websockets, with the Jekyll server sending a command to the browser when the build is done. This of course, would only be used locally.

benbalter:
Also to note, if you want to play around with this idea, check out @jldec's awesome [pub server](https://github.com/jldec/pub-server), which does this very well.

chrisfinazzo:
I asked @parkr this [question][] a while back and he suggested that in lieu of a plugin, this might work better as a subcommand. I know that in prior versions, people have [used][] Guard in interesting ways to try and make it work. Originally, my thought was to try and make a plugin, especially now that Mercenary will do a fair amount of the initial setup for you.

I'll look into [pub-server][] when I have a moment, that's one I haven't seen before. 

[question]: https://github.com/jekyll/jekyll/issues/2354
[used]: http://stackoverflow.com/questions/8395526/using-live-reload-with-jekyll/19173204#19173204
[pub-server]: https://github.com/jldec/pub-server

parkr:
Yeah, I'd definitely say not to put this in core (extra dependencies), but maybe as a plugin we recommend?

[livereload](https://rubygems.org/gems/livereload) and [rack-livereload](https://rubygems.org/gems/rack-livereload) both seem like good places to start. `jekyll-livereload` is free on rubygems.org! :smile:

jaybe-jekyll:
http://livejs.com

Client/js-resource-based solution, and/but quite simple and effective.

@benbalter the inline/online editing capabilities within pub server is fascinating and intriguing.

Exciting!

DirtyF:
This is It's true that [Hugo comes with LiveReload built in](http://gohugo.io/extras/livereload/) and it definitively improves the developer experience. 

Another related feature that would be useful as a plugin : Browsersync. (See [Netlify's Browsersync gulp task](https://github.com/netlify/netlify-home/blob/master/gulpfile.js#L28) for a possible implementation with Jekyll.)

envygeeks:
[...]

PR #5142, opened 2016-07-27
Add LiveReload functionality to Jekyll.

awood:
Invoke using `jekyll serve --livereload`.  The Serve command will inject
some necessary JavaScript links into pages.  When loaded in a browser,
this JavaScript opens a connection over port 35729 using WebSockets to
an EventMachine reactor listening to that port in a separate thread.

When Jekyll finishes rendering a page, `--livereload` hooks create a
JSON message with the relative URL of the refreshed page.  This message
is then pushed to the browser over the WebSockets connection.  The
client-side JavaScript then reloads the page if the URL in the message
matches the URL of the current page.

Implements #4644


awood:
This PR doesn't include any significant documentation on the feature, but I figured I'd wait until some of the details get hammered out before writing anything.

Other notes:
* It may be desirable to prune the number of options available to configure LiveReload.  Specifically, the Flash version for [browsers that don't support WebSockets](http://caniuse.com/#search=websockets) might be worth dropping since no one is going to be developing there Jekyll site in Android and the overlap of IE 8 only users and Jekyll users is probably pretty small.
* The semantics around `--ignore` are weird could probably lead to false bug reports.  From my experiments and reading of the JavaScript (which I did not write), if any portion of URL the browser is pointed to matches the URL to reload in the JSON message, the browser will refresh.  This is especially noticeable on index pages ("/blog/" would match every refresh message for pages in the directory)
* The tests are very much functional tests instead of unit tests.  In my opinion the functionality is complex enough to merit it, but others may disagree.

lib/jekyll/commands/serve.rb
```diff
@@ -13,9 +15,22 @@ class << self
           "show_dir_listing"   => ["--show-dir-listing",
             "Show a directory listing instead of loading your index file."],
           "skip_initial_build" => ["skip_initial_build", "--skip-initial-build",
-            "Skips the initial site build which occurs before the server is started."]
+            "Skips the initial site build which occurs before the server is started."],
+          "livereload"         => ["-l", "--livereload",
+            "Use LiveReload to automatically refresh browsers"],
+          "swf"                => ["--swf", "Use Flash for WebSockets support"],
```
envygeeks: This is a no-go.  Everybody is dropping support for flash faster than we want to turn around and have to remove it.

lib/jekyll/commands/serve.rb
```diff
@@ -13,9 +15,22 @@ class << self
...
+          "ignore"             => ["--ignore GLOB1[,GLOB2[,GLOB3...]]", Array,
```
envygeeks: If these options belong to live-reload directly and not to Jekyll globally then please prefix them with `--live-reload-`

lib/jekyll/commands/serve.rb
```diff
@@ -33,26 +48,117 @@ def init_with_program(prog)
...
+        end
+
+        #
+
+        def shutdown
+          @server.shutdown if running?
+        end
+
+        private
+        # rubocop:disable Metrics/CyclomaticComplexity
```
envygeeks: This is a no-go.  Please cleanup your method, it's pretty complex.

lib/jekyll/commands/serve.rb
```diff
@@ -173,31 +281,51 @@ def enable_logging(opts)
         # forget to add one of the certificates.
 
         private
-        # rubocop:disable Metrics/AbcSize
         def enable_ssl(opts)
-          return if !opts[:JekyllOptions]["ssl_cert"] && !opts[:JekyllOptions]["ssl_key"]
-          if !opts[:JekyllOptions]["ssl_cert"] || !opts[:JekyllOptions]["ssl_key"]
+          jekyll_opts = opts[:JekyllOptions]
```
envygeeks: Why is this necessary?

awood:
@envygeeks do you want revisions as separate commits to squash merge or for me to just adjust this current commit?

envygeeks:
@awood whatever works best for you and allows you to flow the best! We normally only ask for a rebase just before we merge it, before that, it's whatever workflow you wish to use since Github now has that fancy "view all files changed".

lib/jekyll/commands/serve.rb
awood: > Why is this necessary?

It's not strictly necessary but it keeps the line length down, makes the code more readable (IMHO), and saves dereferencing the opts variable over and over for the same thing.  I can put it back if you want.

envygeeks:
I kinda like this, maybe we should add an updater like we've for normalize.css so that we can auto-update it every once in a while as it'll be quite easy to forget it exists.  Other than that just a few comments and we can do another deeper review.

awood:
> I kinda like this, maybe we should add an updater like we've for normalize.css so that we can auto-update it every once in a while as it'll be quite easy to forget it exists. Other than that just a few comments and we can do another deeper review.

So you automatically pull the file every so often?  That's a really good idea.  I think I will need to modify the code somewhat because it's been modified slightly to read from variables defined in the inserted script tags, but I think I can figure something else out so the stock livereload.js will work

lib/jekyll/commands/serve.rb
awood: > This is a no-go. Please cleanup your method, it's pretty complex.

Okay, the complexity decreased below the threshold with the latest commit, but it is still fairly ugly.  I'm not sure about the best strategy.  Maybe just drop the warning that `--livereload` and `--no-watch` don't make any sense?

awood:
> So you automatically pull the file every so often? That's a really good idea. I think I will need to modify the code somewhat because it's been modified slightly to read from variables defined in the inserted script tags, but I think I can figure something else out so the stock livereload.js will work

Okay, I've implemented the above.  The PR now just uses the stock version of the most recent released `livereload.js`.  Pulling it in via `bower` might be a good way of making sure it stays up to date, but that is something for the maintainers to decide.

awood:
Looks like this is failing in JRuby.  I think the issue has to do with the WebSockets server thread introduced in this PR.  I'll do some investigation, but if anyone out there has some experience with the subtleties of JRuby, I'd wouldn't mind a second opinion.

awood:
So I've made some changes for JRuby, but JRuby's TLS support [is](https://github.com/eventmachine/eventmachine/issues/34) [broken](https://github.com/eventmachine/eventmachine/issues/180).  I can either remove the TLS support for the WebSockets connection entirely or just ignore those options when running under JRuby (not sure if that's an acceptable solution though).

Additionally, while the tests pass, the actual functionality under JRuby is broken.  For whatever reason, the message to reload is sent but the browser doesn't react.  I'll look into this issue later and try and get some tests that actually ensure receipt of the reload command.

awood:
I have tracked the JRuby functionality issue down to a problem within either EM-Websocket or EventMachine.  I've filed an [issue](https://github.com/igrigorik/em-websocket/issues/140) with EM-Websocket.

awood:
@envygeeks the Travis failures for this PR appear to be code style releated in code that I haven't modified.  Otherwise I think this PR is ready for review at least.  There's an outstanding issue with EM-Websocket/EventMachine under JRuby but I don't think that should block the review process.  I can go ahead and address any issues the maintenance team sees while waiting to hear back from on the issue I filed for the JRuby bug.

envygeeks:
@awood I'll take some time this week to start reviewing /cc @jekyll/core @jekyll/ecosystem

awood:
Looks like the JRuby issue is a [known problem](https://github.com/eventmachine/eventmachine/issues/477) with an EventMachine reactor running under a thread.

awood:
An alternative to having LiveReload just not work under JRuby would be to switch from WEBrick to an EventMachine compatible http server.  Jekyll could then just start the EventMachine reactor when it's time to start serving and the reactor would manage both the LiveReload server on 35729 and the http server on 4000.

I actually have a working proof-of-concept using Thin as the http server, and it's pretty nice since it removes all the need for threading from Jekyll proper and just delegates everything to the EventMachine event loop.  Plus EventMachine can daemonize so LiveReload would continue to work in daemon mode (whereas it does not in this PR since that would require Jekyll spinning off two processes which I thought a little dodgy especially if one process dies and the other doesn't).

The downside is that Thin doesn't run under JRuby at all.  There are JRuby compatible http server (notably Puma) but they aren't EventMachine based and so there would have to be some sort of low-level shim layer that brokered between EventMachine's Connection class and the http server which would probably end up being a nightmare.

But if Thin ever runs under JRuby, I think the strategy described above would be a worthwhile refactoring.

lib/jekyll/commands/serve.rb
```diff
@@ -31,27 +49,118 @@ def init_with_program(prog)
...
-              Build.process(opts)
-              opts["config"] = config
-              Serve.process(opts)
+
+              # Check to make sure the specified options make sense logically
+              # before setting defaults.
+              validate_options(opts)
+
+              opts["watch"] = true unless opts.key?("watch")
+              opts["livereload_port"] ||= LIVERELOAD_PORT
```
spudowiar: You've used `Hash#key?` for `"watch"` and `||=` for `"livereload_port"`

lib/jekyll/commands/serve.rb
```diff
@@ -31,27 +49,118 @@ def init_with_program(prog)
...
+            if opts["detach"]
+              Jekyll.logger.abort_with "Error:",
+                "--detach and --livereload are mutually exclusive"
+            end
+            unless opts["watch"]
+              Jekyll.logger.warn "Using --livereload without --watch defeats the purpose"\
+                " of LiveReload."
+            end
+          elsif opts["livereload_min_delay"] ||
+              opts["liverealod_max_delay"]   ||
```
spudowiar: Typo!

```patch
-              opts["liverealod_max_delay"]   ||
+              opts["livereload_max_delay"]   ||
```

lib/jekyll/commands/serve.rb
```diff
@@ -31,27 +49,118 @@ def init_with_program(prog)
...
+              opts["livereload_ignore"]      ||
+              opts["livereload_port"]
+            Jekyll.logger.warn "The --livereload-min-delay, --livereload-max-delay, "\
+               "--livereoload-ignore, and --livereload-port options require the "\
```
spudowiar: Typo!

```diff
-               "--livereoload-ignore, and --livereload-port options require the "\
+               "--livereload-ignore, and --livereload-port options require the "\
```

lib/jekyll/commands/serve.rb
spudowiar: I'd suggest reverting this and splitting the method up into multiple methods.

lib/jekyll/commands/serve.rb
```diff
@@ -172,31 +283,51 @@ def enable_logging(opts)
...
           opts[:SSLEnable] = true
         end
 
         private
-
         def start_callback(detached)
           unless detached
             proc do
-              Jekyll.logger.info("Server running...", "press ctrl-c to stop.")
+              @running << "."
```
spudowiar: I feel that `Queue` is the wrong data type for this usage. Correct me if I'm wrong, but you never seem to actually wait for the `Queue` to become populated and even if you did, there are probably better data types for this usage.

spudowiar:
@awood 

> An alternative to having LiveReload just not work under JRuby would be to switch from WEBrick to an EventMachine compatible http server.
> The downside is that Thin doesn't run under JRuby at all.

:laughing:

I've never liked LiveReload but I'm sure it'll be a very useful feature for many Jekyll users.

envygeeks:
I've never really found it useful, I can F5 as fast as that thing loads lol.  But at this point we need to bring in @jekyll/ecosystem and @jekyll/core so they can comment on your suggestions.

lib/jekyll/commands/serve.rb
awood: Yeah, I did that because "watch" can legitimately be `false`

spudowiar: Oops, my mistake! :blush:

lib/jekyll/commands/serve.rb
awood: I'm using it as a mechanism to block the test from running until the server(s) can successfully start.  The only place it's actually waiting for the queue to be populated is in `test_commands_serve.rb`.  The test starts a `Serve` thread and `Serve` starts a WEBrick thread and joins it so there's not another good way I could think of to know whether WEBrick is actually up and running.

I think Queue is the appropriate choice here; the docs say "[The Queue] class provides a way to synchronize communication between threads."

That being said, I do think that it's kind of crummy to have cruft in there that's only really used for the tests.  I'm totally open to suggestions on a better way to handle this situation.

spudowiar:
But you're polling in the test, not using the actual `Queue`!

awood:
> But you're polling in the test, not using the actual `Queue`!

The test polls`running?` which is in turn based on whether or not the `@running` Queue is empty or nil.

spudowiar:
@awood that defeats the purpose of using `Queue`, a `Queue` is designed so you can `Queue#pop` without polling

awood:
> @awood that defeats the purpose of using Queue, a Queue is designed so you can Queue#pop without polling

Yeah, I'm using the Queue more like "if the Queue has anything in it, that means Webrick and EventMachine are running" and it only gets cleared when the servers shutdown.  So the queue is really just a semaphore type thing I suppose.  I'm open to suggestions on the implementation.  My main goal is to avoid sporadic test failures due to tests running before the server is up.  Maybe using a Mutex would make more sense?

awood:
Is there anyway to unregister a hook once it's been registered?  The test failures are order dependent (seed 52792 fails while seed 59920 passes).  I think what's happening is that the LiveReload hook is getting registered and then later `Serve` is getting reinitialized and resetting a variable that the hook references to nil.  I can work around that by just always initializing a LiveReload reactor but only start it when the `--livereload` option is given, but I think it'd be better to have my tests clean up after themselves.

envygeeks:
@awood there isn't, because hooks aren't named.

awood:
I'm not sure how to interpret those Appveyor build errors.  Most of them look like code I haven't touched.

ashmaroli:
@awood, Appveyor has sprung errors all over the Jekyll codebase. I've extracted the relevant error for this PR to below:

```
Error:
TestCommandsServe#test_: using LiveReload should serve livereload.js over HTTPS. :
Errno::ECONNREFUSED: No connection could be made because the target machine actively refused it.  
- connect(2) for "localhost" port 35729 (localhost:35729)
    C:/projects/jekyll/vendor/bundle/ruby/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient/session.rb:604:in `initialize'
    C:/projects/jekyll/vendor/bundle/ruby/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient/session.rb:604:in `new'
    C:/projects/jekyll/vendor/bundle/ruby/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient/session.rb:604:in `create_socket'
    C:/projects/jekyll/vendor/bundle/ruby/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient/ssl_socket.rb:18:in `create_socket'
    C:/projects/jekyll/vendor/bundle/ruby/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient/session.rb:746:in `block in connect'
    C:/Ruby23/lib/ruby/2.3.0/timeout.rb:91:in `block in timeout'
    C:/Ruby23/lib/ruby/2.3.0/timeout.rb:101:in `timeout'
    C:/projects/jekyll/vendor/bundle/ruby/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient/session.rb:742:in `connect'
    C:/projects/jekyll/vendor/bundle/ruby/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient/session.rb:504:in `query'
    C:/projects/jekyll/vendor/bundle/ruby/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient/session.rb:174:in `query'
    C:/projects/jekyll/vendor/bundle/ruby/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient.rb:1240:in `do_get_block'
    C:/projects/jekyll/vendor/bundle/ruby/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient.rb:1017:in `block in do_request'
    C:/projects/jekyll/vendor/bundle/ruby/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient.rb:1131:in `protect_keep_alive_disconnected'
    C:/projects/jekyll/vendor/bundle/ruby/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient.rb:1012:in `do_request'
    C:/projects/jekyll/vendor/bundle/ruby/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient.rb:1102:in `follow_redirect'
    C:/projects/jekyll/vendor/bundle/ruby/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient.rb:653:in `get_content'
    C:/projects/jekyll/test/test_commands_serve.rb:110:in `block (2 levels) in <class:TestCommandsServe>'
    C:/projects/jekyll/vendor/bundle/ruby/2.3.0/gems/shoulda-context-1.2.1/lib/shoulda/context/context.rb:413:in `instance_exec'
    C:/projects/jekyll/vendor/bundle/ruby/2.3.0/gems/shoulda-context-1.2.1/lib/shoulda/context/context.rb:413:in `block in create_test_from_should_hash'
```

awood:
@ashmaroli Does Appveyor have firewall rules or something that are keeping that port closed?  I am totally clueless when it comes to development under Windows so I think I will need some hand-holding here. :confused:

ashmaroli:
@XhmikosR will be able to answer your queries related to Appveyor better.

FeodorFitsner:
So, it can't connect to the localhost?

awood:
@FeodorFitsner That's what it looks like to me at least.

FeodorFitsner:
Connecting to localhost should be possible per se. Look at this AppVeyor website test, for example: https://ci.appveyor.com/project/AppVeyor-Website/website/build/1.0.614#L129

Try logging in via RDP and see what's going on - if there is process listening on that port is started and listening: https://www.appveyor.com/docs/how-to/rdp-to-build-worker/

awood:
@FeodorFitsner I fixed a concurrency issue I found in the tests where the LiveReload server was starting after the test so I believe that addresses the "No connection" error.  Looks like the AppVeyor errors now are

```
C:/Ruby21/lib/ruby/site_ruby/2.1.0/rubygems.rb:261: warning: previous definition of
find_spec_for_exe was here terminate called after throwing an instance of 'std::runtime_error' 
what():  Encryption not available on this event-machine
```

Which I'm guessing is related to the TLS support for the LiveReload connection, but Jekyll already supports TLS so I am a bit confused.  Any insight would be much appreciated!

FeodorFitsner:
@awood Don't have a clue what that could mean, sorry 😞 My Ruby skills are really basic.

awood:
@XhmikosR Looks like EventMachine is getting installed on those Windows machines without OpenSSL support.  See [here](https://github.com/eventmachine/eventmachine/wiki/Building-EventMachine-with-SSL-on-Windows).  What's the best way to get OpenSSL installed on those build nodes?  And then I have to figure the best way to pass compile flags through the gemspec to bundler.

XhmikosR:
@awood: You only need the DLLs? See https://www.appveyor.com/docs/installed-software/#tools

awood:
@XhmikosR I'm pretty much at the end of my rope here.  Eventmachine isn't being installed with native extensions when installed by bundler (em-websockets pulls in eventmachine).  Setting the `--with-ssl-dir` via `bundler config` doesn't seem to do anything ([Looks](https://github.com/eventmachine/eventmachine/wiki/Building-EventMachine-with-SSL-on-Windows) like I'm not the first to hit that issue).

I can install eventmachine with native extensions via gem, but if I install before bundler runs, eventmachine is installed into the GEM_HOME for the user and not the `vendor/bundle` GEM_HOME that bundler is using.  I'm guessing that some alteration is needed to the GEM_PATH or to temporarily set GEM_HOME to the `vendor/bundle` directory when gem installs eventmachine.

If I try to uninstall eventmachine and reinstall it after bundler runs, I get the enigmatic message that "Could not find eventmachine-1.2.0.1-x86-mingw32 in any of the sources".

I've run about 50 different builds with various attempts to link to OpenSSL on Windows but without any real luck.  Any assistance would be appreciated.  (This is not to mention that actual Windows users will probably wind up a very tedious situation where they have to download and install OpenSSL and then link eventmachine to it themselves if they want to have a functional Jekyll install).  How does Jekyll currently link to OpenSSL?  `jekyll serve` references `OpenSSL::X509::Certificate` which I thought was just a binding to OpenSSL.

If there are any other Windows folks on the team, I'd appreciate some insight.  As it stands, I'm about ready to throw in the towel.

See also https://github.com/eventmachine/eventmachine/issues/689

XhmikosR:
I can't help you anymore with this myself since I haven't checked out the
source changes. It's not a bad idea and I'm pretty sure there will be a
solution if you still want to have this merged.

awood:
@XhmikosR Yeah, I'd definitely like to get it merged.  Just not sure where to go with the Windows build errors.  Any experienced Ruby developer on Windows could probably shed some light on things.  Mostly I worry that, if this code hits master, Windows users will end up with a complicated installation since eventmachine is going to require OpenSSL but it seems like it's only a soft requirement.  Let me reach out to the eventmachine people and see if they have any ideas.

awood:
@XhmikosR Okay, so I'm going to give up on eventmachine+SSL for Windows.  I haven't figured anything out after talking briefly to the EventMachine people and if it's this complicated for me, then it's probably hopelessly complicated for just a user.

So I have an alternative thought: I could just drop SSL/TLS support for the eventmachine server responsible for the websockets stuff.  Running Jekyll serve over TLS and running with LiveReload would simply be mutually exclusive operations since fetching the LiveReload JS over HTTP would create a browser warning about insecure content.

I like providing people with the tools to use SSL/TLS if they want to, but the SSL/TLS usage is already a bit wonky because if you are using a self-signed cert, you have to have an exception for both port 4000 and for the LiveReload port.

What do you think?

ashmaroli:
in case you didnt know it already, [`jekyll-admin`](https://github.com/jekyll/jekyll-admin) plugin has a feature akin to LiveReload, built-in, and it works great on Windows.
Perhaps, you could write a similar Ruby+NodeJS plugin for Jekyll core..?

awood:
@DirtyF You can remove the help-wanted label.  I've elected to drop TLS support for LiveReload altogether.  It's just too problematic in the Windows environment.

awood:
@envygeeks any idea what 

```
W, [2016-11-04T17:46:58.154891 #5275]  WARN -- :       This usage of the Code Climate Test Reporter is now deprecated. Since version

      1.0, we now require you to run `SimpleCov` in your test/spec helper, and then
```

means in the CI failures?  I'm not seeing that locally even with a new bundle install including codeclimate-test-reporter 1.0.0.  Not having the test dependencies somewhat locked down is pretty frustrating.  I've had this same issue with new RuboCop versions being released and suddenly code unrelated to my PR start getting flagged.  It makes the CI environment feel like spinning a roulette wheel.

envygeeks:
@awood if the tests don't fail you can safely ignore that, we do include SimpleCov along with their test reporter afaiw, but that's just our coverage reporter letting us know that something has changed.  That shouldn't cause the test to fail through if the tests are failing it's probably elsewhere.

awood:
> That shouldn't cause the test to fail through if the tests are failing it's probably elsewhere.

The test script aborts almost immediately.  Here's the snippet from Travis.

```
+ruby -S bundle exec rake TESTOPTS=--profile test
/home/travis/.rvm/rubies/ruby-2.3.1/bin/ruby -w -I"lib:lib:test" -I"/home/travis/build/jekyll/jekyll/vendor/bundle/ruby/2.3.0/gems/rake-11.3.0/lib" "/home/travis/build/jekyll/jekyll/vendor/bundle/ruby/2.3.0/gems/rake-11.3.0/lib/rake/rake_test_loader.rb" "test/**/test_*.rb" --profile
# -------------------------------------------------------------
# SPECS AND TESTS ARE RUNNING WITH WARNINGS OFF.
# SEE: https://github.com/Shopify/liquid/issues/730
# SEE: https://github.com/jekyll/jekyll/issues/4719
# -------------------------------------------------------------
W, [2016-11-04T17:46:44.805938 #5249]  WARN -- :       This usage of the Code Climate Test Reporter is now deprecated. Since version

      1.0, we now require you to run `SimpleCov` in your test/spec helper, and then
      run the provided `codeclimate-ruby` binary separately to report your results
      to Code Climate.
      More information here: https://github.com/codeclimate/ruby-test-reporter/blob/master/README.md
```

envygeeks:
I have a meeting in a couple of minutes, I'll fix that error up for ya after the meeting.

ashmaroli:
@awood, the codeclimate-test-reporter gem issue has been temporarily resolved on the `master` branch. You can either `rebase` to `upstream/master` and force-push to this branch, to trigger a new build or wait to have one of the maintainers restart the current build..

awood:
@ashmaroli @envygeeks  Thanks for all your help on this.  Looks like all the checks are passing now!  I think this is ready for final review.  My branch is in kind of a mess so once I get a pass on the review I can tidy up the commit history if necessary.

DirtyF:
@awood Thanks for your continuous effort on this, @jekyll/core should review this.

If @ashmaroli or @XhmikosR could test it under Windows, it would be much appreciated.

ashmaroli:
I did a couple of manual testing on **Windows** using **a newly created Jekyll gem that incorporated this feature**.

Ruby: **ruby 2.3.1p112 (2016-04-26 revision 54768) [i386-mingw32]**
Browser: Google Chrome

Below is a representation of what I expected and what I got: 
( sample site served with `bundle exec jekyll s --livereload` )

Status | Description
------ | --------------
:x:  | Log output at info if LiveReload has been enabled
:x:  | Log output at debug if LiveReload has been enabled ( `--verbose` added )
:white_check_mark: | warn when additional `--no-watch` switch is used
:white_check_mark:  | `<script></script>` injected into sample webpage
:white_check_mark:  | only `livereload.js` requested
:x:  | Reload webpage **automatically** when a `post` content has been changed.
:x:  | Reload webpage **automatically** when a `page` content has been changed.

--
/cc @jekyll/windows to replicate the results above in a different Windows env
@pathawks or @DirtyF to test similar steps on non-Windows

awood:
@ashmaroli Darn, that's disappointing that it's not working for you.  I'll have to set up a Windows dev environment to investigate, but would you mind telling me which browser you're using so I can be on the same page with you.

awood:
@ashmaroli I finally got a windows setup going and didn't have a problem getting pages to refresh in Firefox (didn't check posts, but you were seeing failures for both pages and posts).  Would you mind testing again and emailing me your console output if you are still seeing failures.

I'm wondering if you were running into a situation where your browser had not yet connected to the LiveReload server when you made your changes.  LiveReload won't really do anything until the browser connects.  You'll see the console message "LiveReload: Browser connected" once the browser connects.  You can also add a "?LR-verbose=true" onto the URL and LiveReload will log some messages to Firefox's Web Developer Console (or the Chrome equivalent).

The other thing I can think of is maybe this is a Chrome issue instead of a Windows issue.  (I forgot to test in Chrome on my Windows machine at home 😦 ).  But if you can get me your console output and/or browser console output, that would help me out.

ashmaroli:
@awood If you were able to have your browser refresh automatically, then its probably Chrome's issue.. or my network issue..
the following is what Chrome tells me:

lr

awood:
@ashmaroli Thanks;   That screenshot is helpful.  I'll take a look with Chrome when I get home.

awood:
Okay, I can confirm the warning @ashmaroli is seeing is due to Chrome thinking that 127.0.0.1 is different from localhost and then flagging the document.write as being cross-origin.  That's probably not too difficult to fix.  Maybe figure out the hostname in javascript and drop it in instead of using 127.0.0.1?  (The error goes away BTW, if you point your browser to 127.0.0.1:4000)

The content length mismatch is more of a mystery to me.  @ashmaroli if you retry your tests using 127.0.0.1 as the hostname do you get any error messages from Jekyll?  I'm seeing this when the message to refresh is sent to the browser

```
[2016-12-14 22:01:23] ERROR Errno::ECONNABORTED: An established connection was a
borted by the software in your host machine. @ io_fillbuf - fd:10
        C:/Ruby22-x64/lib/ruby/2.2.0/webrick/httpserver.rb:80:in `eof?'
        C:/Ruby22-x64/lib/ruby/2.2.0/webrick/httpserver.rb:80:in `run'
        C:/Ruby22-x64/lib/ruby/2.2.0/webrick/server.rb:294:in `block in start_thread'
```

but it doesn't seem to actually adversely affect the LiveReload functionality.

ashmaroli:
> if you retry your tests using 127.0.0.1 as the hostname do you get any error messages from Jekyll?

@awood the cross-origin error resolved on pointing to `http://127.0.0.1:4000/` but still doesn't reload the browser automatically..
When I check the `Network` tab in Chrome console, I see that livereload.js has failed `GET` due to `ERR_CONTENT_LENGTH_MISMATCH`

ashmaroli:
A google search on this issue returned many stackoverflow posts that seem to ask `have you checked that the server is sending out a correct content length in the content-length header`

awood:
@ashmaroli Mind testing one more time?  I made a small correction to how the content-length header is calculated.  (I've not been able to reproduce the issue on my Windows machine, otherwise I'd do it.)

ashmaroli:
@awood I don't mind testing on my machine. Its just that since noone else is able to support my claims, I'm not sure if I the error is with the code or my system..

On using the latest code, `livereload.js` doesn't seem to be downloaded at all.. Viewing the source-code of the generated pages have no trace of the following:
```
    <script>
      document.write(
        '<script src="http://' +
        (location.host || 'localhost').split(':')[0] +
        ':<%=@options["livereload_port"] %>/livereload.js?snipver=1<%= livereload_args %>"' +
        '></' +
        'script>');
    </script>
```

awood:
@jekyll/core How's this PR doing besides the Windows issue @ashmaroli is encountering?  I'd like to get some feedback on the Linux side of things and maybe some more testers for Windows since I haven't been able to reproduce the Windows issue.

RobertDeRose:
@awood I wish I would have noticed your work on this back in July.  I was hoping we could just improve Jekyll's plugin capabilities to allow plugins to hook into core more easily without the need for Monkey patching, this would allow things like Livereloading to stay a plugin.  But heck, if this gets merged, great work!  It looks like a ton of effort!

awood:
@parkr  How does this look to you?  I'd like to at least get the Linux/OSX sign off before I devote additional time to fixing what appeared to me to be a Windows issue unique to one tester.

@RobertDeRose No worries.  I agree that it probably would be best if we could just bundle this as an additional "official" Jekyll plugin like "jekyll-sitemap".  But I really want it rolled into the `serve` command as an option instead of as a separate command.

DirtyF:
@awood Sorry for not getting back to you sooner, I really wish we come to a decision on this.

Some conflicts appeared in the meantime, if you could fix those, that would make the review easier. Thanks for your time and patience. I'm currently a user of the `jekyll-livereload` plugin and it would greatly fit into officials plugins.

/cc @jekyll/plugin-core @jekyll/ecosystem

ashmaroli:
@awood Great News! I revisited this feature on the same machine I had tested earlier.. the difference being  that I used a fresh clone of this PR..
I decided to test this out on the `docs` folder of the repo.. and to my surprise( even though, my system has remained the same.. and Jekyll version on this branch is at `3.4.0`) , my browser tab automatically reloaded with the changes I made in my text-editor.. Guess, new dependencies were released in the interim..

**So whatever I said earlier is now moot..!!** :tada:

lib/jekyll/commands/serve.rb
```diff
@@ -51,6 +107,70 @@ def process(opts)
...
+        def shutdown
+          @server.shutdown if @running
+        end
+
+        private
+        def validate_options(opts)
+          if opts["livereload"]
+            if opts["detach"]
+              Jekyll.logger.abort_with "Error:",
+                "--detach and --livereload are mutually exclusive"
```
mattr-: I'd like to go out on a limb here and have the `--livereload` option take precedence over the `--detach` option so that if both are specified together, we simply don't detach.

lib/jekyll/commands/serve.rb
```diff
@@ -51,6 +107,70 @@ def process(opts)
...
+            end
+            if opts["ssl_cert"] || opts["ssl_key"]
+              # This is not technically true.  LiveReload works fine over SSL, but
+              # EventMachine's SSL support in Windows requires building the gem's
+              # native extensions against OpenSSL and that proved to be a process
+              # so tedious that expecting users to do it is a non-starter.
+              Jekyll.logger.abort_with "Error:", "LiveReload does not support SSL"
+            end
+            unless opts["watch"]
+              Jekyll.logger.warn "Using --livereload without --watch defeats the purpose"\
```
mattr-: I think it makes sense for `--livereload` to also implicitly add `--watch` to the command line options and both do the watching and the livereloading. It feels like not being pedantic about the options here is a better experience for the user.

lib/jekyll/commands/serve/servlet.rb
```diff
@@ -3,6 +3,128 @@
 module Jekyll
   module Commands
     class Serve
+      # This class is used to determine if the Servlet should modify a served file
+      # to insert the LiveReload script tags
+      class SkipAnalyzer
+        BAD_USER_AGENTS = [%r!MSIE!].freeze
+
+        def self.skip_processing?(req, res, options)
```
mattr-: I'd appreciate seeing some more expressive variable names here. I can parse that `req` means `request` and `res` means `response` but I think it's better if we just use those variable names rather than having people who come along later reading the code to parse them too.

lib/jekyll/commands/serve/websockets.rb
```diff
@@ -0,0 +1,220 @@
...
+              "",
+            ].join("\r\n")
+            send_data(headers)
+            send_data(body)
+            close_connection_after_writing
+          end
+        end
+      end
+
+      class LiveReloadReactor
```
mattr-: I'd prefer to see these two classes split out into their own separate files. I know Ruby doesn't care about it but it makes it easier for discoverability if it's a single class per file so that people can assume relatively safely that there's a `live_reload_reactor.rb` file to go with the `LiveReloadReactor` class.

lib/jekyll/commands/serve/websockets.rb
```diff
@@ -0,0 +1,220 @@
...
+        # http://feedback.livereload.com/knowledgebase/articles/86174-livereload-protocol
+        def reload(pages)
+          pages.each do |p|
+            msg = {
+              :command => "reload",
+              :path    => p.url,
+              :liveCSS => true,
+            }
+
+            Jekyll.logger.debug("LiveReload:", "Reloading #{p.url}")
```
mattr-: What do you think about making this an info message vs. a debug message?

awood:
@ashmaroli Wonderful.  This PR is pretty messy, so I'm working on starting over on master and then reimplimenting my changes so that the changes are more succinct.

@mattr Thanks for the review.  I'll try to incorporate your suggestions as I rebase/rewrite.

awood:
@mattr- I believe I have addressed your requests with one exception.  Personally speaking, I think setting the `Jekyll.logger.debug("LiveReload:", "Reloading #{p.url}")` message to `info` would be undesirable because when Jekyll regenerates a site, that loop runs over a long list of pages and I wouldn't want to spam the console.  If you feel differently, I'll willing to change it to `info`.

awood:
The appveyor failures seem possibly related to https://github.com/eventmachine/eventmachine/issues/806 but I'll have to see if `gem install eventmachine --platform ruby` fixes the problem in Windows with Ruby 2.4.

But the code itself in this PR shouldn't need to change since it's a dependency issue.

lib/jekyll/commands/serve.rb
```diff
@@ -108,8 +108,9 @@ def shutdown
         def validate_options(opts)
           if opts["livereload"]
             if opts["detach"]
-              Jekyll.logger.abort_with "Error:",
-                "--detach and --livereload are mutually exclusive"
+              Jekyll.logger.warn "Warning:",
+                "--detach and --livereload are mutually exclusive. Choosing --livereload"
```
mattr-: :heart::heart::heart:

lib/jekyll/commands/serve.rb
```diff
@@ -1,24 +1,55 @@
 # frozen_string_literal: true
 
+require "thread"
+
 module Jekyll
   module Commands
     class Serve < Command
+      # Similar to the pattern in Utils::ThreadEvent except we are maintaining the
+      # state of @running instead of just signaling and event.  We have to maintain this
```
mattr-: Typo in `signaling and event`?

test/test_commands_serve.rb
```diff
@@ -12,6 +15,125 @@ def custom_opts(what)
...
+      @client.connect_timeout = 5
+      @standard_options = {
+        "port"            => 4000,
+        "host"            => "localhost",
+        "baseurl"         => "",
+        "detach"          => false,
+        "livereload"      => true,
+        "source"          => @temp_dir,
+        "destination"     => @destination,
+        "livereload_port" => Jekyll::Commands::Serve.singleton_class::LIVERELOAD_PORT,
```
mattr-: If we have a default built in already, is there a reason to specify it here in the test?

lib/jekyll/commands/serve/live_reload_reactor.rb
```diff
@@ -0,0 +1,153 @@
...
+            end
+          end
+        end
+
+        private
+        def connect(ws, handshake)
+          @connections_count += 1
+          if @connections_count == 1
+            message = "Browser connected"
+            message += " over SSL/TLS" if handshake.secure?
```
Crunch09: Just curious: Could this ever be true now that we disabled support of SSL with LiveReload because of Windows issues? (Fine for me to keep this line in, in any case, to be future proof)

lib/jekyll/commands/serve.rb
```diff
@@ -1,24 +1,55 @@
...
+          "livereload_min_delay" => ["--livereload-min-delay [SECONDS]",
+            "Minimum reload delay",],
+          "livereload_max_delay" => ["--livereload-max-delay [SECONDS]",
+            "Maximum reload delay",],
+          "livereload_port"      => ["--livereload-port [PORT]", Integer,
+            "Port for LiveReload to listen on",],
         }.freeze
 
+        LIVERELOAD_PORT = 35_729
+        LIVERELOAD_DIR = File.join(File.dirname(__FILE__), "serve", "livereload_assets")
```
ashmaroli: Jekyll has started using `__dir__` instead of `__FILE__` throughout the codebase now..

lib/jekyll/commands/serve.rb
```diff
@@ -1,24 +1,55 @@
...
+
+        attr_reader :mutex, :run_cond
         #
 
+        def running?
+          @running
+        end
```
ashmaroli: I think this can be simplified using a combo of `attr_accessor :running` and `alias` / `alias_method`..

lib/jekyll/commands/serve/live_reload_reactor.rb
awood: Yeah, I was keeping it around in case SSL/TLS support gets better with EventMachine and Windows.

awood:
> 👍 Amazing work! 💯 👏 I don't feel comfortable enough with EventMachine and LiveReload, that's the only reason for not-approving.
> Would be great if we could improve our hooks so similar things can be built as plugins without the need of monkey patching.

I completely understand if the team decides not to accept this patch since it adds a lot of new stuff: multi-threading, Javascript, WebSockets.  And all three of those new things can be extremely vexatious.

If the team doesn't want to accept this, then my request would be to modify Jekyll in such a way that plugins can actually replace a core command (`serve`, `build`, etc) (and switch the core command methods to `protected`).  That way I could continue to maintain [Hawkins](https://github.com/awood/hawkins) but I would be able to properly replace the stock `serve` command with my own that extends the existing command.  I am not a fan of monkey patching so today Hawkins adds a new command `liveserve` which is a little hokey.

Also a way to unregister plugins would be helpful for testing.

DirtyF:
@awood I just tested it with Jekyll's website and it works perfectly if I run `bundle exec jekyll serve --livereload` from the `docs` folder. What should we require in our [`site.rake`](https://github.com/jekyll/jekyll/blob/4e561a84e4db0035e988593fbeff8e7a49c0d333/rake/site.rake#L13-L35) task to be able to use livereload option?

ashmaroli:
@awood instead of me adding more review comments to this PR here, are you fine with having to review my pull request to your branch that refactors certain portions of this PR?

mattr-:
@ashmaroli Id like to merge this PR after running it through some
acceptance tests later. Can your changes wait until it lands in master?

ashmaroli:
> Can your changes wait until it lands in `master`?

Yes, it can.. I'm just shooting to DRY things up a bit..
FWIW, I too was able to **successfully** test this **branch against the Jekyll Docs** like Frank did ( except that I did so on **Windows** with **Ruby 2.3** )

awood:
> @awood instead of me adding more review comments to this PR here, are you fine with having to review my pull request to your branch that refactors certain portions of this PR?

Yes, that would be wonderful.

ashmaroli:
> Yes, that would be wonderful.

I've decided to postpone that in order to avoid a re-review of this PR.. Rest assured, I'll mention you when I start the PR here on Jekyll.
Additionally, there's an active proposal to enable some disabled Rubocop checks.. so depending on what gets merged first, most of what I'm thinking about here is gonna be cleared in due course.. :smiley:

lib/jekyll/commands/serve/livereload_assets/livereload.js
```diff
@@ -0,0 +1,1183 @@
+(function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
```
pathawks: 😨  What is this?

Can we have a non-minified version of this somewhere? Or at least a comment with a link to where the original source can be found?

DirtyF: FWIW you should be able to format it in you browser developer tools.

pathawks: I just mean that all the variable names have been replaced with single letters. Even formatted, this is not readable code.

ashmaroli: We can de-obfuscate in-browser? I did not know that..!!

DirtyF: https://github.com/livereload/livereload-js/blob/master/dist/livereload.js

awood: Yes, that's the source of it.  Depending on how y'all like to do things, it might make sense to pull that in via whatever the JS dependency tool of the month is.

pathawks:
How often is the JavaScript library updated? Does it require updates more frequently than Jekyll updates?

I feel a bit uncomfortable including a big JavaScript file in our repo that I have no way of auditing.

Does this need to be in Jekyll Core, or would it be better as a plugin? Releasing this as a plugin would allow the JavaScript dependencies to be updated as often as possible, without waiting on a new Jekyll release.

Alternatively, maybe this could be packaged with the **jekyll-watch** gem. This feature doesn't really make sense without **jekyll-watch**, right?

ashmaroli:
> How often is the JavaScript library updated?

From the look of the library's [Releases](https://github.com/livereload/livereload-js/releases) page, I'd say, *"not that frquently.."*. The library doesn't seem to be in active development either. The last commit there was on `Oct 1, 2015`

> would it be better as a plugin?

The community has already developed plugins along similar lines:
 - Alex Wood's own [`Hawkins`](https://github.com/awood/hawkins)
 - [`jekyll-livereload`](https://github.com/RobertDeRose/jekyll-livereload)

DirtyF:
@pathawks As long as `serve` command will be part of core, such `--livereload` option makes sense. 

This is the default in most today building tools, e.g. Hugo has a [fast-render option](http://gohugo.io/news/0.30-relnotes/) activated by default now.

The question is: is this the right way to do it? Could we avoid relying on that Javascript? How should be trigger the brower to refresh on changes?

At least it's a first step and it's working. most users would appreciated to benefit from this feature given it plays well with `--incremental` and al. 

Just my 2 cents.

pathawks:
> As long as serve command will be part of core, such --livereload option makes sense.

:+1:

awood:
> I feel a bit uncomfortable including a big JavaScript file in our repo that I have no way of auditing.

I definitely appreciate this feeling.  The [Nikola](https://github.com/getnikola/nikola/blob/948df1e2928777557a1862cd33ac4bf22f49749e/bower_components/livereload-js/.bower.json) static site generator includes the file wholesale as well, but using Bower it looks like.  A looser method of inclusion (as opposed to just dropping the JS file right in the repo) might be worth looking at?

DirtyF:
It seems like [Hugo also embed this JavaScript](https://github.com/gohugoio/hugo/blob/master/livereload/livereload.go#L105) for now.

mattr-:
Including the livereload.js file in Jekyll seems bad but we don't have another way to go about having those sorts of dependencies in a nicer way today. I feel like it's an acceptable risk to take though, as you have to tell Jekyll to activate livereload explicitly instead of it being on all the time.

Once this gets merged, should we should follow up with a PR to the maintainer docs in docs/_docs/maintaining about external code bases to keep an eye on?

ashmaroli:
Are we going to wait on EventMachine to release binary support for Ruby 2.4 on Windows or just merge this right away?

DirtyF:
Related: A new plugin is out https://github.com/anomaly/jekyll-reload

ashmaroli:
> A new plugin is out

It doesn't seem to tested on Ruby 2.4 on Windows, which is what is blocking this..

DirtyF:
As far as I know, Jekyll never officially supported Windows.

ashmaroli:
> Jekyll never officially supported Windows.

Of course.. It still doesn't..
So is this going to be included in `3.7.0`? I'm in favor of including it..

DirtyF:
@awood Could you please fix the remaining conflicts? Sorry for not merging this sooner, it got approved and should ship in next minor version.

awood:
@DirtyF Conflicts resolved

parkr:
Can you skip the specs that use LiveReload on Windows if they won't work?

awood:
> Can you skip the specs that use LiveReload on Windows if they won't work?

Done!

DirtyF:
💯 A big thanks @awood for all the work you put into this, I hope this really improve the developer experience for a majority of Jekyll users.

@jekyllbot: merge +minor

DirtyF:
@awood 

If I addd the `--livereload` option on the command line this works fine.
If I add the livereload: true option to [our rake file](https://github.com/jekyll/jekyll/blob/master/rake/site.rake#L27-L34) then the following message appears:

```sh
rake aborted!
NoMethodError: undefined method `start' for nil:NilClass
/Users/frank/code/jekyll/jekyll/lib/jekyll/commands/serve.rb:224:in `start_up_webrick'
/Users/frank/code/jekyll/jekyll/lib/jekyll/commands/serve.rb:104:in `process'
/Users/frank/code/jekyll/jekyll/rake/site.rake:37:in `block (2 levels) in <top (required)>'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.3.0/lib/rake/task.rb:251:in `block in execute'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.3.0/lib/rake/task.rb:251:in `each'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.3.0/lib/rake/task.rb:251:in `execute'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.3.0/lib/rake/task.rb:195:in `block in invoke_with_call_chain'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/2.4.0/monitor.rb:214:in `mon_synchronize'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.3.0/lib/rake/task.rb:188:in `invoke_with_call_chain'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.3.0/lib/rake/task.rb:181:in `invoke'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.3.0/lib/rake/application.rb:160:in `invoke_task'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.3.0/lib/rake/application.rb:116:in `block (2 levels) in top_level'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.3.0/lib/rake/application.rb:116:in `each'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.3.0/lib/rake/application.rb:116:in `block in top_level'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.3.0/lib/rake/application.rb:125:in `run_with_threads'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.3.0/lib/rake/application.rb:110:in `top_level'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.3.0/lib/rake/application.rb:83:in `block in run'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.3.0/lib/rake/application.rb:186:in `standard_exception_handling'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.3.0/lib/rake/application.rb:80:in `run'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.3.0/exe/rake:27:in `<top (required)>'
/Users/frank/.rbenv/versions/2.4.2/bin/rake:23:in `load'
/Users/frank/.rbenv/versions/2.4.2/bin/rake:23:in `<top (required)>'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/site_ruby/2.4.0/bundler/cli/exec.rb:75:in `load'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/site_ruby/2.4.0/bundler/cli/exec.rb:75:in `kernel_load'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/site_ruby/2.4.0/bundler/cli/exec.rb:28:in `run'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/site_ruby/2.4.0/bundler/cli.rb:424:in `exec'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/site_ruby/2.4.0/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/site_ruby/2.4.0/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/site_ruby/2.4.0/bundler/vendor/thor/lib/thor.rb:387:in `dispatch'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/site_ruby/2.4.0/bundler/cli.rb:27:in `dispatch'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/site_ruby/2.4.0/bundler/vendor/thor/lib/thor/base.rb:466:in `start'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/site_ruby/2.4.0/bundler/cli.rb:18:in `start'
/Users/frank/.rbenv/versions/2.4.2/bin/bundle:30:in `block in <main>'
/Users/frank/.rbenv/versions/2.4.2/lib/ruby/site_ruby/2.4.0/bundler/friendly_errors.rb:122:in `with_friendly_errors'
/Users/frank/.rbenv/versions/2.4.2/bin/bundle:22:in `<main>'
Tasks: TOP => site:preview
```

burrich:
Hi, can you confirm that livereload is not supported on Windows please ?

ashmaroli:
@burrich `--livereload` is definitely supported on Windows.. just not with `--ssl_cert` and `--ssl_key` as well..

burrich:
Ok thanks @ashmaroli , i will try.

awood:
@burrich If I recall correctly, it won't work with Ruby 2.4 on Windows due to a dependency (EventMachine) not working.

ashmaroli:
> won't work with Ruby 2.4 on Windows due to a dependency (EventMachine) not working.

@burrich That issue is being tracked here :point_right: https://github.com/eventmachine/eventmachine/issues/806

lib/jekyll/commands/serve.rb
```diff
@@ -63,6 +104,76 @@ def process(opts)
...
+              Jekyll.logger.warn "Warning:",
+                "--detach and --livereload are mutually exclusive. Choosing --livereload"
+              opts["detach"] = false
+            end
+            if opts["ssl_cert"] || opts["ssl_key"]
+              # This is not technically true.  LiveReload works fine over SSL, but
+              # EventMachine's SSL support in Windows requires building the gem's
+              # native extensions against OpenSSL and that proved to be a process
+              # so tedious that expecting users to do it is a non-starter.
+              Jekyll.logger.abort_with "Error:", "LiveReload does not support SSL"
```
RobertDeRose: I've stepped way back from this PR, never had the time to look at it was my plugin was just working for me.  However, Livereload can support SSL, I have support for it in my plugin https://github.com/RobertDeRose/jekyll-livereload/commit/34df507e4dadee09554ca06479de514899b9af39

DirtyF: Thanks @RobertDeRose, I was using your plugin until now :)
Are you interested in submitting a patch?

RobertDeRose: I might, if I can find the time.  So, that's a big maybe

awood: @RobertDeRose The problem wasn't with LiveReload not supporting SSL; it was with getting EventMachine and OpenSSL to cooperated on Windows.  I actually had it working fine in Linux, but getting the OpenSSL libraries to install and work on the Windows CI environment was a nightmare, so I just gave up 😭.  I figured most people would only be using LiveReload for rapid prototyping so I didn't consider SSL support to be a must-have.

RobertDeRose: @awood that's a fair point, I never did testing for Windows, and to be honest, not a lot of benefit for it when developing locally, unless you really need to ensure SSL isn't messing with thing, but at that point, you don't really need live reload.  OK ignore my extremely late to the party comments then.

Issue #5039, opened 2016-06-28
jekyllrb.com: Add custom 404 page with helpful info 


parkr:
If you visit a page on jekyllrb.com that does not exist, you get the generic GitHub Pages 404 page.

Instead of this, we should put together a custom page with helpful pointers for finding answers.

/cc @jekyll/documentation

adampetrie:
Im happy to take this on and I took an initial stab in the PR mentioned above. Its not much to look at but hey, it's as start.

What information do you think would be most valuable to include? Right now the main nav gets included but that's it.

benbalter:
Maybe a link to `/docs/` and `/help/`?

PR #5143, opened 2016-07-27
Initial 404 page

adampetrie:
This is a first go at addressing https://github.com/jekyll/jekyll/issues/5039. Not sure exactly what we're looking for here so I figured I'd open a PR and use the feedback to help this take shape.

### DIFF
--- a/site/404.html
+++ b/site/404.html
@@ -7,7 +7,7 @@ sitemap: false
 <section class="intro">
   <div class="grid">
     <div class="unit whole align-center">
-      <p class="first">404 - Not Found</p>
+      <p class="first">Huh. It seems that page is<br/>Hyde-ing...</p>
     </div>
   </div>
 </section>
--- a/site/404.html
+++ b/site/404.html
@@ -1,5 +1,5 @@
 ---
-layout: default
+layout: error
 permalink: /404.html
 sitemap: false
 ---
--- a/site/404.html
+++ b/site/404.html
@@ -7,7 +7,7 @@ sitemap: false
 <section class="intro">
   <div class="grid">
     <div class="unit whole center-on-mobiles">
-      <p class="first">404 - Page Not Found</p>
+      <p class="first">404 - Not Found</p>
     </div>
   </div>
 </section>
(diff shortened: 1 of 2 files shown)


site/404.html
```diff
@@ -0,0 +1,22 @@
+---
+layout: default
+permalink: /404.html
+---
+
+<section class="intro">
+  <div class="grid">
+    <div class="unit whole center-on-mobiles">
+      <p class="first">404 - Page Not Found</p>
```
envygeeks: This should be generic "not found" not everything is a page.

site/404.html
```diff
@@ -0,0 +1,22 @@
+---
+layout: default
+permalink: /404.html
+---
```
benbalter: You'll want to add `sitemap: false` to prevent it from being indexed.

adampetrie:
Here it is now:

screen shot 2016-07-27 at 7 09 58 pm

A bit more error-pagey IMO.

benbalter:
This is terrible, but instead of "404 - not found", what about, "Huh. It seems that page is Hyde-ing..."?

parkr:
@adampetrie *Much* better than before! I wonder if we should provide a short description of each page, e.g. `Community – Connect with other Jekyllites` etc? Giving a little insight into what's on the page would fill out the blank space here and provide a bit more help getting to the resource they are looking for.

site/404.html
```diff
@@ -0,0 +1,38 @@
...
+<section class="error">
+  <div class="grid">
+    <div class="unit whole center-on-mobile">
+      <p>The resource you requested was not found. Here are some links to help you find your way:</p>
+      <ul>
+        <li>
+          <a href="/">Home</a>
+        </li>
+        <li>
+          <a href="/docs/home/">Docs</a>
```
parkr: I'd call this `Documentation` – it's `Docs` in the header because it's too wide 😄

adampetrie:
I am pro puns so the 'hyde-ing' bit is awesome. There should probably be a 404 mention somewhere on the page though?

adampetrie:
How's this? I'm just spit-ballin' here...

screen shot 2016-07-27 at 7 51 27 pm

parkr:
LGTM. Let's ship it!

@jekyllbot: merge +site

PR #5152, opened 2016-07-28
Include theme directories in default gemspec

benbalter:
See https://github.com/jekyll/minima/pull/6.

We need to *include* the theme directories, not *exclude* them.

### DIFF
--- a/lib/theme_template/theme.gemspec.erb
+++ b/lib/theme_template/theme.gemspec.erb
@@ -10,7 +10,7 @@ Gem::Specification.new do |spec|
   spec.homepage      = "TODO: Put your gem's website or public repo URL here."
   spec.license       = "MIT"
 
-  spec.files         = `git ls-files -z`.split("\x0").select { |f| f.match(%r{^(<%= theme_directories.join("|") %>)/}) }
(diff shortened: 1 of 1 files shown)


benbalter:
As written, this PR will *only* include files in the theme directories. Do we want to include the LICENSE? I think the other files are safe to not include.

parkr:
Durrrrr. Thanks @benbalter!

> Do we want to include the LICENSE? I think the other files are safe to not include.

Yeah! If we're generating a license and readme, we should include them.

Thank you!

parkr:
LGTM.

benbalter:
@parkr Can I LGTM my own PR?

benbalter:
Also to note, we should update minima once this merges.

parkr:
> @parkr Can I LGTM my own PR?

Yep!

> Also to note, we should update minima once this merges.

To use this method? Sure!

LGTM.

benbalter:
LGTM.

parkr:
@jekyllbot: merge +bug

wommy:
https://github.com/jekyll/minima/commit/7629d352a05f809fe1cda3e10c62bbd4a495e173

wommy:
@benbalter @parkr thoughts on my above?

benbalter:
@wommy you're spot on your with commit comment. Mind opening a dedicated issue in this repo to figure out which approach is best to harmonize the three?

Issue #5153, opened 2016-07-28
Default site fails to build on Jekyll 3.2.x


benbalter:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I Updated to the latest Jekyll (or) if on Github Pages to the latest `github-pages`
- [x] I Read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [ ] This is a feature request.

---

- [x] I am on (or have tested on) ***macOS*** 10+
- [ ] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux
- [ ] I am on (or have tested on) ***Fedora*** GNU/Linux
- [ ] I am on (or have tested on) ***Arch*** GNU/Linux
- [ ] I am on (or have tested on) ***Other*** GNU/Linux
- [ ] I am on (or have tested on) ***Windows*** 10+

---

- [ ] I was trying to install.
- [ ] There is a broken Plugin API.
- [ ] I had an error on Github Pages, and I have not tested locally.
- [ ] I had an error on Github Pages, and Github Support said it was a Jekyll Bug.
- [ ] I had an error on Github Pages and I did not test it locally.
- [ ] I was trying to build.
- [x] It was another bug.

## My Reproduction Steps

```
➜  jekyll git:(master) git pull
Already up-to-date.
➜  jekyll git:(master) be jekyll new test-site
New jekyll site installed in /Users/benbalter/projects/jekyll/test-site.
➜  jekyll git:(master) ✗ cd test-site
➜  test-site git:(master) ✗ be jekyll serve --verbose
```

## The Output I Wanted

🎉 your site is built!

## The Output I got

```
Logging at level: debug
Configuration file: /Users/benbalter/projects/jekyll/test-site/_config.yml
         Requiring: kramdown
            Source: /Users/benbalter/projects/jekyll/test-site
       Destination: /Users/benbalter/projects/jekyll/test-site/_site
 Incremental build: disabled. Enable with --incremental
      Generating...
           Reading: _posts/2016-07-28-welcome-to-jekyll.markdown
         Rendering: _posts/2016-07-28-welcome-to-jekyll.markdown
  Pre-Render Hooks: _posts/2016-07-28-welcome-to-jekyll.markdown
  Rendering Liquid: _posts/2016-07-28-welcome-to-jekyll.markdown
         Requiring: rouge
  Rendering Markup: _posts/2016-07-28-welcome-to-jekyll.markdown
         Requiring: kramdown
[...]

PR #5154, opened 2016-07-28
Add test to build the default site

benbalter:
Motivated by https://github.com/jekyll/jekyll/issues/5153, and inspired by our tests on the Pages Gem, this PR adds a purposely (for now) failing test to generate and build Jekyll's default site.

The test is quick to run (2.3s, including Ruby's boot on my machine), and would provide a good smoke test / sanity check that the default Jekyll action works out of the box.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,12 @@ matrix:
       env: TEST_SUITE=fmt
     - rvm: jruby-9.0.5.0
       env: TEST_SUITE=test
+    - rvm: 2.3.0
+      env: TEST_SUITE=default-site
 env:
   matrix:
     - TEST_SUITE=test
     - TEST_SUITE=cucumber
-    - TEST_SUITE=default-site
 branches:
   only:
     - master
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -19,7 +19,7 @@ env:
   matrix:
     - TEST_SUITE=test
     - TEST_SUITE=cucumber
-
+    - TEST_SUITE=default-site
```
parkr: Does this need to run 3 times, once per Ruby version?

benbalter: Your call. I'm fine with it running once? If so, what's the best way to do that?

parkr: To add it only once, choose your fave Ruby version and add it to the `matrix.include` section above this one (see `fmt`):

```yaml
- rvm: 2.3.0
  env: TEST_SUITE=test-jekyll-new
```

Or whatever 😄

parkr:
Travis is doing weird things. :cool:

@jekyllbot: merge +dev

Issue #5145, opened 2016-07-28
Can't start local jekyll server


dlin303:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I Updated to the latest Jekyll (or) if on Github Pages to the latest `github-pages`
- [x] I Read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [ ] This is a feature request.

---

- [x] I am on (or have tested on) ***macOS*** 10+
- [ ] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux
- [ ] I am on (or have tested on) ***Fedora*** GNU/Linux
- [ ] I am on (or have tested on) ***Arch*** GNU/Linux
- [ ] I am on (or have tested on) ***Other*** GNU/Linux
- [ ] I am on (or have tested on) ***Windows*** 10+

---

- [ ] I was trying to install.
- [ ] There is a broken Plugin API.
- [ ] I had an error on Github Pages, and I have not tested locally.
- [ ] I had an error on Github Pages, and Github Support said it was a Jekyll Bug.
- [ ] I had an error on Github Pages and I did not test it locally.
- [x] I was trying to build.
- [ ] It was another bug.

## My Reproduction Steps
Did a fresh gem install of jekyll 3.2.0 and tried to serve it

```
$ gem install jekyll
$ jekyll new testblog
$ cd testblog
$ jekyll serve
```
Got this error:

```
Configuration file: /Users/Daron/Development/testblog/_config.yml
            Source: /Users/Daron/Development/testblog
       Destination: /Users/Daron/Development/testblog/_site
 Incremental build: disabled. Enable with --incremental
      Generating...
     Build Warning: Layout 'post' requested in _posts/2016-07-28-welcome-to-jekyll.markdown does not exist.
  Liquid Exception: no implicit conversion of nil into String in about.md
jekyll 3.2.0 | Error:  no implicit conversion of nil into String
```

## The Output I Wanted
But if i `gem uninstall jekyll` and reinstall v3.1.6 `gem install jekyll -v 3.1.6`, everything works fine.

```
$ rm -rf testblog
$ gem uninstall jekyll
$ gem install jekyll -v 3.1.6
$ jekyll new testblog
$ cd testblog
$ jekyll serve
```
I get what I expect:

```
Configuration file: /Users/Daron/Development/testblog/_config.yml
            Source: /Users/Daron/Development/testblog
       Destination: /Users/Daron/Development/testblog/_site
[...]

PR #5156, opened 2016-07-28
Fix for symlinked themes

benbalter:
If a user uses Rbenv, their theme is going to be symlinked. We run `File.realpath` on the path before we grab things like the includes folder, but not on the theme root, meaning when the two are passed to `Jekyll.sanitized_path`, Jekyll will think the unresolved root is the real root, and that the path (with resolved root already prepended) is an unsafe path outside the directory, thus prepending the unresolved theme path.

This PR resolves that by resolving the theme root path before passing things to sanitized path, ensuring the string comparison happens on both resolved paths, not one symlink and one resolved path.

Fixes https://github.com/jekyll/jekyll/issues/5145.

### DIFF
--- a/lib/jekyll/theme.rb
+++ b/lib/jekyll/theme.rb
@@ -34,11 +34,8 @@ module Jekyll
     private
 
     def path_for(folder)
-      resolved_dir = realpath_for(folder)
-      return unless resolved_dir
-
-      path = Jekyll.sanitized_path(root, resolved_dir)
-      path if File.directory?(path)
+      path = realpath_for(folder)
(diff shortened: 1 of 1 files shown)


parkr:
LGTM. Is there a test we could add for this? I'm fine either way,

benbalter:
I think https://github.com/jekyll/jekyll/pull/5154 should be the test for this... at least I can't get `jekyll new` to build using the default theme on Travis without this PR.

parkr:
:cool: Thanks!

@jekyllbot: merge +bug

PR #5157, opened 2016-07-28
Run hooks in priority order.

stevecheckoway:
Low priority hooks are being run before higher priority hooks. This is easy to
demonstrate with the following plugin:

    1.upto(10).each do |n|
      Jekyll::Hooks.register :site, :after_reset, priority: Jekyll::Hooks::PRIORITY_MAP[:low] do
        puts "Low #{n}"
      end
      Jekyll::Hooks.register :site, :after_reset, priority: Jekyll::Hooks::PRIORITY_MAP[:normal] do
        puts "Normal #{n}"
      end
      Jekyll::Hooks.register :site, :after_reset, priority: Jekyll::Hooks::PRIORITY_MAP[:high] do
        puts "High #{n}"
      end
    end

Sorting by the negative of the priority and then by the order the hook was
added does the right thing.

### DIFF
--- a/features/hooks.feature
+++ b/features/hooks.feature
@@ -235,7 +235,7 @@ Feature: Hooks
       owner.output = "1 #{owner.output.chomp}"
     end
     Jekyll::Hooks.register :pages, :post_render, priority: :high do |owner|
-      # high runs last
+      # high runs first
       owner.output = "2 #{owner.output.chomp}"
     end
     Jekyll::Hooks.register :pages, :post_render do |owner|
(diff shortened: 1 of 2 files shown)


stevecheckoway:
Hmm. Confusingly, the feature file indicates the _low_ priority hooks should run first. That's counter-intuitive and disagrees with plugin priority meanings.

Either the feature file should be fixed, or the `-` in the patch removed so that the hook priority is `[priority, @hook_priority.size]`.

Either way, comparing as a floating point number gives seriously broken results. From the plugin in the commit message, without the patch, I get

```
Low 1
Low 5
Low 6
Low 7
Low 8
Low 9
Low 10
Low 2
Low 3
Low 4
Normal 1
Normal 4
Normal 5
Normal 6
Normal 7
Normal 8
Normal 9
Normal 10
Normal 2
Normal 3
High 4
High 5
High 6
High 7
High 1
High 8
High 9
High 10
High 2
High 3
```

parkr:
Agreed - high priority plugins should be run first, if only to be consistent with the other plugin infrastructure which does high to low.

envygeeks:
plugin priority makes little sense for hooks and plugins, I don't even know why that is there.  Everybody can just delegate themselves as high or low, and most don't even matter.  For hooks this is even more so, hooks should have a name so that you can insert before or after a specific hook if you need to, but priorities make little sense because everybody will just be high or low.

Ultimately priorities that are not weighted by the user, have little meaning without named meaning.

stevecheckoway:
Can't users get the same behavior of named hooks simply by using distinct priorities? They're not constrained to `:low`, `:normal`, or `:high`.

parkr:
@jekyllbot: merge +minor

PR #5167, opened 2016-07-30
Improve tests for Jekyll::PluginManager

Crunch09:
Hi,
this tries to improve the tests for the `Jekyll::PluginManager` by doing two things:
 - migrate the existing tests to `should` syntax (those were the last in the `test` dir with `def test_*` syntax
 - add missing tests

When running `script/test test/test_plugin_manager.rb`:

**Before**
screen shot 2016-07-30 at 11 22 31

**After**
screen shot 2016-07-30 at 15 16 21

### DIFF
--- a/test/test_plugin_manager.rb
+++ b/test/test_plugin_manager.rb
@@ -21,7 +21,7 @@ class TestPluginManager < JekyllUnitTest
   context "JEKYLL_NO_BUNDLER_REQUIRE set to `true`" do
     should "not require from bundler" do
       with_env("JEKYLL_NO_BUNDLER_REQUIRE", "true") do
-        assert_equal false, Jekyll::PluginManager.require_from_bundler,
+        refute Jekyll::PluginManager.require_from_bundler,
                      "Gemfile plugins were required but shouldn't have been"
         assert ENV["JEKYLL_NO_BUNDLER_REQUIRE"]
       end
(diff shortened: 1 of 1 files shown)


test/test_plugin_manager.rb
```diff
@@ -8,29 +8,141 @@ def with_no_gemfile
...
 
-  def test_blocking_requiring_from_bundler
-    with_env("JEKYLL_NO_BUNDLER_REQUIRE", "true") do
-      assert_equal false, Jekyll::PluginManager.require_from_bundler,
-                   "Gemfile plugins were required but shouldn't have been"
-      assert ENV["JEKYLL_NO_BUNDLER_REQUIRE"]
+  context "JEKYLL_NO_BUNDLER_REQUIRE set to `true`" do
+    should "not require from bundler" do
+      with_env("JEKYLL_NO_BUNDLER_REQUIRE", "true") do
+        assert_equal false, Jekyll::PluginManager.require_from_bundler,
```
parkr: I believe you can use `refute` here, which is the same as asserting false.

test/test_plugin_manager.rb
```diff
@@ -8,29 +8,141 @@ def with_no_gemfile
...
   end
+
+  context "require gems" do
+    should "check if configured gems are allowed" do
+      allow(Jekyll::External).to receive(:require_with_graceful_fail).and_return(nil)
+      site = double(:gems => %w(jemoji foobar))
+      plugin_manager = PluginManager.new(site)
+
+      expect(plugin_manager).to receive(:plugin_allowed?).with("jemoji")
+      expect(plugin_manager).to receive(:plugin_allowed?).with("foobar")
```
parkr: Why does this assert that the plugins were actually required?

parkr:
This is wonderful! Thank you! Two general comments above, otherwise :+1:

test/test_plugin_manager.rb
Crunch09: Good point! I'll update this test to proof they are required instead of allowed.

Crunch09:
@parkr Thanks for the review! I tried to address your comments in the added commits

parkr:
LGTM. Thank you! 👍 

@jekyllbot: merge +dev

PR #5168, opened 2016-07-30
docs/quickstart: update instructions to require bundler

parkr:
With gem-based themes being bundled in the new site via the `Gemfile`, we
should ask folks to use Bundler wherever possible. This should lead to more
successful installations and getting the base site setup properly.

The only trouble this introduces is it puts a dependency on Bundler. That
said, I'm totally fine with requiring everyone use Bundler for this site.
How could we best install bundler in these instructions?

/cc @jekyll/documentation

### DIFF
--- a/site/_docs/quickstart.md
+++ b/site/_docs/quickstart.md
@@ -10,7 +10,8 @@ For the impatient, here's how to get a boilerplate Jekyll site up and running.
 ~ $ gem install jekyll
 ~ $ jekyll new myblog
 ~ $ cd myblog
-~/myblog $ jekyll serve
+~/myblog $ bundle install
+~/myblog $ bundle exec jekyll serve
 # => Now browse to http://localhost:4000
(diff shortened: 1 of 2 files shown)


chrisfinazzo:
This was discussed a while back (ultimately put on hold) in #4368. For documentation, that's a starting point to build on. As far as the internals go, I recall you mentioned potentially having the ability to install specific versions similar to how the prereleases [work][].

With a stable v3 now available, is this a viable strategy?

[work]: https://jekyllrb.com/docs/installation/#pre-releases

parkr:
@jekyllbot: merge +site

parkr:
@chrisfinazzo We could definitely encourage people to use Gemfiles from now on. It's the best way to get things working. For prereleases, we could suggest updating their gemfile to say:

```ruby
gem "jekyll", "<THE-VERSION>"
```

etc etc etc

PR #5176, opened 2016-08-01
Rename .markdown to .md

ashmaroli:
This PR is to adopt `.md` as the extension for all Markdown files except History.markdown, .markdown files in /test folder and /features folder

- [x] Rename all existing .markdown files ( except History.markdown ) as .md files
- [x] Replace instances of `.markdown` within files to `.md`
- [x] Green Build with Travis CI


parkr:
... Oh wow, you didn't make it clear that you wanted to change the tests, too. I thought you were just talking about the documentation! I'm find moving the documentation, but let's keep the tests as-is for now.

ashmaroli:
Oh well, my bad.. I should've been clearer. I'll update the PR.

test/source/win/_posts/2009-05-24-yaml-linebreak.markdown
```diff
@@ -1,7 +1,7 @@
...
-tag: "Ruby"
----
-
+---
+layout: post
+title: "Test title"
+tag: "Ruby"
+---
+
 This is the content
```
ashmaroli: wonder why this is present as a diff... :confused:

features/step_definitions.rb
```diff
@@ -73,7 +73,7 @@
 Given(%r!^I have the following (draft|page|post)s?(?: (in|under) "([^"]+)")?:$!) do |status, direction, folder, table|
   table.hashes.each do |input_hash|
     title = slug(input_hash["title"])
-    ext = input_hash["type"] || "markdown"
+    ext = input_hash["type"] || "md"
     filename = "#{title}.#{ext}" if %w(draft page).include?(status)
```
ashmaroli: Am i wrong in changing this? :confused:

envygeeks: Yes, features is a business spec.  They are tests.

test/test_new_command.rb
```diff
@@ -54,7 +54,7 @@ def site_template
       capture_stdout { Jekyll::Commands::New.process(@args) }
 
       new_site_files = dir_contents(@full_path).reject do |f|
-        File.extname(f) == ".markdown"
+        File.extname(f) == ".md"
       end
```
ashmaroli: included this coz, this test runs against the template in /lib

test/source/win/_posts/2009-05-24-yaml-linebreak.markdown
XhmikosR: Line endings maybe?

ashmaroli: dunno.. I copied the file from master branch and overwrote the file on this branch.. still no success..
perhaps if I rebase to upstream........

ashmaroli:
Any idea why the Gemfile is not being included as part of `static_template_files` in test env `TEST_SUITE=test` (assuming thats whats causing the tests to fail..)?

ashmaroli:
-- To be considered for merge only after merge of PR-5203 --

benbalter:
Noticed that in the template site we had `.md` for pages and `.markdown` for the post and was going to open an issue suggesting the same. Nice work!

ashmaroli:
@benbalter Thanks :smiley:

ashmaroli:
@parkr Hi, your new release-post-draft reminded me of this PR.. 
Is now a good time to consider this PR? It concerns with:
- renaming `.markdown` files to `.md` files, 
- modifying `jekyll new` command template to have new posts generated with a `.md` extn
- modifying `test_new_command.rb` to accommodate this
- modifying the `Rakefile` to parse `.md` files for _jekyllrb_

parkr:
This is an omnibus PR – it does a lot of things that are only marginally related to each other. We shouldn't be updating the file extensions of documentation in the same PR where we make updates to the `jekyll new` site template, for example. I think `.markdown` is a better extension than `.md` for newcomers ("what the heck is md?!" vs. "ok, i see that this file is markdown because it ends in .markdown"), so I'm going to leave things as-is. Thank you!

ashmaroli:
> We shouldn't be updating the file extensions of documentation in the same PR where we make updates to the jekyll new site template. I think `.markdown` is a better extension than `.md` for newcomers

@parkr FWIW, this seems like a contradictory situation to me.. 
`jekyll new` in v3.3 gives `index.md` and `about.md` pages at root so users are going to realize what `.md` files are anyways.. but theirs posts in `_posts` have a`.markdown` extension. Now, wouldn't they be confused, as to why the double standards?

envygeeks:
We seem to be confusing double standard with inconsistency here.

ashmaroli:
> We seem to be confusing double standard with inconsistency here.

:grin: we need to resolve that nevertheless, no?

sskylar:
FWIW `.markdown` is the extension endorsed by the creator of Markdown: http://daringfireball.net/linked/2014/01/08/markdown-extension

Would love to see Jekyll adopt one (or the other) as part of a project style guide to avoid inconsistencies as noted above.

envygeeks:
I prefer `.md` but if I'm honest, I don't care either way, but I do care that we are consistent, so one or the other should be removed from our own stuff, not that we need to stop supporting one or the other, just make one or the other consistent in our own software.

ashmaroli:
I prefer `.md` too.. I don't like long extensions

XhmikosR:
Just to share my random opinion, I prefer `.md` too, because it's shorter and consistent with most file extensions out there. Sure, there are long ones too, but like 95% there are small. I do see the point that `.markdown` is self explanatory,` but here, you don't create a new standard :p

DirtyF:
I suggest we add a note on http://jekyllrb.com/docs/posts/ that mention that both extensions are valid. So we can perfectly stick to `.md` for the site docs if we decide to adopt this convention, and use `.markdown` files for educational purpose to users in themes or examples.

Issue #5179, opened 2016-08-02
Get Jekyll Tests Working on Windows


parkr:
Hello, valiant members of @jekyll/windows. We have a quest for you all. @XhmikosR recently added AppVeyor as a CI target for Jekyll so we might be better equipped to maintain compatibility with Jekyll in the longer term. When it was introduced, AppVeyor was constantly failing. We knew this would be a big lift, so we have not been checking failures yet.

Well, 3.2.0 came out and what do you know: a Windows incompatibility. Jekyll 3.2.0 didn't work at all on Windows. Taking a look at the AppVeyor failures, they had the same errors that users were seeing! We could have caught this. I was disappointed (as were all of you) and so now it is time to figure them out once and for all. 

According to [this build](https://ci.appveyor.com/project/jekyll/jekyll/build/108), there are 9 failures (mostly symlink handling or paths):
- [ ] configuring sites should have an array for plugins if passed as a string. (`test/test_site.rb:17`) (Expected: `["/tmp/plugins"]`, Actual: `["C:/tmp/plugins"]`)
- [ ] configuring sites should have an array for plugins if passed as an array. (`test/test_site.rb:24`) (Expected `["/tmp/plugins", "/tmp/otherplugins"]`, Actual: `["C:/tmp/plugins", "C:/tmp/otherplugins"]`)
- [ ] creating sites should sort pages alphabetically.  (`test/test_site.rb:201`) 
  
  ```
  Expected 
  ["%# +.md", ".htaccess", "about.html", "bar.html", "coffeescript.coffee", "contacts.html", "deal.with.dots.html", "dynamic_file.php", "environment.html", "exploit.md", "foo.md", "humans.txt", "index.html", "index.html", "main.scss", "main.scss", "properties.html", "sitemap.xml", "static_files.html", "symlinked-file"]
  but got
  ["%# +.md", ".htaccess", "about.html", "bar.html", "coffeescript.coffee", "contacts.html", "deal.with.dots.html", "dynamic_file.php", "environment.html", "exploit.md", "foo.md", "humans.txt", "index.html", "index.html", "main.scss", "properties.html", "sitemap.xml", "static_files.html"]
  ```
- [ ] filters group_by filter should successfully group array of Jekyll::Page's.  (`test/test_filters.rb:466`) (Expected: `4`, Actual: `5`)
[...]

PR #5180, opened 2016-08-02
Update appveyor.yml and fix optional deps for Ruby x64

XhmikosR:
* rename variables
* add `default-site` target
* remove `fast_finish` in order to match Travis CI behavior

A side note, @parkr, should we do something with a .gemrc to skip docs on `bundle install`?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,39 +12,33 @@ branches:
 build: off
 
 install:
-  - SET PATH=C:\Ruby%ruby_folder_version%\bin;%PATH%
+  - SET PATH=C:\Ruby%RUBY_FOLDER_VER%\bin;%PATH%
   - bundle install --retry 5
 
 environment:
   BUNDLE_WITHOUT: "benchmark:site:development"
   matrix:
-    - ruby_folder_version: "23"
-      ruby_gems_folder: "2.3.0"
+    - RUBY_FOLDER_VER: "23"
+      GEMS_FOLDER_VER: "2.3.0"
       TEST_SUITE: "test"
-    - ruby_folder_version: "23"
-      ruby_gems_folder: "2.3.0"
+    - RUBY_FOLDER_VER: "23"
+      GEMS_FOLDER_VER: "2.3.0"
       TEST_SUITE: "cucumber"
-    - ruby_folder_version: "23"
-      ruby_gems_folder: "2.3.0"
+    - RUBY_FOLDER_VER: "23"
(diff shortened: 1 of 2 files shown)


XhmikosR:
A side note, @parkr, should we do something with a .gemrc to skip docs on `bundle install`?

XhmikosR:
OK so it seems `default-site` fails with

```
script/default-site: setting up tmp directory
script/default-site: creating new default site
bundler: command not found: exe/jekyll
Install missing gem executables with `bundle install`
```

XhmikosR:
I noticed that on Ruby x64, rdiscount wasn't installed so I specified `x64-mingw32` in Gemfile.

XhmikosR:
Another though, how about running cucumber only for the latest (2.3.0) Ruby? It should be -3 builds.

parkr:
> A side note, @parkr, should we do something with a .gemrc to skip docs on bundle install?

That's a terrific idea!

> Another though, how about running cucumber only for the latest (2.3.0) Ruby? It should be -3 builds.

Another great idea. Fewer checks is definitely better.

/cc #5179

XhmikosR:
For now I reduced the cucumber tests.

Do you have a working solution for the no doc on install? I tried one earlier but didn't notice any speed improvement.

parkr:
> Do you have a working solution for the no doc on install? I tried one earlier but didn't notice any speed improvement.

Hm, I googled it and found "Bundler doesn't install RI or RDoc." [source](http://stackoverflow.com/questions/7535737/using-no-rdoc-and-no-ri-with-bundler). So maybe we're ok.

appveyor.yml
```diff
@@ -12,49 +12,41 @@ branches:
...
+  - SET PATH=C:\Ruby%RUBY_FOLDER_VER%\bin;%PATH%
   - bundle install --retry 5
 
 environment:
   BUNDLE_WITHOUT: "benchmark:site:development"
   matrix:
-    - ruby_folder_version: "23"
-      ruby_gems_folder: "2.3.0"
+    - RUBY_FOLDER_VER: "23"
+      GEMS_FOLDER_VER: "2.3.0"
```
parkr: Where is the GEMS_FOLDER_VER used?

XhmikosR: In caching.

parkr: Cool, thanks.

parkr:
LGTM.

Issue #5159, opened 2016-07-28
Release v3.2.1


parkr:
Just going through the issues from yesterday's release. Glad everyone is upgrading! I want to get a v3.2.1 out ASAP, probably Monday as I'm traveling this weekend.

- [x] Fix theme generator so theme files are _included_ rather than rejected :facepalm: (#5152)
- [x] Fix rejection of symlinked themes (e.g. with a Homebrewed installation of Ruby) (#5156)
- [x] Fix Windows compatibility (#5146)
- [x] Release post
- [x] Update history, bump version, tag it & release gem

jleclanche:
So you know, your 3.2.0 release is not showing up in Releases:

https://github.com/jekyll/jekyll/releases

It's only in tags, marking the latest release as 3.1.6 on github.

parkr:
Fixes #4677.

parkr:
Released. http://jekyllrb.com/news/2016/08/02/jekyll-3-2-1-released/

PR #5187, opened 2016-08-02
Site updates for v3.2.1

parkr:
/cc @jekyll/core #5159


parkr:
LGTM, I guess.

Gemfile
```diff
@@ -63,11 +63,11 @@ group :jekyll_optional_dependencies do
...
-  gem "jekyll-gist", "~> 1.0"
-  gem "jekyll-feed", "~> 0.1.3"
-  gem "jekyll-coffeescript", "~> 1.0"
-  gem "jekyll-redirect-from", "~> 0.9.1"
-  gem "jekyll-paginate", "~> 1.0"
+  gem "jekyll-gist"
+  gem "jekyll-feed"
+  gem "jekyll-coffeescript"
+  gem "jekyll-redirect-from"
+  gem "jekyll-paginate"
```
mattr-: Is there a reason we don't want to pin these anymore?

mattr-:
:-1: until the builds pass.

parkr:
> 👎 until the builds pass.

@mattr- Would you take a look at #5188 for me? It worked when I applied that patch to this branch locally. 😄

Gemfile
parkr: :wave: @mattr-!!

A couple of them are used in our tests but are not runtime dependencies. By pinning them, we open ourselves up to issues where we release a new version of one of these, say jekyll-paginate, and we don't catch that it is broken in these tests as we upgrade this code (e.g. we could break an API)

parkr:
LGTM. Travis is passing.

Gemfile
mattr-: :+1:

PR #5189, opened 2016-08-03
Site#configure_theme: do not set theme unless it's a string

### LINKED ISSUE #5163: Undefined method `downcase' error with the latest version
First of all, thank you so much for all your hard work for maintaining and upgrading Jekyll! I just came across something that I believe to be a bug with the latest version of jekyll, since it was resolved when I downgraded to jekyll -v 3.1.6 from the latest version. I would really appreciate it if you could please look into this. (The detailed explanation follows.)

## My Reproduction Steps

**1** I cloned this jekyll template in my local machine: https://github.com/Bloc/portfolio-iro
**2** I ran these 3 commands:
```
$ gem install jekyll
$ gem install jekyll-paginate
$ gem install pygments.rb
``` 
**3** I received the following error when I ran `jekyll serve`.
```
Configuration file: /Users/bloc/code/portfolio-iro/_config.yml
jekyll 3.2.0 | Error:  undefined method `downcase' for #<Hash:0x007fb6bcce5b60>
```
**4** I also tried cloning two other jekyll templates. 
   https://github.com/Bloc/portfolio-kami/ 
   https://github.com/hcbviolet/hcbviolet.github.io

   I repeated the step 2 & 3 with each, but as I ran `jekyll serve`, the same "undefined method `downcase` ..." was the result.

**5** I uninstalled jekyll, reinstalled it by doing `gem install jekyll -v 3.1.6`. Now with this downgraged version, everything seems to be working.

Thank you again for all your help, and please let me know if you need additional information. 

---
- [ ] I Read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/

---

parkr:
Some previous ad-hoc 'themes' used this configuration option to store a hash of values.
In that case, we should simply pretend we have no theme.

Fixes #5163.

/cc @benbalter @emico7 @getaaron @wreichard

### DIFF
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -423,11 +423,14 @@ module Jekyll
 
     private
     def configure_theme
+      self.theme = nil
+      return if config["theme"].nil?
+
       self.theme =
         if config["theme"].is_a?(String)
           Jekyll::Theme.new(config["theme"])
         else
-          Jekyll.logger.warn "Theme:", "value of 'theme' in config should be "
+          Jekyll.logger.warn "Theme:", "value of 'theme' in config should be " \
           "String to use gem-based themes, but got #{config["theme"].class}"
           nil
         end
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -424,7 +424,7 @@ module Jekyll
     private
     def configure_theme
       self.theme = nil
-      self.theme = Jekyll::Theme.new(config["theme"]) if config["theme"]
+      self.theme = Jekyll::Theme.new(config["theme"]) if config["theme"].is_a?(String)
     end
 
     private
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -424,7 +424,14 @@ module Jekyll
     private
     def configure_theme
       self.theme = nil
-      self.theme = Jekyll::Theme.new(config["theme"]) if config["theme"].is_a?(String)
+      return unless config["theme"]
+
+      if config["theme"].is_a?(String)
+        self.theme = Jekyll::Theme.new(config["theme"])
+      else
+        Jekyll.logger.warn "Theme:",
+          "value of 'theme' in config should be String, but got #{config["theme"].class}"
+      end
     end
 
     private
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -423,15 +423,14 @@ module Jekyll
 
     private
     def configure_theme
-      self.theme = nil
-      return unless config["theme"]
-
-      if config["theme"].is_a?(String)
-        self.theme = Jekyll::Theme.new(config["theme"])
-      else
-        Jekyll.logger.warn "Theme:",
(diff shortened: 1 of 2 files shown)


getaaron:
Would it be better to inform the developer of the new use of `theme` instead of silently falling back to the old behavior? I sort of prefer the current crash to this because you know there's a problem. I suggest raising a helpfully worded exception if the object exists but is not a string.

If you disagree and still want to support the old behavior, then this solution looks good to me.

Thanks for fixing so quickly. 👍

lib/jekyll/site.rb
```diff
@@ -424,7 +424,14 @@ def configure_plugins
...
     def configure_theme
       self.theme = nil
-      self.theme = Jekyll::Theme.new(config["theme"]) if config["theme"]
+      return unless config["theme"]
+
+      if config["theme"].is_a?(String)
+        self.theme = Jekyll::Theme.new(config["theme"])
+      else
+        Jekyll.logger.warn "Theme:",
+          "value of 'theme' in config should be String, but got #{config["theme"].class}"
```
parkr: What do you think about this? /cc @benbalter @pathawks

pathawks: LGTM

getaaron: 👍🏼

benbalter: 👍

benbalter:
LGTM.

lib/jekyll/site.rb
```diff
@@ -424,7 +424,7 @@ def configure_plugins
     private
     def configure_theme
       self.theme = nil
-      self.theme = Jekyll::Theme.new(config["theme"]) if config["theme"]
+      self.theme = Jekyll::Theme.new(config["theme"]) if config["theme"].is_a?(String)
```
envygeeks:
```ruby
self.theme = if ...
end
```

Will result in the same thing as those two lines.  That'll rely on you using the if on the setter rather than setting if the condition is met.  However it's a little more optimized that way and results in less messages.

lib/jekyll/site.rb
```diff
@@ -424,7 +424,14 @@ def configure_plugins
     private
     def configure_theme
       self.theme = nil
-      self.theme = Jekyll::Theme.new(config["theme"]) if config["theme"]
+      return unless config["theme"]
```
envygeeks: You can optimize this:

```ruby
private
def configure_theme
  self.theme = 
    if config["theme"].is_a?(String)
      Jekyll::Theme.new(config[
        "theme"
      ])
    else
      Jekyll.logger.warn "Theme:", "value of 'theme' in config should be "
        "String, but got #{config["theme"].class}"
      nil
    end
end
```

parkr: That prints the warning if you don't set `theme` at all, which is not desired. i'll update it.

parkr:
Can I get another LGTM, please? /cc @benbalter

benbalter:
LGTM.

parkr:
@jekyllbot: merge +bug

PR #5195, opened 2016-08-03
Add debug output to theme rendering

benbalter:
Trying to debug some theme issues locally, I found this debug output helpful, and thought rather than silently swallowing errors, it'd be better to at least make warning level notices for missing files or invalid directories.


lib/jekyll/renderer.rb
```diff
@@ -139,6 +139,11 @@ def place_in_layouts(content, payload, info)
       output = content.dup
       layout = site.layouts[document.data["layout"]]
 
+      if layout
+        layout_source = layout.path.start_with?(site.source) ? :site : :theme
+        Jekyll.logger.debug("Layout source:", layout_source)
```
envygeeks: I'd much rather this be folded off into a new line without the variable being created for the GC to handle.

lib/jekyll/theme.rb
```diff
@@ -6,6 +6,8 @@ class Theme
 
     def initialize(name)
       @name = name.downcase.strip
+      Jekyll.logger.debug "Theme:", name
+      Jekyll.logger.debug "Theme source:", root
       configure_sass
```
ashmaroli: I added one more line in my local copy to check if path to layouts was as expected in Windows:
```ruby
Jekyll.logger.debug "Layouts Source:", layouts_path
```

parkr: 👍

parkr:
@benbalter I love all of this! Have a sec to fix this up?

ashmaroli:
@parkr I've resolved the conflicts (albeit made minor changes), keeping @benbalter 's original concept intact.

DirtyF:
@ashmaroli latest rubygem seems to check that the files mentioned in the gemspec are present in the directory, I added the default file listing so fix the `test-theme` cucumber test.

DirtyF:
`gem build` is now checking the validity of the values in the gemspec:

```
WARNING:  See http://guides.rubygems.org/specification-reference/ for help
ERROR:  While executing gem ... (Gem::InvalidSpecificationException)
    "FIXME" or "TODO" is not a summary
    "Put your gem's website or public repo URL here." is not a valid HTTP URI
```

ashmaroli:
@DirtyF Thank you for pushing the fix but IMHO, it should be handled in a separate PR since the error is reproducible on the `master` branch as well.

DirtyF:
@ashmaroli fix in #6784 has been merged on master

lib/jekyll/renderer.rb
```diff
@@ -169,12 +169,16 @@ def place_in_layouts(content, payload, info)
...
-        "in #{document.relative_path} does not exist."
-      )
+      if invalid_layout?(layout)
+        Jekyll.logger.warn(
+          "Build Warning:",
+          "Layout '#{document.data["layout"]}' requested "\
+          "in #{document.relative_path} does not exist."
+        )
+      elsif !layout.nil?
+        layout_source = layout.path.start_with?(site.source) ? :site : :theme
```
parkr: What do you think about adding this method to the `Layout` class, so you could ask the layout where it's from? 😄

ashmaroli: I was planning to open a PR after this merged to do just what you asked.. (but very different from what's here), I was aiming to make most minimal changes to what @benbalter had originally authored..

DirtyF:
@jekyllbot: merge +dev

PR #5196, opened 2016-08-03
Use jekyll-feed to generate the default site's RSS feed

benbalter:
Rather than encouraging each new Jekyll use to hard code a RSS feed template in their generated site (where it's frozen in carbonite and will never receive bug fixes or new features), we should instead rely on the office Jekyll Feed plugin to generate the default site's RSS feed. 

As an added benefit, it gives us an opportunity to show users how to use themes (and further hints that they exist and that they should use them, rather than hard-coding templates they found on the internet).

Practically, Jekyll Feed is a drop in replacement, but is a bit more battle tested, in that it handles a few edge cases surfaced from being used in some many different environments, supports relative URLs in posts, includes author metadata, post summaries, and images (all if present).

You can see the [Jekyll Feed Template](https://github.com/jekyll/jekyll-feed/blob/master/lib/feed.xml) to compare.


parkr:
Nokogiri has C dependencies, so I'm 👎  on this. It will break the default site for most Windows users.

benbalter:
>  It will break the default site for most Windows users.

Does it matter that [it's a dev dependency](https://github.com/jekyll/jekyll-feed/blob/master/jekyll-feed.gemspec#L22) thus shouldn't be a problem for building the site, only for contributing to `jekyll-feed`?

parkr:
> Does it matter that it's a dev dependency thus shouldn't be a problem for building the site, only for contributing to jekyll-feed?

YOU ARE TOTALLY RIGHT. My bad. I just took a look [at the RubyGems.org page](https://rubygems.org/gems/jekyll-feed), saw it and left that comment. Sorry!

LGTM.

benbalter:
:metal: Will give it a bit to sit, in case others want to weigh in.

benbalter:
Also to note, I purposely did not use `{% feed_meta %}` in the template, as that would break the build if for some reason `jekyll-feed` wasn't loaded (e.g., no bundler, removed from Gemfile, etc.). Better to 404 than to fail the build.

On a related note, it'd be awesome to be able to check if a plugin's installed, e.g., `{% if site.plugins | include: "jekyll-feed" %}`.

pathawks:
Cool :+1:

envygeeks:
I'm :+1:, LGTM.

parkr:
@jekyllbot: merge +bug

pathawks:
I wonder if `jekyll-feed` should either _only_ be included in the `Gemfile`, or required in the Gemfile but not in the `jekyll-plugins` group.

I was confused when I removed `jekyll-feed` from `_config.yml` but the feed was still generated.

ashmaroli:
I vote for ```required in the `Gemfile` but not in the `:jekyll-plugins` group.```

PR #5197, opened 2016-08-03
fix path to exe/jekyll for windows

Crunch09:
I hope this fixes `bundler: command not found: exe/jekyll` as seen for example on the [last AppVeyor build](https://ci.appveyor.com/project/jekyll/jekyll/build/153/job/kx1av8u99re60dls#L131). As i do not have a Windows OS i'll only know for sure when AppVeyor runs for this PR.

### DIFF
--- a/script/default-site
+++ b/script/default-site
@@ -10,7 +10,7 @@ rm -Rf ./tmp/default-site
 workdir=$(pwd)
 
 echo "$0: creating new default site"
-bundle exec exe/jekyll new tmp/default-site
+bundle exec ./exe/jekyll new tmp/default-site
 pushd tmp/default-site
 
 echo "$0: respecifying the jekyll install location"


Crunch09:
Well, it didn't 😞 
> bundler: command not found: ./exe/jekyll

Anybody else has an idea?

envygeeks:
AFAIW this is a bundler bug and we have the same problem for some stuff on Linux sometimes.  There is a discussion there and I don't know if it's worth it to try and work around the problem here only to have them fix it when they finally find out what's going on with their bin stuff.

Crunch09:
@envygeeks Thanks! Closing then for now.

ashmaroli:
I bypassed bundler just for this test.
ref: [commit](https://github.com/jekyll/jekyll/pull/5241/commits/c6673e0baf848ee132b8c0df22b86b48fffd9708)

Issue #5192, opened 2016-08-03
Blank new blog generated with Jekyll-3.2.1 on Windows


ashmaroli:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I Updated to the latest Jekyll 3.2.1

---

- [x] I am on (or have tested on) ***Windows*** 7
- [x] I am running Ruby 2.3.1p112 (2016-04-26 revision 54768) [i386-mingw32]

---

- [x] I was trying to create a new blog.

## My Reproduction Steps

```
$ jekyll new my_blog
$ cd my_blog
$ bundle install
$ bundle exec jekyll s
(--redacted--)
generating...
D:/RailsInstaller/Ruby2.3.0/lib/ruby/gems/2.3.0/gems/minima-1.0.1/_layouts/default.html: No such file or directory @ rb_sysopen - /RailsInstaller/Ruby2.3.0/lib/ruby/gems/2.3.0/gems/minima-1.0.1/_layouts/default.html
D:/RailsInstaller/Ruby2.3.0/lib/ruby/gems/2.3.0/gems/minima-1.0.1/_layouts/page.html: No such file or directory @ rb_sysopen - /RailsInstaller/Ruby2.3.0/lib/ruby/gems/2.3.0/gems/minima-1.0.1/_layouts/page.html
D:/RailsInstaller/Ruby2.3.0/lib/ruby/gems/2.3.0/gems/minima-1.0.1/_layouts/post.html: No such file or directory @ rb_sysopen - /RailsInstaller/Ruby2.3.0/lib/ruby/gems/2.3.0/gems/minima-1.0.1/_layouts/post.html
  done in 0.702 seconds.
  Please add the following to your Gemfile to avoid polling for changes:
    gem 'wdm', '>= 0.1.0' if Gem.win_platform?
(--redacted--)
  Server address: http://127.0.0.1:4000/
  Server running... press ctrl-c to stop.
```
When I point my browser to http://localhost:4000/, all I see is a blank page

## The Output I Wanted

http://localhost:4000/ to show my_blog/index.html rendered with minima-1.0.1 theme

parkr:
Where are your gems installed to? What permissions are those directories? It's possible that you installed with `sudo` and Jekyll doesn't have permissions to read those files.

/cc @benbalter

ashmaroli:
My gems are set to path, so they can be accessed from anywhere on the windows machine.
I'm on an Administrator account, so full access to all files/folders.
[...]

PR #5202, opened 2016-08-05
fix theme layout-path for Windows

ashmaroli:
Proposed fix for #5192

- [x] Fix theme layout-path
- [x] Remove redundant existing code in `site.rb`
- [x] Pass CI


parkr:
@ashmaroli This is going to be impossible for us for security reasons. We use `Jekyll.sanitized_path` to mitigate path traversal attacks (e.g. pulling the content of `/etc/passwd` from a Jekyll build). If you can come up with a fix for this which preserves security against these attacks, then I'm 👍 ! In the meantime, this current ref is a no-go for us, unfortunately. Thanks!

PR #5203, opened 2016-08-05
modifying jekyll new command

ashmaroli:
This PR is to showcase an idea I thought would be great if fruitful. It consists of :
- [x] Store the argument(s) passed to `jekyll new` as the assumed name of new blog.
- [x] Call this stored data at various places :
  - [x] Jekyll Info Logger
  - [x] _config.yml -> Live Blog
- [x] Add an ERB template for _config.yml to facilitate above
- [x] Create Gemfile from ERB template instead of within process method
- [x] Modify existing test files to test above changes
- [x] Green build with Travis CI


lib/jekyll/commands/new.rb
```diff
@@ -21,6 +21,7 @@ def init_with_program(prog)
         def process(args, options = {})
           raise ArgumentError, "You must specify a path." if args.empty?
 
+          new_blog_title = args[0]
```
DirtyF: shouldn't it be `arg.first` ?

DirtyF:
@ashmaroli I like the idea of enhancing the `new` command!

Should we change directory and run `bundle install` for the user while we're at it?

ashmaroli:
@DirtyF, As of now, I haven't tested the "serve" part. Apart from the change in the generation process, user interaction hasn't changed. Its the same Gemfile. Though the _config file will currently be having issues. (has `nil` instead of the blog's title value).

so yes, `cd my_blog` & `bundle install`; followed by `bundle exec jekyll s`

lib/jekyll/commands/new.rb
```diff
@@ -34,7 +35,7 @@ def process(args, options = {})
             create_site new_blog_path
           end
 
-          Jekyll.logger.info "New jekyll site installed in #{new_blog_path}."
+          Jekyll.logger.info "New jekyll site \"#{new_blog_title}\" installed in #{new_blog_path}."
```
envygeeks: `%Q{}`

lib/site_template/Gemfile.erb
```diff
@@ -0,0 +1,24 @@
...
+
+# Hello! This is where you manage which Jekyll version is used to run.
+# When you want to use a different version, change it below, save the
+# file and run `bundle install`. Run Jekyll with `bundle exec`, like so:
+#
+#     bundle exec jekyll serve
+#
+# This will help ensure the proper Jekyll version is running.
+# Happy Jekylling!
+gem "jekyll", "<%= Jekyll::VERSION %>"
```
DirtyF: gem "jekyll", "~><%= Jekyll::VERSION %>" ?

ashmaroli: good idea.
~> will only poll for TEENY version updates, lets go for >=version instead. Whatsay?
Or play safe with ~> only?

envygeeks: No.

envygeeks: By that I mean "no" on the ">=", it is not our job to break a users site, "~>" will give them the updates they need should there be a security issue or a major bug in that version.  ">=" will give them 4.0 before they are even ready and break everything, this lands on us and we get the blame.

ashmaroli: Thanks. Will stick with "~>"

lib/jekyll/commands/new.rb
```diff
@@ -21,6 +22,12 @@ def init_with_program(prog)
         def process(args, options = {})
           raise ArgumentError, "You must specify a path." if args.empty?
 
+          if args.size == 1
+            new_blog_title = args[0]
```
chrisfinazzo: I know it's not the "UNIX" thing to do, but can you `.upcase` the title here? Lowercase directory names just don't look right.

Yes, it's a nitpick, but it's _my_ nitpick 😏

envygeeks: What kind of madness is this?

ashmaroli: madness?? Whats so wrong?

envygeeks: Sorry @ashmaroli that was for @chrisfinazzo

lib/jekyll/commands/new.rb
ashmaroli: test.fmt is raising offence warnings for %Q{ } usage. So switching over to nested single quotes,

lib/jekyll/commands/new.rb
chrisfinazzo: The kind of madness where a `new_blog_title` of "blog" should be converted to "Blog". You could get into all sorts of inane arguments about what is an acceptable directory name (because Unicode), however, I am of the opinion that all representations of that name should be uppercase by default. If people want to change it, fine, there's nothing stopping you.

ashmaroli: There is nothing stopping you from having an uppercase dir name anyways.
`the_blue_horn`, `The Blue Horn`, `THE BLUE HORN` is all supported. 
Pass what you like as arguments.

chrisfinazzo: @ashmaroli 
hug it out

ashmaroli:
Almost there.. I need help figuring out how to fix the only remaining failing test.

ashmaroli:
Also Considering:
- Add new option `"classic", "--classic", "Create a new Blog w/o themes like ol' times"`

> Calling this option would simply copy contents from an entirely new directory `lib/classic_template` containing files from Jekyll 3.1.6. 
Expecting it to work like Jekyll 3.1.6.

Crunch09:
> Almost there.. I need help figuring out how to fix the only remaining failing test.

Hi, IMHO `assert_contains` / `assert_includes` only checks if `obj` is in a `collection`, not if all elements of a `sub_collection` are included in `collection`.

[This](https://gist.github.com/Crunch09/f1a417ed62c9bf31398f1f6f6fad9246) should make this test pass again.

ashmaroli:
@Crunch09, Thank you for the assistance. I've pushed a new commit based on your gist. You're correct about the difference between the two assertions. I was just checking if they could get the test to pass.

ashmaroli:
@parkr, Hi, Is it okay if I add a new option `--classic` to the `new` command. This option is for the minor section of users who may like to stick with the < 3.2.x way of Jekylling. 
If yes, append to this PR or create another?

I will be squashing all the commits into one after I get approval for this PR.

pathawks:
It seems that, since creating a new site is something that only happens once, their should be One True Way to do so. I don’t think their needs to be a `--classic` switch.

ashmaroli:
Note: The debate on adding a new `--classic` switch moved to PR - 5217

ashmaroli:
@envygeeks, I was going through the code to see if there was a way to 'colorize' interpolated 
texts ( `#{new_blog_title}` ) within _logger.info_ messages. I didnt find any, so devised a hack. It works locally, but doesn't feel right.
My hack was to add the color method right within the interpolation ( `#{new_blog_title.cyan}` )
Is there another way defined within Jekyll to extend it to this situation?

parkr:
Hey @ashmaroli, I'm gonna have to ask you to split up these changes into discrete PR's. Doing 5+ significant changes in the same PR will slow down review and could result in complete rejection even if some of the ideas are good.

Issue #5207, opened 2016-08-06
Data files as a default collection


benbalter:
For historical reasons, we currently have three data types, each with their own reading and rendering logic:

* Pages
* Documents
* Data files

This causes headaches for anyone trying to build on top of Jekyll in any way that iterates over content, because logic has to be implemented two, often three times, in very different ways (as an example Document has a `path` method, while on Pages you need to call `File.join(@base, @dir, name)` to achieve the same result.

Then you get to data files, which unlike documents and pages, aren't distinct objects and are read directly into the site's config as a nested hash. You can't call `to_liquid` on them and there's no programatic way to get the document's path (since the key is sanitized). 

Now here's the silly thing: collection documents [can be YAML files](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/document.rb#L264), which begs the question: *why don't we just make data files a default collection*?

If I'm reading the code correctly, it should just be a matter of:

1. Teaching documents to [read in CSV files](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/readers/data_reader.rb#L53-L61)

2. Alias `site.data` to `site.collections.data.docs` for backwards compatibility

Theoretically, this should preserve existing behavior, but would allow each data file to be a distinct object and interacted with by plugins, not to mention, would allow us to DRY up the codebase a bunch.

See also https://github.com/jekyll/jekyll/issues/5064, a related issue on being able to disable default collections (which assumes `_data` *is* a collection).

/cc @jekyll/build

shawnbot:
I couldn't find any reference to this in other issues, but would/could this address the desire to treat data files as collections, so that you could have, for instance, a page for each row in a CSV file? E.g. in a config:

```yaml
collections:
  parcels:
    # or maybe this could be a key in site.data?
    source: _data/parcels.csv
    output: true
    permalink: /parcels/:blkgroup/

defaults:
- scope:
    path: ''
    type: parcels
  values:
    layout: parcel
```

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>3.1-stable</code> or <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

[...]

Issue #2983, opened 2014-10-08
Enhancement: "data collections"


phillipadsmith:
Howdy Jekyll Folks,

(Apologies if this idea isn't fully baked, or the right place to ask -- just want to get it out of my head while I'm thinking of it...)

I've been working a bit with [collections](https://github.com/phillipadsmith/everycandidate.org/tree/master/_toronto_city_council) and [data files](https://github.com/phillipadsmith/everycandidate.org/blob/master/_data/toronto_council.csv) recently and I'm curious if anyone else would find the idea of a "data collection" to be useful.

A data collection would simply be a collection where the source for the output of the collection is a data file. For example, where a CSV (or JSON, or YML) file [like this](https://github.com/phillipadsmith/everycandidate.org/blob/master/_data/toronto_council.csv) could result in the same output that is currently generated by [these files](https://github.com/phillipadsmith/everycandidate.org/tree/master/_toronto_city_council). 

Thinking it through a bit, there would likely need to be a way to specify what layout to use to output the collection. Perhaps that could be done in the configuration file, e.g.:

```
collections:
  toronto_city_council:
    output: true
    data: toronto_council.csv
    layout: candidate.html
    permalink: /:some/:val/:from/:data/:file.html
```

The use case, as [demonstrated here](https://github.com/phillipadsmith/everycandidate.org/blob/master/_scripts/generate_candidates.pl), is that I'm often generating [hundreds of stub files for a collection](https://github.com/phillipadsmith/everycandidate.org/tree/master/_toronto_city_council) from the [same CSV](https://github.com/phillipadsmith/everycandidate.org/blob/master/_data/toronto_council.csv) that is later used to [populate much of the data for that collection's output](https://github.com/phillipadsmith/everycandidate.org/blob/master/_layouts/candidate.html).

Let me know if any of this makes sense. And, if so, what would be helpful in thinking it through further.

Phillip.

davidrleonard:
Not to resurrect a dead thread there, but wondering if any more thought was given to this? This is also a big need for me in designing large Jekyll sites that have a lot of pages with very similar content (i.e. bio pages for hundreds of people). I'd love to be able to manage this data elsewhere, like a Google Sheet, and export it to a single file then build my pages out from there.

envygeeks:
This ticket is being locked. If you have an issue (or feature request) please file a new ticket instead of practicing the art of necromancy. It has hard to maintain actual issues against comments if this happens.

PR #5208, opened 2016-08-06
WIP: Convert data dir to a collection

benbalter:
Testing out the idea of converting the `_data` directory to a default collection so that data files are first-class objects. See https://github.com/jekyll/jekyll/issues/5207 for additional context.

Right now, it *theoretically* works (tests pass locally), and *should be* fully backwards compatible.

This is accomplished by initializing `data` as a default collection, moving the reading logic from DataReader to Document, and by creating a new DataDrop, which we expose as `site.data` to avoid a breaking change.

In this sense, existing `site.data.foo` and `site.data["foo"]` access methods *should* work.

It was actually a smaller diff than I thought. The biggest change/noise here is updating the existing tests to expect a second default collection, but otherwise, things Just Work.

There is, however, one gotcha: In various places, Jekyll assumes a document's data will be a hash. Theoretically a document's front matter *could* be an array (it's valid YAML), and a data file, definitely can be an array (and many of our fixtures are). For now, since this is just an early proof of concept, I short circuited the drop, and just returned the raw hash when `to_liquid` is called to prevent errors.

Thoughts? Dumb idea?

### DIFF
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -131,11 +131,32 @@ module Jekyll
         relative_path[0..-extname.length - 1].sub(collection.relative_directory, "")
     end
 
+    # Returns an array of sub directories that represents the document's path
+    # relative to the collection directory
+    def subdirs
+      cleaned_relative_path.split(File::SEPARATOR)
+        .reject { |c| c.empty? || c.eql?(basename_without_ext) }
+    end
+
     # Determine whether the document is a YAML file.
     #
     # Returns true if the extname is either .yml or .yaml, false otherwise.
     def yaml_file?
       %w(.yaml .yml).include?(extname)
+    end
+
+    # Determine whether the document is a JSON file.
+    #
+    # Returns true if extname == .json, false otherwise.
+    def json_file?
+      ".json" == extname
+    end
+
+    # Determine whether the document is a CSV file.
+    #
+    # Returns true if extname == .csv, false otherwise.
+    def csv_file?
+      ".csv" == extname
     end
 
     # Determine whether the document is an asset file.
@@ -166,7 +187,7 @@ module Jekyll
     # Returns false if the document is either an asset file or a yaml file,
     #   true otherwise.
     def render_with_liquid?
-      !(coffeescript_file? || yaml_file?)
+      !(coffeescript_file? || yaml_file? || json_file? || csv_file?)
     end
 
     # Determine whether the file should be placed into layouts.
@@ -174,7 +195,7 @@ module Jekyll
     # Returns false if the document is either an asset file or a yaml file,
     #   true otherwise.
     def place_in_layout?
-      !(asset_file? || yaml_file?)
+      !(asset_file? || yaml_file? || yaml_file? || json_file? || csv_file?)
     end
 
     # The URL template where the document would be accessible.
@@ -260,8 +281,13 @@ module Jekyll
     def read(opts = {})
       Jekyll.logger.debug "Reading:", relative_path
 
-      if yaml_file?
+      if yaml_file? || json_file?
         @data = SafeYAML.load_file(path)
+      elsif csv_file?
+        @data = CSV.read(path, {
+          :headers  => true,
+          :encoding => site.config["encoding"]
+        }).map(&:to_hash)
       else
         begin
           defaults = @site.frontmatter_defaults.all(
@@ -341,6 +367,7 @@ module Jekyll
     #
     # Returns a Hash representing this Document's data.
     def to_liquid
+      return data if data.is_a?(Array)
       @to_liquid ||= Drops::DocumentDrop.new(self)
     end
 
@@ -366,7 +393,9 @@ module Jekyll
     #   equal or greater than the other doc's path. See String#<=> for more details.
     def <=>(other)
       return nil unless other.respond_to?(:data)
-      cmp = data["date"] <=> other.data["date"]
+      if data.is_a?(Hash) && other.data.is_a?(Hash)
+        cmp = data["date"] <=> other.data["date"]
+      end
       cmp = path <=> other.path if cmp.nil? || cmp.zero?
       cmp
     end
@@ -428,12 +457,12 @@ module Jekyll
     # Override of normal respond_to? to match method_missing's logic for
     # looking in @data.
     def respond_to?(method, include_private = false)
-      data.key?(method.to_s) || super
+      data.is_a?(Hash) && data.key?(method.to_s) || super
     end
 
     # Override of method_missing to check in @data for the key.
     def method_missing(method, *args, &blck)
-      if data.key?(method.to_s)
+      if data.is_a?(Hash) && data.key?(method.to_s)
         Jekyll::Deprecator.deprecation_message "Document##{method} is now a key "\
                            "in the #data hash."
         Jekyll::Deprecator.deprecation_message "Called by #{caller.first}."
--- a/lib/jekyll/readers/data_reader.rb
+++ b/lib/jekyll/readers/data_reader.rb
@@ -1,70 +1,29 @@
 module Jekyll
   class DataReader
     attr_reader :site, :content
+
+    EXTENSIONS = %w(.yaml .yml .json .csv).freeze
+
     def initialize(site)
       @site = site
-      @content = {}
       @entry_filter = EntryFilter.new(site)
     end
 
-    # Read all the files in <source>/<dir>/_drafts and create a new Draft
-    # object with each one.
-    #
-    # dir - The String relative path of the directory to read.
-    #
-    # Returns nothing.
-    def read(dir)
-      base = site.in_source_dir(dir)
-      read_data_to(base, @content)
-      @content
-    end
+    def read_data(dir)
+      site.reader.get_entries(dir, site.config["data_dir"]).map do |entry|
+        next unless EXTENSIONS.include?(File.extname(entry))
 
-    # Read and parse all yaml files under <dir> and add them to the
-    # <data> variable.
-    #
-    # dir - The string absolute path of the directory to read.
-    # data - The variable to which data will be added.
-    #
-    # Returns nothing
-    def read_data_to(dir, data)
-      return unless File.directory?(dir) && !@entry_filter.symlink?(dir)
-
-      entries = Dir.chdir(dir) do
-        Dir["*.{yaml,yml,json,csv}"] + Dir["*"].select { |fn| File.directory?(fn) }
-      end
-
-      entries.each do |entry|
-        path = @site.in_source_dir(dir, entry)
+        path = @site.in_source_dir(File.join(dir, site.config["data_dir"], entry))
         next if @entry_filter.symlink?(path)
 
-        key = sanitize_filename(File.basename(entry, ".*"))
-        if File.directory?(path)
-          read_data_to(path, data[key] = {})
-        else
-          data[key] = read_data_file(path)
-        end
-      end
-    end
+        doc = Document.new(path, {
+          :site       => @site,
+          :collection => @site.data_collection
+        })
+        doc.read
 
-    # Determines how to read a data file.
-    #
-    # Returns the contents of the data file.
-    def read_data_file(path)
-      case File.extname(path).downcase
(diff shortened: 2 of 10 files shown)


lib/jekyll/document.rb
```diff
@@ -166,15 +187,15 @@ def coffeescript_file?
...
+      !(coffeescript_file? || yaml_file? || json_file? || csv_file?)
     end
 
     # Determine whether the file should be placed into layouts.
     #
     # Returns false if the document is either an asset file or a yaml file,
     #   true otherwise.
     def place_in_layout?
-      !(asset_file? || yaml_file?)
+      !(asset_file? || yaml_file? || yaml_file? || json_file? || csv_file?)
```
DirtyF: `yaml_file?` is called twice

spudowiar: @DirtyF is it a `yaml_file?` or is it **REALLY* a `yaml_file?` 😏

kenman345: Perhaps `yml_file?`

pathawks:
Just merged `master`.

Tests are failing because a `_data/` file can be an `Array` which does not respond to `.key?` as is expected for a `Document`.

lib/jekyll/document.rb
```diff
@@ -180,15 +180,24 @@ def coffeescript_file?
     #   or if the document doesn't contain any Liquid Tags or Variables,
     #   true otherwise.
     def render_with_liquid?
-      !(coffeescript_file? || yaml_file? || json_file? || csv_file?)
+      return false if data["render_with_liquid"] == false
+      return false if coffeescript_file? || yaml_file? || json_file? || csv_file?
+      Utils.has_liquid_construct?(content)
```
pathawks: I broke these two lines up because Rubocop was complaining about line length. This reads better to me anyway.

lib/jekyll/document.rb
```diff
@@ -175,7 +197,7 @@ def no_layout?
     # Returns false if the document is set to `layouts: none`, or is either an
     #   asset file or a yaml file. Returns true otherwise.
     def place_in_layout?
-      !(asset_file? || yaml_file? || no_layout?)
+      !(asset_file? || yaml_file? || yaml_file? || json_file? || csv_file? || no_layout?)
```
kenman345: `yaml_file?` twice?

lib/jekyll/document.rb
```diff
@@ -262,8 +284,12 @@ def published?
     def read(opts = {})
       Jekyll.logger.debug "Reading:", relative_path
 
-      if yaml_file?
+      if yaml_file? || json_file?
```
kenman345: maybe combine this with the `csv_file?` as a `data_file?` method. Then check for if its a `csv_file?` and process it accordingly.

lib/jekyll/document.rb
```diff
@@ -123,13 +123,34 @@ def cleaned_relative_path
...
     #
     # Returns true if the extname is either .yml or .yaml, false otherwise.
     def yaml_file?
       %w(.yaml .yml).include?(extname)
     end
 
+    # Determine whether the document is a JSON file.
+    #
+    # Returns true if extname == .json, false otherwise.
+    def json_file?
```
kenman345: I'm wondering if this should really exist or exist in its current form. YAML files are very close to JSON files and will be processed the same way. I am wondering if `json_file?` should include ` || yaml_file?` so any place it is used can be reduced since they're always side by side.

kenman345:
I am going to take a guess @pathawks  and say that the tests might be failing in part due to erroring  not loading properly. When Collections evaluates static files or generated/data files, it looks for the frontmatter on files. 

In [`data.feature`](https://github.com/jekyll/jekyll/blob/master/features/data.feature) we see that the first thing after the `When I run jekyll build` is not an assert that it exited with status 0. Instead it goes right into saying things exist.

The `data.feature` definitions do not include frontmatter/`---` on the first line of the yaml files. This is what I noted in my PR thats open (#7147). 

To fix this we may need to add a check in `collection.rb` to not treat the `.yaml` and `.yml` files, as well as `.json` files as static files and read them into the collection.

kenman345:
@pathawks Here we go:
https://github.com/jekyll/jekyll/blob/af6320940427acfe0968c58f33e04526cda17914/lib/jekyll/collection.rb#L54-L68

Note line 61 where we only check if it has a yaml_header? but not if the ending is of a data type. So the error about a missing key/value is because the page its trying to generate is trying to use a key/property of `products` on `site.data` in the first feature definition that is not even getting loaded after the change is made to load in files as a collection.

kenman345:
@benbalter @pathawks @ashmaroli I am happy to submit a PR for this branches PR to fix the issues, and potentially bring up to speed some enhancements that go along with the noted situations in the referenced issues if I can get some confirmation:

1. Noticed .tsv is removed from datareader as being recognized. Is this a miss or deliberate change?
2. Everything the datareader sees as a data file should be also included by collection.rb to process as a data file?
---
**Below is partially off topic as it pertains to other issues/PRs but also on topic as it directly relates to this code and currently is an issue for the same reasons this PR is failing tests:**

As a personal request, I'd love to get this answered while we're on the topic:

3. Enhancement on top of PR: If we're moving collections to process all data types like `_data` does now, can we also change `@data = SafeYaml.load_file(path)` to store to a variable like `read_content` does so that it can merge the data files into the collections hash? This would resolve the issue of same name/different extension data files having the same `id` in the collection hash. It also addresses a solution to #2983 and to my own PR (#7147) where users can set a .md file with the same name as any data file within the same folder/collection to have it merge the data sets. 

This allows for script/large data exports to be used directly by Jekyll without having to modify the data for it to be recognized by Jekyll. This can understandably be done as a PR after this is merged and would allow me to close my PR that's more focused on only adding in YAML files with a more focus driven solution that makes sense for Jekyll from my understanding.

kenman345:
If I can get some feedback I can make a PR for this branch to help get it up to speed.

lib/jekyll/readers/data_reader.rb
```diff
@@ -3,73 +3,28 @@
...
-
-      entries.each do |entry|
-        path = @site.in_source_dir(dir, entry)
+        path = @site.in_source_dir(File.join(dir, site.config["data_dir"], entry))
         next if @entry_filter.symlink?(path)
 
-        if File.directory?(path)
-          read_data_to(path, data[sanitize_filename(entry)] = {})
-        else
-          key = sanitize_filename(File.basename(entry, ".*"))
```
kenman345: @pathawks After some testing locally, I believe the translation of this to collections is needed. The key used to be generated here and used in the line below, but with the collection handling it, we do not have a key being set. So if I want data to exist I need to have something like 
```yaml
---
products:
- name: sugar
  price: 5.3
- name: salt
  price: 2.5
```

in `_data/products.yaml` in order for `site.data.products` to be properly initiliazed, and then it still would be a bit different than how it is now I believe. Give it a try and have a index.html with "{{site.data | jsonify}}" in it. Take a look at the output compared to how it is currently without this Branch/PR

lib/jekyll/document.rb
```diff
@@ -262,8 +284,12 @@ def published?
     def read(opts = {})
       Jekyll.logger.debug "Reading:", relative_path
 
-      if yaml_file?
+      if yaml_file? || json_file?
         @data = SafeYAML.load_file(path)
+      elsif csv_file?
+        @data = CSV.read(path,
```
kenman345: We can check for .tsv here and use "," for .csv and "\t" for .tsv as the col_sep so we dont have to make two different paths. 

```ruby
:col_sep  => column_delimiter, 
```

and the method:
```ruby


    # Determine the column delimiter for ___SeparatedValue files.
    #
    # Returns "\t" if extname is `.tsv`, returns "," otherwise 
    def column_delimiter
      return "\t" if extname == ".tsv"
      ","
    end
```

kenman345: this also allows for future updates if separator is bar or colon separated...

lib/jekyll/readers/data_reader.rb
kenman345: @pathawks to see what I’m saying easily, try making a new site with `_data/[anything here is fine].yml` With it being a valid yaml data set like that in the data.feature definitions.

Make an index.html that contains {{site.data | jsonify }} and {{ site.collections | jsonify }}

You will see the difference I’m seeing where the collections appears to get the data but the site.data hash only seems to get the name of the data files as a key but no values.

kenman345:
@pathawks I took a crack at it but its definitely still incomplete. Heres the commit on my personal fork of this repo: https://github.com/kenman345/jekyll/commit/24018d8ef5fae9fb52258ea4e4d2eead3b7bf0d4

This solves most of the data collection issues (whoops, noticed I committed with some commented things still there, but ignore those for now.) but still leaves a few, which is why no PR to merge to this PR.

1. the configuration `data:` collection seems to not be backwards compatible with these changes after what I've done.
2. I would always get tons of errors even compiling if I simply had it set @data = the data file content. I guess I am not finding something that allows us to control it looking for a layout and rendering the page, so if we address that then changing up the hash location we store the data at wont be necessary?!?
3. Since I am not storing the data information to the root of site.data in ruby, we are potentially able to get some breaking of plugins built around this. So we would want to address a proper fix or documentation on how to convert to Jekyll 4.0 methodology if this is how it had to be.

I hope that it helps you guys figure out whats wrong and how to proceed from there. I am still new to all this Ruby stuff, let alone Liquid/Jekyll nuances, so I understand its rough. Just trying to help.

kenman345:
May need to merge master back into this PR to resolve the conflicts. 

Also, @benbalter what do you think of the revisions I made? I believe this brings your PR code complete unless we are able to figure out a way to not have any breaking changes to the string normalization.

lib/jekyll/document.rb
```diff
@@ -123,13 +123,42 @@ def cleaned_relative_path
...
+    def delimiter_separated_file?
+      %w(.csv .tsv).include?(extname)
+    end
+
+    # Determine the column delimiter for ___SeparatedValue files.
+    #
+    # Returns "\t" if extname is `.tsv`, otherwise returns ",".
+    def column_delimiter
+      return "\t" if extname == ".tsv"
+      ","
```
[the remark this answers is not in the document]
DirtyF: Fine, we can deal with this in a dedicated PR.

test/test_site.rb
```diff
@@ -508,9 +497,9 @@ def convert(*_args)
                                             source_dir, "_data", "categories.01", "dairy.yaml"
                                           ))
 
-        assert_equal site.data["categories01"]["dairy"], file_content
+        assert_equal site.data["categories.01"]["dairy"], file_content
```
DirtyF: Why did you change the test here?

lib/jekyll/document.rb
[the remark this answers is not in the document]
kenman345: Yea, I figured no need to add a configuration option until someone reports an issue as it really is a rare thing. I figured it might be good to point out why I decided on this code pattern for a return.

test/test_site.rb
kenman345: This was one of the major things I had a concern about and was unsure how to address.

Due to the substitution changes/removal from data_reader we have this naming of property change to consider.

I think we might be able to address this though so the only breaking change is the index of the collections property on `site` instead.

kenman345: Looking it over its likely a fix would need to exist in the data_drop.rb but I was unsure what the change should be. If not there, then the sanitize_Filename method that got removed would be the right place. 

Either way, I think this is an avoidable breaking change given someone that is a bit more familiar with the sanitization we are doing. I do however like the idea that this makes data collections inline with the property references of  a normal collection, which means if you find yourself using data collection for too much and decide to break it out into its own collection later on then you have much less to worry about in terms of referencing the data.

kenman345:
beep beep, reminder this is still a PR. If its going to get no love then perhaps it be indicated that its being backburnered until after 4.0

DirtyF:
@kenman345 we have 70 of these, no need to nudge maitainers, we'll proceed when we can.

kenman345:
I am not sure what the conflicts are. I dont see any that should exist and the merge resolution is not something I have access to

PR #5209, opened 2016-08-07
WIP: Allow front matter to be optional for pages

benbalter:
This changes the reader logic to allow markdown Pages to be rendered without front matter.

The idea being, that with Themes available in 3.2, a site could be as simple as a config file that requests a theme, and a series of Markdown files.

For simple "hello world" sites, this flag would remove the step of having to learn what front matter is, and add it to a markdown file.

Right now I implemented a `front_matter_optional` config flag, which default to `false`. When set, any file, for which a converter is available, regardless of if it has YAML front matter, will be read in as a page.

If there's interest in this feature, I can work on adding tests, but wanted to pause before I went too far down the rabbit hole.

For context, I'd like to add this feature because with GitHub Pages, we can set a default theme, and a default template in our configuration defaults. That means that a user could have their first website by literally just creating a markdown file, no need to learn what YAML front matter is, or worry about adding it to each file.

This could be alternatively implemented as a plugin, but I think it's cleaner to adjust how files are read in, rather than reading them in as a second pass and rendering them independently.

### DIFF
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -21,6 +21,7 @@ module Jekyll
       "keep_files"        => [".git", ".svn"],
       "encoding"          => "utf-8",
       "markdown_ext"      => "markdown,mkdown,mkdn,mkd,md",
+      "front_matter_optional" => false,
 
       # Filtering Content
       "show_drafts"       => nil,
--- a/lib/jekyll/reader.rb
+++ b/lib/jekyll/reader.rb
@@ -40,8 +40,12 @@ module Jekyll
       dot = Dir.chdir(base) { filter_entries(Dir.entries("."), base) }
       dot_dirs = dot.select { |file| File.directory?(@site.in_source_dir(base, file)) }
       dot_files = (dot - dot_dirs)
+
+      converters = site.converters.reject { |c| c.class == Converters::Identity }
       dot_pages = dot_files.select do |file|
-        Utils.has_yaml_header?(@site.in_source_dir(base, file))
+        Utils.has_yaml_header?(@site.in_source_dir(base, file)) ||
+          site.config["front_matter_optional"] &&
+            converters.any? { |c| c.matches(File.extname(file)) }
       end
       dot_static_files = dot_files - dot_pages
 


envygeeks:
I'm on the fence with this, not because I don't like the idea, but because it also makes way more sense for posts too, where most of the data can be inferred from the path, unless the user wants to get ultra specific.  I would like to see this extended to posts as well.

jake-low:
@envygeeks: based on my investigations in #5064 (here's a [direct link](https://github.com/jekyll/jekyll/issues/5064#issuecomment-236484168)) frontmatter for posts is already optional. This was just a simple test of one version of Jekyll though, it's certainly possible that there's complexity I failed to discover.

Anyway, I'm in favor of a change like this, because as far as I can tell it reduces the difference between `_posts` and other document types. My initial foray was related to collections (correct me if I'm wrong, but I believe `pages` is not a collection), but in general making Jekyll more flexible for advanced use cases by reducing the "magic" around posts, pages, and data (but keeping sane defaults for a blog-plus-pages type site, to remain friendly for new users) seems like a worthy goal to me.

Wrt. configuration, would it be easy to make the setting controlling whether frontmatter is optional scoped per-collection, rather than globally? It seems to me that this would be more friendly to backwards-compatibility since the default for `_posts` could be true, but other collections could default to false.

spudowiar:
I'm :-1: on this, front matter is explicit, when you omit it you're telling Jekyll it's a plain file that it shouldn't touch.

benbalter:
>  you're telling Jekyll it's a plain file that it shouldn't touch.

@spudowiar Can you talk a bit more about when you think a user would have a `.md` file in their site that they'd want served as raw plaintext?

frickler01:
Big `+1` @benbalter

>  no need to learn what YAML front matter is

since that was a noticable road block to overcome when adopting Jekyll as a newcomer. For non-programmers I assume it to be a real obstacle.

Also `+1` @jake-low 

> reducing the "magic" around posts, pages, and data

I still think that they all should be the same. Posts have a date in the filename, everything else is a page. Data has data converters e.g. for `.json` or `.yaml` - then it just goes to the data dict. More data in the headers. If there is no converter for the file type, copy verbatim.

tgragnato:
Static site generators are a great small step into a post-CMS era,
but if the _aim_ is to complete a giant leap for **tech unsavvy users**,
then Jekyll is hard in many other subtle ways other than `YAML`.

For someone is interested in Jekyll, front matter should not be that adoption barrier.
> designed for human readability and interaction 😄 😄 😄 😄 😄 😄 

Anyway, no one use markdown as plaintext, so it can be useful.

frickler01:
@tgragnato non-programmer ≠ tech-unsavvy

spudowiar:
> non-programmer ≠ tech-unsavvy

:+1:

Non-programmer might be our target audience, but not tech-unsavvy

tgragnato:
I agree on that

Still prefer a separated "advanced" _fronted_ ( gui / **web** / whatelse ) ,
that allows you to format .md with proper front matter upon creation,
rather than placing the user in a _backend_.

Style matters xD

benbalter:
While the appropriateness and scope of this proposal for Jekyll core is being discussed, I created a lightweight plugin to implement the behavior: https://github.com/benbalter/jekyll-optional-front-matter.

jackphel:
I can see the usefulness for a lot of people. However, this is a
substantial backwards incompatible change to a core invariant in Jekyll's
processing. There is a pretty strong statement in the docs:
https://jekyllrb.com/docs/plugins/#converters that this will never happen.
It's a *big deal* to abandon something like that.

parkr:
@benbalter You wrote a plugin for this – does that suffice for you or do you still think this should be a core feature? I could really go either way. It must be opt-in if it is placed in core. Thanks!

benbalter:
> do you still think this should be a core feature?

I never understood why Markdown Pages required front matter (possibly "do what you ask, no more no less"), but do not have the time to see this through myself. If someone else would like to pick things up, I'd be 👍 but am not strongly opinionated.

parkr:
> I never understood why Markdown Pages required front matter (possibly "do what you ask, no more no less")

Yeah this dates back to well before I started. The idea was that conversion was opt-in, I think. I'd have to ask an old-timer but they might not even remember.

Will close this out for now then as the plugin satisfies this. Thanks!

PR #5210, opened 2016-08-07
Exclude node_modules by default

parkr:
If no 'exclude' directive is specified, exclude node_modules by default.

https://twitter.com/mxstbr/status/761856359579185153

/cc @jekyll/build

### DIFF
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -17,7 +17,7 @@ module Jekyll
       # Handling Reading
       "safe"              => false,
       "include"           => [".htaccess"],
-      "exclude"           => ["vendor"],
+      "exclude"           => %w(node_modules vendor),
       "keep_files"        => [".git", ".svn"],
       "encoding"          => "utf-8",
       "markdown_ext"      => "markdown,mkdown,mkdn,mkd,md",
--- a/site/_docs/configuration.md
+++ b/site/_docs/configuration.md
@@ -594,7 +594,7 @@ collections:
 # Handling Reading
 safe:         false
 include:      [".htaccess"]
-exclude:      ["vendor"]
+exclude:      ["node_modules", "vendor"]
 keep_files:   [".git", ".svn"]
 encoding:     "utf-8"
 markdown_ext: "markdown,mkdown,mkdn,mkd,md"
--- a/test/test_configuration.rb
+++ b/test/test_configuration.rb
@@ -48,6 +48,12 @@ class TestConfiguration < JekyllUnitTest
     end
   end
 
+  context "the defaults" do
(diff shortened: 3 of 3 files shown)


mxstbr:
No way, this is awesome! 🙌 Glad to have inspired a positive change in Jekyll!

spudowiar:
I don't see why we should have defaults like this, people will only ask to make the list bigger. This doesn't affect people in normal usage of Jekyll so there's no need to have it as default, in my opinion.

spudowiar:
If you do decide to go through with this, please document it

envygeeks:
If this is gonna happen, can we please also do the same with vendor since it's also a sore spot?

parkr:
> If this is gonna happen, can we please also do the same with vendor since it's also a sore spot?

@envygeeks That's a great idea!

> I don't see why we should have defaults like this, people will only ask to make the list bigger. This doesn't affect people in normal usage of Jekyll so there's no need to have it as default, in my opinion.

@spudowiar So many people are using Gulp and Grunt to run Jekyll and confused when it takes like 30s to generate their site. Ultimately, it's a sad result of using NPM that you end up with 15,000 files in a `node_modules` directory. At least this way we don't unnecessarily slow builds – you basically *never* want this directory in your resulting site. 😄

spudowiar:
OK, I see your point. Well `vendor` and `node_modules` do seem like sane defaults, just make sure you document it! :)

Strangehill:
on the topic of reasonable defaults: Gemfile, Gemfile.lock, package.json

Strangehill:
Why 
> If no 'exclude' directive is specified, exclude node_modules by default.

will specifying an 'exclude' require

parkr:
This is ready for another round of LGTM's.

parkr:
@jekyllbot: merge +minor

arechsteiner:
@parkr: How to *include* node_modules now that this change has been released?

I've tried this in my `_config.yml`:

    include: ["node_modules"]

However, the folder node_modules is still excluded.

I'm not very well-versed in Ruby but I think your commit actually forces node_modules to be excluded, instead of making it the default behavior that can be overridden by the user. This is not the intention, or is it?

mxstbr:
I think just setting `exclude` to nothing should work perfectly fine:

```
exclude: []
```

arechsteiner:
Thanks @mxstbr, this works. It's not very intuitive IMHO. Maybe this could be improved some time to honor the include directive and be in line with the documentation:

> **Include**
> 
> **Force inclusion of directories** and/or files in the conversion. .htaccess is a good example since dotfiles are excluded by default. 

That's definitely not what happens with this example.

PR #5211, opened 2016-08-08
WIP: Pages as Documents

benbalter:
This pull request is intended to convert Pages into fully backward compatible Documents to give them parity with other content types. The motivation stems from having to implement everything twice, and in two slightly different ways, in support of [Jekyll Admin](https://github.com/jekyll/jekyll-admin).

As of right now, unit tests pass, and most integration tests do. The example plugin test, for example, fails, because the test monkey patches `Pages#initialize` (which I'm fine not supporting).

A few things to note:
- `Page` now extends `Document`
- `Page.new` can be called with 2 arguments (as a Document) or 4 (as a Page) to maintain backwards compatiblity
- Traditionally Pages did not have titles, but Document did. This has been fixed.
- Traditionally Pages were sorted by their `basename` in Document speak (slug without reference to parent diretories). Documents without dates are sorted by their full path, thus `/a/index.html` comes before `/b/index.html`, which I'd call a bug fix/improvement
- `site.pages` works as expected, but with a deprecation warning
- Traditionally, `Page#path` was _relative_ while `Document#path` was absolute. To resolve this:
  - Documents now have `absolute_path` and `relative_path` methods to be explicit
  - Calling `Page#path` will return `relative_path`, but throw a deprecation warning
  - All internal references to `Document#path` were change to `Document#absolute_path`
- All Pages methods should still work as expected, and thus _shouldn't_ be a breaking change


alfredxing:
I did some work on this also in #4822. The one breaking-change issue we had was with plugins and hooks subclassing Page and overriding the constructor (as with the test plugin -- see the discussion in that PR).

parkr:
> I'm thinking about moving Page to use Drop before we make this change since this will likely be breaking...

:+1:

pathawks:
A lot has changed since this PR was created.
I have done my best to merge `master` into this PR, but there are still several failing tests.

A strange failure that I keep seeing locally:

```text
Minitest::Assertion: No visible difference in the Hash#inspect output.
You should look at the implementation of #== on Hash or its members.
```

It's just comparing two hashes, and the diff shows no differences. What gives?

letrastudio:
From what I can gather by looking at the comments and commits, this change seems mostly concerned with simplifying implementation. But I have a few questions about potential effects on behavior. Seeing as how 4.0 is the milestone now, I’m hoping for some breaking changes over strict backwards compatibility, if it means getting rid of odd quirks.

- Would this allow moving pages to a `_pages` folder? (or maybe just some pages, with everything being merged into `_pages` at build time?)
- When using `collections_dir` would pages go in or stay out?
- Would `{{ page }}` variables get unified once and for all? The [documentation](https://jekyllrb.com/docs/variables/#page-variables) confusingly refers to “page” variables of everything _but_ pages

PR #5217, opened 2016-08-09
Add --classic switch to 'jekyll new' command

ashmaroli:
This PR proposes to extend the `jekyll new` command with following:
- [x] Add a new `--classic` switch for those not interested in the new theme functionality.

Calling this option would simply copy contents from an entirely new directory `lib/site_classic_template` containing files from Jekyll 3.1.6 ( though slightly modified to contain recent updates). This move hopes to minimize the confusion over missing folders from generating a new blog with Jekyll 3.2., at the same time opening up the choice to have the best of both worlds _(pre-v3.2 & current release)_ with one gem. 

**No need of downgrading to v3.1.6.**
**One Gem. Many choices.**


geraldb:
Hello,
  A repost to your forum post. Have you tried `drjekyll new starter`?

It does what you want e.g. it lets you create a new "classic" style theme - in fact you can choose any of 100+.

[All the code is open source and public domain](https://github.com/drjekyllthemes/drjekyll). Try:

    $ gem install drjekyll
    $ drjekyll new starter

Of course - than you use "plain" jekyll

    $ cd starter
    $ jekyll build

etc.

Another example with its own new starter command is [ye good old octopress](https://github.com/octopress/octopress), definitely needs more love/care/contributions. Cheers.

PS: If `drjekyll` and `octopress` do NOT work, you can always create your own e.g. use `yeoldjekyll` or something ;-)

PPS: The Jekyll team is - of course - invited to incorporate the drjekyll and mrhyde command line tools - if that is holding you back (e.g. you want a "blessed" official command).

DirtyF:
@geraldb you shouldn't have to install another gem to bootstrap a jekyll project, that's a core feature.

@ashmaroli thanks for implementing `bundle install` in the `new` command, maybe it would be better to do a separate PR for this, don't you think?

ashmaroli:
@DirtyF, I included both the new features into the same PR for following:
 - I already have another PR modifying the new command.
 - Both new features of this PR involve changes to the same file, it would be easier for me and the maintainers to track the changes / conflicts.

@parkr, whats your take on this?

ashmaroli:
I also think Bundler should now be added to the .gemspec as runtime dependency so that it installs automatically with `gem install jekyll`

geraldb:
> @geraldb you shouldn't have to install another gem to bootstrap a jekyll project, that's a core feature.

Just add the drjekyll gem to the jekyll gem dependency than it's a core feature. ;-) It's that easy. Jekyll needs a (theme) package manager.

Crunch09:
I'm with @pathawks on this one, [don't think we need a classic switch](https://github.com/jekyll/jekyll/pull/5203#issuecomment-238117347).

ashmaroli:
Okay.. so that's 2 thumbs-down against my own one +1..

envygeeks:
What exactly is this switch intended to solve?  What happens if the user isn't using bundler?

On the "should this be two pull requests" that's an ideal situation as parallel discussions need to be had for both of them, which can ultimately lead to problems with this pull request but it's really not a big deal no matter what, considering it's easy to rebase.  It's all good.

ashmaroli:
@envygeeks, A few users expressed their confusion over the missing folders, and inability to use Jekyl like the were used to, at Jekyll Talk. So the `--classic` switch is intended for such similar users, a very small minority albeit. Using the classic switch while generating a new blog creates the same folder structure last seen in v3.1.6. That blog doesnt need to be bundle installed as there is no extra gems required to be `served` 
I was hoping to have bundler added as a dependency. Then it would get installed alongwith the Jekyll gem. So didnt add a step to raise an error and gracefully abort if 'bundler' is not installed in the user's path.

I'll branch off the 'bundle install feature' to a new PR and devote this PR just for discussion on the 
--classic switch

ashmaroli:
Note: Adding `bundle install` feature to `jekyll new` command has been moved to #5225

ashmaroli:
updated this PR to be a standalone pull instead of an extension of another PR ( as was initially ).
I agree this is not very DRY, but helps do a quicker review, to get the point across.

ashmaroli:
updated this PR to be a standalone pull instead of an extension of another PR ( as was initially ).
I agree this is not very DRY, but helps do a quicker review, to get the point across.

ashmaroli:
@parkr, is there even a remote possibility that this PR may be considered before v3.3?

parkr:
@ashmaroli I don't think this will be able to make it into v3.3.

ashmaroli:
closing.. as this has been moved to another gem [`jekyll-plus`](https://github.com/ashmaroli/jekyll-plus) entirely

PR #5220, opened 2016-08-10
update appveyor.yml

ashmaroli:
Updating appveyor.yml to [support latest releases of Ruby vers](https://github.com/appveyor/website/blob/4cf4be275b18140b86ebfcadc4b8cceed851e769/NJekyll/site/_updates/2016-08-09.md) 
(... and experimenting with Build Cache)

\cc @XhmikosR

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,32 +12,32 @@ branches:
 build: off
 
 install:
-  - SET PATH=C:\Ruby%RUBY_FOLDER_VER%\bin;%PATH%
-  - bundle install --retry 5
+  - SET PATH=C:\Ruby%RUBY_SERIES%\bin;%PATH%
+  - bundle install --retry 3
 
 environment:
   BUNDLE_WITHOUT: "benchmark:site:development"
   matrix:
-    - RUBY_FOLDER_VER: "23"
-      GEMS_FOLDER_VER: "2.3.0"
+    - RUBY_SERIES: "23"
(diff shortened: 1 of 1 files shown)


XhmikosR:
I don't see the need for those changes.

ashmaroli:
A big :-1: ?
 :sob: 

oh well.. :walking: :broken_heart:

XhmikosR:
Because the changes are very bad.

1. The gems path is the same so your change broke that
2. The cache changes will prevent invalidating it

I hope you see now how useless and wrong this PR is.

ashmaroli:
I admit that some changes in this PR are unnecessary.
I changed the folder names coz I didnt like seeing `RUBY_FOLDER..` `GEMS_FOLDER..` in the logs.
I changed the cache to see if clearing the `-> <file>` would cause installed gems to be cached for consequent tests..
While I was on that, I saw that Appveyor had pushed support for latest rubies, n bundler ( both of which were initiated by you ). So I thought, y not try updating them here as well..

Only after your latest comment, do I even realize that the changes were wrong. My apologies.
Thank you for the clarification, though.

envygeeks:
> Because the changes are very bad.
> The gems path is the same so your change broke that
> The cache changes will prevent invalidating it
> I hope you see now how useless and wrong this PR is.

These kind of comments are unacceptable, please be more respectful when declining an idea.

PR #5223, opened 2016-08-10
remove redundant `private` declarations

karloescota:
remove redundant `private` declarations

### DIFF
--- a/lib/jekyll/collection.rb
+++ b/lib/jekyll/collection.rb
@@ -207,7 +207,6 @@ module Jekyll
       end
     end
 
-    private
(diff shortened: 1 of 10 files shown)


envygeeks:
You call them redundant, I call them necessary when you audit code as much as I do and want things to be very explicit so there are no mistakes made.  No thanks on this.

parkr:
Yeah, I'm going to agree with @envygeeks here. 😄 They _are_ redundant, but they help make the code a bit more clear for folks skimming it, knowing what's public and what's private. Thank you though!!

PR #5229, opened 2016-08-11
Fix grid-width for large screens

ashmaroli:
Fix stylesheet to expand the grid wrap a bit further for large screens.

### DIFF
--- a/site/_includes/docs_contents.html
+++ b/site/_includes/docs_contents.html
@@ -1,4 +1,4 @@
-<div class="unit one-fifth hide-on-mobiles">
+<div class="unit one-fourth hide-on-mobiles">
   <aside>
     {% for section in site.data.docs %}
     <h4>{{ section.title }}</h4>
--- a/site/_includes/news_contents.html
+++ b/site/_includes/news_contents.html
@@ -1,4 +1,4 @@
-<div class="unit one-fifth hide-on-mobiles">
+<div class="unit one-fourth hide-on-mobiles">
   <aside>
     <ul>
       <li class="{% if page.title == 'News' %}current{% endif %}">
--- a/site/_layouts/docs.html
+++ b/site/_layouts/docs.html
@@ -7,7 +7,7 @@ layout: default
 
       {% include docs_contents_mobile.html %}
 
-      <div class="unit four-fifths">
+      <div class="unit three-fourths">
         <article>
           <div class="improve right hide-on-mobiles">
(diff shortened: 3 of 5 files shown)


parkr:
Can you please provide screenshots? This seems to be a change to a fundamental part of this grid system.

ashmaroli:
### before change. (current live):
before

**Browser:** Google Chrome
**Resolution:** 1366 x 768px

---

### after:
extending grid-width to a max 1180px.
(simulating browsers that do not work with CSS calc())
  


**Browser:** Google Chrome
**Resolution:** 1366 x 768px

---

### after (2):
extending grid-width further.
(use calc() to fluidly alter max-width to 60px less than max-screen-width)
[reference for calc() support](http://caniuse.com/#search=calc)

calc-width

**Browser:** Google Chrome
**Resolution:** 1366 x 768px

envygeeks:
I'm on the fence, on one aspect I like this, on the other, I like consistent font widths.

envygeeks:
But AFAIC I'm :+1: on this if others like it.

DirtyF:
I'm  👎  because, the line length was designed with optimized readability in mind, and this just break it without any other reason than occupy the whole space.

Maybe @jglovier could give some insight on how scale the design on larger screen width?

nhoizey:
I agree with @DirtyF, if the lines are to long, it hurts readability.
Common best practices put 50 to 70 characters per line.

Another way to use large viewports would be to put the main menu vertical on the left side.

ashmaroli:
New PR - 5243 adding in more changes..

ashmaroli:
Amended to remove increased line-length issue with initial commit.
The amend involves changing grid from using five-columns to four-columns (more breathing space for aside element)  and confining the section element to a max-width of 1024px

mattr-:
Seems like it would be good. Could we get new before and after screenshots one more time?

ashmaroli:
## Current Live: HEADER AREA
Current

## PR-branch served by Jekyll: HEADER AREA
Header-New

---

## Current Live: FOOTER AREA & ASIDE
Current

## PR-branch served by Jekyll: FOOTER AREA & ASIDE
Footer-Aside

mattr-:
Now that I'm seeing the screenshots, I'm not a fan of this one. It breaks the concept that the whole site has one width that it fits in.

As far as larger screens go, the site works fine in larger screens right now. I'd rather not break the design to just to make sure we take the full width of the screen. Closing this one out.

jglovier:
Sorry for the delayed response, but I agree with @mattr- on keeping the existing structure for the same reasons he cited.

Issue #4891, opened 2016-05-13
Release Jekyll v3.2


parkr:
This is a checklist for releasing v3.2.0.

- [x] Documentation for gem-based themes
- [x] Write `jekyll new-theme` command ( #4848 )
- [x] Port changes from #4911 forward to `master` branch.
- [x] Fix `jekyll new-theme` and theme template based on feedback ( #5042 )
- [x] Convert Jekyll-new theme to a gem-based theme ( #4922 )
- [x] Cull list of issues in the milestone & punt where possible
- [x] Test Jekyll against Pathutil to find breakage.
- [x] ThemeBuilder: Remove example site from theme (#5141)
- [x] Release 1.0 of `minima` (https://github.com/jekyll/minima/issues/1)
- [x] Publish release post ( #5140 )
- [x] Release `jekyll` gem
- [x] Release `jekyll-docs` for v3.2.0

/cc @jekyll/stability @jekyll/build @jekyll/ecosystem

envygeeks:
Add `Pathutil#cleanpath` to that list because we have `Pathutil` in there now and I need to make sure it's completely backwards compatible.

benbalter:
> Documentation for gem-based themes

https://jekyllrb.com/docs/themes/

parkr:
> https://jekyllrb.com/docs/themes/

HOW DID THAT GET THERE?!

envygeeks:
@parkr @benbalter that's why I commented that the release of the documentation is part of a much larger issue, every time we update the site the beta documentation lands on there indiscriminately and it's led to quite a few people asking questions about unreleased features and being confused.

benbalter:
> HOW DID THAT GET THERE?! 

[I wrote them!](https://github.com/jekyll/jekyll/issues/4595)

chrisfinazzo:
In cases like this, would it be preferable to ask that changes to documentation land in a separate branch? Currently, everything goes straight to master and sometimes things aren't ready. Whether it gets tagged as beta, or something like `{{feature}}-docs-vx` where `vx` is the anticipated milestone for when it should be integrated.

This is probably too formal, curious to hear what @jekyll/documentation thinks.

envygeeks:
[...]

PR #5230, opened 2016-08-11
bump rouge to 2.0.x

Crunch09:
Hi,

this bumps `rouge` to 2.0.x as discussed in https://github.com/jekyll/jekyll/issues/4891#issuecomment-234981633 .

I used the `HTMLLegacy` formatter to avoid breaking changes but there are some minor ones as you can see in the tests (no `text-align: right;` on the `td` of the line number for example). If this is not okay,  i think we would have to write our own formatter.

### DIFF
--- a/jekyll.gemspec
+++ b/jekyll.gemspec
@@ -34,7 +34,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('mercenary', '~> 0.3.3')
   s.add_runtime_dependency('safe_yaml', '~> 1.0')
   s.add_runtime_dependency('colorator', '~> 1.0')
-  s.add_runtime_dependency('rouge', '~> 1.7')
+  s.add_runtime_dependency('rouge', '~> 2.0.5')
   s.add_runtime_dependency('jekyll-sass-converter', '~> 1.0')
   s.add_runtime_dependency('jekyll-watch', '~> 1.1')
   s.add_runtime_dependency("pathutil", "~> 0.9")
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -30,6 +30,7 @@ require "safe_yaml/load"
 require "liquid"
 require "kramdown"
 require "colorator"
+require "rouge"
 
 SafeYAML::OPTIONS[:suppress_warnings] = true
 
--- a/lib/jekyll/converters/markdown/redcarpet_parser.rb
+++ b/lib/jekyll/converters/markdown/redcarpet_parser.rb
@@ -55,7 +55,7 @@ class Jekyll::Converters::Markdown::RedcarpetParser
 
     protected
     def rouge_formatter(_lexer)
-      Rouge::Formatters::HTML.new(:wrap => false)
+      Rouge::Formatters::HTMLLegacy.new(:wrap => false)
     end
   end
 
--- a/lib/jekyll/tags/highlight.rb
+++ b/lib/jekyll/tags/highlight.rb
@@ -110,9 +110,12 @@ eos
 
       def render_rouge(code)
         Jekyll::External.require_with_graceful_fail("rouge")
-        formatter = Rouge::Formatters::HTML.new(
+        formatter = Rouge::Formatters::HTMLLegacy.new(
           :line_numbers => @highlight_options[:linenos],
-          :wrap         => false
+          :wrap         => false,
+          :css_class    => "highlight",
+          :gutter_class => "gutter",
+          :code_class   => "code"
         )
         lexer = Rouge::Lexer.find_fancy(@lang, code) || Rouge::Lexers::PlainText
         formatter.format(lexer.lex(code))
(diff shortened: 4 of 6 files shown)


jekyll.gemspec
```diff
@@ -34,7 +34,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('mercenary', '~> 0.3.3')
   s.add_runtime_dependency('safe_yaml', '~> 1.0')
   s.add_runtime_dependency('colorator', '~> 1.0')
-  s.add_runtime_dependency('rouge', '~> 1.7')
+  s.add_runtime_dependency('rouge', '~> 2.0.5')
```
Crunch09: Not sure here if we should choose `~> 2.0.5` or `~> 2.0`

envygeeks:
I'm sorry but this pull in it's current form is unacceptable for 3.x.  You ***must*** allow 1.x (Rouge) to remain and work around Rouges changes in order for this to be accepted in 3.x as it's a breaking change that affects people, plugins and other stuff.  This is what I had tried to imply was the problem on talk.jekyllrb.com, it's just too much work, without a wrapper of which we should have built long ago.

envygeeks:
Either way, if you are willing to fix the problem so that both 1.x and 2.x work then it can be accepted in 3.x but until then it has a :-1: from me, because I do not wish to deal with upgrading the hundreds of plugins we have on our own sites that use Rouge 1.x in some way before absolutely necessary and even if I do, I do not wish to have to do it overnight because Jekyll decided to randomly upgrade a major dependency.

jekyll.gemspec
envygeeks: `s.add_runtime_dependency(">= 1.7", "< 3")` == "1.x", "2.x"

parkr:
@envygeeks Point taken – I proposed we do this for 3.3 after it was shot down for 3.2 due to timing (https://github.com/jekyll/jekyll/issues/4891#issuecomment-234810772). I'm 👍 on doing this if it's backwards-compatible. Could we check `Rouge::VERSION`?

Crunch09:
Thanks for the feedback! Will try to make it work with rouge 1.x

Crunch09:
Making some good progress on this!
But i am not sure which [Formatter](https://github.com/jneen/rouge#formatters) we should choose for 2.x ? I guess we don't need to use the `Rouge::Formatters::HTMLLegacy` formatter because users who want this behaviour are still free to use Rouge `~> 1.7`.

pathawks:
Fixes #5556

jneen:
Hi! What I recommend doing is actually defining your own formatter, especially if there are special things you want it to do. You can look through the existing formatters as well as the [Gitlab Formatter](https://gitlab.com/gitlab-org/gitlab-ce/blob/master/lib/rouge/formatters/html_gitlab.rb). Alternately, you can use `Formatters::HTML` to generate plain unadorned HTML and add stuff around it as you like. Looking at the code you have, you might also consider using `Formatters::HTMLTable`, which takes an inner formatter (`Formatters::HTML` in your case) and a number of line numbering options.

jneen:
Ah yes, and if you want to support both 1.x and 2.x, you'll have to define a wrapper, as the APIs are different.

Crunch09:
@jneen Thank you!

fliiiix:
@Crunch09 are you working on it?

Crunch09:
@fliiiix Yes. Sorry for the delay, it's been a busy couple of weeks but there should be an update by the end of the weekend.

Crunch09:
@jekyll/core Unfortunately i still have some open questions regarding this and i hope you can help to clear them up so i can push an update:
- If we write our own formatter, should it produce the exact same output as now with Rouge 1.7? If not, what should / is allowed to change compared to the current version? I guess if we were to change some things this wouldn't be a breaking change for users as they still could use the old Rouge version.
- If we want to land this in jekyll 3.4 and support both Rouge versions should there be different CI builds to test against each version?

parkr:
Fixes https://github.com/jekyll/jekyll/issues/5774.

parkr:
Fixes #5774.

willclarktech:
Hi, is there any movement on this? Currently code highlighting of ES6 template literals are broken on GitHub Pages sites because rouge 1.x doesn’t support them.

jneen:
Also, I'm not clear on the incompatibility story for users who wrote their own formatters - if they wrote their own, how are they affected by the existing ones changing?

pathawks:
> If we write our own formatter, should it produce the exact same output as now with Rouge 1.7? If not, what should / is allowed to change compared to the current version? I guess if we were to change some things this wouldn't be a breaking change for users as they still could use the old Rouge version.

If a user has Rouge 1.x installed, the output should be unchanged. If a user has Rouge 2.x installed, they should be aware that the output might be different.

> If we want to land this in jekyll 3.4 and support both Rouge versions should there be different CI builds to test against each version?

I don't think we need to double the size of our matrix, but we should have at least one that tests with the old Rouge and at least one that tests with the new. Probably only have one that tests against Rouge 1.x and move everything else to 2.x?

Crunch09:
@pathawks Thank you for the clarification, that should help!

guymorita:
Can we get this merged in? Would help lots. Thanks.

Crunch09:
@guymorita Hi, please see #6150 for updates.

PR #5232, opened 2016-08-11
Clarify language around installation requirements

stkent:


envygeeks:
If you want my opinion on this matter, drop any mention of Jekyll 2.x we no longer support it at all, in any form, Jekyll Pages is on 3.x and quickly updates now, and we've no desire to issue any fixes for 2.x and those still on it should update.

envygeeks:
AFAIW we also still require NodeJS, MiniRacer or RubyRacer for CoffeeScript too, even in 3.x yes @jekyll/plugins? I just think the only difference here is it uses ExecJS for flexibility now yes?

parkr:
> AFAIW we also still require NodeJS, MiniRacer or RubyRacer for CoffeeScript too, even in 3.x yes @jekyll/plugins? I just think the only difference here is it uses ExecJS for flexibility now yes?

Yep, jekyll-coffeescript requires a JavaScript runtime, so either therubyracer or a Node.js installation. It's all baked into the `coffeescript` gem which `jekyll-coffeescript` depends on.

stkent:
This was just meant to be a tweak of existing language; didn't realize the content was out of date! I'll close this PR and let someone (much) more familiar with the current dependencies take a stab at it.

PR #5235, opened 2016-08-12
Fix #5233: Increase our ability to detect Windows.

### LINKED ISSUE #5233: jekyll serve not working on Windows Subsystem for Linux

---

---

- [x] It was another bug.

## My Reproduction Steps

```
jekyll serve
Configuration file: /path/to/_config.yml
            Source: /path/to
       Destination: /path/to/_site
      Generating...
                    done in 0.062 seconds.
jekyll 3.1.6 | Error:  Invalid argument - Failed to watch "/path/to/.git/hooks": the given event mask contains no legal events; or fd is not an inotify file descriptor.
```

This is almost certainly due to https://github.com/Microsoft/BashOnWindows/issues/216. I'm not sure if there's any viable workaround at the moment, but figured (1) it was worth bringing this to the jekyll team's attention, and (2) having a public issue for other users who run into this problem.

envygeeks:
This increases our ability to detect Windows, and to detect Windows+Bash.  It also adds a message to Windows for users who try to "--watch", also noting to to them to check out the Windows ticket so eventually somebody pings us if this issue is fixed. /cc @TAGraves

/cc @jekyll/windows @jekyll/core @parkr @benbalter @jekyll/build

### DIFF
--- a/lib/jekyll/commands/build.rb
+++ b/lib/jekyll/commands/build.rb
@@ -71,12 +71,23 @@ module Jekyll
         #
         # Returns nothing.
         def watch(site, options)
-          External.require_with_graceful_fail "jekyll-watch"
-          watch_method = Jekyll::Watcher.method(:watch)
-          if watch_method.parameters.size == 1
-            watch_method.call(options)
+          if Utils::Platforms.windows?
+            Jekyll.logger.warn "", "--watch arg is unsupported on Windows. "
+            Jekyll.logger.warn "", "If you are on Windows Bash, please see: " \
+              "https://github.com/Microsoft/BashOnWindows/issues/216"
+
           else
-            watch_method.call(options, site)
+            External.require_with_graceful_fail "jekyll-watch"
+            watch_method = Jekyll::Watcher.method(:watch)
+            if watch_method.parameters.size == 1
+              watch_method.call(
+                options
+              )
+            else
+              watch_method.call(
+                options, site
+              )
+            end
           end
         end
       end # end of class << self
--- a/lib/jekyll/utils/platforms.rb
+++ b/lib/jekyll/utils/platforms.rb
@@ -13,16 +13,53 @@ module Jekyll
         end
       end
 
+      # --
+      # Allows you to detect "real" Windows, or what we would consider
+      # "real" Windows.  That is, that we can pass the basic test and the
+      # /proc/version returns nothing to us.
+      # --
+
+      def really_windows?
+        RbConfig::CONFIG["host_os"] =~ %r!mswin|mingw|cygwin!i && \
+          !proc_version
+      end
+
+      #
+
+      def windows?
+        RbConfig::CONFIG["host_os"] =~ %r!mswin|mingw|cygwin!i || \
+          proc_version =~ %r!microsoft!i
+      end
+
+      #
+
+      def linux?
+        RbConfig::CONFIG["host_os"] =~ %r!linux! && \
+          proc_version !~ %r!microsoft!i
+      end
+
       # Provides windows?, linux?, osx?, unix? so that we can detect
       # platforms. This is mostly useful for `jekyll doctor` and for testing
       # where we kick off certain tests based on the platform.
 
-      { :windows? => %r!mswin|mingw|cygwin!, :linux? => %r!linux!, \
-          :osx? => %r!darwin|mac os!, :unix? => %r!solaris|bsd! }.each do |k, v|
+      { :osx? => %r!darwin|mac os!, :unix? => %r!solaris|bsd! }.each do |k, v|
         define_method k do
           !!(
             RbConfig::CONFIG["host_os"] =~ v
           )
+        end
+      end
+
+      #
+
+      private
+      def proc_version
+        @cached_proc_version ||= begin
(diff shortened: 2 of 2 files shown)


lib/jekyll/utils/platforms.rb
```diff
@@ -13,18 +13,40 @@ module Platforms
...
+      def windows?
+        !!(RbConfig::CONFIG["host_os"] =~ %r!mswin|mingw|cygwin!i || \
+          proc_version =~ %r!microsoft!i)
+      end
+
+      #
+
+      def linux?
+        RbConfig::CONFIG["host_host"] =~ %r!linux! && \
+          !proc_version =~ %r!microsoft!i
```
parkr: I might be remembering Lua, but is this `!~`?

envygeeks: You're right, I was in a hurry and doing it on my tablet, I'll ship a correction.

parkr:
Love it!!!!!

lib/jekyll/utils/platforms.rb
envygeeks: Fixed.

lib/jekyll/utils/platforms.rb
```diff
@@ -13,18 +13,44 @@ module Platforms
         end
       end
 
+      #
+
+      def windows?
+        !!(RbConfig::CONFIG["host_os"] =~ %r!mswin|mingw|cygwin!i || \
```
XhmikosR: How about `x64_mingw`? Does this work right there?

spudowiar: Looks to me like the regex isn't anchored so it'd match `x64_mingw`

lib/jekyll/commands/build.rb
```diff
@@ -71,12 +71,23 @@ def build(site, options)
         #
         # Returns nothing.
         def watch(site, options)
-          External.require_with_graceful_fail "jekyll-watch"
-          watch_method = Jekyll::Watcher.method(:watch)
-          if watch_method.parameters.size == 1
-            watch_method.call(options)
+          if Utils::Platforms.windows?
+            Jekyll.logger.warn "", "--watch unsupported on Windows. "
```
benbalter: Maybe "The --watch flag is unsupported..." for a bit of polish.

cori: Date: Fri, 12 Aug 2016 07:28:39 -0700

lib/jekyll/utils/platforms.rb
envygeeks: @XhmikosR @spudowiar we don't care if it's x64, as @spudowiar pointed out, that's why it's unanchored, we are only interested if we find a specific keyword.  There could be dozens of builds.

XhmikosR: As long as it detects `x64_mingw` as Windows, that's fine.

lib/jekyll/commands/build.rb
envygeeks: Done.

envygeeks:
Updated to add `#really_windows?` so that @ashmaroli can skip certain tests if it's actually Windows.

lib/jekyll/utils/platforms.rb
```diff
@@ -13,18 +13,55 @@ module Platforms
...
             RbConfig::CONFIG["host_os"] =~ v
           )
         end
       end
+
+      #
+
+      private
+      def proc_version
+        @cacahed_proc_version ||= begin
```
ashmaroli: did you intend it to be @cached_proc_version ?

envygeeks: Fixed.

parkr:
LGTM. Any objections from @jekyll/Windows?

envygeeks:
LGTM.

envygeeks:
From our own Windows users this works for them, so it has a LGTM.

parkr:
@jekyllbot: merge +minor

ashmaroli:
@envygeeks, is there a way to detect Windows version or detect if I'm on the default Windows `cmd.exe` instead of bash shell? I miss `autoregeneration` now..
It works properly when using Jekyll 3.2.1 in `cmd.exe` on Windows 7.

envygeeks:
@ashmaroli we do detect tell the difference between BashOnWindows and Windows itself in cmd.exe/PowerShell, the problem is probably in my detector which is a bit extreme in it's handling, it excludes the entirety of Windows even if there is the win32 extension that works with our watchers.

Later today after I get done with some stuff I'll ship a pull that adds `bash_on_windows?` and adjusts https://github.com/jekyll/jekyll/pull/5235/files#diff-47014fa0275ac66de0d2217d362529ffR74 to use that method so that watchers still work on normal Windows in cmd/PowerShell.

ashmaroli:
Awesome!! Thanks :smiley:

parkr:
@envygeeks Planning to release 3.3 tomorrow so if you want that change in 3.3, we'll need the pull today. If not, then we can release later.

envygeeks:
I'll push it sometime this evening after I get done working.

PR #5236, opened 2016-08-12
Enhance new command to include argument passed, as blog-title

ashmaroli:
This PR deals with the idea of automatically titling a newly generated blog with the argument passed to `jekyll new`

illustrating with an example on windows platform:

```
$ jekyll new Exploring Ruby

#: New jekyll site 'Exploring Ruby' installed in D:/Projects/Exploring Ruby.
```

Add Blog-Title

served-view

ref: #5203 
/cc @jekyll/ecosystem

### DIFF
--- a/lib/jekyll/commands/new.rb
+++ b/lib/jekyll/commands/new.rb
@@ -55,6 +55,7 @@ module Jekyll
 
         def create_config_file(title, path)
           @blog_title = title
+          @user_email = user_email
 
           config_template = File.expand_path("_config.yml.erb", site_template)
           config_copy = ERB.new(File.read(config_template)).result(binding)
@@ -159,6 +160,11 @@ RUBY
           Dir.chdir(path) do
             system("bundle", "install")
           end
+        end
+
+        def user_email
+          gitconfig_email = `git config user.email`.chomp
+          gitconfig_email.empty? ? "your-email@domain.com" : gitconfig_email
         end
       end
     end
--- a/lib/jekyll/commands/new.rb
+++ b/lib/jekyll/commands/new.rb
@@ -22,8 +22,11 @@ module Jekyll
         def process(args, options = {})
           raise ArgumentError, "You must specify a path." if args.empty?
 
+          new_blog_title = args.join(" ")
           new_blog_path = File.expand_path(args.join(" "), Dir.pwd)
+
           FileUtils.mkdir_p new_blog_path
+
           if preserve_source_location?(new_blog_path, options)
             Jekyll.logger.abort_with "Conflict:",
                       "#{new_blog_path} exists and is not empty."
@@ -33,9 +36,10 @@ module Jekyll
             create_blank_site new_blog_path
           else
             create_site new_blog_path
+            create_config_file(new_blog_title, new_blog_path)
           end
 
-          after_install(new_blog_path, options)
+          after_install(new_blog_title, new_blog_path, options)
         end
 
         def create_blank_site(path)
@@ -47,6 +51,17 @@ module Jekyll
 
         def scaffold_post_content
           ERB.new(File.read(File.expand_path(scaffold_path, site_template))).result
+        end
+
+        def create_config_file(title, path)
+          @blog_title = title
+
+          config_template = File.expand_path("_config.yml.erb", site_template)
+          config_copy = ERB.new(File.read(config_template)).result(binding)
+
+          File.open(File.expand_path("_config.yml", path), "w") do |f|
+            f.write(config_copy)
+          end
         end
 
         # Internal: Gets the filename of the sample post to be created
@@ -103,9 +118,18 @@ RUBY
           !options["force"] && !Dir["#{path}/**/*"].empty?
         end
 
+        def erb_files
+          erb_file = File.join("*", "*.erb")
+          Dir.glob(erb_file)
+        end
+
         def create_sample_files(path)
           FileUtils.cp_r site_template + "/.", path
           FileUtils.rm File.expand_path(scaffold_path, path)
+
+          erb_files.each do |file|
+            FileUtils.rm file
+          end
         end
 
         def site_template
@@ -120,8 +144,8 @@ RUBY
         # then automatically execute bundle install from within the new blog dir
         # unless the user opts to generate a blank blog or skip 'bundle install'.
 
-        def after_install(path, options = {})
-          Jekyll.logger.info "New jekyll site installed in #{path.cyan}."
+        def after_install(title, path, options = {})
+          Jekyll.logger.info "New jekyll site #{title.cyan} installed in #{path.cyan}."
           Jekyll.logger.info "Bundle install skipped." if options["skip-bundle"]
 
           unless options["blank"] || options["skip-bundle"]
--- a/lib/jekyll/commands/new.rb
+++ b/lib/jekyll/commands/new.rb
@@ -7,7 +7,8 @@ module Jekyll
         def init_with_program(prog)
           prog.command(:new) do |c|
             c.syntax "new PATH"
-            c.description "Creates a new Jekyll site scaffold in PATH"
+            c.description "Creates a new Jekyll site scaffold in PATH, and " \
+                        "automatically assigns site's title based on argument passed."
 
             c.option "force", "--force", "Force creation even if PATH already exists"
             c.option "blank", "--blank", "Creates scaffolding but with empty files"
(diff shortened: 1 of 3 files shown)


ashmaroli:
travis failing only on Ruby v2.1.9 .. unrelated bug elsewhere?

envygeeks:
All is well, sometimes it happens, I restarted the test for you and it passed.

ashmaroli:
Now supports fetching title from path passed as argument

Support for path

ashmaroli:
would like more :eyes: on this, pls
/cc @jekyll/ecosystem

lib/jekyll/commands/new.rb
```diff
@@ -7,7 +7,8 @@ class << self
         def init_with_program(prog)
           prog.command(:new) do |c|
             c.syntax "new PATH"
-            c.description "Creates a new Jekyll site scaffold in PATH"
+            c.description "Creates a new Jekyll site scaffold in PATH, and\n\t\t\t" \
```
pathawks: Do we really need three tabs here? Is this done anywhere else in Jekyll?

lib/jekyll/commands/new.rb
```diff
@@ -31,10 +35,11 @@ def process(args, options = {})
           if options["blank"]
             create_blank_site new_blog_path
           else
-            create_site new_blog_path
+            create_site new_blog_title, new_blog_path
```
pathawks: Note to self: `create_site` is `private`, so changing its signature is no problem :+1:

lib/jekyll/commands/new.rb
```diff
@@ -48,6 +53,18 @@ def scaffold_post_content
           ERB.new(File.read(File.expand_path(scaffold_path, site_template))).result
         end
 
+        def create_config_file(title, path)
+          @blog_title = title
+          @user_email = user_email
```
pathawks: Where is `user_email` defined?

lib/jekyll/commands/new.rb
```diff
@@ -114,6 +145,11 @@ def site_template
         def scaffold_path
           "_posts/0000-00-00-welcome-to-jekyll.markdown.erb"
         end
+
+        def user_email
```
pathawks: This is [similar to `Jekyll::ThemeBuilder.user_email`](https://github.com/jekyll/jekyll/blob/f4c3c5afd6f7d051934eaf67d60ebe413b7214a8/lib/jekyll/theme_builder.rb#L90-L92). Is there any way to DRY up?

test/test_new_command.rb
```diff
@@ -89,9 +100,11 @@ def site_template
...
 
     should "force created folder" do
-      capture_stdout { Jekyll::Commands::New.process(@args) }
-      output = capture_stdout { Jekyll::Commands::New.process(@args, "--force") }
-      assert_match(%r!New jekyll site installed in!, output)
+      capture_stdout { Jekyll::Commands::New.process(@args, "--force") }
+      output = Jekyll.logger.messages.last
+      success_message = "New jekyll site #{@title.cyan} " \
+      "installed in #{@full_path.cyan}."
+      assert_includes output, success_message
```
pathawks: I am uncomfortable changing from `assert_match` to `assert_includes`.

lib/jekyll/commands/new.rb
ashmaroli: Its not done anywhere else in Jekyll. I added 3 tabs to keep it in-line with the first line. I agree, its a bit ugly.

lib/jekyll/commands/new.rb
ashmaroli: Yes, I added this when I saw that block in theme_builder.rb
But they're both used differently. So DRYing it would be another PR.

pathawks: > But they're both used differently. So DRYing it would be another PR.

I'm not sure I understand. Can you elaborate?

test/test_new_command.rb
ashmaroli: The output would be displayed only is the script executed successfully upto that point. So, in theory, it shouldn't matter. `assert_includes` is more tolerant of the ANSI escape codes present in the output.
The context for default path also [checks](https://github.com/jekyll/jekyll/blob/master/test/test_new_command.rb#L45) via an `assert_includes`.

ashmaroli:
Thank you for the comments, @pathawks :smiley:

test/test_new_command.rb
pathawks: Since we are modifying an existing test here, rather than creating a new one, it makes me uncomfortable to go from `assert_match` to `assert_includes`, but I guess the old test wasn't even checking for the path, so it's not as if the new test is any more lenient.

:+1:

lib/jekyll/commands/new.rb
ashmaroli: In this PR, I'm assigning user_email to an instance variable to be output by the ERB file.
In theme_builder, user_email is defined as a symbol which is forwarded to another subclass ERBRenderer. Since that symbol needn't be in the memory at the instant `jekyll new` is called, i decided to add a separate private method to this command.

pathawks: You're right. We _should_ DRY up, but it should be in a separate PR.

pathawks:
LGTM

lib/jekyll/commands/new.rb
```diff
@@ -21,8 +22,11 @@ def init_with_program(prog)
         def process(args, options = {})
           raise ArgumentError, "You must specify a path." if args.empty?
 
+          new_blog_title = sanitized args
```
mattr-: `sanitized` is new and is used in only one place. I would prefer to see `new_blog_title` become a method that does the extraction on the args, so that I don't have to wonder what's going on here. Instead, I just see this (from further down):

`create_site new_blog_title(args), new_blog_path`

which is much clearer, IMHO.

`new_blog_title` then becomes:

```ruby
def new_blog_title(args)
    a = args.join(" ")
    a.tr("\\", "/").split("/").last
end
```

lib/jekyll/commands/new.rb
```diff
@@ -57,6 +74,12 @@ def initialized_post_name
 
         private
 
+        def sanitized(args)
```
mattr-: I'm concerned that you're not using things from the Ruby standard library here and are instead using string manipulation. Is there something that prevents you from using `File`, `Dir`, `Pathname`, or `FileUtils`?

mattr-:
I am really uncomfortable with changing the site template to use ERB files that then need to be processed to make a template site. This for two reasons:

1. Introducing ERB to the site_template complicates the process of generating the new site from the template.
2. More importantly, by doing this, we've started generating more content for the user than the minimum necessary to get a user going. Jekyll is not Rails and it's ok for us to do things differently from them. It's a template, let's have it be a full on template and make it completely obvious that it's the user's job to put their own content in there. Jekyll does not and should not have any control over the user's content outside of generating the scaffolding necessary to get them going or to keep them going. This means we shouldn't be generating actual content for the user unless it's useful as part of the scaffolding and meant to be removed/changed, like the site template pages are here.

I'm :+1: on the ability to put the new site in a different directory/folder from the current directory/folder. I'm :-1: on doing any further customization to the template.

lib/jekyll/commands/new.rb
ashmaroli: the reason i delegated extraction to a standalone method is because it got too complicated (for me) further ahead.

by the time the following queued up for processing, `new_blog_title` was already defined.
```ruby
def create_site(new_blog_title, new_blog_path)
  create_sample_files new_blog_title, new_blog_path
  [...]
end
```

But if were to have `new_blog_title` itself be a method, then (with all due respect), things go beyond my newbie-skill-level. Sorry.
The max I can do to this (as of this writing) is to rename `sanitized` to `extract_title`

lib/jekyll/commands/new.rb
ashmaroli: The reason is, because I **felt** this method to be faster. But if your expertise says its wrong, I'll look at standard library.

ashmaroli:
Thanks for your feedback, @mattr-.
The main purpose behind this PR was to have the blog-title automatically defined from the arguments supplied. But since U feel it goes against Jekyll's simplicity, ... 

You're welcome to close this.

ashmaroli:
closing.. as this has been moved to another gem [`jekyll-plus`](https://github.com/ashmaroli/jekyll-plus) entirely

PR #5225, opened 2016-08-10
Add bundle install to jekyll new command

ashmaroli:
This PR - an extension of PR #5203, proposes to extend the `jekyll new` command with following:
- [x] Automatically `cd my_blog` and initiate an interactive prompt to run `bundle install` 

> After you run `jekyll new my_blog` the command shell automatically `cd` into the new directory and initiates a user prompt whether to run bundle install right away. 
>If bundler is not installed, an exception is raised gracefully.
> The user may choose to skip 'bundle install' if they plan to edit their new gemfile and change the theme gem, or if they want to inspect the new directory before running the `bundle install` command

PR #5237, opened 2016-08-12
Add bundle install to jekyll new command

ashmaroli:
This PR proposes to extend the `jekyll new` command with following:
- Automatically `cd my_blog` and initiate `bundle install` 

> After you run `jekyll new my_blog` the command shell automatically `cd` into the new directory and runs `bundle install`.
> If bundler is not installed, an exception is raised gracefully.
> The user may choose to skip 'bundle install' if they plan to edit their new gemfile and change the theme gem, or if they want to inspect the new directory before running the `bundle install` command by using the new switch `--skip-bundle`

ref: #5225

### DIFF
--- a/lib/jekyll/commands/new.rb
+++ b/lib/jekyll/commands/new.rb
@@ -11,6 +11,7 @@ module Jekyll
 
             c.option "force", "--force", "Force creation even if PATH already exists"
             c.option "blank", "--blank", "Creates scaffolding but with empty files"
+            c.option "skip-bundle", "--skip-bundle", "Skip 'bundle install'"
 
             c.action do |args, options|
               Jekyll::Commands::New.process(args, options)
@@ -34,7 +35,7 @@ module Jekyll
             create_site new_blog_path
           end
 
-          Jekyll.logger.info "New jekyll site installed in #{new_blog_path.cyan}."
+          after_install(new_blog_path, options)
         end
 
         def create_blank_site(path)
@@ -113,6 +114,27 @@ RUBY
 
         def scaffold_path
           "_posts/0000-00-00-welcome-to-jekyll.markdown.erb"
+        end
+
+        # After a new blog has been created, print a success notification and
+        # then automatically execute bundle install from within the new blog dir
+        # unless the user opts to generate a blank blog or skip 'bundle install'.
+
+        def after_install(path, options = {})
+          Jekyll.logger.info "New jekyll site installed in #{path.cyan}."
+          Jekyll.logger.info "Bundle install skipped." if options["skip-bundle"]
+
+          unless options["blank"] || options["skip-bundle"]
+            bundle_install path
+          end
+        end
+
+        def bundle_install(path)
+          Jekyll::External.require_with_graceful_fail "bundler"
+          Jekyll.logger.info "Running bundle install in #{path.cyan}..."
+          Dir.chdir(path) do
+            system("bundle", "install")
+          end
         end
       end
     end
--- a/test/test_new_command.rb
+++ b/test/test_new_command.rb
@@ -40,9 +40,12 @@ class TestNewCommand < JekyllUnitTest
 
     should "display a success message" do
       Jekyll::Commands::New.process(@args)
-      output = Jekyll.logger.messages.last
+      output = Jekyll.logger.messages[-3]
+      output_last = Jekyll.logger.messages.last
       success_message = "New jekyll site installed in #{@full_path.cyan}."
+      bundle_message = "Running bundle install in #{@full_path.cyan}..."
       assert_includes output, success_message
+      assert_includes output_last, bundle_message
     end
 
     should "copy the static files in site template to the new directory" do
@@ -85,13 +88,23 @@ class TestNewCommand < JekyllUnitTest
     should "create blank project" do
       blank_contents = %w(/_drafts /_layouts /_posts /index.html)
       capture_stdout { Jekyll::Commands::New.process(@args, "--blank") }
+      output = Jekyll.logger.messages.last
+      bundle_message = "Running bundle install in #{@full_path.cyan}..."
       assert_same_elements blank_contents, dir_contents(@full_path)
+      refute_includes output, bundle_message
     end
 
     should "force created folder" do
       capture_stdout { Jekyll::Commands::New.process(@args) }
       output = capture_stdout { Jekyll::Commands::New.process(@args, "--force") }
       assert_match(%r!New jekyll site installed in!, output)
+    end
+
+    should "skip bundle install when opted to" do
+      capture_stdout { Jekyll::Commands::New.process(@args, "--skip-bundle") }
+      output = Jekyll.logger.messages.last
+      bundle_message = "Bundle install skipped."
+      assert_includes output, bundle_message
     end
   end
 


lib/jekyll/commands/new.rb
```diff
@@ -114,6 +116,30 @@ def site_template
         def scaffold_path
           "_posts/0000-00-00-welcome-to-jekyll.markdown.erb"
         end
+
+        def bundle_install(path)
+          Jekyll::External.require_with_graceful_fail "bundler"
+          $stdout.print "\nRun 'bundle install' with default theme? (Y/N)  "
```
envygeeks: Please use the Jekyll Logger.

lib/jekyll/commands/new.rb
```diff
@@ -114,6 +116,30 @@ def site_template
...
+          @ans = ENV["CI"] ? "N" : $stdin.gets.chomp
+          until @ans == "Y" || @ans == "N"
```
envygeeks: A standard CLI must understand all cases of "y" and "yes" and "n" and "no", "y/n" is not good enough here.  I often type yes, rather than y, I am a very explicit person in what my intentions are, I'm not a keyboard basher.  I would be fuming if it rejected "yes" and probably start a Twitter storm.

lib/jekyll/commands/new.rb
```diff
@@ -114,6 +116,30 @@ def site_template
...
+        end
+
+        def stdinput(path)
+          if @ans == "Y"
+            Jekyll.logger.info "Running bundle install in", path.to_s + "..."
+            Dir.chdir(path.to_s) do
+              system("bundle install")
+            end
+          elsif @ans == "N"
+            Jekyll.logger.warn "\nBundle install skipped.\n" \
```
envygeeks: This shouldn't be a warning.  It's a valid choice.

lib/jekyll/commands/new.rb
```diff
@@ -114,6 +116,30 @@ def site_template
...
+            "Edit your Gemfile and run 'bundle install' " \
```
envygeeks: I don't know that this is useful tbh, any reason why this is here?

lib/jekyll/commands/new.rb
ashmaroli: Jekyll logger inserts a newline at the end.. I proposed _print_ to have gets on same line

lib/jekyll/commands/new.rb
ashmaroli: skipping b i _may not_ install minima for first-time users. They wont be able to jekyll serve till minima is found in the gemfile.lock

envygeeks: Technically they won't be able to do anything Jekyll until they bundle install.  This should be obvious though I will concede to new users it might not be obvious, so I don't disagree if you choose to leave it because I can see it's benefits.

lib/jekyll/commands/new.rb
ashmaroli: invalid 'coz its looking for Y/N instead of y/n.. but your comment above explained the need to remove this

ashmaroli: I added it as a warning as bundle install is a necessary step, yet not mandatory at that very instant.

envygeeks: It shouldn't be a warning no matter how you phrase it.  It's the users choice when they install.  It should simply say "Skipping Bundle Install: You should do it later" or something to that extend.  Shaming the user for making a choice makes them feel like your choice is the only choice.  A user will do things their way, in their order.

envygeeks: And that "Skipping bundle install" should be an informational message, not a warning.

ashmaroli:
Add bundler as a runtime dependency to gemspec?

envygeeks:
I don't think that's a good idea, that could lead to problems on a system.

parkr:
Thanks for this PR! I'm not sure how I feel about this either. The current code is interactive, which I'm not such a big fan of. `rails new` always ran `bundle install` which I liked as a newcomer, but the Rails community is pretty different from the Jekyll community, I think. Not sure how I feel about this.

In general, we are headed in the direction of "always use bundler" so this isn't problematic from that perspective. But we should gracefully fail if bundler isn't available and mention that the user should run `bundle install` before running `jekyll`. 😄

ashmaroli:
Hi @parkr, I'm pretty confused.. 
> But we should gracefully fail if bundler isn't available

PR does that already..
> mention that the user should run bundle install before running jekyll. 

??
There wont be a Gemfile to run `bundle` against until `jekyll new` has been called.
Running bundle after blog install will install _minima_ theme gem for first time users; update minima if necessary for existing customers on J3.2.
Running bundle interactively lets the user edit their Gemfile / view the installed directory contents before having a request sent to rubygems.org. Once they make their decision, the new theme has to be bundled before being served.

mattr-:
I'm :+1: with the idea of running bundler when you create a new Jekyll site. Let's just run it all the time though, instead of asking the user if we should run it. I see no use case where a user would not want to run bundler after getting a new site.

ashmaroli:
@mattr- Thanks for the +1. :smiley: 

> I see no use case where a user would not want to run bundler after getting a new site.

I mentioned a possible scenario in the PR desc.:
_The user may choose to skip 'bundle install' if they plan to edit their new gemfile and change the theme gem, or if they want to inspect the new directory before running the bundle install command_

mattr-:
Those are technical use cases that I don't want to solve for, especially the directory inspection. While I expect that a large set of Jekyll's users are going to be on the more technical side, it's ridiculously easy to run `bundle` after we've done it as part of site generation after editing the Gemfile. I see where you're coming from on this, but I feel like it's only going to be used in a tiny set of cases and I'd rather come back and add it later if it's a highly requested feature.

ashmaroli:
Okay. I'll modify the PR.

spudowiar:
I'd prefer the simple option of an argument `--no-bundler`, there's no need to have an interactive prompt, Jekyll doesn't do this already for anything AFAIK. Other that that, :+1:!

ashmaroli:
Replaced interactive bundle with running `bundle install` by default. Can be skipped by using a switch `--skip_bundle`

lib/jekyll/commands/new.rb
```diff
@@ -114,6 +115,29 @@ def site_template
...
+            bundle_install path
+          end
+        end
+
+        def bundle_install(path)
+          Jekyll::External.require_with_graceful_fail "bundler"
+          Jekyll.logger.info "Running bundle install in #{path.cyan}..."
+          Dir.chdir(path) do
+            unless ENV["CI"] || ENV["TEST"] # skip `bundle` during local tests
+              system("bundle install")
```
ashmaroli: I'd like to skip `bundle install` when running tests. Could find the appropriate ENV key/value.

envygeeks: I don't know about all that but that should be `system("bundle", "install")`

envygeeks: Meh, actually, tbh, it doesn't matter.  I mix the two anyways I wouldn't blame you if you don't care to change it either.

spudowiar: Doesn't _really_ matter until you start passing arguments dynamically :wink: Although I do prefer the variadic argument form

ashmaroli: it works.. so leaving it as is.. 
Thanks for the info, anyways :smiley:

envygeeks: @spudowiar it's that kind of terrible logic that leads to security issues in software.  It always matters.  Whether it works or not is irrelevant to the point, the statement stands and the option was left to leave it as is out of ease, not out of it "not mattering."

spudowiar: @envygeeks I'm saying that although you should always use the variadic form, in this case not changing it doesn't matter too much. I always use the variadic form

lib/jekyll/commands/new.rb
```diff
@@ -11,6 +11,7 @@ def init_with_program(prog)
 
             c.option "force", "--force", "Force creation even if PATH already exists"
             c.option "blank", "--blank", "Creates scaffolding but with empty files"
+            c.option "skip_bundle", "--skip_bundle", "Skip 'bundle install'"
```
parkr: I would prefer `--skip-bundle` here. Consistency of hyphens seems more intuitive.

lib/jekyll/commands/new.rb
```diff
@@ -114,6 +115,29 @@ def site_template
...
+          unless options["blank"] || options["skip_bundle"]
+            bundle_install path
+          end
+        end
+
+        def bundle_install(path)
+          Jekyll::External.require_with_graceful_fail "bundler"
+          Jekyll.logger.info "Running bundle install in #{path.cyan}..."
+          Dir.chdir(path) do
+            unless ENV["CI"] || ENV["TEST"] # skip `bundle` during local tests
```
parkr: I think we should be testing all our functionality in the tests. Ideally we'd just pass `--skip-bundle` if we really didn't want to bundle in CI.

lib/jekyll/commands/new.rb
parkr: I would prefer `system("bundle", "install")` as well. You may also have to specify the `BUNDLER_GEMFILE` variable as well.

lib/jekyll/commands/new.rb
spudowiar: I agree with @parkr, running `bundle install` won't hurt in tests, we can always skip it if time is precious

ashmaroli:
@jekyll/ecosystem :  Since, this PR adds bundle install by default, should `script/default-site` be modified to avoid switching directories and then executing `bundle install`, as well?

ashmaroli:
*Updates:*
  - changed skip switch to `--skip-bundle`
  - `system(bundle install)` replaced with `system("bundle", "install")`
  - modified `test/test_new_command` to test bundle-install-action and skip-switch

lib/jekyll/commands/new.rb
```diff
@@ -114,6 +115,27 @@ def site_template
...
           "_posts/0000-00-00-welcome-to-jekyll.markdown.erb"
         end
+
+        # After a new blog has been created, print a success notification and
+        # then automatically execute bundle install from within the new blog dir
+        # unless the user opts to generate a blank blog or skip 'bundle install'.
+
+        def after_install(path, options = {})
+          Jekyll.logger.info "New jekyll site installed in #{path.cyan}."
+          Jekyll.logger.info "Bundle install skipped." if options["skip-bundle"]
```
spudowiar: Seems superfluous since `--skip-bundle` is an explicitly chosen command line option, unaffected by the environment in any way

parkr: :+1:

benbalter:
LGTM.

parkr:
@jekyllbot: merge +minor

ashmaroli:
**TODO:**
- Update online documentation regarding this new feature. : 
  - add a para under `Quickstart`

--
/cc @DirtyF 
Whatsay? Is there any other place I should update?

ashmaroli:
@parkr et al, All those bundle (show) reports in test logs might get frustrating at some point in future.. 
Your opinion?

parkr:
> All those bundle (show) reports in test logs might get frustrating at some point in future.. 
> Your opinion?

Yeah definitely. If you can capture those and output them using `Jekyll.logger` then our current `capture_stdout` should ensure they're not written to actual stdout.

DirtyF:
@ashmaroli yeah, [homepage snippet](https://github.com/jekyll/jekyll/blob/master/site/index.html#L60-L64) will also be affected by this change, be sure to reflect changes here too.

PR #5239, opened 2016-08-12
Colorize interpolated output in logger.info

ashmaroli:
Colorize interpolated output in Jekyll.logger.info as a way of highlighting relevant text.

blog_path

theme_name n path

### DIFF
--- a/lib/jekyll/commands/new_theme.rb
+++ b/lib/jekyll/commands/new_theme.rb
@@ -16,6 +16,7 @@ class Jekyll::Commands::NewTheme < Jekyll::Command
       end
     end
 
+    # rubocop:disable Metrics/AbcSize
     def process(args, opts)
       if !args || args.empty?
         raise Jekyll::Errors::InvalidThemeName, "You must specify a theme name."
@@ -28,9 +29,10 @@ class Jekyll::Commands::NewTheme < Jekyll::Command
       end
 
       theme.create!
-      Jekyll.logger.info "Your new Jekyll theme, #{theme.name}," \
-        " is ready for you in #{theme.path}!"
+      Jekyll.logger.info "Your new Jekyll theme, #{theme.name.cyan}," \
+        " is ready for you in #{theme.path.to_s.cyan}!"
       Jekyll.logger.info "For help getting started, read #{theme.path}/README.md."
     end
+    # rubocop:enable Metrics/AbcSize
   end
 end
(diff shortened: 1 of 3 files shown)


envygeeks:
Please interpolate instead of doing "+", that's why we enforce double quotes.

mattr-:
LGTM.

ashmaroli:
squashing commits..

lib/jekyll/commands/new_theme.rb
```diff
@@ -22,14 +22,20 @@ def process(args, opts)
       end
 
       new_theme_name = args.join("_")
-      theme = Jekyll::ThemeBuilder.new(new_theme_name, opts)
+      @theme = Jekyll::ThemeBuilder.new(new_theme_name, opts)
+
+      update_logger
+    end
+
+    def update_logger
```
mattr-: I didn't notice this before, but the addition of the `update_logger` method complicates this command quite a bit. I think we're better off without the extra method. Please make that change, and I'll reapprove.

ashmaroli: I agree, it was added to satisfy rubocop with an earlier commit. I forgot to remove it when I reverted back to using interpolation within the strings.

ashmaroli:
Looks like Rubocop doesnt want to keep it simple. :frowning:

mattr-:
I recommend wrapping that method to disable that Rubocop check for now:

```ruby
# rubocop:disable Metrics/AbcSize
def method
...
end
# rubocop:enable Metrics/AbcSize
```

envygeeks:
You'll need to squash before we can approve @ashmaroli, looks like Jekyllbot resets your approvals once you commit again, that's pretty clever and great programming by @parkr.

ashmaroli:
Squashed..
I like Jekyllbot resetting approvals. It makes sense too, as the new commit may have brought in changes not acceptable to the maintainers.
+1 to Parker for that.

mattr-:
LGTM. Thanks for making the updates! :metal:

ashmaroli:
You're welcome :smiley:  :metal:

ashmaroli:
There's that travis bug again.. Restart required..

mattr-:
LGTM. :+1:

envygeeks:
LGTM.  @jekyllbot: merge +minor

PR #5240, opened 2016-08-13
Update AppVeyor config.

XhmikosR:
Install gems in vendor and cache those only.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,3 @@
-# https://www.appveyor.com/docs/appveyor-yml
-
 version: "{build}"
 
 clone_depth: 10
@@ -13,31 +11,24 @@ build: off
 
 install:
   - SET PATH=C:\Ruby%RUBY_FOLDER_VER%\bin;%PATH%
-  - bundle install --retry 5
+  - bundle install --retry 5 --jobs=%NUMBER_OF_PROCESSORS% --clean --path vendor\bundle
 
 environment:
   BUNDLE_WITHOUT: "benchmark:site:development"
   matrix:
     - RUBY_FOLDER_VER: "23"
-      GEMS_FOLDER_VER: "2.3.0"
       TEST_SUITE: "test"
     - RUBY_FOLDER_VER: "23"
-      GEMS_FOLDER_VER: "2.3.0"
       TEST_SUITE: "cucumber"
     - RUBY_FOLDER_VER: "23"
-      GEMS_FOLDER_VER: "2.3.0"
       TEST_SUITE: "fmt"
     - RUBY_FOLDER_VER: "23"
-      GEMS_FOLDER_VER: "2.3.0"
       TEST_SUITE: "default-site"
     - RUBY_FOLDER_VER: "23-x64"
-      GEMS_FOLDER_VER: "2.3.0"
       TEST_SUITE: "test"
     - RUBY_FOLDER_VER: "22"
-      GEMS_FOLDER_VER: "2.2.0"
       TEST_SUITE: "test"
     - RUBY_FOLDER_VER: "21"
-      GEMS_FOLDER_VER: "2.1.0"
       TEST_SUITE: "test"
 
 test_script:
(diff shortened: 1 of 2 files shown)


XhmikosR:
Hmm, so it seems this can't work without a Gemfile.lock. Too bad, because the current solution to cache the whole bin folder is bad.

I guess I will just remove caching...

appveyor.yml
```diff
@@ -46,7 +49,6 @@ test_script:
   - bundler --version
   - bash ./script/cibuild
 
-cache:
+#cache:
   # If one of the files after the right arrow changes, cache will be skipped
-  - C:\Ruby%RUBY_FOLDER_VER%\bin -> Gemfile,jekyll.gemspec,appveyor.yml
-  - C:\Ruby%RUBY_FOLDER_VER%\lib\ruby\gems\%GEMS_FOLDER_VER% -> Gemfile,jekyll.gemspec,appveyor.yml
+#  - 'vendor -> appveyor.yml,Gemfile,jekyll.gemspec'
```
parkr: Are we then *not* caching `vendor`? Or are we relying on the automatic inclusion of this directory?

parkr:
Thank you, @XhmikosR!

appveyor.yml
XhmikosR: We can't use a custom deployment location because bundler complains that there is no Gemfile.lock. See https://ci.appveyor.com/project/jekyll/jekyll/build/259

I'm no expert with Rubygems so perhaps there is another way to do it instead of using `bundler --deployment`.

XhmikosR: I can go back to the previous solution but to me caching the whole bin folder seems bad. Or the system wide gems folder...

parkr: Haha yeah caching the system-wide gems folder is not ideal. You shouldn't need `--deployment` for us (that's for running `bundle install` before you boot a server and it's why heroku, for example, requires the `Gemfile.lock`) – can we just do a custom list of `--without` to make sure we only install what we need?

appveyor.yml
```diff
@@ -13,7 +13,10 @@ build: off
 
 install:
   - SET PATH=C:\Ruby%RUBY_FOLDER_VER%\bin;%PATH%
-  - bundle install --retry 5
+  # Bundler downgrade should be removed when the issue is fixed upstream
+#  - gem uninstall bundler -x
+#  - gem install bundler -v "=1.10.6" --no-document
+  - bundle install --retry 5 --jobs=%NUMBER_OF_PROCESSORS%
```
parkr: We should be able to use `--without` and specify the `vendor` dir via `--path vendor`. Is that invalid?

appveyor.yml
XhmikosR: I'm reading http://bundler.io/v1.12/bundle_install.html and deployment seems the case for us.

We already have `without` https://github.com/jekyll/jekyll/blob/master/appveyor.yml#L19

So basically, we need to instruct bundle to install the gems in a custom folder which we can cache.

appveyor.yml
XhmikosR: I haven't tried the `path` option. Let me try that.

XhmikosR:
@parkr: `path` seems to do the job as far as I can tell:

```
Bundle complete! 40 Gemfile dependencies, 90 gems now installed.
Gems in the groups benchmark, site and development were not installed.
Bundled gems are installed into ./vendor/bundle.
```

Note that caching isn't actually active since the builds don't pass.

XhmikosR:
On a side note @parkr, the `site-default` failure seems to be because Windows doesn't recognize exe/jekyll as ruby executable file.

Doing `bundle exec ruby exe/jekyll new tmp/default-site` should work. Should I try this in this PR?

ashmaroli:
@XhmikosR, default-site script seems to  run with simply `exec` instead of `bundle exec'
(I'm aware, its not entirely correct.)
[Appveyor log](https://ci.appveyor.com/project/jekyll/jekyll/build/352) showing passed default-site test.

XhmikosR:
I pushed my change to use `ruby exe/jekyll` in this PR and works, but it fails later  https://ci.appveyor.com/project/jekyll/jekyll/build/349/job/x73k0syp9xmuvib3

ashmaroli:
Ah yes.. I see it now.. double-drive-letter in the path..
`bundle exec ruby` is the better solution as it does not create a 'no-return-to-previous console' like exec does.. 
do you mind if I add `ruby` to my PR as well?

XhmikosR:
You can cherry pick it if you want. Or wait until this PR is merged and rebase then.

ashmaroli:
I'll push a commit manually referencing yours..
Thanks.

XhmikosR:
That is not the proper way to do it. Read about cherry picking or like I said, wait until this is merged and rebase.

ashmaroli:
okay.. I'm aware of cherry-picking. I'll wait for this to get merged.

parkr:
@jekyllbot: merge +dev

PR #5241, opened 2016-08-13
Edit files to pass Appveyor --Experimental--

ashmaroli:
This PR deals with modifying files purely to try and achieve minimal errors/failures with Appveyor ( & secondarily, with Travis )

### DIFF
--- a/lib/jekyll/plugin_manager.rb
+++ b/lib/jekyll/plugin_manager.rb
@@ -82,6 +82,16 @@ module Jekyll
       if site.config["plugins_dir"].eql? Jekyll::Configuration::DEFAULTS["plugins_dir"]
         [site.in_source_dir(site.config["plugins_dir"])]
       else
+        custom_plugins_path
+      end
+    end
+
+    def custom_plugins_path
+      if Jekyll::Utils::Platforms.windows?
+        Jekyll.logger.warn "Conflict:", "Jekyll doesn't currently support " \
+                  "setting plugins_dir to a different dir on Windows"
+        Array(site.config["plugins_dir"])
+      else
         Array(site.config["plugins_dir"]).map { |d| File.expand_path(d) }
       end
     end
--- a/test/test_collections.rb
+++ b/test/test_collections.rb
@@ -178,6 +178,8 @@ class TestCollections < JekyllUnitTest
       })
       @site.process
       @collection = @site.collections["methods"]
+
+      skip_if_windows
     end
 
     should "include the symlinked file as it resolves to inside site.source" do
--- a/test/test_filters.rb
+++ b/test/test_filters.rb
@@ -561,6 +561,10 @@ class TestFilters < JekyllUnitTest
     end
 
     context "group_by filter" do
+      setup do
+        skip_if_windows "skipped in favour of test designed for Windows"
+      end
+
       should "successfully group array of Jekyll::Page's" do
         @filter.site.process
         grouping = @filter.group_by(@filter.site.pages, "layout")
@@ -588,6 +592,55 @@ class TestFilters < JekyllUnitTest
               "The list of grouped items for '' is not an Array."
             )
             assert_equal 15, g["items"].size
+          end
+        end
+      end
+
+      should "include the size of each grouping" do
+        grouping = @filter.group_by(@filter.site.pages, "layout")
+        grouping.each do |g|
+          p g
+          assert_equal(
+            g["items"].size,
+            g["size"],
+            "The size property for '#{g["name"]}' doesn't match the size of the Array."
+          )
+        end
+      end
+    end
+
+    context "group_by filter on Windows" do
+      setup do
+        skip_unless_windows
+      end
+
+      should "successfully group array of Jekyll::Page's" do
+        @filter.site.process
+        grouping = @filter.group_by(@filter.site.pages, "layout")
+        grouping.each do |g|
+          assert(
+            ["default", "nil", ""].include?(g["name"]),
+            "#{g["name"]} isn't a valid grouping."
+          )
+          case g["name"]
+          when "default"
+            assert(
+              g["items"].is_a?(Array),
+              "The list of grouped items for 'default' is not an Array."
+            )
+            assert_equal 4, g["items"].size
+          when "nil"
+            assert(
+              g["items"].is_a?(Array),
+              "The list of grouped items for 'nil' is not an Array."
+            )
+            assert_equal 2, g["items"].size
+          when ""
+            assert(
+              g["items"].is_a?(Array),
+              "The list of grouped items for '' is not an Array."
+            )
+            assert_equal 14, g["items"].size
           end
         end
       end
--- a/test/test_plugin_manager.rb
+++ b/test/test_plugin_manager.rb
@@ -110,6 +110,10 @@ class TestPluginManager < JekyllUnitTest
   end
 
   context "plugins_dir is set to a different dir" do
+    setup do
+      skip_if_windows "Jekyll doesn't currently support " \
+      "setting plugins_dir to a different dir on Windows"
+    end
     should "expand plugin path" do
       site = double({ :config => { "plugins_dir" => "some_other_plugins_path" } })
       plugin_manager = PluginManager.new(site)
--- a/test/test_site.rb
+++ b/test/test_site.rb
@@ -170,6 +170,7 @@ class TestSite < JekyllUnitTest
     end
 
     should "sort pages alphabetically" do
+      skip_if_windows
       method = Dir.method(:entries)
       allow(Dir).to receive(:entries) do |*args, &block|
         method.call(*args, &block).reverse
@@ -199,6 +200,38 @@ class TestSite < JekyllUnitTest
         sitemap.xml
         static_files.html
         symlinked-file
+      )
+      assert_equal sorted_pages, @site.pages.map(&:name)
+    end
+
+    should "sort pages alphabetically on Windows" do
+      skip_unless_windows
+      method = Dir.method(:entries)
+      allow(Dir).to receive(:entries) do |*args, &block|
+        method.call(*args, &block).reverse
+      end
+      @site.process
+      sorted_pages = %w(
+        %#\ +.md
+        .htaccess
+        about.html
+        application.coffee
+        bar.html
+        coffeescript.coffee
+        contacts.html
+        deal.with.dots.html
+        dynamic_file.php
+        environment.html
+        exploit.md
+        foo.md
+        humans.txt
+        index.html
+        index.html
+        info.md
+        main.scss
+        properties.html
+        sitemap.xml
+        static_files.html
       )
       assert_equal sorted_pages, @site.pages.map(&:name)
     end
@@ -294,10 +327,12 @@ class TestSite < JekyllUnitTest
 
       should "remove orphaned files in destination" do
         @site.process
-        refute_exist dest_dir("obsolete.html")
-        refute_exist dest_dir("qux")
-        refute_exist dest_dir("quux")
-        assert_exist dest_dir(".git")
+        unless Jekyll::Utils::Platforms.really_windows?
+          refute_exist dest_dir("obsolete.html")
+          refute_exist dest_dir("qux")
+          refute_exist dest_dir("quux")
+          assert_exist dest_dir(".git")
+        end
         assert_exist dest_dir(".git", "HEAD")
       end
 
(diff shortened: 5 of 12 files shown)


ashmaroli:
@envygeeks, Appveyor results down to 3 failures. Please verify if the changes I proposed is acceptable.

lib/jekyll/plugin_manager.rb
```diff
@@ -81,6 +81,13 @@ def require_plugin_files
     def plugins_path
       if site.config["plugins_dir"].eql? Jekyll::Configuration::DEFAULTS["plugins_dir"]
         [site.in_source_dir(site.config["plugins_dir"])]
+      else plugins_path_custom
```
envygeeks: This should be indented and put on it's own line please.

test/test_collections.rb
```diff
@@ -180,13 +180,15 @@ class TestCollections < JekyllUnitTest
       @collection = @site.collections["methods"]
     end
 
-    should "include the symlinked file as it resolves to inside site.source" do
-      assert_includes @collection.filtered_entries, "um_hi.md"
-      refute_includes @collection.filtered_entries, "/um_hi.md"
-    end
+    unless Jekyll::Utils::Platforms.windows?
+      should "include the symlinked file as it resolves to inside site.source" do
+        assert_includes @collection.filtered_entries, "um_hi.md"
```
envygeeks: You should be able to do `skip "Unsupported test on Windows" if Jekyll::Utils::Platforms.windows?` inside the test.  The reason this is preferable over the swallowing if is it produces inconsistent test numbers, and doesn't let people know why it's that way.

envygeeks:
It looks good to me in theory but I'm confused as to why the lot of these tests need to be hidden, considering most of them ***should*** work on Windows. If they don't I'd really like the skip message to reflect why they don't work.

ashmaroli:
The difference could be because of how Windows n Unix-based OS create and handle symlinks.
If you compare this [Appveyor log](https://ci.appveyor.com/project/jekyll/jekyll/build/273/job/miv69c8e1gt0m9ra) vs [Travis log](https://travis-ci.org/jekyll/jekyll/jobs/152053611), you'll see a duplicate block entry of the post-listings in Travis. I suspect, this could be due to symlinks as well.

test/test_filters.rb
```diff
@@ -463,7 +463,13 @@ def to_liquid
               g["items"].is_a?(Array),
               "The list of grouped items for 'default' is not an Array."
             )
-            assert_equal 5, g["items"].size
+            hr = "<---==========================================================--->"
```
envygeeks: Why did this need to be added?

test/test_filters.rb
```diff
@@ -475,7 +481,11 @@ def to_liquid
               g["items"].is_a?(Array),
               "The list of grouped items for '' is not an Array."
             )
-            assert_equal 13, g["items"].size
+            if Jekyll::Utils::Platforms.windows?
```
envygeeks: This should be a unique test.  This kind of tight coupling is terrifying (not your fault.)

test/test_plugin_manager.rb
```diff
@@ -111,6 +111,7 @@ def with_no_gemfile
 
   context "plugins_dir is set to a different dir" do
     should "expand plugin path" do
+      skip "Jekyll doesn't support setting plugins_dir to a different dir on Windows"
```
envygeeks: This message is misleading.  We would like to support it, it's because of Ruby bugs we can't.

test/test_filters.rb
ashmaroli: so that I could differentiate between various grouped items in the log, and hence count them easily..
That will be discarded in the final squash commit.

test/test_site.rb
```diff
@@ -192,12 +192,14 @@ def generate(site)
         index.html
         index.html
         main.scss
-        main.scss
         properties.html
         sitemap.xml
         static_files.html
-        symlinked-file
       )
+      unless Jekyll::Utils::Platforms.windows?
```
envygeeks: This entire test should be skipped at this point no?

test/test_plugin_manager.rb
envygeeks: Note: Eventually I plan on implementing Pathutil here and making Pathutil cleaner so it can support this for Windows users.  I've a lot of Windows users here that use Jekyll so it is on my plate but since I'm a systems guy by job, I don't find much time to work on anything more than scaling Jekyll.

test/test_site.rb
ashmaroli: why skip? i dont get it..  
Its unless windows? ..

envygeeks: Because as mentioned before, those kind of couplings are terrifying, and they're gonna make our eventual move to RSpec harder.  If you need a conditional test use contexts:

```
context "when not on Windows"
   it "should do this stuff"
     skip_if_windows

context "on Windows"
   it "should do stuff"
```

Sorry I didn't elaborate that.

envygeeks: You can choose whether to `skip_when_not_windows`, dealers choice on that, considering the test will pass on Linux as well, so... yeah.

ashmaroli:
I hope I implemented the really_windows? method as @envygeeks  intended.. 
Also, the remaining 2 fails are quite puzzling.. They dont occur when testing locally.

lib/jekyll/plugin_manager.rb
```diff
@@ -82,6 +82,14 @@ def plugins_path
       if site.config["plugins_dir"].eql? Jekyll::Configuration::DEFAULTS["plugins_dir"]
         [site.in_source_dir(site.config["plugins_dir"])]
       else
+        custom_plugins_path
+      end
+    end
+
+    def custom_plugins_path
+      if Jekyll::Utils::Platforms.windows?
+        Array(site.config["plugins_dir"])
```
parkr: What makes the `File.expand_path(d)` call invalid? I believe what needs to change is the _test_, not this code.

test/test_plugin_manager.rb
```diff
@@ -110,6 +110,10 @@ def with_no_gemfile
   end
 
   context "plugins_dir is set to a different dir" do
+    setup do
+      skip_if_windows "Jekyll doesn't currently support " \
+      "setting plugins_dir to a different dir on Windows"
```
parkr: If we don't support it, then this should be caught in the application code, not told in the test code.

parkr:
Hm, I'm not sure I like the direction of this PR at this point. A number of the failing tests on `master` on Windows are related to the differences between Windows and non-Windows filesystems. I think we should work to make them essentially work the same. For example, we set `plugins: /tmp/plugins` or something and we should simply expect `D:/tmp/plugins` or something (e.g. if windows, add the drive name, if not, then don't). For the symlinks, I think we would want to use the `win32-file` gem which shims a number of `File` methods with proper Windows system calls.

Thanks for this PR! I really want to get these tests green before v3.3 if possible.

lib/jekyll/plugin_manager.rb
ashmaroli: Why I changed the _code_ is because when the defined Array gets mapped to file.expand_path(d), paths such as C:/tmp/.. etc are created instead of /tmp/...

test/test_plugin_manager.rb
ashmaroli: Agreed. But I dont know enough Ruby to do that..

lib/jekyll/plugin_manager.rb
parkr: Interesting, I thought that is what we wanted!

envygeeks: @parkr to an extent it is, except Pathname is entirely broken on Windows when it comes to paths so it starts complaining and breaking Jekyll.

ashmaroli:
I tried that route (adding drive name if windows..) didnt quite get there..
will check win32-file gem.

lib/jekyll/plugin_manager.rb
ashmaroli: if 'tmp/plugins', 'tmp/blah' etc are dirs relative to the root of the blog, 'C:/tmp/plugins', 'C:/tmp/blah are not anywhere near the generated blog's root.

test/test_plugin_manager.rb
ashmaroli: Oh.. by stating that, did you mean a strict default should be set for windows; and raise an exception if otherwise?

ashmaroli:
> For example, we set _plugins: /tmp/plugins_ or something and we should simply expect _D:/tmp/plugins_ or something

@parkr  By **_/tmp/plugins_**, did you mean **_/TMPDIR/plugins_** or **_blog-root/tmp/plugins_** ?

ashmaroli:
@parkr, I'm not justifying the wrongs in this PR.. but after skipping so many tests,  6 out of 7 are in the green.. :smile:

script/default-site
```diff
@@ -12,7 +12,7 @@ workdir=$(pwd)
 echo "$0: creating new default site"
 if [[ "$(uname -s)" != "Linux" ]]
 then 
-  exec exe/jekyll new tmp/default-site
+   `exe/jekyll new tmp/default-site`
```
envygeeks: Why did you wrap this to backticks? That matters little in a bash script, and I would prefer exec stay as we've no reason to hold our own script open rather than replace it, the latter should have also of had exec too, because we care little after we launch the script and their exit status is the status to rule them all.  In any event the backtick is entirely unnecessary.

ashmaroli: Ah.. I was just testing this change... 
exec did not seem to run the code that followed.. 
I'd reverted back locally.. just didnt push to remote..

envygeeks: Looking at the logic it seems that bit is meant to run on OS X. /cc @benbalter and @parkr who originally created it so they can further explain the logic behind it.

ashmaroli: I added the if..fi block there..
The revert has been pushed.. and it passed Appveyor..

envygeeks:
Superficial facial edits should be constrained to their own pull request, if this is a utility pull that fixes a problem, it should not add facial edits that are not necessarily relevant to the case being handled because it only adds more stuff we need to review and might not want.

ashmaroli:
those commits serve no purpose to fix a problem. Hence refrained from adding to PR of their own.
I added them to see if it helped me in current PR. But alas, they only work in travis.
I'll be dropping such commits when I squash at the end.

ashmaroli:
yay! exited script/default-site successfully. :tada:

script/default-site
```diff
@@ -2,16 +2,16 @@
...
 echo "$0: setting up tmp directory"
 mkdir -p ./tmp
 rm -Rf ./tmp/default-site
 
-workdir=$(pwd)
-
 echo "$0: creating new default site"
-bundle exec exe/jekyll new tmp/default-site
+bundle exec jekyll new tmp/default-site
+
```
ashmaroli: It seems like the executable exe/jekyll is unnecessary. Appveyor & Travis both, are able to run Jekyll as available at source . 
Do correct me, if it really is necessary

ashmaroli:
*Update:* Reduced the no.of failing errors in Cucumber test scenarios by letting ruby execute Jekyll binary instead of shell executing it.
The remaining errors in Cucumber suite pertains to usage of backslashes in windows paths.

ashmaroli:
for `features/collections.feature`, the array in `_site/index.html` is
```diff
--- expected
+++ actual

    Collections: 
    _methods/collection/entries 
    _methods/configuration.md 
    _methods/escape-+ #%20[].md 
    _methods/sanitized_path.md 
    _methods/site/generate.md 
    _methods/site/initialize.md 
--- _methods/um_hi.md

    Collections: 
    _methods/collection/entries 
    _methods/configuration.md 
    _methods/escape-+ #%20[].md 
    _methods/sanitized_path.md 
    _methods/site/generate.md 
    _methods/site/initialize.md 
+++ _methods/yaml_with_dots.md
```

ashmaroli:
for case: `features/site_configuration.feature:158`
there seems to be a discrepancy with the Timezone data.. hence:

```diff
--- expected
+++ actual

--- Post Layout: <p>content for entry1.</p>
---  built at 2013-04-09T23:22:00-04:00

+++ Post Layout: <p>content for entry1.</p>
+++  built at 2013-04-10T03:22:00+00:00
```

If I were to pass `UTC-04:00` as the TimeZone Data, then:
```diff
+++ actual

+++ Post Layout: <p>content for entry1.</p>
+++  built at 2013-04-10T07:22:00+04:00
```

If I were instead to pass `UTC-04:00` as the TimeZone Data, 
but have the FrontMatter modified to `04:00` instead of `-04:00`, then:
```diff
+++ actual

+++ Post Layout: <p>content for entry1.</p>
+++  built at 2013-04-09T23:22:00+04:00
```

But then if I modify TZ to `UTC+04:00` and use `-04:00` in FrontMatter, then:
```diff
+++ actual

+++ Post Layout: <p>content for entry1.</p>
+++  built at 2013-04-09T23:22:00-04:00
```

ashmaroli:
:tada:  Appveyor is finally Green!! (Travis needs a restart on one of the suites)
@parkr & @envygeeks would love some feedback on this..

ashmaroli:
*Note:* This PR is not for merging.. but rather to serve as a Case Study Log..

PR #5244, opened 2016-08-14
Fix a link in deployment-methods.md

mkasberg:
The link to the scp deploy script was broken - it appears someone changed the link text and did not update the matching link text in the markdown footer. To prevent this problem from happening again, and for consistency with the rest of this markdown file, I changed both links to inline style links.

### DIFF
--- a/site/_docs/deployment-methods.md
+++ b/site/_docs/deployment-methods.md
@@ -98,13 +98,12 @@ Another way to deploy your Jekyll site is to use [Rake](https://github.com/ruby/
 ### scp
 
 Once you’ve generated the `_site` directory, you can easily scp it using a
-`tasks/deploy` shell script similar to [this deploy script][]. You’d obviously
-need to change the values to reflect your site’s details. There is even [a
-matching TextMate command][] that will help you run this script.
+`tasks/deploy` shell script similar to this:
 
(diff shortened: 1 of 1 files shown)


envygeeks:
:-1: that script uses scss directly, it's preferred to do it via Jekyll itself since it's embedded.

envygeeks:
To note, I already know it was there, but now that it's been brought to attention, I question if it's still viable.

mkasberg:
That makes sense. I actually removed that line from my own version of the script. I was just trying to fix the broken link, since it looks bad on the website right now.

Is there a more preferable implementation of that script somewhere that we could link to?

Alternatively, since it's basically a one-liner, what if we just actually write the script into the markdown? Like this:

    scp -r _site/* user@server:/home/user/public_html

envygeeks:
I am :+1: on just inlining it. That is what I thought when I seen it.

mkasberg:
Just updated the pull request to have it inlined.

envygeeks:
LGTM.

DirtyF:
LGTM.

mattr-:
LGTM 

@jekyllbot: merge +site

albertvolkman:
Also, the text immediately following is an exact duplicate.

screenshot 2016-09-15 13 20 45

PR #5253, opened 2016-08-17
Format config file for easy editing

ashmaroli:
This PR is to format _config.yml to increase readability and enable easy editing.

### DIFF
--- a/lib/site_template/_config.yml
+++ b/lib/site_template/_config.yml
@@ -7,25 +7,31 @@
 #
 # For technical reasons, this file is *NOT* reloaded automatically when you use
 # 'jekyll serve'. If you change this file, please restart the server process.
-
+#
+# ------------------------------------------------------------------------------
+#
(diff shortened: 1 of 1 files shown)


envygeeks:
All this does is force me to look right.

DirtyF:
If you want to make it easier to scan, properties should be right aligned:
http://uxmovement.com/forms/form-label-proximity-right-aligned-is-easier-to-scan/

That's what @envygeeks expressed. Sorry but it's a 👎  for me on this one.

mdxprograms:
I'm ok with it as is. I don't think reading a config file is that difficult in the first place. Especially using YAML.

mdxprograms:
👎 for me

mattr-:
Thanks for the edits. I can understand your point of view. We'll keep the file as is though.

ashmaroli:
Thank you all for your feedback.

PR #5256, opened 2016-08-18
Fix issue where Windows drive name is stripped from Jekyll.sanitized_path incorrectly

kwokfu:
Strip drive name only when necessary.


ashmaroli:
This works on Windows, so a :+1:  from me.

ashmaroli:
\cc @parkr, (just incase you've nt examined this PR already)

DirtyF:
@ashmaroli No need to mention maintainers, they are notified, and please give them some time to review, especially when they're on vacation ;)

parkr:
If you're fixing #5192, can you please add a test for this to our test suite so we know it's fixed?

This is very sensitive code so I'd love a 👍 from @jekyll/security as well. Thank you!

benbalter:
Ran into this bug this week. Thanks for fixing this! :smile:

lib/jekyll.rb
```diff
@@ -160,11 +160,11 @@ def sanitized_path(base_directory, questionable_path)
 
       questionable_path.insert(0, "/") if questionable_path.start_with?("~")
       clean_path = File.expand_path(questionable_path, "/")
-      clean_path.sub!(%r!\A\w:/!, "/")
 
-      if clean_path.start_with?(base_directory.sub(%r!\A\w:/!, "/"))
+      if clean_path.start_with?(base_directory)
         clean_path
       else
+        clean_path.sub!(%r!\A\w:/!, "/")
```
ptoomey3: Is there a test/scenario that shows where this would be useful; can we just drop the whole `clean_path.sub!(%r!\A\w:/!, "/")` bit? I'm trying to think of a legit case where we receive `questionable_path` with a leading drive letter AND `clean_path.start_with?(base_directory)` returns false AND where `File.join(base_directory, clean_path.sub!(%r!\A\w:/!, "/"))` is expected to do something sensible. It seems like, if `clean_path` contains a drive letter`, it is a fully absolute path by definition. So, doing this substitution and then appending it on to `base_directory` would always result in an invalid path.

envygeeks: @ptoomey3 it'll be a valid path in the sense that it'll work for home users, but it'll break every one of my installs of Jekyll.  Our Jekyll users here have 4 drives, plus another 8 on the network depending on who they are.  If they are working from their secondary drive (which is where we expect them to place their personal data as that is their personal drive to keep at no matter what) and this strips the path, Jekyll now thinks C:/ is the path, it is not, Jekyll just broke.

This fixes a problem while exacerbating a larger problem for other users and I can see potential security issues downstream (I'd have to test this because I'm not quite sure but I have a feeling this would affect Jekyll Assets which relies on this to ensure nothing silly is happening in a round-about-way and to keep Sprockets in check, along with the users.)

ashmaroli:
why dont we apply a patch only for windows users and leave as is for every other platform, by using conditional statements?

parkr:
In any event, we should write a test that demonstrates the breakage on AppVeyor so we can confirm this fixes things.

test/test_path_sanitization.rb
```diff
@@ -28,4 +28,35 @@ class TestPathSanitization < JekyllUnitTest
...
+
+  context "on Windows with file path has matching prefix" do
+    setup do
+      @base_path = "D:/site"
+      @file_path = "D:/sitemap.xml"
+      allow(Dir).to receive(:pwd).and_return("D:/")
+    end
+
+    should "not strip base path" do
+      assert_equal "D:/site/sitemap.xml",
```
ptoomey3: I'm still unclear on when this test's behavior is desirable. It seems like the path to `sitemap.xml` is absolute, so under what condition does it make sense to transform it into a relative path that is appended on to the base path? I understand that this is largely just testing existing behavior, but I'm wondering if that behavior is needed.

kwokfu: @ptoomey3, I'm adding this test based on #5276.

The scenario is if we configured source to D:/ and destination to D:/site, the file sitemap.xml will be created in the source path before copying to destination (i.e copy D:/sitemap.xml to D:/site/sitemap.xml). That is where Jekyll build fail in my test bed.

parkr: @ptoomey3 this method is like File.join, but enforces the base path :) it was written for that!

parkr:
I shudder to think of how many plugins and such we'd break if we changed `Jekyll.sanitized_path` to not just gracefully bolt the `base_directory` before the `questionable_path`, so I think we need to keep that behaviour.

I agree that the code at present is hella difficult to keep track of and reason about. A lot of that confusion comes down to the fact that the mysterious regexp and so on is not described through comments. For this critical piece of the codebase, maybe a line-by-line explanation would be helpful.

kwokfu:
@parkr I think the latest fix for #5276 should obsolete the regex that strip away drive name. I'm not sure if there is a valid reason to strip away any drive name. I think it is also a good to remove the regex line that strip drive name, as it really doesn't make any sense.

ashmaroli:
Hello @kwokfu, any updates on this?

lib/jekyll.rb
ashmaroli: > I'd have to test this because I'm not quite sure but I have a feeling...

@envygeeks : Did you get the time to test if this had any side-effects..?

kwokfu:
@parkr, these 2 test cases passed in Windows environment but failed *nix environment.

The reason behind that 2 fail cases is because of how the File.expand_path behaves differently between Windows (gives "D:/sitemap.xml") and *nix (gives "/D:/sitemap.xml").

I do not want to add too much decision in Jekyll.sanitized_path, which might make it harder to read. I am thinking if there is a better way to determine if it is running in Windows environment.

ashmaroli:
@kwokfu, why not use conditional filtering?
See: `jekyll/utils/platform.rb` => [really_windows?](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/utils/platforms.rb#L22-L25) method

ashmaroli:
When I mentioned conditional filtering, I implied that the changes you made be conditionally applied.. not test conditionally.. 
how can we be sure that this will not break on an unixoid users system?

kwokfu:
@ashmaroli, thats because both the tests were written specifically with Windows environment in mind. Or maybe I should just assert differently between Windows and *nix (i.e. D:/site/sitemap.xml for Windows and D:/site/D:/sitemap.xml for *nix). What you say?

ashmaroli:
say, a non-windows users runs `jekyll serve` using a release with this change incorporated. What do you expect to happen? 
continue serving or abort like travis previously did?

kwokfu:
@ashmaroli, I think serving shouldn't be a problem because of the changes barely introducing any new feature, but additional checking on the path prefix before actually stripping the drive letter.

Do let me know if you have any scenario pop out that I've missed, especially related to theme where files are sitting on another path.

parkr:
@XhmikosR Would you be willing to give this branch a try using a gem-based theme?

parkr:
An alternative here, which might make things a lot easier to grok, is to have a Windows-specific sanitized_path method:

```ruby
def windows_sanitized_path(base, questionable)
  # Assume drive names, do not strip them.
end

def sanitized_path(base, questionable)
  if Jekyll::Utils::Platforms.windows?
    return windows_sanitized_path(base, questionable)
  end

  # We know we're on Linux or a Mac now.
  # Assume no drive names.
end
```

What do you think about that?

ashmaroli:
> Would you be willing to give this branch a try using a gem-based theme?

The branch works fine with minima on Windows.. 
What I doubted was if this branch would work equally fine on a Linux or like platforms..

parkr:
@kwokfu Hoping to release v3.3.0 out here soon, what do you think about my suggestion above?

kwokfu:
@parkr, I'm still exploring this and thinking what if someone really did create a directory in *nix with a name that looks like Windows drive name (for whatever reason) and what will be the case that goes into the sanitized_path.

ashmaroli:
> what if someone really did create a directory in *nix with a name that looks like Windows drive name

I dont think someone can do something like that to affect the regex. Even if they did, the `santized_path` method strips only the initial drive name. It should be similar to this [block](https://github.com/jekyll/jekyll/blob/master/test/test_path_sanitization.rb#L15-L18) in the test_file

kwokfu:
@ashmaroli, thanks for pointing that out, but what I'm really interested is if the questionable path in the test start with a drive name instead of what it is now.

I'm still working on to have all my new tests passed in both platforms. Lets hope that God strike me an idea later today so that I can get a better test results tonight.

kwokfu:
@parkr, @ashmaroli, I believe current changes made is sufficient to handle both the reported bugs without altering the original design of the function. I think introducing the suggested logic for Windows-only will introduce duplicate codes or changing the designate behaviour of the function, hence, not really wanted to incorporate in this patch (may be in another patch with discussion on how this function should behave?).

parkr:
>  I think introducing the suggested logic for Windows-only will introduce duplicate codes or changing the designate behaviour of the function, hence, not really wanted to incorporate in this patch (may be in another patch with discussion on how this function should behave?).

@kwokfu My primary concern is ensuring that all possible path traversal vectors are covered. This function exists to prevent path traversal. My secondary concern is doing The Right Thing with regard to drive names. What if you request `C:/mysite` and your site source is at `D:/Users/My Documents/mysite`? What is the expected behaviour there? Being able to tailor our path traversal checking to also respect drive names (instead of ignoring them as this PR does) is worth the potential code duplication.

@ptoomey3 If you have time to load this up and try to crack it, I'd love to have your approval before I merge. FWIW, it passes our checks in this test suite. Thanks!

kwokfu:
I would also like to know is there any documentation or discussion restricting where can we create a new Jekyll site, like under /css? I'm getting error while generating site under that path. If that were not discussed previously, I might just file a bug and work from there.

ashmaroli:
> I'm getting error while generating site under that path.

@kwokfu Can you please elaborate further? What exactly was the error raised?

kwokfu:
@ashmaroli, just ran the same test in Windows and getting the same error, looks like same is happening to both platform. Following is the output for Windows. I'll paste the trace in issue if confirmed bug.

    D:\>jekyll new css
    D:\>cd css
    D:\css> jekyll build
    ..ignored..
    jekyll 3.2.1 | Error:  No such file or directory @ rb_file_s_stat - D:/css/css/Gemfile

ashmaroli:
@kwokfu That bug is not related to Jekyll 3.2.1 but rather this branch alone.
I'm able to build and serve a new site created at `css`

kwokfu:
@ashmaroli, thats weird. I'm not create a new site using this branch. Are you creating the new site under root? It has to be under root directory, in either Windows or *nix.

ashmaroli:
Interesting find.. was able to reproduce the bug on **Windows** at `C:\` using `3.2.1` and `master`. Surely warrants a dedicated ticket. Feel free to do so, if you have time..

ashmaroli:
@kwokfu, the bug you discovered seems to happen due to the presence of a subdirectory also named `css`. Deleting that directory resolves this edge-case-scenario. 
Jekyll 3.3 no longer requires that directory to be present in the root. It'll be handled by the theme gem.

ptoomey3:
> @ptoomey3 If you have time to load this up and try to crack it, I'd love to have your approval before I merge. FWIW, it passes our checks in this test suite. Thanks!

This looks ok. In the end, the path either has to be a prefix of the base or gets joined to the base. 👍

parkr:
> This looks ok. In the end, the path either has to be a prefix of the base or gets joined to the base. 👍

Thank you SO much, @ptoomey3! 🔒

parkr:
@jekyllbot: merge +bug

PR #5262, opened 2016-08-20
Replace liquid highlight tag with backticks

ashmaroli:
Replace {% highlight ---- %} and {% endhighlight %} with backticks to keep markdown syntax highlight intact for .md docs within site/_docs folder.

{% highlight --- %} is kept intact for lines code blocks involving optional `linenos` arg and those args not supported in GFM.

### DIFF
--- a/site/_docs/assets.md
+++ b/site/_docs/assets.md
@@ -9,14 +9,14 @@ a Ruby gem. In order to use them, you must first create a file with the
 proper extension name (one of `.sass`, `.scss`, or `.coffee`) and ***start the
 file with two lines of triple dashes***, like this:
 
-{% highlight sass %}
+```sass
 ---
 ---
 
 // start content
 .my-definition
   font-size: 1.2em
-{% endhighlight %}
+```
 
 Jekyll treats these files the same as a regular page, in that the output file
 will be placed in the same directory that it came from. For instance, if you
@@ -46,10 +46,10 @@ If you are using Sass `@import` statements, you'll need to ensure that your
 `sass_dir` is set to the base directory that contains your Sass files. You
 can do that thusly:
 
-{% highlight yaml %}
+```yaml
 sass:
     sass_dir: _sass
-{% endhighlight %}
+```
 
 The Sass converter will default the `sass_dir` configuration option to
 `_sass`.
@@ -72,10 +72,10 @@ The Sass converter will default the `sass_dir` configuration option to
 You may also specify the output style with the `style` option in your
 `_config.yml` file:
 
-{% highlight yaml %}
+```yaml
 sass:
     style: compressed
-{% endhighlight %}
+```
 
 These are passed to Sass, so any output style options Sass supports are valid
 here, too.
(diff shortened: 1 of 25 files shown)


crispgm:
I agree with you. But are there any reason why jekyll should provide a highlight tag?

ashmaroli:
The highlight tag additionally supports including line no.s not natively present in GFM.
All it takes is to add `linenos` after the language code.

So its presence is advantageous.

envygeeks:
The highlight tag is a relic of days gone past.  However, this needs some discussion.  I am for this as I often browse the documentation from Github directly so this gives me an advantage that others have had when they browse the actual website, but only @parkr and @mattr- (both of whom have been around far longer than I or @alfredxing) can tell us if they are even still necessary considering Github defaults to Kramdown now (I don't remember if Github's main site still uses GFM -- I thought it was only comments that did that) and Kramdown has the same support that Jekyll does for Rouge and Pygments, I think it's a good idea to modernize the site that way and remove old relics... put them in the museum (git history.)

envygeeks:
I should also /cc @jglovier and @benbalter who could also probably provide some insight too.

mattr-:
Back in the day, Jekyll used to support both Textile and Markdown as a markup language to use for pages and posts so the `{% highlight %}` tag was the original solution to provide markup agnostic syntax highlighting. The Jekyll documentation site was created while we still had support for both markup languages and the markdown backend situation was a bit less settled.

Since we're removed support for Textile and have started defaulting to Kramdown as the markdown generator, the `{% highlight %}` tag is a bit less useful. At this point, I believe its only useful use case is if you want to add line numbers to your code listing, which isn't supported by GFM's extensions to standard Markdown.

I'm :+1: on this change for the Jekyll site.

LGTM

benbalter:
:+1: to changing our own documentation, and what we suggest users use in our own documentation.

👎 to removing the highlight tag entirely, as it's still necessary for line numbers and for non-Kramdown markdown engines.

@envygeeks out of curiosity, if you had a question, why did you assign instead of cc?

ashmaroli:
@benbalter, I've already kept 'highlight' intact for existing codes for linenos; and for the documentation regarding the same.

envygeeks:
@benbalter mostly because I wanted to test out the new assignment stuff :stuck_out_tongue_winking_eye:

envygeeks:
LGTM.

parkr:
IIRC this changes the look of the blocks due to some CSS weirdness. Let me give it a try locally.

DirtyF:
@parkr it does not seem to be the case:

highlighter-backticks

parkr:
@DirtyF If you look closely, you'll see a margin is off there. Alas, the HTML is wildly different:

`jekyllrb.com`

screen shot 2016-08-23 at 2 53 38 pm

`this PR`

screen shot 2016-08-23 at 2 53 30 pm

ashmaroli:
I've made some edits to the `_style.scss` to remove the discrepancy mentioned above.
Add to this PR or create another?

\cc @jekyll/performance

envygeeks:
This is not a performance problem.

ashmaroli:
the edit is simply:
```sass
.highlighter-rouge .highlight {
  @extend .highlight;
  margin: 0;
  padding: 10px 0.5em;
}
```

mattr-:
@ashmaroli just add to this one.

envygeeks:
Can we get some screenshots of the new stuff? For approval.  Thanks!

ashmaroli:
## Screenshots

### Current
live

### PR
pr

parkr:
LGTM.

mattr-:
LGTM

@jekyllbot: merge +site

PR #5264, opened 2016-08-20
Add `static_file.name` and `.basename` Liquid attributes

### LINKED ISSUE #5263: Get static file name

---

---

It would be nice if `site.static_files` gave a `file.name` variable that is just the name of the file without the path or extension.  As a use case, I do this to add different favicon sizes to the head (naming the favicon `dimension.png`). Currently, my workaround is:

```liquid
{% for file in site.static_files %}
{% if file.path contains '/favicon/' %}
{% assign name = file.path | remove: '/favicon/' | remove: file.extname %}
<link rel='icon' href='{{ file.path | prepend: site.url }}' sizes='{{ name }}x{{ name }}' type='image/png'>
{% endif %}
{% endfor %}
```

DirtyF:
_Disclaimer: These are my first ruby lines_ 😅 

A proposal to fix #5263

### DIFF
--- a/lib/jekyll/static_file.rb
+++ b/lib/jekyll/static_file.rb
@@ -97,7 +97,7 @@ module Jekyll
 
     def to_liquid
       {
-        "filename"      => File.basename(name, extname),
+        "name"          => File.basename(name, extname),
         "extname"       => extname,
         "modified_time" => modified_time,
         "path"          => File.join("", relative_path)
--- a/lib/jekyll/static_file.rb
+++ b/lib/jekyll/static_file.rb
@@ -1,6 +1,6 @@
 module Jekyll
   class StaticFile
-    attr_reader :relative_path, :extname
+    attr_reader :relative_path, :extname, :name
 
     class << self
       # The cache of last modification times [path] -> mtime.
@@ -97,6 +97,7 @@ module Jekyll
 
     def to_liquid
       {
+        "filename"      => File.basename(name, extname),
         "extname"       => extname,
         "modified_time" => modified_time,
         "path"          => File.join("", relative_path)
--- a/lib/jekyll/static_file.rb
+++ b/lib/jekyll/static_file.rb
@@ -97,7 +97,8 @@ module Jekyll
 
     def to_liquid
       {
-        "name"          => File.basename(name, extname),
+        "basename"      => File.basename(name, extname),
+        "name"          => name,
         "extname"       => extname,
         "modified_time" => modified_time,
         "path"          => File.join("", relative_path)
--- a/site/_docs/static_files.md
+++ b/site/_docs/static_files.md
@@ -42,7 +42,15 @@ following metadata:
       <td><p><code>file.name</code></p></td>
       <td><p>
 
-        The string name of the file without the extension, e.g. <code>image</code> for <code>image.jpg</code>
+        The string name of the file e.g. <code>image.jpg</code> for <code>image.jpg</code>
+
+      </p></td>
+    </tr>
+    <tr>
+      <td><p><code>file.basename</code></p></td>
+      <td><p>
+
+        The string basename of the file e.g. <code>image</code> for <code>image.jpg</code>
 
       </p></td>
     </tr>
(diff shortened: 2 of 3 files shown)


lib/jekyll/static_file.rb
```diff
@@ -97,6 +97,7 @@ def write(dest)
 
     def to_liquid
       {
+        "filename"      => File.basename(name, extname),
```
DirtyF: Or simply use `name` in order to simple call `file.name` in liquid?

lib/jekyll/static_file.rb
```diff
@@ -1,6 +1,6 @@
 module Jekyll
   class StaticFile
-    attr_reader :relative_path, :extname
+    attr_reader :relative_path, :extname, :name
```
parkr: This is a short way of defining:

```ruby
def name
  @name
end
```

Therefore, the best thing to do here is to set `@name` in the `initialize` method using the code you have on line 100 and then, when you are constructing `to_liquid` below, you can just call the `name` method instead of re-calculating its value. Does that make sense?

parkr:
Well done, @DirtyF! This is great. I left a comment above which should help you get this ready for a merge. 😄 And well done adding a test! I would love to also see a test for the `name` `attr_reader` you added, too, just to make sure it keeps working as we expect.

🎉

lib/jekyll/static_file.rb
DirtyF: yes, but then it breaks the tests as 'name' used to be the whole name of the file including the extension.

envygeeks:
I'm not against this but the LGTM is being witheld since this is a breaking change and there is no backwards compatibility being kept until 4.0, can somebody explain to me what the ramifications of this change are?

DirtyF:
@envygeeks I didn't intend a breaking change, only wanted to add an extra Liquid shortcut to access the name of a static file without having to use liquid filters to remove the extra parts. I guess I'm doin' it wrong. 😞

envygeeks:
@DirtyF Well, the question really is, can I still access it the old way if I want to?

lib/jekyll/static_file.rb
```diff
@@ -97,6 +97,7 @@ def write(dest)
 
     def to_liquid
       {
+        "name"          => File.basename(name, extname),
```
parkr: I'm fine with adding this to Liquid. When you do `static_file.name`, it expects something else, though?

fene:
@envygeeks what do you mean 'the old way'? all i see in this change is additions, nothing breaking

lib/jekyll/static_file.rb
DirtyF: From what I understand, in the initialize function, name is the complete name of the file including the extension.

I just modified the output in Liquid, to be able to get only the name without the extension, already available as extname. I did not want to modify the current behavior, as it would break all current tests. 

As a total beginner, I have no clue how to properly achieve this. I just thought that we got to keep `name` as it is, and just add `filename` in the `to_liquid` function.

parkr:
Interesting, this change LGTM. I think the main issue was the possibility that `StaticFile#name` (in Ruby) would be changed if he implemented my suggestion.

@DirtyF, would you please leave a comment next to your addition to `to_liquid` explaining, essentially, that the value is to match the `name` value in `Page`?

DirtyF:
@parkr I don't get what you mean 😕 

* `{{ page.name }}` returns `about.md`
* `{{ file.name }`} would return `static_file` not `static_file.txt` (cf. use case in #5263)

I completed the documentation with explicit examples for liquid metadata available for static files.

spudowiar:
@DirtyF I'd suggest naming it `basename` (and providing the `name` property as well).

Other, unrelated note: would we see an improvement using Jekyll Drops for these things?

DirtyF:
@spudowiar you're right `basename` is far less confusing and more explicit.

Documentation updated to reflect the changes.

docs-static-files

envygeeks:
LGTM. /cc @jekyll/core

DirtyF:
@parkr yes it is.

parkr:
This should eventually become a Drop. For now, LGTM.

@jekyllbot: merge +minor

PR #5270, opened 2016-08-22
0.12.1 release

MAYAKARYA:


parkr:
Thanks! But we're leaving this branch unmerged as it conflicts badly and is very old. 😄

PR #5281, opened 2016-08-24
Site documentation section links always point to https://jekyllrb.com

Anthony-Gaudino:
On the site documentation section, links to documentation sections always point to the `jekyllrb.com` website, this means that users testing changes might get confused because they will see the official external website page instead of their local website upon clicking those links.

**Please check if this change doesn't break the links on the official website on `https://jekyllrb.com` before accepting the pull request.**

----------

@jekyll/documentation

### DIFF
--- a/site/_includes/docs_option.html
+++ b/site/_includes/docs_option.html
@@ -1,5 +1,5 @@
 {% for item in include.items %}
   {% assign item_url = item | prepend:"/docs/" | append:"/" %}
   {% assign doc = site.docs | where: "url", item_url | first %}
-  <option value="{{ site.url }}{{ doc.url }}">{{ doc.title }}</option>
+  <option value="{{ doc.url }}">{{ doc.title }}</option>
 {% endfor %}
--- a/site/_includes/docs_ul.html
+++ b/site/_includes/docs_ul.html
@@ -2,6 +2,6 @@
 {% for item in include.items %}
   {% assign item_url = item | prepend:"/docs/" | append:"/" %}
   {% assign p = site.docs | where:"url", item_url | first %}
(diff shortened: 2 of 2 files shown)


Anthony-Gaudino:
@jekyll/documentation

ashmaroli:
Hi..
I think there's another way around this issue. But, it requires an extra step from the testing user's end.
What they need to do is edit the url parameter inside **site/\_config.yml** from `https://jekyllrb.com` to `http://localhost:4000`

ashmaroli:
@parkr, jekyllbot commented twice here. Once without description, then again with proper description when another user (myself) commented. 
**Bug?**

DirtyF:
I guess we could use a `_development.yml` config file to override `_config.yml` `url` value but I don't know how to pass a --config option to the `[site.rake](https://github.com/jekyll/jekyll/blob/master/rake/site.rake#L25)`

Anthony-Gaudino:
Maybe the changes on this pull request won't do any harm, only the links on the documentation section have the `{{ site.url }}` and the changes only affect the list of links to documentation sections.

It's a good idea for someone that knows how the site is hosted to look at this pull request just to make sure this won't break the links on the official website.

Another option is to just accept the pull request and check the site after rebuild, if something is wrong, revert.

parkr:
This works fine for me! LGTM.

/cc @jekyll/documentation

parkr:
LOL, @jekyllbot. 😄

parkr:
@jekyllbot: merge +site

DirtyF:
@Anthony-Gaudino it's fine on http://jekyllrb.com/ ❤️  thanks!

PR #5293, opened 2016-08-25
Site template: exclude Gemfile and Gemfile.lock in site config

crispgm:
`Gemfile` and `Gemfile.lock` is generated after `jekyll new`, both of which should exclude in `_config.yml`. Otherwise, they can be downloaded from site.

### DIFF
--- a/lib/site_template/_config.yml
+++ b/lib/site_template/_config.yml
@@ -29,3 +29,6 @@ markdown: kramdown
 theme: minima
 gems:
   - jekyll-feed
+exclude:
+  - Gemfile
+  - Gemfile.lock


ashmaroli:
Can you explain the need to explicitly exclude `Gemfile` and `Gemfile.lock` ? Gemfile.lock is generated only after bundler runs.

crispgm:
@ashmaroli `Gemfile` and `Gemfile.lock` will be served in develop mode, and be generated to destination. However, they are used by develop, not what we really want to generate to our site.

parkr:
LGTM.

/cc @jekyll/ecosystem

mattr-:
LGTM

@jekyllbot: merge +minor

:boom:

mattr-:
i got the tag wrong, but ¯\\\_(ツ)\_/¯

envygeeks:
Yeah my first few times with it I got it wrong too.  It's whatever IMO.

PR #5295, opened 2016-08-28
Execute jekyll from clone instead of defined binary when running 'script/default-site'

ashmaroli:
While working on #5241, I noticed that its not required to run `jekyll new` from **exe/jekyll** binary and rather works well with Jekyll _cloned with the build_. 
I opened this PR to have all discussions and updates separate from the other PR.

/cc @jekyll/ecosystem
/cc @jekyll/windows

### DIFF
--- a/script/default-site
+++ b/script/default-site
@@ -10,7 +10,7 @@ rm -Rf ./tmp/default-site
 workdir=$(pwd)
 
 echo "$0: creating new default site"
-bundle exec ruby exe/jekyll new tmp/default-site
+bundle exec jekyll new tmp/default-site
 pushd tmp/default-site
 
 echo "$0: respecifying the jekyll install location"
--- a/script/default-site
+++ b/script/default-site
@@ -10,10 +10,9 @@ rm -Rf ./tmp/default-site
 echo "$0: creating new default site"
 bundle exec jekyll new tmp/default-site
 pushd tmp/default-site
-workdir="../../"
 
 echo "$0: respecifying the jekyll install location"
-ruby -e "contents = File.read('Gemfile'); File.write('Gemfile', contents.sub(/gem \"jekyll\".*\\n/, 'gem \"jekyll\", path: \"$workdir\"'))"
+ruby -e "contents = File.read('Gemfile'); File.write('Gemfile', contents.sub(/gem \"jekyll\".*\\n/, 'gem \"jekyll\", path: \"../../\"'))"
 echo "$0: installing default site dependencies"
 BUNDLE_GEMFILE=Gemfile bundle install
 echo "$0: building the default site"
--- a/script/default-site
+++ b/script/default-site
@@ -7,11 +7,10 @@ echo "$0: setting up tmp directory"
 mkdir -p ./tmp
 rm -Rf ./tmp/default-site
 
-workdir=$(pwd)
-
 echo "$0: creating new default site"
 bundle exec jekyll new tmp/default-site
 pushd tmp/default-site
+workdir="../../"
 
 echo "$0: respecifying the jekyll install location"
 ruby -e "contents = File.read('Gemfile'); File.write('Gemfile', contents.sub(/gem \"jekyll\".*\\n/, 'gem \"jekyll\", path: \"$workdir\"'))"


benbalter:
@ashmaroli Can you describe a bit more what you're trying to do here? I'm having trouble following the diff.

ashmaroli:
okay, I am running `bundle exec jekyll new` here instead of `bundle exec exe/jekyll new`
Additionally, changed `workdir=$(pwd)` to `workdir="../../"` 
I also removed `$0` from every `echo`

ashmaroli:
If its really necessary, I'll split the commit into bite-size ones.

XhmikosR:
First of all, I already have https://github.com/jekyll/jekyll/pull/5240 which includes the necessary part of this patch.

The rest of the changes seem unneeded to me at this point.

ashmaroli:
@XhmikosR, I'm aware of that PR. this PR deals with discussing `bundle exec jekyll new` instead of `bundle exec exe/jekyll`, and the relevance of `exe/jekyll` in this script. 
It has nothing to do with `bundle exec ruby exe/jekyll new`

XhmikosR:
Then just make the needed changes only instead of changing the log messages for example.

ashmaroli:
Roger that, Captain! :smile:

parkr:
> Additionally, changed `workdir=$(pwd)` to `workdir="../../"`

Out of curiosity, why did you need to make this change? The updated code in this PR is harder to follow in my opinion so I would love to know why it is needed. Thanks!

ashmaroli:
> why did you need to make this change?

@parkr, I'll try to explain things to the best of my ability.  
The main reason why you are unable to understand this is because you've personally not tested it out on a windows-machine per se. So I'll include references to Appveyor logs that you can verify.

value `$(pwd)` ***before*** `$0: creating new default site ` == ***/c/projects/jekyll***

say, I move assigning to `$(pwd)` further down (after `pushd temp/default-site`), then
`$(pwd)` == ***/c/projects/jekyll/tmp/default-site*** ( An incorrect path on Windows )

_But.._ when I assign workdir to simply be two directories up ( ***../../*** ) ***after*** `pushd tmp/default-site`, then 
in-theory, its a place where one can expect to find Jekyll gemspec, on any platform.

Rightly so, Both Appveyor and Travis exits green, after executing every line. 

This explained why I changed the value of `workdir`

---

Refs:
**Failed test in Appveyor**: https://ci.appveyor.com/project/jekyll/jekyll/build/386/job/vyjf304g5ovbkogn
**Passed test in Appveyor**: https://ci.appveyor.com/project/jekyll/jekyll/build/403/job/ryc5t0pkc9egovox

ashmaroli:
Now to explain the change in `bundle exec ... ...` line in this script and the very reason for this PR:

on Windows `bundle exec exe/jekyll new ...` fails as exe/jekyll is not identified as an executable. As of this writing, a patch has been applied to run `bundle exec ruby exe/jekyll new ... ` instead. And it works! So, excellent. But..

**my enquiry here is :** is having to run ***exe/jekyll executable*** really necessary for _this script_?

CI uses Jekyll at source . for every build. I tested to see if that source is sufficient to run this script.
Both Travis & Appveyor says, they're cool with executing `bundle exec jekyll new tmp/default-site` using Jekyll cloned at source . (They both passed this test).
Now, I need a human perspective to that observation.

ashmaroli:
Referencing CI Build Jobs as of latest commit 53f0404 ( before I remove merge-conflicts and rebase )

***Travis CI:*** https://travis-ci.org/jekyll/jekyll/jobs/155967689
***Appveyor CI:*** https://ci.appveyor.com/project/jekyll/jekyll/build/403/job/ryc5t0pkc9egovox

script/default-site
```diff
@@ -7,11 +7,10 @@ echo "$0: setting up tmp directory"
...
-workdir=$(pwd)
-
 echo "$0: creating new default site"
-bundle exec ruby exe/jekyll new tmp/default-site
+bundle exec jekyll new tmp/default-site
 pushd tmp/default-site
+workdir="../../"
 
 echo "$0: respecifying the jekyll install location"
 ruby -e "contents = File.read('Gemfile'); File.write('Gemfile', contents.sub(/gem \"jekyll\".*\\n/, 'gem \"jekyll\", path: \"$workdir\"'))"
```
parkr: Instead of using `$workdir` at all, what about just changing this line to use `path: \"../../\"`?

ashmaroli: Essentially, the same thing, but much better.
Removes the variable and shortens the line here by 2 characters :stuck_out_tongue_winking_eye: 

I'll update shortly.

parkr:
Seems to work! Thanks. LGTM.

envygeeks:
LGTM.

parkr:
@jekyllbot: merge +dev

PR #5304, opened 2016-08-30
Site: exclude README.md and .gitignore [ci skip]

crispgm:
😄

### DIFF
--- a/site/_config.yml
+++ b/site/_config.yml
@@ -32,3 +32,7 @@ gems:
   - jekyll-seo-tag
   - jekyll-avatar
   - jekyll-mentions
+
+exclude:
+  - README.md
+  - .gitignore


envygeeks:
I see no reason not to.  LGTM.

site/_config.yml
```diff
@@ -32,3 +32,7 @@ gems:
   - jekyll-seo-tag
   - jekyll-avatar
   - jekyll-mentions
+
+exclude:
+  - README.md
+  - .gitignore
```
parkr: Interesting! This should be removed automatically.

parkr:
LGTM.

site/_config.yml
crispgm: @parkr if these files were generated with `jekyll new` by default, we could add them to template of `_config.yml`. Repos with jekyll site have large possibilities contain such files as `README.md`, `LICENSE`, `Rakefile`, `.gitignore`, etc.

ashmaroli: @crispgm, currently the files [here](https://github.com/jekyll/jekyll/tree/master/lib/site_template) plua a Gemfile are copied to a new site's root by `jekyll new`

ashmaroli: so, is there really a need to have this exclusion, by default?

envygeeks:
@jekyllbot: merge +site

ashmaroli:
@DirtyF, I cant find the documentation that explains the principle behind the **exclude array** in a _config.yml. Please help.

envygeeks:
@ashmaroli the exclude array prevents a file from being written.  So say you don't want your Gemfile, or README.md or vendor/ folder to end up being processed, you exclude them and they aren't processed or written. On a side note: I just seen your comment, sorry I merged before we talked about what you wanted to talk about.

See: https://jekyllrb.com/docs/configuration/#global-configuration

ashmaroli:
@envygeeks Thats alright, nothing serious, simply wondering aloud

site/_config.yml
parkr: @crispgm Yes! I mean that Jekyll should ignore it when building. See [EntryFilter](https://github.com/jekyll/jekyll/blob/b5ee87b4ca91fefe86c42526fee230a3c0294eca/lib/jekyll/entry_filter.rb#L4-L6).

crispgm: @parkr sounds that `.gitignore` should be exclude by `EntryFilter`. I will inspect this later.

PR #5052, opened 2016-07-03
rubocop: lib/jekyll/renderer.rb complexity fixes

ayastreb:
#4885

In this PR I refactored `Renderer` class, removed duplicated code from `Convertible` and changed `Page` class to use `Renderer.run` instead of `Convertible.do_layout`

PR #5308, opened 2016-08-30
Proxy a number of Convertible methods to Renderer

parkr:
This simplifies the code a bit -- much less duplication.

We might consider providing a way to set the layouts for a Renderer so we can override it if necessary. It seems a bit aggressive to want/have to change `site.layouts` in order for this to work.

#4885

### DIFF
--- a/lib/jekyll/convertible.rb
+++ b/lib/jekyll/convertible.rb
@@ -190,7 +190,10 @@ module Jekyll
     #
     # Returns nothing
     def render_all_layouts(layouts, payload, info)
+      _renderer.layouts = layouts
       _renderer.place_in_layouts(output, payload, info)
+    ensure
+      @_renderer = nil # this will allow the modifications above to disappear
     end
 
     # Add any necessary layouts to this convertible document.
@@ -200,11 +203,16 @@ module Jekyll
     #
     # Returns nothing.
     def do_layout(payload, layouts)
-      @_renderer = Jekyll::Renderer.new(site, self, payload)
-      _renderer.run
+      _renderer.tap do |renderer|
+        renderer.layouts = layouts
+        renderer.payload = payload
+        renderer.run
+      end
 
       Jekyll.logger.debug "Post-Render Hooks:", self.relative_path
       Jekyll::Hooks.trigger hook_owner, :post_render, self
+    ensure
+      @_renderer = nil # this will allow the modifications above to disappear
     end
 
     # Write the generated page file to the destination directory.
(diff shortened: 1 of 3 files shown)


parkr:
I am indebted to #5052 which gave me the idea for this.

lib/jekyll/renderer.rb
```diff
@@ -2,20 +2,41 @@
 
 module Jekyll
   class Renderer
-    attr_reader :document, :site, :payload
+    attr_reader :document, :site
+    attr_writer :layouts, :payload
 
-    def initialize(site, document, site_payload = nil)
+    def initialize(site, document, site_payload = nil, layouts: nil)
```
parkr: @envygeeks Are named parameters not allowed in jruby? https://travis-ci.org/jekyll/jekyll/jobs/156357192

parkr:
Works! Turns out jruby doesn't like named parameters. https://github.com/jruby/jruby/issues/1403

envygeeks:
Do you mean keywords @parkr? Because JRuby has full support for keywords.  What error were you getting?

parkr:
@envygeeks Yeah, I thought that too and the error was weird. The build I saw it in was https://travis-ci.org/jekyll/jekyll/jobs/156357192 and it was fixed by 4af0f0262ec119fd2a6049b236bba97314b2976e.

envygeeks:
I'll look at the error but what I do know is we use keywords in JRuby (we
mostly deploy with JRuby.)

parkr:
@jekyllbot: merge +dev

lib/jekyll/convertible.rb
```diff
@@ -76,34 +78,23 @@ def validate_permalink!(filename)
...
-      Jekyll::Renderer.new(site, self).output_ext
+      _renderer.output_ext
     end
 
     # Determine which converter to use based on this convertible's
     # extension.
     #
     # Returns the Converter instance.
     def converters
-      @converters ||= site.converters.select { |c| c.matches(ext) }.sort
```
parkr: Hey! It's been ages since you commented, but I'm circling back now. A new PR (or an issue) will definitely get more of a response than a comment on a really old PR. But here we are!

So you're saying we took away `extname` and everything exploded?

joallard: Hey @parkr, sorry about that! I was on the hunt for what prevented Jekyll-Haml from working, but I've moved on since. Looks like they ended up finding a way!

PR #5309, opened 2016-08-31
Add Gemfile template

ashmaroli:
- moving the contents of the Gemfile to an ERB template of its own reduces clutter in new.rb and allows easier edits to those contents in future.
- ~~rename create_sample_files to create_static_files~~
- also update the test file to assert minima gem in Gemfile.

### DIFF
--- a/lib/jekyll/commands/new.rb
+++ b/lib/jekyll/commands/new.rb
@@ -23,6 +23,7 @@ module Jekyll
           raise ArgumentError, "You must specify a path." if args.empty?
 
           new_blog_path = File.expand_path(args.join(" "), Dir.pwd)
+
           FileUtils.mkdir_p new_blog_path
           if preserve_source_location?(new_blog_path, options)
             Jekyll.logger.abort_with "Conflict:",
@@ -49,6 +50,15 @@ module Jekyll
           ERB.new(File.read(File.expand_path(scaffold_path, site_template))).result
         end
 
+        def create_gemfile(path)
+          gemfile_template = File.expand_path("Gemfile.erb", site_template)
+          gemfile_copy = ERB.new(File.read(gemfile_template)).result(binding)
+
+          File.open(File.expand_path("Gemfile", path), "w") do |f|
+            f.write(gemfile_copy)
+          end
+        end
+
         # Internal: Gets the filename of the sample post to be created
         #
         # Returns the filename of the sample post, as a String
@@ -58,44 +68,11 @@ module Jekyll
 
         private
 
-        def gemfile_contents
-          <<-RUBY
-source "https://rubygems.org"
-ruby RUBY_VERSION
-
-# Hello! This is where you manage which Jekyll version is used to run.
-# When you want to use a different version, change it below, save the
-# file and run `bundle install`. Run Jekyll with `bundle exec`, like so:
-#
-#     bundle exec jekyll serve
-#
-# This will help ensure the proper Jekyll version is running.
-# Happy Jekylling!
-gem "jekyll", "#{Jekyll::VERSION}"
-
-# This is the default theme for new Jekyll sites. You may change this to anything you like.
-gem "minima"
-
-# If you want to use GitHub Pages, remove the "gem "jekyll"" above and
-# uncomment the line below. To upgrade, run `bundle update github-pages`.
-# gem "github-pages", group: :jekyll_plugins
-
-# If you have any plugins, put them here!
-group :jekyll_plugins do
-   gem "jekyll-feed", "~> 0.6"
-end
-RUBY
-        end
-
         def create_site(new_blog_path)
           create_sample_files new_blog_path
 
           File.open(File.expand_path(initialized_post_name, new_blog_path), "w") do |f|
             f.write(scaffold_post_content)
-          end
-
-          File.open(File.expand_path("Gemfile", new_blog_path), "w") do |f|
-            f.write(gemfile_contents)
           end
         end
 
@@ -103,8 +80,17 @@ RUBY
           !options["force"] && !Dir["#{path}/**/*"].empty?
         end
 
+        def erb_files
+          erb_file = File.join("*", "*.erb")
+          Dir.glob(erb_file)
+        end
+
         def create_sample_files(path)
           FileUtils.cp_r site_template + "/.", path
+          create_gemfile path
+          erb_files.each do |file|
+            FileUtils.rm file
+          end
           FileUtils.rm File.expand_path(scaffold_path, path)
         end
 
--- a/lib/jekyll/commands/new.rb
+++ b/lib/jekyll/commands/new.rb
@@ -50,15 +50,6 @@ module Jekyll
           ERB.new(File.read(File.expand_path(scaffold_path, site_template))).result
         end
 
-        def create_gemfile(path)
-          gemfile_template = File.expand_path("Gemfile.erb", site_template)
-          gemfile_copy = ERB.new(File.read(gemfile_template)).result(binding)
-
-          File.open(File.expand_path("Gemfile", path), "w") do |f|
-            f.write(gemfile_copy)
-          end
-        end
-
         # Internal: Gets the filename of the sample post to be created
         #
         # Returns the filename of the sample post, as a String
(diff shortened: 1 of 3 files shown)


lib/site_template/Gemfile.erb
```diff
@@ -0,0 +1,24 @@
...
+# This will help ensure the proper Jekyll version is running.
+# Happy Jekylling!
+gem "jekyll", "~> <%= Jekyll::VERSION %>"
+
+# This is the default theme for new Jekyll sites. You may change this to anything you like.
+gem "minima"
+
+# If you want to use GitHub Pages, remove the "gem "jekyll"" above and
+# uncomment the line below. To upgrade, run `bundle update github-pages`.
+# gem "github-pages", group: :jekyll_plugins
```
envygeeks: No.

lib/jekyll/commands/new.rb
```diff
@@ -57,53 +67,29 @@ def initialized_post_name
...
-# If you have any plugins, put them here!
-group :jekyll_plugins do
-   gem "jekyll-feed", "~> 0.6"
-end
-RUBY
-        end
-
         def create_site(new_blog_path)
-          create_sample_files new_blog_path
+          create_static_files new_blog_path
```
envygeeks: This is not Jekyll 4, please put back it's original name.

ashmaroli:
Initially planned to be pushed after merge of #5236 as it dealt with an erb template for config file. ( So, there will be a few overlaps. )
Now, refitted to be presented as a standalone PR.

lib/site_template/Gemfile.erb
ashmaroli: No?

lib/jekyll/commands/new.rb
ashmaroli: okay..

envygeeks:
It seems to me that with the addition of all thees ERB files, there needs to be a proper system to deal with all this, so far this is turning into a nightmare.  If there is no system to deal with an ulimited amount of templates with certain data being passed in and out (or otherwise) and each one is a method, then the new system is no better than just having the data directly inside of methods if you ask me.  We've not reduced complexity, we've only shifted it to be more confusing if you ask me.

/cc @jekyll/core

lib/site_template/Gemfile.erb
envygeeks: Sorry, that seemed kind of blunt and rude.  Github Pages is it's own project independent of Jekyll and they should deal with this themselves (and we should have a system that lets them embed their own templates into Jekyll...) That said, Github Pages is not a Jekyll plugin, you choose either Github Pages or you choose Jekyll.  One replaces the other.

My vote moves towards allowing authors of gems like github-pages or our own Jekyll gem hosted on our private servers to provide their own base templates rather than including stuff that isn't a concern of core into core.

lib/jekyll/commands/new.rb
envygeeks: Again I seemed rude without intending to be.  We've already tied into this API a bit, which means we expect it to remain stable, and that also means that I would assume other users might have already tied into it and expect it to remain stable.

API changes like this should be left to major versions rather than minor versions.

ashmaroli:
> this is turning into a nightmare.

nightmare for maintainers or for end-users?

> We've not reduced complexity, we've only shifted it to be more confusing if you ask me.

I feel once all the templates are set, one solid refactoring should DRY things up. ,IMHO

lib/site_template/Gemfile.erb
ashmaroli: The entire contents here were simply _moved_ from new.rb to gemfile.erb with minor adaptations.

envygeeks:
> nightmare for maintainers or for end-users?

I make no differentiation when it comes to open-source.

> I feel once all the templates are set, one solid refactoring should DRY things up. ,IMHO

Procrastination is the evil that destroys all good software.

lib/site_template/Gemfile.erb
envygeeks: That doesn't change my point :wink:

envygeeks: And looking at the diff, I don't see that it was originally there, my point being that it shouldn't be there, we should offer a system for them to add their own Gemfile to override ours not add their gem when their gem is a replacement for ours with additions and their Gem is not a concern to core.

ashmaroli:
> Procrastination is the evil that destroys all good software.

:heart:  quote :+1: 
should make a banner out of it.. 

Humor apart, to make things more complicated, was never my intention. But I guess, I did not look at the bigger picture when I thought about all these edits.

envygeeks:
> Humor apart, to make things more complicated, was never my intention. But I guess, I did not look at the bigger picture when I thought about all these edits.

I don't think you mean to make it more complicated, sorry if I implied that! I think what you're doing is great, but at the same time, I can see a disaster in the making so I expressed that. And that isn't your fault.  I don't think this pull should be stopped either, however, I do think that if these types of moves are to continue we should definitely fix the issue of streaming lining these edits through a better designed system that can handle them.

lib/site_template/Gemfile.erb
ashmaroli: > And looking at the diff, I don't see that it was originally there...

I understood your point clearly.
This comment is only to show, that those comments were [added months ago](https://github.com/jekyll/jekyll/commit/09f9f193d8e5fb714a6852c9276be1c6e55efbe8).

lib/jekyll/commands/new.rb
```diff
@@ -57,53 +67,29 @@ def initialized_post_name
...
+        def erb_files
+          erb_file = File.join("*", "*.erb")
+          Dir.glob(erb_file)
+        end
+
         def create_sample_files(path)
           FileUtils.cp_r site_template + "/.", path
+          create_gemfile path
+          erb_files.each do |file|
+            FileUtils.rm file
```
parkr: Would make more sense to, in that `cp_r` command, simply exclude the `*.erb` files.

ashmaroli: `site_template` is actually a string object. To exclude specific files, I'll need to use `Dir.glob` which wont include `.gitignore` so that would require adding File::FNM_DOTMATCH constant and the whole step chained to another step to copy those files.
I hence feel adding an extra method `erb_files` is not so wrong.

envygeeks: Then Dir.glob properly.

parkr: @ashmaroli I realize that it's not easy. That said, filtering out before the files are copied is preferable to copying the whole directory and trying to remember what to delete. Always good to save on FS operations, too. If you don't mind, I'd love to see a `site_template_files` method which returns the value of `Dir.glob` for the `site_template` directory properly excluding `.erb` files and so forth. It should be relatively simple:

```ruby
def site_template_files
  Jekyll::Utils.safe_glob(site_template, ["**", "*"], File::FNM_DOTMATCH).reject do |file|
    file.end_with?(".erb")
  end
end
```

ashmaroli: > If you don't mind, I'd love to see a `site_template_files` method which returns the value of `Dir.glob` for the `site_template` directory properly excluding `.erb` files and so forth.

@parkr I dont mind at all, infact I learned of the `safe_glob()` method only after seeing the suggested code above. So you're most welcome to point me to the right directions.
The problem with that `site_template_files` method is since it returns all the dotfiles (incl. ".", ".."), the `cp_r` method copies `site_template/.`, `site_template/..` to the `new_blog_path` as well thereby rendering the new method pointless

parkr: We have [a thing for that!](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/entry_filter.rb):

```ruby
def site_template_files
  filter = EntryFilter.new(nil)
  Jekyll::Utils.safe_glob(site_template, ["**", "*"], File::FNM_DOTMATCH).reject do |file|
    file.end_with?(".erb") || filter.special?(file) || filter.backup?(file) || filter.symlink?(file)
  end
end
```

ashmaroli: @parkr, thanks once again. I've enabled *access_to_maintainers* for this branch, you can legitimately commit to this PR if you have the time. 

--
**P.S. :**  The modified `site_template_files` method (with EntryFilter) is throwing the following:
**Error:  undefined method `source' for nil:NilClass** 
(`source` in that error opens with a backtick but closes with a single-quote)

ashmaroli:
@parkr, I changed the overall code (related to just this PR) based on ThemeBuilder. This avoids copying the erb file(s) first and then deleting them. 
(Moreover, this opens a way to turn _config.yml to an ERB template if my other PR gets approval as well..; and also for a new Class that may be easily introduced to handle (similar) methods in `Commands::New` and `ThemeBuilder`)

Requesting a fresh review..

ashmaroli:
closing.. as this has been moved to another gem [`jekyll-plus`](https://github.com/ashmaroli/jekyll-plus) entirely.

Issue #5311, opened 2016-08-31
[Proposal] [Discussion] Theme builder should generate minima 1.0.1 files by default


ashmaroli:
Hi, 
As absurd as it sounds, I think it would be a lot more helpful to the front-end-novice when he is provided with a starting point. And that starting point is our _minima_.
Currently, only layout files (barely-there-barebones) are provided. _includes and _sass dirs are simply empty.

While he could copy those files from his gems_folder, its a lot easier to run `jekyll new-theme xyz`

files under _layouts and _includes are essentially static, hence could remain the same.  What can be treated as dynamic are the contents of _sass dir.
It contains a minima.scss and a minima dir with three partials.

In theory, Jekyll should be able to easily rename them using value from `theme.name` to **/_sass/theme_name.scss** and **/_sass/theme_name/[files]** at the same time rename **@ import** instances programmatically.

parkr:
This defeats the purpose of gem-based themes, so I disagree that this should be the case by default.

I was thinking about proposing a command like `jekyll copy-theme-file _includes/head.html` that copies the contents of the theme's `_includes/head.html` into your site's `_includes/head.html` so you can begin to modify it with just one command. Would that satisfy your needs?

spudowiar:
@parkr I think @ashmaroli meant that when you create a theme it should be pre-populated with the files for [minima](https://github.com/jekyll/minima)

parkr:
@spudowiar You're right! My bad. I agree that a better starting point for `jekyll new-theme` is a great idea. Want to put together a pull request of the pieces you think we need and we can talk them over? I like the idea of starter sass, though I'm not sure about the includes. 😄

jaybe-jekyll:
Basic {% include %}' such as required HTML, snippets, structure, etc. are valuable, sensible, and introduces the user to key (while advanced) features of the engine and ways to be more-organized and DRY.

ashmaroli:
> I think @ashmaroli meant that when you create a theme using jekyll new-theme it should be pre-populated with the files for minima

@spudowiar, Thanks for simplifying the case into a single sentence :smiley:

jaybe-jekyll:
https://github.com/jekyll/jekyll/pull/5317#issuecomment-244446015

Scenario:

I want to quickly spin up a new, default jekyll theme (which is minima) as my own, to start hacking on right away, locally; I don't want a gem-based theme at this moment-- I want it local and in my new jekyll source.

I'd like to, i.e.:

[...]

PR #5317, opened 2016-09-02
add minima files to theme template

ashmaroli:
Ref: #5311 
adds minima-1.0.1 files to theme_template.
extends existing methods.

@jekyll/ecosystem


pathawks:
If this is the best way to start a new theme, the workflow should be to just go download the latest version of minima and start hacking away.

Including all those files inside of Jekyll just seems like madness. :-1:

jaybe-jekyll:
Scenario:

I want to quickly spin up a new, default jekyll theme (which is minima) as my own, to start hacking on right away, locally; I don't want a gem-based theme at this moment-- I want it local and in my new jekyll source.

I'd like to, i.e.:
- `jekyll new --local mylocaltheme`

I'd like that to produce a new directory, `./mylocaltheme`, containing `jekyll new` defaults (sans `theme: minima` in _config.yml) along with local copies of minima gem and related directories/files (includes, etc.) within `./mylocaltheme`.

I `cd ./mylocaltheme` and begin configuring/changing my new theme source starting point.

Ref: #5311

spudowiar:
@pathawks I disagree, `minima` itself comes with a lot of extra stuff, this is a reasonable middle ground

jaybe-jekyll:
@spudowiar For clarity, I meant to suggest including/copying locally only the relevant and minimally-needed minima jekyll theme files and folders, not the entire gem contents.

spudowiar:
Yeah, that's my only qualm, I can imagine that we should make `jekyll` soft depend on `minima` and pull the files from the Gem itself

spudowiar:
@jaybe-jekyll sorry, didn't see your comment before, I understand what you mean and that's what I imagine too

jaybe-jekyll:
Indeed; depending on minima gem/theme implements and promotes using themes, to @parkr point, while providing the user a way to easily "clone" the [current and latest] version/code of the minima theme in their source workspace.

On deck/heads up/next:

> I created an awesome theme with this new command and functionality, now I want to create/bundle/package my theme into a gem so it can be a shared and installable jekyll gem-based theme! How do I do that?

ashmaroli:
> So, will any change to these files in the future require a PR here in addition to a PR for minima?

My intention with this PR was to simply have the files included with `jekyll new-theme` I did consider the above situation when opening the ticket to discuss this. Hence included the version number in the title.
i.e. **my intention was to have minima-1.0.1 files preloaded in the new theme folder when i ran** `jekyll new-theme xyz` **because it provided the necessary starting point for development.**
Minima has since been given new features and grown slightly :wink:  but that's Minima. Let it grow. Jekyll NewTheme need not concern itself with that...

If users do want to fork off the lastest Minima or any other theme, then I think the following may be a nice approach: (but beyond the scope of current PR.)
assume myTheme == Minima or any other theme, then
```
jekyll new-theme NEW THEME --develop myTheme
```
would fetch [directories] from that gem.

spudowiar:
@ashmaroli if we have a feature like this, this seems redundant so I'm currently :-1: on this

ashmaroli:
> if we have a feature like that, this seems redundant

By feature, if you meant this line: `jekyll new-theme NEW THEME --develop myTheme` 
its only a concept. Till that becomes a reality, this PR may be a welcome change.

lib/jekyll/theme_builder.rb
```diff
@@ -56,12 +56,27 @@ def write_file(filename, contents)
...
     mkdir_p(SCAFFOLD_DIRECTORIES)
+    mkdir_p("_sass/#{name}")
   end
 
   def create_starter_files
     %w(page post default).each do |layout|
       write_file("_layouts/#{layout}.html", template("_layouts/#{layout}.html"))
     end
+
+    %w(footer head header).each do |incl|
```
parkr: I personally don't like `head` and `header` because they cause confusion. Maybe `banner` or `top` for the user-facing heading

lib/jekyll/theme_builder.rb
```diff
@@ -56,12 +56,27 @@ def write_file(filename, contents)
...
+      write_file("_includes/#{incl}.html", template("_includes/#{incl}.html"))
+    end
+
+    %w(github twitter).each do |icon|
+      write_file("_includes/icon-#{icon}.svg", template("_includes/icon-#{icon}.svg"))
+      write_file("_includes/icon-#{icon}.html", template("_includes/icon-#{icon}.html"))
+    end
```
parkr: I don't want these assets in the default theme. I think they're superfluous.

lib/jekyll/theme_builder.rb
```diff
@@ -56,12 +56,27 @@ def write_file(filename, contents)
...
+
+    write_file("_sass/#{name}.scss", template("_sass/theme.scss.erb"))
+    %w(base layout syntax-highlighting).each do |sass|
+      write_file("_sass/#{name}/_#{sass}.scss", template("_sass/theme/_#{sass}.scss"))
```
parkr: Why include `base` and `layout`? I like `syntax-highlighting` though – nice power feature in a way. Folks end up copying that from Rouge or something in the end anyway.

lib/jekyll/theme_builder.rb
ashmaroli: I never liked that choice of naming either. But I kept it on the assumption that:
- Joel must have named them according to corresponding HTML5 elements.
- The community never objected to it till now. (Theme developers must've been copying from them minima gem dir anyway. )

I'll change after other maintainers / captains weigh in on this..

lib/jekyll/theme_builder.rb
ashmaroli: I added this because, I'm copying all required files from `minima` without any change.
Since, minima's footer include references these assets, I added them as well.

lib/jekyll/theme_builder.rb
ashmaroli: again, `base` and `layout` are `minima` files. So.. added them too.

ashmaroli:
@parkr, while this PR deals with adding vanilla `minima` files to `new-theme` template, I'm open to changing the direction to include `minima`**-esque** files to the template.

[Awaiting feedback from others in this conversation..]

lib/jekyll/theme_builder.rb
parkr: What do you think, @jekyll/ecosystem?

spudowiar: @parkr I've always used `head`, `header` and `footer` since they match the HTML elements. I'm sure most who write themes for Jekyll will recognise this.

PR #5319, opened 2016-09-03
Accept a base theme for `new-theme` to import files from

### LINKED ISSUE #5311: [Proposal] [Discussion] Theme builder should generate minima 1.0.1 files by default

spudowiar:
# Pull Request

This provides two features to resolve #5311

* The ability to create a theme based on another theme
* Documentation in `Gemfile` and a few tweaks to allow creating a site with a custom theme (and valid Gem) in `_theme` based on another theme

# Usage

Create a theme based on Minima called `custom-theme` in the directory `custom-theme`
```
bundle exec jekyll new-theme -b minima custom-theme
```

Create a site using a mutable theme. This creates a theme based on Minima in `_theme` called `custom-theme` and sets it as the Jekyll theme. When the user is ready to release their theme, they can do so with ease and `_theme` is a valid Gem (since they used `new-theme`)
```
bundle exec jekyll new .
bundle exec jekyll new-theme -b minima custom-theme _theme
echo 'gem "custom-theme", path: "_theme"' >> Gemfile
sed 's/minima/custom-theme/g' -i _config.yml
```
The latter case has been documented in the `Gemfile`

# FIXME

* Needs tests
* Needs documentation for `jekyllrb`


lib/jekyll/commands/new.rb
```diff
@@ -75,6 +75,15 @@ def gemfile_contents
 # This is the default theme for new Jekyll sites. You may change this to anything you like.
 gem "minima"
 
+# If you want to develop a custom theme, first create a new theme, like so
+#
+#     bundle exec jekyll new-theme -b minima --name custom-theme _theme
```
ashmaroli: wont it work w/o `bundle exec` prefix?

spudowiar: We call all `jekyll` commands with `bundler` since this is documentation... in a `Gemfile` :smile:

lib/jekyll/commands/new_theme.rb
```diff
@@ -22,7 +26,8 @@ def process(args, opts)
         raise Jekyll::Errors::InvalidThemeName, "You must specify a theme name."
       end
 
-      new_theme_name = args.join("_")
+      new_theme_name = args.first
+      opts["theme_path"] = args[1..-1].join(" ") if args.size > 1
```
ashmaroli: suppose I enter **Long Island** as the theme_name.. how would this translate as the .gemspec filename?
**long_island.gemspec** or **long island.gemspec**?

ashmaroli: Nope. As of now, local tests show this does not happen.

spudowiar: Ah, I see the reasoning for the old behaviour now, let me amend this commit :smile:

On 3 Sep 2016 14:25, ashmaroli <notifications@github.com> wrote:

In lib/jekyll/commands/new_theme.rb<https://github.com/jekyll/jekyll/pull/5319#discussion_r77436860>:

> @@ -22,7 +26,8 @@ def process(args, opts)
>          raise Jekyll::Errors::InvalidThemeName, "You must specify a theme name."
>        end
>
> -      new_theme_name = args.join("_")
> +      new_theme_name = args.first
> +      opts["theme_path"] = args[1..-1].join(" ") if args.size > 1


suppose I enter Long Island as the theme_name.. how would this translate as the .gemspec filename?
long_island.gemspec or long island.gemspec?

envygeeks: It should be dashes not underscores as is the most common, and it should never be spaces.

spudowiar: @envygeeks I agree with your point entirely, I was aiming for consistency with `Jekyll::Commands::New` (which uses spaces in the directory name) and I didn't fully understand this command's existing logic. @ashmaroli's example has explained it to me :smile:

spudowiar:
@ashmaroli @envygeeks rebased, should settle your qualms for now :smile:

ashmaroli:
Can we have the c.options listed alphabetically too? :smile:  Would be better IMHO.

ashmaroli:
I also think tests should be written to test theme_builder. But let's wait for @parkr's opinion on that.

spudowiar:
> Can we have the c.options listed alphabetically too?

@ashmaroli Rebased.

ashmaroli:
the options are now:
```
-o, --output
-b, --base
-c, -code-of-conduct
```
I meant:
```
-b, --base
-c, -code-of-conduct
-o, --output
```

DirtyF:
@ashmaroli I'd rather see the options listed in a logical, i.e you start from a base theme, you specify the ouput and occasionnaly a code of conduct.

```
-b, --base
-o, --output
-c, -code-of-conduct
```

pathawks:
How is this better than `git clone https://github.com/jekyll/minima.git custom-theme`?

spudowiar:
@DirtyF I think `--output` is logically first since you may just want `jekyll new-theme custom-theme -o _theme`, after that I think they should be in alphabetical order

@pathawks This is generated as a user's theme rather than just cloning and modifying themes. I'm imagining the future possibility of passing all the configuration options to the `new-theme` command (e.g. summary, URL, etc.) and you can have a ready to go theme in a second :smile:

parkr:
Hm, to me this does not resolve #5311 but is entirely different. My initial comment (as you referenced) was due to a misinterpretation on my part of the original post of that issue. I agree a `jekyll copy-from-theme` command or something would be useful, but this PR goes way beyond that.

benbalter:
> The ability to create a theme based on another theme

I'm 110% on board with the problem. I'm not sure this is the right solution. I think there we're actually trying to solve two problems here:

1. Give me a boilerplate theme, not to use, but to build my theme on. This is similar to `bundle gem` or `rails generate` in concept. For this, the folder structure, blank layouts, and a Gemspec should be enough.

2. I want to base my theme on another theme and actually use that theme's files. For this https://github.com/jekyll/jekyll/issues/4510 hinted at the idea of having theme dependencies, in which Theme A can add theme B as a dependency, sass includes would work as expected, but if a file didn't exist in theme A, we'd look to theme B.

If I'm reading the diff correctly, I'm not sure that hard-coding the files from Minima (which is intended to be an end-user theme, not a boilerplate), is the best solution to either problem, and gets us further from the idea of not hard-coding layouts into each site (because then a bug in Minima would be propagated to every child theme, without the ability to fix it downstream).

spudowiar:
@parkr People were talking about having this command, I implemented it, just in case. I wasn't trying to implement your comment, if you see my initial comment I said (extra emphasis added)

> This is **_not_** intended for one-off changes to a theme file, that should be provided by jekyll copy-from-theme or a similar command [...]

I agree that hierarchical themes seem like a great idea (as I said in #4510), I also think that something to this effect might be useful. If you wanted to create a theme based on another theme and, over time, modify it, this might be a nice solution.

Issue #5323, opened 2016-09-04
theme.gemspec doesn't build theme gem properly 


ashmaroli:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [x] This is **not** a feature request.

---

- [x] I am on (or have tested on) ***Windows 7***

---

- [x] I was trying to build the theme gem.

## My Reproduction Steps

- Create new theme, **titan**:
```
$ jekyll new-theme titan
```
- cd into theme dir and add a file each to empty subdirs.
- edit *titan.gemspec* to prepare for building gem
- update git index by running:
```
$ git add .
```
- Build gem.
```sh
$ gem build titan.gemspec

  Successfully built RubyGem
  Name: titan
  Version: 0.1.0
  File: titan-0.1.0.gem
```
- unpack gem contents to verify build before publishing:
```sh
$ gem unpack titan-0.1.0.gem
Unpacked gem: ~/titan-0.1.0
```
- `cd` into the unpacked dir and inspect.

## The Output I Wanted

see the necessary files in the unpacked directory.

## The Output I Got

an empty directory which means my theme gem was essentially empty.

---
I reached a solution by simply changing the `spec.files` line in the gemspec to:
```ruby
  spec.files  = `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
```
and additionally adding a file each to empty dirs.

---
/cc @jekyll/ecosystem

benbalter:
> I reached a solution by simply changing the spec.files line in the gemspec to:

@ashmaroli this is a great bug report, thank you.

Any interested in submitting a pull request to make that change?

/cc @jekyll/ecosystem

ashmaroli:
@benbalter thanks :smiley:  I've already opened a PR at #5325.

benbalter:
See an alternative approach for minima via https://github.com/jekyll/minima/pull/31 (and related issue https://github.com/jekyll/minima/issues/29.

ashmaroli:
Fixed in #5325

PR #5325, opened 2016-09-05
update gemspec template to include only required theme files

ashmaroli:
Ref: #5323 

Modify gemspec template to include all non-empty directories within the theme gem.

Currently, trying to build with the generated theme.gemspec results in an empty gem (at least on Windows 7 + Ruby 2.3.1 + RubyGems 2.6.6)

### DIFF
--- a/lib/theme_template/theme.gemspec.erb
+++ b/lib/theme_template/theme.gemspec.erb
@@ -10,7 +10,7 @@ Gem::Specification.new do |spec|
   spec.homepage      = "TODO: Put your gem's website or public repo URL here."
   spec.license       = "MIT"
 
-  spec.files         = `git ls-files -z`.split("\x0").select { |f| f.match(%r{^(<%= theme_directories.join("|") %>|LICENSE|README)/i}) }
+  spec.files         = `git ls-files -z`.split("\x0").select { |f| f.match(%r{^(<%= theme_directories.join("|") %>|LICENSE|README)}i) }
 
   spec.add_development_dependency "jekyll", "~> <%= jekyll_version_with_minor %>"
   spec.add_development_dependency "bundler", "~> 1.12"


benbalter:
> Modify gemspec template to include all non-empty directories within the theme gem.

We want to be a bit more strict here. We don't want to include, e.g., test files, example files, etc. Themes must conform to a specific spec, and we can lean on that (with the proper regex) to enforce that spec.

benbalter:
@jekyll/ecosystem would it be possible to include the regex as a constant as e.g., `Jekyll::Theme::REGEX`? Would that crete a circular dependency? That way we can enforce the theme regex upstream.

ashmaroli:
I'm opting to have the gem-build **reject directories / files we don't need to include with the gem**, over **including  a set of folders**, which would **_require a complex regex_**.

benbalter:
@ashmaroli I think this is an example of where we (the Jekyll maintainers) want to absorb the complexity (e.g., maintain a complex regex), rather than push that complexity on theme maintainers or end users. 

We want the default to be a whitelist, not a blacklist. In the above example, what if I use rspec, and have a `spec` folder instead of a `test` folder? Or accidentally include a `vendor` folder, which has a handful of large, unusable assets? We have an opportunity to use the default Gemspec (and hopefully if we can include it by reference, all theme gemspecs), as an enforcement mechanism to ensure themes conform to a shared specification (and help theme creators to do so). Put another way, we can tell computers to do some of the heavy lifting, rather than expecting humans to.

parkr:
> We want the default to be a whitelist, not a blacklist.

I agree here – a whitelist would be more useful. We use `ThemeBuilder#write_file` to write files so you could hook into that, or simply add a new method to the `ThemeBuilder::ERBRenderer` for these other scaffold files and include them in the regexp. 👍

benbalter:
> for these other scaffold files and include them in the regexp.

@parkr would it create a circular dependency for the theme gemspec to require Jekyll (or a specific jekyll file) and include the Regex by reference?

ashmaroli:
@benbalter, Suppose if the regex is made a constant, how do we modify it as required (Constant?) ?
say.., I want to include an `assets` folder as well in my theme gem.. How would that work?

benbalter:
> how do we modify it as required (Constant?) ?
say.., I want to include an assets folder as well in my theme gem.. How would that work?

@ashmaroli You can't, and that's the idea. Jekyll is creating a standard for what a theme should look like, and is enforcing that standard by deciding what's distributed with the compiled Gem. 

You can add a `/assets` folder to your Gem, but since Jekyll doesn't (yet) support it, it won't be used, and thus is just wasted bandwidth and space. When Jekyll does include assets (e.g., updates the theme spec), we just update the Regex, and any site with assets would be able to distribute those assets simply by releasing a new version of their Gem.

spudowiar:
@benbalter I always imagined a Jekyll DSL for themes, e.g. `Jekyll.theme do [...] end` where people could do various Jekyll specific repetitive stuff, but there currently isn't much available. If Jekyll themes were to get far more complicated, I can see that becoming useful.

parkr:
I'd like to include this in 3.2.2. @ashmaroli Can you please move this back to a `select` and include the missing files in the whitelist in the `ThemeBuilder::ERBRenderer` in `theme_builder.rb`? Thanks!

ashmaroli:
I've currently pushed a simple fix to just get this working: 68aa5ca

Also under consideration: Regex accepted at *minima* repo:
```ruby
  spec.files         = `git ls-files -z`.split("\x0").select do |f|
    f.match(%r{^(_(includes|layouts|sass)/|(LICENSE|README)((\.(txt|md|markdown)|$)))}i)
  end
```
> include the missing files in the whitelist in the ThemeBuilder::ERBRenderer in theme_builder.rb

I do not see the need to modify `theme_builder.rb`, but I'm open to incorporate ideas if required.

ashmaroli:
I was waiting to have a decision made regarding the regex being defined as a constant.

benbalter:
> @parkr would it create a circular dependency for the theme gemspec to require Jekyll (or a specific jekyll file) and include the Regex by reference?

☝️ @parkr (or anyone else) know if this is the case?

spudowiar:
No, it wouldn't, as long as Jekyll is installed, since Jekyll doesn't depend on our theme.

envygeeks:
That's why Jekyll doesn't depend on a theme in the upper Gemspec, themes should depend on a version of Jekyll they support, rather than have no version of Jekyll in their Gemspec at all.  RubyGems can deal with that, and still continue on resolve (if I remember right) but bundler, it'll poop on your party, so it's best to avoid those kind of situations and just let the theme demand it's version of Jekyll.

spudowiar:
It also makes absolutely no sense to depend on a theme, a theme is completely non essential to using Jekyll.

envygeeks:
> It also makes absolutely no sense to depend on a theme

That's a subjective assessment that on it's face value is wrong, it makes a lot of sense, people tend to globally install and use Jekyll, having it in there would solve a problem that might occur otherwise (a problem that has already occurred for some users.)

benbalter:
:metal: @ashmaroli would you be able to move the regex to a constant on the Theme model, and then include it by reference in `theme.gemspec.erb`?

spudowiar:
@envygeeks it also makes absolutely no sense to listen to my advice when it comes to UX since I think like a developer and that never even crossed my mind :smile:

envygeeks:
@spudowiar yeah tbh I wouldn't have thought about it either if I hadn't of saw a few threads about it on the forums.  Otherwise I wouldn't have noticed... or cared :laughing:

ashmaroli:
@benbalter Is b2d3695 what you had in mind?

benbalter:
> Is b2d3695 what you had in mind?

Yes! Two questions:

1. What do you think about calling it `FILE_REGEX` (or something similar) to make the intention a bit more clear?

2. Is there any downside to adding it to `theme.rb` and requiring `jekyll` instead of the individual file?

ashmaroli:
> What do you think about calling it FILE_REGEX (or something similar) to make the intention a bit more clear?

I like how `Jekyll::Theme::REGEX` sounds like jekyll_theme_regex but to make intentions more clear, hows `Jekyll::Theme::GEM_REGEX` ?

> Is there any downside to adding it to `theme.rb` and requiring `jekyll` instead of the individual file?

There's no downside per se. Just thought future updates to *the constant* would be better if on a separate file.

benbalter:
> I like how Jekyll::Theme::REGEX sounds like jekyll_theme_regex but to make intentions more clear, hows Jekyll::Theme::GEM_REGEX ?

👍 

> There's no downside per se. Just thought future updates to the constant would be better if on a separate file.

I think to be safe, it's probably better to include `jekyll` and add it to the existing theme file. Otherwise, once the require statement was hard coded in a theme, we'd never be able to adjust our file structure (or e.g., move that constant to another file). This gives us more flexibility and makes it easier for users to add to their gemspec (remembering just Jekyll, rather than a long path).

ashmaroli:
**Update:** This PR has been rebased to simply *fix* the concerned issue as *upgrading* the regex to a `Jekyll::Constant` has been handled better in another pull.

PR #5326, opened 2016-09-05
update gitignore template to ignore theme gems built by user

ashmaroli:
add *.gem to have git ignore theme gems built by the user.


envygeeks:
I'm okay with this, LGTM.

parkr:
LGTM.

@jekyllbot: merge +minor

PR #5329, opened 2016-09-06
Docs: update permalink for pretty urls

cycomachead:
Currently the permalink `other/` generates the file `other/index.html`, and not `/other`.

### DIFF
--- a/site/_docs/pages.md
+++ b/site/_docs/pages.md
@@ -95,6 +95,6 @@ simple and it works. In the end, the decision is yours!
     matter variable. In the example above, using the first method, you can
     get URL <code>http://example.com/other</code> for the file
     <code>other.md</code> by setting this at the top of the file:
-    <code>permalink: /other</code>
+    <code>permalink: other/</code>
   </p>
 </div>


DirtyF:
cc @jekyll/documentation

benbalter:
LGTM.

DirtyF:
👎 

* `/other` will generate a other.html file with canonical `/other`
* `other/` will generate an other/index.html file with canonical `/other/`

site/_docs/pages.md
```diff
@@ -95,6 +95,6 @@ simple and it works. In the end, the decision is yours!
     matter variable. In the example above, using the first method, you can
     get URL <code>http://example.com/other</code> for the file
```
Strangehill: If adding the trailing slash on line 98 wouldn't you also need it here?

cycomachead: I suppose it should be there too--I just forget to change that, but browsers redirect `/other` → `/other/` automatically, so it still does work.

cycomachead:
> /other will generate a other.html file with canonical /other

When I do this that doesn't work...`other.html` gets generated, but the URL `/other` is a 404 because the browser is adding a trailing /.

But in any case, I think the "ProTip" section should either match the preceding section about creating a structure that complies to `other/index.html` or be written to be clear that the output of `/other` is _not_ the same as the directory structure shown above.

site/_docs/pages.md
pathawks: Is it the _browser_ that adds the slash, or the _server_?

parkr: `other.html` _should_ be written here, not `other/index.html`. Check `_site`.

Issue #5144, opened 2016-07-28
[jekyll 3.2.0]static_file.rb 149 `utime': No such file or directory


jnozsc:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I Updated to the latest Jekyll (or) if on Github Pages to the latest `github-pages`
- [x] I Read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [ ] This is a feature request.

---

- [ ] I am on (or have tested on) ***macOS*** 10+
- [x] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux
- [ ] I am on (or have tested on) ***Fedora*** GNU/Linux
- [ ] I am on (or have tested on) ***Arch*** GNU/Linux
- [ ] I am on (or have tested on) ***Other*** GNU/Linux
- [ ] I am on (or have tested on) ***Windows*** 10+

---

- [ ] I was trying to install.
- [ ] There is a broken Plugin API.
- [ ] I had an error on Github Pages, and I have not tested locally.
- [ ] I had an error on Github Pages, and Github Support said it was a Jekyll Bug.
- [ ] I had an error on Github Pages and I did not test it locally.
- [x] I was trying to build.
- [ ] It was another bug.

## My Reproduction Steps
```
2016-07-28T02:51:37.297835+00:00 app[web.1]: Generating site: /app -> /app/_site
2016-07-28T02:51:37.371747+00:00 heroku[web.1]: Process exited with status 1
2016-07-28T02:51:37.386510+00:00 heroku[web.1]: State changed from starting to crashed
2016-07-28T02:51:37.387493+00:00 heroku[web.1]: State changed from crashed to starting
2016-07-28T02:51:39.161559+00:00 heroku[web.1]: Starting process with command `bundle exec rackup config.ru -p 30370`
2016-07-28T02:51:52.026687+00:00 app[web.1]: /app/vendor/bundle/ruby/2.3.0/gems/jekyll-3.2.0/lib/jekyll/static_file.rb:149:in `utime': No such file or directory @ utime_internal - /app/_site/bin/erb (Errno::ENOENT)
2016-07-28T02:51:52.026709+00:00 app[web.1]: 	from /app/vendor/bundle/ruby/2.3.0/gems/jekyll-3.2.0/lib/jekyll/static_file.rb:149:in `copy_file'
[...]

PR #5334, opened 2016-09-06
[docs] add help for missing static_file e.g. on heroku

Crunch09:
Hi,
as described in #5144 somehow on Heroku a `bin/erb` file couldn't be found during build. This was a symlinked file and it couldn't be copied because of the reasons @parkr pointed out in https://github.com/jekyll/jekyll/pull/4640#discussion_r57260002 .
I wasn't sure if i should also rescue the exception in the `static_file.rb` and provide a better error message in the log, so for now there is only a new section in the troubleshooting doc.

I would be glad if @xuanxu and / or @stomar could confirm if their site is building again after they set their jekyll environment to `production`.

cc: @jekyll/documentation 

screen shot 2016-09-06 at 22 54 20

### DIFF
--- a/site/_docs/troubleshooting.md
+++ b/site/_docs/troubleshooting.md
@@ -13,6 +13,7 @@ that might be of help. If the problem you’re experiencing isn’t covered belo
 - [Base-URL Problems](#base-url-problems)
 - [Configuration problems](#configuration-problems)
 - [Markup Problems](#markup-problems)
+- [Production Problems](#production-problems)
 
 ## Installation Problems
 
@@ -208,6 +209,13 @@ v1.1.0, Jekyll also passes these excerpts through Liquid, which can cause
 strange errors where references don't exist or a tag hasn't been closed. If you
 run into these errors, try setting `excerpt_separator: ""` in your
 `_config.yml`, or set it to some nonsense string.
+
+## Production Problems
+
+If you run into an issue that a static file can't be found in your
+production environment during build since v3.2.0 you should set your
+[environment to `production`](../configuration/#specifying-a-jekyll-environment-at-build-time).
+The issue is caused by trying to copy a non-existing symlink.
 
 <div class="note">
   <h5>Please report issues you encounter!</h5>


site/_docs/troubleshooting.md
```diff
@@ -209,6 +210,13 @@ strange errors where references don't exist or a tag hasn't been closed. If you
 run into these errors, try setting `excerpt_separator: ""` in your
 `_config.yml`, or set it to some nonsense string.
 
+## Production Problems
+
+If you run into an issue that a static file can't be found in your
+production environment during build since v3.2.0 you should set your
+[environment to `production`](https://jekyllrb.com/docs/configuration/#specifying-a-jekyll-environment-at-build-time).
```
DirtyF: Could we use relative link here instead please? `(../configuration/#specifying-a-jekyll-environment-at-build-time)` Thanks.

Crunch09: Thanks! Updated it.

envygeeks: I agree with removing the domain but I don't get the logic behind a relative link? Unless Jekyll has some magic that makes these absolute they should be absolute, having links relative makes it hard to change the name of a page and while it might be rare it can happen and this makes that move more expensive than it ever should be.

xuanxu:
Confirmed, that solved the issue for me, I set the JEKYLL_ENV to production in heroku and build was successfull. Thanks @Crunch09!

parkr:
LGTM.

parkr:
@jekyllbot: merge +site

site/_docs/troubleshooting.md
DirtyF: @envygeeks Got it.  I'll open another PR to fix this globally, because there are already some relative links elsewhere in the docs.

stomar:
IMO the description is a bit unclear. It's not specifically a production problem but rather a problem for possibly **any site with symlinks** that break when copied elsewhere (namely to `_site`), e.g. relative symlinks to files outside the site's directory. Building such a site in a **non-production** JEKYLL_ENV environment (e.g. development) fails, because the symlinks are copied as symlinks and the target can not be found.

On Heroku this happens also for sites without any symlinks, since Heroku seems to add some.

@envygeeks The statement in #5144 that this problem is caused by 0f74db4 is wrong; I did some further tests and wanted to add a corresponding comment there, but the issue is locked now (why?!?).
Could you please unlock?

Crunch09:
@envygeeks Yes, you're right, the issue was introduced by https://github.com/jekyll/jekyll/pull/4640

envygeeks:
@stomar @Crunch09 my bot locks issues when people do ":+1:" or "I have this issue too" (or the like.) It is unlocked now so that you can continue the discussion, and yes my bot probably was wrong, it was just in the beginning phases of being able to do blames.

PR #5335, opened 2016-09-07
Explain how to copy a theme's files

### LINKED ISSUE #30: Jekyll gets confused with two post  at the same date
Jekyll gets confused, with two posts  at the same date. This leads (for example) to a broken "next" / "previous" - list for the posts. 

To fix this the slug is used for equality tests, if two posts share the same publication date. This nicely solves the problem.

Please pull from my repository (Changeset 0de96b8e9dec4b351b3db21a51f29ede5d45d7e1)

DirtyF:
Fix https://github.com/jekyll/minima/issues/30

cc/ @jekyll/documentation

### DIFF
--- a/site/_docs/themes.md
+++ b/site/_docs/themes.md
@@ -34,6 +34,10 @@ Jekyll will look first to your site's content, before looking to the theme's def
 Refer to your selected theme's documentation and source repository for more information on what files you can override.
 {: .note .info}
 
+To locate theme's files on your computer, run `bundle show` followed by
+the name of the theme's gem, e.g. `bundle show minima` for default Jekyll's
+theme. Then copy the files you want to override, from the returned path to your root folder.
+
 ## Creating a theme
 
 Jekyll themes are distributed as Ruby gems. Don't worry, Jekyll will help you scaffold a new theme with the `new-theme` command. Just run `jekyll new-theme` with the theme name as an argument:


pathawks:
@DirtyF Which issue does this fix?

ashmaroli:
Is there a way to `--watch` those theme directories directly w/o copying them over to working blog directory?

site/_docs/themes.md
```diff
@@ -34,6 +34,10 @@ Jekyll will look first to your site's content, before looking to the theme's def
 Refer to your selected theme's documentation and source repository for more information on what files you can override.
 {: .note .info}
 
+To locate theme's files on your computer, run `bundle show` followed by
+the name of the theme's gem, e.g. `bundle show minima` for default Jekyll's
+theme. Then copy the files you want to override, from the returned path to your root folder.
+
```
benbalter: Assuming the theme's hosted someplace like GitHub, do we want to also direct users to the themes website (which will likely have things like docs and examples)?

benbalter:
> Is there a way to --watch those theme directories directly w/o copying them over to working blog directory?

You shouldn't have to, because the only way that directory should change is when the Gem gets updated. If you'd like to make local changes to those files, you should manually copy them into your site.

site/_docs/themes.md
parkr: In theory this would be in the RubyGems metadata as `homepage`.

parkr:
LGTM.

👍 on the idea to go visit the URL, though that could cause issues if you have an old version of the theme and you look at the master branch on GitHub.

envygeeks:
LGTM.

envygeeks:
@jekyllbot: merge +site

PR #5336, opened 2016-09-07
update theme installation instructions

ashmaroli:
This PR is to update documentation on installing themes regarding importing theme partials.
Individual theme documentations do mention this. I'm just making sure that this info is also present in the upstream docs. 

/cc @jekyll/documentation, @DirtyF

### DIFF
--- a/site/_docs/themes.md
+++ b/site/_docs/themes.md
@@ -14,7 +14,11 @@ Jekyll has an extensive theme system, which allows you to leverage community-mai
 
 2. Save the changes to your `Gemfile`
 3. Run the command `bundle install` to install the theme
-4. Finally, activate the theme by adding the following to your site's `_config.yml`:
+4. Set your `/css/main.scss` to import your theme partials using the `@import` directive:
+
+        @import "my-awesome-jekyll-theme";
+
+5. Finally, activate the theme by adding the following to your site's `_config.yml`:
 
         theme: my-awesome-jekyll-theme
 


site/_docs/themes.md
```diff
@@ -14,7 +14,11 @@ Jekyll has an extensive theme system, which allows you to leverage community-mai
 
 2. Save the changes to your `Gemfile`
 3. Run the command `bundle install` to install the theme
-4. Finally, activate the theme by adding the following to your site's `_config.yml`:
+4. Set your `/css/main.scss` to import your theme partials using the `@import` directive:
+
+        @import: my-awesome-jekyll-theme
```
parkr: I have never seen `@import` with a colon before. Is this new?

DirtyF: Could we stick with the `@import "my-awesome-theme";` here?

ashmaroli: @parkr sorry, I did not mean to include the colon. must've been half-asleep.. :grin: 

@DirtyF, If you check the full page, you'll see that `my-awesome-jekyll-theme` was the example in the steps preceding this.
I'll update after you confirm.

DirtyF: @ashmaroli yes sorry, I was only refering to the Sass syntax, of course we wanna keep the same exemple.

site/_docs/themes.md
```diff
@@ -14,7 +14,11 @@ Jekyll has an extensive theme system, which allows you to leverage community-mai
 
 2. Save the changes to your `Gemfile`
 3. Run the command `bundle install` to install the theme
-4. Finally, activate the theme by adding the following to your site's `_config.yml`:
+4. Set your `/css/main.scss` to import your theme partials using the `@import` directive:
```
benbalter: Should we create the design pattern of using `/assets/style.scss? That way, when themes can support other files, we can just copy the assets folder (with images, js, stylesheets, etc.). Additionally, given this design pattern, I'm not sure how many sites will have multiple CSS files (maybe sass, but that'd be in `_sass`) to make the dedicated folder necessary.

ashmaroli: Wow! I was just thinking about something similar..
Thats a great idea. We need to work out a proper schema..

Anyways, I feel this is all that's required to be documented as of now. We can always update the docs when necessary.

ashmaroli:
@jekyllbot: remove `pending-feedback`

site/_docs/themes.md
parkr: I think the theme's default layout requires using the `css` directory for now.

parkr:
LGTM.

site/_docs/themes.md
parkr: We now have `/assets` support so we can move this to `/assets/<theme>.scss` which imports `_sass/<theme>.scss`?

ashmaroli: @parkr Can you create a blog-post *detailing* the addition of the assets folder as part of `SCAFFOLD_DIRECTORIES` like how to use this new feature.

Right now, I'm confused if moving `css/main.scss' generated by `jekyll new` can be moved to `/assets` and it'll still be read and processed..

DirtyF: @ashmaroli e.g https://github.com/broccolini/swiss uses the assets folder, and https://github.com/broccolini/swiss/blob/master/assets/style.scss using `site.theme` to import styles. Site template will be updated for Jekyll 3.3 to remove css/main.css

ashmaroli: > Site template will be updated for Jekyll 3.3 to remove css/main.css

I see.. in that case I think its best to update the documentation within that PR.. and close this after that gets merged..

parkr: This will need to be updated with the outcome of https://github.com/jekyll/minima/pull/43.

ashmaroli: > This will need to be updated with the outcome of jekyll/minima#43.

@parkr Actually, there's no longer a need for this PR.. It was to inform users the need to set `css/main.scss` to import **<theme-name>** as part of the installation. Its no longer the case. This is now expected to be *pre-configured* within every jekyll-theme-gem itself.

Hence, closing this one out..

ashmaroli:
Closing, as this is no longer needed..

Issue #5328, opened 2016-09-06
Jekyll HEAD has a breaking change that breaks some plugins.


envygeeks:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [ ] This is a feature request.

---

- [x] I am on (or have tested on) ***macOS*** 10+
- [x] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux
- [ ] I am on (or have tested on) ***Fedora*** GNU/Linux
- [ ] I am on (or have tested on) ***Arch*** GNU/Linux
- [ ] I am on (or have tested on) ***Other*** GNU/Linux
- [x] I am on (or have tested on) ***Windows*** 10+

---

See: https://github.com/jekyll/jekyll-sitemap/issues/123

ashmaroli:
I have traced this error to [this PR](https://github.com/jekyll/jekyll/pull/5308)

parkr:
I just noticed that when running `bundle exec rake site:publish`! Fix in #5337.

Crunch09:
I guess this can be closed now?

envygeeks:
Yessum, thanks @parkr! And thanks for pinging @Crunch09.

PR #5337, opened 2016-09-07
Convertible: set self.output in #render_all_layouts and #do_layout

parkr:
In #5308, I modified `Convertible` to essentially delegate to the `Renderer`. This removed a lot of duplicated code (😸) but introduced a slight difference: the `Renderer` doesn't really modify the document it's rendering–it's meant only to render and return the result. Sadly, then, the methods aren't *entirely* equivalent. This PR sets `self.output` in two places to ensure the old functionality is preserved.

Without this, the `jekyll-sitemap` gem barfed when rendering the source of `jekyllrb.com`. It [uses the `Page` API's rather than the `Renderer`](https://github.com/jekyll/jekyll-sitemap/blob/bc9f90e032c2bad2f72b451f9ccc3dc8523c6af4/lib/jekyll/jekyll-sitemap.rb#L56-L57). Without these changes the `site_map.output` value is `nil` on the latest `jekyll/jekyll` master.

/cc @jekyll/stability

### DIFF
--- a/lib/jekyll/convertible.rb
+++ b/lib/jekyll/convertible.rb
@@ -193,7 +193,7 @@ module Jekyll
     # Returns nothing
     def render_all_layouts(layouts, payload, info)
       _renderer.layouts = layouts
-      _renderer.place_in_layouts(output, payload, info)
+      self.output = _renderer.place_in_layouts(output, payload, info)
     ensure
       @_renderer = nil # this will allow the modifications above to disappear
     end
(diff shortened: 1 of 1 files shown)


parkr:
Fixes https://github.com/jekyll/jekyll/issues/5328 and https://github.com/jekyll/jekyll-sitemap/issues/123.

parkr:
@envygeeks would you mind taking a 👀 ?

envygeeks:
This is LGTM!

parkr:
Thanks. :) LGTM.

parkr:
@jekyllbot: merge +bug

pathawks:
:tada:

Issue #5070, opened 2016-07-08
Filter to get absolute URL


pathawks:
- [ ] I believe this to be a bug, not a question about using Jekyll.
- [x] I Updated to the latest Jekyll (or) if on Github Pages to the latest `github-pages`
- [x] I Read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [x] This is a feature request.

---

It would be nice if there were some way on a Liquid template to get the absolute URL of a Jekyll document.

Currently, several plugins have different incantations of prepending `site.url + site.baseurl` or `site.github.url`. It would be nice if core included One True Way to facilitate this. This way, a Liquid template could just do something equivelant to:

```text
{{ page.url | absolute }}
```

or

```text
{{ page.absolute_url }}
```

For background, see [jekyll/jekyll-sitemap#112 (comment)](https://github.com/jekyll/jekyll-sitemap/issues/112#issuecomment-222330371)

I would be happy to submit a pull request if others agree this would be of value.

envygeeks:
Personally I feel this is entirely out of scope for Jekyll.

pathawks:
> Personally I feel this is entirely out of scope for Jekyll.

The other two alternatives I see are:

 1. Each plugin/theme develops its own (probably quirky) way to handle URLs.
 2. A plugin is developed that handles URLs, and any plugin/theme that needs absolute URLs can reference this plugin as a dependency.

The current situation (Alternative 1) has caused users to run into edge cases where their URLs are handled correctly by some plugins and not others, and differing expectations of how they even _should_ be handled.

So it sounds like Alternative 2 is your recommendation?

envygeeks:
Two, it sounds okay.  If @jekyll/core and @jekyll/plugin-core agree we could make it a default plugin too.

parkr:
I think hyperlinks are a core part of what Jekyll does. It builds webpages and other content and in doing so we have always offered template-based URL building. [My post about `baseurl`](https://byparker.com/blog/2014/clearing-up-confusion-around-baseurl/) is easily one of the most-referenced pages on my site. `site.baseurl`, `site.url` and now `site.github.url` cause a lot of confusion. It'd be nice to provide a unified/centralized way of doing this.

[...]

PR #5338, opened 2016-09-08
Site template: set empty url in config file by default

ashmaroli:
Addresses [minima/#34](https://github.com/jekyll/minima/issues/34) at the upstream.
This bug occurs due a recent change shipped with minima-1.1.0

/cc @jekyll/ecosystem
/cc @pathawks

### DIFF
--- a/lib/site_template/_config.yml
+++ b/lib/site_template/_config.yml
@@ -20,7 +20,7 @@ description: > # this means to ignore newlines until "baseurl:"
   line in _config.yml. It will appear in your document head meta (for
   Google search results) and in your feed.xml site description.
 baseurl: "" # the subpath of your site, e.g. /blog
-url: "http://example.com" # the base hostname & protocol for your site
+url: "" # the base hostname & protocol for your site, e.g. http://example.com
 twitter_username: jekyllrb
 github_username:  jekyll
 


lib/site_template/_config.yml
```diff
@@ -20,7 +20,8 @@ description: > # this means to ignore newlines until "baseurl:"
   line in _config.yml. It will appear in your document head meta (for
   Google search results) and in your feed.xml site description.
 baseurl: "" # the subpath of your site, e.g. /blog
-url: "http://example.com" # the base hostname & protocol for your site
+url: "" # the base hostname & protocol for your site, e.g. http://example.com
```
envygeeks: This isn't required by Jekyll.  Put it downstream.

ashmaroli: but Jordon, `jekyll new` creates the `_config.yml` with the `url` included.

pathawks:
Thanks @ashmaroli for opening this here. This is not a minima issue.

If the site template is going to have a `url` in config, it should be an empty string with instructions on how to configure.

Jekyll is a static site generator, and generating URLs is foundational to that. Many many plugins (and now themes) rely on `site.url`.

See also [this comment from parkr](https://github.com/jekyll/jekyll/issues/5070#issuecomment-231502898)

Big :+1: from me.

lib/site_template/_config.yml
```diff
@@ -20,7 +20,8 @@ description: > # this means to ignore newlines until "baseurl:"
...
+        # [Required] for production environment.
```
pathawks: I don't know if this comment is necessary. Some plugins require absolute URLs, but most themes should not.

ashmaroli: Added that based on [DirtyF's comment](https://github.com/jekyll/minima/issues/34#issuecomment-245447733)

pathawks: I assume when he said "you're gonna have to set this value" just meant to change it from `example.com`

ashmaroli: Righhtt... that makes sense :+1:  :smiley: 
so.. remove that comment entirely or update to: 
```
Update this value before going into production.
```
?

pathawks: My vote is to just remove [line 24](https://github.com/ashmaroli/jekyll/blob/debd1b7beaa0307a7b00c70f53c844306a59517c/lib/site_template/_config.yml#L24) and leave the comment on 23.

envygeeks:
It appears you are right, so it's got a LGTM from me.

envygeeks:
LGTM.

ashmaroli:
@envygeeks, Thanks. Whats your take on the comment at [Line 24](https://github.com/ashmaroli/jekyll/blob/debd1b7beaa0307a7b00c70f53c844306a59517c/lib/site_template/_config.yml#L24)? Update or Discard?

envygeeks:
It's not required, none of my sites use that variable and it's not even in any of my sites configuration's.

DirtyF:
@envygeeks 😮  interesting, I figured out `site.url` was needed as some point or another.

But if it's not useful, why embed it by default in `_config_yml `?

envygeeks:
>  interesting, I figured out site.url was needed as some point or another.

@DirtyF I'll concede that I pretty much program all my own plugins and even had to fork Jekyll Sitemap and Jekyll Feed because of the way it handle's URL's, so I probably haven't run into that problem because I do tend to fork and customize a lot, most people don't.

> But if it's not useful, why embed it by default in _config_yml?

I don't know if it's useful or not, that's up to the community, for me it's not useful, but for you it is, you've found that eventually you need it, that means that we should warn about it in some context and this is probably the best context to warn about it in.

ashmaroli:
Taking [abishov.com](http://abishov.com/)'s [config file](https://github.com/ArazAbishov/arazabishov.github.io/blob/master/_config.yml#L9) ( the line with `url: ` variable has been commented-out ), as a working example, my thoughts echo that of Jordon's comment above.

the `url: ` variable seems to be an unnecessary definition. 
So, I now propose to have that variable and its references removed from the default template.

ashmaroli:
> Doesn't site.url provide for explicit linkages within feed.xml?

@jaybe-jekyll, Sorry, I do understand what you are asking.. Can you please rephrase?

jaybe-jekyll:
@ashmaroli It was asked where `site.url` is used/employed/needed. The default/new minima theme' `feed.xml` references `site.url` to build proper feed links.

ashmaroli:
Jekyll (HEAD) now uses **jekyll-feed** instead of hard-coded `feed.xml`. Since the plugin doesn't depend on `site.url`'s presence to generate a site's feeds, even if we were to move in favor of removing the key from `_config.yml` altogether, it wont break things.
But giving a benefit of doubt to its dependency from other plugins, this PR is good to :ship:

pathawks:
> Since the plugin doesn't depend on `site.url`'s presence to generate a site's feeds…

Yes it does, unless the site is hosted on GitHub Pages. Relative URLs in a feed cause problems.

ashmaroli:
lobbying to have this merged a.s.a.p if its LGTM to the maintainers and release a pre-release version. At least we can give some respite to the users in despair

envygeeks:
@ashmaroli Github doesn't work on our release cycle and us not theirs, so even if we did, it only fixes the problem locally really and I don't know that it would affect Github Pages unless they update Gem themes everyday.

ashmaroli:
> so even if we did, it only fixes the problem locally really

This would have us meet our users halfway in solving the issue. Pushing a new site into production with an empty `url:` is less punishing than having one with `http://www.example.com`
(IMHO)

parkr:
@DirtyF @jaybe-jekyll Check out #5399. You might find you like that PR. 😄 

@jekyllbot: merge +bug

PR #5344, opened 2016-09-09
Make Jekyll hyper-parallel, how about we make Jekyll fast, yes?

envygeeks:
This commit introduces parallel processing into Jekyll, which in essence will speed up builds for people with larges sites... considerably. By throwing out many forked Jekyll's to process each piece independently marshaling and bringing back in pieces of data as necessary Jekyll can get exponential (possibly) increases of processing power with better IO and more cores.  How much will this speed up Jekyll?

```
Command being timed: "bundle exec ruby /usr/bin/jekyll b"
User time (seconds): 173.07
System time (seconds): 2.68
Percent of CPU this job got: 1064%
Elapsed (wall clock) time (h:mm:ss or m:ss): 0m 12.74s
Average shared text size (kbytes): 0
Average unshared data size (kbytes): 0
Average stack size (kbytes): 0
Average total size (kbytes): 0
Maximum resident set size (kbytes): 876848
Average resident set size (kbytes): 0
Major (requiring I/O) page faults: 0
Minor (reclaiming a frame) page faults: 1794500
Voluntary context switches: 250534
Involuntary context switches: 4696
Swaps: 0
File system inputs: 80
File system outputs: 50696
Socket messages sent: 0
Socket messages received: 0
Signals delivered: 0
Page size (bytes): 4096
```

That is https://github.com/kubernetes/kubernetes.github.io and it was built on my personal 24 core server.  Jekyll consumed little over 10 of my cores and built the site in 2 seconds compared to the 36 seconds it normally takes to build that site.  On Jekyll's own site, the build time goes from 9 seconds average to 3-4 seconds on my 4 core Skylake on Linux.

**_This pull request still needs work, however, it works for most sites as it stands, we hav tested it across a dozen or so sites, however there are some stuff that still needs to be accounted for.**_

---

This feature was paid for and sponsored by @forestryio (@jpatters and @scottgallant) who wanted it for Forestry.io and to also hand it back over to the open-source community so it could benefit the entire world and not just them.

### DIFF
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -32,6 +32,9 @@ require "kramdown"
 require "colorator"
 
 SafeYAML::OPTIONS[:suppress_warnings] = true
+ENV["PARALLEL_UNITS"] ||= Pathutil.new("/proc/cpuinfo").read
+  .scan(/^processor\s+:\s+\d/m).size.to_s
+
 
 module Jekyll
   # internal requires
--- a/jekyll.gemspec
+++ b/jekyll.gemspec
@@ -38,4 +38,5 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('jekyll-sass-converter', '~> 1.0')
   s.add_runtime_dependency('jekyll-watch', '~> 1.1')
   s.add_runtime_dependency("pathutil", "~> 0.9")
+  s.add_runtime_dependency("parallel", "~> 1.9")
 end
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -32,11 +32,50 @@ module Jekyll
         categories_from_path(collection.relative_directory)
       end
 
-      data.default_proc = proc do |_, key|
-        site.frontmatter_defaults.find(relative_path, collection.label, key)
+      data.default_proc = ddp
+      trigger_hooks(:post_init)
+    end
+
+    def ddp
+      return proc do |_, key|
+        @site.frontmatter_defaults.find(
+          relative_path, @collection.label, key
+        )
+      end
+    end
+
+    def restore_state(site, collection)
+      @site = site
+      @collection = collection
+      @data["excerpt"] = Excerpt.new(
+        self
+      )
+    end
+
+    def marshal_dump
+      @data.default_proc = nil; @data["excerpt"] = nil
+      blacklist = [:@site, :@collection, :@to_liquid]
+      Jekyll.logger.debug "Skipping the marshal of vars #{
+        blacklist
+      }."
+
+      (instance_variables - blacklist).each_with_object({}) do |var, obj|
+        Jekyll.logger.debug "Marshaling the variable #{var} in Document"
+        obj[var] = instance_variable_get(
+          var
+        )
+      end
+    end
+
+    def marshal_load(data)
+      data.each do |k, v|
+        instance_variable_set(
+          k, v
+        )
       end
 
-      trigger_hooks(:post_init)
+      @data.default_proc = ddp
+      self
     end
 
     # Fetch the Document's data.
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -44,12 +44,58 @@ module Jekyll
 
       process(name)
       read_yaml(File.join(base, dir), name)
+      data.default_proc = ddp
 
-      data.default_proc = proc do |_, key|
-        site.frontmatter_defaults.find(File.join(dir, name), type, key)
+      Jekyll::Hooks.trigger :pages,
+        :post_init, self
+    end
+
+    def ddp
+      proc do |_, key|
+        @site.frontmatter_defaults.find(
+          File.join(@dir, @name), type, key
+        )
+      end
+    end
+
+    def marshal
+      Marshal.dump(
+        self
+      )
+    end
+
+    def restore_state(site)
+      @site = site
+    end
+
+    def marshal_dump
+      @data.default_proc = nil
+      blacklist = [
+        :@site, :@_renderer
+      ]
+
+      Jekyll.logger.debug "Skipping the marshal of vars #{
+        blacklist
+      }."
+
+      (instance_variables - blacklist).each_with_object({}) do |var, obj|
+        Jekyll.logger.debug "Marshaling the variable #{var} in Document"
+        obj[var] = instance_variable_get(
+          var
+        )
+      end
+    end
+
+    def marshal_load(data)
+      data.each do |k, v|
+        instance_variable_set(
+          k, v
+        )
       end
 
-      Jekyll::Hooks.trigger :pages, :post_init, self
+      @data.
+        default_proc = ddp
+      self
     end
 
     # The generated directory into which the page will be placed
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -1,4 +1,5 @@
 # encoding: UTF-8
+require "parallel"
 require "csv"
 
 module Jekyll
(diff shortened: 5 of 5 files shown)


ashmaroli:
Thank you @envygeeks. 
Thank you @forestryio.
:smiley:

crispgm:
👍  `jekyllrb.com` will also benefit from this feature a lot.

lib/jekyll.rb
```diff
@@ -32,6 +32,9 @@ def require_all(path)
 require "colorator"
 
 SafeYAML::OPTIONS[:suppress_warnings] = true
+ENV["PARALLEL_UNITS"] ||= Pathutil.new("/proc/cpuinfo").read
```
spudowiar: Really? :smirk:

Try `Parallel.processor_count`

envygeeks: I'll stick with what I chose as I pleases me.

spudowiar: `Parallel.processor_count` does the same thing, it's cross-platform though. You're using `parallel` anyway, so it makes sense to use it?

https://github.com/grosser/parallel/blob/master/lib/parallel/processor_count.rb

envygeeks: I know what it does. I audited the code before I even used the library. It
existence doesn't mean I have to use it or even accept it. I made my choice
and it was to build my own as noted.

spudowiar: OK, I was just wondering about Windows support

On 10 Sep 2016 11:58, Jordon Bedwell <notifications@github.com> wrote:

In lib/jekyll.rb<https://github.com/jekyll/jekyll/pull/5344#discussion_r78275614>:

> @@ -32,6 +32,9 @@ def require_all(path)
>  require "colorator"
>
>  SafeYAML::OPTIONS[:suppress_warnings] = true
> +ENV["PARALLEL_UNITS"] ||= Pathutil.new("/proc/cpuinfo").read


I know what it does. I audited the code before I even used the library. It existence doesn't mean I have to use it or even accept it. I made my choice and it was to build my own as noted.
…

envygeeks: There will be windows support. But it'll be a much simpler method we use as
we don't need such wide support. This was originally built for a Docker
image so everything here reflects that.

spudowiar: Ah, OK, thanks for the explanation :smile:

benbalter: @envygeeks as this gets reviewed for core inclusion, can we now use the higher level method provided by Parallel to ensure cross-platform support?

benbalter:
@envygeeks can you explain a bit how and why you're using Marshal here?

Also, is there any risk of one file referencing another file, but that file not being rendered/read yet?

spudowiar:
@benbalter I thought we could use lazy evaluation but @envygeeks knows more about this than me so I'll happily go along with his implementation :)

envygeeks:
@benbalter @spudowiar The data is processed outside of the main process where shared memory does not happen.  We get a copy of the memory with our own bits if we choose to edit something, but those changes ***do not*** land back in the main process, we must marshal and ship the data back, well technically upstream marshals it and ships it back, we do have a marshal and unmarshal method but those are relics for testing what needed to be there when development to see what needed to be excluded from the marshal because it was too complex (read: hashes with default procs or anything with a proc that we cared about.)

There are two ways we can do things: YAML/JSON, or Marshal.  I will not be implementing YAML/JSON... ever.  It's too slow, and it will make this pull request meaningless, Marshal is and remains the fastest way to marshal data from a sub-process back into the main process.

parkr:
This is great! My previous attempts required shared memory, or assumed some kind of marshalling would be built into the library I used. Glad you got this working.

> There are two ways we can do things: YAML/JSON, or Marshal. I will not be implementing YAML/JSON... ever. It's too slow, and it will make this pull request meaningless, Marshal is and remains the fastest way to marshal data from a sub-process back into the main process.

`Marshal` is unfortunately not a safe way of handling this user-provided data so this would be a show-stopper for GitHub Pages. MsgPack is about the same speed (at the very least, much, *much* faster than SafeYAML or JSON) and is a safe alternative. Would you consider using that here?

Here is the [Marshal documentation](http://docs.ruby-lang.org/en/2.1.0/security_rdoc.html#label-Marshal.load) in case you are interested:

> Never use Marshal.load to deserialize untrusted or user supplied data. Because Marshal can deserialize to almost any Ruby object and has full control over instance variables, it is possible to craft a malicious payload that executes code shortly after deserialization.

spudowiar:
@parkr I'm not sure this really is an issue **(I'm NOT a security expert, though!)** since the "user-provided" data is actually just from a fork of Jekyll **as far as I understand**

parkr:
> I'm not sure this really is an issue (I'm NOT a security expert, though!) since the "user-provided" data is actually just from a fork of the Jekyll process as far as I understand

@spudowiar If you had malicious Markdown and was able to execute arbitrary Ruby code on any computer that ran Jekyll because Jekyll was using an unsafe library for inter-process communication, we'd be in deep trouble. The problem is that your data as the user is included in these communications and could thus take advantage of the vulnerabilities in the Marshal library.

spudowiar:
@parkr I thought the issue is that someone could `Marshal` an arbitrary object, for example, and when it's `Marshal.load`ed, it would instantiate the arbitrary object (e.g. a subprocess object). I believe that `Marshal.dump` is safe and `Marshal.load` is safe as long as YOU `dump`ed its input

envygeeks:
The risk of that happening is pretty low and a surface that only exists for Github because of the way that infrastructure is done (though I don't see why that is still a problem? Seccomp/Apparmor?) Either way, I can switch to msgpack but that'll take some time as I'll have to create my own dump and load methods on each given class, that said, msgpack also throws out data it cannot handle instead of throwing, so there is a problem with that too.)

That'll require some work.

lib/jekyll.rb
envygeeks: @benbalter I will ***not*** be using Parallels method, as I stated I will be using my own. // @mattr-

envygeeks: @mattr- it'll have basic-cross platform support, as much as Jekyll has, I just haven't had time to come back to this as we've bigger problems and this can wait.

untra:
What is the status of this PR? I understand parallel build support is a tricky problem with a major security concern. I'm very curious as to what the correct implementation approach should be, and I'm very excited for this new feature.
Thanks! 🍸

envygeeks:
It's still being worked on but being tested on our build servers to spot problems that need to be addressed.  It won't be finished for a while yet because there are lots of edges in the way Jekyll deals with data (namely  using default proc instead of a wrapper class,) which prevents us from pushing this out faster.

untra:
Is this feature still under development, or has it been pushed to the wayside?

mattr-:
At this point, it's safe to assume that it's been pushed to the wayside, as there's still open issues with the implementation and there hasn't been an update in about 8 months.

envygeeks:
There are no issues with this pull request, there are issues with Jekyll. It's Jekyll that stalled this pull request because of the things it does.  Jekyll can never be parallel until it fixes it's ways.

mattr-:
@envygeeks I'm sorry, but I believe you're mistaken. I've counted four issues (or three if you combine the two issues below about `Marshal` into one) with this implementation:
1. No cross-platform support
2. Open questions about the use of `Marshal`
3. Request to not use `Marshal` that hasn't been addressed.
4. No fixes to the way Jekyll does things so that it can be more parallel.

I get that there's an argument for that last one not being an issue with this particular implementation, but I'm honestly confused as to how the first three aren't actual problems with this PR (and therefore the implementation). Where did I go wrong? 

Would you prefer to cut your losses now and abandon this work since your obligation to @forestryio is completed?

envygeeks:
I never had an obligation to @forestryio on this patch, they gave it to you out of the kindess of their hearts after I suggested it. Read your own CoC and stop making this personal.  You have no idea what you are talking about. Piece of shit.

mattr-:
I'm sorry, I took

>This feature was paid for and sponsored by @forestryio (@jpatters and @scottgallant) who wanted it for Forestry.io 

to mean that they'd paid you for the work and were having you contribute it back to the community. Based on your comment above, I was wrong.

benbalter:
Re-opened by mistake, with apologies for the noise.

PR #5360, opened 2016-09-15
Themes: Gem `files` RegExp as constant

spudowiar:
See #5323 and #5325 for earlier discussion.

/cc @jekyll/ecosystem @ashmaroli @benbalter


lib/jekyll/theme_builder.rb
```diff
@@ -3,6 +3,15 @@ class Jekyll::ThemeBuilder
     _layouts _includes _sass
   ).freeze
 
+  DOCUMENTATION_FILES = %w(
+    README LICENSE
+  ).freeze
+
+  GEM_INCLUSION_REGEX = Regexp.union(
+    %r!\A#{Regexp.union(SCAFFOLD_DIRECTORIES)}/!,
+    %r!\A#{Regexp.union(DOCUMENTATION_FILES)}(?:\..*)?\z!i
```
spudowiar: Slightly looser than previously (`README.*` vs `README.(md|txt|markdown)`)

benbalter:
Nice. We'll need to reconcile this with @ashmaroli's efforts over in https://github.com/jekyll/jekyll/pull/5325.

One question, does it make more sense to have the Regex in `theme.rb`, rather than `theme_builder.rb` to keep like-logic together? Although the constant is used by the theme builder, once, it's used by the theme itself (outside the theme builder experience) every time after that.

lib/jekyll/theme_builder.rb
```diff
@@ -3,6 +3,15 @@ class Jekyll::ThemeBuilder
     _layouts _includes _sass
   ).freeze
 
+  DOCUMENTATION_FILES = %w(
+    README LICENSE
+  ).freeze
+
+  GEM_INCLUSION_REGEX = Regexp.union(
+    %r!\A#{Regexp.union(SCAFFOLD_DIRECTORIES)}/!,
+    %r!\A#{Regexp.union(DOCUMENTATION_FILES)}(?:\.[^/]+)?\z!i
```
spudowiar: Slightly different than previously, (e.g. `README.*` vs `README.{md,txt,markdown}`)

spudowiar:
@benbalter I left it `Jekyll::ThemeBuilder` since `SCAFFOLD_DIRECTORIES` was there, I can move both to `Jekyll::Theme` (rename `SCAFFOLD_DIRECTORIES` to `THEME_DIRECTORIES`) if people would prefer that

spudowiar:
@benbalter I don't really see reconcilation of this patch with #5325 as an option :wink:, which PR to merge is really just a matter of preference for you and the other maintainers now

ashmaroli:
I'm :+1: with including `SCAFFOLD_DIRECTORIES` constant in the regex.
I'm fine with either of the PRs getting merged, as long as the end-user enjoy the benefits at the earliest.

spudowiar:
>  I'm fine with either of the PRs getting merged, as long as the end-user enjoy the benefits at the earliest.

:+1: Glad to know we share the same opinion :smile:

benbalter:
LGTM. @jekyll/ecosystem?

parkr:
I really don't think things will work well when the gemspec references a dependency. I think we're asking for trouble here.

benbalter:
> I really don't think things will work well when the gemspec references a dependency. I think we're asking for trouble here.

Any other way we can centralize the regex? Is a cURL to jeykllrb terrible? It should only be used when building the gem, right?

parkr:
@benbalter My reasoning is that if we shipped some regexp that broke, all releases using that version of Jekyll would silently break. That's a scary scenario. I'd much rather have this be a manual update for theme creators. Yes, it's more headache for them, but it's way safer.

spudowiar:
@parkr The RegExp is only used at build time so we're not depending on Jekyll when it's installed by the user

I personally think there are far more things likely to break and, as long as my RegExp calculation is correct, we shouldn't really have to touch the RegExp itself

spudowiar:
@parkr Also, we lock the gem to building with a specific version of Jekyll, so the developer knows when they've changed that, it's not a silent change

parkr:
> @parkr Also, we lock the gem to building with a specific version of Jekyll, so the developer knows when they've changed that, it's not a silent change

This isn't true entirely. We lock to a MAJOR version via the [pessimistic version constraint](http://guides.rubygems.org/patterns/#pessimistic-version-constraint), e.g `~> 3.2`. This translates to `>= 3.2, < 4`. This means if we break this regexp in Jekyll 3.2.1, then 

> @parkr The RegExp is only used at build time so we're not depending on Jekyll when it's installed by the user

That's true, but there's race condition weirdness when you depend on Jekyll and require it in your gemspec. I think we'll find folks can't run this reliably.

> we shouldn't really have to touch the RegExp itself

I think we like changing things and I expect this regexp to be modified right away after 3.3 is released.

lib/theme_template/theme.gemspec.erb
```diff
@@ -10,9 +11,7 @@ Gem::Specification.new do |spec|
   spec.homepage      = "TODO: Put your gem's website or public repo URL here."
   spec.license       = "MIT"
 
-  spec.files         = `git ls-files -z`.split("\x0").select { |f| f.match(%r{^(<%= theme_directories.join("|") %>|LICENSE|README)/i}) }
+  spec.files         = `git ls-files -z`.split("\x0").grep(Jekyll::ThemeBuilder::GEM_INCLUSION_REGEX)
 
   spec.add_development_dependency "jekyll", "~> <%= jekyll_version_with_minor %>"
-  spec.add_development_dependency "bundler", "~> 1.12"
-  spec.add_development_dependency "rake", "~> 10.0"
```
parkr: These are required. Why did you remove these? The release process is `bundle exec rake release`.

spudowiar: `jekyll new-theme` doesn't create a `Rakefile`

spudowiar:
@parkr When the Gem is built, the RegExp is baked into the gemspec. Unless the developer is building a Jekyll theme without Jekyll, there shouldn't be any race condition weirdness, although I'd have to check how Bundler handles this.

parkr:
> @parkr When the Gem is built, the RegExp is baked into the gemspec. Unless the developer is building a Jekyll theme without Jekyll, there shouldn't be any race condition weirdness, although I'd have to check how Bundler handles this.

@spudowiar Right, I get that. What isn't clear to me is when you run `bundle exec jekyll serve` in your theme, Bundler reads in your `Gemfile` which specifies your `.gemspec`, which requires `Jekyll` but Jekyll isn't resolved yet so Bundler goes down in a shower of flames. Also, running `bundle install` will surely fail when Jekyll isn't already installed. This makes the UX for these tools much worse and perhaps impossible.

spudowiar:
@parkr I see, that should be easy to solve. Simply change `require "jekyll" ` and `spec.files = ...` have `unless Bundler` and `unless ::Bundler`, respectively. Expect a push later today.

On 21 Sep 2016 22:39, Parker Moore <notifications@github.com> wrote:

@parkr<https://github.com/parkr> When the Gem is built, the RegExp is baked into the gemspec. Unless the developer is building a Jekyll theme without Jekyll, there shouldn't be any race condition weirdness, although I'd have to check how Bundler handles this.

@spudowiar<https://github.com/spudowiar> Right, I get that. What isn't clear to me is when you run bundle exec jekyll serve in your theme, Bundler reads in your Gemfile which specifies your .gemspec, which requires Jekyll but Jekyll isn't resolved yet so Bundler goes down in a shower of flames. Also, running bundle install will surely fail when Jekyll isn't already installed. This makes the UX for these tools much worse and perhaps impossible.

parkr:
This still something you want?

spudowiar:
@benbalter if you still want this, I'll fix it up

benbalter:
I'd definitely still like to see this, if we can find a way around @parkr's circular dependency concerns.

spudowiar:
@benbalter that was what my last comment fixed, I can push a commit for that then

PR #5364, opened 2016-09-16
Add ThemeAssetsReader which reads assets from a theme

parkr:
If the theme includes the 'assets' directory, it will be walked and items will be added to the site based on the normal rules of Jekyll: if there is YAML front matter, it will be added as a (convertible) Page, otherwise it will be added as a StaticFile. It will output to the `assets/` directory.

[Image available, but airplane WiFi is spotty.]

- [x] Add basic functionality
- [x] Ensure `jekyll new-theme` looks for `assets/` if present
- [x] Add tests
- [x] Update the docs

/cc @jekyll/ecosystem @benbalter @broccolini


.rubocop.yml
```diff
@@ -16,7 +16,7 @@ Lint/UnreachableCode:
 Lint/UselessAccessModifier:
   Enabled: false
 Metrics/AbcSize:
-  Max: 20
+  Max: 21
```
spudowiar: That's cheating! :wink:

parkr: 😇

lib/jekyll/page.rb
```diff
@@ -40,7 +40,11 @@ def initialize(site, base, dir, name)
       @base = base
       @dir  = dir
       @name = name
-      @path = site.in_source_dir(base, dir, name)
+      @path = if site.in_theme_dir(base) == base # we're in a theme
```
benbalter: Can you explain a bit more as to what's going on here?

lib/jekyll/readers/theme_assets_reader.rb
```diff
@@ -0,0 +1,47 @@
+module Jekyll
+  class ThemeAssetsReader
+    attr_reader :site
+    def initialize(site)
```
benbalter: Whitespace here?

lib/jekyll/readers/theme_assets_reader.rb
```diff
@@ -0,0 +1,47 @@
...
+    private
+    def read_theme_asset(path)
+      base = site.theme.root
+      dir = File.dirname(path.sub("#{site.theme.root}/", ""))
+      name = File.basename(path)
+
+      if Utils.has_yaml_header?(path)
+        append_unless_exists site.pages,
+          Jekyll::Page.new(site, base, dir, name)
+      else
```
benbalter: Really elegant implementation.

lib/jekyll/theme.rb
```diff
@@ -43,7 +47,7 @@ def path_for(folder)
     end
 
     def realpath_for(folder)
-      File.realpath(Jekyll.sanitized_path(root, "_#{folder}"))
+      File.realpath(Jekyll.sanitized_path(root, folder.to_s))
```
benbalter: I'm torn here. Do we want to pass vanilla symbols (without the `_`) and special case `assets` so that we treat all of them as named folders, rather than paths?

test/test_theme_assets_reader.rb
```diff
@@ -0,0 +1,61 @@
...
+    }, "Site should not have read in the #{relative_path} file, " \
+      "but it was found in #{haystack.inspect}"
+  end
+
+  context "with a valid theme" do
+    should "read all assets" do
+      ThemeAssetsReader.new(@site).read
+      assert_file_with_relative_path @site.static_files, "assets/img/logo.png"
+      assert_file_with_relative_path @site.pages, "assets/style.scss"
+    end
```
benbalter: Do you want to test that it doesn't clobber site files?

lib/jekyll/theme.rb
spudowiar: If we're passing symbols like `:_layouts`, why is it called `realpath_for`? It'd be something like `in_theme_dir` which makes more sense. I think `realpath_for` should be an abstraction (special case `:assets`, for example)

lib/jekyll/page.rb
spudowiar: @benbalter It essentially makes sure that we're normalising into the theme directory, rather than into the source directory, otherwise you might get `/home/user/work/home/user/.gems/[...]/theme.html` instead of `/home/user/.gems/[...]/theme.html`.

@parkr Have you tried this out with various symbolic links? I'm worried it might be a bit weird, depending on how `base` is set (I haven't checked, myself)

lib/jekyll/readers/theme_assets_reader.rb
```diff
@@ -0,0 +1,47 @@
...
+        else
+          read_theme_asset(path)
+        end
+      end
+    end
+
+    private
+    def read_theme_asset(path)
+      base = site.theme.root
+      dir = File.dirname(path.sub("#{site.theme.root}/", ""))
```
spudowiar: We do this same thing at least 2 different ways elsewhere, in the `Jekyll::Layout` we use the equivalent of `path.sub(site.theme.root, "")` and in the `Jekyll::Document` we use `Pathutil` and there are probably more ways we're doing this. :confounded:

test/test_theme.rb
```diff
@@ -56,7 +56,7 @@ def setup
 
     should "return the resolved path when a symlink & resolved path exists" do
       expected = File.expand_path("./_layouts", @expected_root)
-      assert_equal expected, @theme.send(:path_for, :symlink)
```
spudowiar: Not happy with this at all, `Theme#path_for` sounds like it should be an abstraction, looks like it's just a sugary version of `Site#in_theme_dir`

lib/jekyll/page.rb
parkr: @benbalter When you include `assets/styles.scss` with YAML front matter, it's a `Jekyll::Page`. To ensure it's read properly, we need to make sure we're looking in the theme's directory rather than your site's directory. This `if` says "if this Page is inside our theme directory then ensure the `@path` gets set properly."

lib/jekyll/readers/theme_assets_reader.rb
parkr: This is copied from another reader. I don't have a personal preference and we don't have it enforced in our Rubocop config.

lib/jekyll/readers/theme_assets_reader.rb
parkr: Would be nice to have a "relative_from" method or something but it depends on what you need. In this case, I want all directories within the theme directory without the forward slash at the beginning (e.g. `assets`). If you do `path.sub(site.theme.root, "") `, you get `/assets` and we can't use Pathutil because it doesn't work on Windows the way we're using it.

lib/jekyll/theme.rb
parkr: Special casing is a non-starter for me in this already complicated codebase. I want a way to fetch the path on disk for a directory. If the directory starts with a `_`, then let's ask for it that way. It's way more direct to ask for `_sass` and get `_sass` then to ask for `sass` and get `_sass`.

@spudowiar The difference here is that this uses `File.realpath` which resolves symlinks. `Site#in_theme_dir` does not.

test/test_theme.rb
parkr: It's not, as I mention above.

test/test_theme_assets_reader.rb
parkr: Clobber?

parkr:
@benbalter I think we're ready for another review. I left a few comments. As for "Update the default site to not to write a CSS file (and add that file to minima instead)", this will have to be in another PR which results from a minima release. I don't think that's a good idea to do here.

lib/jekyll/theme.rb
benbalter: Makes sense. Would a string be a better fit stylistically then?

test/test_theme_assets_reader.rb
benbalter: If I have `/assets/foo.js` in my site, and `/assets/foo.js` in my theme, to confirm `/assests/foo.js` in the resulting site contain's the site's content, not the theme's.

parkr:
> One question about if we need a test to ensure the files in the site's assets folder are given preference over files in the theme's asset folder

Done in 29d8fee.

parkr:
@jekyllbot: merge +minor

PR #5380, opened 2016-09-19
Only complain about `kramdown.coderay` if it is actually in the config

ethomson:
After upgrading to jekyll 3, I wanted to continue using coderay as my highlighter.  So I updated my configuration to:

```
kramdown:
  syntax_highlighter: coderay
  syntax_highlighter_opts:
    line_numbers: inline
    css: class
```

And I was surprised when jekyll complained:

```
Deprecation: You are using 'kramdown.coderay' in your configuration, please use 'syntax_highlighter_opts' instead.
```

It seems that jekyll is raising this notice whenever `highlighter == "coderay"`, even when the `kramdown.coderay` configuration setting is nonexistent.

Switching this to check for a non-empty `kramdown.coderay` hash keeps this warning when the user has `kramdown.coderay` settings, but suppresses this warning when the configuration has switched to the new style `kramdown.syntax_highlighter_opts`.

### DIFF
--- a/lib/jekyll/converters/markdown/kramdown_parser.rb
+++ b/lib/jekyll/converters/markdown/kramdown_parser.rb
@@ -104,7 +104,7 @@ module Jekyll
 
         private
         def modernize_coderay_config
-          if highlighter == "coderay"
+          unless @config["coderay"].empty?
             Jekyll::Deprecator.deprecation_message(
               "You are using 'kramdown.coderay' in your configuration, " \
               "please use 'syntax_highlighter_opts' instead."


lib/jekyll/converters/markdown/kramdown_parser.rb
```diff
@@ -104,7 +104,7 @@ def strip_coderay_prefix(hash)
 
         private
         def modernize_coderay_config
-          if highlighter == "coderay"
+          if !@config["coderay"].empty?
```
pathawks: I can't speak to the wider implications of this change, but as far as style goes, could we use `unless` instead of `if !`?
I think Rubocop should be complaining about this.

ethomson: Sure; fixed.

ashmaroli: Rubocop does infact complain about this..

parkr:
Rubocop errors are fixed on `master`. We can merge without them.

parkr:
@jekyllbot: merge +buf

ethomson:
Thanks @parkr !  ❤️

PR #5381, opened 2016-09-19
Appease Rubocop

pathawks:
Rubocop 0.43.0 seems to have some new expectations.

`rubocop -a`

### DIFF
--- a/lib/jekyll/hooks.rb
+++ b/lib/jekyll/hooks.rb
@@ -54,7 +54,7 @@ module Jekyll
 
     # Ensure the priority is a Fixnum
     def self.priority_value(priority)
-      return priority if priority.is_a?(Fixnum)
+      return priority if priority.is_a?(Integer)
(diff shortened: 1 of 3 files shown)


lib/jekyll/hooks.rb
```diff
@@ -54,7 +54,7 @@ def self.register(owners, event, priority: DEFAULT_PRIORITY, &block)
 
     # Ensure the priority is a Fixnum
```
pathawks: Should I replace `Fixnum` with `Integer` in this comment too?

envygeeks: Technically it's still right regardless, it's up to you at that point tbh.

parkr:
@jekyllbot: merge +dev

parkr:
FWIW the thing about `10` being in the filename was a false positive. Tracked in bbatsov/rubocop#3514.

PR #5389, opened 2016-09-21
remove features' directories on windows with proper access

ashmaroli:
Feature directories generated during cucumber tests currently throw a `Errno::EACCES` exception after every scenario runs on windows.
This Pull resolves this issue.

--
/cc @jekyll/windows

### DIFF
--- a/features/step_definitions.rb
+++ b/features/step_definitions.rb
@@ -1,4 +1,5 @@
 Before do
+  FileUtils.rm_rf(Paths.test_dir) if Paths.test_dir.exist?
   FileUtils.mkdir_p(Paths.test_dir) unless Paths.test_dir.directory?
   Dir.chdir(Paths.test_dir)
 end
@@ -6,7 +7,7 @@ end
 #
 
 After do
-  Paths.test_dir.rmtree if Paths.test_dir.exist?
+  FileUtils.rm_rf(Paths.test_dir) if Paths.test_dir.exist?
   Paths.output_file.delete if Paths.output_file.exist?
   Paths.status_file.delete if Paths.status_file.exist?
   Dir.chdir(Paths.test_dir.parent)


XhmikosR:
Which tests does this fix exactly? The cucumber one?

ashmaroli:
> Which tests does this fix exactly? The cucumber one?

yes.. thought you'd understand if I just mentioned *features*
I'll update the body..

XhmikosR:
It works as far as I can tell

before:

```
201 scenarios (201 failed)
1638 steps (7 failed, 2 skipped, 1629 passed)
```

after:

```
201 scenarios (7 failed, 194 passed)
1638 steps (7 failed, 2 skipped, 1629 passed)
```

XhmikosR:
But there seems to be a failure in Travis.

features/step_definitions.rb
```diff
@@ -1,12 +1,13 @@
 Before do
+  FileUtils.rm_rf(Paths.test_dir) if Paths.test_dir.exist?
```
ashmaroli: I added this line to *ensure* every scenario starts with a clean slate

ashmaroli:
> But there seems to be a failure in Travis.

Thats a known bug.. happens randomly to random suites.. Maintainers can restart that test to get it to pass..
Cucumber passed with Green..

features/step_definitions.rb
```diff
@@ -1,12 +1,13 @@
...
   FileUtils.mkdir_p(Paths.test_dir) unless Paths.test_dir.directory?
   Dir.chdir(Paths.test_dir)
 end
 
 #
 
 After do
-  Paths.test_dir.rmtree if Paths.test_dir.exist?
```
parkr: Out of curiosity, why does `.rmtree` fail? Can you paste the error or describe the issue? It should be identical to `FileUtils.rm_rf` IIRC.

parkr:
@jekyllbot: merge +dev

features/step_definitions.rb
ashmaroli: `.rmtree` does not fail per se. The feature directories do get removed completely (verified locally), but on Windows it also raises an alarm:
```
  Permission denied @ dir_s_rmdir - C:/projects/jekyll/tmp/jekyll (Errno::EACCES)
  C:/projects/jekyll/features/step_definitions.rb:9:in `After'
```
I stumbled upon the fix when I tried the *identical*  `FileUtils.rm_rf` method. This method manipulates files rather than try to modify the directory attribute.

parkr: Interesting! @envygeeks, is this a part of Pathutil?

PR #5391, opened 2016-09-21
Support HTML comment front matter

trans:
Adjusted YAML_FRONT_MATTER_REGEXP to support front matter in XML/HTML comment.

Useful for using page/post sources that might not otherwise support YAML front matter, e.g. a wiki.


trans:
I overlooked an obvious mistake when inserting the adjusted regexp, so I am going to close this and resubmit. Sorry.

Issue #5370, opened 2016-09-18
jekyllrb development [watch] on windows


ashmaroli:
Its a known fact that `--watch` is a bit buggy on windows platform.
`build.rb` warns windows users about it : 
```ruby
# lib/jekyll/commands/build.rb#L75-L77
if Utils::Platforms.windows?
  Jekyll.logger.warn "", "--watch arg is unsupported on Windows. "
  Jekyll.logger.warn "", "If you are on Windows Bash, please see: " \
    "https://github.com/Microsoft/BashOnWindows/issues/216"
```
This warning also pops up when `jekyllrb site` is locally previewed using `bundle exec rake site:preview` .
But then, there is no problem with auto-regeneration, when a **newly generated blog** is first served and then edited live, on a **windows** machine. 

So I set out to have `site/` mimic the same process used by a newly-generated blog being served.
Instead of modifying `rake/site.rake` I placed a Gemfile at `site/` and `exclude`d it and its `.lock` from processing.
```
~/site> bundle exec jekyll serve
```
then allowed me to `watch` the `site/` dir contents for any changes.

--

Now, the question is, 
```
"Is this technique valid enough to be considered for a merge?"
```
:smile: 

--
/cc @jekyll/documentation

DirtyF:
I couldn't tell, I don't use Windows. 
cc/ @jekyll/windows for a feedback on this.

parkr:
> But then, there is no problem with auto-regeneration, when a newly generated blog is first served and then edited live, on a windows machine.

Can you explain how this is? From what I understand, this is not the case because the underlying library that `listen` uses to track filesystem changes is not particularly robust on Windows.

ashmaroli:
> Can you explain how this is?

I would gladly do so if I could. But unfortunately, no. I've not dug out the answer yet.
But, I can provide more info surrounding this:

--
Since pristine `Jekyll 3.2.1` cant render layouts properly on Windows, my local version has been tweaked to enable that.
I'm on x64 Windows 7 with a 32-bit Ruby 2.3.1 and its allies.

--

This is the console log after serving a newly generated blog:
```
bundle exec jekyll serve


Configuration file: C:/qwertyui/_config.yml
            Source: C:/qwertyui
       Destination: C:/qwertyui/_site
 Incremental build: disabled. Enable with --incremental
      Generating...
                    done in 0.796 seconds.
  Please add the following to your Gemfile to avoid polling for changes:
[...]

PR #5393, opened 2016-09-22
add a Gemfile to enable serving jekyllrb.com through bundler

ashmaroli:
Ref: #5370 

Since `--watch` is known to not work on Windows, previewing an *active* local copy of `https://jekyllrb.com` is not possible by conventional methods.
This can be bypassed by adding a Gemfile to the site directory and then having bundler start the jekyll server.

--
/cc @jekyll/windows

### DIFF
--- a/site/Gemfile
+++ b/site/Gemfile
@@ -1,0 +1,13 @@
+source "https://rubygems.org"
+
+gem "jekyll", "~> 3.2"
+
+group :jekyll_plugins do
+  gem "jemoji"
+  gem "jekyll-sitemap"
+  gem "jekyll-seo-tag"
+  gem "jekyll-avatar"
+  gem "jekyll-mentions"
+  gem "jekyll-feed"
+  gem "jekyll-redirect-from"
+end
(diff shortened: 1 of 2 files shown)


envygeeks:
@ashmaroli you know that bundler is smart enough to find the Gemfile in `../`? Did you `bundle install --with=site`? These dependencies should be in the parent Gemfile and then from the `site/` folder you can `bundle exec jekyll` and it pick up the Gemfile in `../`, this is uncessary since we've already a Gemfile, any changes should go there.

ashmaroli:
> you know that bundler is smart enough to find the Gemfile in ../?

I honestly did not know that..

--
But on trying if this would be a viable solution, it returned the known warning.

ashmaroli:
Sorry @DirtyF, I shouldnt have tagged the documentation team on this.. its related to a `Windows` issue..

ashmaroli:
*update:*
just found out that if I use Jekyll repo HEAD instead of jekyll-3.2.1, I loose the ability gained from the change proposed here.. so closing this...

ashmaroli:
@envygeeks Do you have a Windows machine at hand? I would like to know what exactly changed with **HEAD** that removed this feature from Jekyll 3.3.. 
(autoregeneration is still working fine on my system with up-to-date gems and Jekyll 3.2.1)

PR #5399, opened 2016-09-22
Add absolute_url and relative_url filters.

parkr:
When creating sites and themes, there's a lot of confusion around `site.url` and `site.baseurl`. These new filters attempt to unify the way we calculate relative and absolute URL's. The new filters are:

- `relative_url` - takes a URL and returns a url with `site.baseurl` properly prepended. This is useful when you want to be host-agnostic.
- `absolute_url` - takes a URL and returns a url with `site.baseurl` and `site.url` properly appended.

Examples, assuming `site.baseurl = "/project"` and `site.url = "http://example.com"`:

```liquid
{{ "/css/main.scss" | relative_url }}
<!-- "/project/css/main.scss" -->

{{ "/css/main.scss" | absolute_url }}
<!-- "http://example.com/project/css/main.scss" -->
```

If your `site.baseurl` or `site.url` values aren't set, it will just append what it can dependeing on the desired output of the filter. Read the tests in the diff for more on this.

/cc @jekyll/ecosystem @jekyll/core @jekyll/stability

PR #5405, opened 2016-09-23
Add documentation for relative_url and absolute_url

parkr:
@benbalter @pathawks I can't *believe* I forgot this! Follow-up to #5399 for documentation.

screen shot 2016-09-23 at 2 27 23 pm

### DIFF
--- a/site/_docs/templates.md
+++ b/site/_docs/templates.md
@@ -23,6 +23,34 @@ common tasks easier.
   <tbody>
     <tr>
       <td>
+        <p class="name"><strong>Relative URL</strong></p>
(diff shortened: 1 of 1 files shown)


parkr:
Oh! Forgot to /cc @jekyll/documentation on this. Slated for Jekyll 3.3.

parkr:
@jekyllbot: merge +site

QXSoftware:
The absolute_url is good, but it can only return **http** url. What about a **https** absolute_url ?

pathawks:
> The absolute_url is good, but it can only return **http** url. What about a **https** absolute_url?

I'm not sure I understand. When does it _not_ return an **https** url?

PR #5408, opened 2016-09-24
swallow bundle output from `jekyll new` while in CI

ashmaroli:
patches `commands/new.rb` to swallow outputs by `bundle install` when running in a CI environment.
This removes unnecessary clogging of CI logs.

--
/cc @jekyll/ecosystem


ashmaroli:
I would also like the outputs to be hidden in local tests. 
Any suggestions?

--
/cc @jekyll/ecosystem

parkr:
Thanks!

@jekyllbot: merge +dev

PR #5410, opened 2016-09-24
Fix handling of non-ASCII characters in new `*_url` filters

pathawks:
How/where should normalization be handled? I would love if the filters could also be responsible for normalization.

### DIFF
--- a/jekyll.gemspec
+++ b/jekyll.gemspec
@@ -38,4 +38,5 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('jekyll-sass-converter', '~> 1.0')
   s.add_runtime_dependency('jekyll-watch', '~> 1.1')
   s.add_runtime_dependency("pathutil", "~> 0.9")
+  s.add_runtime_dependency('addressable', '~> 2.4')
 end
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -1,3 +1,5 @@
+require "addressable/uri"
+
 module Jekyll
   module Filters
     module URLFilters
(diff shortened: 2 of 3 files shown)


parkr:
Punycode, you are the worst. Encoding URL's is notorious. I think this could be a fix for after 3.3 just because it's so gnarly.

Does Ruby stdlib have anything for this?

jekyll.gemspec
```diff
@@ -38,4 +38,5 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('jekyll-sass-converter', '~> 1.0')
   s.add_runtime_dependency('jekyll-watch', '~> 1.1')
   s.add_runtime_dependency("pathutil", "~> 0.9")
+  s.add_runtime_dependency('addressable', '~> 2.4')
```
pathawks: **Addressable** was already being installed implicitly, so this is not _actually_ adding a new dependency.

pathawks:
> I think this could be a fix for after 3.3 just because it's so gnarly.

Oh please no. I feel like plugins won't be able to rely on this filter until this is resolved.

> Does Ruby stdlib have anything for this?

I don't think so, and that's why [Addressable](https://github.com/sporkmonger/addressable) exists. It was already an implicit dependency, so I don't see a problem using it here.

jekyll.gemspec
parkr: Doesn't look like it's a dependency of jekyll, but it a dependency of `jekyll-sitemap` per a Gemfile.lock of mine using both. So it is technically not used yet.

pathawks: Ok, that makes sense.

pathawks:
Travis likes it! ✅

parkr:
I'm OK with this. @jekyll/stability, any thoughts?

parkr:
@jekyllbot: merge +bug

PR #5411, opened 2016-09-24
Replace classic box-sizing reset with inheritance reset

zdroid:
Reference: https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/

### DIFF
--- a/site/_sass/_style.scss
+++ b/site/_sass/_style.scss
@@ -1,10 +1,14 @@
 /* Base */
 
-* {
-  -webkit-box-sizing: border-box;
-  -moz-box-sizing: border-box;
-  box-sizing: border-box;
-}
+ html {
+   box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


site/_sass/_style.scss
```diff
@@ -1,10 +1,14 @@
 /* Base */
 
-* {
-  -webkit-box-sizing: border-box;
-  -moz-box-sizing: border-box;
-  box-sizing: border-box;
-}
+ html {
+   box-sizing: border-box;
```
parkr: Why not `-webkit` and `-mox`?

zdroid: Check it out here :wink:: http://caniuse.com/#search=box-sizing

parkr:
@jekyllbot: merge +site

PR #5412, opened 2016-09-24
Add .editorconfig

zdroid:
As suggested in jekyll/minima#44 discussion.

Humans make mistakes, that's simple. It's easy to prevent the most common mistakes with adding one standardized file of 14 lines.

Another editorconfig-related thread: jekyll/jekyll#3527

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false


parkr:
The reason we originally rejected this was because it isn't standard – it's a plugin for everything. Ir reminds me [of this comic](https://xkcd.com/927/). It's hard to actually enforce. We have rubocop checks for this now. I accepted it into jekyll/minima because it doesn't have any checks like we have here with rubocop.

I'm ambivalent. Does anyone in @jekyll/core have a strong preference?

zdroid:
It is hard to be enforced to all users, but trust me, it'd help a lot. Myriad projects tried it out and it works fine. Atom, Brackets and Sublime support it, and they're surely the three most used code editors in the GitHub community.

parkr:
Ok, we'll tweak it if it starts to interfere with any rubocop rules.

@jekyllbot: merge +dev

zdroid:
Glad to hear that. :+1:

PR #5413, opened 2016-09-25
Fix Travis.ci documentation

tf2manu994:
You can [see](https://travis-ci.org/tf2manu994/tf2manu994.github.io/builds/162424169) that without this fix, it crashes and burns.

`Gem::InstallError: activesupport requires Ruby version >= 2.2.2.`

Which is a dependency.

### DIFF
--- a/site/_docs/continuous-integration.md
+++ b/site/_docs/continuous-integration.md
@@ -94,7 +94,7 @@ Your `.travis.yml` file should look like this:
 ```yaml
 language: ruby
 rvm:
-- 2.2.2
+- 2.2.5
 
 before_script:
  - chmod +x ./script/cibuild # or do this locally and commit
@@ -127,7 +127,7 @@ access to Bundler, RubyGems, and a Ruby runtime.
 
 ```yaml
 rvm:
-- 2.2.2
+- 2.2.5
 ```
 
 RVM is a popular Ruby Version Manager (like rbenv, chruby, etc). This
--- a/site/_docs/continuous-integration.md
+++ b/site/_docs/continuous-integration.md
@@ -94,7 +94,7 @@ Your `.travis.yml` file should look like this:
 ```yaml
 language: ruby
 rvm:
-- 2.1
+- 2.2.2
 
 before_script:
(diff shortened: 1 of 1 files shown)


ashmaroli:
Jekyll doesn't depend on activesupport-5+ actually.

--
/cc @jekyll/documentation

tf2manu994:
@ashmaroli html-proofer does though. 

```
$ gem install html-proofer
Fetching: concurrent-ruby-1.0.2.gem (100%)
Successfully installed concurrent-ruby-1.0.2
Fetching: minitest-5.9.0.gem (100%)
Successfully installed minitest-5.9.0
Fetching: thread_safe-0.3.5.gem (100%)
Successfully installed thread_safe-0.3.5
Fetching: tzinfo-1.2.2.gem (100%)
Successfully installed tzinfo-1.2.2
Fetching: i18n-0.7.0.gem (100%)
Successfully installed i18n-0.7.0
Fetching: activesupport-5.0.0.1.gem (100%)
ERROR:  Error installing html-proofer:
	activesupport requires Ruby version >= 2.2.2.
```

ashmaroli:
I saw that too..
jekyll doesnt directly *require* `html-proofer` to be installed either.. That gem recently released a new version that depended on a higher version of activesupport.
I cant see why jekyll documentation has to be altered for that change...

tf2manu994:
Updated to shave some time off build if that was the concern

ashmaroli:
> shave some time off build

Please explain more..

tf2manu994:
I assumed the reason we were using the older version of ruby was to take advantage of the fact that it was preinstalled on Travis. Travis already preinstalls 2.2.5, so it's more efficient to install 2.2.5 than 2.2.2

DirtyF:
Please bear in mind that according to https://pages.github.com/versions/ ruby version is still 2.1 for GitHub Pages.

tf2manu994:
If we're going by that, should we not use the github-pages gem?

peternewman:
It looks like it needs a note in the .travis.yml to update the docs when the Travis config is updated.

parkr:
@jekyllbot: merge +dev

Issue #5414, opened 2016-09-25
Faulty handling of `timezone` key on Windows


ashmaroli:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I updated to the latest Jekyll
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
## 
- [x] I am on (or have tested on) **_Windows**_ 7+
## 
- [x] It was another bug.
## Steps to Reproduce this Bug:
- `jekyll new my_blog`
- `cd my_blog`
- add `timezone: America/New_York` to `_config.yml`
- `bundle exec jekyll serve`
- Goto `http://localhost:4000/jekyll/update/2016/09/25/welcome-to-jekyll.html`
- Inspect the `post-meta`  
## The Output I Wanted

Expected the `post-meta` to show correct TZ data

---

**N.B. :** This bug doesn't occur as long as I don't set the `timezone:` key in `_config.yml` 
         I have tried adding gem `tzinfo-data` to my Gemfile and bundling.  
**Addional Experiments:** [PR-comment](https://github.com/jekyll/jekyll/pull/5241#issuecomment-249405858)
## 

/cc @jekyll/windows, @jekyll/stability

parkr:
This is because we use the `TZ` environment variable in `Jekyll.set_timezone.` If you can find a way to make this work on Windows, that'd be dope.

ashmaroli:
## Temporary Workaround *
*Required only for Windows users or for cross-platform compatibility*
set the Timezone **based** on [GNU TZ Guidelines](http://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html) for POSIX Systems ** :

If you're in a timezone that's **4 hours behind** UTC, then set `timezone: UTC+4` or `UTC4` or `UTC+04:00`
If you're in a timezone that's **4 hours ahead of** UTC, then set `timezone: UTC-4` or `UTC-04:00`
Similarly, if you're in a timezone that's **5 1/2 hours ahead of** UTC, then set `timezone: UTC-05:30`

--
_* Does not guarantee this will hold its ground when DST is not in effect_
_** The 3-letter-word has no particular effect on the variable. `UTC` == `XYZ` but chose it for semantics_

parkr:
> If you're in a timezone that's 4 hours behind UTC, then set timezone: UTC+4 or UTC4 or UTC+04:00
[...]

PR #5416, opened 2016-09-25
site_configuration.feature: use UTC format in timezone

ashmaroli:
switch to date/time in POSIX format as value to `timezone:` key in `features/site_configuration.feature` since Windows does not seem to understand tz-name specified currently.
Ref: #5414 

--
/cc @jekyll/windows

### DIFF
--- a/features/site_configuration.feature
+++ b/features/site_configuration.feature
@@ -161,8 +161,8 @@ Feature: Site configuration
       And I have a post layout that contains "Post Layout: {{ content }} built at {{ page.date | date_to_xmlschema }}"
       And I have an "index.html" page with layout "page" that contains "site index page"
       And I have a configuration file with:
-        | key         | value            |
-        | timezone    | America/New_York |
+        | key         | value        |
+        | timezone    | UTC+04:00    |
       And I have a _posts directory
       And I have the following posts:
         | title     | date                   | layout  | content             |
@@ -181,8 +181,8 @@ Feature: Site configuration
       And I have a post layout that contains "Post Layout: {{ content }} built at {{ page.date | date_to_xmlschema }}"
       And I have an "index.html" page with layout "page" that contains "site index page"
       And I have a configuration file with:
-        | key         | value               |
-        | timezone    | Pacific/Honolulu    |
+        | key         | value        |
+        | timezone    | UTC+10:00    |
       And I have a _posts directory
       And I have the following posts:
         | title     | date                   | layout  | content             |


parkr:
LGTM.

@jekyllbot: merge +dev

ashmaroli:
**Note to Self:** *Wait till November (end of DST) to see if this causes any CI failure and then publish documentation.*

budparr:
How does this affect DST? Does this change mean I'd have to change the UTC time on all my sites every year? And is it backwards compatible with all the sites I have this set for?

ashmaroli:
@budparr, the change has been made to just this one feature.  It will not affect any of the Jekyll sites created till date, and yet to be created. 
Besides, you don't have to change anything at all if you don't plan on previewing the site on Windows 

> How does this affect DST?

I'm yet to confirm.. In theory, it would create problems, but it may not, too.. 
I was planning to publish related documentation for Windows users who currently face the issue outlined in #5414 but decided to put off till I have solid evidence to back it.

> Is this change really necessary?

Yes, if you're in **development** mode on **Windows** and for Appveyor to pass that feature.

budparr:
Not sure, I understand. I use the timezone setting in config in all my sites and use critical date-based functionality. Are you saying that one can continue to use `America/New_York`? as my time-zone setting, and not worry about DST?

ashmaroli:
`America/New_York` is still valid (and automatically handles DST), except on Windows.
I had issues with it on my Windows system. If you're on Windows too, and had this setting working all this while, I'm willing to consider looking deeper into my system configuration.

budparr:
> America/New_York is still valid (and automatically handles DST)

Okay, thanks. That's what I wanted to know.

features/site_configuration.feature
```diff
@@ -161,8 +161,8 @@ Feature: Site configuration
       And I have a post layout that contains "Post Layout: {{ content }} built at {{ page.date | date_to_xmlschema }}"
       And I have an "index.html" page with layout "page" that contains "site index page"
       And I have a configuration file with:
-        | key         | value            |
-        | timezone    | America/New_York |
+        | key         | value        |
+        | timezone    | UTC+04:00    |
```
parkr: This should be UTC-4 if it is meant to be New York but you notice how the posts have their own offset so they are unaffected by this:

```text
| entry2    | 2013-04-10 03:14 -0400 | post    | content for entry2. |
```

becomes: 

```text
Post Layout: <p>content for entry2.</p>\n built at 2013-04-10T03:14:00-04:00
```

That seems to invalidate this test. I believe we'd need to remove the timezone offsets in the `date` segment of the posts.

ashmaroli: A [thread on StackExchange](http://unix.stackexchange.com/questions/104088/why-does-tz-utc-8-produce-dates-that-are-utc8) for reference..

ashmaroli: > I believe we'd need to remove the timezone offsets in the date segment of the posts.

local tests show, there's no difference to whether the `post.date` is present or not..

ashmaroli:
@parkr may I suggest an edit to `History.markdown` along the lines of the new title..?

PR #5420, opened 2016-09-27
Site template: Move contents of the index.html page to the 'home' layout

DirtyF:
Reflects changes in https://github.com/jekyll/minima/pull/48

/cc @jekyll/core


ashmaroli:
@DirtyF, few questions:
- what do you say to adding a comment  telling users they *need not* edit the empty `index.html` they now see..?
- is using index.md instead of `index.html` a bad-idea? 
If done so, we could add a `title: Home; permalink: /` and minima will automatically add this page to the top-nav

DirtyF:
@ashmaroli good catch, applied your recommendations, except for title and permalink as It seems perfectly fine to me as it is.

ashmaroli:
> except for title and permalink as it seems perfectly fine to me as it is.

Thanks Frank, but according to [this](https://github.com/jekyll/minima/blob/master/_includes/header.html#L20-L22), unless those parameters are set, the top-nav wont have the link defined. 
I wont press on including them though, as this [line](https://github.com/jekyll/minima/blob/master/_includes/header.html#L7) already links to the blog's root.

parkr:
@jekyllbot: merge +minor

PR #5428, opened 2016-09-29
[docs] Update documentation regarding `bundle install` after `jekyll new`

ashmaroli:
Update the docs to mention that `jekyll new` now runs `bundle install` automatically for every new site and that they may choose to skip that with the `--skip-bundle` switch.

Ref: #5237

--
/cc @jekyll/documentation


parkr:
LGTM.

@jekyllbot: merge +site

parkr:
@jekyllbot: merge +site

PR #5431, opened 2016-09-30
set site.url in dev environment to `http://localhost:4000`

Crunch09:


test/test_commands_serve.rb
```diff
@@ -19,6 +19,11 @@ def custom_opts(what)
           p
         )
       end
+      allow(SafeYAML).to receive(:load_file).and_return({})
+      allow(Jekyll::Commands::Build).to receive(:build).and_return("")
+    end
+    teardown do
+      Jekyll.instance_variable_set(:@sites, [])
```
parkr: You should be able to do [`Jekyll.sites.clear`](https://docs.ruby-lang.org/en/2.0.0/Array.html#method-i-clear) here.

lib/jekyll/commands/serve.rb
```diff
@@ -33,6 +33,8 @@ def init_with_program(prog)
               opts["serving"] = true
               opts["watch"  ] = true unless opts.key?("watch")
               config = opts["config"]
+              default_url = "http://localhost:4000"
```
parkr: One can set the `host` and `port` options in the command line or via your config file. If I set my `--port` to `3999`, I should get my default url as `http://localhost:3999`. Same with `--host`.

test/test_commands_serve.rb
Crunch09: 🙈 thanks!

lib/jekyll/commands/serve.rb
Crunch09: @parkr Thanks! Should i also take `baseurl` and `ssl` options into consideration?

parkr: `ssl` yes, `baseurl` nope!

envygeeks:
I'm confused as to what is going on here? We already do all this? See: https://github.com/jekyll/jekyll/blob/master/lib/jekyll/commands/serve.rb#L119 that has been in there for quite some time and should automatically change itself based on everything in the options.

envygeeks:
Oh, I see what's going on, okay, yeah, can you just use our helper method that will generate everything for you and put it onto site? That would make your life much easier @Crunch09!

envygeeks:
Oh, you need baseurl not to be there as per @parkr's specs, you can adjust that method to exclude it if you want, add a conditional into the params and then use that conditional in the builder and you can exclude it!

lib/jekyll/commands/serve.rb
```diff
@@ -33,6 +33,8 @@ def init_with_program(prog)
               opts["serving"] = true
               opts["watch"  ] = true unless opts.key?("watch")
               config = opts["config"]
+              url = default_url(opts)
+              opts = opts.merge({ "url" => url }) if Jekyll.env == "development"
```
parkr: Let's avoid the `url` variable assignment here and just do 

```ruby
opts.merge!({ "url" => default_url(opts) }) if Jekyll.env == "development"
```

test/test_commands_serve.rb
```diff
@@ -79,16 +85,53 @@ def custom_opts(what)
         custom_options = {
           "config"  => %w(_config.yml _development.yml),
           "serving" => true,
-          "watch"   => false # for not having guard output when running the tests
+          "watch"   => false, # for not having guard output when running the tests
+          "url"     => "http://localhost:4000"
```
parkr: Why is this added?

lib/jekyll/commands/serve.rb
Crunch09: Ah thanks, i only tried it with the non-bang version and then the line was too long :)

test/test_commands_serve.rb
Crunch09: This test checks which options are passed to `Serve.process` and the URL is now also included in that options hash.

lib/jekyll/commands/serve.rb
parkr: Rubocop corrected me further:

```ruby
opts["url"] = default_url(opts) if Jekyll.env == "development"
```

Crunch09: Thanks, just wanted to post the same thing 😆 Updated commit coming in a few minutes

Crunch09:
@envygeeks I refactored it now to also use the `server_address` method.

parkr:
@jekyllbot: merge +minor

benbalter:
Just to note, and this doesn't need to hold up 3.3., that we went from passing `server` and a hash of named arguments, to passing four unnamed arguments to `server_address` (meaning every option we add now has to be an argument). I'd think passing the options hash around, rather than the individual arguments would be a bit cleaner and future proof.

(Noticed because of https://github.com/jekyll/jekyll/issues/5451).

Frikki:
Why were there no test suite that had shown this error https://github.com/jekyll/jekyll/issues/5451 when these changes were made?

benbalter:
> Why wasn’t there a test suite showning this error #5451 when these changes were made?

Because JekyllAdmin is a plugin and distributed/versioned independently.

Crunch09:
> I'd think passing the options hash around, rather than the individual arguments would be a bit cleaner and future proof.

@benbalter Yes, it probably would be more future proof to just use an option hash. But on the other hand i didn't think of any additional arguments we could pass to create a server address and it also just was a private method.

parkr:
@benbalter @Crunch09 I think the best option forward is to revert the signature change, then. It was just for aesthetics AFAICT.

Crunch09:
@parkr You're right, just to DRY things up. @benbalter

PR #5433, opened 2016-10-01
Fix loading data from subdir with a period in name

### LINKED ISSUE #5429: Jekyll incorrectly processes _data subdirs with a period in the name

---

---

## Issue
I'm unable to access data in a  `_data/` subdirectory via the `site.data` array if that subdirectory includes a period `.` in the name/path. It seems jekyll strips everything after the period when storing in the `site.data` array.

## My Reproduction Steps
```
mkdir _data/0.1
touch  _data/0.1/this.yml
```

Then I use this code to simply print `site.data`:
```
{{ site.data }}
```
I see this as the output:
```
{"0"=>{"this"=>false}}
```

I expect this:
```
{"0.1"=>{"this"=>false}}
```

ayastreb:
Fix #5429
If we're processing a directory - use it's name as a key.

### DIFF
--- a/lib/jekyll/readers/data_reader.rb
+++ b/lib/jekyll/readers/data_reader.rb
@@ -37,10 +37,10 @@ module Jekyll
         path = @site.in_source_dir(dir, entry)
         next if @entry_filter.symlink?(path)
 
-        key = sanitize_filename(File.basename(entry, ".*"))
         if File.directory?(path)
-          read_data_to(path, data[key] = {})
+          read_data_to(path, data[entry] = {})
         else
+          key = sanitize_filename(File.basename(entry, ".*"))
           data[key] = read_data_file(path)
         end
       end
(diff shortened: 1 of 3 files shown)


lib/jekyll/readers/data_reader.rb
```diff
@@ -37,10 +37,10 @@ def read_data_to(dir, data)
         path = @site.in_source_dir(dir, entry)
         next if @entry_filter.symlink?(path)
 
-        key = sanitize_filename(File.basename(entry, ".*"))
         if File.directory?(path)
-          read_data_to(path, data[key] = {})
+          read_data_to(path, data[entry] = {})
```
parkr: I think we might still need to do `sanitize_filename` here to make this accessible in Liquid.

ayastreb: ok, so "categories.01" will become "categories01", right?

parkr:
@jekyllbot: merge +bug

PR #5441, opened 2016-10-03
Remove autoload of Draft which no longer exists.

mhw:
Referencing Jekyll::Draft raises 'LoadError: cannot load such file -- jekyll/draft'


parkr:
@jekyllbot: merge +bug

Issue #5407, opened 2016-09-24
[Feature Request] semantic liquid filters for assets


ashmaroli:
Placing a request to have semantic filters to pull in assets from a theme gem.
Close on the heels of the newly added `{{ 'abc' | absolute_url }}` & `{{ 'abc' | relative_url }}`, I'd like to suggest the following as well:
- [asset_url](#asset-url)
- [img_url](#img-url)
- [javascript_url](#javascript-url)
- [stylesheet_url](#stylesheet-url)

---
## asset_url

this filter adds any file that is an immediate child of `assets/` directory:

``` html
<!-- 
look for "assets/style.css" or "assets/logo.png" or 
"assets/script.js" or "assets/font.ttf"
 -->

{{ 'style.css' | asset_url }}
{{ 'logo.png' | asset_url }}
{{ 'script.js' | asset_url }}
{{ 'font.ttf' | asset_url }}
```
## 
## img_url

this filter adds an image file from within `assets/img/` or `assets/image/` :

``` html
<!-- look for "assets/img/logo.png" or "assets/images/logo.png" -->

{{ 'logo.png' | img_url }}
```
## 
## javascript_url

this filter adds a script file from `assets/js/` or `assets/scripts/`

``` html
<!-- look for "assets/js/script.js" or "assets/scripts/script.js" -->

{{ 'script.js' | javascript_url }}
```
## 
## stylesheet_url

this filter adds the stylesheet from `assets/css/` or `assets/styles/`

``` html
<!-- look for "assets/css/style.css" or "assets/styles/style.css" -->

{{ 'style.css' | stylesheet_url }}
```

---

N.B. I'm aware of existing similars via `jekyll-assets` plugin
## 

/cc @jekyll/ecosystem

mdxprograms:
Could this be added in a more general sense if someone supplies `javascript_path` in `_config.yml` or is this already an option?

ashmaroli:
I have not tried that yet, but if possible, that sounds like a great choice.

benbalter:
[...]

PR #5445, opened 2016-10-04
add `asset_url` filter

ashmaroli:
Add `asset_url` filter that prepends `/[baseurl]/assets/` to the `relative_url` of the given asset-file-path.

```
{{ "css/main.css" | asset_url }}
# -> "/baseurl/assets/css/main.css"
```
- [x] Add `asset_url` filter
- [x] Add tests
- [x] Add documentation

Ref: #5407 
## 

/cc @jekyll/ecosystem

### DIFF
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -29,6 +29,20 @@ module Jekyll
         ).normalize.to_s
       end
 
+      # Produces the path to the given asset relative to the assets directory at the
+      # domain root based on site.baseurl
+      #
+      # input - the asset file relative to the 'assets' directory. The string may also
+      #         include any subfolder(s) the asset may be a part of.
+      #           e.g. "main.css" or "css/main.css"
+      #
+      # Returns a path relative to the assets directory as a String.
+      def asset_url(input)
+        relative_url(
+          "/assets" + ensure_leading_slash(input.to_s)
+        )
+      end
+
       private
       def ensure_leading_slash(input)
         return input if input.nil? || input.empty? || input.start_with?("/")
--- a/site/_docs/templates.md
+++ b/site/_docs/templates.md
@@ -50,6 +50,30 @@ common tasks easier.
       </td>
     </tr>
     <tr>
+      <td rowspan="2">
+        <p class="name"><strong>Asset URL</strong></p>
+        <p>Prepend <code>baseurl</code> value and string <code>/assets</code> to the input. This will produce a relative URL to just the assets directory. Subdirectories, if present, may be passed along with the input as well.</p>
+      </td>
+      <td class="align-center">
+        <p>
+         <code class="filter">{% raw %}{{ "style.css" | asset_url }}{% endraw %}</code>
+        </p>
+        <p>
+          <code class="output">/my-baseurl/assets/style.css</code>
+        </p>
+      </td>
+    </tr>
+    <tr>
+      <td class="align-center">
+        <p>
+         <code class="filter">{% raw %}{{ "css/style.css" | asset_url }}{% endraw %}</code>
+        </p>
+        <p>
+          <code class="output">/my-baseurl/assets/css/style.css</code>
+        </p>
+      </td>
+    </tr>    
+    <tr>
       <td>
         <p class="name"><strong>Date to XML Schema</strong></p>
         <p>Convert a Date into XML Schema (ISO 8601) format.</p>
--- a/test/test_filters.rb
+++ b/test/test_filters.rb
@@ -419,6 +419,47 @@ class TestFilters < JekyllUnitTest
       end
     end
 
+    context "asset_url filter" do
+      should "produce a relative URL for a given asset file" do
+        asset_file = "test.css"
+        assert_equal "/base/assets/#{asset_file}", @filter.asset_url(asset_file)
+      end
+
+      should "produce a relative URL for a given asset file within a subdirectory" do
+        asset_file = "css/test.css"
+        assert_equal "/base/assets/#{asset_file}", @filter.asset_url(asset_file)
+      end
+
+      should "ensure the leading slash between baseurl, assets_dir and input" do
+        input = "test.css"
+        assert_equal "/base/assets/#{input}", @filter.asset_url(input)
+      end
+
+      should "ensure the leading slash for the baseurl" do
+        asset_file = "test.css"
+        filter = make_filter_mock({ "baseurl" => "base" })
+        assert_equal "/base/assets/#{asset_file}", filter.asset_url(asset_file)
+      end
+
+      should "be ok with a nil 'baseurl'" do
+        asset_file = "test.css"
+        filter = make_filter_mock({
+          "url"     => "http://example.com",
+          "baseurl" => nil
+        })
+        assert_equal "/assets/#{asset_file}", filter.asset_url(asset_file)
+      end
+
+      should "not prepend a forward slash if input is empty" do
+        input = ""
+        filter = make_filter_mock({
+          "url"     => "http://example.com",
+          "baseurl" => "/base"
+        })
+        assert_equal "/base/assets", filter.asset_url(input)
+      end
+    end
+
     context "jsonify filter" do
       should "convert hash to json" do
         assert_equal "{\"age\":18}", @filter.jsonify({ :age => 18 })


benbalter:
Needs tests and docs as you indicated, but looking good!

/cc https://github.com/jekyll/jekyll/issues/5407

lib/jekyll/filters/url_filters.rb
```diff
@@ -29,6 +29,23 @@ def relative_url(input)
...
+      # Produces the path to the given asset relative to the assets directory at the
+      # domain root based on site.baseurl
+      #
+      # input - the asset file relative to the 'assets' directory. The string may also
+      #         include any subfolder(s) the asset may be a part of.
+      #           e.g. "main.css" or "css/main.css"
+      #
+      # Returns a path relative to the assets directory as a String.
+      def asset_url(input)
+        return if input.nil?
```
envygeeks: If I'm honest, I think filters should just throw instead of returning unexpected results if the user makes a mistake.  AKA if I use the wrong variable name, or something, this will give an expected result when it should have thrown.

lib/jekyll/filters/url_filters.rb
```diff
@@ -29,6 +29,23 @@ def relative_url(input)
...
+        site = @context.registers[:site]
+        return ensure_leading_slash(input.to_s) if site.config["baseurl"].nil?
```
envygeeks: You are so close to the edge of the method it should just be:

```
if site.config["baseurl"].nil?
  # This
else
  # That
end
```

Explicit returns have a cost in Ruby that they shouldn't (but they do) and the `if` + `return` will result in a higher cost than an `if/else` in this case... especially when at the edge of the method.

parkr:
What is the motivation for this?

envygeeks:
@parkr I think this is a result of the discussion in the ticket about theme assets.

benbalter:
> What is the motivation for this?

See #5407.

lib/jekyll/filters/url_filters.rb
ashmaroli: Yes, you're right. It should tell something's not right..

lib/jekyll/filters/url_filters.rb
ashmaroli: I'm not able to see the "cost" that results because of this.. how high is it?

parkr: Why not lean on "relative_url" here to do the dirty work of the baseurl stuff?

ashmaroli: > Why not lean on "relative_url" here to do the dirty work of the baseurl stuff?

Do you mean something similar to:
```
asset_url = "/assets" + relative_url("main.css")
```

lib/jekyll/filters/url_filters.rb
ashmaroli: Keeping the same behavior other filters follow..

pathawks:
I'm still :-1: on this. I understand when `absolute_url` or `relative_url` might produce different results based on config, but it seems that `asset_url` has nothing to do with the config, and is just exactly the same as `{{ url | prepend: '/assets' | relative_url }}`.

This might make for syntactic sugar, but I worry it overcomplicates things. It seems to me to imply that there is some sort of magic going on behind the scenes (though there isn't), and magic is difficult to understand.

It seems like this filter would be most used by theme creators, but adds complexity for all users. All for something that can be accomplished with a simple `prepend`? Is it really worth it?

ashmaroli:
screenshot of updated docs

af

lib/jekyll/filters/url_filters.rb
```diff
@@ -29,6 +29,23 @@ def relative_url(input)
...
+        Addressable::URI.parse(
+          ensure_leading_slash(File.join(site.config["baseurl"], "assets", input.to_s))
```
pathawks: `File.join` = Bad
Please use `Addressable::URI.parse` instead.

site/_docs/templates.md
```diff
@@ -50,6 +50,30 @@ common tasks easier.
       </td>
     </tr>
     <tr>
+      <td rowspan="2">
+        <p class="name"><strong>Asset URL</strong></p>
+        <p>Prepend and <code>baseurl</code> value and string <code>/assets/</code> to the input. This will produce a relative URL to just the assets directory. Subdirectories, if present, may be passed along with the input as well.</p>
```
pathawks: I think there is extra “and” in that first sentence.

ashmaroli:
> Is it really worth it?

Tough question there.. @pathawks  
Why I really wanted it was for those who may get confused with the placement of `prepend` filter in the liquid example you quoted.. 
{{ url | prepend: '/assets' | relative_url }} is not the same as {{ url | relative_url | prepend '/assets' }}.

lib/jekyll/filters/url_filters.rb
ashmaroli: `Addressable::URI.parse` is being used at the line above it..

site/_docs/templates.md
ashmaroli: Ah yes.. 
fixup coming up..

pathawks:
How often will this be used with a variable? Seems like it is most likely to be used with a static string:

```text
{{ "style.css" | asset_url }}
{{ "/assets/style.css" | relative_url }}
```

As somebody who is not a beginner, but can remember what it was like to be a beginner, the second seems to make it more clear _where_ the file is located.

ashmaroli:
> As somebody who is not a beginner, but can remember what it was like to be a beginner, the second seems to make it more clear where the file is located.

when there are subdirectories,
{{ "/assets/css/style.css" | relative_url }} or {{ "css/style.css" | prepend: "/assets" | relative_url }} can get a little irksome..

lib/jekyll/filters/url_filters.rb
ashmaroli: Is File.join worse than concat strings with  `+` ?

pathawks:
But _nobody_ would write `{{ "css/style.css" | prepend: "/assets" | relative_url }}`

`{{ "/assets/css/style.css" | relative_url }}` tells me _exactly_ where to find the file.

The only time this would not be the case is when the url is a variable rather than a static string, but I can't think of any compelling examples of when it would be done that way, and doing so would just introduce more magic.

I’ve said my piece. If I just can't understand how wonderful this is thrn don't let me stand in the way.

ashmaroli:
I didnt mean to be difficult to reason with..

lib/jekyll/filters/url_filters.rb
envygeeks: The cost matters little, please change it.

ashmaroli: done..

lib/jekyll/filters/url_filters.rb
envygeeks: @pathawks can you elaborate while `File#join` is bad in this context?
@ashmaroli Depends. For systems work yes.  For URL's no, but that said, I think @pathawks is trying to say that Addressable has a method to deal with that kind of stuff, I know it has `#convert_path` but maybe there is something else I don't remember.

pathawks: I have seen all problems using `File.join` for URLs when working with non-ASCII characters.

jekyll/jekyll-sitemap#106

I feel like there have been more issues, but cannot find them at the moment.

pathawks: As long as `input` **does _not_** have a leading slash:

```ruby
relative_url(
  Addressable::URI.join("assets/", input.to_s).normalize
)
```

pathawks: Here it is:

```ruby
relative_url(
  "assets" + ensure_leading_slash(input.to_s)
)
```

ashmaroli: nicee.. clean and simple.. :+1:

envygeeks:
Does this need a review again? Make sure to ping  when you think it's ready again because I think I'm the one blocking this merge.

ashmaroli:
@envygeeks , yes Jordon, this requires a `re-review` from _all of you_ actually.. the filter definition has changed entirely.
Its only with @pathawks most recent input, that I clearly understood what @parkr meant when he said
> Why not lean on "relative_url" here to do the dirty work of the baseurl stuff?

envygeeks:
Done, it's okay from what I can tell, thanks!

ashmaroli:
@parkr, one question: why does all the `input` arguments to `ensure_leading_slash` have to be explicitly converted to a string when interpolation with `#{input}` does that natively?

DirtyF:
IMHO if we wanna go down this road of providing sugax syntax for default assets path, we may as well learn from `jekyll-assets` and provide default liquid tags like `style` or `script`. This feels kind of a half baked feature to me, but I don't mind if this gets merged, it's a first step to adopt convention for storing assets in Jekyll, and it's a good thing.

ashmaroli:
Tags were never under consideration.
At the initial stages of creating this pull, I had envisioned:

def `asset_path` : `relative_url of asset_file`
def `asset_url` : `absolute_url of asset_file`

if **image_file**, prepend: `/assets/img/`
if **stylesheet**, prepend: `/assets/css/`
if **script_file**, prepend: `/assets/js/`
if **font_file**, prepend: `/assets/fonts/`
otherwise, prepend: `/assets/`
and enforce corresponding sub-directory usage. 

But over time, it was decided to not enforce sub-directories, by default 
That boiled down to just the basic two filters. I decided to settle for the relative_url derivative and classify it as an URLFilter (hence `asset_url`)
The relevance of this filter in its present state, is still under dispute (in context of simply using the `relative_url` filter on input `"assets/main.css"`).

PR #5452, opened 2016-10-06
Update Wikipedia YAML list link

danayel:
List no longer exists on that page as an id.  Basic_components holds the lists now, and at the top conveniently :v:

I also added a second link under the tags heading.


parkr:
@jekyllbot: merge +site

PR #5456, opened 2016-10-06
Revert Commands::Serve#server_address signature change.

parkr:
This reverts a tiny part of #5431, which introduced a backwards-incompatible change that JekyllAdmin was relying on. It's never fun to break things and jekyll-admin must have a stunning reputation for not breaking. 😄 

This is also necessary because `launch_browser` also used `server_address` and wasn't updated. So we have to handle that anyway.

/cc @jekyll/stability

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -104,12 +104,7 @@ module Jekyll
         def start_up_webrick(opts, destination)
           server = WEBrick::HTTPServer.new(webrick_opts(opts)).tap { |o| o.unmount("") }
           server.mount(opts["baseurl"], Servlet, destination, file_handler_opts)
-          Jekyll.logger.info "Server address:", server_address(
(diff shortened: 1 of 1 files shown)


benbalter:
👍  this doesn't fix the breaking change for Jekyll Admin, but we were monkey patching the `process` method, so I'm fine with pushing the complexity into JekyllAdmin here, doing what's best for Jekyll, and making the necessary change upstream.

parkr:
The breaking change was the creation of the `start_webrick` method, which I think we should keep. Monkey-patching is not really something we should feel we need to support, anyway, and Ben is fine with pushing it upstream. I'd love to get a jekyll-admin fix out before 3.3.

ashmaroli:
FWIW, `v3.3.0` doesnt break `jekyll-admin` with this change incorporated..

parkr:
@jekyllbot: merge +bug

PR #5457, opened 2016-10-06
`jekyll new-theme` should specify Jekyll as a runtime dependency for the theme

ashmaroli:
Themes created with Jekyll 3.3.x will run with only Jekyll 3.3.x

--
/cc @jekyll/ecosystem


parkr:
Last time I tried this Bundler gave me a "Circular dependency" error, but let me know if you don't get the same. 😄

ashmaroli:
> Last time I tried this Bundler gave me a "Circular dependency" error,

For this to happen, Jekyll has to _depend_ on a theme too.. but since it doesnt, I dont think bundler will care..

parkr:
@jekyllbot: merge +bug

Issue #5434, opened 2016-10-01
The Future of Theme Gems [Discussion]


ashmaroli:
The future of theme gems as I see it..:
**Requires Jekyll 4.0**
## Self-Contained Modules

Theme gems will be self-contained _insta-jekyll_-modules. All they need would be **content**. This allows users to freely switch between themes without breaking their site or losing their customizations. For this to happen a few things will be prescribed and expected to be followed. A theme-gem ecosystem would be:
- SCAFFOLD DIRECTORIES [ `_layouts` | `_includes` | `_sass` | `_data` | `assets` ]
- every theme gem shall contain a **configuration file** named the same as the gem-name e.g. `minima.yml` and it will reside inside `_data/`
- core styles and settings will be defined in `minima.yml` and referenced in templates with liquid using the `theme` namespace. e.g. `{{ theme.myvariable }}`
- the layout names will be standard for all theme-gems:
  - `home` for index
  - `post` for posts
  - `page` for pages
  - `default` as the parent layout for above layouts

When overriding defaults, the user creates the required file / copies from the gem_dir. into a directory named the same as the theme gem in use.  e.g. `minima/_data/minima.yml` | `minima/_layouts/` | `minima/_includes/` etc 

Jekyll 4.0 shall look for these directories first in the `site.root` and then in the gem.

Following this _convention_ ensures easy switching between theme gems at the same time preserving their customizations.

vwochnik:
+1

oncleben31:
`_data` theme folder could host locale strings files for themes with i18n features.

ashmaroli:
@parkr, what do you think about modifying `lib/~/data_reader.rb` to include `_data/` inside theme gems..?

mmistakes:
I'm with @oncleben31 and @ashmaroli. 

Bundling in `_data` files would be very useful. One of my theme's currently uses a data file for all of the [localized text](https://github.com/mmistakes/minimal-mistakes/blob/master/_data/ui-text.yml) found in it's UI elements. Asking users to manually copy over that file when using a theme gem is kind of a bummer.

jaybe-jekyll:
This is a neat topic of discussion. In recent related threads, themes have been enhanced to include seeded assets/, etc. for the purposes of providing a full, complete, and out-of-box-functional theme/site. Extending inclusion to other, additional areas such as pre-seeded/configured _data/ is exciting.

ashmaroli:
[...]

Issue #5350, opened 2016-09-12
_config.yml for themes


vwochnik:
As @envygeeks suggested in #5341 I am opening an issue to discuss an approach to themes that would satisfy most requirements that have been stated after the basic theme support was released:

Why not have a `_config.yml` for each theme containing include and exclude where every file other then _*-prefixed directories is being read into a StaticFile with the ability for a site including this theme to provide their own version, a.k.a. overriding?

Also, config files can then include the plugins the theme depends on making Jekyll automatically load those to support the build process of the theme.

So basically, a theme will be 'sucked' into the build process of a site and whenever the site provides a file with the same relative path as the one provided by the theme, the site's one is to be used.

It would at least solve the problem with theme assets, with or without `jekyll-assets`.

/cc @jekyll-core @envygeeks

benbalter:
> Why not have a _config.yml for each theme containing include and exclude where every file other then _*-prefixed directories is being read into a StaticFile with the ability for a site including this theme to provide their own version, a.k.a. overriding?

This would be in the theme itself? if so, that should be the role of the Gemspec's file matcher.

Why not just include everything in the theme's `assets` folder, if the same file does not already exist in the site (e.g., allowing overrides like we do for layouts and includes). That way, you could have `assets/logo.img` or `assets/style.scss`, both of which would be copied, without the need to specify additional configuration.

vwochnik:
It seems too generic to me. What if the theme wants to have a `_data` file or something else? Maybe a robots.txt supplied? Or a `favicon.ico`? Having a configurable theme which can itself use plugins via the config `gems`, as well as `include` and `exclude` files makes it possible to completely separate content from theme which wouldn't be possible with just a single `assets` folder.

benbalter:
>  Having a configurable theme which can itself use plugins via the config gems

benbalter:
>  Having a configurable theme which can itself use plugins via the config gems

Ahh, so you're suggesting that the theme's config be used as a default and merged into the user's config?

> What if the theme wants to have a _data file or something else? 

Can you explain a bit more how the theme would use the data file?

> Maybe a robots.txt supplied? 

Since that is logic, not presentation, and isn't theme specific, wouldn't that be better handled by a plugin?

> Or a favicon.ico?

I don't think it's appropriate for the theme to package a favicon. What would that favicon be that would be appropriate for any site?

vwochnik:
[...]

PR #5470, opened 2016-10-09
Allow reading data files bundled within theme-gems

ashmaroli:
If a theme gem contains a `_data` folder at its root, its files will be read and added to the site's internal data hash. The theme-gem's data hash will be overridden by data at source if the same 'key-value' pair exists there.
## 

TODO:
- [x] Add tests
- [x] Add cucumber feature
- [x] Add Documentation
## 

Ref: #5434 
/cc @jekyll/ecosystem, @envygeeks, @pathawks


pathawks:
For what purpose? Seems like a violation of separation of design from content.

I feel like there must be a better way to do i18n.

ashmaroli:
> For what purpose?

Its not just for i18n _per se_.. but having this can lead to creating a more functional theme-gem.. I dont have a solid use case for this as of now.. will be fleshed out it in the referenced ticket (no. 5434)

--
> Seems like a violation of separation of design from content.

No.. not all.. 
though my original intention was to use it for structural elements like define a navigation system using `navigation.yml` , etc.. I now see the possibilities for built-in i18n as well..

--
P.S.  Will be away for a few days. Feedback welcome.

benbalter:
@ashmaroli thanks for opening up this PR. I really like the vision for themes that you laid out in #5434. 

> I dont have a solid use case for this as of now.

Before we move forward, I'd feel a lot more comfortable if we understood *why* we were building the feature, rather than just building it for a sense of symetry with support for other Jekyll-specific folders, or because it's easy to do so. As was mentioned over in https://github.com/jekyll/jekyll/issues/5350, I share @pathawks concerns that themes should only know about presentation, not content (to ensure theme fungability).

We really, really need to figure out a solid I18N implemntation, but I don't know that `_data` folder support for themes is it.

spudowiar:
> No.. not all..
> though my original intention was to use it for structural elements like define a navigation system using navigation.yml , etc.. I now see the possibilities for built-in i18n as well..

Seems like a good use case... for when you are using `<site root>/_data/navigation.yml` (which already works), **not** `<theme root>/_data/navigation.yml` :smile:

ashmaroli:
> I share Pat Hawks' concerns that themes should only know about presentation, not content (to ensure theme fungibility).

I care about theme fungibility too and this feature does not mess with content in the manner you guys fear. Content, _to me,_ comes from markdown files at source, **supported** by files in an _optional_ `_data` folder.

oncleben31:
> I feel like there must be a better way to do i18n.

> We really, really need to figure out a solid I18N implemntation, 

Have you some guideline as I'm trying to add basic i18n features to minima template (in https://github.com/jekyll/minima/pull/60) ?
 Currently I feel the best option is to use locale string file in  `_data` folder. But I'm new to Jekyll so if you have better ideas please share it.

dummied:
Did #5491 (and ashmaroli/jekyll-data) render this moot?

ashmaroli:
@dummied This is still active, provided there's a demand for this to be included in the core

Swiftrien:
I have a theme in which I provide default values for configuration and theme-translation via the _data directory.
If not in the _data directory, how else could I provide default values?

ashmaroli:
@Swiftrien you may also give the [jekyll-data gem](https://github.com/ashmaroli/jekyll-data) a try in the meantime.. 
It provides the same functionality as this pull, and a li'l bit more.. :smiley:

Swiftrien:
Thanks ashmaroli

However that would mean that every user of my theme also has to install it. Thus this is not the answer to this issue..

ashmaroli:
> However that would mean that every user of my theme also has to install it.

@Swiftrien I'm not trying to promote a plugin I authored, but if you add `jekyll-data` as a `runtime_dependency` to your theme's gemspec, it will be automatically installed with your theme-gem.
Its entirely your call. I won't mind if you decide otherwise.

But I'd definitely like it if you gave it a try, and provide a feedback.
I could then update this pull based on that.

Either way, Happy Jekylling :smiley:

Swiftrien:
I have nothing against promotion :-)
[classic-jekyll-theme](https://rubygems.org/gems/classic-jekyll-theme) :-)

I just added the jekyll-data (0.3.0) gem to a website of mine, removed the _data directory, rebuild the site, and everything is as it should be. Thus jekyll-data seems to work fine. Note that I did not use the "theme configuration" only the "regular data" files.

My bigger issue with external contributions is that is takes away my control. And I have had several experiences where external dependencies costs me huge amounts of time. So in general I try to avoid them where possible. As it is, I think that the jekyll-data gem adds enough value that I will probably include it in the next release. (Mainly because I hope it will be part of Jekyll 3.4 anyhow... ;-))

yatil:
I support this PR for the reasons outlined in #6684. 

The actual practical example for us at W3C/WAI is that we have several independent resources that are then sub-moduled into the big site (W3C/wai-website). Every resource can access information in the data files (minor translations but mostly at this stage WCAG 2.0 Success Criteria and techniques). Moving those source files to the theme gives us the possibility to make the changes once and re-use the data in every resource.

mattr-:
This looks ok to me at first glance. I'd be happy to take a deeper look once this up to date with master and the conflicts are resolved.

parkr:
The purpose of themes and gem-based themes especially is to make the look-and-feel of websites portable. Can you help me understand how data files, which I usually see used for content, fall under the category of look-and-feel?

DirtyF:
@parkr Data files are often use to store config options, that would allow theme authors to embed those options with their thems instead of asking users to copy a sample `_config.yml`. Right now gem-based themes can only ship layouts and assets and often that's not all it takes to have a working theme out of the box, e.g. @mmistakes: https://github.com/mmistakes/minimal-mistakes/tree/master/_data

Balancingrock:
In the classic-jekyll-theme I use the data files for:

- Configuration (as DirtyF commented)
- Translations (User can provide their own translations for titles and headers elements)
- Data for site elements.

The last one is indeed content data, but the theme will process this data to build the user's site. That imo is definitely a theme-job, but needs user generated data.

yatil:
- Also example data: Here is an example navigation that you can copy to your data directory and edit. 
- Text descriptions for page elements instead of having them in the template or in the _config.yml

ellemenno:
@parkr my look and feel use case is icons: I'm working on a theme that embeds icons as svg, rather than using an icon font.

I want to provide the theme icon shape paths in the _data folder, and provide an _include macro for the theme and the user to embed icons wherever needed. By retrieving icon data on demand at build time, I can avoid forcing the inclusion of all icons in the generated page html, which saves significant kb.

Ryuno-Ki:
@ellemenno As @mattr- wrote in https://github.com/jekyll/jekyll/pull/5470#issuecomment-358536052

Failing builds + conflicts in the first place :-)

ashmaroli:
> Scheduling this for inclusion in 4.1

That's awesome news @mattr-. Thank you.

lib/jekyll/reader.rb
```diff
@@ -183,5 +183,21 @@ def read_included_file(entry_path)
         site.static_files.concat(StaticFileReader.new(site, dir).read(file))
       end
     end
+
+    # Read data files within current site.
+    #
+    # If the site uses a theme and has a `_data` directory at its root, read the files within
```
mattr-, line 189:
```suggestion
    # If the site uses a theme and the theme has a `_data` directory at its root, read the files within
```

DirtyF:
Waiting for the latest tests to pass to merge...

ashmaroli:
In light of recently acquired knowledge that unlike layouts or includes which are overridden entirely by the namesake at source, data read-in may at times persist partially warranting additional user-intervention to override the persisting  data, I'm withdrawing this submission.

Those interested in this functionality may opt to use the [`jekyll-data`](https://rubygems.org/gems/jekyll-data) plugin instead.

ashmaroli:
@yatil While I understand the sentiment, please consider the following scenario:
A theme has bundled the following data file named `social_links.yml`:
```yaml
twitter: jekyllrb
github:  jekyll
```
And the theme uses the following HTML *include* to render social media icons:
```html
{%- assign social = site.data.social_links -%}
<ul class="social-media-list">
  {% if social.github %}<li><a href="https://github.com/{{ social.github }}" title="{{ social.github  }}"><svg class="svg-icon"><use xlink:href="{{ '/assets/imgs/social-icons.svg#github' | relative_url }}"></use></svg></a></li>{% endif %}
  {% if social.twitter %}<li><a href="https://twitter.com/{{ social.twitter }}" title="{{ social.twitter  }}"><svg class="svg-icon"><use xlink:href="{{ '/assets/imgs/social-icons.svg#twitter' | relative_url }}"></use></svg></a></li>{%- endif -%}
{% if social.linkedin %}<li><a  href="https://www.linkedin.com/in/{{ social.linkedin  }}" title="{{ social.linkedin }}"><svg class="svg-icon"><use xlink:href="{{ '/assets/imgs/social-icons.svg#linkedin' | relative_url }}"></use></svg></a></li>{% endif %}
  <!-- ... so on ... -->
</ul>
```
Now, suppose one of the theme's user doesn't have a GitHub account or a Twitter handle.. they think that all they need to do is override the theme's data file with one at their source:
```yaml
# source/_data/social_links.yml

linkedin: johndoe
```

*However, when they build their site, all three social icons are rendered by default.*
The user has to then explicitly set `github` and `twitter` keys to `null` to not render them, A couple of keys may not be a hassle. But imagine if the theme supports over 50 social accounts.., the user would have to manually disable over 4 dozens of keys...........

yatil:
@ashmaroli 

> However, when they build their site, all three social icons are rendered by default.

Because both, the theme's `social_links.yml` and the site’s `social_links.yml` are interpreted. But, theoretically, wouldn’t it be possible to say “if the site has a social_links.yml, ignore the theme’s social_links.yml”?

I personally can see both behaviors to be beneficial, but one would then need to be able to configure it and that might create problems.

Thanks for exploring the idea, it is much appreciated.

ashmaroli:
> would then need to be able to configure it and that might create problems

Jekyll 4.0 introduced the ability to import *configuration* from a theme-gem and merge it with the one from the source. It led to problems similar to what I described above.
And now Jekyll 4.1 will make amends by providing users with an option to disable loading the configuration from the theme.

> theoretically, wouldn’t it be possible to say “if the site has a social_links.yml, ignore the theme’s social_links.yml”?

Yes, that is possible. I had considered that but there's catch though.
`_data/foo.yml` containing top-level key `bar:` and `_data/foo/bar.yml` are equivalent and it is not  possible to know about *top-level keys* without parsing the data file first.

PR #5483, opened 2016-10-13
Documentation fix: jekyll.environment to site.environment

jmcglone:
On the latest version of Jekyll, using `jekyll.environment` in a liquid tag did not work as described in the documentation. However, `site.environment` did work, and was correctly mapped to my environment configuration contained in my jekyll config file.  I'm guessing `jekyll.environment` is just left over from a previous version of Jekyll and therefore needs to be updated in the docs?

I did not search through the rest of the documentation to see where else `jekyll.environment` is referenced.

/cc @jekyll/documentation

### DIFF
--- a/site/_docs/configuration.md
+++ b/site/_docs/configuration.md
@@ -422,7 +422,7 @@ For example, suppose you set this conditional statement in your code:
 
 ```liquid
 {% raw %}
-{% if jekyll.environment == "production" %}
+{% if site.environment == "production" %}
    {% include disqus.html %}
 {% endif %}
 {% endraw %}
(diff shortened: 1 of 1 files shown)


DirtyF:
@jmcglone I don't reproduce this behaviour with Jekyll 3.3 on Mac OS X, can you provide more info?

pathawks:
I think this is definitely supposed to be `jekyll.environment`

If that is not working, we need to fix it.

jmcglone:
@DirtyF @pathawks Well, that is good to know!

First, I apologize, I was wrong about my Jekyll version for this project -- it's using 3.1.3.

I also have a very customized Jekyll site that is integrated into a Rails application and built during a Capistrano deploy [using a rake task](https://github.com/mlibrary/heliotrope/blob/master/lib/tasks/jekyll.rake#L37), which could also be the cause.

Other things:

OS: Linux 3.16.0-4-amd64 #1 SMP Debian 3.16.36-1+deb8u1 (2016-09-03) x86_64 GNU/Linux
Ruby: 2.3.0

pathawks:
[The relevant test](https://github.com/jekyll/jekyll/blob/v3.3.0/features/rendering.feature#L16)

I am going to close this PR as we do not want to change the documentation.

If you can reproduce this issue with the latest version of Jekyll, please do open an issue and we will investigate further :+1:

jmcglone:
Thanks!

Issue #5467, opened 2016-10-08
Extra log entries in Jekyll 3.3.0


ashmaroli:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I updated to the latest Jekyll
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
## 
- [x] I am on (or have tested on) **_Windows**_ 7+
## 
- [x] It was another bug.
## 
## My Reproduction Steps

```
jekyll new my_blog
bundle exec jekyll serve
```
## The Output I Got:

```
D:\Projects\my_blog>bundle exec jekyll serve
Configuration file: D:/Projects/my_blog/_config.yml
Configuration file: D:/Projects/my_blog/_config.yml
            Source: D:/Projects/my_blog
       Destination: D:/Projects/my_blog/_site
 Incremental build: disabled. Enable with --incremental
      Generating...
                    done in 0.91 seconds.
                    --watch arg is unsupported on Windows. 
                    If you are on Windows Bash, please see: 
                    https://github.com/Microsoft/BashOnWindows/issues/216
Configuration file: D:/Projects/my_blog/_config.yml
    Server address: http://127.0.0.1:4000/
  Server running... press ctrl-c to stop.
```

3 instances of `Configuration file: D:/Projects/my_blog/_config.yml`
in total..

and if I run with `--verbose`:

```
  Logging at level: debug
Configuration file: D:/Projects/my_blog/_config.yml
  Logging at level: debug
  Logging at level: debug
Configuration file: D:/Projects/my_blog/_config.yml
         Requiring: jekyll-feed
         Requiring: kramdown
            Source: D:/Projects/my_blog
       Destination: D:/Projects/my_blog/_site
 Incremental build: disabled. Enable with --incremental
      Generating...
```

2 instances of `Configuration file: D:/Projects/my_blog/_config.yml`
3 instances of `Logging at level: debug`
in this block.
## 

/cc @jekyll/ecosystem

Crunch09:
This is caused by the two `configuration_from_options` calls within `Jekyll::Commands::Serve` [here](https://github.com/jekyll/jekyll/blob/53d472b7e297f2e5267aebbd3e160cd9e88c82bd/lib/jekyll/commands/serve.rb#L47) and [here](https://github.com/jekyll/jekyll/blob/53d472b7e297f2e5267aebbd3e160cd9e88c82bd/lib/jekyll/commands/serve.rb#L150).
[...]

PR #5487, opened 2016-10-16
create configuration from options only once in the boot process

Crunch09:
Hi,

before, `Jekyll.configuration` was called 3 times during  the `serve` command as described in #5467 . With this patch it is called only once:

screen shot 2016-10-16 at 22 48 22

I also [made sure](https://github.com/jekyll/jekyll/blob/a6580f978e6f826c9c4a369c183a47ce5c222d0a/lib/jekyll/commands/serve.rb#L53-L57) that `jekyll-admin` still works with this, to avoid the disaster from last time 😊 .

@jekyll/build

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -32,19 +32,35 @@ module Jekyll
             cmd.action do |_, opts|
               opts["serving"] = true
               opts["watch"  ] = true unless opts.key?("watch")
-              config = opts["config"]
-              opts["url"] = default_url(opts) if Jekyll.env == "development"
-              Build.process(opts)
-              opts["config"] = config
-              Serve.process(opts)
+              read_config = configuration_from_options(opts)
+              if Jekyll.env == "development"
+                opts["url"] = read_config["url"] = default_url(opts, read_config)
+              end
+
+              build_and_serve(opts, read_config)
             end
           end
         end
 
         #
 
-        def process(opts)
-          opts = configuration_from_options(opts)
+        def build_and_serve(opts, read_config)
+          if Build.method(:process).arity == -2
+            Build.process(opts, read_config)
+          else
+            Build.process(opts)
+          end
+          if Serve.method(:process).arity == -2
+            Serve.process(opts, read_config)
+          else
+            Serve.process(opts)
+          end
+        end
+
+        #
+
+        def process(opts, read_config = nil)
+          opts = read_config.nil? ? configuration_from_options(opts) : read_config
           destination = opts["destination"]
           setup(destination)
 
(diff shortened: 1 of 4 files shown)


ashmaroli:
whats the status on this?

Crunch09:
Hey @ashmaroli ,

because i edited my first post, i think jekyllbot didn't ping the correct team. So pinging again now 😆 

/cc @jekyll/build

parkr:
I kind of wish we could just make `configuration_from_options` idempotent – to have some guard that says if it's already a setup `Jekyll::Configuration` instance, then just return it.

```ruby
def configuration_from_options(opts)
  return opts if opts.is_a?(Jekyll::Configuration) && opts.complete?

  # do more stuff
end
```

Then we'd pass along the completed `Jekyll::Configuration` instance.

Also, when we create a site, the site is added to `Jekyll.sites`. If a site already exists, we should just use that one so we don't recreate them. That already has the config on it!

Crunch09:
@parkr Thanks for your feedback, sounds like a good approach to me. I'll try to come up with a solution like this.

Crunch09:
@parkr Hi, I updated the code as you suggested but left out the `complete?` check because i wasn't sure when a Configuration should be considered complete and also the test suite already passed (at least on minitest 5.9 🙄 ).

> Also, when we create a site, the site is added to Jekyll.sites. If a site already exists, we should just use that one so we don't recreate them. That already has the config on it!

I thought it would be best to create a separate PR to tackle this issue after this PR got merged?

Crunch09:
@parkr Could you take another look at this one if you have a few minutes, please?

Crunch09:
@parkr Would be great if you could take another look at this one if you have time. Would love to see this getting merged. Let me know if there is anything missing.

DirtyF:
@Crunch09 FWIW I tested your PR and it works fine by me.

Crunch09:
@DirtyF Great, happy to hear that :)

lib/jekyll/commands/serve.rb
```diff
@@ -32,11 +32,12 @@ def init_with_program(prog)
...
-              opts["url"] = default_url(opts) if Jekyll.env == "development"
-              Build.process(opts)
-              opts["config"] = config
-              Serve.process(opts)
+
+              config = configuration_from_options(opts)
+              if Jekyll.env == "development"
+                config["url"] = default_url(config)
+              end
+              build_and_serve(config)
```
parkr: You could get really fancy and do

```ruby
[Build, Serve].each { |klass| klass.process(config) }

pathawks:
LGTM :+1:

parkr:
@jekyllbot: merge +minor

Crunch09:
Thanks!

parkr:
Good work @Crunch09!

PR #5489, opened 2016-10-17
fix rubocop errors on testing with Rubocop 0.44

ashmaroli:
Rubocop 0.44 introduced a new cop called `Metrics/BlockLength` that excludes spec files and rakefiles by default. This PR excludes our minitest files as well.
Also excludes 2 other files that got caught in [this job](https://travis-ci.org/jekyll/jekyll/jobs/168120013).
  - have the new `Metrics/BlockLength` cop ignore test files and
    `jekyll/configuration.rb`
  - have `AllCops` ignore Jekyll Executable which is not going to be
    altered in the near future.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,6 +6,7 @@ AllCops:
   Exclude:
     - lib/jekyll/renderer.rb
     - bin/**/*
+    - exe/**/*
     - benchmark/**/*
     - script/**/*
     - vendor/**/*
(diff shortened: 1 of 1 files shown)


parkr:
@jekyllbot: merge +dev

pathawks:
I wonder if @jekyllbot could notice Rubocop errors and maybe try to automatically fix them.

ashmaroli:
> I wonder if @jekyllbot could notice Rubocop errors and maybe try to automatically fix them.

or.. we could have a conservative version-lock for Rubocop gem in our Gemfile..

PR #5491, opened 2016-10-18
Add jekyll-data to the list of plugins

ashmaroli:
While the basic functionality is the same as that I proposed in #5470, the plugin further adds a `theme` (equivalent to `site.data[theme-name]`) namespace to the `UnifiedPayloadDrop` class and outputs the data read into an easily readable format in `jekyll build --verbose`

--
/cc @jekyll/documentation


parkr:
@jekyllbot: merge +site

PR #5494, opened 2016-10-18
Collapse `gsub`

pathawks:
I found a couple places where I could combine two `gsub`s into one :+1:


pathawks:
@jekyllbot: merge +minor

PR #5495, opened 2016-10-19
Do not swallow all exceptions on render

### LINKED ISSUE #5104: Site.render() swallows all Errno::ENOENT exceptions, leading to silent build failures

---
- [x] I am on (or have tested on) **_macOS**_ 10+
- [x] I am on (or have tested on) **_Other**_ GNU/Linux

---
- [x] I was trying to build.

## My Reproduction Steps

Create a new site using `jekyll new my-test-site`. Then create a custom Liquid tag which only throws missing file errors when run:

``` ruby
# in _plugins/mytag.rb
require 'liquid'

module Jekyll
  class MyCustomTag < Liquid::Tag

    def initialize(tag_name, argument, tokens)
      super
    end

    def render(context)
      puts "Rendering my_tag"
      File.open("/some/file/that/does/not/exist").read.encode("UTF-8")
    end
  end
end

Liquid::Template.register_tag('my_custom_tag', Jekyll::MyCustomTag)
```

Next, update a post to use this tag:

``` md
{% my_custom_tag some_argument %}
```

Finally, render the site with `jekyll build`.
## The Output I Wanted

I wanted Jekyll to exit with a non-zero exit code because the uncaught exception in my tag prevents the site from rendering properly.

What _actually_ happens is that the error is logged and Jekyll reports that the site was successfully built, but the actual `_site/` output is missing several posts, including the page with the bad tag _and_ other pages:

```
Configuration file: /private/tmp/error-handling-bug/_config.yml
            Source: /private/tmp/error-handling-bug
       Destination: /private/tmp/error-handling-bug/_site
      Generating...
Rendering my_tag
  Liquid Exception: No such file or directory - /some/file/that/does/not/exist in /private/tmp/error-handling-bug/_posts/2016-07-14-welcome-to-jekyll.markdown
                    done in 0.031 seconds.
```
## Cause of bug

When Ruby's `File.open()` fails due to a missing file, it throws `Errno::ENOENT`. This exception bubbles up to Jekyll's `Site.render()` method, [**where it is silently ignored**](https://github.com/jekyll/jekyll/blob/8a3ea59141d721b09a762d7ae26e677b173bfcfa/lib/jekyll/site.rb#L192):

```
    rescue Errno::ENOENT
      # ignore missing layout dir
    end
```

This is sloppy error-handling; at a minimum, the `rescue` statement should log a warning, but ideally the "ignore missing layout dir" logic would be scoped more narrowly so that it doesn't catch and ignore `Errno::ENOENT` exceptions thrown by other parts of the page rendering code.

pathawks:
Fixes #5104

### DIFF
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -191,11 +191,7 @@ module Jekyll
       render_pages(payload)
 
       Jekyll::Hooks.trigger :site, :post_render, self, payload
-    # rubocop: disable HandleExceptions
-    rescue Errno::ENOENT
-      # ignore missing layout dir
     end
-    # rubocop: enable HandleExceptions
 
     # Remove orphaned files and empty directories in destination.
     #


pathawks:
Digging into this a bit deeper, I can trace this all the way back to commit number two, when this was still called **autoblog** 7dfe32a597758bb485cc0104b01906781740d977

[Here is what the function looked like:](https://github.com/jekyll/jekyll/blob/7dfe32a597758bb485cc0104b01906781740d977/lib/autoblog/site.rb#L14-L25)

```ruby
    def read_layouts
      base = File.join(self.root, "_layouts")
      dir = Dir.new(base)
      dir.each do |f|
        unless %w{. ..}.include?(f)
          name = f.split(".")[0..-2].join(".")
          self.layouts[name] = File.read(File.join(base, f))
        end
      end
    rescue Errno::ENOENT => e
      # ignore missing layout dir
    end
```

So, it looks like this was not ever meant to swallow some exception in a called function, it was doing all the business right there. I guess we've just carried the cruft along and now here we are, _eight_ years later.

parkr:
Probably from the `File.read` call. I'd be interested in just doing a `File.exist?` call there instead.

pathawks:
> Probably from the `File.read` call. I'd be interested in just doing a File.exist? call there instead.

Sorry to confuse; the code that I posted there is way old and gone. Was just trying to show why this `rescue` was originally added. [We currently don't have any `File.read` calls](https://github.com/jekyll/jekyll/blob/51d9be83a10a976499c42339310e519855f684e4/lib/jekyll/site.rb#L180-L194) :+1:

pathawks:
> when CI passes

You mean appveyor? `master` [is failing on appveyor](https://ci.appveyor.com/project/jekyll/jekyll/build/769)

parkr:
@pathawks Nope! Travis was failing earlier. I kicked it and it passed. 😄 

@jekyllbot: merge +bug

PR #5496, opened 2016-10-19
Restrict Rubocop version

pathawks:
Sometimes Rubocop will add cops in a minor version and all of a sudden CI starts breaking for reasons unrelated to any recent changes. This locks Rubocop to the current minor version so that we can start using new cops on _our_ schedule :+1:

/cc: @jekyll/core @ashmaroli

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,7 +18,7 @@ end
 #
 
 group :test do
-  gem "rubocop"
+  gem "rubocop", "~> 0.44.1"
   gem "cucumber", "~> 2.1"
   gem "jekyll_test_plugin"
   gem "jekyll_test_plugin_malicious"


parkr:
I don't mind that this happens, to some degree. It helps us stay current. Whenever we lock this, we end up just falling behind and it makes updates really painful because it's not just 1 or 2 cops added, it's like 38. That said, it hurts contributors who want a green CI and don't understand why it's failing.

I'll defer here.

pathawks:
Anyway @jekyllbot could open an issue/PR anytime this needs to be bumped? Subscribe to updates from Rubocop?

parkr:
@pathawks That's a good idea. jekyllbot is currently stateless but I bet this would be fruitful work. We could do this for other dependencies as well by checking RubyGems against our Gemfile and gemspec.

ashmaroli:
> Is there any way @jekyllbot could open an issue/PR anytime this needs to be bumped? Subscribe to updates from Rubocop?

Love this idea.. 
Till date, JekyllBot needs explicit commands from the maintainers here to do something.. The only thing I found it does automatically is mark issues as `stale` and warn users that `JekyllHelp` is no longer functional.. 
JekyllBot opening a PR automatically would be so cool.. Hope it would also understand that we currently (as of then) do not intend to bump a dependency version..

pathawks:
* parkr/auto-reply#9
 * parkr/auto-reply#10

parkr:
@jekyllbot: merge +dev

PR #5500, opened 2016-10-19
removing the line: $ jekyll build --watch

d3netxer:
jekyll watches automatically, so having this line implies that it doesn't. Therefore it should be removed.

$ jekyll build --watch
# => The current folder will be generated into ./_site,
#    watched for changes, and regenerated automatically.


pathawks:
Jekyll does not watch by default during `build`, only during `serve`.

d3netxer:
good call nevermind then

pathawks:
Thank you for the PR though :+1:

PR #5501, opened 2016-10-20
Update to reflect jekyll-hook is outdated as deployment method

nicw:
The jekyll-hook method is outdated, but the deployment methods page hasn't been updated. Added text.

### DIFF
--- a/site/_docs/deployment-methods.md
+++ b/site/_docs/deployment-methods.md
@@ -72,7 +72,8 @@ Deploying is now as easy as telling nginx or Apache to look at
 laptops$ git push deploy master
 ```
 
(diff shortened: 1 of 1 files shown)


pathawks:
Would it be better to just remove it, since it's no longer supported?

nicw:
As an outside user, I'd prefer that approach. I wasn't sure if there was a historical reason to keep it around. Shall I submit another?

parkr:
Removing it sounds good to me. 👍

DirtyF:
@nicw It's taken care of, thanks for pointing out 👍

nicw:
Great, thanks!

PR #5435, opened 2016-10-01
Replace backticks within HTML blocks with HTML tags

alexmalik:
Prior to the change backticks were used in an attempt to create a
code block. The problem is that inside block level HTML tags Markdown
is not supported. I have replaced the backticks with a combination of
HTML tags in order to approximately simulate the appearance of a code
block. The docs suggest possible use of span tags in place of the
surrounding div tags as a solution to getting the Markdown to render.
I tried this but no success.

This change improves the readers understanding of the information,
because the reader doesn't have to make sense of raw markdown.

PR #5503, opened 2016-10-21
Update github-pages.md

baiyangcao:
Fix the code style problem


pathawks:
Thank you for submitting this. This is a real problem we need to address.

Actually though, this is a duplicate of #5435. Let's focus our attention on getting that merged. :+1:

PR #5505, opened 2016-10-22
include a hashbang for all benchmark scripts & make them executable

fene:
i have no idea what the circumstances of the benchmark scripts are (i cant even find their usage anywhere in the repo) but it's weird how some files have a hashbang, and even then, none of them are executable. this PR adds hashbangs (`#!/usr/bin/env ruby`, to be precise) to all benchmark files and makes them executable.


parkr:
@jekyllbot: merge +dev

fene:
it's just for consistency anyway!

Issue #3571, opened 2015-03-14
Add a parameter to the markdownify filter for "inline" rendering


hubgit:
When Markdown is used for section headings, for example, the `markdownify` filter produces invalid (and inappropriate) HTML by wrapping the output in a paragraph tag.

I'm curently working around this by doing these conversions client-side using [markdown-it](https://github.com/markdown-it/markdown-it#usage-examples) which has a `renderInline` method, but it's a bit of a hack and is dependent on JavaScript.

Following the syntax of other filters that take optional parameters, `| markdownify: 'inline'` would be ideal.

envygeeks:
Please post the HTML that you get without that plugin.

hubgit:
Input:

```html
{% capture headingText %}Example heading{% endcapture %}
<h2>{{ headingText | markdownify }}</h2>
```

Desired output:

```html
<h2>Example heading</h2>
```

Actual output:

```html
<h2><p>Example heading</p></h2>
```

I've just realised, though, that it would actually make more sense to tell markdownify/kramdown that it's a section heading and let it handle this by itself, like this:

```html
{% capture heading %}## {{ headingText  }}{% endcapture %}
{{ heading | markdownify }}
```

envygeeks:
Hmmm... that is highly undesirable.  I don't think we need an option for it though, this is an actually bug we need to fix... you should in theory get back the `<h2>` no matter what.  There is either a bug in our markdown library or there is a bug in what we send to the markdown library but that's a pretty serious bug no matter what because all things considered that is absolutely invalid HTML it's spitting back to you and shouldn't have even been output no matter what.

hubgit:
The Markdown converter doesn't know it's inside an element where a paragraph isn't allowed, so it's not too surprising that it doesn't handle it automatically - it might only be possible to do this correctly by allowing the "inline" preference to be specified.

On the other hand, I'm totally fine with using concatenating `##` to the front and passing the whole heading to `markdownify`, now that I've realised it's possible, so don't consider this a bug on my behalf.

envygeeks:
Actually you're right, I didn't notice that you were putting it that way.  I do apologize for not paying attention without coffee running through me.  That said, I don't think our preferred markdown converter supports such things (and I don't know that we should either since Redcarpets maintainer points out that the spec states it is not meant to be used in block level elements... https://github.com/vmg/redcarpet/issues/13) 

[...]

PR #5509, opened 2016-10-23
Provide an "inline" mode to markdownify filter

davidstosik:
# Problem

Often, editors want to use markdown to format data (coming from a YML file for example), but do not want the result to be wrapped around a paragraph `<p>` tag.

The Cucumber feature testing `markdownify` is actually a great example of the situation: with `By {{ '_Obi-Wan_' | markdownify }}`, I doubt the result most people would like is "By" and "Obi-Wan" on two separate lines.

Myself, when I used the following HTML earlier:

``` html
<div class="lead">
  {{ post.lead | markdownify }}
</div>
```

I did not expect, nor want, to wrap my lead text around a paragraph. (I do not want the margin that the `<p>` adds.)
Sometimes, you might even want control over the attributes of the generated paragraph:

``` html
<p class="warning">
  {{ post.warning | markdownify }}
</p>
```

This one would actually produce invalid HTML, as it would nest `<p>` elements.
# Proposed solution

My idea is to provide an optional "inline mode" to `markdownify`, off by default, but that developers can turn on easily when writing a layout.

In the solution I propose, when inline mode is on, all `<p>` elements are wiped of the output string.

Here is how to use it:

```
{{ post.lead | markdownify:'inline' }}
```

Please note that I am not altering `markdownify`'s default behaviour, thus not breaking any existing layout.

I'm quite new to Jekyll, so I hope this PR actually makes sense. After browsing the issue queue, it looked like I was not the only person needing this, so I gave it a try. My PR might not be perfect (I'm of course open to suggestions and requests), but at least it is focused, answers a specific needs, and even has tests! (Both Minitest and Cucumber.)

### DIFF
--- a/features/embed_filters.feature
+++ b/features/embed_filters.feature
@@ -57,11 +57,11 @@ Feature: Embed filters
     And I have the following post:
       | title     | date       | layout  | content                                     |
       | Star Wars | 2009-03-27 | default | These aren't the droids you're looking for. |
-    And I have a default layout that contains "{{ 'By _Obi-wan_' | markdownify }}"
+    And I have a default layout that contains "By {{ '_Obi-wan_' | markdownify }}"
     When I run jekyll build
     Then I should get a zero exit status
     And the _site directory should exist
-    And I should see "<p>By <em>Obi-wan</em></p>" in "_site/2009/03/27/star-wars.html"
+    And I should see "By <p><em>Obi-wan</em></p>" in "_site/2009/03/27/star-wars.html"
 
   Scenario: Markdownify a given inline string
     Given I have a _posts directory
--- a/features/embed_filters.feature
+++ b/features/embed_filters.feature
@@ -57,11 +57,23 @@ Feature: Embed filters
     And I have the following post:
       | title     | date       | layout  | content                                     |
       | Star Wars | 2009-03-27 | default | These aren't the droids you're looking for. |
-    And I have a default layout that contains "By {{ '_Obi-wan_' | markdownify }}"
+    And I have a default layout that contains "{{ 'By _Obi-wan_' | markdownify }}"
     When I run jekyll build
     Then I should get a zero exit status
     And the _site directory should exist
-    And I should see "By <p><em>Obi-wan</em></p>" in "_site/2009/03/27/star-wars.html"
+    And I should see "<p>By <em>Obi-wan</em></p>" in "_site/2009/03/27/star-wars.html"
+
+  Scenario: Markdownify a given inline string
+    Given I have a _posts directory
+    And I have a _layouts directory
+    And I have the following post:
+      | title     | date       | layout  | content                                     |
+      | Star Wars | 2009-03-27 | default | These aren't the droids you're looking for. |
+    And I have a default layout that contains "By {{ '_Obi-wan_' | markdownify:'inline' }}"
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "By <em>Obi-wan</em>" in "_site/2009/03/27/star-wars.html"
 
   Scenario: Sort by an arbitrary variable
     Given I have a _layouts directory
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -17,9 +17,7 @@ module Jekyll
       site = @context.registers[:site]
       converter = site.find_converter_instance(Jekyll::Converters::Markdown)
       output = converter.convert(input.to_s)
-      if mode.to_s == 'inline'
-        output = output.gsub(/<p>(.*?)<\/p>\n/im, '\1')
-      end
+      output = output.gsub(%r!<p>(.*?)<\/p>\n!im, '\1') if mode.to_s == "inline"
       output
     end
 
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -13,10 +13,14 @@ module Jekyll
     # input - The Markdown String to convert.
     #
     # Returns the HTML formatted String.
-    def markdownify(input)
+    def markdownify(input, mode = nil)
       site = @context.registers[:site]
       converter = site.find_converter_instance(Jekyll::Converters::Markdown)
-      converter.convert(input.to_s)
+      output = converter.convert(input.to_s)
+      if mode.to_s == 'inline'
+        output = output.gsub(/<p>(.*?)<\/p>\n/im, '\1')
+      end
+      output
     end
 
     # Convert quotes into smart quotes.
--- a/test/test_filters.rb
+++ b/test/test_filters.rb
@@ -57,6 +57,20 @@ class TestFilters < JekyllUnitTest
       )
     end
 
+    should "markdownify with simple string in inline mode" do
+      assert_equal(
+        "one <em>simple</em> string",
+        @filter.markdownify("one _simple_ string", :inline)
+      )
+    end
+
+    should "markdownify with multi-line string in inline mode" do
+      assert_equal(
+        "first line\nsecond line",
+        @filter.markdownify("first line\n\nsecond line", :inline)
+      )
+    end
+
     context "smartify filter" do
       should "convert quotes and typographic characters" do
         assert_equal(


crispgm:
This can be quite useful. But, I think `kramdown` generates wraps with `<p>...</p>`.

features/embed_filters.feature
```diff
@@ -57,11 +57,23 @@ Feature: Embed filters
     And I have the following post:
       | title     | date       | layout  | content                                     |
       | Star Wars | 2009-03-27 | default | These aren't the droids you're looking for. |
-    And I have a default layout that contains "By {{ '_Obi-wan_' | markdownify }}"
+    And I have a default layout that contains "{{ 'By _Obi-wan_' | markdownify }}"
```
pathawks: We _must not_ modify the behavior of existing tests when adding a feature.

pathawks:
I do not see the utility of this. Perhaps you could provide a little more context from real world examples?

One simple way around the problem is

```text
{{ '_Obi-Wan_' | prepend: 'By ' | markdownify }}
```

_If_ we are going to do this, it should be more sophisticated than a `gsub` to remove paragraph tags. Kramdown can be configured to only transform inline elements and not output block tags. See the `SmartyPants` converter for an example of this kind of configuration.

davidstosik:
@pathawks Context is what I describe in "Problem" at the top of this PR, I thought it was clear: I sometimes do not what `markdownify` to wrap its output with a `<p>` tag.
The reason may be:

- because I'm wrapping the output in a `<p>` myself, and I want to have control over its attributes (classes, etc):

  ```html
  <p class="lead">
    {{ post.lead }}
  </p>
  ```

- because I'm outputting the result of this markdown in an inline tag (such as `<h2>`), and putting a `<p>` in an inline tag is invalid

  ```html
  <h2>{{ post.title | markdownify }}</h2>
  ```

- I just don't want that `<p>` tag and its default margins, etc. (nor do I want to write more CSS to correct that).

  ```html
  <div><!-- please give me no extra margin -->
    {{ post.excerpt | markdownify }}
  </div>
  ```

I saw a couple issues about this before starting working on the PR:

- #3571
- #1162

So obviously, I'm not the only person for whom that forced `<p>` is a pain.

I'm definitely not certain myself that using regular expressions is the best way to achieve this (cf. [You can't parse [X]HTML with regex.](http://stackoverflow.com/a/1732454/2341409)), but I thought this specific use-case would be fine (when using `inline` mode, one should be aware that `<p>` tags are going to get stripped, and `<p>` tags generated from markdown are pretty straightforward to match).

If there's any better way to do it, I'll be happy to work on the PR. At least, that's one step further than just opening an issue: there's some code to work and discuss on.

crispgm:
It is used when you want to embed text-based Markdown generations inside HTML blocks.

But IMO, this could be implemented by Markdown converter or by creating Liquid filters like:

`{{ 'markdown here' | markdownify | ltrim: '<p>' | rtrim: '</p>' `

pathawks:
I guess I was looking for a little bit more detail, but looking at your examples:

 * `{{ post.lead | markdownify | replace:'<p>','<p class="lead">' }}`
 * `{{ post.title | prepend:'## ' | markdownify }}`
 * The third example sounds more like a CSS problem

Some instances where just stripping `<p>` would break things:

 * `<h2>{{ '#hashtag' | markdownify:'inline' }}</h2>` => `<h2><h1>hashtag</h1></h2>`
 * `<h2>{{ '* list item' | markdownify:'inline' }}</h2>` => `<h2><ul><li>list item</li></ul></h2>`

As stated in https://github.com/vmg/redcarpet/issues/13#issuecomment-1054113 (which is linked to from the issue you linked to):

> Either way, mixing HTML and Markdown is very often a bad idea, and source of plenty of ambiguous parsing and strange formatting.

If you take a look at [our SmartyPants converter](https://github.com/jekyll/jekyll/blob/v3.3.0/lib/jekyll/converters/smartypants.rb), you can see that it is in fact quite possible [to specify different options for Kramdown](https://github.com/jekyll/jekyll/blob/v3.3.0/lib/jekyll/converters/smartypants.rb#L4-L5). If there is going to be an `inline` option for the `markdownify` filter, it should actually do what it says on the tin and render all span elements and not render any block elements.

I'm sorry but I am still :-1: on this, but I will step back and let others discuss 😺

/cc: @jekyll/core @jekyll/ecosystem

lib/jekyll/filters.rb
```diff
@@ -13,10 +13,12 @@ module Filters
     # input - The Markdown String to convert.
     #
     # Returns the HTML formatted String.
-    def markdownify(input)
+    def markdownify(input, mode = nil)
       site = @context.registers[:site]
       converter = site.find_converter_instance(Jekyll::Converters::Markdown)
-      converter.convert(input.to_s)
+      output = converter.convert(input.to_s)
+      output = output.gsub(%r!<p>(.*?)<\/p>\n!im, '\1') if mode.to_s == "inline"
```
parkr: What if I have a paragraph inside of this code being passed through `markdownify`? We'd want to ensure this is the first and last `p` tags, not random ones.

davidstosik: This is a possibility I overlooked, but you're right there is a problem here.
Removing only the first and last one `p` tag would not really make it better, as it'd potentially leave unopened/unclosed tags in the middle...

davidstosik:
After some reflection, and your useful comments, I realize that what I really need, apparently, is a way to restrict Markdown parsing to inline tags. In other words ignore:

- Paragraphs
- Headers
- Blockquotes
- Lists
- Code Blocks
- Horizontal Rules

I found this article talking about such an approach with Redcarpet: [Render Single-Line Markdown Text with Redcarpet](https://www.icelab.com.au/notes/render-single-line-markdown-text-with-redcarpet). Would such an approach have more chances?

pathawks:
Requiring Redcarpet is a nonstarter.

Like I mentioned above, this can and should be done by setting some kramdown options to only parse inline tags.

It is not a technical problem that prevents me from jumping on board; I just fail to see how adding this complexity would be a benefit to a majority of Jekyll users.

davidstosik:
@pathawks I'm sorry I was not clear. I'm not saying "this would require Redcarpet to work". I'm asking whether filter that says "I'm not gonna output any block-level tag" makes more sense than just cutting out one or many `p` tags?

To achieve that goal, I would take an approach similar to the one that I linked above: inherit from Redcarpet/Kramdown/whatever, and make the block parsers innocuous. For Redcarpet, it means overriding a few methods in an herited class (see [this link](https://www.icelab.com.au/notes/render-single-line-markdown-text-with-redcarpet) I shared previously), for Kramdown it seems it would just required setting [`@block_parsers`](https://github.com/gettalong/kramdown/blob/master/lib/kramdown/parser/kramdown.rb#L75) to an empty list.

Does this make any sense?

test/test_filters.rb
```diff
@@ -57,6 +57,20 @@ def select; end
...
+      assert_equal(
+        "one <em>simple</em> string",
+        @filter.markdownify("one _simple_ string", :inline)
+      )
+    end
+
+    should "markdownify with multi-line string in inline mode" do
+      assert_equal(
+        "first line\nsecond line",
+        @filter.markdownify("first line\n\nsecond line", :inline)
```
pathawks: What can you tell me about this test? Why two `\n` in the input and one `\n` in the output?

pathawks:
@davidstosik Here is how I imagine it would work b0c591a3f2c8f47bdfc8a54774f1bb8e76f6f2a7

test/test_filters.rb
envygeeks: `one\n\ntwo` -> `<p>one</p>\n<p>two</p>` (stripped `p` tags) -> `one\ntwo`... this shouldn't be happening.

envygeeks:
```ruby
def markdownify(input, inline = false)
  if !inline
    Kramdown::Document.new(input).to_html
  else
    input.each_line.map(&method(__method__)).join(
      "\n"
    )
  end
end
```

parkr:
Where are we with this? Any outstanding issues, @pathawks? 

I'm happy to add this feature in as long as we can accomplish it without too great a complexity burden.

Please rebase so we can see green tests on AppVeyor when you have a minute. Thanks!

pathawks:
I am not a fan of the strategy used here: passing rendered Markdown through a simple regex to strip `<p>` tags. I outlined how I think it should be done in b0c591a3f2c8f47bdfc8a54774f1bb8e76f6f2a7

However, I’m just one guy, so I don't want to block this if others believe this to be the better approach `:shrug:`

jhirth:
I also think that only supporting "inlines" (as CommonMark calls it) would be the more robust and therefore preferred solution.

For what it's worth, my use case are programming language keywords in titles. I'd like to use a monospace font for those as I do everywhere else. For the time being, I just strip the paragraph tags via the "remove" filter:

    {{ page.title | markdownify | remove: '<p>' | remove: '</p>' }}

@davidstosik 

> I do not want the margin that the `<p>` adds.

There is a CSS solution for that:

```css
p:not(:first-child) {
  margin-top: 1rem;
}
```

I do the spacing of all blocks (headings, tables, paragraphs, lists, widgets, ...) like that. This way there is only spacing between elements and not at the very top and bottom, which means those elements will also look great inside uniformly padded boxes or modals.

parkr:
I think this is still an important filter to have, but the way it currently removes the `p` tags is unacceptably brittle. We essentially need "convert to markdown, remove first `<p>`, remove last `</p>`" and that'd do it. The current code removes any `<p>` and any `</p>` so it is not specific enough for our requirements.

pathawks:
https://github.com/jekyll/jekyll/commit/b0c591a3f2c8f47bdfc8a54774f1bb8e76f6f2a7

jhirth:
@parkr

Removing the first `<p>` and last `</p>` would be actually *more* brittle since you now can have an unmatched closing `</p>` followed by an unmatched opening `<p>` in the middle. Trashing all `p` tags indiscriminately is much better as far as nasty hacks are considered. @pathawks is definitely on the right track. "Inlines" only is the way to go.

lib/jekyll/filters.rb
pathawks: `gsub` is too fragile for something like this. We have discussed better solutions.

pathawks:
How should we proceed with this? I've given an example of how _I_ would implement this without `gsub` in b0c591a3f2c8f47bdfc8a54774f1bb8e76f6f2a7

I feel bad that it has been half a year and we still haven't solved @davidstosik's problem 😞 

/cc: @jekyll/stability
/cc: @jekyll/ecosystem

parkr:
@pathawks I could get on board with b0c591a3f2c8f47bdfc8a54774f1bb8e76f6f2a7 if properly tested. 😄

kevinstadler:
+1 on getting this merged, I currently use @jhirth's ```{{ page.title | markdownify | remove: '<p>' | remove: '</p>' }}``` since I'm also only dealing with single-line post titles, but clearly there must be a nicer way to do this...

perennialmind:
As an end user, @pathawks version look eminently sane, a sharp contrast to the regex workarounds.

parkr:
A PR to implement @pathawks suggestion from https://github.com/jekyll/jekyll/commit/b0c591a3f2c8f47bdfc8a54774f1bb8e76f6f2a7 would be gladly accepted by me (with tests!). ❤️

ashmaroli:
> A PR to implement @pathawks suggestion from https://github.com/jekyll/jekyll/commit/b0c591a3f2c8f47bdfc8a54774f1bb8e76f6f2a7 would be gladly accepted by me (with tests!). ❤️ 

Reminder-ping @davidstosik and @pathawks

pathawks:
I have no need for this feature. If somebody wants to add some documentation to b0c591a3f2c8f47bdfc8a54774f1bb8e76f6f2a7 and create a PR, I would support that, but I don't have time to do that work myself.

pathawks:
But this PR cannot be merged as long as it relies on `gsub`. Should we close it?

ashmaroli:
> but I don't have time to do that work myself.

k.. I'll cherry-pick that commit and create a new PR.

> Should we close it?

If its not merge-able, yes.. I vote to close..

pathawks:
Closing due to lack of interest.

PR #5515, opened 2016-10-28
github-pages gem should replace all Gemfile stuff

silvenon:
When using github-pages gem, I think you have to remove other lines as well, not just the jekyll one, because github-pages uses its own version of minima and plugins.

Please correct me if I'm wrong. :)


parkr:
Hey @silvenon! You're definitely right that the group isn't 100% necessary to include them in the bundle, the group is an alternate way you add these plugins to your site. Instead of 

```yaml
# in _config.yml
gems:
  - jekyll-gist
  - jemoji
```

you write:

```ruby
# in Gemfile
group :jekyll_plugins do
  gem "jekyll-gist"
  gem "jemoji"
end
```

I think I'll leave the comment as-is for now. Thanks for the patch, though!

PR #5519, opened 2016-10-29
Remove Glynn as deployment option

neudabei:
Remove Glynn as deployment option since that gem is no longer actively maintained. See notes in readme in repository: https://github.com/dmathieu/glynn and https://github.com/dmathieu/glynn/issues/67


DirtyF:
@neudabei 👍 

/cc @jekyll/documentation

neudabei:
😉 Maybe it can be re-considered once the glynn project is maintained again for those who still want to use FTP or rather SFTP with jekyll.

DirtyF:
@jekyllbot: merge +site

Phlow:
Damn, that's sad. I use *Glynn* for small websites on my own webspace. It was so easy to use…

neudabei:
@Phlow +1, I used it too and was very glad when I found it.

PR #5521, opened 2016-10-30
tested on two windows comps, both needed this extra step

ajdamico:


site/_docs/windows.md
```diff
@@ -16,6 +16,8 @@ A quick way to install Jekyll is to follow the [installation instructions by Dav
  2. Install Ruby via Chocolatey: `choco install ruby -y`
  3. Reopen a command prompt and install Jekyll: `gem install jekyll`
 
+Note: if step 3 fails due to an SSL error, save the file [cacert.perm](https://curl.haxx.se/ca/cacert.pem) into your rubygems folder "C:\tools\rubyXX\lib\ruby\X.X.X\rubygems\ssl_certs" and retry step 3.
```
pathawks: 😱 I'm not going to trust a random certificate from `haxx.se` :-1:

There must be a better way to solve this problem.

sa3dany:
The certificates that come with mingw64 which is bundled with git for windows work as well.

git\mingw64\ssl\cert.pem

parkr:
@shiftkey Hello good sir! 👋 Quick question for you. Have you ever encountered SSL errors in Ruby programs on Windows before? Is there a "safe" fix you'd recommend? This affects our documentation but if there's a better programmatic fix we can also try that. Thank you! 😸

shiftkey:
@parkr do you have a link to an issue or discussion around what this failure looks like? I'm not sure how I feel about this workaround, and I think it should be something that gets fixed upstream...

parkr:
@shiftkey They usually look like this: https://github.com/jekyll/jekyll-gist/issues/30

parkr:
Downloading root certs from the Internet doesn't feel safe so we're going to try to find another way. Thanks for the patch though, @ajdamico! 😄

shiftkey:
So I sat down with `ruby 2.3.1p112 (2016-04-26 revision 54768) [x64-mingw32]` and very quickly triggered this issue:

```
C:\Users\shiftkey>gem install jekyll
ERROR:  Could not find a valid gem 'jekyll' (>= 0), here is why:
          Unable to download data from https://rubygems.org/ - SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B: certificate verify failed (https://api.rubygems.org/specs.4.8.gz)

C:\Users\shiftkey>gem --version
2.5.1
```

I do not believe this is something that needs to be fixed in `jekyll`. I was hopeful that the latest stable installer from rubyinstaller.org would have fixed this, but this still appears to be an issue.
I looked at the workaround listed [here](https://gist.github.com/shiftkey/c5a95e8ae7cae12c3cc5a01ce3419a43) - kindly mirrored from @luislavena) but it appears that a set of certificates are installed at the expected location:



I don't have the bandwidth currently to dig into what needs to be fixed here, but I agree that this workaround should not be documented here.

shiftkey:
Scratch that, I found https://github.com/mislav/ssl-tools which gave me some interesting info:

```
C:\Users\shiftkey\src\ssl-tools>ruby doctor.rb api.rubygems.org:443
C:/Ruby23-x64/bin/ruby (2.3.1-p112)
OpenSSL 1.0.1l 15 Jan 2015: C:/Users/Justin/Projects/knap-build/var/knapsack/software/x64-windows/openssl/1.0.1l/ssl
SSL_CERT_DIR=""
SSL_CERT_FILE=""

HEAD https://api.rubygems.org:443
OpenSSL::SSL::SSLError: SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B: certificate verify failed

The server presented a certificate that could not be verified:
  subject: /C=BE/O=GlobalSign nv-sa/CN=GlobalSign Organization Validation CA - SHA256 - G2
  issuer: /C=BE/O=GlobalSign nv-sa/OU=Root CA/CN=GlobalSign Root CA
  error code 20: unable to get local issuer certificate

Possible causes:
  `C:/Users/Justin/Projects/knap-build/var/knapsack/software/x64-windows/openssl/1.0.1l/ssl/cert.pem' does not exist
  `C:/Users/Justin/Projects/knap-build/var/knapsack/software/x64-windows/openssl/1.0.1l/ssl/certs/' is empty
```

luislavena:
@shiftkey avoid manual patching of RubyGems to place the `.pem` files, manually download latest RubyGems, install locally and run the update.

Official instructions here: 

http://guides.rubygems.org/ssl-certificate-update/#installing-using-update-packages

(Replace 2.6.7 in the links and instructions with 2.6.8)

Hope that helps.

shiftkey:
@luislavena that's just what I needed - thanks! I'm not familiar enough with the rubyinstaller.org team to help get this fixed upstream, but hopefully this is enough information to get Jekyll users unblocked!

PR #5524, opened 2016-11-01
URL: warn if key doesn't exist in url drop

### LINKED ISSUE #5332: URL: check to see if the key exists or rescue NoMethodError
If you have URL template like `/:year/:headline`, you'll see a NoMethodError, like:

``` text
  Liquid Exception: undefined method `headline' for #<Jekyll::Drops::UrlDrop:0x007ff42b74aa28> in sitemap.xml
NoMethodError: undefined method `headline' for #<Jekyll::Drops::UrlDrop:0x007ff42b74aa28>
  /Users/parkr/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/jekyll-3.2.1/lib/jekyll/url.rb:89:in `public_send'
  /Users/parkr/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/jekyll-3.2.1/lib/jekyll/url.rb:89:in `block in generate_url_from_drop'
  /Users/parkr/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/jekyll-3.2.1/lib/jekyll/url.rb:88:in `gsub'
  /Users/parkr/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/jekyll-3.2.1/lib/jekyll/url.rb:88:in `generate_url_from_drop'
  /Users/parkr/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/jekyll-3.2.1/lib/jekyll/url.rb:69:in `generate_url'
  /Users/parkr/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/jekyll-3.2.1/lib/jekyll/url.rb:60:in `generated_url'
  /Users/parkr/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/jekyll-3.2.1/lib/jekyll/url.rb:40:in `to_s'
  /Users/parkr/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/jekyll-3.2.1/lib/jekyll/document.rb:211:in `url'
  /Users/parkr/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/jekyll-3.2.1/lib/jekyll/drops/drop.rb:52:in `public_send'
  /Users/parkr/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/jekyll-3.2.1/lib/jekyll/drops/drop.rb:52:in `[]'
```

This is because [`URL#generate_url_from_drop`](https://github.com/jekyll/jekyll/blob/v3.2.1/lib/jekyll/url.rb#L89) uses `public_send` to look for all these values. It should either use `#[]` or warn and ignore a missing method like this.

fene:
fixes #5332 

the tests only cover not throwing a `NoMethodError` because i can't figure out how to capture the logged message to check if it logged the correct message (minitest's `capture_io` only works with kernel `puts`, weirdly enough)

### DIFF
--- a/lib/jekyll/url.rb
+++ b/lib/jekyll/url.rb
@@ -86,7 +86,11 @@ module Jekyll
 
     def generate_url_from_drop(template)
       template.gsub(%r!:([a-z_]+)!) do |match|
-        replacement = @placeholders.public_send(match.sub(":".freeze, "".freeze))
+        begin
+          replacement = @placeholders.public_send(match.sub(":".freeze, "".freeze))
+        rescue NoMethodError
+          Jekyll.logger.warn "", "#{match} is not defined!"
+        end
         if replacement.nil?
           "".freeze
         else
--- a/lib/jekyll/url.rb
+++ b/lib/jekyll/url.rb
@@ -86,15 +86,14 @@ module Jekyll
 
     def generate_url_from_drop(template)
       template.gsub(%r!:([a-z_]+)!) do |match|
-        begin
-          replacement = @placeholders.public_send(match.sub(":".freeze, "".freeze))
-        rescue NoMethodError
-          Jekyll.logger.warn "", "#{match} is not defined!"
+        key = match.sub(":".freeze, "".freeze)
+        unless @placeholders.key?(key)
+          raise NoMethodError, "The URL template key #{key} doesn't exist!"
         end
-        if replacement.nil?
+        if @placeholders[key].nil?
           "".freeze
         else
-          self.class.escape_path(replacement)
+          self.class.escape_path(@placeholders[key])
         end
       end.gsub(%r!//!, "/".freeze)
     end
--- a/lib/jekyll/drops/url_drop.rb
+++ b/lib/jekyll/drops/url_drop.rb
@@ -78,6 +78,11 @@ module Jekyll
       def y_day
         @obj.date.strftime("%j")
       end
+
+      private
+      def fallback_data
+        {}
+      end
     end
   end
 end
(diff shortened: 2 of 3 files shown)


lib/jekyll/url.rb
```diff
@@ -86,7 +86,11 @@ def generate_url_from_hash(template)
 
     def generate_url_from_drop(template)
       template.gsub(%r!:([a-z_]+)!) do |match|
-        replacement = @placeholders.public_send(match.sub(":".freeze, "".freeze))
+        begin
+          replacement = @placeholders.public_send(match.sub(":".freeze, "".freeze))
+        rescue NoMethodError
```
parkr: Instead of a `begin .. rescue`, what about checking for the key first, like

```ruby
template.gsub(%r!:([a-z_]+)!) do |match|
  key = match.sub(":".freeze, "".freeze)
  raise NoMethodError, "No URL template key #{key} exists!" unless @placeholders.key?(key)
  if @placeholders[key].nil?
    "".freeze
  else
    self.class.escape_path(@placeholders[key])
  end
end.gsub(%r!//!, "/".freeze)
```

fene: hmm, by using this approach, it looks like the URLDrop (at least the one i use in my tests) throws an unrelated error because it can't find `fallback_data` when using key? or `#[]`

parkr: Ah, ok. No problem. Define `def fallback_data` in `UrlDrop` as an empty hash.

```ruby
class UrlDrop
  private
  def fallback_data
    {}
  end
end
```

fene:
updated according to @parkr's comments

fene:
@jekyll/core is this ok to merge? its been sitting for a while

parkr:
@jekyllbot: merge +minor

PR #5526, opened 2016-11-02
Fix tests to get script/test to pass on Windows

ashmaroli:
- [x] Fix `TestGeneratedSite`
- [x] Fix `TestFilters`
- [x] Fix `TestSite` (only partially)
- [x] Add new helper method to skip testing symlinks on Windows

--

ref: #5241 
/cc @jekyll/windows

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -158,4 +158,11 @@ class JekyllUnitTest < Minitest::Test
       str
     )
   end
+
+  def skip_if_windows(msg = nil)
+    if Utils::Platforms.really_windows?
+      msg ||= "Jekyll does not currently support this feature on Windows."
+      skip msg.to_s.magenta
+    end
+  end
 end
--- a/test/test_collections.rb
+++ b/test/test_collections.rb
@@ -186,6 +186,9 @@ class TestCollections < JekyllUnitTest
     end
 
     should "include the symlinked file from site.source in the list of docs" do
+      # no support for including symlinked file on Windows
+      skip_if_windows "Jekyll does not currently support symlinks on Windows."
+
       assert_includes @collection.docs.map(&:relative_path), "_methods/um_hi.md"
     end
   end
--- a/test/test_entry_filter.rb
+++ b/test/test_entry_filter.rb
@@ -82,6 +82,9 @@ class TestEntryFilter < JekyllUnitTest
 
     # rubocop:disable Performance/FixedSize
     should "include only safe symlinks in safe mode" do
+      # no support for symlinks on Windows
+      skip_if_windows "Jekyll does not currently support symlinks on Windows."
+
       site = Site.new(site_configuration("safe" => true))
       site.reader.read_directories("symlink-test")
 
@@ -91,6 +94,9 @@ class TestEntryFilter < JekyllUnitTest
     # rubocop:enable Performance/FixedSize
 
     should "include symlinks in unsafe mode" do
+      # no support for symlinks on Windows
+      skip_if_windows "Jekyll does not currently support symlinks on Windows."
+
       site = Site.new(site_configuration)
 
       site.reader.read_directories("symlink-test")
--- a/test/test_generated_site.rb
+++ b/test/test_generated_site.rb
@@ -50,11 +50,13 @@ class TestGeneratedSite < JekyllUnitTest
 
     should "print a nice list of static files" do
       time_regexp = "\\d+:\\d+"
+      #
+      # adding a pipe character at the beginning preserves formatting with newlines
       expected_output = Regexp.new <<-OUTPUT
-- /css/screen.css last edited at #{time_regexp} with extname .css
-- /pgp.key last edited at #{time_regexp} with extname .key
-- /products.yml last edited at #{time_regexp} with extname .yml
-- /symlink-test/symlinked-dir/screen.css last edited at #{time_regexp} with extname .css
+| - /css/screen.css last edited at #{time_regexp} with extname .css
+  - /pgp.key last edited at #{time_regexp} with extname .key
+  - /products.yml last edited at #{time_regexp} with extname .yml
+  - /symlink-test/symlinked-dir/screen.css last edited at #{time_regexp} with extname .css
 OUTPUT
       assert_match expected_output, File.read(dest_dir("static_files.html"))
     end
--- a/test/test_site.rb
+++ b/test/test_site.rb
@@ -277,19 +277,19 @@ class TestSite < JekyllUnitTest
         @site.process
         # generate some orphaned files:
         # single file
-        File.open(dest_dir("obsolete.html"), "w")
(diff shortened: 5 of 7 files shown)


ashmaroli:
This PR is a refined version of the referenced pull.

test/test_site.rb
```diff
@@ -14,14 +14,20 @@ class TestSite < JekyllUnitTest
 
     should "have an array for plugins if passed as a string" do
       site = Site.new(site_configuration({ "plugins_dir" => "/tmp/plugins" }))
-      assert_equal ["/tmp/plugins"], site.plugins
+      array = Utils::Platforms.really_windows? ? ["C:/tmp/plugins"] : ["/tmp/plugins"]
```
XhmikosR: Why hardcode this? You should just use the environment variable.

ashmaroli: > just use the environment variable

please explain more.

ashmaroli: the path is not important here actually. This step and the next is to simply test if **plugins** are correctly detected **as specified**.
e.g. if were to set `plugins_dir => "/apps/bazinga"`, the array returned would be:
```shell
["C:/apps/bazinga"] # on Windows and AppVeyor CI
["/apps/bazinga"] # on Unixoids and Travis CI
```

XhmikosR: I'm just saying, it's bad practice hardcoding paths. Especially when you can use something a lot more reliable to detect the path.

ashmaroli: the more I think about this, I wonder if this is a bug or intended..
when I set, `plugins_dir => "/apps/plugins"`
I should get a relative path to `plugins` dir inside my site, or an absolute_path to `plugins` dir **with** my site.

ashmaroli: moving this segment to a dedicated issue ticket..

test/helper.rb
```diff
@@ -158,4 +158,11 @@ def nokogiri_fragment(str)
       str
     )
   end
+
+  def skip_if_windows(msg = nil)
+    if Utils::Platforms.really_windows?
+      msg ||= "Jekyll doesn't currently support Symlinks on Windows"
```
parkr: This seems like a strange default. Is this only used in symlinks-related code? If not, I'd suggest something like `Jekyll does not currently support this feature on Windows.`

test/test_collections.rb
```diff
@@ -186,6 +186,9 @@ class TestCollections < JekyllUnitTest
     end
 
     should "include the symlinked file from site.source in the list of docs" do
+      # no support for including symlinked file on Windows
+      skip_if_windows
```
parkr: Then here I'd be specific with my message and say `skip_if_windows, "Jekyll does not currently support symlinsk on Windows."`.

test/test_entry_filter.rb
```diff
@@ -82,6 +82,9 @@ class TestEntryFilter < JekyllUnitTest
 
     # rubocop:disable Performance/FixedSize
     should "include only safe symlinks in safe mode" do
+      # no support for symlinks on Windows
+      skip_if_windows
```
parkr: Same here and everywhere you use this method – make that comment the message instead of a comment so when I run the tests, this information is surfaced.

test/test_generated_site.rb
```diff
@@ -50,11 +50,13 @@ class TestGeneratedSite < JekyllUnitTest
...
+      # adding a pipe character at the beginning preserves formatting with newlines
       expected_output = Regexp.new <<-OUTPUT
-- /css/screen.css last edited at #{time_regexp} with extname .css
-- /pgp.key last edited at #{time_regexp} with extname .key
-- /products.yml last edited at #{time_regexp} with extname .yml
-- /symlink-test/symlinked-dir/screen.css last edited at #{time_regexp} with extname .css
+| - /css/screen.css last edited at #{time_regexp} with extname .css
+  - /pgp.key last edited at #{time_regexp} with extname .key
+  - /products.yml last edited at #{time_regexp} with extname .yml
+  - /symlink-test/symlinked-dir/screen.css last edited at #{time_regexp} with extname .css
```
parkr: Why is this necessary? Can you paste the test failure in a comment below?

test/test_site.rb
```diff
@@ -175,7 +175,6 @@ def generate(site)
         method.call(*args, &block).reverse
       end
       @site.process
-      # files in symlinked directories may appear twice
```
parkr: Maybe say here than symlinked directories are excluded from this list and added in when not Windows.

test/test_generated_site.rb
ashmaroli: @parkr :point_down: 
```
Failure:
TestGeneratedSite#test_: generated sites should print a nice list of static files.  [C:/projects/jekyll/test/test_generated_site.rb:59]
Minitest::Assertion: Expected /- \/css\/screen.css last edited at \d+:\d+ with extname .css
- \/pgp.key last edited at \d+:\d+ with extname .key
- \/products.yml last edited at \d+:\d+ with extname .yml
- \/symlink-test\/symlinked-dir\/screen.css last edited at \d+:\d+ with extname .css
/ to match "\n- /css/screen.css last edited at 23:11 with extname .css\n- /pgp.key last edited at 23:11 with extname .key\n- /products.yml last edited at 23:11 with extname .yml\n- /symlink-test/symlinked-dir last edited at 23:11 with extname \n- /symlink-test/symlinked-file last edited at 23:11 with extname \n".
```

ashmaroli:
> Out of curiosity, why are we using really_windows? instead of just windows?

Because `::Platforms.really_windows?` is *more strict* than `::Platforms.windows?`

parkr:
> Because ::Platforms.really_windows? is more strict than ::Platforms.windows?

Why do we want it to be more strict?

ashmaroli:
> Why do we want it to be more strict?

:smiley: Because currently, symlinks is known to not have native support in Windows env. But then it may work in Windows subsystem on Linux

ashmaroli:
TODO: Undo Gemfile changes when `codeclimate-test-reporter` issue is fixed on `master`

parkr:
This is looking excellent!

Gemfile
```diff
@@ -22,7 +22,7 @@ group :test do
   gem "cucumber", "~> 2.1"
   gem "jekyll_test_plugin"
   gem "jekyll_test_plugin_malicious"
-  gem "codeclimate-test-reporter"
+  gem "codeclimate-test-reporter", "0.6.0"
```
parkr: Want to submit a separate PR for this?

ashmaroli: @parkr, if you think, going with `"~> 0.6.0"` is the better way for now, I shall definitely move this to a dedicated PR. Otherwise, this will remain here to have the tests run correctly..

parkr:
@jekyllbot: merge +dev

PR #5528, opened 2016-11-02
Fix bug in plugins_path with trailing slash

### LINKED ISSUE #5527: bug in `plugins_path` with trailing `/` on windows
There exists a bug in how Jekyll handles array of plugins if a new jekyll site on **Windows** is configured to use a **non-default path** to `/plugins`

## To reproduce this bug:
- simply run the test-file `test/test_site.rb` on a Windows machine.

You'll get:  `["<Drive-letter:>/tmp/plugins"]`
Instead of:  `["<abs_path_to_local_Jekyll_Repo>/tmp/plugins"]`

## Workaround:
Change https://github.com/jekyll/jekyll/blob/master/test/test_site.rb#L16 by deleting the trailing `/` of the string; and then run the test again.

You'll get: `["<abs_path_to_local_jekyll_repo>/tmp/plugins"]` as expected, even though, the test fails.

--
/cc @jekyll/windows 
/cc @jekyll/ecosystem

ashmaroli:
fix #5527

### DIFF
--- a/test/test_site.rb
+++ b/test/test_site.rb
@@ -14,14 +14,17 @@ class TestSite < JekyllUnitTest
 
     should "have an array for plugins if passed as a string" do
       site = Site.new(site_configuration({ "plugins_dir" => "/tmp/plugins" }))
-      assert_equal ["/tmp/plugins"], site.plugins
+      assert_equal [File.join(Dir.pwd, "/tmp/plugins")], site.plugins
     end
 
     should "have an array for plugins if passed as an array" do
       site = Site.new(site_configuration({
-        "plugins_dir" => ["/tmp/plugins", "/tmp/otherplugins"]
+        "plugins_dir" => ["/tmp/plugins", "tmp/otherplugins"]
       }))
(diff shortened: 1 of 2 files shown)


ashmaroli:
/cc @jekyll/windows
/cc @parkr 

Ready for review

test/test_site.rb
```diff
@@ -14,14 +14,17 @@ class TestSite < JekyllUnitTest
...
     should "have an array for plugins if passed as a string" do
       site = Site.new(site_configuration({ "plugins_dir" => "/tmp/plugins" }))
-      assert_equal ["/tmp/plugins"], site.plugins
+      assert_equal [File.join(Dir.pwd, "/tmp/plugins")], site.plugins
     end
 
     should "have an array for plugins if passed as an array" do
       site = Site.new(site_configuration({
-        "plugins_dir" => ["/tmp/plugins", "/tmp/otherplugins"]
+        "plugins_dir" => ["/tmp/plugins", "tmp/otherplugins"]
```
ashmaroli: checks against both strings. one with trailing `/` and another without..

PR #5530, opened 2016-11-02
Use only the used Font Awesome icons.

XhmikosR:
The font is generated with https://icomoon.io/app/.

This saves ~50KB.

A couple of questions:

1. Is IE < 9 still supported? If not we could skip .eot.
2. Do you want the json file from the IcoMoon app committed too?


docs/_sass/_font-awesome.scss
```diff
@@ -1,11 +1,10 @@
...
- */
 @font-face {
   font-family: 'FontAwesome';
-  src: url('../fonts/fontawesome-webfont.eot?v=4.4.0');
-  src: url('../fonts/fontawesome-webfont.eot?#iefix&v=4.4.0') format('embedded-opentype'), url('../fonts/fontawesome-webfont.woff2?v=4.4.0') format('woff2'), url('../fonts/fontawesome-webfont.woff?v=4.4.0') format('woff'), url('../fonts/fontawesome-webfont.ttf?v=4.4.0') format('truetype'), url('../fonts/fontawesome-webfont.svg?v=4.4.0#fontawesomeregular') format('svg');
+  src: url('../fonts/FontAwesome.eot?9h6hxj');
+  src: url('../fonts/FontAwesome.eot?9h6hxj#iefix') format('embedded-opentype'),
+    url('../fonts/FontAwesome.ttf?9h6hxj') format('truetype'),
+    url('../fonts/FontAwesome.woff?9h6hxj') format('woff'),
+    url('../fonts/FontAwesome.svg?9h6hxj#FontAwesome') format('svg');
```
pathawks: WOFF [is supported by 92% of browsers](http://caniuse.com/#feat=woff). What users are you targeting with all these new files?

XhmikosR: I suggest that you read about the bulletproof fontface.

DirtyF: I thought [SVG icons were the way to go](https://github.com/blog/2112-delivering-octicons-with-svg) now.

XhmikosR:
Again guys that's a different issue. I simply made this pull request to
reduce the font size. Everything else is for a separate PR which I have no
time to submit myself.

DirtyF:
@XhmikosR I hear you, thanks for pointing that this could be improved, wasn''t aware of this before your PR.

pathawks:
@XhmikosR Reducing file size is certainly a good thing, and I am all :+1: on that.

The fact remains, however, that this PR is adding three additional font formats to the CSS, and that is something that cannot be done without good reason. I'm not saying it mustn't be done; I’' only saying that the motivation is unclear to me.

DirtyF:
@pathawks yeah, I find it too bad that woff2 has gone 😕  
Saving bytes is one thing but browser support is another.

XhmikosR:
@pathawks: where on earth do you see this PR is adding 3 new formats? Please don't spread BS.

XhmikosR:
And seriously, I have no interest on discussing this except for the points I make on my OP.

@parkr: please have a look and cmerge or close.

pathawks:
@XhmikosR Previously `_font-awesome.scss` had only EOT. This PR adds TTF, WOFF, and SVG. I am only discussing the changes you have proposed in this PR. If this change was a mistake, it should be fixed. If I am misunderstanding this change, I would like an explanation.

I am not trying to be difficult, only trying to understand the implications of this PR.

XhmikosR:
No it does **not** have only eot. Scroll to the right.

pathawks:
_That’s_ what I was missing! So sorry for the confusion.

XhmikosR:
Yeah, that is my whole point of asking if IE < 9 is supported ;) If not, I could skip .eot.

parkr:
> Use only the used Font Awesome icons

If we want to use another icon, we have to regenerate this, right? How do you feel about the trade-off between saving KB and adding development time? Not sure how frequently we even need to add 

> Is IE < 9 still supported? If not we could skip .eot.

I don't believe IE < 9 is supported, no. If it doesn't add any bandwidth to the request for IE > 9, Safari, Firefox, and Chrome, then I'd be fine leaving it in as it sounds like we had .eot before this PR.

> Do you want the json file from the IcoMoon app committed too?

What is this JSON file? Can you paste in a gist and paste the link in a reply here? How does IcoMoon play in? Maybe we need documentation in the CSS about how to updated FontAwesome icons – I have no idea what IcoMoon is and how to use it.

XhmikosR:
The json file is what you feed to the app to generate the font without
doing everything from scratch.

Generally this is a fine solution IMO. JUST LOAD THE JSON file, choose the
icons you want and you are done.

DirtyF:
Saving 50kb is not the right measurement here IMHO.

Since this PR is intended to address performance issues, I just ran a [test on webpagetest](https://www.webpagetest.org/result/161109_1S_186/2/details/#waterfall_view_step1) of the current loading on mobile. Chrome load woff2 format by default as all other modern browsers do except IE.

@XhmikosR have you ran some tests that shows that rendering is faster with this PR?

XhmikosR:
Of course it's faster. And in the worst case it simply saves 59 kb of crap
that are not used so it saves bandwidth.

DirtyF:
My point here it's that it seems to it's more important to keep woff2 format for a majority of browsers, and that this will assure browsers download the lighter format.

XhmikosR:
Your point is moot. It's not lighter.

parkr:
Thanks for the PR @XhmikosR. I'd be happy to accept it if you are willing to write a quick README or comment in the generated file which explains how to modify the file. If there is configuration required, then let's commit that here (or put it in the README) as well so that isn't lost for future folks.

XhmikosR:
@parkr: where do you want me to place the info? I can upload the json file in the fonts folder or do you have something else in mind?

parkr:
> @parkr: where do you want me to place the info? I can upload the json file in the fonts folder or do you have something else in mind?

JSON file in the fonts folder would work, or make a README with instructions. I don't even know how to update this! A README with instructions would be lovely.

XhmikosR:
@parkr: can you check if now everything is as you want?

parkr:
AppVeyor failed because the `fmt` test timed out. No clue why that would happen as I don't know the nuances of bash on AppVeyor Windows.

parkr:
@jekyllbot: merge +site

parkr:
Thanks for sticking with this, @XhmikosR! Hooray for saving bandwidth and reducing page size. 👏

PR #5536, opened 2016-11-03
Fix HTML rendering

malaterre:
Previously the YAML example would contains the string 'yaml' at the top of the file. Very confusing for newcomers.


DirtyF:
@malaterre good catch, thanks!

parkr:
Thank you!

@jekyllbot: merge +site

PR #5540, opened 2016-11-03
Update documentation on jekyllrb.com

matthewdu:
The previous doc about `categories` and `tags` does not seem to be true.

For
```
--- # Front Matter
...
categories: piano, music
...
---
```
and
```
--- # Front Matter
...
categories: "piano, music"
...
---
```
the parsed output for `categories` is
```
"categories"=>["piano,", "music"] # there's an extra comma
```
The same applies with `tags`

You can verify this by logging `data` at https://github.com/jekyll/jekyll/blob/master/lib/jekyll/document.rb#L480

Space-delimited string works though. Better yet, this should probably only say YAML list.


DirtyF:
@jekyllbot: merge +site

Issue #5541, opened 2016-11-03
Passing `/` to the absolute URL filter results in a double trailing slash


benbalter:
`{{ "/" | absolute_url }}` generates `http://[DOMAIN]//` (two trailing slashes). Instead, `absolute_url` (and `relative_url`) should account for this edge case by stripping the double trailing slash.

e0i:
I could not reproduce the issue and proposed tests at #5542 related to the issue also pass. Perhaps both issues can now be closed?

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>3.3-stable</code> or <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, please consider building it first as a plugin. Jekyll 3 introduced [hooks](http://jekyllrb.com/docs/plugins/#hooks) which provide convenient access points throughout the Jekyll build pipeline whereby most needs can be fulfilled. If this is something that cannot be built as a plugin, then please provide more information about why in order to keep this issue open.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

pathawks:
I know #5542 didn't reveal this bug, but I can reproduce it when following these instructions: https://github.com/jekyll/jekyll-feed/issues/158#issue-208191943

pathawks:
`{{ "/" | absolute_url }}` generates the proper URL as expected.

`{{ "/" | absolute_url | xml_escape }}` adds an extra slash to the end of the URL.

pathawks:
There's a trailing slash _after_ the Liquid. I am banging my head against a wall right now.

https://github.com/jekyll/jekyll-feed/blob/v0.8.0/lib/feed.xml#L8

PR #5542, opened 2016-11-03
test double slash when input = '/'

ashmaroli:
test #5541


test/test_filters.rb
```diff
@@ -368,6 +368,24 @@ def select; end
...
+        refute_equal "http://example.com/base//", filter.absolute_url(page_url)
+      end
+
+      should "not append a forward slash if input is '/' and nil 'baseurl'" do
+        page_url = "/"
+        filter = make_filter_mock({
+          "url"     => "http://example.com",
+          "baseurl" => nil
+        })
+        refute_equal "http://example.com//", filter.absolute_url(page_url)
```
pathawks: Same as above

test/test_filters.rb
```diff
@@ -368,6 +368,24 @@ def select; end
         assert_equal "http://example.com/base", filter.absolute_url(page_url)
       end
 
+      should "not append a forward slash if input is '/'" do
+        page_url = "/"
+        filter = make_filter_mock({
+          "url"     => "http://example.com",
+          "baseurl" => "/base"
+        })
+        refute_equal "http://example.com/base//", filter.absolute_url(page_url)
```
pathawks: Can we `assert` that it is proper, rather than refuting this particular wrong case?

ashmaroli: done..

parkr:
@benbalter, does this look good to you? I don't see a failing test in the history and I don't trust my reading well enough since I was the one who let this slip through in the first place. 😸

pathawks:
Shouldn't these tests currently be failing? Was something else causing jekyll/jekyll-feed#145?

ashmaroli:
I was not able to reproduce the error on my Windows machine. Hence wrote these tests to have Travis check it..
an edge case I didn't include in the tests : 
`test: if the filter produces "//" when the input and baseurl equals "/"`

e0i:
I have applied the changes manually and checks didn't fail at all. Perhaps it was indeed something else that caused the situation described at #5541 as @pathawks mentioned?

pathawks:
What's up with AppVeyor?

ashmaroli:
> What's up with AppVeyor?

@pathawks the test log pre-dates fixing AppVeyor on the `master`. Restarting the build should sort this. To be on the safer side, I'll update the branch to latest `master` and initiate a new build altogether.

parkr:
Always good to have extra tests for things like this.

@jekyllbot: merge +bug

ashmaroli:
@parkr I think it would be better that @jekyllbot adds a single label `bug-fix` instead of the current two.. mainly because tagging a PR with `bug` doesn't seem that right.

PR #5559, opened 2016-11-08
Use site.baseurl before link and post_url tags

ryandesign:
This PR updates the documentation to show that `{{ site.baseurl }}` must be used before the `{% link %}` and `{% post_url %}` tags.

### DIFF
--- a/docs/_docs/templates.md
+++ b/docs/_docs/templates.md
@@ -543,7 +543,7 @@ will generate the correct permalink URL for the post you specify.
 
 ```liquid
 {% raw %}
-{% post_url 2010-07-21-name-of-post %}
+{{ site.baseurl }}{% post_url 2010-07-21-name-of-post %}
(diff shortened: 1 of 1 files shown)


ryandesign:
> Let's just use the new `relative_url` filter.

That's what I tried first. I couldn't get it to work. When I researched why, it seemed that filters can only be used with `{{ ... }}` notation, not with `{% ... %}` notation.

parkr:
@pathawks We'd have to make sure `link` calls that filter explicitly. Want to submit a PR to that effect?

pathawks:
#5569

pathawks:
> We'd have to make sure link calls that filter explicitly.

Since that'd be a breaking change, let's update the docs first :+1:

pathawks:
Thank you @ryandesign :+1:

@jekyllbot: merge +site

PR #5564, opened 2016-11-09
Be much more specific about ignoring specific vendored directories.

parkr:
An alternative to https://github.com/github/pages-gem/pull/352, but it probably won't work.

~~The way exclusion works in Jekyll is really pretty naïve. In `EntryFilter#filter`, it iterates through an array of pathnames and checks to see if each pathname is included, excluded, etc. The way reading works in Jekyll, this list of pathnames is just the basename, so `vendor/cache` becomes first `vendor`, then (if not excluded), we get the array of files and directories inside of `vendor` as basenames, e.g. `cache` instead of `vendor/cache`.~~

~~I'd love to rethink the way we do reading so we get the full pathname from the site source, or allow `EntryFilter`'s `base_directory` path to be included in exclusion testing.~~

I checked and the code for exclusion already does this! 🎉 

```ruby
def excluded?(entry)
  glob_include?(site.exclude, relative_to_source(entry))
end
```

### DIFF
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -17,7 +17,9 @@ module Jekyll
       # Handling Reading
       "safe"              => false,
       "include"           => [".htaccess"],
-      "exclude"           => %w(node_modules vendor/bundle vendor/ruby vendor/cache),
+      "exclude"           => %w(
+        node_modules vendor/bundle/ vendor/cache/ vendor/gems/ vendor/ruby/
+      ),
       "keep_files"        => [".git", ".svn"],
       "encoding"          => "utf-8",
       "markdown_ext"      => "markdown,mkdown,mkdn,mkd,md",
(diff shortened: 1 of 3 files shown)


lib/jekyll/configuration.rb
```diff
@@ -17,7 +17,7 @@ class Configuration < Hash
       # Handling Reading
       "safe"              => false,
       "include"           => [".htaccess"],
-      "exclude"           => %w(node_modules vendor),
+      "exclude"           => %w(node_modules vendor/bundle vendor/ruby vendor/cache),
```
ashmaroli: is there a way to use wildcards to exclude any future sub-directory under `/vendor/`

crispgm:
What about making the exclusion more explicitly?

parkr:
> What about making the exclusion more explicitly?

@crispgm What do you mean? We are making the exclusion more explicit here. By adding the `/`, then we are saying "ignore these directories." We don't want to exclude all of `vendor/`.

parkr:
@jekyllbot: merge +bug

PR #5566, opened 2016-11-09
Use an older version of codeclimate-test-reporter gem

ashmaroli:
only to get the ball rolling on this issue..


ashmaroli:
oops! late by a few minutes.. handled already..

PR #5514, opened 2016-10-25
Add jekyll-pinboard to list of third-party plugins

georgemandis:
Added link to my Jekyll plugin: https://github.com/snaptortoise/jekyll-pinboard-plugin. Interfaces with the Pinboard API to make data for specified tags available to the template, sort of similar to the built-in Data Files support.

PR #5570, opened 2016-11-10
Fix link to jekyll-pinboard plugin

DirtyF:
fix #5514 

/cc @parkr


DirtyF:
@jekyllbot: merge +site

PR #5580, opened 2016-11-15
Document necessary local site.github.url override

jeffbowen:
When previewing your site locally, `site.github.url` needs to point to localhost. This seems to require an override in `config.yml` and is the recommended approach mentioned in @benbalter's [Jekyll Style guide](http://ben.balter.com/jekyll-style-guide/config/#url).

### DIFF
--- a/docs/_docs/github-pages.md
+++ b/docs/_docs/github-pages.md
@@ -31,9 +31,14 @@ site builds properly, use `site.github.url` in your URL's.
 {% endraw %}
 ```
 
-This way you can preview your site locally from the site root on localhost,
-but when GitHub generates your pages from the gh-pages branch all the URLs
-will resolve properly.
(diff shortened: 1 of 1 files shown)


DirtyF:
@jeffbowen Thanks for the PR, but since Jekyll 3.3 this is taken care of. (see #5431)

You might want to update Ben's styleguide instead ;)

jeffbowen:
Interesting. I'm using 3.3 and it doesn't work without my override. My remote origin is set correctly and I have confirmed that `{{ jekyll.environment }}` is `development`. Will keep looking into why it doesn't work automatically for me.

DirtyF:
Sorry to hear that @jeffbowen, feel free to reopen this issue if you have more details to submit.

What is the ouput of `--trace` if you don't fill `site.url` ou `site.github.url`?

PR #5582, opened 2016-11-15
Remove instructions to install Jekyll 2 on Windows

DirtyF:
As Jekyll 2 is no more supported, we only mention instructions for Jekyll 3.

/cc @jekyll/documentation


DirtyF:
@jekyllbot: merge +site

PR #5587, opened 2016-11-17
Fix bad config YAML in collections example

happycollision:
If you add the previous version to your config file, you'd get an error on build/serve.

### DIFF
--- a/docs/_docs/configuration.md
+++ b/docs/_docs/configuration.md
@@ -513,8 +513,8 @@ With these defaults, all posts would use the `my-site` layout. Any html files th
 
 ```yaml
 collections:
-  - my_collection:
-      output: true
+  my_collection:
+    output: true
 
 defaults:
   -


happycollision:
That's what I thought, but putting this into the config file:

```yaml
collections:
  - my_collection:
      output: true
```

results in this error in the terminal:

```txt
Configuration file: /Users/Don/Git Projects/some-site/_config.yml
            Source: /Users/Don/Git Projects/some-site
       Destination: /Users/Don/Git Projects/some-site/_site
      Generating... 
jekyll 3.0.2 | Error:  undefined method `gsub' for {"my_collection"=>{"output"=>true}}:Hash
```

Conversely, this

```yaml
collections:
  my_collection:
    output: true
```

produces no errors and behaves as expected (the collection outputs and I can access it at `site.my_collection`).

This also seems to be exactly in line with the [current documentation for collections](http://jekyllrb.com/docs/collections/#step-3-optionally-render-your-collections-documents-into-independent-files). (Or that same [collection documentation from the latest commit on master](https://github.com/jekyll/jekyll/blob/6be62def90e1b379b72f9b656094117d0dad4a74/docs/_docs/collections.md#step-3-optionally-render-your-collections-documents-into-independent-files))

Just above that link on those pages is the "Step 1" section, which seems to suggest that if you want to specify metadata for a collection, you must not use an array. From the documentation: 

> Add the following to your site's `_config.yml` file, replacing `my_collection`
> with the name of your collection:
> 
> ```yaml
> collections:
> - my_collection
> ```
> 
> You can optionally specify metadata for your collection in the configuration:
> 
> ```yaml
> collections:
>   my_collection:
>     foo: bar
> ```

Am I missing something?

parkr:
Thank you!

@jekyllbot: merge +site

PR #5591, opened 2016-11-20
add 'jekyll extract' command to library

ashmaroli:
this command allows one to easily copy files or directories from within the theme-gem in use, to the
site-source.

Example usage:
  - `bundle exec jekyll extract _layouts`
  - `bundle exec jekyll extract _layouts/default.html`
  - `bundle exec jekyll extract _layouts --force`

--
Ref: A [comment](https://github.com/jekyll/jekyll/issues/5311#issuecomment-244173059) made by @parkr 
/cc @jekyll/ecosystem

### DIFF
--- a/features/extract_command.feature
+++ b/features/extract_command.feature
@@ -1,0 +1,91 @@
+Feature: Extracting theme contents to source
+  As a hacker who likes to customize stuff
+  I want to be able to easily obtain files from my theme-gem
+  In order to override the default theme layouts and styles
+
+  Scenario: Extracting an entire directory
+    Given I have a configuration file with "theme" set to "test-theme"
+    When I run jekyll extract _layouts
+    Then I should get a zero exit status
+    And the _layouts directory should exist
+    And the "_layouts/default.html" file should exist
+
+  Scenario: Extracting an entire directory with subdirectories
+    Given I have a configuration file with "theme" set to "test-theme"
+    When I run jekyll extract assets
+    Then I should get a zero exit status
+    And the assets directory should exist
+    And the "assets/style.scss" file should exist
+    And the "assets/img/logo.png" file should exist
+
+  Scenario: Extracting multiple directories
+    Given I have a configuration file with "theme" set to "test-theme"
+    When I run jekyll extract _layouts assets
+    Then I should get a zero exit status
+    And the _layouts directory should exist
+    And the "_layouts/default.html" file should exist
+    And the assets directory should exist
+    And the "assets/style.scss" file should exist
+    And the "assets/img/logo.png" file should exist
+
+  Scenario: Extracting only a specific file
+    Given I have a configuration file with "theme" set to "test-theme"
+    When I run jekyll extract assets/img/logo.png
+    Then I should get a zero exit status
+    And the assets directory should exist
+    And the "assets/img/logo.png" file should exist
+    And the "assets/style.scss" file should not exist
+
+  Scenario: Extracting specific files
+    Given I have a configuration file with "theme" set to "test-theme"
+    When I run jekyll extract _layouts/default.html assets/img/logo.png
+    Then I should get a zero exit status
+    And the _layouts directory should exist
+    And the "_layouts/default.html" file should exist
+    And the assets directory should exist
+    And the "assets/img/logo.png" file should exist
+    And the "assets/style.scss" file should not exist
+
+  Scenario: Extracting a combo of directory and specific files
+    Given I have a configuration file with "theme" set to "test-theme"
+    When I run jekyll extract _layouts assets/img/logo.png assets/style.scss
+    Then I should get a zero exit status
+    And the _layouts directory should exist
+    And the "_layouts/default.html" file should exist
+    And the assets directory should exist
+    And the "assets/img/logo.png" file should exist
+    And the "assets/style.scss" file should exist
+
+  Scenario: Extracting a file outside the theme-gem
+    Given I have a configuration file with "theme" set to "test-theme"
+    When I run jekyll extract ../../pwd
+    Then I should get a non-zero exit status
+
+  Scenario: Extracting a non-existent file
+    Given I have a configuration file with "theme" set to "test-theme"
+    When I run jekyll extract _layouts/test.html
+    Then I should get a non-zero exit status
+    And I should see "Error: Specified file or directory doesn't exist" in the build output
+    And the "_layouts/test.html" file should not exist
+
+  Scenario: Extracting a file that already exists at destination
+    Given I have a configuration file with "theme" set to "test-theme"
+    And I have a _layouts directory
+    And I have a "_layouts/default.html" page that contains "test-layout"
+    When I run jekyll extract _layouts/default.html
+    Then I should see "Error: 'default.html' already exists at destination. Use --force to overwrite." in the build output
+
+  Scenario: Force extraction of a file that already exists at destination
+    Given I have a configuration file with "theme" set to "test-theme"
+    And I have a _layouts directory
+    And I have a "_layouts/default.html" page that contains "test-layout"
+    When I run jekyll extract _layouts/default.html --force
+    Then I should get a zero exit status
+    And I should see "Extract: /_layouts/default.html" in the build output
+    And I should see "default.html from test-theme:" in "_layouts/default.html"
+
+  Scenario: Extracting with nothing specified
+    Given I have a configuration file with "theme" set to "test-theme"
+    When I run jekyll extract
+    Then I should get a non-zero exit status
+    And I should see "Error: You must specify a theme directory or a file path." in the build output
--- a/features/extract_command.feature
+++ b/features/extract_command.feature
@@ -3,6 +3,30 @@ Feature: Extracting theme contents to source
   I want to be able to easily obtain files from my theme-gem
   In order to override the default theme layouts and styles
 
+  Scenario: Inspecting the contents of a directory to be extracted
+    Given I have a configuration file with "theme" set to "test-theme"
+    When I run jekyll extract _layouts --show
+    Then I should get a zero exit status
+    And I should see "Listing: Contents of '/_layouts' in theme gem..." in the build output
+    And I should see "/_layouts/default.html" in the build output
+
+  Scenario: Inspecting the contents of multiple directories to be extracted
+    Given I have a configuration file with "theme" set to "test-theme"
+    When I run jekyll extract _layouts assets --show
+    Then I should get a zero exit status
+    And I should see "Listing: Contents of '/_layouts' in theme gem..." in the build output
+    And I should see "/_layouts/default.html" in the build output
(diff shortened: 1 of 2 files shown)


ashmaroli:
Milestone: `v3.4`
TODO: Add minitests and documentation, if this pull generates interest among @jekyll/core

mattr-:
Is there a reason why one wouldn't run `bundle open`, `bundle show` or something similar to get at the files in their theme gem?

features/extract_command.feature
```diff
@@ -0,0 +1,62 @@
+Feature: Extracting theme contents to source
```
mattr-: I'm not seeing a test for any of the sad paths that could be here. Could you add a few please?

ashmaroli:
> Is there a reason why one wouldn't run bundle open, bundle show or something similar to get at the files in their theme gem?

The user is still at their liberty to run `bundle show <theme-name>` to get the path and `bundle open <theme-name>` to edit the theme-gem contents. 
This command **copies** the requested file or entire director(ies) over to the site's `source_dir`

features/extract_command.feature
ashmaroli: if by **test**, you meant minitest module, I encountered some roadblocks while running it locally and decided to settle for a cucumber feature.

parkr:
I like the idea of this. I think we should come up with a different name (`jekyll theme-copy <path>` perhaps? I'm open to ideas here, but I think `extract` isn't the right one), and we should think through the user experience of having to know the path (should we do an all-or-nothing copy, or should we allow people to specify `--layout default` instead of `_layouts/default.html`?) Remember that gem-based themes are really focused on not-very-technical content creators, so using a common language without having to know the paths might be helpful.

ashmaroli:
I'm open to suggestions as well.
I opted for `extract` to allude the idea of 'extracting' the specified content from the theme gem..

`jekyll theme extract` seems like an argument is passed to `jekyll theme` command.. not a bad idea.. but may end up confusing..

This command also copies over entire directory if a user isn't aware of what file to copy over.. just that, copying a directory over is destructive. It doesn't matter if a file within already exists at source.
All they have to do is `jekyll extract _layouts` and voila! they have the entire `/_layouts` copied over to their source.

ashmaroli:
Since we don't have a strict guideline enforced on what a layout template should be named, the only things that are going to be *constant* are the directory names: `_layouts`, `_includes`, `_sass` & `assets` 

I fail to see how `--layout default` is better than the simpler and scalable, `_layouts/default.html` ..

need to copy over the scss file? run `jekyll extract assets/main.scss`
need to copy over the page layout? run `jekyll extract _layouts/page.html`
need to copy over the entire scaffold over? run `jekyll extract _layouts _includes _sass assets`

features/extract_command.feature
ashmaroli: oh those.. good point.. :+1: 
I'll update the feature.

parkr:
> I fail to see how `--layout default` is better than the simpler and scalable,` _layouts/default.html` ..

When you setup a new site and all you have is a `_config.yml` file with `theme: minima` and some files in `_posts` and maybe an `index.html`, it won't even occur to you that you should extract `_layouts/default.html` when you want to modify the default layout. The reason it's better is because it doesn't assume that you know where layouts are placed in your Jekyll site. This is more approachable because the more previous knowledge you need, the higher the learning curve. We want this feature to be accessible for those who are just learning, right?

features/extract_command.feature
ashmaroli: done..

ashmaroli:
> it won't even occur to you that you should extract _layouts/default.html when you want to modify the default layout.

@parkr I've added a `--show` switch that when used along with a `<directory>` argument, prints the contents of the directory thus informing a (new) user about the `relative_path(s)_to_<template>`. They can then proceed with the *extraction* as intended..
I agree, that is not what you suggested.

I decided against implementing *switches* to determine what to extract in light of the following points favoring current route:
 - Multiple arguments allow one to *extract* random layout file, include file, asset file, etc together in one step
`jekyll extract _layouts/default.html _includes/head.html assets/main.scss _sass/minima.scss`
 - Should the `_data` folder be included in the future editions of a theme, there's no need to implement another switch.
 - [Edge-case] Any valid file within the theme gem can be *extracted* to the source by specifying the file's path relative to the theme-gem root directory. 
(say a theme shipped with some files that Jekyll 3.4 may not read-in (e.g. `_data/locales/fr.yml` or a pre-defined `_config.yml`) but can be *extracted* and used anyways.)

---
@mattr- I've added the various possible negative scenarios to the associated cucumber test.
@DirtyF, To what file do i add the documentation for this command?

ashmaroli:
@parkr *et al*, hows `extract-theme` for the renamed command? 
- it goes well with existing `jekyll new-theme`..
- `jekyll extract-theme _layouts/default.html` keeps the original notion of having a theme-file extracted to `source_dir` as well..

DirtyF:
@ashmaroli if you ask me, I would tend to prefix the command with `theme` or even [make it an option of a more global theme command.](https://github.com/jekyll/jekyll/pull/4848#issuecomment-215167195)

Some examples based on @parkr proposal:

```sh
jekyll theme-copy --layout default
jekyll theme-copy --include head
jekyll theme-copy --sass main
jekyll theme-copy --assets img/logo
```

nice to have : 
User could omit extensions.
autocompletion  (if you specify --layout, we search in the _layouts folder of the theme)

Shortened commands:

```sh
jekyll theme-copy -l default
jekyll theme-copy -i head
jekyll theme-copy -s main
jekyll theme-copy -a img/logo
```

Oneliners:

```sh
jekyll theme-copy -l default -i head -s main -a img/logo
jekyll theme --copy -l default -i head -s variables
jekyll theme -c -l default -i head -s variables
jekyll t -c -l default -i head -s variables
```

ashmaroli:
@DirtyF Thank you for reminding me about that *comment*.. had forgotten about it.. I loved it then.. n still like it.. but IMHO, its more suited for v4.0 than a v3.x.. with altering `new-theme` to `theme --new` even if it doesnt break anything..

I'm not so excited about your current suggestion though.. to me it is like adding more layers of complexity to an otherwise simple step..
- for every --switch called, the overall execution *may* get slowed down considerably.
- we would have to create a switch for `<theme>/_data` should it be read in a future release.
- we would have to define another switch to *copy a file from theme-root (namely a pre-defined `_config.yml` within the theme gem.)*
- add extra computation to check for `assets/avatar.jpg` and `assets/avatar.png` should both exist.
- add extra computation to check for `_includes/head.html` and `_includes/head.svg` should both exist.
- add extra computation to differentiate _sass/minima.scss and _sass/minima/ when I run `jekyll theme -c -s minima`

ashmaroli:
hello guys, I'm thinking of moving this to another gem [`jekyll-plus`](https://github.com/ashmaroli/jekyll-plus) as (`jekyll extract-theme command`) and start a new pull here, to have `jekyll.rb` with:
```
Jekyll::External.require_if_present("jekyll-plus")
```
Is that a welcome route?  
I'm opting for it against having this merged to core  or have that gem as a `blessed-gem` is so that the commands provided by `jekyll-plus` **be optional** *(available only if its installed && present in their Gemfile)*
This also allows Parker to implement his idea to core without conflict, following which, the jekyll-plus users though will just have to choose either of the commands.
I agree, its sorta against the `"there should be only one way to have a thing be done"` mantra

/cc @jekyll/core

parkr:
I don't think we'd require `jekyll-plus` unless we wrote it, we'd rather see the user add `gem "jekyll-plus", group: :jekyll_plugins` to their Gemfile.

I think this command is necessary and should be in Jekyll core, but I don't like the `extract-theme` name and there are open questions about whether we should use `--layout default` or `_layouts/default.html` or support both.

ashmaroli:
> I don't think we'd require `jekyll-plus` unless we wrote it

Thank you @parkr  for clarifying it

ashmaroli:
any updates on this?
No hurry.. just asking.. :innocent:

pathawks:
There are still some pretty foundational questions outstanding about The Right Way to do this. I really don't want to ship something half-baked; it's important that we get this right.

It seems like we are still trying to figure out how users will configure themes in a somewhat standard fashion; a user shouldn't have to guess which of these hundred files they need to modify if they just want to add a logo or something.

ashmaroli:
> a user shouldn't have to guess which of these hundred files they need to modify if they just want to add a logo or something.

shouldn't the theme's `README` tell them all those details?

parkr:
This makes me yearn for a `jekyll edit` command, which would pull in from the theme, copy to the user's site, and open in the editor. The only reason I can think of that someone would want to copy this file is to edit it. Would this be a better command?

ashmaroli:
I like `jekyll edit`.
Its short, and conveys the intent as well.
That resolves the dispute on the name.

--
Now onto tackling the "`foundational questions outstanding about The Right Way to do this`" :wink:

pathawks:
If I want to edit the footer that came with my theme, how do I do that?

* `jekyll edit --include footer`
* `jekyll edit --layout footer`
* `jekyll edit --layout default`

Do I need to understand all the details of how the theme organizes its files? Will all themes organize those files in similar ways?

ashmaroli:
> Will all themes organize those files in similar ways?

Its the lack of this "enforcement" that has me turning down Parker's suggestion.
Instead my base approach works fine here:
  - Minima's documentation tells me (in reality, not yet) that the `footer` is an *include* that can be found at `_includes/footer.html`
  - so I just run `$ bundle exec jekyll edit _includes/footer.html` and voila! there is an `_includes/footer.html` at my `source`. (soon: opened in my default editor)

ashmaroli:
@parkr some minor hiccups with implementing the `jekyll edit` suggestion. Would like your feedback.
  - the main intention of this command is to have files extracted from the theme-gem to `source` (so that they can be easily *edited* as needed.) The names `extract`, `extract-theme`, `theme-copy` all suggest that this command does not relate to files at `source`. But `jekyll-edit` would send users the wrong message that even template files present at `source` (user-created or manually copied from the theme) can be edited by this command as well. The remedy would be to opt for **`jekyll edit-theme`** instead.
  - if `jekyll edit-theme` would open a file in default editor automatically, how do we go about using the command with directories as argument? e.g. `jekyll edit-theme _includes`?
    - Open all files within `_includes` in multiple instances of the default editor?
    Edge-case: Users with low available memory may face issues with opening numerous instances of editor especially if the said directory contains numerous files of considerable size. (or)
    - Restrict the open-in-default-editor for just files as arguments?

I now think that instead of opening the files in the editor by default, it would be better to have that feature as an option.

```sh
$ bundle exec jekyll copy-theme _layouts --edit

# (or)

$ bundle exec jekyll extract _layouts --edit
```

mattr-:
Due to

>There are still some pretty foundational questions outstanding about The Right Way to do this

I'm going to close this pull request. It's not because we don't want the feature but I think a discussion first about a larger `edit` type command should happen before we start slinging around more code for this.

PR #5592, opened 2016-11-20
Fix example URL inconsistency

brainscript:
In case of the GitHub example links URLs like `<a href="https://github.com/{{ org.username }}">` with the protocol, the domain **and** the variable were used, but for the Twitter example link only the `{{  author.name }}` was set as `href`. To make it more consistent I added the `https://twitter.com/` to the URL.


DirtyF:
@jekyllbot: merge +site

DirtyF:
@brainscript thanks!

PR #5594, opened 2016-11-22
Disable `Style/EndOfLine` cop for dev on Windows

ashmaroli:
disable the `Style/EndOfLine` cop to allow developers on Windows with git configured to use Windows- style-line-endings to exec `script/fmt` and not receive *unnecessary* offense warnings.

*Note: This may be a lone issue with my configuration.. or not..* 

--
/cc @jekyll/windows  
label: `windows`

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -71,6 +71,8 @@ Style/EmptyLinesAroundAccessModifier:
   Enabled: false
 Style/EmptyLinesAroundModuleBody:
   Enabled: false
+Style/EndOfLine:
+  Enabled: false
 Style/ExtraSpacing:
   AllowForAlignment: true
 Style/FileName:


pathawks:
I can see maybe turning this off when testing locally, it feels like it should be enabled when testing on Travis to make sure that no Windows line endings get merged, right?

ashmaroli:
> it feels like it should be enabled when testing on Travis to make sure that no Windows line endings get merged, right?

Valid point. 
I keep getting  `*.rb:1:1: C: Style/EndOfLine: Carriage return character detected.` for every existing .rb file in the library,  But never see this occurring on AppVeyor though.. 

If Windows line-endings are indeed detrimental to Jekyll's execution, you may close this pull.

envygeeks:
You shouldn't be using carriage returns at all in our source.

XhmikosR:
Just set up .gitattributes properly.

ashmaroli:
I have the following in my `.gitconfig`, is that the issue?
```
[core]
    autocrlf = true
```

XhmikosR:
Yes, but you shouldn't need to change that. Instead, the repo should enforce LF via .gitattributes.

ashmaroli:
@envygeeks, when I run `script/fmt` on my Windows 7 system, against the `master` branch, Rubocop checks 118 files and calls the offense on **all 118 files**. At first I thought, its an issue with Rubocop. But, I dont see any issues filed for this. So, I assumed this to be an edge-case scenario with just this repo.

XhmikosR:
https://help.github.com/articles/dealing-with-line-endings/

So basically, list the filetypes you want and set them to use LF. That's for another PR BTW.

ashmaroli:
Thanks @XhmikosR, you may create the new PR as you see fit.

PR #5573, opened 2016-11-12
Bring documentation on 'Directory Structure' up-to-date

ashmaroli:
The documentation on 'Directory Structure' contains a few dated information and errors that can confuse a new Jekyll user. This pull fixes those errors.
- [x] Add a note on differing directory structure for sites created with Jekyll v3.2+
- [x] Add a note about `_sass` directory contents.
- [x] Fix any remaining typos or errors in general.

--

### Screenshot of the new info panel:
dirst

--
/cc @jekyll/documentation

PR #5602, opened 2016-11-24
Add an excluded 'README.txt' to `/site_template`

ashmaroli:
This pull adds a README.txt to the site_template that will be ignored by git and excluded from `site.process`.
The README mainly serves to alleviate confusion among experienced users upgrading to `v3.2+` around gem-based themes and missing directories.; secondarily easing new users to getting acquainted with using Jekyll.

--
#### Why a `README.txt` instead of `README.md` and why it would be ignored by Git and Jekyll: 
 - This document was visioned to be a sort-of welcome-note to the user, instead of being a manual on the user's blog repo. They may choose to include a dedicated `README.md` for that purpose or convert this document to a markdown document, but that wouldn't be done by default.
 - A `README.md` on a Jekyll Site - Source repo usually contains only one line:
`"Source to my personal blog powered by Jekyll"`
 - A user needn't concern themselves with having to edit **this file** as a README is not so important for their Jekyll Blog Source Repo., especially, if it would be ignored by git and Jekyll. 

--
/cc @jekyll/ecosystem 
@DirtyF for the similarity to #5573


lib/site_template/README.txt
```diff
@@ -0,0 +1,23 @@
+===============================
+Welcome to your new Jekyll Site
+===============================
```
parkr: Can we do a standard H1 or H2 with `#` or `##`?

lib/site_template/.gitignore
```diff
@@ -1,3 +1,4 @@
 _site
 .sass-cache
 .jekyll-metadata
+README.txt
```
parkr: ?

lib/site_template/README.txt
```diff
@@ -0,0 +1,23 @@
...
+
+This file generated by Jekyll will neither be included in the site built
+when you run `jekyll build` (or `jekyll serve`) nor will it be read by
+git if you plan to have your site set up as a git repository.
```
parkr: I see why we wouldn't want it in the compiled site, but why ignore it from your git source tree? Most repositories nowadays have a README and I'd argue it's all the better that they do 😄

lib/site_template/_config.yml
```diff
@@ -32,3 +32,4 @@ gems:
 exclude:
   - Gemfile
   - Gemfile.lock
+  - README.txt
```
parkr: 👍 Can we call it `README.md` please?

lib/site_template/README.txt
```diff
@@ -0,0 +1,23 @@
...
+
+For the new users of Jekyll, Welcome to Jekyll! :) The directory you just
+generated is a self-sufficient blog. You should be able to view this blog
+in your browser by simply running `bundle exec jekyll serve` in your
+command-terminal and pointing your browser to http://localhost:4000/. For
+more information on using Jekyll, goto: https://jekyllrb.com/docs/home/
```
parkr: s/goto/go to/

lib/site_template/README.txt
```diff
@@ -0,0 +1,23 @@
...
+
+For the experienced users of Jekyll who just recently upgraded from v3.1.x
+and older, don't be confused by the missing folders in your blog. They've
+been bundled together as a theme-gem called 'minima' which your site will
```
parkr: You might link out to `minima` here 😄

lib/site_template/README.txt
```diff
@@ -0,0 +1,23 @@
...
+look-up to during the 'jekyll build' process. This configuration can be
```
parkr: "during Jekyll's build process." Using CLI commands as short-hand isn't needed :)

lib/site_template/README.txt
```diff
@@ -0,0 +1,23 @@
...
+changed by editing the "theme: minima" line in your '_config.yml' and the
```
parkr: Let's be consistent with either single or double quotes, not both unless they're nested in which case the double quotes are on the outside and single quotes inside.

lib/site_template/README.txt
```diff
@@ -0,0 +1,23 @@
...
+"gem 'minima', '~> 2.0'" line in your 'Gemfile'
+To know the path to the installed theme-gem, simply run
+`bundle show minima` in your terminal. For more information on gem-based
+themes, goto: https://jekyllrb.com/docs/themes/
```
parkr: s/goto/go to/

parkr: You might also consider linking them as Markdown.

lib/site_template/README.txt
pathawks: :+1:

ashmaroli: I never explained it in the PR body, but there's a reason why I chose this route: :smile: 
 - This document was visioned to be a sort-of welcome-note to the user, instead of being a manual on the user's blog repo. They may choose to include a dedicated `README.md` for that purpose or convert this document to a markdown document, but that wouldn't be done by default.
 - A `README.md` on a Jekyll Site - Source repo usually contains only one line:
`"Source to my personal blog powered by Jekyll"`
 - A user needn't concern themselves with having to edit this file as a README is not so important for his Jekyll Blog Source Repo., especially, if it would be ignored by git and Jekyll. 

So I decided, its best that git and Jekyll ignore this document (`README.txt`)

lib/site_template/README.txt
ashmaroli: my bad.. will fix this and other instances..

ashmaroli:
I've updated the document content as requested. 
If you feel the points I listed in this [comment](https://github.com/jekyll/jekyll/pull/5602#discussion_r89946606) can be ignored, let me know. I'll convert the document to a Markdown file.

fene:
@ashmaroli could you convert the links to markdown-reference-style links? see this:https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links

lib/site_template/.gitignore
fene: i see no reason why we wouldn't want to commit this to a git repository too, many other site* generators do it

lib/site_template/README.txt
fene: hmm, i'm fine with it not being checked into git, but i'd like for it to be a markdown document, since that's basically the universal standard for readmes these days

lib/site_template/_config.yml
crispgm: I think we should convert it to `README.md` and I will put Jekyll badges in.

ashmaroli: guys, I opted for a `.txt` file because I did not want this file to be checked into the user's git index.
Since `README` holds a special value in repositories, I'm going to rename this file as `Welcome.txt`

@crispgm you're free to create a PR that initializes a proper `README.md` with `jekyll new`. Just that there will also be this `Welcome.txt` that'll be ignored by both Git and Jekyll, if this gets merged.

parkr:
I like the general idea of telling the user what's up – is it better to make another file many users won't open, or is it better to make this simply STDOUT when running `jekyll new`? The purpose seems to be filling in the user on the new way of doing things with theme gems, which could be said once per new site, perhaps a bit more concisely. Does that sound like a reasonable compromise? I'm not sure I see how this file fits into the natural flow, because it'd be so easy for users to ignore it. If we really want them to read it, let's give it to them as STDOUT. 👍 

What do you think?

ashmaroli:
> The purpose seems to be filling in the user on the new way of doing things with theme gems

yes, that is the intention.


> perhaps a bit more concisely.

I'm open to having the "content" edited.


> If we really want them to read it, let's give it to them as STDOUT.

I disagree with this suggestion though. STDOUT is volatile. Its great for simple messages like the `success_message` we display at the end of `jekyll new`, and to a lesser extent, for the `post_install_message` a new gem may display after install.
But when faced with a volley of information, (the `success_message`, the auto-initiation of `bundle install`, the consequent bundle output, I feel my main intention to **` provide the new user time to read and comprehend the situation at their pace and refer our online docs for further info. `** is compromised.
A more physical entity such as a file would be a great stand-in for that purpose.

quiltbag:
i agree with @ashmaroli on this. `jekyll new` already provides enough output as is, i don't think we need another extensive message. maybe we can just put something like `For more information, check the README.txt/md in <path>`?

DirtyF:
Yeah, we should lean to respect our user's limited attention.
What is the minimum of necessary information we should give after creating a new site?

- reassurance, minimal guidance
- invitation to preview the site
- invitation to add a post
- invitation to have a look at our docs to change or modify default's theme. (Let's explain this once and for all at a unique place)

Therefore I'll be more incline to something like : 

> Congrats, you've just setup a new Jekyll website! 👌 
> Now run `bundle exec jekyll serve`, open http://localhost:4000 in your browser, write your first post (https://jekyllrb.com/docs/posts/) and review your post before publishing.
> You want to change the current theme or learn how to tweak it? Take a look at our documentation (https://jekyllrb.com/docs/themes/), we got you covered.
> Happy Jekylling!

ashmaroli:
> Therefore I'll be more incline to something like : `[...]`

okay.. I get what you're suggesting. 
But nowhere in the copy you suggested reassure the new user that things have changed and there's no need to be alarmed about it. 
I can say from past experience that there's still a good probability that the confused user (who is actually following an outdated tutorial) simply logs in to Jekyll Talk and requests for assistance instead of checking the updated `Directory Structure` section and the `note info` just below it, in our docs or the page on customizing themes..

parkr:
If there is too much information for stdout then it belongs in our documentation on jekyllrb.com.

pathawks:
> If there is too much information for stdout then it belongs in our documentation on jekyllrb.com.

Agreed. This is a better way of communicating with the user than creating magic files that are ignored by Jekyll and Git.

DirtyF:
Friendly reminder: the best way to communicate with users is to give them early feedback, i.e. while they're interacting with your software, RTFM is not the only answer you can come up with.

A friend of mine used to say:
> If you read the docs, you're amongst the top 1% of developers

PR #5604, opened 2016-11-26
Update upgrading.md to mention usage of `bundle update`

merrua:
I've added the case for a simple Jekyll updates so beginners do not need to google for it, or ask.

### DIFF
--- a/docs/_docs/upgrading.md
+++ b/docs/_docs/upgrading.md
@@ -9,4 +9,5 @@ Upgrading from an older version of Jekyll? Upgrading to a new major version of J
 - [From 0.x to 1.x and 2.x](/docs/upgrading/0-to-2/)
 - [From 2.x to 3.x](/docs/upgrading/2-to-3/)
 
-If you are making a minor update (for example from 3.3.1  to 3.3.2 ) run 'gem update jekyll' when in your site directory. 
+If you are making a minor update (for example from 3.3.1 to the latest version at the time 3.3.2) run 'bundle update jekyll' when in your site directory. 
+If you would like to update all your gems, run 'bundle update' when in your site directory.
--- a/docs/_docs/upgrading.md
+++ b/docs/_docs/upgrading.md
@@ -8,3 +8,5 @@ Upgrading from an older version of Jekyll? Upgrading to a new major version of J
 
 - [From 0.x to 1.x and 2.x](/docs/upgrading/0-to-2/)
 - [From 2.x to 3.x](/docs/upgrading/2-to-3/)
+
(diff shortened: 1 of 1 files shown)


docs/_docs/upgrading.md
```diff
@@ -8,3 +8,5 @@ Upgrading from an older version of Jekyll? Upgrading to a new major version of J
 
 - [From 0.x to 1.x and 2.x](/docs/upgrading/0-to-2/)
 - [From 2.x to 3.x](/docs/upgrading/2-to-3/)
+
+If you are making a minor update (for example from 3.3.1  to 3.3.2 ) run 'gem update jekyll' when in your site directory.
```
DirtyF: ## Keeping Jekyll up to date

If you followed our setup recommendations and installed Bundler, run `bundle update jekyll` or simply `bundle update` and all your gems will update to the latest versions.
If you don't have Bundler installed, run `gem update jekyll`

fene: minor nit, but could you take out some of the spaces so that it looks like `from 3.3.1 to 3.3.2)`?

fene: whoops, i guess this becomes superfluous if you implement @DirtyF's suggestion

merrua:
Please take a look at the changed version that gives the bundler command.

docs/_docs/upgrading.md
```diff
@@ -8,3 +8,6 @@ Upgrading from an older version of Jekyll? Upgrading to a new major version of J
 
 - [From 0.x to 1.x and 2.x](/docs/upgrading/0-to-2/)
 - [From 2.x to 3.x](/docs/upgrading/2-to-3/)
+
+If you are making a minor update (for example from 3.3.1 to the latest version at the time 3.3.2) run 'bundle update jekyll' when in your site directory.
```
pathawks: I don't think this only applies when making a minor update. Depending on the `Gemfile`, this could also even update to a new _major_ version of Jekyll.

parkr: Correct – this applies to any updates. 😄

merrua: Do you have suggestted text, that would not suggest for anyone to try it for something like 3.9.7 to 4.0.0? I wanted the impression a new person should check in those cases without the need for them to understand semantic versioning already ? I don't want to relay on all breaking changes occuring with documentation. As most documentation tends to trail behind.

parkr:
Thank you! If someone thinks of a better way, I'm sure they'll come along and submit a PR sooner or later. For now, this is great.

@jekyllbot: merge +docs

Issue #3486, opened 2015-02-20
post_url does not like post name with parenthesis


tkrotoff:
Example:

```
{% post_url 2014-02-11-Prevention-et-gestion-des-troubles-musculosquelettiques-(TMS) %}
```

Result:

```
Deprecation: A call to '{{ post_url jekyll::tags::posturl }}' did not match a post using the new matching method of checking name (path-date-slug) equality. Please make sure that you change this tag to match the post's name exactly.
```

The HTML link generated by post_url works. Without the parenthesis around the word TMS => no deprecation warning.

Edit: FYI
- List of valid URL characters: [`ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~:/?#[]@!$&'()*+,;=`](http://stackoverflow.com/a/1547940/990356)
- List of invalid filename characters under Windows: [`\/:*?"<>|`](https://msdn.microsoft.com/en-us/library/aa365247)
- List of invalid filename characters under UNIX (depends on the filesystem): [`NUL (\0)` and `/`](http://en.wikipedia.org/wiki/Ext4)

parkr:
Hm. Not sure if we should support parentheses.

willnorris:
why not?  they're [perfectly valid](https://url.spec.whatwg.org/#url-writing).

alfredxing:
Wouldn't a simple [`Regexp.escape`](http://ruby-doc.org//core-2.2.0/Regexp.html#method-c-escape) work?

mattr-:
@alfredxing I think `URI.escape` might work better in this case. What do you think?

alfredxing:
I don't think that would work, since it doesn't seem to escape parentheses (which is what we need since we're turning the string into a regular expression).

parkr:
Regexp.escape? But really we should just be matching "post.name". "Are you strings equal??"

alfredxing:
We're matching using a Regex (https://github.com/jekyll/jekyll/blob/master/lib/jekyll/tags/post_url.rb#L13-L17), which is why the parens don't work (the parens are not escaped in `/#{slug}/`)

parkr:
No strong opinions on this one.

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least
three months.

The resources of the Jekyll team are limited, and so we are asking for your help.

[...]

PR #5605, opened 2016-11-26
Escaped regular expressions when using post_url.

Lunderberg:
Previously, the post_url function would give error messages when the post being listed contained special characters for use in regular expressions.  These special characters are now escaped using
`Regexp.escape`.

This is in reference to issue #3486.

### DIFF
--- a/lib/jekyll/tags/post_url.rb
+++ b/lib/jekyll/tags/post_url.rb
@@ -16,7 +16,7 @@ module Jekyll
 
         escaped_slug = Regexp.escape(slug)
         @name_regex = %r!^_posts/#{path}#{date}-#{escaped_slug}\.[^.]+|
-          ^#{path}_posts/?#{date}-#{slug}\.[^.]+!x
+          ^#{path}_posts/?#{date}-#{escaped_slug}\.[^.]+!x
       end
 
       def post_date
--- a/lib/jekyll/tags/post_url.rb
+++ b/lib/jekyll/tags/post_url.rb
@@ -14,7 +14,8 @@ module Jekyll
             "'#{name}' does not contain valid date and/or title."
         end
 
-        @name_regex = %r!^_posts/#{path}#{date}-#{slug}\.[^.]+|
+        escaped_slug = Regexp.escape(slug)
+        @name_regex = %r!^_posts/#{path}#{date}-#{escaped_slug}\.[^.]+|
           ^#{path}_posts/?#{date}-#{slug}\.[^.]+!x
       end
 
--- a/test/test_tags.rb
+++ b/test/test_tags.rb
@@ -575,7 +575,7 @@ CONTENT
       refute_match(%r!markdown\-html\-error!, @result)
     end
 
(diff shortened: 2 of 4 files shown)


parkr:
Hello! Thanks so much for the pull request. In order to evaluate this change and to prevent it from regressing in the future, we would like to see a unit test added to `test/test_tags.rb`. You should see examples for this tag in that file. Let us know if we can help!

lib/jekyll/tags/post_url.rb
```diff
@@ -14,7 +14,8 @@ def initialize(name)
             "'#{name}' does not contain valid date and/or title."
         end
 
-        @name_regex = %r!^_posts/#{path}#{date}-#{slug}\.[^.]+|
+        escaped_slug = Regexp.escape(slug)
+        @name_regex = %r!^_posts/#{path}#{date}-#{escaped_slug}\.[^.]+|
           ^#{path}_posts/?#{date}-#{slug}\.[^.]+!x
```
dotnil: Shouldn't the latter `slug` be escaped too?

Lunderberg: Thank you.  I had missed that one, and the updated pull request has both `slug`s escaped.

Lunderberg:
There is now a unit test in `test/test_tags.rb` to test the escaping of special characters.  The test succeeds with the change applied, and fails when the change is not present.  I haven't used ruby very much before, so it is mainly a copy-paste-modify of an existing test, with small modifications.

ashmaroli:
@Lunderberg please update this [test](https://github.com/jekyll/jekyll/blob/master/test/test_generated_site.rb#L14) as well to have our CI go green.. Thanks! :smiley:

test/test_tags.rb
```diff
@@ -554,6 +554,32 @@ def highlight_block_with_opts(options_string)
...
+        "destination" => dest_dir,
+        "read_posts"  => true
+      })
+    end
+
+    should "not cause an error" do
+      refute_match(%r!markdown\-html\-error!, @result)
+    end
+
+    should "have the URL to the \"complex\" post from 2008-11-21" do
```
ashmaroli: can single quotes be used here, (for cleanliness)?

Lunderberg: Good points, both changes have been made.

ashmaroli: :smile: I was actually asking Parker if that was allowed, but since you @Lunderberg implemented it anyways, can we switch the quotes to have it `"have the URL to the 'special-chars' post from 2016-11-26"` for homogeneity with other unit-tests.. also, there's more instances of `\""` inside **test names** within the same test-file.. can you patch that as well? Thanks

ashmaroli: *cleaning* other *unit-test-names* is perhaps beyond the scope of this pull.. so we'll tackle those in another PR..

parkr:
Wonderful work, thank you!

@jekyllbot: merge +bug

PR #5609, opened 2016-11-29
fix date parsing in file names to be stricter

### LINKED ISSUE #5603: Jekyll parses any dash-separated numeric filename as date
- [x] I updated to the latest Jekyll

---

- [x] I am on ***Debian/Ubuntu*** GNU/Linux

---

- [x] I was trying to build.

## My Reproduction Steps

1.  Create a collection, set default layout to page, e.g:

        collections:
          docs:
            output: true

2.  Create a file in the collection with a dash-separated filename in which the first three fields are numeric, e.g:

        84093135-42842323-42000001-b890-136270f7e5f1.md

3.  Try to build.

## The Output I Wanted

1.  Don't parse filenames as dates outside of the `posts` type scope.
2.  Build `84093135-42842323-42000001-b890-136270f7e5f1.html`

## The Output I Got

        Invalid date '84093135-42842323-42000001': Document '_m/work/84093135-42842323-42000001-b890-136270f7e5f1.md' does not have a valid date in the filename.

## Notes

*    Tested on Jekyll 3.3.1 and 3.0.1, the result is the same. 
*   If I change one of the characters in the first three fields (separated by `-`) in the filename to a letter, the site builds.
*   If I add a date in the frontmatter the site builds, however this collection has no use for dates.
*   The result is the same with `error_mode` set to `strict`, `warn` or `lax`.

## Workaround

*   The files in this collection are generated automatically by a service. I could get the other service to strip the dashes from the filename, and set the `permalink` in the frontmatter so that the rendered file ends up where it is supposed to end up (with the dashes), but this removes some of the semantics.

fene:
this fixes #5603


test/test_document.rb
```diff
@@ -492,4 +492,33 @@ def assert_equal_value(key, one, other)
       assert_equal true, File.file?(@dest_file)
     end
   end
+
+  context "a document in a collection with dash-separated numeric file name" do
```
fene: we don't really _need_ these tests because most others in this spec file will fail if the regex is wrong, but i added them for consistency

ashmaroli:
Before this pull, the expected `@site.posts.size` was `51` in [test/test_generated_site.rb](https://github.com/jekyll/jekyll/blob/master/test/test_generated_site.rb#L14)
With this pull, a new file was added to `/methods` collection ( albeit ignored by Jekyll ) yet `@site.posts.size` fell to a `49` ..
which other files got ignored??

fene:
@ashmaroli okay, that _is_ weird. i'll have a look

ashmaroli:
@fene these two files got ignored with this pull:
```
category/_posts/2008-9-23-categories.markdown
z_category/_posts/2008-9-23-categories.markdown
```
They were ignored because they have `9` instead of `09`..

This could potentially break some existing sites.. so a documentation-update or a deprecation-release is warranted..

fene:
ping @jekyll/core @jekyll/documentation what do you think about this?

ashmaroli:
I guess we're covered documentation-wise.

posts

and that section has been around for [years now](https://github.com/jekyll/jekyll/blob/cc73f04eb7c5c81b8a09e6a0802640c1c7cabcc0/site/_posts/2012-07-01-posts.md#creating-post-files)

fene:
^ weird travis error on the latest build

ashmaroli:
Don't worry about it.. its a bug.. occurs randomly..
it'll go away when one of the maintainers will restart that lone test for you..

parkr:
@jekyllbot: merge +bug

PR #5612, opened 2016-11-30
Add a module to re-define `ENV["TZ"]` in Windows

ashmaroli:
Finally managed to crack this issue.
With this in place, Jekyll on Windows will be able to reset `ENV["TZ"]` to the IANA TimeZone specified in the `config.yml`

### Points To Consider:
- Windows users need to have the following added to their Gemfile (and installed):
  - `gem "tzinfo"`
  - `gem "tzinfo-data"`
- A side-effect to this is that now **only valid IANA Timezones** will be accepted. Earlier Windows users were able to pass `UTC-04:00` etc as a workaround. Those sites will essentially require a re-configuring.

--
**Addresses:** #5414 
/cc @jekyll/windows


ashmaroli:
### DST & Failing Test..
- Travis which doesn't use this module **passes** all test leading to the inference that **`America/New_York`** on Travis is still at **`UTC-04:00`** even though DST is **not active** currently.
- AppVeyor which sets `TZ` using this module **fails** as it uses the correct timezone-offset at **`UTC-05:00`**

lib/jekyll/commands/new.rb
```diff
@@ -84,6 +84,10 @@ def gemfile_contents
 group :jekyll_plugins do
    gem "jekyll-feed", "~> 0.6"
 end
+
+# Windows does not include zoneinfo files, so bundle the tzinfo-data gem
+gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
```
parkr: How does this work when you don't have the gen installed?

ashmaroli:
@DirtyF I've added a small piece of documentation to go along with this pull.

parkr:
@jekyllbot: merge +bug

PR #5614, opened 2016-12-01
Update _font-awesome.scss

XhmikosR:
Move .woff file before .ttf so that it's preferred by supported browsers.

Missed this in #5530, sorry :/


parkr:
@jekyllbot: merge +site

PR #5615, opened 2016-12-01
lock minitest gem to v5.9

ashmaroli:
Our tests seem to break with the newly released `minitest gem v5.10.0`
This pull locks the gem we use to `v5.9.x`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -42,7 +42,7 @@ group :test_legacy do
   gem "simplecov"
   gem "minitest-reporters"
   gem "minitest-profile"
-  gem "minitest"
+  gem "minitest", "~> 5.9.1"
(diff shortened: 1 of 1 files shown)


Crunch09:
We should be able remove the lock again when [this](https://github.com/kern/minitest-reporters/pull/220) is merged and released.

zenspider:
I just released minitest 5.10.1 to patch this up until this gets fixed.

ashmaroli:
This pull is no longer necessary as the issue has been addressed upstream. Thanks @zenspider 
Leaving it open, though.. just in case.... :wink:

parkr:
Thanks, all! We should be good.

PR #5640, opened 2016-12-07
Write Jekyll::Utils::Exec.run for running shell commands.

### LINKED ISSUE #5463: `rake spec` (`script/test`) outputs bundle commands
I think somebody did something wrong.

parkr:
This fixes #5463 by taking @envygeeks's `run_in_shell` command and port it directly into the Jekyll codebase.

Any commands Jekyll has to run should be run with this method.

/cc @jekyll/core @ashmaroli

### DIFF
--- a/features/support/helpers.rb
+++ b/features/support/helpers.rb
@@ -102,8 +102,6 @@ def run_jekyll(args)
 end
 
 #
-
-# rubocop:disable Metrics/AbcSize
 def run_in_shell(*args)
   p, output = Jekyll::Utils::Exec.run(*args)
 
--- a/features/support/helpers.rb
+++ b/features/support/helpers.rb
@@ -1,6 +1,5 @@
 require "fileutils"
 require "jekyll"
-require "open3"
 require "time"
 require "safe_yaml/load"
 
@@ -106,21 +105,16 @@ end
 
 # rubocop:disable Metrics/AbcSize
 def run_in_shell(*args)
-  i, o, e, p = Open3.popen3(*args)
-  out = o.read.strip
-  err = e.read.strip
+  p, output = Jekyll::Utils::Exec.run(*args)
 
-  [i, o, e].each(&:close)
-
-  File.write(Paths.status_file, p.value.exitstatus)
+  File.write(Paths.status_file, p.exitstatus)
   File.open(Paths.output_file, "wb") do |f|
     f.puts "$ " << args.join(" ")
-    f.puts out
-    f.puts err
-    f.puts "EXIT STATUS: #{p.value.exitstatus}"
+    f.puts output
+    f.puts "EXIT STATUS: #{p.exitstatus}"
   end
 
-  p.value
+  p
 end
 # rubocop:enable Metrics/AbcSize
 
--- a/features/support/helpers.rb
+++ b/features/support/helpers.rb
@@ -114,7 +114,6 @@ def run_in_shell(*args)
 
   p
 end
-# rubocop:enable Metrics/AbcSize
 
 #
 
--- a/lib/jekyll/commands/new.rb
+++ b/lib/jekyll/commands/new.rb
@@ -133,10 +133,9 @@ RUBY
           Jekyll::External.require_with_graceful_fail "bundler"
           Jekyll.logger.info "Running bundle install in #{path.cyan}..."
           Dir.chdir(path) do
-            if ENV["CI"]
-              system("bundle", "install", "--quiet")
-            else
-              system("bundle", "install")
+            _, output = Jekyll::Utils::Exec.run("bundle", "install", "--quiet")
+            output.to_s.each_line do |line|
+              Jekyll.logger.info(line)
             end
           end
         end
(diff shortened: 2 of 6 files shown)


features/support/helpers.rb
```diff
@@ -106,21 +105,16 @@ def run_jekyll(args)
 
 # rubocop:disable Metrics/AbcSize
```
mattr-: can we remove the disabling of this check for this method now that it's been refactored?

lib/jekyll/commands/new.rb
```diff
@@ -133,10 +133,9 @@ def bundle_install(path)
           Jekyll::External.require_with_graceful_fail "bundler"
           Jekyll.logger.info "Running bundle install in #{path.cyan}..."
           Dir.chdir(path) do
-            if ENV["CI"]
-              system("bundle", "install", "--quiet")
-            else
-              system("bundle", "install")
+            _, output = Jekyll::Utils::Exec.run("bundle", "install", "--quiet")
```
mattr-: why do we not care about the return value here?

envygeeks: More importantly, why do you care about the output and not the status when you tell bundler to shut up while it's doing work?

parkr:
Updated.

lib/jekyll/commands/new.rb
```diff
@@ -121,23 +121,23 @@ def scaffold_path
...
-          Jekyll.logger.info "New jekyll site installed in #{path.cyan}."
-          Jekyll.logger.info "Bundle install skipped." if options["skip-bundle"]
-
           unless options["blank"] || options["skip-bundle"]
             bundle_install path
           end
+
+          Jekyll.logger.info "New jekyll site installed in #{path.cyan}."
+          Jekyll.logger.info "Bundle install skipped." if options["skip-bundle"]
         end
```
ashmaroli: out of curiosity, why was this moved to further ahead?
It was initially:
```
install directory > display success message > initiate 'bundle install'
```
now its:
```
install directory > initiate 'bundle install' > display install success message
```

lib/jekyll/commands/new.rb
```diff
@@ -121,23 +121,23 @@ def scaffold_path
...
           Jekyll::External.require_with_graceful_fail "bundler"
           Jekyll.logger.info "Running bundle install in #{path.cyan}..."
           Dir.chdir(path) do
-            if ENV["CI"]
-              system("bundle", "install", "--quiet")
-            else
-              system("bundle", "install")
+            process, output = Jekyll::Utils::Exec.run("bundle", "install")
+            output.to_s.each_line do |line|
+              Jekyll.logger.info("Bundler:".green, line.strip) unless line.to_s.empty?
```
ashmaroli: I'm always :+1:  for a little color out there.. but is `Bundler:` really necessary? (Though it doesn't hurt to be extra descriptive..)

lib/jekyll/commands/new.rb
ashmaroli: This change here would warrant a change to the test file [here](https://github.com/jekyll/jekyll/blob/master/test/test_new_command.rb#L44-L48)

lib/jekyll/commands/new.rb
pathawks: > but is `Bundler:` really necessary?

Yes. Nice to know where the output is coming from.

lib/jekyll/commands/new.rb
pathawks: Doesn't it seem like we should actually install Jekyll before we claim that Jekyll was installed?

We _could_ add an additional message before `bundle_install` that the directory was setup, but is it worth it?

ashmaroli: > Doesn't it seem like we should actually install Jekyll before we claim that Jekyll was installed?

*One wouldn't be able to run `jekyll new ` if Jekyll wasn't installed already.*  :grin:   
(I get it that was not your point.. :grinning: )  

By the time `install-success message` is displayed, the new project site would have already been created at the path specified.
The user is then informed of the initiation of the automated `bundle install` .. (which runs in context to the newly generated `Gemfile`)

But like I mentioned in the previous comment, if Parker decides to stick with his vision, he'll have edit the test file accordingly as well.

features/support/helpers.rb
```diff
@@ -103,24 +102,17 @@ def run_jekyll(args)
...
-    f.puts err
-    f.puts "EXIT STATUS: #{p.value.exitstatus}"
+    f.puts output
+    f.puts "EXIT STATUS: #{p.exitstatus}"
   end
 
-  p.value
+  p
 end
 # rubocop:enable Metrics/AbcSize
```
pathawks: Since we removed `rubocop:disable`, we could probably remove `rubocop:enable` as well

pathawks:
Updated the tests, as pointed out by @ashmaroli :+1:

lib/jekyll/commands/new.rb
DirtyF: Should not we colorize the output rather than `Bundler:`?

jekyll-new-output

ashmaroli: I like how it is in `Before`.. In `After`, I need to consciously ignore the `Bundler: ` .. but thats just my nit.....

parkr:
@jekyllbot: merge +minor

PR #5641, opened 2016-12-08
use backticks for Gemfile for consistency since in the next sentence …

daattali:
…_config.yml file has backtick


quiltbag:
@jekyllbot: merge +docs

PR #5660, opened 2016-12-15
add 'bundler' as a dependency

ashmaroli:
Technically, Bundler is only a `development_dependency` but specifying it as a `runtime_dependency` has it installed along with Jekyll.

This removes the need for 
```
$ gem install jekyll bundler
```
and possible errors otherwise:
```
~/core_ext/kernel_require.rb:55:in `require': cannot load such file -- bundler (LoadError)
```


ashmaroli:
An earlier [suggestion](https://github.com/jekyll/jekyll/pull/5237#issuecomment-239393692) was [turned down](https://github.com/jekyll/jekyll/pull/5237#issuecomment-239564097). 
(Requesting a proper clarification.)

/cc @jekyll/core

ashmaroli:
any updates on this?
/cc @parkr @pathawks

pathawks:
@ashmaroli I'm afraid this sort of decision is beyond my expertise.

parkr:
Please see https://github.com/jekyll/jekyll/issues/5684#issuecomment-290616936 – in short, the hard reliance on a `Gemfile` makes this impossible, as many users don't use `Gemfile`s. Thanks for the PR!

pathawks:
Currently, `jekyll new` fails spectacularly without `bundler` installed:

> ```
>   Dependency Error: Yikes! It looks like you don't have bundler or one of
> its dependencies installed. In order to use Jekyll as currently configured,
> you'll need to install this gem. The full error message from Ruby is:
> 'cannot load such file -- bundler' If you run into trouble, you can find
> helpful resources at https://jekyllrb.com/help/! 
> jekyll 3.5.2 | Error:  bundler
> ```

It seems like there must be _some_ better way to handle this.

ashmaroli:
we can perhaps disable automatic `bundle install` if the `LoadError` is raised..?

pathawks:
Closing this in favor of #6377

PR #5666, opened 2016-12-18
Rubocop -a on lib/jekyll

### LINKED ISSUE #5665: Jekyll has invalid formatting for deprecation message.
https://github.com/jekyll/jekyll/blame/master/lib/jekyll/collection.rb#L35
/cc @DirtyF

DirtyF:
Fix #5665

### DIFF
--- a/lib/jekyll/collection.rb
+++ b/lib/jekyll/collection.rb
@@ -32,8 +32,8 @@ module Jekyll
     # Override of method_missing to check in @data for the key.
     def method_missing(method, *args, &blck)
       if docs.respond_to?(method.to_sym)
-        Jekyll.logger.warn "Deprecation:", "#{label}.#{method} should be changed to" \
-          "#{label}.docs.#{method}."
+        Jekyll.logger.warn "Deprecation:",
+          "#{label}.#{method} should be changed to #{label}.docs.#{method}."
         Jekyll.logger.warn "", "Called by #{caller.first}."
         docs.public_send(method.to_sym, *args, &blck)
       else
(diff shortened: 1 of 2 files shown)


lib/jekyll/collection.rb
```diff
@@ -208,14 +209,15 @@ def read_document(full_path)
...
+
     def read_static_file(file_path, full_path)
       relative_dir = Jekyll.sanitized_path(
         relative_directory,
         File.dirname(file_path)
       ).chomp("/.")
 
       files << StaticFile.new(site, site.source, relative_dir,
-        File.basename(full_path), self)
+                              File.basename(full_path), self)
```
ashmaroli: whoa! rubocop did this?
may I suggest: ?
```ruby
files << StaticFile.new(
  site,
  site.source,
  relative_dir, 
  File.basename(full_path),
  self
)
```

lib/jekyll/excerpt.rb
```diff
@@ -30,8 +30,7 @@ def data
       @data
     end
 
-    def trigger_hooks(*)
-    end
+    def trigger_hooks(*); end
```
ashmaroli: jus curious. why do we have an empty method?

pathawks: So it can be overridden :+1:

ashmaroli:
noticed this just now.. PRs shouldn't have a `bug` label.. that's better for issue tickets. PRs should be sporting a `bug-fix` label instead.. :smiley:

DirtyF:
@jekyllbot: merge +dev

PR #5668, opened 2016-12-19
Use each instead of map to actually return nothing

kainjow:


quiltbag:
@jekyllbot: merge +bug

PR #5669, opened 2016-12-19
Remove unused _base argument

kainjow:


parkr:
Thanks! I don't think we can change the arity of our functions until Jekyll 4, as people lean on internal API's in all sorts of ways that we don't know.

I'm also curious about how this parameter came to be unused...

lib/jekyll/reader.rb
```diff
@@ -70,7 +70,7 @@ def retrieve_posts(dir)
     # dot_dirs - The Array of subdirectories in the dir.
     #
     # Returns nothing.
-    def retrieve_dirs(_base, dir, dot_dirs)
+    def retrieve_dirs(dir, dot_dirs)
```
pathawks: Sorry, no can do. These methods are not `private` which means this changes the method signature in a breaking way.

_If_ we want to do this, we need to wait for 4.0

lib/jekyll/reader.rb
```diff
@@ -47,7 +47,7 @@ def read_directories(dir = "")
       dot_static_files = dot_files - dot_pages
 
       retrieve_posts(dir)
-      retrieve_dirs(base, dir, dot_dirs)
+      retrieve_dirs(dir, dot_dirs)
```
pathawks: We can prefix this as `_base`, but removing it altogether changes the method signature.

parkr:
Gonna close this out as we can't modify the arity this function and I have no plans for a Jekyll 4.0.

parkr:
Thanks for the PR, though!

PR #5679, opened 2016-12-22
[liquid 4] Utils

pathawks:
As long as we are upgrading to Liquid 4, let's take advantage of some of the utility methods provided by Liquid.

In the long run, this should help our behavior and code more closely match that of Liquid core.

Note that this PR is against `liquid-4` rather than `master`.
#4362

 * [`to_integer `](https://github.com/Shopify/liquid/blob/v4.0.0/lib/liquid/utils.rb#L36-L44)
 * [`to_date`](https://github.com/Shopify/liquid/blob/v4.0.0/lib/liquid/utils.rb#L63-L81)

/cc @parkr

### DIFF
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -345,16 +345,9 @@ module Jekyll
 
     private
     def time(input)
-      case input
-      when Time
-        input.clone
-      when Date
-        input.to_time
-      when String
-        Time.parse(input) rescue Time.at(input.to_i)
-      when Numeric
-        Time.at(input)
-      else
+      begin
+        Liquid::Utils.to_date(input).to_time
+      rescue NoMethodError
         raise Errors::InvalidDateError,
           "Invalid Date: '#{input.inspect}' is not a valid datetime."
       end.localtime
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -347,12 +347,12 @@ module Jekyll
 
     private
     def time(input)
-      begin
-        Liquid::Utils.to_date(input).to_time
-      rescue NoMethodError
+      date = Liquid::Utils.to_date(input)
+      unless date.respond_to?(:strftime)
         raise Errors::InvalidDateError,
           "Invalid Date: '#{input.inspect}' is not a valid datetime."
-      end.localtime
+      end
+      date.to_time.localtime
     end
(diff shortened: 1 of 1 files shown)


pathawks:
You can see this convention used in [Liquid's own filters](https://github.com/Shopify/liquid/blob/v4.0.0/lib/liquid/standardfilters.rb#L63-L72):

```ruby
# Truncate a string down to x characters
def truncate(input, length = 50, truncate_string = "...".freeze)
  …
  length = Utils.to_integer(length)
  …
```

lib/jekyll/filters.rb
```diff
@@ -345,16 +347,9 @@ def sort_input(input, property, order)
...
-      when Date
-        input.to_time
-      when String
-        Time.parse(input) rescue Time.at(input.to_i)
-      when Numeric
-        Time.at(input)
-      else
+      begin
+        Liquid::Utils.to_date(input).to_time
+      rescue NoMethodError
```
parkr: I'd rather us check a return value of to_date than rescue such a generic error here.

pathawks: :+1:

lib/jekyll/filters.rb
```diff
@@ -345,19 +347,12 @@ def sort_input(input, property, order)
...
-        input.clone
-      when Date
-        input.to_time
-      when String
-        Time.parse(input) rescue Time.at(input.to_i)
-      when Numeric
-        Time.at(input)
-      else
+      date = Liquid::Utils.to_date(input)
+      unless date.respond_to?(:strftime)
```
Crunch09: Don't want to be nitpicking but could we instead check for `:to_time` ? Because ultimately that's what is going to be invoked on the `date` object.

pathawks: You are totally right. Good catch. 🍻

pathawks:
The date stuff could use a little work if we want to implement #5680.

Should I do that here, or in another PR?

parkr:
@pathawks I'd do that in a separate PR if you don't mind. Those changes don't depend on Liquid 4.

pathawks:
> @pathawks I'd do that in a separate PR if you don't mind.

Will do, boss 👍

pathawks:
Would you like me to squash?

Issue #5674, opened 2016-12-21
Allow site.post_attr_hash to process non-Enumerable attributes


CMessinides:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [ ] This is a feature request.

---

- [ ] I am on (or have tested on) ***macOS*** 10+
- [x] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux
- [ ] I am on (or have tested on) ***Fedora*** GNU/Linux
- [ ] I am on (or have tested on) ***Arch*** GNU/Linux
- [ ] I am on (or have tested on) ***Other*** GNU/Linux
- [ ] I am on (or have tested on) ***Windows*** 10+

---

- [ ] I was trying to install.
- [ ] There is a broken Plugin API.
- [ ] I had an error on GitHub Pages, and I have reproduced it locally.
- [ ] I had an error on GitHub Pages, and GitHub Support said it was a Jekyll Bug.
- [ ] I had an error on GitHub Pages and I did not test it locally.
- [x] I was trying to build.
- [ ] It was another bug.

## My Reproduction Steps

In any post in the `_posts` directory:

```markdown
---
author: Dr Jekyll
---
Content...
```

In `_plugins/plugin.rb`:

```ruby
index_by_author = site.post_attr_hash('author')
```

`jekyll build --trace` throws an error and the build fails:

```
      Generating... 
/var/lib/gems/2.3.0/gems/jekyll-3.3.1/lib/jekyll/site.rb:236:in `block in post_attr_hash': undefined method `each' for "Dr Jekyll":String (NoMethodError)
```
## The Output I Wanted
I would like the method `site.posts_attr_hash` to handle Strings, Numbers, and Booleans the same way it handles Arrays and Hashes. At the moment, `site.posts_attr_hash` applies `.each` to the attribute it receives, so Jekyll throws a NoMethodError and quits when that attribute is not Enumerable. In the above example, I would rather it output this when given a String attribute:

```ruby
{
  'Dr Jekyll' => <Jekyll::Document ...>
}
```

I believe it could accomplish that (without breaking dependent implementations in `site.tags` and `site.categories`) if it looked something like this:

```ruby
    def post_attr_hash(post_attr)
      # Build a hash map based on the specified post attribute ( post attr =>
[...]

PR #5685, opened 2016-12-24
Index Posts by any attribute

CMessinides:
In reference to issue #5674 

# Why I made this change
In the process of making a plugin, I wanted to index posts by attribute `author`, which was always a String. `site.post_attr_hash` seemed like the method to accomplish what I wanted, but instead it threw an error. This change expands the usefulness of `post_attr_hash` to all attribute types while preserving its original function so `site.tags` and `site.categories` work as they already did.

# What impact it will have
The commented documentation of this method reads:
```ruby
# Construct a Hash of Posts indexed by the specified Post attribute.
    #
    # post_attr - The String name of the Post attribute.
...
 # Returns the Hash: { attr => posts } where
    #   attr  - One of the values for the requested attribute.
    #   posts - The Array of Posts with the given attr value.
```
The documentation does not mention that `post_attr` must respond to `.each` for the method to work. I feel this is misleading for users, in particular plugin authors. This change will make the method function as the documentation says it will. In addition, broadening the functionality of this method will help plugin authors create helpers and generators that, for instance, create lists of posts indexed by a Boolean attribute (ex. `featured`) or String attribute (ex. `author`) with a built-in method rather than having to roll their own.

# What it does
`site.post_attr_hash` now checks the value of parameter `post_attr` to see if it responds to `.each` or not. If it does, the method works as it did before. If not, it uses the value of `p.data[post_attr]` as the key in the `hash` that the method will return.

### DIFF
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -234,10 +234,10 @@ module Jekyll
       hash = Hash.new { |h, key| h[key] = [] }
       posts.docs.each do |p|
         attr_value = p.data[post_attr]
-        if attr_value.respond_to? :each 
+        if attr_value.respond_to? :each
           attr_value.each { |t| hash[t] << p }
         elsif !attr_value.nil?
-          hash[attr_value] << p            
+          hash[attr_value] << p
         end
       end
       hash.values.each { |posts| posts.sort!.reverse! }
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -233,7 +233,12 @@ module Jekyll
       # array of posts ) then sort each array in reverse order.
       hash = Hash.new { |h, key| h[key] = [] }
       posts.docs.each do |p|
-        p.data[post_attr].each { |t| hash[t] << p } if p.data[post_attr]
+        attr_value = p.data[post_attr]
+        if attr_value.respond_to? :each 
+          attr_value.each { |t| hash[t] << p }
+        elsif !attr_value.nil?
+          hash[attr_value] << p            
+        end
       end
       hash.values.each { |posts| posts.sort!.reverse! }
       hash


ashmaroli:
I would recommend that you add a test to prove that the changes work as you intended..

benbalter:
If you wanted a list of post authors, could you not use `site.posts.map { |p| p.data["author"] }`?

CMessinides:
@benbalter The behavior I wanted to add to this method was more along the lines of `site.posts.group_by { |p| p.data['author'] }`. That's roughly what I expected the method to do when I read its documentation. I was confused when it raised a NoMethodError.

But, using the `group_by` approach (which I didn't know about when I made this pull request) is a valid solution, and probably the better one. In hindsight, this change isn't as vital (or efficient) as I thought it was. Considering how simple the alternative is, I wouldn't have any complaints if we closed this PR.

lib/jekyll/site.rb
```diff
@@ -237,7 +237,12 @@ def post_attr_hash(post_attr)
       # array of posts ) then sort each array in reverse order.
       hash = Hash.new { |h, key| h[key] = [] }
       posts.docs.each do |p|
-        p.data[post_attr].each { |t| hash[t] << p } if p.data[post_attr]
+        attr_value = p.data[post_attr]
+        if attr_value.respond_to? :each
+          attr_value.each { |t| hash[t] << p }
+        elsif !attr_value.nil?
+          hash[attr_value] << p
```
parkr: I think you can use `Kernel#Array` here! It just wraps it in an array in case it's not already an array.

```ruby
Array(p.data[post_attr]).each { |attribute| hash[attribute] << p } if p.data[post_attr]
```

nebulousdog:
this would be great! @parkr is the only holdup your recommendation [here](https://github.com/jekyll/jekyll/pull/5685#discussion_r131310717)? if so, let's refresh the PR :)

nebulousdog:
I could use some advice on how to get `jekyll/jekyll` as a base branch to make a PR to. I tried again here https://github.com/nebulousdog/jekyll/pull/1 but failed.

DirtyF:
@nebulousdog thanks for the reminder, I rebased the branch and applied @parkr recommendation

ashmaroli:
Updated it a bit to conform with our present style-guide.. and earmarked it for `v4.0.0` so that we don't forget about this again.. :wink:

PR #5687, opened 2016-12-25
Test against Ruby 2.4.0

ashmaroli:
Test against the latest version of Ruby: v2.4.0


.travis.yml
```diff
@@ -5,9 +5,10 @@ language: ruby
 sudo: false
 
 rvm:
-  - &ruby1 2.3.1
-  - &ruby2 2.2.5
-  - &ruby3 2.1.9
+  - &ruby1 2.4.0
+  - &ruby2 2.3.1
+  - &ruby3 2.2.5
+  - &ruby4 2.1.9
```
yous: How about bumping previous versions? Current versions are 2.3.3, 2.2.6, and 2.1.10. https://www.ruby-lang.org/en/downloads/

ashmaroli: that's for a separate PR. This PR is only about Ruby 2.4.  
You're welcome to create a PR to bump the rvm versions we currently test against.  
:smiley:

parkr:
It looks like the pygments.rb update is causing issues. Does pygments.rb 1.0 work on Ruby 2.1, 2.2, and 2.3, too? If it does, then maybe we should just update!

Gemfile
```diff
@@ -3,6 +3,13 @@ gemspec :name => "jekyll"
 
 gem "rake", "~> 12.0"
 
+if RUBY_VERSION >= '2.4'
+  gem "json", "~> 2.0"
+  gem "pygments.rb", "~> 1.1"
+else
+  gem "pygments.rb", "~> 0.6.0" unless RUBY_ENGINE == "jruby"
```
parkr: Can we use pygments.rb v1 with Ruby 2.1,2,3,4 instead of using 2 versions?

ashmaroli: we can.. but I wasn't sure if the change would affect Jekyll internally in any way..
I pushed a new commit to use just one version of `pygments.rb`

From the logs, it looks like we just need to change our assertions to match the new pygments syntax. But I don't want to undertake that task.

So, you're free to close this PR or assign someone to directly modify this branch.

parkr:
Thank you! We should definitely upgrade the pygments.rb dependency as you already did, but it has caused a number of test failures. It would be preferable to see the upgrade & test failures fix in a separate PR so we can keep this focused on the 2.4 upgrade.

It also occurs to me that we should check 2.4 on AppVeyor as well! If that's a possibility, it would be nice to keep Travis & AppVeyor in harmony.

ashmaroli:
I think we can't test with `Ruby 2.4.0` on AppVeyor yet as the RubyInstaller project has not yet released a corresponding binary for Windows.

/cc @XhmikosR

XhmikosR:
Yeah, not possible. We need 2.4.0, but see https://github.com/oneclick/rubyinstaller/issues/348

parkr:
@ashmaroli @XhmikosR That settles it then! 👍  We'll wait on rubyinstaller for AppVeyor support and in the meantime we'll hope that the 2.4.0 tests on unix-like platforms will work for any brave souls who install Ruby 2.4.0 on their Windows machines.

ashmaroli:
From the look of the test failures, it looks like the only thing that going to break is the pygments HTML output.. doesn't seem to be a serious issue, IMO

parkr:
> From the look of the test failures, it looks like the only thing that going to break is the pygments HTML output.. doesn't seem to be a serious issue, IMO

Yeah, agreed. I would like to contain that change in one separate PR so we can always trace back. Feel free to simply fix the tests so the expectations match! The HTML that pygments outputs seems good enough to me.

parkr:
Thank you!

PR #5630, opened 2016-12-04
Updates to permalink documentation

tomjohnson1492:
I reviewed the existing permalink documentation and made some updates to improve the clarity and accuracy. There are several pages that I updated in this pull request:

**permalinks.** I heavily edited this page to be more thorough and cover some more scenarios. I don't think can just get a sense of my updates by reading the side-by-side comparison in the Github interface. I think you need to download the patch branch and build it locally to see the whole context.

**collections.** I'm not sure if permalink details for collections belongs in the Collections topic or Permalinks topic, but I left it in Collections for now. I included a number of scenarios to clarify permalinks. Some details were just confusing (which you can see by reading this [Jekyll talk thread](https://talk.jekyllrb.com/t/how-can-i-change-the-filename-extension-from-html-to-xhtml/3341/8)).

**pages.** Just made a few updates. One detail I think was outdated in the existed docs is the practice of manually inserting index.html files inside of folders. Using the right permalink styles makes this unnecessary. Really the update that started this whole pull request was the ability to flatten pages in root. I spent months trying all kinds of workarounds until I stumbled across this technique. 

(Overall, too much of the Jekyll documentation focuses on blogging scenarios. There is a huge trend of using Jekyll for documentation sites, so the emphasis on pages needs expansion. With doc sites, users don't just have a few pages. They potentially have hundreds, and the workflow is different. You often submit content to an internal environment before pushing it out live. In my doc contributions, I want to help expand the doc site scenario.)

**templates.** No new updates here -- this is just the same content I changed and submitted in my previous pull request (https://github.com/jekyll/jekyll/pull/5610). If you process that pull request first, this should not show any conflicts. Or actually, you might want to close the previous pull request and just process this request. I provided more detail about using includes in a more advanced way. I tried to explain includes, especially passing parameters to includes, with more clarity.

**quickstart.** I updated the quickstart page to provide more information about all the missing files and directories that people have been wondering about ever since the introduction of gem-based themes. I feel we really need this detail clarified in the quick start. I don't know how many questions I've seen in the forum where people ask where these files are, but it's a frequent question.

Also, I think given that Github Pages supports only Minima, there needs to be more emphasis on installing non-gem-based themes as well. 

I also explain what the Gemfile and Gemfile lock, as well as bundle exec commands, actually do, since these now come into play more prominently with gem-based themes.

**.gitignore.** I use WebStorm to work with Jekyll. I need to omit the .idea files that WebStorm generates to index the content.

[...]

PR #5696, opened 2016-12-26
Added new includes.md topic to docs

tomjohnson1492:
Added new includes.md topic.

See https://github.com/jekyll/jekyll/pull/5630 for more details on the update. 

@jekyll/documentation

### DIFF
--- a/docs/_docs/includes.md
+++ b/docs/_docs/includes.md
@@ -32,7 +32,7 @@ such as variables.
 
 ### Using variables names for the include file
 
-The name of the file you want to embed can be specified as a variable instead of an actual file name. For example, suppose you defined a variable in your page's frontmatter like this:
+The name of the file you want to embed can be specified as a variable instead of an actual file name. For example, suppose you defined a variable in your page's front matter like this:
 
 ```yaml
 ---
@@ -44,7 +44,7 @@ my_variable: footer_company_a.html
 You could then reference that variable in your include:
 
 ```liquid
-{% raw %}{% include {{ page.my_variable }} %}{% endraw %}
+{% raw %}{% include {{page.my_variable}} %}{% endraw %}
 ```
 
 In this example, the include would insert the file `footer_company_a.html` from the `_includes/footer_company_a.html` directory.
@@ -56,17 +56,17 @@ You can also pass parameters to an include. For example, suppose you have a file
 ```liquid
 {% raw %}<div markdown="span" class="alert alert-info" role="alert">
 <i class="fa fa-info-circle"></i> <b>Note:</b>
-{{ include.content }}
+{{include.content}}
 </div>{% endraw %}
 ```
 
-The `{% raw %}{{ include.content }}{% endraw %}` is a parameter gets populated when you call the include and specify a value for that parameter, like this:
+The `{% raw %}{{include.content}}{% endraw %}` is a parameter that gets populated when you call the include and specify a value for that parameter, like this:
 
 ```liquid
 {% raw %}{% include note.html content="This is my sample note." %} {% endraw %}
 ```
 
-The value of `content` (which is `This is my sample note`) will be inserted into the {% raw %}`{{ include.content }}`{% endraw %} parameter.
+The value of `content` (which is `This is my sample note`) will be inserted into the {% raw %}`{{include.content}}`{% endraw %} parameter.
 
 Passing parameters to includes is especially helpful when you want to hide away complex formatting from your Markdown content.
 
@@ -85,10 +85,10 @@ You could templatize this content in your include and make each value available 
 
 ```liquid
 {% raw %}<figure>
-   <a href="{{ include.url }}">
-   <img src="{{ include.file }}" style="max-width: {{ include.max-width }};"
-      alt="{{ include.alt }}"/>
-   <figcaption>{{ include.caption }}</figcaption>
+   <a href="{{include.url}}">
+   <img src="{{include.file}}" style="max-width: {{include.max-width}};"
+      alt="{{include.alt}}"/>
+   <figcaption>{{include.caption}}</figcaption>
 </figure>{% endraw %}
 ```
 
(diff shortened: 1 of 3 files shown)


docs/_docs/includes.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Passing parameters to includes
+
+You can also pass parameters to an include. For example, suppose you have a file called `note.html` in your `_includes` folder that contains this formatting:
+
+```liquid
+{% raw %}<div markdown="span" class="alert alert-info" role="alert">
+<i class="fa fa-info-circle"></i> <b>Note:</b> {{include.content}}
+</div>{% endraw %}
+```
```
ashmaroli: update to the comment above:
**`"HTML can split to a newline for better readability:"`**
```liquid
{% raw %}
  <div markdown="span" class="alert alert-info" role="alert">
    <i class="fa fa-info-circle"></i> 
    <b>Note:</b> 
    {{ include.content }}
  </div>
{% endraw %}
```
once Jekyll migrates to Liquid v4, we'll use its new whitespace-control syntax to remove the resulting clear lines from  tags.

tomjohnson1492: I can split up the HTML onto separate lines, but if you put the raw tags on their own lines, they create a blank space there, which makes the pre formatting look like it has a bunch of empty space at the top and bottom.

tomjohnson1492: re the space around tags such as `{{ include.content }}`, I made this update. At some point, i'd like to create a style guide for the site that defines conventions like these.

docs/_docs/includes.md
```diff
@@ -0,0 +1,175 @@
...
+
+The {% raw %}`{{include.content}}`{% endraw %} is a parameter gets populated when you call the include and specify a value for that parameter, like this:
```
tomjohnson1492: updated

tomjohnson1492:
I made all the updates requested on this topic, so you can re-review and potentially merge the patch now.

docs/_docs/includes.md
ashmaroli: > but if you put the raw tags on their own lines, they create a blank space there, which makes the pre formatting look like it has a bunch of empty space at the top and bottom.

Hence the reason why i mentioned about Liquid 4's ability to remove that whitespace..

ashmaroli: > At some point, i'd like to create a style guide for the site that defines conventions like these.

Ben has already set of guidelines defined at https://github.com/benbalter/jekyll-style-guide. Add to it if not already defined..

docs/_docs/includes.md
```diff
@@ -97,15 +103,18 @@ This include contains 5 parameters:
...
-</figcaption>{% endif %}</figure>{% endraw %}
+{% raw %}<figure>
+   {% if {{ include.url }} %}<a href="{{ include.url }}">{% endif %}
+   <img src="{{ include.file }}" {% if {{ include.max-width }} %}
+   style="max-width: {{ include.max-width }}" {% endif %}
+   alt="{{ include.alt }}" />{% if {{ include.url }} %}</a>{% endif %}
+   {% if {{ include.caption }} %}
+   <figcaption>{{ include.caption }}</figcaption>
+   {% endif %}
+</figure>{% endraw %}
```
ashmaroli: I'm having a hard-time reading the code here (slight exaggeration).. consider indenting lines as required..

you cant have `{{ }}` inside `{% %}` have you tested it locally?

docs/_docs/includes.md
```diff
@@ -55,35 +55,41 @@ You can also pass parameters to an include. For example, suppose you have a file
...
 ```liquid
 {% raw %}<div markdown="span" class="alert alert-info" role="alert">
-<i class="fa fa-info-circle"></i> <b>Note:</b> {{include.content}}
+<i class="fa fa-info-circle"></i> <b>Note:</b>
+{{ include.content }}
 </div>{% endraw %}
 ```
 
-The {% raw %}`{{include.content}}`{% endraw %} is a parameter gets populated when you call the include and specify a value for that parameter, like this:
+The `{% raw %}{{ include.content }}{% endraw %}` is a parameter gets populated when you call the include and specify a value for that parameter, like this:
```
ashmaroli: I made a mis-judgement regarding this the last time.. please revert it..
{% raw %}`{{ include.content }}`{% endraw %} and adjust all other instances for consistency, (if any)

tomjohnson1492: can you remind me what the misjudgement was? I simply forgot to put the raw tags inside the backticks. it renders fine.

docs/_docs/includes.md
tomjohnson1492: ahh, now we're coming into something interesting territory. I tested the code locally and have been using includes like this for months in my doc projects. But I didn't have spaces around the variables. It turns out this spacing matters when the variable appears inside an if statement.

This works:

```
<figure>
   {% if {{include.url}} %}<a href="{{include.url}}">{% endif %}
   <img src="{{include.file}}" {% if {{include.max-width}} %}
   style="max-width: {{include.max-width }}" {% endif %}
   alt="{{include.alt}}" />{% if {{include.url}} %}</a>{% endif %}
   {% if {{include.caption}} %}
   <figcaption>{{include.caption}}</figcaption>
   {% endif %}
</figure>
```

This doesn't:

```
<figure>
   {% if {{ include.url }} %}<a href="{{ include.url }}">{% endif %}
   <img src="{{ include.file }}" {% if {{ include.max-width }} %}
   style="max-width: {{ include.max-width }}" {% endif %}
   alt="{{ include.alt }}" />{% if {{ include.url }} %}</a>{% endif %}
   {% if {{ include.caption }} %}
   <figcaption>{{ include.caption }}</figcaption>
   {% endif %}
</figure>
```

The difference is with the spacing in the page variables. 

I was under the impression that `{% if {{include.url}} %}` and `{% if {{ include.url }} %}` would be rendered the same, since having the include outside of an `{% if ... %}` doesn't impose the same requirements about spacing. Outside of if statements, `{{ include.file }}` and `{{include.file}}` work equally well. 

I can remove the spacing for variables inside of `if` statements in the code and add a little explanatory note. Is this a bug in Jekyll, or a feature?

tomjohnson1492:
I updated the doc to remove the problematic spacing. Submitting for re-review.

docs/_docs/includes.md
ashmaroli: never mind..

docs/_docs/includes.md
ashmaroli: removing the space around variables may "work" within `{% if }` blocks but Liquid should be issuing a warning it's wrong nonetheless..
I'm not in favor of Jekyll's official documentation stating this.

/cc @pathawks  your opinion?

pathawks:
I have a request.

When we are reviewing, can we try really hard to perform one thorough review, and then approve once all suggested changes have been made? I recognize that this is more work for the reviewer, but @tomjohnson1492 has put a lot of work in improving our documentation, and I don't want those changes to be lost to bureaucracy.

Do point out things that must be changed before the PR can be merged, but also recognize that things can be further refined in future pull requests. There are some much needed improvements in these pull requests, and I would like to see them land in our documentation as soon as they can be reviewed.

ashmaroli:
> I have a request.

okay @pathawks I'll make sure I'll do a thorough scrutiny before submitting the review

docs/_docs/includes.md
tomjohnson1492: I think it would be best if I just removed the section about conditionalizing the includes. I can take out this part: "You could also use `if` tags to safeguard scenarios where either the parameter is optional or where the author doesn't include the parameter:....."

I'm guessing the correct approach is to use the `default` filter, right?

If this really should be an error from Liquid, then yeah, the jekyll docs shouldn't promote bad code. 

I'll also omit the paragraph I added about spacing versus non-spacing and errors.

tomjohnson1492:
I made 2 fixes here: 

- removed the erroneous liquid code and hack note
- added back the spacing inside curly braces

Submitting for review.

ashmaroli:
Since I do not have write-access to the repo, my reviews should not block the merge :innocent:

DirtyF:
@jekyllbot: merge +docs

pathawks:
:tada:

Issue #4227, opened 2015-12-05
Consolidate caches and create a _cache folder for *all* to use.


envygeeks:
Today we did an audit for our biggest Jekyll site and noticed it has like 8 cache folders, this is pretty big and unwieldy, it also makes it harder for us to control the caching on our build servers.  We should create a `_cache` folder and consolidate all Jekyll caching into that folder, and then ask plugin authors to place their caches in that folder too with or without prefixes.  Exp:
- `_cache/jekyll/incremental`
- `_cache/jekyll-assets/assets`
- `_cache/jekyll/sass`

Now cleanup went from "let me figure out what is what" to `rm -rf _cache`

parkr:
Let's add the removal of these caches to the `jekyll clean` command.

alfredxing:
:+1: Good idea! It'd be easier for users in terms of ignoring these files for source control as well.

parkr:
Are any sites using `_cache` for content? That's what worries me here.

envygeeks:
@parkr make it configurable then, out of all the configuration options that a system should have, whether it caches is a high-priority unless it uses system caches, and especially if it uses local directory caching.  We allow it on Jekyll-Assets which gives users flexibility on where they place their cache and how, it's not a bad idea for Jekyll and I don't know the YACO applies here because for people like me who build hundreds of sites putting our caches in a global directory and controlling them gives us much more flexibility in our build systems, how we build all that, and it gives the average user more flexibility in how dirty their folder is at the root.

envygeeks:
Reopening at the request of @nhoizey
/cc @parkr @jekyll/core @jekyll/ecosystem

envygeeks:
Oops, thought I did reopen it, thanks for doing it @parkr.

mattr-:
:+1: for this idea from me.

pathawks:
I very much approve of this idea :+1:

> Are any sites using _cache for content?

Yeah, how would this work? Don’t want anybody to upgrade without reading the docs and end up having some of their data "cleaned" away by mistake.

envygeeks:
I think we would have to do a pre-4.0 pre-flight warning, to let people we know that we are implementing a new folder and that it could be destroyed by "jekyll clean" but we could always use `.jekyll-cache` or something like that which in theory shouldn't conflict and if it does I'm okay with that the latter folder conflicting because one would expect any `.jekyll` prefixed folder to be reserved.

envygeeks:
[...]

PR #5701, opened 2016-12-28
Fix bug in 'jekyll clean' regarding .sass-cache

stepchowfun:
Hello! This was a quick fix so I didn't file an issue, but I'm happy to if needed.

The bug is that SASS creates a temporary folder called `.sass-cache` in the _current_ directory (regardless of the specified `source`), but `jekyll clean` incorrectly removes it from `source` (regardless of the current directory).

This small change fixes the bug. I tested it on my machine. I'm using Jekyll 3.3.1 on macOS Sierra (x86_64) with Ruby 2.3.0.

Related issue: https://github.com/jekyll/jekyll/issues/4649
Related PR: https://github.com/jekyll/jekyll/pull/4652 (introduced the bug)

Happy holidays!

### DIFF
--- a/lib/jekyll/commands/clean.rb
+++ b/lib/jekyll/commands/clean.rb
@@ -20,7 +20,7 @@ module Jekyll
           options = configuration_from_options(options)
           destination = options["destination"]
           metadata_file = File.join(options["source"], ".jekyll-metadata")
-          sass_cache = File.join(options["source"], ".sass-cache")
+          sass_cache = ".sass-cache"
 
           remove(destination, :checker_func => :directory?)
           remove(metadata_file, :checker_func => :file?)


ashmaroli:
I think its not a bug with `jekyll clean` that it cleans only `.sass-cache` at `source` but rather with `.sass-cache` being created at current-directory irrespective of the `source` config

/cc @pathawks

pathawks:
> I think its not a bug with jekyll clean that it cleans only .sass-cache at source but rather with .sass-cache being created at current-directory irrespective of the source config

Yup.

See also #4227

stepchowfun:
Makes sense. I'll go ahead and close this PR, assuming https://github.com/jekyll/jekyll/issues/4227 is also addressing it.

PR #5708, opened 2016-12-30
Update Classifier-Reborn

Ch4s3:
There are some serious improvements between 2.0 and 2.0.5, and no breaking changes, so I might be good to bump this up.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -73,7 +73,7 @@ group :jekyll_optional_dependencies do
   gem "toml", "~> 0.1.0"
 
   platform :ruby, :mswin, :mingw, :x64_mingw do
-    gem "classifier-reborn", "~> 2.0"
+    gem "classifier-reborn", "~> 2.0.5"
     gem "liquid-c", "~> 3.0"
     gem "pygments.rb", "~> 0.6.0"
     gem "rdiscount", "~> 2.0"


quiltbag:
@Ch4s3 looks like the build failed because of said update. could you look into that?

ashmaroli:
@Ch4s3 all of the passing tests in classifier-reborn's repo report just one thing:
```
warning: global variable `$GSL' not initialized
```

Ch4s3:
@quiltbag It should be easy to fix. It looks like the failures are hapening in [test_related_posts.rb](https://github.com/jekyll/jekyll/blob/master/test/test_related_posts.rb) THis particular failure happens when you try to compare posts that only have words like "a and the", or posts that are only a few words long. It used to just fail silently.

I could update the posts on the fixture site to be more realistic... I see it referenced [here](https://github.com/jekyll/jekyll/blob/master/test/test_related_posts.rb#L6) but I can't find the file.

@ashmaroli That just means that the c library for GSL isn't installed and the Ruby lib is being used. would it be better for you folks if this didn't throw a warning?

Ch4s3:
Looks like fixture_sites are setup [here](https://github.com/jekyll/jekyll/blob/57fd5f887da1189a16bdfbb982d75f725c38d725/script/console) but I'm not sure how to view the fixture posts.

Ch4s3:
It looks like the test posts only contain `url: {{ page.url }}\ndate: {{ page.date }}\nid: {{ page.id }}` which would never have worked with LSI. I'm a little surprised this never caused a problem before.

Ch4s3:
I should probably update [jekyll/related_posts.rb](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/related_posts.rb) to account for this as well.

Ch4s3:
If someone could tell me how to add meaningful posts to the test, I could have this fixed pretty quickly.

parkr:
Sounds good to me!

Ch4s3:
Aren't you on vacation Parker? ;)

Ch4s3:
closing in favor of #5711

PR #5720, opened 2017-01-05
Document --unpublished build option

sskylar:
The `--unpublished` build option ([found here](https://github.com/jekyll/jekyll/blob/57fd5f887da1189a16bdfbb982d75f725c38d725/lib/jekyll/command.rb#L63-L64)) was missing from the [docs](https://jekyllrb.com/docs/configuration/#build-command-options).

This PR fixes that.


DirtyF:
@jekyllbot: merge +docs

Issue #1082, opened 2013-05-11
Suggestion: Read metadata in separate files.


sequitur:
Hakyll currently reads metadata from separate files, i.e. a file called `foo.md` would be stapled to the YAML front matter of `foo.md.metadata` in the same directory.

Having this in Jekyll would be helpful both for projects building their content in more formats than one (E.g., HTML via Jekyll and PDF via Pandoc). It also makes it possible to attach metadata to non-text files, which may or may not be useful at present, but would be helpful for future features or plugins.

Ajedi32:
:+1: for this. Unless I'm misunderstanding here, it's currently impossible to use standard markdown-formatted files with Jekyll and have them render properly (same for other file formats too). Unless you include weird, Jekyll-specific metadata at the top of all your files, Jekyll refuses to process them. That's kind of a deal breaker for me at the moment.

micapam:
Me too, @Ajedi32. I'm going to use nanoc for now, but I'd rather use Jekyll if possible as there seems to be wider community support etc. So I'll watch this thread.

My thoughts on the 'simplicity' comments above: there is more than one kind of simplicity. I like Markdown because it's minimal, and system-agnostic. Simple for the user != simple for the system.

parkr:
Are any of you using a plugin for this? What does it look like? I think we'll ship Jekyll 3 without this and revisit later. I have not heard anything for a while so it's becoming lower priority.

Ajedi32:
A plugin would be great. Personally, my interest has shifted to [Metalsmith](http://www.metalsmith.io/) though so I don't really have the motivation to write one myself at the moment (especially given that I'm not very familiar with Jekyll or its plugin system).

If there is a plugin which implements this functionality though I'd certainly love to hear about it.

micapam:
@parkr Beware self-selection bias. You not be hearing much about it because those people for whom it's a deal-breaker are going elsewhere. I considered Jekyll but decided on nanoc for this very reason.

jooize:
I'm using [Hardwired](https://github.com/nathanaeljones/hardwired) much because of its support for omitting the YAML “---” and allowing capital letters in the front matter tags “Layout: legal-page”.

envygeeks:
[...]

PR #5733, opened 2017-01-07
Add support for detached front matter (optional)

darwin:
Say you have $site/some/path/page.md. With this patch you can optionally
create $site/some/path/page.md.fm.yml (or .yaml) and it will be
recognized as a "detached YAML front matter". That means that data found
in this yaml file will be merged with data in normal inlined front
matter (inlined data override detached data).

Why?
My IntelliJ IDE is too "advanced" and cannot cope with YAML front
matters in various files. For example with .md , .scss or even .html it
breaks inspections[1] and bugs me about invalid syntax all day. This
allows me to remove inlined front matters and put them into separate
files and stop fighting with IntelliJ.

Why <page>.fm.yml?
This is by convention. File name prefix must match the associated
page file name. I don't want to introduce a clash with potentially
existing yaml files in exiting sites. And I need a way how
to detect these special yaml files and filter them out from site output.
Also the extension must be yml or yaml for IDEs to treat it as a proper
yaml file.

Notes:
1. this should be transparent to existing Jekyll users
2. interested users may incrementally move their inlined front matters
   into detached files as needed

[1] https://www.jetbrains.com/help/idea/2016.3/code-inspection.html


DirtyF:
Interesting. IMHO that looks like a edge case to me and more related to IntelliJ than Jekyll itself, that's the reason why I would rather suggest you implement this in a plugin.

/cc @jekyll/core

darwin:
This wasn't a problem for me when using TextMate or Sublime (5+ years using jekyll). Because front matter just confused syntax highlighting at the beginning of files. No big deal. But when I recently switched to IntelliJ with most of my web-dev work this started to be a real PITA, because additionally I rely on inspections being "clean" for my whole project. I guess most people use text editors with jekyll so they don't have issue with inlined front matters. And people using IDEs with deep introspections didn't care enough to fix this (AFAIK).

I don't see an easy way how to implement something like this with a plugin. And even if it was possible it would definitely involve heavy monkey-patching. btw. I have implemented [several Jekyll plugins](https://github.com/binaryage/site/tree/master/_plugins) over past years.

Just to be clear. This is just a proposal. I'm perfectly ok if this PR gets rejected. It does not block me, I will [simply roll my patches on top of jekyll](https://github.com/binaryage/site/blob/3f019f062cf0348ee3421ac59bc0fbe5c61d7efd/Gemfile#L3) master in [my own fork](https://github.com/binaryage/jekyll). Just wanted to share this for other folks to benefit from.

parkr:
Prior art: https://github.com/jekyll/jekyll/issues/1082

We have considered this a few years ago and ultimately didn't feel like it was a good idea. This is a fairly fundamental change that is too big to just accept for the sake of helping IntelliJ users from annoyances.

I'm interested in the merits of why this feature should be considered *separate* from the IDE/text editor/syntax highlighter argument. What value does it add to separate YAML front matter from metadata?

darwin:
Fair enough, thanks for taking time to reply @parkr.

PR #5734, opened 2017-01-08
BUGFIX: fix pagination link for page 1

oxnz:
the old paginator generate link for page 1 always the same as paginator.previous_page_path, which is apparently wrong when page number is bigger that 2

### DIFF
--- a/docs/_docs/pagination.md
+++ b/docs/_docs/pagination.md
@@ -207,7 +207,7 @@ page with links to all but the current page.
     {% if page == paginator.page %}
       <em>{{ page }}</em>
     {% elsif page == 1 %}
-      <a href="{{ paginator.previous_page_path | prepend: site.baseurl | replace: '//', '/' }}">{{ page }}</a>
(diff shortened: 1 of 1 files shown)


ParthKolekar:
Any update on this?

docs/_docs/pagination.md
```diff
@@ -207,7 +207,7 @@ page with links to all but the current page.
     {% if page == paginator.page %}
       <em>{{ page }}</em>
     {% elsif page == 1 %}
-      <a href="{{ paginator.previous_page_path | prepend: site.baseurl | replace: '//', '/' }}">{{ page }}</a>
+      <a href="{{ "/index.html" | prepend: site.baseurl | replace: '//', '/' }}">{{ page }}</a>
```
ParthKolekar: <a href="{{ site.paginate_path | prepend: site.baseurl | replace: '//', '/' | replace: '/page/:num', '' }}">{{ page }}</a>

Is the above not better?

ParthKolekar: Each user will need to customize his/her pagination I think. So that makes my example no better than yours.

DirtyF:
Current example is working fine. Closing this.

PR #5724, opened 2017-01-05
Remove broken test

jekyllbot:
PR automatically created for @pathawks.

```text
Remove broken test
```

PR #5473, opened 2016-10-10
Addition of a sample "typical post"

kentonh:
So, I thought this section in the docs would be a good place to put the anatomy of a post .md file.

PR #5735, opened 2017-01-08
Add sample post

DirtyF:
to avoid waiting for @kentonh rebase on ##5473


DirtyF:
CI would pass with #5724

DirtyF:
resolved in #5473

PR #5698, opened 2016-12-27
Documentation on how to build navigation

tomjohnson1492:
I added a documentation page on how to build navigation for your site. This topic is primarily intended for users who have a lot of pages on their site (such as for documentation websites), and want to build a more robust sidebar navigation.

Jekyll combines Liquid with YAML in interesting ways that aren't really documented clearly in the existing docs, except for a brief reference [here](http://jekyllrb.com/docs/datafiles/#the-data-folder). You can read about Liquid on Shopify and YAML in YAML's docs, but exactly how you store YAML files in a Jekyll project and iterate through them using Liquid loops and filters to generate lists of pages is something that isn't clear to a lot of people. (You can see origins of these questions in [previous help issues](https://github.com/jekyll/jekyll-help/issues/266).)  The documentation on navigation would fit well into the Jekyll docs.

@jekyll/documentation

PR #5742, opened 2017-01-09
Retire Jekyll 2 docs

### LINKED ISSUE #5741: Remove requirements for Jekyll 2
http://jekyllrb.com/docs/installation/ still mentions Jekyll 2 dependencies — an old Ruby version, NodeJS and Python. I don't know how many people would like to install Jekyll 2 today, version 3 was published on October 27th 2015, maybe it's time to remove those instructions?

/cc @jekyll/core @jekyll/documentation

chrisfinazzo:
cc/ @jekyll/documentation.


DirtyF:
fixes #5741

docs/_docs/upgrading.md
```diff
@@ -6,5 +6,4 @@ permalink: /docs/upgrading/
 
 Upgrading from an older version of Jekyll? Upgrading to a new major version of Jekyll (e.g. from v2.x to v3.x) may cause some headaches. Take the following guides to aid your upgrade:
```
DirtyF: ➡️  This page could now redirect to  /upgrading/2-to-3/

chrisfinazzo: OK, will update. Am I correct in that this is just another entry inside of `docs/redirects`?

DirtyF: I think these are for external redirects.

docs/_docs/upgrading.md
```diff
@@ -6,5 +6,4 @@ permalink: /docs/upgrading/
...
 
-- [From 0.x to 1.x and 2.x](/docs/upgrading/0-to-2/)
```
parkr: I'd keep this link!

docs/_docs/upgrading/0-to-2.md
```diff
@@ -1,140 +0,0 @@
----
-layout: docs
-title: Upgrading from 0.x to 2.x
-permalink: /docs/upgrading/0-to-2/
----
-
-Upgrading from an older version of Jekyll? A few things have changed in 1.0
```
parkr: Let's leave this document – it's still useful. Believe it or not, there are plenty of folks using Jekyll 0.12.

DirtyF: I was wondering, is there any way to know when a version gets downloaded from Rubygems?

parkr: There aren't any hooks, but we could poll: http://guides.rubygems.org/rubygems-org-api/#gem-version-methods

pathawks: I worry that somebody might inherit a Jekyll site that hasn't been updated for years. Even if we aren't supporting 2.0, we should still provide a path for Jekyll 1.0 users to upgrade their sites to the latest and greatest.

DirtyF: I thought we could assume we provide documentation for latest version by default, but we can point to an older version like 2.0 ou 0.12 docs version.

DirtyF:
FYI [A recent Twitter poll](https://twitter.com/jekyllrb/status/821152178492600321) show that 14% of our users are still using Jekyll v2 or less. That's quiet a lot, those people should be able to access a documentation adapted for the version they still use, even if we strongly invite them to update.

Shouldn't we build a documentation corresponding to latest v2 and hosting it on something like https://v2.jekyllrb.com/ ?

chrisfinazzo:
I think (long-term) that's the direction we need to move in with regard to old versions that are no longer being targeted with new features, but still need support. At the same time, it seems the navigation needs to be rethought given the comments in #5698 - adding more items to the list is untenable.

Using a separate domain would make it clear about how to retire "active" documentation as new versions get released, so I'm 👍🏼 on this.

parkr:
I think we should keep these around. We can do a better job of *isolating* the Jekyll 2 docs so as not to confuse users, but we shouldn't get rid of instructions when some folks are still using them.

Issue #5718, opened 2017-01-05
absolute_url should complain if URL is not absolute


pathawks:
If `site.url` == `example.org`, then `{{ "/" | absolute_url }}` will return `example.org/`.

If the current page (`https://example.org/about/`) includes a link to the above URL, visiting that link will direct the browser to `https://example.org/about/example.org/`.

This is obviously not the intended behavior.

Now that **jekyll-sitemap** and **jekyll-feed** are relying on these filters, a `site.url` that does not contain a protocol can bork a site's entire sitemap or feed.

I don't know if this is best fixed with `jekyll doctor` or by complaining at build time.

parkr:
Duplicate of #5788?

pathawks:
> Duplicate of #5788?

No. Kind of the opposite.

In _this_ case, if `site.url` is not absolute, then the `absolute_url` filter will never work: it will never generate an absolute URL.

ashmaroli:
@pathawks the PR linked to this ticket has been merged. Is this still an issue?

pathawks:
@ashmaroli Good catch :+1:

PR #5760, opened 2017-01-12
Add URL checks to Doctor

pathawks:
If a site does not provide an absolute URL in `site.url`, all sorts of wonky behavior can creep up, especially as sites rely more heavily on themes plugins that assume standard behavior.

To this end, I would like to add some simple URL checks to `jekyll doctor` to make sure that things are in good working order.

### DIFF
--- a/lib/jekyll/commands/doctor.rb
+++ b/lib/jekyll/commands/doctor.rb
@@ -96,9 +96,11 @@ module Jekyll
 
         def proper_site_url?(site)
           url = site.config["url"]
-          url_exists?(url) &&
-            url_valid?(url) &&
-            url_absolute(url)
+          [
+            url_exists?(url),
+            url_valid?(url),
+            url_absolute(url),
+          ].all?
         end
 
         private
--- a/lib/jekyll/commands/doctor.rb
+++ b/lib/jekyll/commands/doctor.rb
@@ -1,3 +1,5 @@
+require "addressable/uri"
+
 module Jekyll
   module Commands
     class Doctor < Command
@@ -36,6 +38,7 @@ module Jekyll
             !deprecated_relative_permalinks(site),
             !conflicting_urls(site),
             !urls_only_differ_by_case(site),
+            proper_site_url?(site)
           ].all?
         end
 
@@ -91,6 +94,13 @@ module Jekyll
           urls_only_differ_by_case
         end
 
+        def proper_site_url?(site)
+          url = site.config["url"]
+          url_exists?(url) &&
+            url_valid?(url) &&
+            url_absolute(url)
+        end
+
         private
         def collect_urls(urls, things, destination)
           things.each do |thing|
@@ -109,6 +119,27 @@ module Jekyll
             dest = thing.destination(destination)
             (memo[dest.downcase] ||= []) << dest
           end
+        end
+
+        def url_exists?(url)
+          return true unless url.nil? || url.empty?
+          Jekyll.logger.warn "Warning:", "Site URL does not appear to be set"
+          false
+        end
+
+        def url_valid?(url)
+          Addressable::URI.parse(url)
+          true
+        rescue
+          Jekyll.logger.warn "Warning:", "Cannot parse site URL: #{url}"
+          false
+        end
+
+        def url_absolute(url)
+          return true if Addressable::URI.parse(url).absolute?
+          Jekyll.logger.warn "Warning:", "Site URL does not appear to be" \
+            " absolute: #{url}"
+          false
         end
       end
     end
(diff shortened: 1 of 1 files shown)


lib/jekyll/commands/doctor.rb
```diff
@@ -35,7 +37,8 @@ def healthy?(site)
             fsnotify_buggy?(site),
             !deprecated_relative_permalinks(site),
             !conflicting_urls(site),
-            !urls_only_differ_by_case(site)
+            !urls_only_differ_by_case(site),
+            proper_site_url?(site)
```
pathawks: If only Rubocop allowed `Style/TrailingCommaInLiteral` :rage:

pathawks:
I should probably add a check for `baseurl` as well, as that is also the source of much frustration.

lib/jekyll/commands/doctor.rb
DirtyF: @pathawks what prevents us from adding `Style/TrailingCommaInLiteral: { EnforcedStyleForMultiline: consistent_comma }`to `.rubocop.yml`?

pathawks: @DirtyF Laziness 😝

pathawks:
Fixes #5718

pathawks:
CI failures have been fixed :+1:

pathawks:
> So now you've got to set `site.url`?

It's not mandatory, but since leaving it blank can cause problems with plugins or themes, users could at least be made aware of the situation.

Note that this only runs when a user runs `jekyll doctor`, not on every build. If a user is running `jekyll doctor` it's probably because something is not working as expected.

If a user is intentionally leaving `site.url` blank, they can safely disregard this warning.

parkr:
> If a user is intentionally leaving site.url blank, they can safely disregard this warning.

Hm, maybe just don't run the check then?

DirtyF:
@pathawks so should we break those tests: one that checks if site.url exists and warn if not, and another that checks the submitted URL is valid?

lib/jekyll/commands/doctor.rb
```diff
@@ -91,6 +94,15 @@ def urls_only_differ_by_case(site)
           urls_only_differ_by_case
         end
 
+        def proper_site_url?(site)
+          url = site.config["url"]
+          [
+            url_exists?(url),
+            url_valid?(url),
```
pathawks: @DirtyF So, replace these two lines with:

```ruby
url_exists?(url) && url_valid?(url)
```

?

DirtyF: I was thinking of moving `url_exists` out of `proper_site_url?` in a separate test to address Parker's concern.  (We dont wan't to warn users that intentionnaly left site.url blank)

pathawks: I don't follow.

We _really_ want `site.url` to be set, either manually by the user or automatically by something like `pages-gem`. If `site.url` is blank, it is likely to cause problems and we should warn the user about this.

If a user has some special reason to use a blank `site.url` then they will understand that this warning does not apply to them.
On the other hand, if a user has simply neglected to set `site.url` and now things aren't working as expected,  this check should alert them to the issue and suggest setting `site.url`. That is the entire point of this PR.

I'm probably not understanding your concerns; can you try to explain to me again now that I've explained my intent?

DirtyF: I get you perfectly and as it's impossible to tell if a blank `site.url` is intentional or not, you prefer to warn in both cases.

Maybe just make the messages more clear then? 

```sh
bundle exec jekyll doctor
Warning: You didn't set an URL in the config file, you may encounter problems with some plugins.
```
```sh
bundle exec jekyll doctor
Warning: The site URL does not seem to be valid, check the value of `url` in your config file.
```

```sh
bundle exec jekyll doctor
Warning: Your site URL does not seem to be absolute, check the value of `url` in your config file.
```

DirtyF:
@pathawks BTW what do you think about adding a "check your configuration" section with `jekyll doctor` in the docs?  I just realized this command is not documented on the website 😕

pathawks:
@DirtyF Good idea. We should also add a request to the PR template that users post the output of `jekyll doctor`

pathawks:
For an example of what I hope to accomplish with this: jekyll/jekyll-sitemap#171

pathawks:
> We should also add a request to the PR template that users post the output of `jekyll doctor`

Thank you @DirtyF for #6169

(Breadcrumbs 🥖)

DirtyF:
@jekyllbot : merge +minor

PR #5769, opened 2017-01-16
Fix a markdown link to look properly on the web

diev:
As I read docs on the web there is:

> If you have installed an older version, you can update the RubyGems package using the directions [here.][ssl-certificate-update] [ssl-certificate-update]: http://guides.rubygems.org/ssl-certificate-update/#installing-using-update-packages

instead of a proper link like others.


DirtyF:
@jekyllbot: merge +doc

PR #5779, opened 2017-01-16
test/helper: fix flaky plugin path test by removing calls to Dir.chdir without a block

parkr:
If we do a `Dir.chdir` before `Configuration::DEFAULTS` is initialized, then its `source` and `destination` values will not be what we expect. We expect that `Dir.pwd` should stay as the root of the repo but there are some errant calls to `Dir.chdir` without a block that are still not yet cleaned up. Remove all bare `Dir.chdir` calls.

Failing test this fixes:

```text
Failure:
TestSite#test_: configuring sites should have an array for plugins by default.  [/home/travis/build/jekyll/jekyll/test/test_site.rb:7]
+++ actual
@@ -1 +1 @@
-["/home/travis/build/jekyll/jekyll/_plugins"]
+["/home/travis/build/jekyll/jekyll/test/source/_plugins"]
```

Example: https://travis-ci.org/jekyll/jekyll/jobs/192169603

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -30,11 +30,6 @@ require "minitest/profile"
 require "rspec/mocks"
 require_relative "../lib/jekyll.rb"
 
-# The default "source" and "destination" values depend on Dir.pwd,
-# which changes when we use Dir.chdir without a block. Initialize
-# it here so it has Dir.pwd = root of this repo.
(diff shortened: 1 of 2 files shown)


pathawks:
Closes #5724

parkr:
@jekyllbot: merge +dev

PR #5780, opened 2017-01-17
Don't include the theme's includes_path if it is nil.

### LINKED ISSUE #5639: All include paths should filter out nil
### Steps to reproduce

1. Create a theme without an `_includes` directory
2. Create a page with an include
3. Build the site

### Expected

Include is loaded from the site directory

### Actual

Site errors out:

```
Liquid Exception: no implicit conversion of nil into String in index.html
Fatal: TypeError
no implicit conversion of nil into String
/hoosegow/.bundle/ruby/2.3.0/gems/jekyll-3.3.1/lib/jekyll/tags/include.rb:115:in `join'
```

The issue is that https://github.com/jekyll/jekyll/blob/master/lib/jekyll/site.rb#L438 returns nil, meaning there's a nil entry in the configure_include_paths array.

For includes (and layouts and sass, and assets) we should run the array through `compact` before returning from the method (to remove the null entries and prevent errors).

//cc @jekyll/ecosystem

parkr:
Fixes #5639.

### DIFF
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -435,7 +435,7 @@ module Jekyll
     private
     def configure_include_paths
       @includes_load_paths = Array(in_source_dir(config["includes_dir"].to_s))
-      @includes_load_paths << theme.includes_path if self.theme
+      @includes_load_paths << theme.includes_path if self.theme && theme.includes_path
     end
 
     private
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -435,7 +435,7 @@ module Jekyll
     private
     def configure_include_paths
       @includes_load_paths = Array(in_source_dir(config["includes_dir"].to_s))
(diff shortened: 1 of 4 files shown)


lib/jekyll/site.rb
```diff
@@ -435,7 +435,7 @@ def configure_theme
     private
     def configure_include_paths
       @includes_load_paths = Array(in_source_dir(config["includes_dir"].to_s))
-      @includes_load_paths << theme.includes_path if self.theme
+      @includes_load_paths << theme.includes_path if self.theme && theme.includes_path
```
benbalter: Does this need to be `self.theme`? If not, can they both be just `theme`?

pathawks:
Is there a test for this?

parkr:
> Is there a test for this?

@pathawks Yes! Pushed it up.

parkr:
@jekyllbot: merge +bug

PR #5781, opened 2017-01-17
Bump htmlproofer

### LINKED ISSUE #5777: Update dependency constraint to allow for html-proofer v3.4.0
Hey there! :wave:

I noticed that the constraint you have for html-proofer doesn't allow for the latest version to be used.

The constraint I found was `~> 2.0`, and the latest version available is `3.4.0`.

DirtyF:
fix #5777 

Note: `script/proof` could be faster with the use of typhoeus.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -88,7 +88,7 @@ end
 
 group :site do
   if ENV["PROOF"]
-    gem "html-proofer", "~> 2.0"
+    gem "html-proofer", "~> 3.4"
   end
 
   gem "jekyll-avatar"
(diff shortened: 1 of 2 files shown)


parkr:
> Note: script/proof could be faster with the use of typhoeus.

How so? html-proofer uses Typhoeus and Hydra to do its work, so I'm wondering how this in fact help us.

DirtyF:
@jekyllbot: merge +dev

DirtyF:
@parkr you're right sorry. But could [Parallel](https://github.com/gjtorikian/html-proofer#configuring-parallel) and a bit of caching help avoid recheck every single link every time we run `script/proof`?

PR #5782, opened 2017-01-17
Use latest jemoji gem

### LINKED ISSUE #5778: Update dependency constraint to allow for jemoji v0.7.0
Hey there! :wave:

I noticed that the constraint you have for jemoji doesn't allow for the latest version to be used.

The constraint I found was `0.5.1`, and the latest version available is `0.7.0`.

DirtyF:
fix #5778


DirtyF:
@jekyllbot: merge +dev

PR #5783, opened 2017-01-17
add `jekyll-admin` by default

ashmaroli:
add `jekyll-admin` to the `Gemfile` and `_config.yml` of a new site installed by `jekyll new`, by default.
@mertkahyaoglu and his team at jekyll-admin will be resolving many Windows-specific bugs in ~~the coming release (most probably~~ `v0.3.0` ~~if not `v0.2.1`)~~ and hence I think it would be safe to be included by default for all users.

milestone: `3.5`, `undetermined`

### DIFF
--- a/lib/jekyll/commands/new.rb
+++ b/lib/jekyll/commands/new.rb
@@ -82,7 +82,8 @@ gem "minima", "~> 2.0"
 
 # If you have any plugins, put them here!
 group :jekyll_plugins do
-   gem "jekyll-feed", "~> 0.6"
+  gem "jekyll-admin", "~> 0.3"
+  gem "jekyll-feed", "~> 0.6"
 end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
--- a/lib/site_template/_config.yml
+++ b/lib/site_template/_config.yml
@@ -28,6 +28,7 @@ github_username:  jekyll
 markdown: kramdown
 theme: minima
 gems:
+  - jekyll-admin
   - jekyll-feed
 exclude:
   - Gemfile


DirtyF:
@ashmaroli rather than imposing to install a lot of dependencies to all of our users, could we invite those who want to use a web interface locally to edit their content? These could be commented for instance.

quiltbag:
i like the idea of putting "recommended plugins" that are commented out by default in the gemfile! perhaps there could be an entire section in there that recommends some plugins and also includes a link to the plugin browser?

DirtyF:
@quiltbag great idea, we can safely assume than most users would want a sitemap, a feed, something to help with seo, etc. 👍

ashmaroli:
> These lines could be commented for instance.

Nice idea. I'll wait for feedback from the Core Team to see if jekyll-plugin can be commented out as well.

> perhaps there could be an entire section in there that recommends some plugins

This is a nice idea too.. 


So this is how I see things could end up:
```yaml
gems:
  - jekyll-feed # Generate an Atom (RSS-like) feed of your Jekyll posts
  - # jekyll-admin # CMS-style graphical interface for your site
  - # jekyll-seo-tag # Add metadata tags for better SEO
```

quiltbag:
the only problem i see with this is that the same commented out lines would have to be both in the `Gemfile` and in the `_config.yml`. it's kind of weird

DirtyF:
If I'm not mistaken, the jekyll_plugins group in Gemfile is supposed to prevent a user to add gem in config. Correct me if I'm wrong.

quiltbag:
@DirtyF is it? that's good to know

ashmaroli:
> it's kind of weird

yeah that's true. One way out is to emulate hows Rails (please excuse me) does it:
```ruby
group :jekyll_plugins do
  # Jekyll plugin that provides users with a traditional CMS-style graphical interface to author content and administer Jekyll sites.
  # https://jekyll.github.io/jekyll-admin
  gem "jekyll-admin", "~> 0.3.0"
end
```

ashmaroli:
> the `jekyll_plugins` group in Gemfile is supposed to prevent a user to add gem in config.

@DirtyF actually no. the config file is for **Jekyll** to `require` these gems, and the `:jekyll_plugins` group is for Bundler.
I'll check and get back on this..

DirtyF:
It's listed as a viable option to install a plugin with bundler in our docs: https://jekyllrb.com/docs/plugins/#installing-a-plugin

ashmaroli:
Ah! looks like my notes are wrong.
This code block says it all:

```ruby
    def self.require_from_bundler
      if !ENV["JEKYLL_NO_BUNDLER_REQUIRE"] && File.file?("Gemfile")
        require "bundler"

        Bundler.setup
        required_gems = Bundler.require(:jekyll_plugins)
        message = "Required #{required_gems.map(&:name).join(", ")}"
        Jekyll.logger.debug("PluginManager:", message)
        ENV["JEKYLL_NO_BUNDLER_REQUIRE"] = "true"

        true
      else
        false
      end
    end
```

pathawks:
> If I'm not mistaken, the jekyll_plugins group in Gemfile is supposed to prevent a user to add gem in config.

This is true and it works great when building locally; I’m not sure it works when the site is built with GitHub Pages.

pnn:
@pathawks i can test that if you want

pathawks:
@pnn Yes please :+1:

pnn:
okay, doing this from github pages is very specific, since the `github-pages` gem locks explicit versions of the jekyll plugins it allows, which are not always the newest ones. you'd have to specify the exact versions in the `jekyll_plugins` group, but this would likely not be worth it

mertkahyaoglu:
The only problem for now is that Jekyll Admin recently disabled `watch` (via [#320](https://github.com/jekyll/jekyll-admin/pull/320)) which makes site regeneration not available outside of Jekyll Admin. Users that are not using Jekyll Admin will not be able to benefit from site regeneration by default.

I'm glad that Jekyll Admin will reach more people with this PR but also we should be ready for incoming issues about this until we find better solution to the problem.

/cc @benbalter

ashmaroli:
> The only problem for now is that Jekyll Admin recently disabled watch (via #320) which makes site regeneration not available outside of Jekyll Admin.

This sure is a deal-breaker for now.

label: `do-not-merge`

DirtyF:
Feel free to reopen when `jekyll-admin` does not disable `watch` by default.

ashmaroli:
> Feel free to reopen when jekyll-admin does not disable watch by default.

Sure.. no probs..

PR #5784, opened 2017-01-17
Bump rubies we test against

ashmaroli:
Bump ruby versions in travis.yml


quiltbag:
@jekyllbot: merge +dev

PR #5792, opened 2017-01-18
Document Jekyll's Philosophy

### LINKED ISSUE #5342: Jekyll's Philosophy
Hey y'all! I wanted to put together a guiding philosophy for the Jekyll project to help make it easier to decide whether to accept a feature request, whether a bug is a bug or actually a feature request, and so on. Jekyll has a massive user base (arguably the largest) and we need to do our best to serve the needs of our users. We have some philosophy verbiage in [the README](https://github.com/jekyll/jekyll#philosophy) and in [the maintainer documentation](https://github.com/jekyll/jekyll/blob/master/docs/triaging-an-issue.md), but I think we should create a `PHILOSOPHY` document everyone can refer to as we iterate on Jekyll to help guide us. This is meant to be a document that _can_ change, but think of it as the Constitution rather than a handful of malleable laws.

I wanted to open an issue to help facilitate a conversation around one question: what is Jekyll's philosophy? In other words, what are our guiding principles?

To start the conversation, here are a few I keep in mind as I look at issues and evaluate pull requests:
- No magic: Jekyll should do what you ask of it and nothing more.
- Easy out-of-the-box experience: If something increases the learning curve, we should be skeptical. Things should "just work" with sane defaults.
- Jekyll is content-focused: what matters is the content. Let the content be king.
- Jekyll is not an airplane cockpit: take on the burden of making hard choices on our users' behalf. Do not give a configuration option for every little thing.
- Jekyll is for Hackers: Jekyll should be easily extensible.

What do you think of as Jekyll's philosophical guiding principles?

Prior art: [WordPress](https://wordpress.org/about/philosophy/).

/cc @jekyll/build @jekyll/documentation @jekyll/ecosystem @jekyll/performance @jekyll/stability @jekyll/windows @jekyll/core

parkr:
Here are 5 points to start with.

Fixes #5342.

/cc @benbalter @pathawks @mattr- for :eyes: especially. :heart:

### DIFF
--- a/docs/philosophy.md
+++ b/docs/philosophy.md
@@ -11,7 +11,8 @@ is about, the following principles should come to mind:
 
 Jekyll is not magic. A user should be able to understand the underlying
 processes that make up the Jekyll build without much reading. It should
-behave "as you'd expect."
+do only what you ask it to and nothing more. When a user takes a certain
+action, the outcome should be easily understandable and focused.
 
 ### 2. It "Just Works"
 
@@ -19,7 +20,7 @@ The out-of-the-box experience should be that it "just works." Run `gem
 install jekyll` and it should build any Jekyll site that it's given.
 Features like auto-regeneration and settings like the markdown renderer
 should represent sane defaults that work perfectly for the vast majority of
-cases. The burden of configuration should not be placed on the user.
+cases. The burden of initial configuration should not be placed on the user.
 
 ### 3. Content is King
 
@@ -31,6 +32,12 @@ should find the management of their content enjoyable and simple.
 
 If a user's site builds today, it should build tomorrow.
 Backwards-compatibility should be strongly preferred over breaking changes.
+Breaking changes should be made to support a strong practical goal, and
+breaking changes should never be made to drive forward "purity" of the
+codebase, or other changes purely to make the maintainers' lives easier.
+Breaking changes provide a significant amount of friction between upgrades
+and reduce the confidence of users in this software, and should thus be
+avoided unless absolutely necessary.
 Upon breaking changes, provide a clear path for users to upgrade.
 
 ### 5. Small & Extensible
@@ -38,4 +45,6 @@ Upon breaking changes, provide a clear path for users to upgrade.
 The core of Jekyll should be simple and small, and extensibility should be
 a first-class feature to provide added functionality from community
 contributors. The core should be kept to features used by at least 90% of
-users–everything else should be provided as a plugin.
+users–everything else should be provided as a plugin. New features should
+be shipped as plugins and focus should be put on creating extensible core
+API's to support rich plugins.
--- a/docs/philosophy.md
+++ b/docs/philosophy.md
@@ -1,0 +1,41 @@
+---
+title: Philosophy
+---
+
+Jekyll offers a unique philosophy when approaching the problem of static
+site generation. This core philosophy drives development and product
+decisions. When a contributor, maintainer, or user asks herself what Jekyll
+is about, the following principles should come to mind:
+
+### 1. No Magic
+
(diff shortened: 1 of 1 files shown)


ashmaroli:
Looks Good! But hows is this document *linked* to rest of the documentation at `jekyllrb.com`
I understand that this pertains to Jekyll Core, but how much of this is applicable to **Jekyll Plugins** created by the contributors?

docs/philosophy.md
```diff
@@ -0,0 +1,41 @@
...
+Jekyll offers a unique philosophy when approaching the problem of static
+site generation. This core philosophy drives development and product
+decisions. When a contributor, maintainer, or user asks herself what Jekyll
+is about, the following principles should come to mind:
+
+### 1. No Magic
+
+Jekyll is not magic. A user should be able to understand the underlying
+processes that make up the Jekyll build without much reading. It should
+behave "as you'd expect."
```
benbalter: @parkr you've described this previously as "Jekyll does exactly what the user asks, no more, no less". Is that still accurate? "I expect this" is very different from "I asked you to do this" or "I understand what's going on under the hood to cause this to happen".

docs/philosophy.md
```diff
@@ -0,0 +1,41 @@
...
+
+### 2. It "Just Works"
+
+The out-of-the-box experience should be that it "just works." Run `gem
+install jekyll` and it should build any Jekyll site that it's given.
+Features like auto-regeneration and settings like the markdown renderer
+should represent sane defaults that work perfectly for the vast majority of
+cases. The burden of configuration should not be placed on the user.
```
benbalter: > The burden of configuration

Initial configuration? Initial setup? Configuration in general?

docs/philosophy.md
```diff
@@ -0,0 +1,41 @@
...
+
+Why is Jekyll so loved by content creators? It focuses on content first and
+foremost, making the process of publishing content on the Web easy. Users
+should find the management of their content enjoyable and simple.
+
+### 4. Stability
+
+If a user's site builds today, it should build tomorrow.
+Backwards-compatibility should be strongly preferred over breaking changes.
+Upon breaking changes, provide a clear path for users to upgrade.
```
benbalter: Also maybe that when it comes to breaking changes, practicality over purity. There should be a strong reason to institution a breaking change, and it needs to outweigh the burden we're pushing on users. "I would have done it this way", "this way is easier to read", or "this is the textbook approach" aren't compelling reasons to break a user's site.

docs/philosophy.md
```diff
@@ -0,0 +1,41 @@
...
+
+### 5. Small & Extensible
+
+The core of Jekyll should be simple and small, and extensibility should be
+a first-class feature to provide added functionality from community
+contributors. The core should be kept to features used by at least 90% of
+users–everything else should be provided as a plugin.
```
benbalter: Maybe to note that (A) Jekyll itself maintains plugins, and (B) the preferred approach for new features is to create the necessary APIs first and then dogfood an implementation, rather that going in the reverse direction. Put another way, Jekyll should use its own plugin API system to develop new features, it's not just for third-party functionality.

docs/philosophy.md
parkr: >  "I expect this" is very different from "I asked you to do this"

This is a great point! I'll update this piece to clarify that.

docs/philosophy.md
parkr: I think **initial configuration** is what I meant here. It's the out-of-the-box experience that I'm getting at.

parkr:
https://jekyllrb.com/philosophy

benbalter:
Nice! Perhaps this warrants a blog post? 😄

parkr:
Good idea!

PR #5797, opened 2017-01-19
Bump rdoc to v5.0

### LINKED ISSUE #5776: Update dependency constraint to allow for rdoc v5.0.0
Hey there! :wave:

I noticed that the constraint you have for rdoc doesn't allow for the latest version to be used.

The constraint I found was `~> 4.2`, and the latest version available is `5.0.0`.

DirtyF:
Let's fix #5776 📖


pnn:
for context, here's the changelog: https://github.com/rdoc/rdoc/blob/master/History.rdoc#500--2016-11-05 includes 4.3 as well

parkr:
@jekyllbot: merge +dev

PR #5803, opened 2017-01-21
Remove `ruby RUBY_VERSION` from generated Gemfile

ashmaroli:
I've had `ruby RUBY_VERSION` in all my Gemfiles generated by `jekyll new` on Windows. Ignored it because it never seemed to cause any issue.
Proposing interpolating the variable based on this [comment](https://github.com/jekyll/jekyll/issues/5719#issuecomment-274227923) to trigger a Travis build and check if this is a fix..

### DIFF
--- a/lib/jekyll/commands/new.rb
+++ b/lib/jekyll/commands/new.rb
@@ -61,7 +61,6 @@ module Jekyll
         def gemfile_contents
           <<-RUBY
 source "https://rubygems.org"
-ruby "#{RUBY_VERSION}"
 
 # Hello! This is where you manage which Jekyll version is used to run.
 # When you want to use a different version, change it below, save the
--- a/lib/jekyll/commands/new.rb
+++ b/lib/jekyll/commands/new.rb
@@ -61,7 +61,7 @@ module Jekyll
         def gemfile_contents
           <<-RUBY
 source "https://rubygems.org"
-ruby RUBY_VERSION
+ruby "#{RUBY_VERSION}"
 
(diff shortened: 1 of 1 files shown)


ashmaroli:
okay.. dont know what burnt Travis. But local tests on my **Windows** system with `Ruby 2.3.3` and `Bundler 1.14.0` did not raise issues with `ruby RUBY_VERSION`

This PR may lead to future issues by having the subsequent blogs generated locked to a certain Ruby version.
Leaving it open for further feedback

label: `Discussion`, `dont-merge-yet`

parkr:
We should probably just get rid of this line. I think it's only useful because Heroku requires it, but even then it's not that useful. I use the `ruby` pragma in `Gemfile` exactly never.

ashmaroli:
> We should probably just get rid of this line

I like that idea. Would it be better if it were commented out instead?

parkr:
> Would it be better if it were commented out instead?

Why comment it out? That seems like useless bloat.

ashmaroli:
> That seems like useless bloat.

LOL! okay, will remove it entirely.

ashmaroli:
The `do-not-merge` label ought to be removed, no?

pathawks:
@ashmaroli Yeah! Who added that anyway?
…oh 😶

DirtyF:
@jekyllbot: merge +dev

parkr:
> @jekyllbot: merge +dev

This isn't a dev change, this is a change to the way `jekyll new` works.

PR #5805, opened 2017-01-21
Update github-pages.md

jbrow78:
I added a few tags that I think you'll need @jekyllbot


pathawks:
Thanks for the pull request.

This file has just a small amount of content that will be inserted into a layout that has the proper HTML tags. You'll notice we don't open `</body>` or `</html>` in this file either.

Always appreciate more eyes on the project though :eyes::+1:

PR #5811, opened 2017-01-22
Add mention of classifier-reborn for LSI

nhoizey:
classifier-reborn is mandatory for LSI since Jekyll 3.0

### DIFF
--- a/docs/_docs/configuration.md
+++ b/docs/_docs/configuration.md
@@ -233,7 +233,7 @@ class="flag">flags</code> (specified on the command-line) that control them.
     <tr class="setting">
       <td>
         <p class="name"><strong>LSI</strong></p>
-        <p class="description">Produce an index for related posts.</p>
+        <p class="description">Produce an index for related posts. The <a href="http://www.classifier-reborn.com/">classifier-reborn</a> plugin must be used.</p>
       </td>
       <td class="align-center">
         <p><code class="option">lsi: BOOL</code></p>
--- a/docs/_docs/configuration.md
+++ b/docs/_docs/configuration.md
@@ -233,7 +233,7 @@ class="flag">flags</code> (specified on the command-line) that control them.
     <tr class="setting">
       <td>
         <p class="name"><strong>LSI</strong></p>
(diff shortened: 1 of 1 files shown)


DirtyF:
Thanks @nhoizey ;)

It's not mentioned on the [variables](https://jekyllrb.com/docs/variables/) page either, but if you try to run a build with the `--lsi` option, Jekyll will ask you to install the gem.

```shell
$ bundle exec jekyll build --lsi
...
      Generating...
  Dependency Error: Yikes! It looks like you don't have classifier-reborn or one of its dependencies installed. In order to use Jekyll as currently configured, you'll need to install this gem. The full error message from Ruby is: 'cannot load such file -- classifier-reborn' If you run into trouble, you can find helpful resources at http://jekyllrb.com/help/!
                    classifier-reborn
```

docs/_docs/configuration.md
```diff
@@ -233,7 +233,7 @@ class="flag">flags</code> (specified on the command-line) that control them.
     <tr class="setting">
       <td>
         <p class="name"><strong>LSI</strong></p>
-        <p class="description">Produce an index for related posts.</p>
+        <p class="description">Produce an index for related posts. The <a href="http://www.classifier-reborn.com/">classifier-reborn</a> plugin must be used.</p>
```
ashmaroli: How about simply: "**( Requires [classifier-reborn](http://www.classifier-reborn.com/) plugin )**" ?

nhoizey:
@DirtyF should I make also a PR for the variables page, then?

Or maybe should the `--lsi` option and `site.related_posts` variable be removed from the documentation, since they're not in Jekyll by default anymore?

pnn:
@nhoizey i believe it should be fine to edit both pages in the scope of this PR

DirtyF:
@nhoizey I understand  you feel it's kind of a false promise as if you install `classifier-reborn` without `gsl`, the indexation takes forever. For now, those links are already of great value for users.

I don't think @jekyll-core intend to remove completely the `--lsi` option and related posts from the core for now, but following the 80/20 principle, this might be something that could be added to the Jekyll 4 roadmap. Not my decision here.

DirtyF:
@jekyllbot: merge +doc

PR #5759, opened 2017-01-12
Update installation.md

BlueberryFoxtrot:
1. Addition of *Running Jekyll on Ubuntu* section, to address Ubuntu stumbling block as per https://github.com/jekyll/jekyll/issues/5719.
2. Restructuring, and I hope I understood correctly when NodeJS/Python are/aren't required.
3. Gentler wording – it's probably not a good idea to tell punters who hit this page because they ran into trouble that installing Jekyll *is* easy and straight-forward; it *ought to be* straight-forward. (There's always the potential for pain and confusion if not all dependencies are in place.)

PR #5817, opened 2017-01-24
Install troubleshooting on Ubuntu

DirtyF:
closes #5759 

- The link to the troubleshooting page is more obvious on installation page
- Ubuntu troubleshooting is on troubleshooting page with other Linux specific instructions

/cc @BlueberryFoxtrot

### DIFF
--- a/docs/_docs/installation.md
+++ b/docs/_docs/installation.md
@@ -10,16 +10,31 @@ encountered and how we might make the process easier.
 
 ### Requirements
 
-Installing Jekyll is easy and straight-forward, but there are a few
-requirements you’ll need to make sure your system has before you start.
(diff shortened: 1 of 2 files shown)


docs/_docs/installation.md
```diff
@@ -10,18 +10,29 @@ encountered and how we might make the process easier.
 
 ### Requirements
 
-Installing Jekyll is easy and straight-forward, but there are a few
-requirements you’ll need to make sure your system has before you start.
+Installing Jekyll ought to be straight-forward if all requirements are met.
```
pnn: can we replace `ought to be` with `should be`? it's a little less complicated

DirtyF:
@jekyllbot: merge +doc

PR #5819, opened 2017-01-25
Fixed inaccuracy in "Built-in permalink styles" docs

arogachev:
Must be either:

> Rather than typing `permalink: /:categories/:year/:month/:day/:title/`, you can just type `permalink: pretty`.

or:

> Rather than typing `permalink: /:categories/:year/:month/:day/:title.html`, you can just type `permalink: date`.

I guess the former was meant to write because the latter was already mentioned in "Where to configure permalinks" section.


pnn:
@jekyllbot: merge +docs

PR #5822, opened 2017-01-26
use logger.info for related posts

joelmeyerhamme:
Imo running `--lsi` should use `Jekyll.logger.info`, so it can be made `--quiet`.


parkr:
@jekyllbot: merge +bug

PR #5823, opened 2017-01-26
WIP: parallel builds

fw42:
This is a rough proof of concept of multi-process parallelism for Jekyll. I want to get some feedback whether or not this idea makes sense or not before spending more time on cleaning it up and writing tests.

Jekyll performance is mostly CPU constrained and Ruby processes are limited to using a single CPU core. The basic idea of this approach is that the parent process spawns N worker processes, each of which will be responsible for roughly 1/N of the files in the site (based on a hash of the filename).

This PR makes a few assumptions that might not be correct.

I haven't done a lot of testing, but for our rather big site, this PR breaks the `jekyll build` time on my 4 core Macbook **down from about 60 seconds to about 20 seconds** (while generating the exact same output). This doesn't currently work with `jekyll serve`, but looking at the code, it seems easy to make that work.

@parkr, @envygeeks, @pushrax, is there anything that you can think of that would be problematic with this approach? Any edge cases I'm not aware of? Has anyone tried this before?

### DIFF
--- a/lib/jekyll/command.rb
+++ b/lib/jekyll/command.rb
@@ -65,6 +65,7 @@ module Jekyll
         c.option "quiet", "-q", "--quiet", "Silence output."
         c.option "verbose", "-V", "--verbose", "Print verbose output."
         c.option "incremental", "-I", "--incremental", "Enable incremental rebuild."
+        c.option "parallel", "-p", "--parallel", "Enable parallel multi-process builds."
       end
     end
   end
--- a/lib/jekyll/commands/build.rb
+++ b/lib/jekyll/commands/build.rb
@@ -55,12 +55,32 @@ module Jekyll
           source      = options["source"]
           destination = options["destination"]
           incremental = options["incremental"]
+          parallel    = options["parallel"]
+
           Jekyll.logger.info "Source:", source
           Jekyll.logger.info "Destination:", destination
           Jekyll.logger.info "Incremental build:",
             (incremental ? "enabled" : "disabled. Enable with --incremental")
-          Jekyll.logger.info "Generating..."
-          process_site(site)
+
+          if parallel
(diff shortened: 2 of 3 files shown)


parkr:
My biggest concern is being able to ensure that multi-core functionality works on Windows or falls back to current 1-core functionality on Windows.

parkr:
I like how clean this approach is.

envygeeks:
Parallel processing won't break on Windows (threaded processing will though.) The only problem here is the same problem we ran into here at Forestry, and that's, that `@site` is laid out everywhere and it uses unmarshalable fallbacks in hashes, so my pull request stalled until 4.0 when we would rethink that whole nasty business and then actually implement some parallel processing.

I already have a pull request open that does this, and we do actively use it on Forestry but because of the `@site` problem we have to fallback on build failures with a specific message we can detect.  The pull request I have open demonstrates some of this with the custom Marshals.  We could make this considerably easier and push it into 3.x too if Jekyll starts storing it's site instances with an ID that can be actively tracked within Jekyll itself, then we could detect every `@site` object, throw it out and unmarshal the site back in pretty efficiently.

parkr:
@fw42 Another concern is user expectations about availability of `{{ post.content }}`. At the moment, collections are all processed first and everything is in alphabetical or date ascending order, thus the `Page` `index.html` will be able to iterate over `site.posts` and use their `post.content` value because all posts will already be rendered.

One this I've been noodling on is simply rendering as requested. Start with your first file, render that and if it requests the `.content` of any other page or post, render that. We keep an ongoing collection of what is being rendered and break if there's a loop. Just-in-time rendering.

fw42:
I see. I guess the false assumption that I made was that rendering one page is independent of rendering another page. Too bad :(

Issue #5485, opened 2016-10-14
Jekyll generates broken output when a page has a yaml syntax error in header


NerdyProjects:
- [X] I believe this to be a bug, not a question about using Jekyll.
- [X] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [ ] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [ ] This is a feature request.

---
- [ ] I am on (or have tested on) **_macOS**_ 10+
- [ ] I am on (or have tested on) **_Debian/Ubuntu**_ GNU/Linux
- [ ] I am on (or have tested on) **_Fedora**_ GNU/Linux
- [X] I am on (or have tested on) **_Arch**_ GNU/Linux
- [ ] I am on (or have tested on) **_Other**_ GNU/Linux
- [ ] I am on (or have tested on) **_Windows**_ 10+

(Jekyll 3.3.0)

---
- [ ] I was trying to install.
- [ ] There is a broken Plugin API.
- [ ] I had an error on GitHub Pages, and I have reproduced it locally.
- [ ] I had an error on GitHub Pages, and GitHub Support said it was a Jekyll Bug.
- [ ] I had an error on GitHub Pages and I did not test it locally.
- [X] I was trying to build.
- [ ] It was another bug.
## My Reproduction Steps
- Clone https://gitlab.com/foodsharing-dev/foodsharing-dev.gitlab.io (or create some empty jekyll blog)
- In one of the posts, invalidate the title by inserting a : without escaping it
- run jekyll build

It gives an error but still passes. The resulting page is broken (Contains an empty blog post without a layout)
Error:

```
 Generating... 
             Error: could not read file /home/matthias/Projekte/lebensmittelretten.de/git/foodsharing-dev.gitlab.io/_posts/2016-10-10-introducing-de.markdown: (<unknown>): mapping values are not allowed in this context at line 3 column 24
done in 0.379 seconds.
 Auto-regeneration: disabled. Use --watch to enable.
purefruit :: lebensmittelretten.de/git/foodsharing-dev.gitlab.io ‹master*› % echo $?                                                                                                                               
0

```
## The Output I Wanted

I would have thought it to fail the build.

DirtyF:
With Jekyll 3.3 on Mac OS X, when I insert `:` in the YFM post title, build runs smoothly and post is well displayed with a title containing `:`.

NerdyProjects:
I just verified again.
```
[...]

Issue #5257, opened 2016-08-18
Jekyll exits with 0 when a build fails


jleclanche:
So I created a post something like the following:

```md
---
layout: post
title: Something: Something else
---
...
```

`jekyll build` fails when parsing the title. However, it fails with **exit code 0**.

```
      Generating... 
             Error: could not read file ...: (<unknown>): mapping values are not allowed in this context at line 3 column 21
                    done in 1.752 seconds.
```

This is really bad behaviour as it means, for example, a CI build will pass despite it failing. Leaves the whole thing in a broken state.

envygeeks:
Quote the title, you can't have a double ":" in YAML without quotes or it assumes it's parsing a k/v. Please ask non-bug user questions at https://talk.jekyllrb.com rather than on this tracker, this tracker is for bugs only.

jleclanche:
@envygeeks That's not what the bug is about. Jekyll is **exiting with exit code 0**.

jleclanche:
@envygeeks As I said in the original post, exiting with 0 is incorrect behaviour - can you reopen please.

envygeeks:
Unfortunately, no.  There is no problem here as the behavior is exactly right, and expected.  You did not receive a fatal error.  Jekyll continued to process what it could and completed it's processing as per normal and exited with status 0 as you can see with the "done in n" under the error.  We only throw on fatal errors and the inability to process a file is not fatal to us so we processed what we could and gave you the error so that you could fix it.

jleclanche:
@envygeeks this is *realy really bad* behaviour. If you want to keep that, a flag should at the very least be introduced to always exit 1 if there is any error, and not try to continue.

envygeeks:
Remember that next time your kernel can't load a module and it continues to boot or one of your PCI devices has an error and the kernel recovers and keeps running in a sane state.  Or the next time your IDE doesn't exit when it has a non-fatal error.

jleclanche:
@envygeeks These are the most ridiculous comparisons I've ever heard.

Don't make such behaviour default if you're concerned about backwards compatibility, but you *should* provide a flag for users who care about their sanity and the sanity of their data...

envygeeks:
Thanks, hope you have a great day.

Mischanix:
[...]

Issue #4713, opened 2016-03-25
`jekyll build` exit code not reflecting errors


felicianotech:
###### What version of Jekyll are you using (`jekyll -v`)?

3.1.2

###### What operating system are you using?

Ubuntu 12.04/Ubuntu 14.04

###### What did you do?
(Please include the content causing the issue, any relevant configuration settings, and the command you ran)

If you run `jekyll build`, sometimes you can get the error "Error: could not read file" when the file's Front Matter is not correct YAML.

###### What did you expect to see?

When this happens, I expect the 'exit code' returned by Jekyll to be something other than 0.

###### What did you see instead?

Exit code '0'.

###### Why this matters?
When running `jekyll build` as a step on my CI environment, when this error occurs the command just returns 0. This makes my CI assume everything went well and continues. By returning a non-zero exit code, this signifies that something went wrong (aka an error) and stops everything.

parkr:
This will come with #4674. We are picking and choosing which errors to raise a stink about and which errors not to raise a stink about. We will likely move toward more errors rather than fewer errors so that we can utilize system indicators like status codes better.

All in due time. 🔜

PR #5832, opened 2017-01-30
Add option to fail a build with front matter syntax errors

jmhooper:
This is a feature request by way of a PR for a problem I've had with Jekyll in past.

I have a number of sites that are deployed from a CI. The setup is functionally equivalent to the results of following [Jekyll's own documentation on using a CI](http://jekyllrb.com/docs/continuous-integration/travis-ci/):

> The simplest test script simply runs jekyll build and ensures that Jekyll doesn’t fail to build the site. It doesn’t check the resulting site, but it does ensure things are built properly.

The issue I run into is that `jekyll build` is not sufficient to verify that a site is building as expected. If there is a syntax error in a page's front matter, Jekyll will print a warning, ignore the page, and continue.

This repeatedly causes problems for me when people add a page and add something like this in the front matter:

```yaml
title: The question: Will this build
description: 'Tough question. It's answer may surprise you'
```

The build succeeds on CI, and then the site is deployed with that page missing, which is not expected behavior.

This situation has been mentioned a few times in issues (jekyll/jekyll#5257, jekyll/jekyll#5485, jekyll/jekyll#4713). Those have all been closed.

Additionally, there is WIP PR that looks like it may pick this up (jekyll/jekyll#4674), but it also looks like there hasn't been any movement there since March 2016.

Here's the message for the commit in this PR which describes what it does:

>This commit adds a config named `strict_front_matter`. This configuration causes a build to fail if there is a syntax error in a page's front matter.

> Prior to this commit, if a page had an error in its front matter the build would print a warning, ignore the file, and continue. This caused problems where pages would have a syntax error and the build would not fail, and instead build an incomplete site.

> This commit adds the option which tells a build to fail when there is a syntax error reading / parsing a page's front matter. This is done by setting the `strict_front_matter` config or using the `--strict_front_matter` CLI option.

> This commit also adds a test case for this behavior and updates the documentation.

Thanks for reading! Hope you'll consider this change!

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -43,8 +43,6 @@ Metrics/MethodLength:
   CountComments: false
   Max: 20
   Severity: error
-  Exclude:
-    - lib/jekyll/command.rb
 Metrics/ModuleLength:
   Max: 240
 Metrics/ParameterLists:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -43,6 +43,8 @@ Metrics/MethodLength:
   CountComments: false
   Max: 20
   Severity: error
+  Exclude:
+    - lib/jekyll/command.rb
 Metrics/ModuleLength:
   Max: 240
 Metrics/ParameterLists:
--- a/docs/_docs/configuration.md
+++ b/docs/_docs/configuration.md
@@ -305,6 +305,18 @@ class="flag">flags</code> (specified on the command-line) that control them.
         <p><code class="flag">--profile</code></p>
       </td>
     </tr>
+    <tr class="setting">
+      <td>
+        <p class="name"><strong>Strict Front Matter</strong></p>
+        <p class="description">
+            Cause a build to fail if there is a YAML syntax error in a page's front matter.
+        </p>
+      </td>
+      <td class="align-center">
+        <p><code class="option">strict_front_matter: BOOL</code></p>
+        <p><code class="flag">--strict_front_matter</code></p>
+      </td>
+    </tr>
   </tbody>
 </table>
 </div>
@@ -601,12 +613,13 @@ collections:
     output:   true
 
 # Handling Reading
-safe:         false
-include:      [".htaccess"]
-exclude:      ["Gemfile", "Gemfile.lock", "node_modules", "vendor/bundle/", "vendor/cache/", "vendor/gems/", "vendor/ruby/"]
-keep_files:   [".git", ".svn"]
-encoding:     "utf-8"
-markdown_ext: "markdown,mkdown,mkdn,mkd,md"
+safe:                 false
+include:              [".htaccess"]
+exclude:              ["Gemfile", "Gemfile.lock", "node_modules", "vendor/bundle/", "vendor/cache/", "vendor/gems/", "vendor/ruby/"]
+keep_files:           [".git", ".svn"]
+encoding:             "utf-8"
+markdown_ext:         "markdown,mkdown,mkdn,mkd,md"
+strict_front_matter: false
 
 # Filtering Content
 show_drafts: null
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -6,71 +6,72 @@ module Jekyll
     # Strings rather than symbols are used for compatibility with YAML.
     DEFAULTS = Configuration[{
       # Where things are
-      "source"            => Dir.pwd,
-      "destination"       => File.join(Dir.pwd, "_site"),
-      "plugins_dir"       => "_plugins",
-      "layouts_dir"       => "_layouts",
-      "data_dir"          => "_data",
-      "includes_dir"      => "_includes",
-      "collections"       => {},
+      "source"              => Dir.pwd,
+      "destination"         => File.join(Dir.pwd, "_site"),
+      "plugins_dir"         => "_plugins",
+      "layouts_dir"         => "_layouts",
+      "data_dir"            => "_data",
+      "includes_dir"        => "_includes",
+      "collections"         => {},
 
       # Handling Reading
-      "safe"              => false,
-      "include"           => [".htaccess"],
-      "exclude"           => %w(
+      "safe"                => false,
+      "include"             => [".htaccess"],
+      "exclude"             => %w(
         Gemfile Gemfile.lock node_modules vendor/bundle/ vendor/cache/ vendor/gems/
         vendor/ruby/
       ),
-      "keep_files"        => [".git", ".svn"],
-      "encoding"          => "utf-8",
-      "markdown_ext"      => "markdown,mkdown,mkdn,mkd,md",
+      "keep_files"          => [".git", ".svn"],
+      "encoding"            => "utf-8",
+      "markdown_ext"        => "markdown,mkdown,mkdn,mkd,md",
+      "strict_front_matter" => false,
 
       # Filtering Content
-      "show_drafts"       => nil,
-      "limit_posts"       => 0,
-      "future"            => false,
-      "unpublished"       => false,
+      "show_drafts"         => nil,
+      "limit_posts"         => 0,
+      "future"              => false,
+      "unpublished"         => false,
 
       # Plugins
-      "whitelist"         => [],
-      "plugins"           => [],
+      "whitelist"           => [],
+      "plugins"             => [],
 
       # Conversion
-      "markdown"          => "kramdown",
-      "highlighter"       => "rouge",
-      "lsi"               => false,
-      "excerpt_separator" => "\n\n",
-      "incremental"       => false,
+      "markdown"            => "kramdown",
+      "highlighter"         => "rouge",
+      "lsi"                 => false,
+      "excerpt_separator"   => "\n\n",
+      "incremental"         => false,
 
       # Serving
-      "detach"            => false, # default to not detaching the server
-      "port"              => "4000",
-      "host"              => "127.0.0.1",
-      "baseurl"           => "",
-      "show_dir_listing"  => false,
+      "detach"              => false, # default to not detaching the server
+      "port"                => "4000",
+      "host"                => "127.0.0.1",
+      "baseurl"             => "",
+      "show_dir_listing"    => false,
 
       # Output Configuration
-      "permalink"         => "date",
-      "paginate_path"     => "/page:num",
-      "timezone"          => nil, # use the local timezone
+      "permalink"           => "date",
+      "paginate_path"       => "/page:num",
+      "timezone"            => nil, # use the local timezone
 
-      "quiet"             => false,
-      "verbose"           => false,
-      "defaults"          => [],
+      "quiet"               => false,
+      "verbose"             => false,
+      "defaults"            => [],
 
-      "liquid"            => {
+      "liquid"              => {
         "error_mode" => "warn",
       },
 
-      "rdiscount"         => {
+      "rdiscount"           => {
         "extensions" => [],
       },
 
-      "redcarpet"         => {
+      "redcarpet"           => {
         "extensions" => [],
       },
 
-      "kramdown"          => {
+      "kramdown"            => {
         "auto_ids"      => true,
         "toc_levels"    => "1..6",
         "entity_output" => "as_char",
--- a/lib/jekyll/convertible.rb
+++ b/lib/jekyll/convertible.rb
@@ -48,8 +48,10 @@ module Jekyll
         end
       rescue SyntaxError => e
         Jekyll.logger.warn "YAML Exception reading #{filename}: #{e.message}"
+        raise e if self.site.config["strict_front_matter"]
       rescue => e
         Jekyll.logger.warn "Error reading file #{filename}: #{e.message}"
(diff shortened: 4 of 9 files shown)


peternewman:
Thanks for this, I'll certainly be grateful of it.

It looks like you've failed some of their lint tests, at least one of which is unrelated to you and has been fixed upstream.

wjdp:
Just to add a thought to this. I debugged some work an editor had done today where they added a front matter variable and then observed their change didn't take effect. What'd happened was a second definition of that variable was present father down the front matter, which hadn't been noticed. This masked any change the editor made.

Long story short: what would be fantastic is for strict checking to also throw an error if a variable is defined twice.

lib/jekyll/convertible.rb
```diff
@@ -48,8 +48,10 @@ def read_yaml(base, name, opts = {})
         end
       rescue SyntaxError => e
         Jekyll.logger.warn "YAML Exception reading #{filename}: #{e.message}"
+        raise e if self.site.config["strict_front_matter"]
       rescue => e
         Jekyll.logger.warn "Error reading file #{filename}: #{e.message}"
+        raise e if self.site.config["strict_front_matter"]
```
parkr: This handling also needs to be added to `Jekyll::Document#read`, I believe.

jmhooper: Cool. Saw that, but wasn't sure where `Jekyll::Document#read` was used so I left it off. Commit to add it is up now.

jmhooper:
Polite bump.

I made the requested changes in f904fcc8d4f0ed2fc50b0fda352d58a9e293856a

peternewman:
I think they'll want you to fix the lint issues Travis has found:
https://travis-ci.org/jekyll/jekyll/jobs/200753547

lib/jekyll/document.rb
```diff
@@ -258,9 +258,11 @@ def read(opts = {})
           read_post_data
         rescue SyntaxError => e
           Jekyll.logger.error "Error:", "YAML Exception reading #{path}: #{e.message}"
+          raise e if self.site.config["strict_front_matter"]
```
ashmaroli: `self` is not required here and in the line further below.
```diff
- raise e if self.site.config["strict_front_matter"]
+ raise e if site.config["strict_front_matter"]
```

I also think multiple `raise`s can be avoided by using the `or` operator:
```ruby
    raise e if (e.is_a? Jekyll::Errors::FatalException || site.config["strict_front_matter"])
    Jekyll.logger.error "Error:", "could not read file #{path}: #{e.message}"
```

/cc @pathawks

docs/_docs/configuration.md
```diff
@@ -601,12 +613,13 @@ collections:
...
 # Handling Reading
-safe:         false
-include:      [".htaccess"]
-exclude:      ["node_modules", "vendor/bundle/", "vendor/cache/", "vendor/gems/", "vendor/ruby/"]
-keep_files:   [".git", ".svn"]
-encoding:     "utf-8"
-markdown_ext: "markdown,mkdown,mkdn,mkd,md"
+safe:                false
+include:             [".htaccess"]
+exclude:             ["node_modules", "vendor/bundle/", "vendor/cache/", "vendor/gems/", "vendor/ruby/"]
```
ashmaroli: Note: To be fixed on `master` or via another pull

`exclude:` array is now `["Gemfile", "Gemfile.lock", "node_modules", "vendor/bundle/", "vendor/cache/", "vendor/gems/", "vendor/ruby/"]`

/cc @DirtyF

parkr:
There seem to be a conflict – can you please take a look?

DirtyF:
@parkr fixed (was because of #5947)

peternewman:
@DirtyF there still seem to be some lint issues I think they'll want to you fix before merging. And a test failure which I suspect isn't your fault.

DirtyF:
@jmhooper We're interested in shipping this one the tests passes. Care to fix those?

jmhooper:
Thanks for all the great feedback! I've put aside some time this weekend for this one

peternewman:
@DirtyF have you seen this, then you can force Travis to be green before merging, or at least clarify things for contributors:
https://docs.travis-ci.com/user/customizing-the-build#Rows-that-are-Allowed-to-Fail

jmhooper:
I pushed a commit to fix some of the linter issues. A few things to note:

1. I added changes I proposed in #5828 to this, along with changes asked for there (rescues from `Psych::SyntaxError` as well as `SyntaxError`)
2. I excluded `command.rb` from the method length cop in `.rubocop.yml`. The `add_build_options` method was over 20 lines and I couldn't see a way to get it below that. Lmk if that's not acceptable and I can break it up, but I am not sure what the best way to do that is.
3. I reduced the complexity of `read` in `document.rb` by adding a private method to handle errors. I decided to go that route to keep the descriptive error message about YAML parsing issues.

The tests and the linter are both green at home for me on ruby 2.3.

jmhooper:
Looks like Travis blew up while trying to install faraday? Not sure if that's something I touched.

ashmaroli:
> Looks like Travis blew up while trying to install faraday? Not sure that's something I touched.

@jmhooper the `checksum` issue with gem `faraday` has been resolved upstream. The tests should run as usual when one of the maintainers here will restart the ci builds for you.

jmhooper:
I amended the commit to remove the first point above per feedback from @parkr on #5832 🙂

.rubocop.yml
```diff
@@ -41,6 +41,8 @@ Metrics/MethodLength:
   CountComments: false
   Max: 20
   Severity: error
+  Exclude:
+    - lib/jekyll/command.rb
```
parkr: Tsk tsk, that's cheating! Haha. What was the error you got here?

lib/jekyll/document.rb
```diff
@@ -260,10 +260,7 @@ def read(opts = {})
           read_content(opts)
           read_post_data
         rescue SyntaxError => e
```
parkr: Doesn't this have to be `rescue => e` so it captures all errors?

lib/jekyll/document.rb
```diff
@@ -458,6 +455,16 @@ def read_post_data
     end
 
     private
+    def handle_read_error(error)
+      if error.is_a? SyntaxError
+        Jekyll.logger.error "Error:", "YAML Exception reading #{path}: #{error.message}"
+      else
+        Jekyll.logger.error "Error:", "could not read file #{path}: #{error.message}"
+      end
+      raise error if site.config["strict_front_matter"]
```
parkr: This also misses the `Jekyll::Errors::FatalException` exception.

.rubocop.yml
jmhooper: This one was tricky for me. The offending method is here:

```ruby
      def add_build_options(c)
        c.option "config", "--config CONFIG_FILE[,CONFIG_FILE2,...]",
          Array, "Custom configuration file"
        c.option "destination", "-d", "--destination DESTINATION",
          "The current folder will be generated into DESTINATION"
        c.option "source", "-s", "--source SOURCE", "Custom source directory"
        c.option "future", "--future", "Publishes posts with a future date"
        c.option "limit_posts", "--limit_posts MAX_POSTS", Integer,
          "Limits the number of posts to parse and publish"
        c.option "watch", "-w", "--[no-]watch", "Watch for changes and rebuild"
        c.option "baseurl", "-b", "--baseurl URL",
          "Serve the website from the given base URL"
        c.option "force_polling", "--force_polling", "Force watch to use polling"
        c.option "lsi", "--lsi", "Use LSI for improved related posts"
        c.option "show_drafts", "-D", "--drafts", "Render posts in the _drafts folder"
        c.option "unpublished", "--unpublished",
          "Render posts that were marked as unpublished"
        c.option "quiet", "-q", "--quiet", "Silence output."
        c.option "verbose", "-V", "--verbose", "Print verbose output."
        c.option "incremental", "-I", "--incremental", "Enable incremental rebuild."
        c.option "strict_front_matter", "--strict_front_matter",
          "Fail if errors are present in front matter"
      end
```

We get dinged b/c it's over 20 lines. It's right at 21 lines. Coalescing any of the options that cover multiple lines brings the line length over 90 characters. Breaking it into `add_build_options_a_through_m` and `add_build_options_n_through_z` is the next best thing I could think of, which would be ridiculous of course.

I'm welcoming thoughts on this b/c I agree that changing the cop is cheating.

jmhooper:
@parkr: I amended the commit to address your last 2 comments. I also commented on the first ones looking for some guidance.

The fact that the last 2 comments weren't picked up by the tests made me realize that when I did this originally I only added tests to cover `Convertible`. I'd like to add tests to cover the code in `Document` as well, but `test_document.rb` doesn't look like a good fit for that oddly enough? I'm currently working on adding them to `test_site.rb` which seems like a better fit. Interested in any thoughts folks here have on where the test should go.

jmhooper:
For the test, I've got something like this in `test_site.rb`:

```ruby
class TestSite < JekyllUnitTest

  # ...

  context "creating sites" do

    # ...

    context "error handling" do

      # ...

      should "raise for bad frontmatter if strict_front_matter is set" do
        site = Site.new(site_configuration(
          "collections" => ["broken"],
          "strict_front_matter" => true,
        ))
        assert_raises do
          site.process
        end
      end

      should "not raise for bad frontmatter if strict_front_matter is not set" do
        site = Site.new(site_configuration(
          "collections" => ["broken"],
          "strict_front_matter" => false,
        ))
        site.process
      end
    end
  end
end
```

Also note that I've added a folder to the test site named "_broken". It has a single markdown file in it with malformed frontmatter.

I haven't committed yet b/c I'm interested to see how folks feel about where I've got this.

.rubocop.yml
ashmaroli: @jmhooper You can split the `def add_build_options` into smaller methods (IMO):
```ruby
    def add_build_options(c)
      c.option "config", "--config CONFIG_FILE[,CONFIG_FILE2,...]",
        Array, "Custom configuration file"
      c.option "destination", "-d", "--destination DESTINATION",
        "The current folder will be generated into DESTINATION"
      c.option "source", "-s", "--source SOURCE", "Custom source directory"
      c.option "future", "--future", "Publishes posts with a future date"
      c.option "limit_posts", "--limit_posts MAX_POSTS", Integer,
        "Limits the number of posts to parse and publish"
      c.option "watch", "-w", "--[no-]watch", "Watch for changes and rebuild"
      c.option "baseurl", "-b", "--baseurl URL",
        "Serve the website from the given base URL"
      c.option "force_polling", "--force_polling", "Force watch to use polling"
      c.option "lsi", "--lsi", "Use LSI for improved related posts"
      c.option "show_drafts", "-D", "--drafts", "Render posts in the _drafts folder"
      c.option "unpublished", "--unpublished",
        "Render posts that were marked as unpublished"
      c.option "quiet", "-q", "--quiet", "Silence output."
      c.option "verbose", "-V", "--verbose", "Print verbose output."

      additional_build_options(c)
    end

    private

    def additional_build_options(c)
      c.option "incremental", "-I", "--incremental", "Enable incremental rebuild."
      c.option "strict_front_matter", "--strict_front_matter",
        "Fail if errors are present in front matter"
    end
      
```

ashmaroli: If that's not acceptable, there's another hack. `c.option "source"` is technically a duplicate as the `jekyll` executable already adds that option **to all Jekyll Commands**. So you can remove that one line from here.. again, just my 2c

pathawks: > You can split the def `add_build_options` into smaller methods

Nope. Not doing that.

Just exclude this one method from the `MethodLength` cop.

```ruby
# rubocop:disable Metrics/MethodLength
def add_build_options(c)
    ....
end
# rubocop:enable Metrics/MethodLength

Comment (pathawks) [inline .rubocop.yml]: I got you @jmhooper :beers:

File: lib/jekyll/configuration.rb [hunk 7]
```diff
@@ -6,71 +6,72 @@ class Configuration < Hash
...
         Gemfile Gemfile.lock node_modules vendor/bundle/ vendor/cache/ vendor/gems/
         vendor/ruby/
       ),
-      "keep_files"        => [".git", ".svn"],
-      "encoding"          => "utf-8",
-      "markdown_ext"      => "markdown,mkdown,mkdn,mkd,md",
+      "keep_files"          => [".git", ".svn"],
+      "encoding"            => "utf-8",
+      "markdown_ext"        => "markdown,mkdown,mkdn,mkd,md",
+      "strict_front_matter" => false,
```
pathawks: This is the only real change in this file; everything else is alignment :+1:

.rubocop.yml
jmhooper: http://gph.is/2jT5Php

jmhooper:
I went ahead and pushed the test for the config on the site that I was talking about above ☝️

jmhooper:
If anyone stumbles on this and needs the functionality, or wants to add it to a site on an older version of Jekyll, I've created a plugin that mimics the behavior: https://github.com/jmhooper/jekyll_strict_front_matter

parkr:
@jekyllbot: merge +minor

PR #5836, opened 2017-01-31
Disable `gems` array in new config file by default

ashmaroli:
I recently realized that its not necessary for a plugin to be listed in both `_config.yml` and `Gemfile` for it to be used in the build process.

If a plugin is listed in the `config.yml` but not in the `Gemfile`, running Jekyll commands with **`bundle exec`** throws an error. 
There's no such event when a plugin is listed under the `:jekyll_group` alone.

This PR aims to promote the `:jekyll_plugins` group in the `Gemfile` as the primary location to define plugins.

`Friction:`  I concede, **theme-gem developers** will find it easier to have the plugins listed in a `_config.yml` and the gems marked as `runtime_dependencies` in the `gemspec`

/cc @jekyll/ecosystem

### DIFF
--- a/lib/jekyll/commands/new.rb
+++ b/lib/jekyll/commands/new.rb
@@ -82,7 +82,7 @@ gem "minima", "~> 2.0"
 
 # If you have any plugins, put them here!
 group :jekyll_plugins do
-   gem "jekyll-feed", "~> 0.6"
+  gem "jekyll-feed", "~> 0.6"
 end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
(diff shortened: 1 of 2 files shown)


pathawks:
What if I deploy to GitHub pages and a plugin is listed in Gemfile but not config?

ashmaroli:
> What if I deploy to GitHub pages and a plugin is listed in Gemfile but not config?

I'm guessing since `github-pages` already uses a constrained set of gems it doesn't hence require `Bundler` to further resolve the various dependencies, ergo, no `PluginManager.require_from_bundler` coming into play, ergo, my proposal here falls flat on its nose.  :disappointed:

parkr:
Thanks for the proposal! As you mentioned, duplicating the activation of plugins here is the ideal way to get things working in both environment – one works well locally and one works for GitHub Pages. Thanks for the PR though! 😄 Keep 'em coming. Your attention to detail is appreciated!

PR #5841, opened 2017-01-31
Use HTTP to allow CSS to display

jhabdas:
The SSL cert or site setup is doing naughty things when landing on the update site over HTTPS. Given there should be a redirect to HTTPS in place when the SSL cert is active moving this back to HTTP should be fine, and improves the current UX.

### DIFF
--- a/docs/_docs/migrations.md
+++ b/docs/_docs/migrations.md
@@ -5,4 +5,4 @@ permalink: /docs/migrations/
 
 If you’re switching to Jekyll from another blogging system, Jekyll’s importers
 can help you with the move. To learn more about importing your site to Jekyll,
-visit our [`jekyll-import` docs site](https://import.jekyllrb.com/docs/home/).
+visit our [`jekyll-import` docs site](http://import.jekyllrb.com/docs/home/).


jhabdas:
Said naughty thing...

screen shot 2017-01-31 at 9 01 23 pm

Without HTTPs not so naughty...

screen shot 2017-01-31 at 9 02 34 pm

**Aside:** Should we get @benbalter's WordPress script listed in here somewhere. I noticed he's been doing a stellar job at keeping it updated and documented.

pathawks:
> The SSL cert or site setup is doing naughty things when landing on the update site over HTTPS.

GitHub Pages is setting `site.url` to `http://import.jekyllrb.com`, so the site is trying to load CSS and scripts over an insecure connection, which the browser is not allowing because mixed content.

> Given there should be a redirect to HTTPS in place when the SSL cert is active moving this back to HTTP should be fine

Obviously such a redirect is _not_ in place. I would rather see this problem fixed at its source.

jhabdas:
I'll leave the branch here in case the pragmatic approach becomes favored. I believe this issue has been occurring for at least half a year or more, which may be very off-putting to those looking at Jekyll as an alternative to WordPress.

PR #5842, opened 2017-01-31
Corrected date for version 3.4.0

alfredmyers:
Year should probably be 2017 instead of 2016


DirtyF:
@jekyllbot: merge +site

Issue #5844, opened 2017-01-31
Contribute a Dockerfile


chb0github:
Hi there,

I have a Dockerfile that I would like to contribute. I noticed there was some talk but I didn't find any action. 

- [x] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [x] This is a feature request.
- [x] I am on (or have tested on) ***macOS*** 10+

Incidentally [This bit me](https://github.com/jekyll/jekyll/pull/3542):

parkr:
We have Dockerfiles all set up in https://github.com/jekyll/docker! You should be able to see them on Docker Hub and download them the normal way. Cheers!

PR #5845, opened 2017-01-31
Create Dockerfile

chb0github:
builds as: `docker build -t jekyll .`
Runs as `docker run -d --name jekyll -p 4000:4000 -v $PWD:/usr/app/jekyll jekyll`

issue: #5844

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM ruby:2.4.0-alpine
+RUN apk add --no-cache gcc musl-dev make
+RUN gem install jekyll bundler redcarpet
+RUN mkdir -p /usr/app/jekyll
+WORKDIR /usr/app/jekyll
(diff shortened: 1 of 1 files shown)


Crunch09:
I am not familiar with docker but is maybe [jekyll/docker](https://github.com/jekyll/docker) the correct place to put this into ?

chb0github:
Oh, I didn't see that. Let me have a look

chb0github:
That repo hasn't had a check-in in 6 months. I submitted a PR but really, it's 1 file. Not sure why anyone would have an entirely different repo for this job. 

Since you're not familiar with it @Crunch09 - basically, once you have this file and build the image, anyone can run jekyll without having to install ruby or any gems as long as they have [docker](https://www.docker.com/).

parkr:
The jekyll/docker repo was updated 3 days ago and I'd like to keep user-facing Dockerfiles there. Thanks!

PR #5848, opened 2017-02-01
hard-wrap long logger-messages neatly

ashmaroli:
Hi,
This PR proposes to alter `Jekyll.logger` messages by hard-wrapping long messages (longer than terminal-window width) neatly as part of `logger[message]`.

PROS:
  - Improves output readability for users on terminal window set to about `80 cols` (default on Windows, at least, it is till Windows 7)
  - opens up outputting `HEREDOC` formatting of messages

CONS:
  - adds one more dependency ([tty-screen](https://github.com/piotrmurach/tty-screen)) to Jekyll's gemspec
  - changes something thats been around for years.

--

### Screencaps:

BEFORE PR -- Default View

cmd

AFTER PR -- Default View

cmd_1b

--
BEFORE PR -- Exception raised

cmd_2

AFTER PR -- Exception raised

cmd_2b

--
/cc @jekyll/core 
/cc @jekyll/ecosystem

### DIFF
--- a/jekyll.gemspec
+++ b/jekyll.gemspec
@@ -39,4 +39,5 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency("pathutil",              "~> 0.9")
   s.add_runtime_dependency("rouge",                 "~> 1.7")
   s.add_runtime_dependency("safe_yaml",             "~> 1.0")
+  s.add_runtime_dependency("tty-screen",            "~> 0.5")
 end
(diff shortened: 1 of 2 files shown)


envygeeks:
It should be 78 chars, data centre consoles (well the older ones -- and older computers too) wrapped at 80 but we always had to set the wrap to 78 because of madness.

ashmaroli:
I stumbled upon the same conclusion.. though indirectly..
I too found it optimal to have the `[message] `string split at `width - rjust value - 2cols` essentially being `80 - 20 - 2` at default setting.

PR #5857, opened 2017-02-04
Enable support for tags and categories across all documents

paulrobertlloyd:
## Problem
While posts are supposedly a default collection type, they still enjoy a number of benefits not available to other documents that appear within custom collections. One such example is that tags or categories added to documents do not get listed under `site.tags.TAG` or `site.categories.CATEGORY`. Furthermore, this inconsistency has manifested itself elsewhere, namely in the [jekyll-archives](https://github.com/jekyll/jekyll-archives/) plugin, which doesn’t support archiving documents outside of the posts collection.

In fact, it was my attempt to fix this issue that led me here. See the conversation and [proof of concept posted to that repo](https://github.com/jekyll/jekyll-archives/pull/88) for details.

## Proposed solution
[Following on from @alfredxing’s suggestion](https://github.com/jekyll/jekyll-archives/pull/88#issuecomment-274953297), I have added a new utility function `doc_attr_hash`. This mimics the function of `post_attr_hash` (which has been refactored to use this new function).

As this PR stands, `site.tags.TAG` and `site.categories.CATEGORY` will now look across all documents, rather than just posts (I have updated the documentation accordingly), while continuing to provide the `post_attr_hash` function for plugins and other consumers that may be referring to it.

This seems like a good compromise, but might this still be considered a breaking change? The alternative would be to leave `site.tags.TAG` and `site.categories.CATEGORY` as is (at least until the next major version), yet provide the new `doc_attr_hash` function for those that require that functionality (such as an improved, collections compatible archive plugin).

I look forward to your comments and review.

### DIFF
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -215,36 +215,45 @@ module Jekyll
       collections["posts"] ||= Collection.new(self, "posts")
     end
 
-    # Construct a Hash of Posts indexed by the specified Post attribute.
+    # Construct a Hash of Documents indexed by the specified Document attribute.
     #
-    # post_attr - The String name of the Post attribute.
+    # doc_attr - The String name of the Document attribute.
+    # documents - The Array of Documents
     #
     # Examples
     #
-    #   post_attr_hash('categories')
-    #   # => { 'tech' => [<Post A>, <Post B>],
-    #   #      'ruby' => [<Post B>] }
+    #   doc_attr_hash('categories')
+    #   # => { 'tech' => [<Document A>, <Document B>],
+    #   #      'ruby' => [<Document B>] }
     #
-    # Returns the Hash: { attr => posts } where
-    #   attr  - One of the values for the requested attribute.
-    #   posts - The Array of Posts with the given attr value.
-    def post_attr_hash(post_attr)
-      # Build a hash map based on the specified post attribute ( post attr =>
-      # array of posts ) then sort each array in reverse order.
+    # Returns the Hash: { attr => docs } where
+    #   attr - One of the values for the requested attribute.
+    #   docs - The Array of Documents with the given attr value.
+    def doc_attr_hash(doc_attr, documents)
+      # Build a hash map based on the specified document attribute
+      # ( doc_attr => array of docs ) then sort each array in reverse order.
       hash = Hash.new { |h, key| h[key] = [] }
-      posts.docs.each do |p|
-        p.data[post_attr].each { |t| hash[t] << p } if p.data[post_attr]
+
+      documents.each do |d|
+        d.data[doc_attr].each { |t| hash[t] << d } if d.data[doc_attr]
       end
-      hash.values.each { |posts| posts.sort!.reverse! }
+
+      hash.values.each { |documents| documents.sort!.reverse! }
       hash
     end
 
+    # As doc_attr_hash, but for posts only
+    # (Provided for backwards compatibility)
+    def post_attr_hash(post_attr)
+      doc_attr_hash(post_attr, posts.docs)
+    end
+
     def tags
-      post_attr_hash("tags")
+      doc_attr_hash("tags", documents)
     end
 
     def categories
-      post_attr_hash("categories")
+      doc_attr_hash("categories", documents)
     end
 
     # Prepare site data for site payload. The method maintains backward compatibility
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -238,7 +238,7 @@ module Jekyll
         d.data[doc_attr].each { |t| hash[t] << d } if d.data[doc_attr]
       end
 
-      hash.values.each { |documents| documents.sort!.reverse! }
+      hash.values.each { |docs| docs.sort!.reverse! }
       hash
     end
 
--- a/docs/_docs/variables.md
+++ b/docs/_docs/variables.md
@@ -173,7 +173,7 @@ following is a reference of the available data.
       <td><p><code>site.categories.CATEGORY</code></p></td>
       <td><p>
 
-        The list of all Posts in category <code>CATEGORY</code>.
+        The list of all Documents in category <code>CATEGORY</code>.
 
       </p></td>
     </tr>
@@ -181,7 +181,7 @@ following is a reference of the available data.
       <td><p><code>site.tags.TAG</code></p></td>
       <td><p>
 
-        The list of all Posts with tag <code>TAG</code>.
+        The list of all Documents with tag <code>TAG</code>.
 
       </p></td>
     </tr>


DirtyF:
Simple, efficient and useful, but IMHO it can be a breaking change for those who don't want to mix posts and collections. 

Even if we added more possibilities to access for instance :

* `site.categories`(`site.posts.categories` + `site.docs.categories`)
* `site.tags`  (`site.posts.tags` + `site.docs.tags` )
* `site.collection.tags`, `site.collection.categories`

/cc @jekyll/core

paulrobertlloyd:
@DirtyF I like the idea of providing more fine-grained options like this, meaning users can scope tag/category parsing to particular collections, although I’m not sure how/where I would provide that functionality in this PR. Maybe a few more eyes could take a look at this, and chime in with suggested routes forward?

parkr:
What's wrong with `{% assign docs_i_want = site.docs | where: "category","myCategory" %}`?

paulrobertlloyd:
@parkr I guess there are two separate (but related) issues being addressed in this PR:

1. Providing a utility function that allows programmatic access to tags and categories across all documents (via `doc_attr_hash`). This is useful for plugin authors wishing to access this information, and seemed like something that should be provided by the core.
2. Clarify the variable space so that `site.tags` and `site.categories` do what they say on the tin. Clearly that is a breaking change, so perhaps something to postpone to a v4.0 release. And as you point out, other methods are possible to achieve the same in the meantime.

Happy to de-scope this PR so that it only addresses the first issue, but thought it worth highlighting another discrepancy that persists throughout Jekyll as we move towards the "posts are a collection" world.

hosnas:
Any update for this feature? I also want to access both post and my collections together by tags

lib/jekyll/site.rb
```diff
@@ -215,36 +215,45 @@ def posts
...
 
+    # As doc_attr_hash, but for posts only
+    # (Provided for backwards compatibility)
+    def post_attr_hash(post_attr)
+      doc_attr_hash(post_attr, posts.docs)
+    end
+
     def tags
-      post_attr_hash("tags")
+      doc_attr_hash("tags", documents)
```
pathawks: > Clearly that is a breaking change, so perhaps something to postpone to a v4.0 release.

Yup.

If `tags` and `categories` were left unchanged, the rest of this would be acceptable.

parkr:
> this is unlikely to ever land before 4.0 in its current state, so i suggest to trim it down in order to land it sooner

@pup @pathawks Can you expound on this further? I've always taken the position that additive changes like what I think I'm reading here are acceptable – they don't break existing behaviour, because in practice no documents other than posts have been allowed to have categories or tags, so there's a fairly high confidence level that if I ran this on 99% of Jekyll sites, nothing would change. The method signatures, of course, cannot be removed (and the change I see at the moment has the `post_attr_hash` method still there). I only see behaviour added, not unexpected & breaking behaviour.

hosnas:
I completely agree with @parkr  . And also I can add that if a collection document has the same tag with a post, perhaps the author want to show the collection in the tag page too. If not, why would a collection have the same tag as other posts?  

I tested this PR and it worked perfectly exactly as I wanted.

pathawks:
@parkr Aside from very obvious bugfixes, I am nervous about any change which would cause a working site to render in a radically different way from what the author intended.

Our documentation clearly states that `site.tags` and `site.categories` deal exclusively with posts. To change that behavior is not merely an additive feature, but changes existing behavior. Just because the new behavior is what one user expects does not mean the new behavior is what _every_ user expects.

**I am not opposed to this change.** I think it would be a good idea to have a way of adding tags and categories to and document. **I only think that this would require a major version bump** to give users a chance to review the new behavior and make any necessary changes to their site.

pathawks:
Now that we're working towards Jekyll 4.0, let's take another look at this one.

ashmaroli:
@paulrobertlloyd I see that you have deleted your fork of the Jekyll Repository.. Can I open a new PR with your changes and build on it or are you willing to revisit this for `Jekyll 4.0`..?

paulrobertlloyd:
@ashmaroli Do what you need to do — excited to see this feature make it into the next version!

ashmaroli:
@paulrobertlloyd Sorry for giving false hope..
As I started writing tests for this, I realized that "my idea" would involve making a lot of changes across multiple places, (therefore breaking current behavior) and the diff you proposed here would end up being just a tiny fraction of the final diff..

So, I withdrew my plan of taking this forward..
Once again, sorry for giving hope..

paulrobertlloyd:
Ah, that’s a shame. Hopefully this feature still makes its way into v4.0.

hosnas:
Is there anything similar or a plugin for this?

MSGoodman:
I needed this functionality on my end, but only for one of my projects. Rather than switch between two local versions of jekyll to use them as needed, my solution (76b55f814670968e1a6836d1933ab42f8654497c) was to add support for two options (<code>categories_for_all_docs</code> and <code>tags_for_all_docs</code>) in a site's <code>_config.yml</code> that would indicate that jekyll should include that site's documents in <code>site.tags</code> and <code>site.categories</code>.

I am unsure whether <code>_config.yml</code> is really the right place for universal settings like this, or where a feature like this would be included in the jekyll documentation, but I figured it might be useful to someone out there and it doesn't seem to me like it would change expected behavior for anyone since it would need to deliberately be "activated," so to speak. I can make a PR for it or fiddle with it further or whatever anyone would like if it seems usable in the main jekyll repo.

In any case, thank you @paulrobertlloyd for the code in this PR as it has been very helpful for me!

paulrobertlloyd:
@MSGoodman Glad you found it useful!

hosnas:
@paulrobertlloyd as it seems this pr isn't going t be merged in near future, can we have this functionality with a plugin?

paulrobertlloyd:
@hosnas Probably, although I’m not sure how. I no longer use Jekyll, so have no interest in creating a plugin, sorry.

ashmaroli:
@hosnas Feel free to open a new full-fledged PR then..
Alternatively, you can open a feature-request ticket detailing *your expected outcome* and some use-case samples

hosnas:
@ashmaroli I finally found [this plugin](https://github.com/sverrirs/jekyll-paginate-v2) exactly do what I need.

ruippeixotog:
I stumbled upon this PR as I was looking for a solution to this problem - I just migrated from using posts to a custom collection, but I noticed that I lost all my tags in the process. Using an alternative expression is not something I can do because many themes use  `site.tags`, as well as many plugins.

One suggestion: have we considered keeping the current default behavior but adding  	an option on `_config.yml` like `collections.places.output_tags`, which would be `false` by default but which users could use to decide which documents they want to produce tags for?

How difficult would it be to implement this? I'd be happy to take this PR if I could get some general guidance on what parts of the code would I need to touch.

parkr:
Could we re-implement something like this? I think it's appropriate to have `site.categories` respond with all documents, as there's not much of a way to use `categories` for non-post documents at this time, so I'd be surprised if lots of folks were using it.

ashmaroli:
While we can easily alter Ruby object `site.tags` to `site.tags(all_docs = true)`, and incur some tech-debt to maintain backwards-compatibility with method calls and returned results, we cannot do the same to Liquid `{{ site.tags }}`.
We will suddenly force our users to implement a filter to separate out posts from other collections (some of which may not even be configured to output to dest).
So we'll need new Liquid variables that do not conflict with existing ones at user's end (because technically, even `{{ site.foo }}` could be conflicting if the user already has a top-level `foo` key in their config file)...

ashmaroli:
IMO, the best route is to have an in-house plugin provide the functionality initially. Once the implementation stabilizes, optionally absorb it into Jekyll Core (or bundle it as a dependency like `jekyll-watch`).

parkr:
An in-house plugin sounds like a good idea! It would monkey-patch things, but I don't think the method has changed in ages. 😄

ashmaroli:
The limiting factor here is not the Ruby front but maintaining compatibility at the Liquid front..

parkr:
Yeah that makes sense! By making it an optional plugin, then it's opt-in and therefore a user would know they are enabling it and that it would affect `site.tags` and `site.categories`.

ashmaroli:
I'm taking a stab at writing the plugin. If someone has already started working on this, do let me know by posting a comment here and I shall assign the task to you.

ashmaroli:
Just shipped an initial release of the plugin:
https://rubygems.org/gems/jekyll-tagories

Issue #5859, opened 2017-02-04
DEFAULTS["exclude"] array overridden by generated site config.


ashmaroli:
It has come to my attention that `vendor/` is not being excluded from being processed when serving the default-site generated by `jekyll new`

On investigating, I realized that this was because the generated `_config.yml` overrode that setting as it contains the following:
```yaml
exclude:
  - Gemfile
  - Gemfile.lock
```
thereby reprogramming Jekyll to only `exclude` the above two files instead of `vendor/*` specified in `Configuration::DEAFULTS`

### To reproduce bug:
  - `jekyll new test-blog`
  - Create `test.md` with following:
```html
<ul>
  {% for item in site.exclude %}
    <li> {{ item }} </li>
  {% endfor %}
</ul>
```
  - `bundle exec jekyll build`
  - inspect `test.html` to find a list of excluded items

--
/cc @jekyll/ecosystem

PR #5860, opened 2017-02-04
Exclude Gemfile by default

ashmaroli:
Exclude `Gemfile` and `Gemfile.lock` by default. I can't think of any situation where they may be needed to be processed.

Additionally, comment out existing `exclude:` array in the generated `_config.yml` and explain what to do with it

Addresses #5859 

--
/cc @jekyll/ecosystem

### DIFF
--- a/lib/site_template/_config.yml
+++ b/lib/site_template/_config.yml
@@ -31,10 +31,8 @@ gems:
   - jekyll-feed
 
 # Exclude from processing.
-#   The items below are excluded by default, to customize the list, uncomment the
-#   lines below and simply add to the list. Items removed will **not** be excluded
-#   from processing.
-#
+# The following items will not be processed, by default. Create a custom list
+# to override the default setting.
 # exclude:
 #   - Gemfile
 #   - Gemfile.lock
--- a/lib/site_template/_config.yml
+++ b/lib/site_template/_config.yml
@@ -29,6 +29,17 @@ markdown: kramdown
 theme: minima
 gems:
   - jekyll-feed
-exclude:
-  - Gemfile
-  - Gemfile.lock
+
+# Exclude from processing.
+#   The items below are excluded by default, to customize the list, uncomment the
+#   lines below and simply add to the list. Items removed will **not** be excluded
+#   from processing.
+#
+# exclude:
+#   - Gemfile
+#   - Gemfile.lock
+#   - node_modules
+#   - vendor/bundle/
+#   - vendor/cache/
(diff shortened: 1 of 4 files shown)


lib/site_template/_config.yml
```diff
@@ -29,6 +29,17 @@ markdown: kramdown
...
 gems:
   - jekyll-feed
-exclude:
-  - Gemfile
-  - Gemfile.lock
+
+# Exclude from processing.
+#   The items below are excluded by default, to customize the list, uncomment the
+#   lines below and simply add to the list. Items removed will **not** be excluded
+#   from processing.
```
parkr: This comment is a little confusing to me. I think what you're driving at is that the defaults go away if you create a custom `exclude` here. Let's try to be really really precise here.

ashmaroli: Yes, that's indeed the intention.

parkr:
@jekyllbot: merge +bug

lib/site_template/_config.yml
```diff
@@ -29,6 +29,15 @@ markdown: kramdown
...
 gems:
   - jekyll-feed
-exclude:
-  - Gemfile
-  - Gemfile.lock
+
+# Exclude from processing.
+# The following items will not be processed, by default. Create a custom list
+# to override the default setting.
+# exclude:
```
pathawks: I wonder if there's anyway this list could be automatically updated with the default exclude list?

ashmaroli: we'll probably have to modify the `deep_merge_hashes` utility method to `append!` when `hash[key].is_a? Array` ...

ashmaroli: .. I never added that patch to `jekyll-data` plugin because then one would never be able to take an item off the list if they needed to..

pathawks: But we would just want it to be comments rather than an actual list.

I just worry about adding files to the exclude list and forgetting to document the change here. We'll just have to be cautious, I guess

ashmaroli: @pathawks, it just struck me.. we could have the list automatically updated if `_config.yml` was generated from an ERB template `_config.yml.erb` : 

```
# exclude:
<% for item in Jekyll::Configuration::DEFAULTS["exclude"] %>
#   - <%= item %>
<% end %>
```
concern: Matt's [comment](https://github.com/jekyll/jekyll/pull/5236#issuecomment-243956535) from many months ago regarding the use of ERB templates..

Your thoughts ?

Issue #5838, opened 2017-01-31
[Proposal] Include `jekyll-default-layout` plugin in core


ashmaroli:
Hi,
I would like to see the `jekyll-default-layout` plugin added to Jekyll Core.

### Why?
  - currently the documents generated by `jekyll new` (**`index.md`, `about.md`, `welcome-to-jekyll.markdown`**) has layouts **tied to Minima** and its not a smooth transition from Minima to one of the other new themes available at GitHub pages.
**Including this plugin in the Core will allow us to generate the concerned documents with the `page.layout` key ( will be commented out, by default ) set by this plugin.**
  - its a lot more practically prompt solution than having each of the new themes to ship named layouts (which will be necessary in the future, nevertheless)
  - already in use at GitHub Pages which is generally vary of running arbitrary code on its servers.

For portability!

Issue #453, opened 2011-12-15
Option for "layout: default" in _config.yml


pdurbin:
For some of my Jekyll sites I'm always using the same layout for all my pages and posts.  It would be nice to be able to specify "layout: default" once in my `_config.yml` and have that layout applied to any file with YAML Front Matter that doesn't have any layout specified explicitly.

I asked a related question on Stack Overflow and answered it with a  plugin that seems to work (i.e. I don't have to write "default: layout" over and over) but I'm no Rubyist and it doesn't handle the `_config.yml` part.  I called it `_plugins/implicit-layout.rb`:

http://stackoverflow.com/questions/8490528/how-can-i-make-jekyll-use-a-layout-without-specifying-it

PR #5863, opened 2017-02-06
Set layouts programmatically

ashmaroli:
This PR proposes to remove explicitly defined `layout` keys from generated pages and have Jekyll rely on @benbalter's [jekyll-default-layout](https://github.com/benbalter/jekyll-default-layout) plugin to have them set programmatically.

### Why? 
  - Allow users to easily *switch* themes using GitHub's **Theme Chooser**
    Currently the feature breaks the build as the provided themes only contain a `default.html` layout instead of separate layouts for various types of documents

### What about third-party theme-gem developers?
  - Since `Jekyll` doesn't enforce a layout naming convention, the various layouts are currently named arbitrarily. This would pave the way to standardize a convention.
  - **NOTE: I have not tested the following statement, but I'm assuming the plugin already takes it into consideration.**
    Instead of having to rename existing theme-gem layout files, the theme's `_config.yml` can set the *default layouts* by using the `defaults:` key:
```yaml
# Defaults
defaults:
  - scope:
      path: ""
      type: posts
    values:
      layout: editorial
  - scope:
      path: ""
      type: pages
    values:
      layout: article
  - scope:
      path: "/"
      type: index
    values:
      layout: home
```
---
Ref: #5838 
/cc @jekyll/ecosystem
/cc @benbalter


ashmaroli:
The next step in solving the "build-break-issue" is to remove the `include` sections within `about.md`

ashmaroli:
label: `downstream:issue`

parkr:
We have tried adding them (https://github.com/jekyll/jekyll/pull/580) and they were removed a short while later (https://github.com/jekyll/jekyll/pull/977). It works fine for GitHub Pages using our themes, but I don't think this should be in Jekyll core. A rationale from many moons ago: https://github.com/jekyll/jekyll/issues/453#issuecomment-16398961

ashmaroli:
Okay.. I agree with the rationale.
Closing.

@parkr One more question: 

---
If a **`layout`** is not to be found during the build process, Jekyll simply warns the user, say: 
```
Build Warning: Layout 'post' requested in _posts/2017-02-11-welcome-to-jekyll.markdown does not
exist
```
and **continues with the processing**. But when an **`include`** is missing, a **`Liquid Exception`** is thrown and **the build process aborted.**
Shouldn't we be a bit more lenient with missing `include`s too, by default? That'd remove the entire *blocking* I mentioned in the PR body.

---

parkr:
> Shouldn't we be a bit more lenient with missing includes too, by default?

The user experience there should really be focused around ensuring the expected behaviour occurs or it's insanely obvious why things aren't working. When we transparently ignore things with only log output, there are plenty of places that you won't catch that until it's too late (like when you deploy your site to any of the hosting providers who don't share build logs). I am generally in the "fail fast and loud" camp so that these kinds of things are caught. Essentially, you're telling the computer to "Include this file" and it can't – it should stop and say "hey! you asked me to do this and I can't so can you please clarify?" It may seem annoying, but it's a lot easier to track down than just a blank page.

ashmaroli:
in that case we'll have to remove the two `include` blocks in the `about.md` template for `jekyll new` Its a source of frustration when switching to a theme other than `minima`

ashmaroli:
> but it's a lot easier to track down than just a blank page

its already a blank page when the `layout` couldn't be loaded.. but to choke when a small snippet of code wasn't found... that's an odd behavior..

PR #5864, opened 2017-02-06
Fixed example filter pluging code

danblakemore:
I generally like following contribution guidelines closely, but this change goes one of two ways: I'm right and it's a tiny doc change, or not.


pathawks:
`module` is correct, as you can [see from our own code](https://github.com/jekyll/jekyll/blob/31178bfe694ba18eeaf11fe0818a82c10bc4df77/lib/jekyll/filters.rb#L9)

Thanks for looking out for us though.

(cloud butt 🤣)

PR #5870, opened 2017-02-08
Add Utils::Internet.connected? to determine whether host machine has internet connection.

parkr:
This PR adds a POC for a common API for checking whether the user has an
Internet connection wherever they're running Jekyll.

Many plugins make calls out into the Internet for data. In order to facilitate a pattern
for gracefully falling back when there is no Internet connection for the user, this module
and module function are added. Plugins which access the Internet should decide whether there
is a path for them to gracefully fallback to some non-Internet default when the host computer
is not connected.

A lot of the work we're doing with GitHub Pages (jekyll/github-metadata,
and things like https://github.com/github/pages-gem/pull/408) require
the internet. We'd like to be able to fall back gracefully when there's
no internet connection, and a common API makes the most sense going
forward.

- [ ] Decide on logical timeouts (connect & open timeouts) and implement
- [ ] Add more robust tests

/cc @jekyll/ecosystem


lib/jekyll/utils/internet.rb
```diff
@@ -0,0 +1,35 @@
...
+      #   else
+      #     Jekyll.logger.warn "Warning:", "Version check has been disabled."
+      #     Jekyll.logger.warn "", "Connect to the Internet to enable it."
+      #     nil
+      #   end
+      #
+      # Returns true if a DNS call can successfully be made, or false if not.
+      module_function
+      def connected?
+        !dns("google.com").nil?
```
pnn: this might be unnecessary, but is google really the right website to ping here? isn't there some sort of website specifically designed for this purpose?

parkr: If you can find it, I will gladly update it! This is usually very fast and Google doesn't often botch their DNS. Seemed like the site I could think of with the best uptime!

envygeeks: `jekyllrb.com`?

mattr-: example.com?

envygeeks: google.com is more available than example.com.  Google is probably the best place to test as it's the most highly available site on the planet.  Just my 2p, but tbqf it's mostly non-relevant, all things considered, how often does example.com get taken out? Only once I know of.

pathawks:
This is good.

I wonder if, in the future, we should add an option to never connect to the internet. Maybe set an environment variable to always return false.

One limitation of this approach is that it will appear there is an internet connection if the user is connected but not authenticated to a hotspot that has a gateway where every request returns a webpage asking the user to agree to terms of service. I think, for our purposes, that is probably not a very big deal.

parkr:
> I wonder if, in the future, we should add an option to never connect to the internet. Maybe set an environment variable to always return false.

That's a good idea. This check doesn't mean we can actually make a TCP connection anywhere useful, but at least we can get situations where a user's wifi is turned off, etc.

@jekyllbot: merge +minor

PR #5871, opened 2017-02-08
Convert StaticFile liquid representation to a Drop & add front matter defaults support to StaticFiles

benbalter:
Per https://github.com/jekyll/jekyll-sitemap/pull/157#issuecomment-278429052, this PR implements a new StaticFileDrop such that static files can inherit front matter defaults.

I believe this is implemented as a non-breaking change. The resulting hash should be identical, with a new `collection` field added (to more closely mimic the DocumentDrop).

I also tried to obtain as much parity with how Documents implement front matter defaults, by spiking out a new `data` method, rather than simply delegating to the existing `defaults` method for fallback data (which we could do if we don't want static files to have user-modifiable metadata beyond defaults).

### DIFF
--- a/lib/jekyll/drops/static_file_drop.rb
+++ b/lib/jekyll/drops/static_file_drop.rb
@@ -1,0 +1,11 @@
+module Jekyll
+  module Drops
+    class StaticFileDrop < Drop
+      extend Forwardable
+      def_delegators :@obj, :name, :extname, :modified_time, :basename
+      def_delegator :@obj, :relative_path, :path
+      def_delegator :@obj, :data, :fallback_data
+      def_delegator :@obj, :type, :collection
+    end
+  end
+end
--- a/test/test_static_file.rb
+++ b/test/test_static_file.rb
@@ -148,8 +148,9 @@ class TestStaticFile < JekyllUnitTest
         "extname"       => ".txt",
         "modified_time" => @static_file.modified_time,
         "path"          => "/static_file.txt",
+        "collection"    => nil
       }
-      assert_equal expected, @static_file.to_liquid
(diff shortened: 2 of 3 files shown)


lib/jekyll/drops/static_file_drop.rb
```diff
@@ -0,0 +1,11 @@
+module Jekyll
+  module Drops
+    class StaticFileDrop < Drop
+      extend Forwardable
+      def_delegators :@obj, :name, :extname, :modified_time, :basename
+      def_delegator :@obj, :relative_path, :path
```
parkr: This may not be 100% the same as below. We have to ensure this has the forward slash at the start!

test/test_static_file.rb
```diff
@@ -148,8 +148,9 @@ def setup_static_file_with_defaults(base, dir, name, defaults)
         "extname"       => ".txt",
         "modified_time" => @static_file.modified_time,
         "path"          => "/static_file.txt",
+        "collection"    => nil
```
parkr: Can we add a test for a static file in a collection, please? That would help clarify this item's behavior.

lib/jekyll/drops/static_file_drop.rb
benbalter: I believe it will, based on https://github.com/jekyll/jekyll/blob/master/lib/jekyll/static_file.rb#L29, no?

benbalter:
>  We should ensure that the collection item is either a string (its label) or the value of Collection#to_liquid

Collection is being called via `type` which is [either nil or the sym label](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/static_file.rb#L135)

parkr:
@jekyllbot: merge +minor

PR #5876, opened 2017-02-10
fix ial css classes in theme doc

marcanuy:
Classes are not being generated in IALs <https://jekyllrb.com/docs/themes/>


DirtyF:
@jekyllbot: merge +docs

PR #5880, opened 2017-02-11
Update installation.md

BlueberryFoxtrot:
It --> Jekyll.  
Avoids misreading that "it is possible to get [Windows] running on Windows."


DirtyF:
@jekyllbot: merge +docs

PR #5881, opened 2017-02-11
New tutorial: Convert an HTML site to Jekyll

tomjohnson1492:
This tutorial explains how to convert an HTML site to Jekyll. We need this tutorial for several reasons. First, theming is one of Jekyll's strengths and should be explained more clearly in the docs. Second, there aren't that many Jekyll themes, but there are thousands of HTML web templates available. Finally, many tech writers that use Jekyll need to clone their company sites to provide seamless branding of their documentation content. They need this tutorial.

I know there are some existing tutorials online for creating Jekyll sites, but here I've tried to simplify the process and make it easier. This tutorial will be added to the Tutorials navigation.

This tutorial should be listed in the Tutorials nav. I'm waiting on [PR 5698](https://github.com/jekyll/jekyll/pull/5698) to go live before I can start adding links to that navigation.


ashmaroli:
I like it.  :+1: 

Now, if only there was a plugin to automate this..........

DirtyF:
That's how I started using Jekyll in the first place, by iteration : 

- add a `_config.yml` file in my existing static site
- split  `{{content}}` and layouts in pages
- split `layouts` in `includes` files when necessary (head, header, footer, etc.)
- use variables in layouts and includes (`_config.yml`, `_data` or `YFM`)
- add basic plugins like jekyll-feed, jekyll-sitemap, jekylll-seo-tag, etc.
- etc.

I liked how you can simply start from HTML and progressively turn your static site into a Jekyll website (and not the other way around).

tomjohnson1492:
This PR hasn't really progressed. Is there anything blocking it from being merged?

_tutorials/convert-html-site-to-jekyll.md
```diff
@@ -0,0 +1,362 @@
...
+* Jekyll looks for files with front matter tags (like those in `index.md`) and [processes the files]({% link _tutorials/orderofinterpretation.md %}) (populating site variables, rendering any Liquid, and converting Markdown to HTML).
+* Jekyll pushes the content from all pages and posts into the `{% raw %}{{ content }}{% endraw %}` tags in the layout specified (`default`) in the front matter tags.
+* The processed files get written as `.html` files in the `_site` directory.
+
+With this basic foundation of how a Jekyll site works, you can convert almost any HTML theme into a Jekyll site. The following sections will walk you through a step-by-step tutorial on converting an HTML template into a Jekyll site.
+
+## 2. Establish the Default Layout
+
+Find your HTML theme and save it as a default layout. If you're converting or cloning (with permission to do so) an existing site, you can right-click the page and view the source code. Copy and paste the source code into a file called `default.html` inside a folder called `_layouts`. This will be the layout template for your pages and posts.
```
pathawks: > (with permission to do so)

This sounds scary. What are you trying to communicate?

_tutorials/convert-html-site-to-jekyll.md
```diff
@@ -0,0 +1,362 @@
...
+
+Note that in looking for templates, you want the HTML output of the template. If the template has PHP tags or other dynamic scripts, these dynamic elements will need to be converted to HTML or stripped out, since Jekyll handles only static content.
+
+Open `default.html` into your browser locally to ensure the site looks and functions like it does online. You will likely need to adjust CSS, JS, and image paths so they work. 
+
+If the paths were relative on the site you copied, you'll need to either download the same assets into your Jekyll site or use absolute paths to the same assets in the cloud. (Note that syntax such as `src="//` requires a prefix such as `src="http://` to work in your local browser.)
```
pathawks: > (Note that syntax such as `src="//` requires a prefix such as `src="http://` to work in your local browser.)

What do you mean? If using `jekyll serve`, the protocol will be `http://` already.
I agree that protocol relative URLs are an anti-pattern, but not sure what that has to do with this.

pathawks:
There is good stuff in here, but it seems more focused on a "My First Jekyll Site" than anything to do specifically with refactoring an existing site into a Jekyll site.

Also remember that anything that is currently a static site is _already_ a Jekyll site. There is nothing special that needs to be done for Jekyll to be able to render a site. This is part of the magic of GitHub Pages: a user can just push a static site and everything just works, even if the user has never heard of Jekyll.

_tutorials/convert-html-site-to-jekyll.md
tomjohnson1492: I'm addressing a scenario where you have an existing static site (like your company site) and you want to convert it into a Jekyll site. For example, I recently did this with a podcast site. The original site looked like this: [writethedocs.org/(http://www.writethedocs.org/). It's a sphinx-generated site. I cloned it to Jekyll to create a podcast-feature on the site: [podcast.writethedocs.org/](http://podcast.writethedocs.org/). The podcast page (Jekyll-powered) is an exact clone of the sphinx theme.

There are loads of people looking to do this. When I was in the WordPress world, I had a [post on cloning your site into a WP site](http://idratherbewriting.com/2008/10/23/seamlessly-integrating-a-blog-into-your-non-blog-website/) -- it was the most popular post on my site for a long time.

_tutorials/convert-html-site-to-jekyll.md
tomjohnson1492: When you copy the source code of a site, it won't render locally because some of the paths to the assets will no longer valid. For example, a path to a CSS reference might look like this in the source code you cloned:

```
src="//assets/mystyles.css"
```

This path won't work when you build your Jekyll site. You would have to either populate it with the http URL before `assets/mystyles.css`, like this: `http://somesite.com/assets/mystyles.css`. Or you'd have to download mystyles.css and use a relative path to it, like this: `/assets/mystyles.css`. Or even `{{site.baseurl}}/assets/mystyles.css`.

tomjohnson1492:
Do you want me to change the title and context? Or should I can close the PR and just publish this tutorial on my own site instead? There's a scarcity of Jekyll themes, and I'm trying to remedy this by helping people understand how easy it is to convert an existing static HTML site into a Jekyll site. Maybe I should include more info up front about how Jekyll can seamlessly plug into any existing site to help automate things?

_tutorials/convert-html-site-to-jekyll.md
pathawks: What I mean is: from whom do I need permission?

pathawks:
@tomjohnson1492 Sorry, I understand you're a bit frustrated that this hasn't gone anywhere.

Perhaps my comments are counter-productive. The last thing that I want to do is discourage you from improving our docs. I am not a member of @jekyll/documentation and I am certainly not a technical writer; I will step aside and let others handle this.

tomjohnson1492:
@pathawks I really do appreciate your feedback and review. You make some valid points. Cloning someone's website does raise immediate red flags. I can probably address this with a revision and maybe approach it from a "Making your first theme" angle, as you suggested. 

Letting a first draft sit for a while is actually good -- it will let me see the content with a fresh perspective, so let me take another crack at it.

parkr:
@DirtyF @pathawks if this looks good to you, feel free to merge.

DirtyF:
I kinda hope @andrewbanchich or @mmistakes would take a quick look at this tutorial before merging 🙏

andrewbanchich:
@DirtyF As in you'd like us to contribute to it?

DirtyF:
@andrewbanchich As someone used to adapt existing themes, what's your opinion on this?  Any feedback? Is it realistic? What's your workflow?

andrewbanchich:
@DirtyF Gotcha, I'll check this out this weekend and see what I can contribute!

tomjohnson1492:
I made some updates to the initial draft to add more detail and clarity. I welcome any reviews or other comments that would help make this tutorial stronger. It's really designed for people who are new to Jekyll and want to convert an HTML template into a Jekyll site. 

In this tutorial, I try to briefly expose readers to different aspects of theming in a basic way, without diving too deeply into details. There are about 13 different steps, so I'm showing more of an end-to-end process so users can get a sense of the whole (rather than just seeing one single, isolated task).

andrewbanchich:
@DirtyF @tomjohnson1492 Nice job! Here's my feedback:

- I haven't used `{% raw %}` in the themes I've converted. Maybe to make it simpler you could only use `{{ content }}` and then explain later the reason why someone may need to use `{% raw %}`?
- For nav menus, I think using `_data` might be overkill for most people. Here is some code I used for a menu:

`
<nav id="menu">
	<ul class="links">
        {% for page in site.pages %}
		    {% if page.layout == "home" %}
		        <li><a href="{{ "" | absolute_url }}">{{ page.title }}</a></li>
	    	{% endif %}
		{% endfor %}
		{% for page in site.pages %}
		    {% if page.layout != "home" %}
		        <li><a href="{{ page.url | absolute_url }}">{{ page.title }}</a></li>
		    {% endif %}
		{% endfor %}
	</ul>
	<ul class="actions vertical">
		<li><a href="#" class="button special fit">Get Started</a></li>
		<li><a href="#" class="button fit">Log In</a></li>
	</ul>
</nav>`

That way, the Home link will always be on top, and the others will automatically appear in the menu when other pages are added.

- One thing I learned to be careful with in Jekyll is paths. I think there were some updates made to Jekyll that smooth some things out, but I do think it would be important to discuss using filters like `relative_url` and `absolute_url`, as well as issues you may come across when setting a site url and base url in `_config.yml`while serving the site locally and then pushing live.

- On that note, maybe it would be nice to have a "The Strange and Wonderful World of Jekyll's Idiosyncrasies" section where you discuss common issues that a developer may run into using Jekyll and Liquid tags.

Nice work! I think this is a great idea to get people to start converting more sites.

DirtyF:
👍  for starting with building a simple navigation without data files here, we can link to the dedicated navigation tutorial anyway 😉 .

tomjohnson1492:
Thanks Andrew. I appreciate your review and suggestions. I'll incorporate these into the tutorial. Re the `{% raw %}` tags, these won't be visible. They're just used to tell Jekyll to ignore the actual code and to display the tag the `raw` tags enclose instead. Otherwise Jekyll tries to execute the statement (e.g., an include) and would complain if the reference file is missing.

I like the code sample for the simplified menu. This kind of highlights how different people use Jekyll. For blogs, you only need a few pages. For doc sites, you have hundreds of pages and often no posts.

tomjohnson1492:
I incorporated edits from Andrew's review. Mainly, I expanded the page navigation section with a more basic example (rather than just iterating through a data file). In the default layout section, I also included examples with relative_url and absolute_url tags. I cleaned up other sentences for clarity. 

Unless anyone has other requests for changes, I'm done with all my edits on this tutorial. I think it's ready to merge.

_tutorials/convert-html-site-to-jekyll.md
```diff
@@ -80,31 +80,56 @@ With this basic foundation of how a Jekyll site works, you can convert almost an
...
 
-Does the local `default.html` page look good in your browser? Are all images, styles, and other elements showing up correctly? If so, great. Keep going.
+Jekyll provides some [filter tags](/docs/templates/) to prepend a site URL before path. For example, you could preface your stylesheet like this:
 
-In the next steps, you'll use this template as the layout for all your pages. In the next section, you'll blank out the content of the layout and replace it with placeholder tags that get populated dynamically with your Jekyll pages.
+```liquid
+{{ "/assets/style.css" | relative_url }}
+```
+
+The `relative_url` tag will prepend the `baseurl` value from your config file to the input. This is useful if your site is hosted at a subpath rather than the root of the domain (for example, `http://mysite.com/blog/`).
```
ashmaroli: `relative_url` and `absolute_url` are `filters` not `tags`

_tutorials/convert-html-site-to-jekyll.md
```diff
@@ -0,0 +1,362 @@
...
+│   └── default.html
+└── index.md
+```
+
+Create these 3 files in a folder called `myjekyllsite`. (Put `default.html` inside a folder called `_layouts`.) Then populate the content of the files as follows:
+
+**_config.yml**
+
+```yaml
+markdown: kramdown
```
DirtyF: This is already set by default: https://github.com/jekyll/jekyll/blob/27ed81547b12d28a60c51961b82a5723981feb7d/lib/jekyll/configuration.rb#L38

We could set website `title` or other common setting here.

_tutorials/convert-html-site-to-jekyll.md
```diff
@@ -0,0 +1,362 @@
...
+* Jekyll pushes the content from all pages and posts into the `{% raw %}{{ content }}{% endraw %}` tags in the layout specified (`default`) in the front matter tags.
+* The processed files get written as `.html` files in the `_site` directory.
+
+With this basic foundation of how a Jekyll site works, you can convert almost any HTML theme into a Jekyll site. The following sections will walk you through a step-by-step tutorial on converting an HTML template into a Jekyll site.
+
+## 2. Establish the Default Layout
+
+Find your HTML theme and save it as a default layout. If you're converting or cloning (with permission to do so) an existing site, you can right-click the page and view the source code. Copy and paste the source code into a file called `default.html` inside a folder called `_layouts`. This will be the layout template for your pages and posts.
+
+Note that in looking for templates, you want the HTML output of the template. If the template has PHP tags or other dynamic scripts, these dynamic elements will need to be converted to HTML or stripped out, since Jekyll handles only static content.
```
DirtyF: Jekyll has its own templating system (Liquid) that allows to retrieve dynamic content.

_tutorials/convert-html-site-to-jekyll.md
```diff
@@ -0,0 +1,362 @@
...
+```
+---
+title: My Awesome Site
+layout: home
+---
+```
+
+When a layout specifies another layout, it means the content of this layout will be stuffed into the `{% raw %}{{ content }}{% endraw %}` tag of the layout specified in the front matter. In this case, the content from `index.md` will be pushed into the `{% raw %}{{ content }}{% endraw %}` tags in the `home` layout. Then the `home` layout will be pushed into the `{% raw %}{{ content }}{% endraw %}` tags of the `default` layout.
+
+(If you don't specify a layout, `default` is used by default. Specifying it in the front matter tags is optional.)
```
DirtyF: Good point, we can use `{: .info }` here to emphase.

_tutorials/convert-html-site-to-jekyll.md
```diff
@@ -0,0 +1,362 @@
...
+
+By the way, the we've only scratched the surface of what you can do with `for` loops in retrieving posts. For example, if you want to display posts from a specific category, you can do so by adding a `categories` property to your post's front matter, and then looking in those categories. Further, you can limit the number of results by adding a `limit` property. Here's an example:
```
DirtyF: By the way, ~~the~~ we've only

_tutorials/convert-html-site-to-jekyll.md
```diff
@@ -0,0 +1,362 @@
...
+If you have more sophisticated requirements around navigation, see the [detailed tutorial on navigation]({% link _tutorials/navigation.md %}). 
+
+
+## 10. RSS feed
+
+Your Jekyll site needs an RSS feed. Here's the [basic RSS feed syntax](http://www.w3schools.com/xml/xml_rss.asp). To create an RSS file in Jekyll, create a file called `feed.xml` and add the following:
+
+```xml
+---
+layout: null
```
DirtyF: `layout: nil`

_tutorials/convert-html-site-to-jekyll.md
```diff
@@ -0,0 +1,362 @@
...
+
+This code uses a `for` loop to look through your last 20 posts. The content from the posts gets escaped and truncated to the last 400 characters using [Liquid filters](https://help.shopify.com/themes/liquid/filters).
+
+In your `default.html` layout, look for a reference to the RSS or Atom feed in your header, and replace it with a reference to the file you just created.
+
+```html
+<link rel="alternate" type="application/rss+xml"  href="/feed.xml" title="{% raw %}{{ site.title }}{% endraw %}">
+```
+
+Note that if you're publishing your site on Github Pages, you can add a Github Pages gem called [`jekyll-feed`][jekyll-feed] to auto-generate your feed.
```
DirtyF: `jekyll-feed` is a gem plugin that works without the need to use GitHub Pages

_tutorials/convert-html-site-to-jekyll.md
```diff
@@ -0,0 +1,362 @@
...
+        <lastmod>{{site.time | date: '%Y-%B-%d' }}</lastmod>
+        <changefreq>daily</changefreq>
+        <priority>0.5</priority>
+    </url>
+    {% endfor %}
+   
+</urlset>{% endraw %}
+```
+
+Note that if you're publishing your site on Github Pages, you can add a Github Pages gem called [`jekyll-sitemap`][jekyll-sitemap] to auto-generate your sitemap.
```
DirtyF: same here `jekyll-sitemap` does not require for you to use GitHub Pages. I think what you mean it's that this gem is whitelisted by GitHub Pages.

_tutorials/convert-html-site-to-jekyll.md
```diff
@@ -0,0 +1,481 @@
...
+---
+title: Blog
+layout: home
+---
+```
+
+When a layout specifies another layout, it means the content of the first layout will be stuffed into the `{% raw %}{{ content }}{% endraw %}` tag of the second layout. As an analogy, think of Russian dolls that fit into each other. Each layout fits into another layout that it specifies.
+
+```
+home --> default
```
ashmaroli: can you point me to the `source_code` for processing above `home --> default`

DirtyF: See above: https://github.com/jekyll/jekyll/pull/5881/files/7e9f4613b9bd913874d63ffbe6f66725e67230c3#diff-faca58a5785fa35d099934be0a3281e5R233

ashmaroli: Thanks @DirtyF  I now see that Tom was only illustrating the concept with an example of Russian Dolls fitting to each other (`home --> default`) and not setting a `layout: home --> default` as I saw it earlier... need to get some rest.....

_tutorials/convert-html-site-to-jekyll.md
```diff
@@ -0,0 +1,481 @@
...
+
+Copy and paste the source code into a file called `default.html`. Put the `default.html` file inside a folder called `_layouts`. This will be the default layout template for your pages and posts &mdash; that is, each page or post will use this layout when Jekyll builds the site.
+
+Note that in looking for templates, you want the HTML output of the template. If the template has PHP tags or other dynamic scripts, these dynamic elements will need to be converted to HTML or stripped out, since Jekyll handles only static content.
+
+Open `default.html` into your browser locally to ensure the site looks and functions like it does online. You will likely need to adjust CSS, JS, and image paths so they work.
+
+For example, if the paths were relative on the site you copied, you'll need to either download the same assets into your Jekyll site or use absolute paths to the same assets in the cloud. Syntax such as `src="//` requires a prefix such as `src="http://` to work in your local browser.
+
+Jekyll provides some [filter tags](/docs/templates/) to prepend a site URL before path. For example, you could preface your stylesheet like this:
```
ashmaroli: Can we not mix two distinct entities here? `filters` and `tags` are two distinct functional entities in `Liquid`. 
```diff
- [filter tags](/docs/templates/)
+ [URL filters](/docs/templates/#filters)
```

_tutorials/convert-html-site-to-jekyll.md
tomjohnson1492: good catch. updated.

_tutorials/convert-html-site-to-jekyll.md
tomjohnson1492: In this tutorial, I'm not having users run `jekyll new site` because that ends up installing a default gem theme with its own styles and such. I have them just create the config file manually, so nothing would be populated by default in it.

_tutorials/convert-html-site-to-jekyll.md
tomjohnson1492: updated

_tutorials/convert-html-site-to-jekyll.md
tomjohnson1492: updated

_tutorials/convert-html-site-to-jekyll.md
tomjohnson1492: updated

_tutorials/convert-html-site-to-jekyll.md
tomjohnson1492: updated

_tutorials/convert-html-site-to-jekyll.md
ashmaroli: > I have them just create the config file manually, so nothing would be populated by default in it

The `DEFAULTS` Frank pointed to **will be set automatically** for an empty `_config.yml`

_tutorials/convert-html-site-to-jekyll.md
tomjohnson1492: updated

_tutorials/convert-html-site-to-jekyll.md
tomjohnson1492: updated

_tutorials/convert-html-site-to-jekyll.md
DirtyF: I get you, but It's not related to the `new`command, Jekyll already sets some configuration defaults for you and use them when your run the `build` command. In theory you don't even need a `_config.yml` to build a jekyll site. Except when you want to override a default value, like `source`or `permalink` you don't need to write those keys in the config file. Convention over Configuration paradigm applies here.

_tutorials/convert-html-site-to-jekyll.md
tomjohnson1492: updated. I'll fix this confusion with a diagram instead. My arrow was a lame attempt at an illustration.

_tutorials/convert-html-site-to-jekyll.md
tomjohnson1492: updated

tomjohnson1492:
I made the updates noted in previous reviews. 

Note: I tried changing the file name from `convert-html-site-to-jekyll.md` to `create-jekyll-theme.md`, but I think this may have been a bad idea because Github wants to retain all the comments on the previous file and its name? I don't think my commit actually deleted/renamed the old file. It seemed to have just added another one. If I need to explicitly remove the `convert-html-site-to-jekyll.md` file, let me know. The `create-jekyll-theme.md` should replace it.

I added 2 other files in this commit -- an update to the list of items on the tutorials nav and an image.

_tutorials/convert-html-site-to-jekyll.md
tomjohnson1492: Actually, when I use `nil` instead of `null`, Jekyll complains in the command line about not being able to find a layout with the name of `nil`. But it doesn't complain when it looks for `null`.

ashmaroli: `layout: null` is correct. A quick [reference](https://github.com/benbalter/jekyll-default-layout#disabling) I could think of

_tutorials/convert-html-site-to-jekyll.md
tomjohnson1492: ahh, I see now. Thanks for clarifying. I made an update to the commit that makes it clear that kramdown is a default setting.

ashmaroli:
I love the concept behind your illustration to explain nested layouts, but from a total-noob's point-of-view, I can't make sense out of it.. specifically, which layout is the `default.html`

tomjohnson1492:
Each layout does indicate the name of the layout directly on it. I'm not
sure how to make it any clearer. If you have suggestions, let me know.

ashmaroli:
> Each layout does indicate the name of the layout directly on it. I'm not sure how to make it any clearer. If you have suggestions, let me know.

```
---yaml
layout: default.html
---
```
indicates that this layout is a descendant of `default.html` i.e. its contents will be inserted into `default.html`
instead you could have the name either **displayed as a caption underneath the boxes**, or as **a comment in the FrontMatter**:
```
---yaml
# _layouts/home.html

# contents of this layout will be injected into 'default.html'
layout: default
---
```

ashmaroli:
> Each layout does indicate the name of the layout directly on it. I'm not sure how to make it any clearer. If you have suggestions, let me know.

```yaml
---
layout: default.html
---
```
indicates that this layout is a descendant of `default.html` i.e. its contents will be inserted into `default.html`
instead you could have the name either **displayed as a caption underneath the boxes**, or as **a comment in the FrontMatter**:
```yaml
---
# _layouts/home.html

# contents of this layout will be injected into 'default.html'
layout: default
---
```

tomjohnson1492:
thanks for the ideas. i'll see if i can work this into the graphic.

tomjohnson1492:
I updated the graphic about how layouts work to make it a little clearer. Hopefully this is better?

DirtyF:
Nice work @tomjohnson1492, let's add this tutorial and see of we have more feedback from the community 🌈 

@jekyllbot: merge +docs

PR #5882, opened 2017-02-11
Add note to collections doc on hard-coded collections.

felicianotech:
As noted in a few Issues including #5064, `posts` is a hard-coded collection that will come up when iterating over `site.collections`. This was a source of confusion for me. I didn't see any mention of this in the docs.

This PR adds a note to http://jekyllrb.com/docs/collections/ that `posts` is a hard-coded collection.


DirtyF:
@jekyllbot: merge +docs

PR #5886, opened 2017-02-16
test 'absolute_url' filter with 'xml_escape' filter

ashmaroli:
test  jekyll/jekyll-feed#158 upstream

### DIFF
--- a/test/test_filters.rb
+++ b/test/test_filters.rb
@@ -391,6 +391,19 @@ class TestFilters < JekyllUnitTest
         assert_equal "http://example.com/", filter.absolute_url(page_url)
       end
 
+      should "not append a forward slash if input is '/' with nil 'baseurl'" \
+             " and finally passed to xml_escape filter" do
+        page_url = "/"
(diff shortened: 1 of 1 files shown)


ashmaroli:
looks like the filter combo only fails within the context of `.xml` files
/cc @pathawks

pathawks:
I too have been trying to create a failing test, and I too have been unsuccessful.

ashmaroli:
its not *meant* to fail.. just checks if it'll fail with the filter-input combo

pathawks:
We need to create a failing test so we can isolate the problem.

pathawks:
There's a trailing slash _after_ the Liquid. I am banging my head against a wall right now.

https://github.com/jekyll/jekyll-feed/blob/v0.8.0/lib/feed.xml#L8

ashmaroli:
The error is in this [line](https://github.com/jekyll/jekyll-feed/blob/master/lib/jekyll-feed/feed.xml#L10)
See the trailing `/` just before the closing tag?

PR #5069, opened 2016-07-08
check if relative URL contains a colon

Crunch09:
fixes #2834 

screen shot 2016-07-08 at 20 35 47

Issue #2834, opened 2014-08-27
URL should validate output


parkr:
Colons, for example, should not be allowed

https://github.com/jekyll/jekyll-paginate/issues/8

kostiakoval:
I still have this issue what should I do ?

rensa:
I get the same error as @tkrotoff (identical except for the path of the jekyll blog) when I'm serving and any of my files regenerate. I can't for the life of me work out which file has an illegal character, though.

parkr:
Can you run with `LISTEN_GEM_DEBUGGING=1`? That env var will show us more about what Listen thinks is happening.

rensa:
Sure. Here's what I'm getting when I `serve`:
```
Configuration file: /Users/rensa/Code/rensa.github.io/_config.yml
            Source: /Users/rensa/Code/rensa.github.io
       Destination: /Users/rensa/Code/rensa.github.io/_site
      Generating... 
                    done.
I, [2015-10-26T20:32:22.635804 #12929]  INFO -- : Celluloid loglevel set to: 1
I, [2015-10-26T20:32:22.635855 #12929]  INFO -- : Listen version: 2.8.4
I, [2015-10-26T20:32:22.681548 #12929]  INFO -- : Record.build(): 0.02466297149658203 seconds
 Auto-regeneration: enabled for '/Users/rensa/Code/rensa.github.io'
Configuration file: /Users/rensa/Code/rensa.github.io/_config.yml
    Server address: http://127.0.0.1:4000
  Server running... press ctrl-c to stop.
```
And then when I save an alteration to a file (in this case, I added the word `Change!` to one my posts):
```
I, [2015-10-26T20:32:52.988417 #12929]  INFO -- : listen: raw changes: [[:modified, "/Users/rensa/Code/rensa.github.io/_posts/writing/2015-09-01-shell-tut.md"]]
I, [2015-10-26T20:32:52.988552 #12929]  INFO -- : listen: final changes: {:modified=>["/Users/rensa/Code/rensa.github.io/_posts/writing/2015-09-01-shell-tut.md"], :added=>[], :removed=>[]}
      Regenerating: 1 files at 2015-10-26 20:32:52 ...done.
[...]

PR #5889, opened 2017-02-17
Don't raise an error if URL contains a colon

### LINKED ISSUE #5284: Can no longer use colons in permalinks
- [X] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [X] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/

---
- [X] I am on (or have tested on) **_macOS**_ 10+
- [X] I am on (or have tested on) **_Debian/Ubuntu**_ GNU/Linux (_via GitHub Pages & TravisCI_)
- [ ] I am on (or have tested on) **_Fedora**_ GNU/Linux (_tested on CentOS 6_)
- [ ] I am on (or have tested on) **_Arch**_ GNU/Linux
- [ ] I am on (or have tested on) **_Other**_ GNU/Linux
- [ ] I am on (or have tested on) **_Windows**_ 10+

---
- [ ] I had an error on GitHub Pages, and I have not tested locally.
- [X] I had an error on GitHub Pages and TESTED it locally.
- [X] I was trying to build.
- [ ] It was another bug.
## My Reproduction Steps

TravisCI log https://travis-ci.org/biopython/biopython.github.io/builds/154676985 from same commit https://github.com/biopython/biopython.github.io/commit/665c0c94aaf43d75bdc775f999156bdf00d3a89c where GitHub Pages broke (unrelated edit in this commit, change is the dependencies):

```
$ bundle exec jekyll build
Configuration file: /home/travis/build/biopython/biopython.github.io/_config.yml
            Source: /home/travis/build/biopython/biopython.github.io
       Destination: /home/travis/build/biopython/biopython.github.io/_site
      Generating... 
  Liquid Exception: The URL /wiki/Category:Cookbook is invalid because it contains a colon. in /_layouts/default.html
                    The URL /wiki/Category:Cookbook is invalid because it contains a colon.
The command "bundle exec jekyll build" exited with 1.
```

We use colons in permalinks to preserve legacy MediaWiki URLs, e.g.:

```
$ head -n 8 wiki/Category_Cookbook.md 

---
title: Cookbook Entries
permalink: wiki/Category:Cookbook
layout: tagpage
tag: Cookbook

---

The entries contained in the Cookbook category are designed to
```

We link to them using e.g. from `wiki/Documentation.md` using:

```
-   [Wiki documentation](Category%3AWiki_Documentation "wikilink")
```

Testing locally, I have simplified `_layout/default.html` to remove the header and sidebar,

``` html
<!DOCTYPE html>
<html lang="en-us">

  <body>

    <div class="content container">
      <hr class="head"/>
      {{ content }}
    </div>

  </body>
</html>
```

Now:

```
$ jekyll build
WARN: Unresolved specs during Gem::Specification.reset:
      rouge (~> 1.7)
      jekyll-watch (~> 1.1)
WARN: Clearing out unresolved specs.
Please report a bug if this causes problems.
Configuration file: /Volumes/Hitachi-3TB-mirrored/repositories/biopython.github.io/_config.yml
            Source: /Volumes/Hitachi-3TB-mirrored/repositories/biopython.github.io
       Destination: /Volumes/Hitachi-3TB-mirrored/repositories/biopython.github.io/_site
      Generating... 
                    The URL /wiki/Category:Cookbook is invalid because it contains a colon.
```
## The Output I Wanted

Want to continue to use colons in permalinks, and be able to link to those pages.

It appears _check if relative URL contains a colon_ (#5069) breaks our use case.

DirtyF:
fix #5284

### DIFF
--- a/lib/jekyll/errors.rb
+++ b/lib/jekyll/errors.rb
@@ -12,6 +12,5 @@ module Jekyll
     InvalidDateError      = Class.new(FatalException)
     InvalidPostNameError  = Class.new(FatalException)
     PostURLError          = Class.new(FatalException)
-    InvalidURLError       = Class.new(FatalException)
   end
 end
--- a/lib/jekyll/errors.rb
+++ b/lib/jekyll/errors.rb
@@ -12,5 +12,6 @@ module Jekyll
     InvalidDateError      = Class.new(FatalException)
     InvalidPostNameError  = Class.new(FatalException)
     PostURLError          = Class.new(FatalException)
+    InvalidURLError       = Class.new(FatalException)
   end
 end
--- a/test/test_url.rb
+++ b/test/test_url.rb
@@ -61,16 +61,6 @@ class TestURL < JekyllUnitTest
       ).to_s
     end
 
-    should "throw an exception if the URL contains a colon" do
-      url = URL.new(
-        :template     => "/:x/:y/:z",
-        :placeholders => { :x => "foo", :z => "bar" }
-      )
-      assert_raises Jekyll::Errors::InvalidURLError do
-        url.to_s
-      end
-    end
-
     should "check for key without trailing underscore" do
       _, matching_doc = fixture_document("_methods/configuration.md")
       assert_equal "/methods/configuration-configuration_methods_configuration", URL.new(
--- a/lib/jekyll/url.rb
+++ b/lib/jekyll/url.rb
@@ -35,15 +35,8 @@ module Jekyll
     # The generated relative URL of the resource
     #
     # Returns the String URL
-    # Raises a Jekyll::Errors::InvalidURLError if the relative URL contains a colon
     def to_s
-      sanitized_url = sanitize_url(generated_permalink || generated_url)
-      if sanitized_url.include?(":")
-        raise Jekyll::Errors::InvalidURLError,
-          "The URL #{sanitized_url} is invalid because it contains a colon."
-      else
-        sanitized_url
-      end
+      sanitize_url(generated_permalink || generated_url)
     end
 
     # Generates a URL from the permalink


lib/jekyll/errors.rb
```diff
@@ -12,6 +12,5 @@ module Errors
     InvalidDateError      = Class.new(FatalException)
     InvalidPostNameError  = Class.new(FatalException)
     PostURLError          = Class.new(FatalException)
-    InvalidURLError       = Class.new(FatalException)
```
pathawks: Let's not get rid of this error; we might want to use it elsewhere.

DirtyF: OK, reverted the removal.

pathawks:
I think it's worth taking a look at where this code came from.

In jekyll/jekyll-paginate#8 Jekyll was choking when trying to create a URL from a template that included invalid keys. `:path/page:num` was not working because `:path` is not a valid permalink key.

In https://github.com/jekyll/jekyll/issues/2834#issuecomment-64732405 Jekyll dies while trying to write a file that includes a colon in the filename.

#5069 was added to prevent these issues.

[workflow_2x](https://xkcd.com/1172/)

If we get rid of this check, we should find some other way to avoid dying catastrophically if a user tries to specify an invalid permalink.

Perhaps the best solution is that, in edge cases where a colon in the permalink is actually desired, the colon be escaped in the permalink https://github.com/jekyll/jekyll/issues/5284#issuecomment-242312942

DirtyF:
@pathawks thanks fro summing things up, I just followed what @parkr said in https://github.com/jekyll/jekyll/issues/5284#issuecomment-280503627

test/test_url.rb
```diff
@@ -61,16 +61,6 @@ class TestURL < JekyllUnitTest
       ).to_s
     end
 
-    should "throw an exception if the URL contains a colon" do
-      url = URL.new(
-        :template     => "/:x/:y/:z",
-        :placeholders => { :x => "foo", :z => "bar" }
-      )
-      assert_raises Jekyll::Errors::InvalidURLError do
```
parkr: What do you and @pathawks think about, if a placeholder doesn't exist, simply removing it?

The real issue we're driving at is: you used an invalid URL placeholder, i.e. `:dat` instead of `:date`.

pathawks:
I think there are two issues in conflict here:

 1. A user is trying to use an invalid URL placeholder, i.e. `:dat` instead of `:date`
 2. A user is trying to use a permalink that includes a colon, i.e. `wiki/Category:Cookbook`

In scenario 1, the user wants to know that something has gone wrong so that the problem can be fixed.
In scenario 2, the user wants the page to be written in such a way that the permalink does include the colon.

Failing the build serves scenario 1, but breaks scenario 2.
Silently dropping invalid URL placeholders breaks scenario 2, and may cause unexpected results in scenario 1.
Simply building URLs that contain colons serves scenario 2, but will certainly cause unexpected results in scenario 1.

I believe that scenario 1 is far more common than scenario 2.

Apparently, with the way things are currently, scenario 2 can still be achieved by [escaping colons in URLs](https://github.com/jekyll/jekyll/issues/5284#issuecomment-242312942)

If we are going to go out of our way to allow URLs to contain colons, we also need to figure out how to write those files to disk in a way that works across the different operating systems we support.

parkr:
@jekyllbot: merge +minor

PR #5891, opened 2017-02-18
Remove duplicate footnote_nr from default config

chunfeilung:
The configuration page lists the footnote_nr kramdown configuration option twice. That seemed a bit much, so I removed one of the two lines.


DirtyF:
@jekyllbot: merge +docs

Issue #5895, opened 2017-02-20
jekyll serve changes url to 127.0.0.1:4000


Ungaminga:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [ ] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [X] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [ ] This is a feature request.

---

- [ ] I am on (or have tested on) ***macOS*** 10+
- [ ] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux
- [ ] I am on (or have tested on) ***Fedora*** GNU/Linux
- [X] I am on (or have tested on) ***Arch*** GNU/Linux
- [ ] I am on (or have tested on) ***Other*** GNU/Linux
- [ ] I am on (or have tested on) ***Windows*** 10+

---

- [ ] I was trying to install.
- [ ] There is a broken Plugin API.
- [ ] I had an error on GitHub Pages, and I have reproduced it locally.
- [ ] I had an error on GitHub Pages, and GitHub Support said it was a Jekyll Bug.
- [ ] I had an error on GitHub Pages and I did not test it locally.
- [ ] I was trying to build.
- [X] It was another bug.

## My Reproduction Steps

run $ jekyll serve
go to your index.html
set some {{site.url}}

## The Output I Wanted

Got 127.0.0.1:4000 instead of _config.yml site.url.

I run jekyll serve and my _config.yml site.url changed to host:port. It's pity, but site.url used in https://github.com/jekyll/jekyll-seo-tag, so i need to build first i deploy. I'm new in ruby and have no idea where jekyll serve changes url. It's akward to have a few new scripts for correct deploy.

crispgm:
Jekyll will not serve as `site.url unless you set `JEKYLL_ENV=production`.

See <https://jekyllrb.com/docs/configuration/#specifying-a-jekyll-environment-at-build-time>

pathawks:
Thanks @crispgm :+1:

Ungaminga:
There are no issue that jekyll not serves as. When i use jekyll serve my site.url _automaticly_ modifided by it.

PR #5896, opened 2017-02-20
Removed JEKYLL_ENV=development site.url change.

Ungaminga:
Here some pull request for https://github.com/jekyll/jekyll/issues/5895. That feature/bug was so annoying that cost me a night.

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -33,11 +33,7 @@ module Jekyll
               opts["serving"] = true
               opts["watch"  ] = true unless opts.key?("watch")
               config = opts["config"]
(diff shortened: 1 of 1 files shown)


pathawks:
This is a documented feature. We do not want to remove it.

Perhaps there is a way we could explain it more clearly?

Ungaminga:
https://jekyllrb.com/docs/variables/ try to search"url" here and find some documantation about JEKYLL_ENV=development. And why anyone need to have some .sh file or put something in bash.bashrc to have correct software work? With default configuration you can only reserve _incorrect_ software result. Is it KISS or Unix-way? It's just retarded, admit it.

pnn:
@Ungaminga the way you implemented it removes all functionality from the `serve` command? also, please refrain from using slurs in your comments.

PR #5900, opened 2017-02-22
Fixed tutorial for publishing gem to include repo.

iBobik:
gem build lists files by git, so it is needed to have them commited.


DirtyF:
Right, thanks for the PR.

/cc @jekyll/documentation

pnn:
@jekyllbot: merge +docs

PR #5903, opened 2017-02-25
Remove dependency on include from default about.md

ashmaroli:
This ensures better portability, allowing to switch and preview themes without having to delete the code block.
(Jekyll aborts build if the include is not found when switching themes)

--
/cc @jekyll/ecosystem


parkr:
@jekyllbot: merge +bug

PR #5904, opened 2017-02-26
Sort collection items by an attribute.

### LINKED ISSUE #5754: Specify order of collection items and set previous/next links accordingly

---

I want to be able to specify the order of items in a collection and have previous/next links respect that.

**Use Case**

I am using Jekyll to build a lesson that contains several topics. Each topic is a separate Markdown file in the `_topics` collection. Topics must be listed on the home page in a particular order, and we want topics to have previous/next links set correctly.

**What Doesn't Work**

We currently enforce ordering using a naming convention: topics are called `01-intro.md`, `02-setup.md`, `03-filter.md`, etc. This works, but when we want to re-order topics, we need to rename many files whose content often hasn't changed, which makes the version control log difficult to interpret.

We have tried using a list in `_config.yml` to specify order:

~~~
topic_order:
  - '/intro'
  - '/setup'
  - '/filter'
~~~

and then using that loop and `where` to create the ordered list in the home page:

~~~
<ul>
{% for ident in site.topic_order %}
  {% assign topic = site.topics | where: "id", ident | first %}
  <a href="{{topic.url}}">{{topic.title}}</a>
  <li>
{% endfor %}
</ul>
~~~

However, his doesn't help with previous/next links within the topics. The only solution I've been able to come up with for that is to use a search loop inside each topic to find it within the `site.topics` list, but that means that building the site is O(N**2), since the search loop over N topics is run once for each topic, and that makes builds noticeably slower for large lessons with 20-30 topics.

**What Might Work**

1.  Support an `order` key in the `_config.yml` metadata for each collection whose value is the ordered list of IDs of collection members. Values in the list that don't correspond to collection members are errors; collection members that are not listed are either an error or omitted.

2.  Set `previous` and `next` links in the `page` for each collection item to respect this ordering.

ashmaroli:
sort collection items by specifying a "sort_by" metadata key in the `_config.yml` and assigning the `key` a `value` that corresponds to a key set in the document's `Front Matter`.

Notes:
  - **all items** in the collection must have the `key` defined.
  - defaults to original `sort!` if none of the items have the key defined.

Example
```
# _config.yml

collections:
  tutorials:
    output: true
    sort_by: index


# _tutorials/get-started.md
---
title: Get Started
index: 1
---

# _tutorials/be-confident.md
---
title: Be Confident
index: 2
---
```
```
# default_docs_array => 
[#<Jekyll::Document _tutorials/be-confident.md collection=tutorials>, #<Jekyll::Document _tutorials/getting-started.md collection=tutorials>]

# docs_array_with_custom_sort => 
[#<Jekyll::Document _tutorials/getting-started.md collection=tutorials>, #<Jekyll::Document _tutorials/be-confident.md collection=tutorials>]
```

Tests will be added later.

/cc @jekyll/ecosystem 

@gvwilson, this implementation may not qualify for your bounty, but in theory should sort the documents and assign the `prev / next` pagers accordingly.

Fixes #5754


gvwilson:
@ashmaroli thanks for the quick turnaround, but this still requires people to modify multiple files when they want to reorder a couple of files or insert a new one.  For example, if the collection includes `alpha.md`, `beta.md`, and `gamma.md` with ordering keys 1, 2, and 3, and I need to change the order to `['beta', 'gamma', 'alpha']`, I have to edit three files (correctly) instead of just `_config.yml`.

ashmaroli:
> but this still requires people to modify multiple files when they want to reorder a couple of files or insert a new one

True. Hence the reason I said that this implementation might not qualify for the bounty.
I chose this method because the array in the `_config.yml` is an **Array of `String`s** while the Collection's array is an **Array of `Jekyll::Document`s**

ashmaroli:
@pathawks I added a basic test step. Will add more if you think this implementation could be beneficial addition to the Core..

pathawks:
I don't think that sorting by an `index` in a documents front matter is very valuable. What happens if I want to add a new page between `1` and `2`? Do I have to edit the index of every page?

I'm not sure what problem this is solving.

ashmaroli:
@pathawks I've added an additional implementation to have the documents array modified by an `order` metadata as @gvwilson specified. I think the latest implementation is a bit inferior in performance in comparison to the earlier. So included both implementations in the PR.

`sort_by` can be used with *stable* collections where the items can be programmatically sorted by specifying a key defined in *all* the items.

`order` can be used for greater control with the arrangements of the array:
  - the `order` metadata should be an `Array` of just the filenames of the documents.
  - the items can be arranged in any desired order and the order will be virtually mapped to the document's `sort_index` Front Matter key, which will be used consequently to sort the array. Hence configuring via this metadata is slower than the former.
  - `order` metadata takes precedence over `sort_by` and will gracefully warn the user if a provided filename doesn't exist in the collection.

will add more tests, if required, after further feedback.

features/collections.feature
```diff
@@ -162,6 +162,79 @@ Feature: Collections
...
+          path: ""
+          type: tutorials
+        values:
+          layout: tuts
+
+    collections:
+      tutorials:
+        output: true
+        order:
+          - getting-started.md
```
gvwilson: These appear to be in the same order as defined by the internal header's `lesson` value, so there's no way to tell whether the sorting is due to the config information or the internal data. Can this test be modified to put the files in a different order so that we can be sure they're being sorted by config spec?

lib/jekyll/collection.rb
```diff
@@ -210,6 +210,29 @@ def read_document(full_path)
...
+        doc_path = "_#{@label}/#{o}"
+        doc = docs.find { |d| d.relative_path == doc_path }
+        if File.exist?(@site.in_source_dir(doc_path))
+          doc.data["sort_index"] = @metadata["order"].index(o)
+        else
+          Jekyll.logger.warn "Error:", "#{o} not found in '#{@label}' collection!"
+        end
+      end
+      docs.sort_by! { |d| d.data["sort_index"] }
+    end
```
gvwilson: 1.  What happens to files that are in the collection directory but *not* specified in the explicit order list in config?
2.  What happens to files that are in the collection directory but do not specify the key being used for sorting (in this case, don't have a `lesson` key)?

features/collections.feature
ashmaroli: the collection would be sorted by `lesson` key only if `sort_by` key is set to "lesson" here. I'll nevertheless edit the order in the test to bring in clarity.

lib/jekyll/collection.rb
ashmaroli: > What happens to files that are in the collection directory.. ..don't have a `lesson` key?

The build process aborts with an `Error`.
I'll push a couple to patches to better handle such situations.

gvwilson:
Failing if there are files in the collection that are not expected 
(i.e., not listed in the ordering in _config.yml) is a useful behavior 
in some cases, but in others, people may want to append them to the 
collection in natural (alphabetic) order - the analogy is with "extra" 
arguments to a function, which can either be an error or can be really 
useful.  Is it worth allowing another parameter in the config for the 
collection to specify which behavior is desired?  That will make it easy 
to expand behaviors in future.

pathawks:
> > What happens to files that are in the collection directory.. ..don't have a lesson key?
>
> The build process aborts with an Error.

This seems like it makes things complicated.

> people may want to append them to the collection in natural (alphabetic) order

This seems like it makes things complicated.

gvwilson:
How about "files not listed in the config are ignored" (and 
equivalently, "files lacking the specified sorting key are ignored")?

ashmaroli:
This is how things will roll now:
  - if there's a file that doesn't have the FrontMatter key specified for `sort_by`, the custom sort rolls back to the default sorting with a graceful warning (for each deviant file).
  - if there's a file missing in the `order` list, the file will be included after the custom list defined in `order`. If there's multiple files missing in the list, then the deviant files will appear after the custom list, in alphabetical order.

gvwilson:
:+1: from me - thanks for the amendment.

ashmaroli:
You're welcome :smiley:

gvwilson:
Can someone please let me know the status of this one? Having it in Jekyll would make it a lot easier to create and manage GitBook-style sites (particularly multi-part lessons).

gvwilson:
@pathawks any further changes required?

pathawks:
@gvwilson I will let @parkr review :+1:

gvwilson:
@parkr any chance you could take a look at this one? thx

gvwilson:
Looks like jekyllbot just closed #5754, but the PR is still open - any chance of merge into the next release? thx.

features/collections.feature
```diff
@@ -162,6 +162,152 @@ Feature: Collections
     And I should see "Collections: this is a test!, Collection#entries, Jekyll.configuration, Jekyll.escape, Jekyll.sanitized_path, Site#generate, Initialize, Site#generate, YAML with Dots" in "_site/index.html" unless Windows
     And I should see "Collections: this is a test!, Collection#entries, Jekyll.configuration, Jekyll.escape, Jekyll.sanitized_path, Site#generate, Initialize, YAML with Dots" in "_site/index.html" if on Windows
 
+  Scenario: Sort entries by Front Matter key
+    Given I have an "index.html" page that contains "Collections: {% assign tutorials = site.tutorials %}{{ tutorials | map:"title" | join: ", " }}"
```
parkr: Why not just `{{ site.tutorials | map: "title" | join: ", " }}`?

features/collections.feature
```diff
@@ -162,6 +162,152 @@ Feature: Collections
...
+    And I have fixture collections
+    And I have a _layouts directory
+    And I have a "_layouts/tuts.html" file with content:
```
parkr: What is `tuts`?

features/collections.feature
```diff
@@ -162,6 +162,152 @@ Feature: Collections
...
+    Then I should get a zero exit status
+    Then the _site directory should exist
+    And I should see "Collections: Getting Started, Let's Roll!, Dive-In and Publish Already!, Tip of the Iceberg, Extending with Plugins, Graduation Day" in "_site/index.html"
+    And I should not see "Previous: Graduation Day" in "_site/tutorials/lets-roll.html"
+    And I should not see "Next: Tip of the Iceberg" in "_site/tutorials/lets-roll.html"
+    But I should see "Previous: Getting Started" in "_site/tutorials/lets-roll.html"
+    And I should see "Next: Dive-In and Publish Already!" in "_site/tutorials/lets-roll.html"
+
+  Scenario: Sort some entries by Front Matter key
+    Given I have an "index.html" page that contains "Collections: {% assign tutorials = site.tutorials %}{{ tutorials | map:"title" | join: ", " }}"
```
parkr: (Identical comment here as L166)

features/collections.feature
```diff
@@ -162,6 +162,152 @@ Feature: Collections
...
+    When I run jekyll build
+    Then I should get a zero exit status
+    Then the _site directory should exist
+    And I should see "Collections: Getting Started, Let's Roll!, Dive-In and Publish Already!, Tip of the Iceberg, Extending with Plugins, Graduation Day" in "_site/index.html"
+    And I should not see "Previous: Graduation Day" in "_site/tutorials/lets-roll.html"
+    And I should not see "Next: Tip of the Iceberg" in "_site/tutorials/lets-roll.html"
+    But I should see "Previous: Getting Started" in "_site/tutorials/lets-roll.html"
+    And I should see "Next: Dive-In and Publish Already!" in "_site/tutorials/lets-roll.html"
+
+  Scenario: Sort some entries by Front Matter key
```
parkr: This doesn't really describe why this test is different from the one above. Can you clarify?

lib/jekyll/collection.rb
```diff
@@ -210,6 +210,65 @@ def read_document(full_path)
 
     private
 
+    def sort_docs!
+      if @metadata["order"] && @metadata["order"].is_a?(Array)
+        rearrange_docs!
+      elsif @metadata["sort_by"]
```
parkr: Use `metadata` attr_reader here instead of instance variable.

lib/jekyll/collection.rb
```diff
@@ -210,6 +210,65 @@ def read_document(full_path)
 
     private
 
+    def sort_docs!
+      if @metadata["order"] && @metadata["order"].is_a?(Array)
```
parkr: Use `metadata` attr_reader here instead of instance variable.

lib/jekyll/collection.rb
```diff
@@ -210,6 +210,65 @@ def read_document(full_path)
...
+        rearrange_docs!
+      elsif @metadata["sort_by"]
+        sort_docs_by_key!
+      else
+        docs.sort!
+      end
+    end
+
+    def sort_docs_by_key!
+      meta_key = @metadata["sort_by"]
```
parkr: Use `metadata` attr_reader here instead of instance variable.

You should handle the `nil` case as well (when `meta_key` is nil)

lib/jekyll/collection.rb
```diff
@@ -210,6 +210,65 @@ def read_document(full_path)
...
+      docs.sort_by! { |d| d.data[meta_key] }
+    rescue
```
parkr: This `rescue` must be specific: which error(s) do you intend to catch?

lib/jekyll/collection.rb
```diff
@@ -210,6 +210,65 @@ def read_document(full_path)
...
+      docs.select { |entry| entry.data[meta_key].nil? }.each do |e|
+        Jekyll.logger.warn "Error:", "'#{meta_key}' not defined in '#{e.relative_path}'"
+      end
+
+      Jekyll.logger.warn "Error:",
+        "Custom sorting skipped due to inconsistent key definition."
+    end
+
+    def rearrange_docs!
+      @metadata["order"].each do |entry|
```
parkr: Use `metadata` attr_reader here instead of instance variable.

Please also wrap in `Array(metadata["order"])` so that a non-array value doesn't explode this.

Please also handle the `nil` case.

lib/jekyll/collection.rb
```diff
@@ -210,6 +210,65 @@ def read_document(full_path)
...
+        doc_path = "_#{@label}/#{entry}"
+        doc = docs.find { |d| d.relative_path == doc_path }
+
+        if File.exist?(@site.in_source_dir(doc_path))
+          doc.data["sort_index"] = @metadata["order"].index(entry)
+        else
+          Jekyll.logger.warn "Error:", "#{entry} not found in '#{@label}' collection!"
```
parkr: Use `label` attr_reader here please.

lib/jekyll/collection.rb
```diff
@@ -210,6 +210,65 @@ def read_document(full_path)
...
+        "Custom sorting skipped due to inconsistent key definition."
+    end
+
+    def rearrange_docs!
+      @metadata["order"].each do |entry|
+        doc_path = "_#{@label}/#{entry}"
+        doc = docs.find { |d| d.relative_path == doc_path }
+
+        if File.exist?(@site.in_source_dir(doc_path))
+          doc.data["sort_index"] = @metadata["order"].index(entry)
```
parkr: I don't want us to be munging the data for the document. Couldn't this method be simplified to just:

```ruby
docs.sort_by! { |doc| metadata["order"].index(doc.relative_path.sub(relative_directory, "")) || docs.size }
```

ashmaroli:
Thank you for the review @parkr. I'll push the changes a.s.a.p

parkr:
@ashmaroli Gladly! Sorry it took so long. Looking forward to using this feature. 😄

ashmaroli:
@parkr I have addressed *some* of the concerns you expressed earlier. The ones I left out are justified below:

1.  **replace `rearrange_docs!` with `docs.sort_by! { |doc| metadata["order"].index(doc.relative_path.sub(relative_directory, "")) || docs.size }`**
While I understand your intent, `|| docs.size` won't work because `sort_by` sorts entries based on entries' properties rather than the array properties.
Also, with `rearrange_docs!`, I get to inform the user that they entered an incorrect path in their configuration.
However, I kinda handled the "data-munging" to an extent by emulating Ruby's `tap` method &mdash; to define a virtual-key (`#{label}_sort_index`) and have it removed at the end of the operation &mdash; a side-effect being that should such a key be defined by the user, it would thenceforth no longer be available for `render`

2. **handle the nil case as well (when `meta_key` in `sort_docs_by_key!` or when `metadata["order"]` is nil)**
These are `private` methods that are run as part of a conditional block that checks for `nil`

3. **wrap in `Array(metadata["order"])`**
Same as `#2`. Part of a `private` method that is only run if the variable is an array.

parkr:
@ashmaroli Thanks for the updates!

1. I will not approve a patch which stores data on users' Documents. `rearrage_docs!` will need to find another way to store this data for its processing. `sort_by` uses whatever return value the block gets so if the document doesn't have a sort value, then `|| docs.size` will put the document at the end of the list, since the first part would return `nil`, which is falsy.
2. This logic should be central to the method (to guard against refactoring that will inevitably remove that nil check in the other method some years down the line), and it's very easy in Ruby to call a private method, so privacy is not a guarantee.
3. This is the easiest way to handle `nil` and we use it throughout this codebase where `nil` or a single value can be passed when we explicitly require an array for the code. This kind of defensive coding is important when coding for our many thousands of users with differing needs and expectations about how things should work.

ashmaroli:
> I will not approve a patch which stores data on users' Documents. rearrage_docs! will need to find another way to store this data for its processing.

The latest patch doesn't store the sort-index value in the document.. instead it gets assigned and deleted..
But yeah, I've been thinking of an alternative..


> This kind of defensive coding is important when coding for our many thousands of users

I agree. But if the user submits  a Hash instead of Array then my implementation would fail with `Array()`. For example:
```yml
collections:
  tutorials:
    output: true
    order:
      foo: bar
      lorem: ipsum
```
Then
```ruby
Array(metadata["order"])
# => [["foo", "bar"],["lorem, "ipsum"]]
```

ashmaroli:
@gvwilson Are you willing to forego a part of this PR's proposal to have the remainder merged in?
i.e., **remove the ability to sort documents by manually listing the order** and instead **be able to sort documents by any defined Front Matter key** in the near future?

The `manual-rearrange` feature can be handled in a future PR..

ashmaroli:
@parkr I've trimmed the patch to only enable sorting documents by their pre-defined front matter keys. No more data-munging..
Fallback to existing sort (`docs.sort`) if the collection does not have the key defined in all documents.

@gvwilson manual rearrangement will make a return if I can come up with a way to achieve it without modifying the documents themselves.

features/collections.feature
```diff
@@ -207,6 +207,73 @@ Feature: Collections
...
+          path: ""
+          type: tutorials
+        values:
+          layout: tutorial
+
+    collections:
+      tutorials:
+        output: true
+        sort_by: lesson
+
```
parkr: Empty space.

features/collections.feature
```diff
@@ -207,6 +207,73 @@ Feature: Collections
...
+    """
+    When I run jekyll build
+    Then I should get a zero exit status
+    Then the _site directory should exist
```
parkr: This should be `And ...` instead of a repeated `Then ...`

features/collections.feature
```diff
@@ -207,6 +207,73 @@ Feature: Collections
...
+          path: ""
+          type: tutorials
+        values:
+          layout: tutorial
+
+    collections:
+      tutorials:
+        output: true
+        sort_by: approx_time
+
```
parkr: Empty line here

lib/jekyll/collection.rb
```diff
@@ -210,7 +210,28 @@ def read_document(full_path)
...
-    private
+    def sort_docs
+      return docs.sort! unless metadata["sort_by"].is_a?(String)
+      sort_docs_by_key!
+    end
+
+    def sort_docs_by_key!
+      meta_key = metadata["sort_by"]
+      docs.sort_by! { |d| d.data[meta_key] }
+    rescue ArgumentError
```
parkr: This is unacceptable to deal with `d.data[meta_key]`. A `nil` value must be dealt with in the `sort_by!` block.

parkr: One way to handle this is to sort by an array of values:

```ruby
# sort by the key's value, then by its path if the first is the same.
docs.sort_by! { |d| [d.data[meta_key], d.path]  }
```

This would be my preferred approach.

```ruby
>> nil <=> nil
=> 0
>> nil <=> "hi"
=> nil
>> "hi" <=> nil
=> nil
```

test/test_collections.rb
```diff
@@ -173,6 +173,48 @@ class TestCollections < JekyllUnitTest
...
+          "tutorials" => {
+            "output"  => true,
+            "sort_by" => "lesson",
+          },
+        },
+      })
+      @site.process
+      @tutorials_collection = @site.collections["tutorials"]
+
+      @actual_array = @tutorials_collection.docs.map(&:relative_path)
```
parkr: Let's call this `@collection_doc_relative_paths` or something. `@actual_array` is a description of the data structure, not the content.

lib/jekyll/collection.rb
ashmaroli: @parkr I think you missed the point why I rescued `ArgumentError` here:
An array of objects **cannot be sorted** by an attribute **if all constituent elements *do not* have the attribute defined**

So while your suggestion can be adapted to allow users to sort by multiple attributes, it'll fail if one of the attribute is not defined in all of the constituent documents..

parkr: @ashmaroli Well, I just spent the last hour digging and digging and I found something interesting. Of course, we _do_ allow sorting by attributes which all constituent elements may not have in `Jekyll::Filters#sort`. We don't use `sort_by` there. We use `sort`. I was digging through the documentation earlier for `sort_by` and I came across a mention of a Schwartzian transform, which is used to speed up sort calls by caching values. I wrote a benchmark and updated `Jekyll::Filters#sort_input` to use this: https://github.com/jekyll/jekyll/pull/6342

Using `rescue ArgumentError` is using a `rescue` as a control flow mechanism. At the very least, we should be checking the properties using `return docs.sort! unless docs.all? { |d| d.data.key?(meta_key) }` instead of hoping that `sort_by` will always raise `ArgumentError`.

I'd propose we fall back to `path` if the key doesn't exist:

```ruby
meta_key = metadata["sort_by"]
docs.map! do |d| 
  Jekyll.logger.warn "Sort warning:", "Missing sort key '#{meta_key}' on document #{d.relative_path}"
  [d.data[meta_key], d]
end.sort! do |apple, orange|
  if !apple.first.nil? && !orange.first.nil?
    # If both documents have the property, sort by that property.
    puts "Sorting by the property"
    apple.first <=> orange.first
  elsif !apple.first.nil? && orange.first.nil?
    -1
  elsif apple.first.nil? && !orange.first.nil?
     1
  else
    # Fall back to Document#<=> if both documents don't have the property.
    apple.last <=> orange.last
  end
end.map! do |apple|
  apple.last # return the Document object
end
```

parkr:
Ruby 2.1.10 consistently fails here :( https://travis-ci.org/jekyll/jekyll/jobs/271100330

ashmaroli:
> Ruby 2.1.10 consistently fails here

... and also on AppVeyor @ Ruby2.4.x ...
Feel free to take this PR off the queue to `Milestone v3.6`

parkr:
Sorry, @ashmaroli 😦 I don't know why it's failing! Maybe something with the `!` methods' return values changed in Ruby 2.1?

gvwilson:
Can someone please take another look at this - it would greatly simplify our lives to have it merged. cc @parkr @ashmaroli @pathawks

ashmaroli:
@gvwilson Over the course of time, the branch [changed ideas](https://github.com/jekyll/jekyll/pull/5904#issuecomment-320465145) and is now *different* from your original *feature-request*.. Even with this PR (as it stands currently) being merged in you'll not be able to rearrange collection documents manually by listing documents in the config file.

gvwilson:
That's very disappointing, and not what we agreed when I paid you for this work.

gvwilson:
That's really unfortunate: the bounty I paid was precisely so that collection documents' order *could* be specified in the configuration file.

ashmaroli:
@gvwilson The Core Team is not willing to incorporate the proposed implementation into Jekyll Core.
The pamphlet associated with the bounty did state that the contributor will not be held responsible if
the PR would not get merged.

Besides, you need not wire your site's source to GitHub Pages to get it built on the fly, there are other alternatives like Netlify.com that allow you to use custom plugins and gems in the build sessions.

PR #5905, opened 2017-02-26
update broken links

Strangehill:
where once there was a working url, there's now a broken redirect
@jekyll/documentation


DirtyF:
@jekyllbot: merge +docs

Issue #5901, opened 2017-02-23
`:path` permalink variable not consistent with documentation – which needs fixing?


rlue:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/

---

- [x] I am on (or have tested on) ***macOS*** 10+

---

- [x] I was trying to build.

---

When specifying a permalink configuration, the `:path` variable behaves differently from what is presented in the documentation.

Specifically, the [documentation](https://jekyllrb.com/docs/collections/#permalinks) presents the `:path` variable as consisting of the entire path relative to the originating `_collection` directory, **excluding** the filename itself. Thus, wherever `:path` appears in an example configuration, it is followed by `/:title`.

In practice, it **includes** the filename, making the inclusion of`/:title` redundant. Thus, Example Configuration 2 (which is supposed to be identical to the default configuration),

```yaml
permalink: /:collection/:path/:title:output_ext
```

produces this:

```
├── apidocs
│   └── mydocs
│       └── doc1             <-- `doc1.html` should be here.
│           └── doc1.html
```

To manually reproduce the default configuration, `/:title:output_ext` must be removed:

```yaml
permalink: /:collection/:path
```

I have confirmed this bug occurs in the latest version of Jekyll (3.4.0), without the Github Pages gem. If the error is the documentation, I'm happy to submit a PR – but if the documentation is describing the correct behavior and the error is in parsing the permalink configuration, I'm afraid this one may be over my head.

/cc @jekyll/build @jekyll/documentation

parkr:
The standard permalink is `/:collection/:path/`, which creates `/apidocs/mydocs/doc1/index.html`. `:path` is everything after the collection name, which defaults to a directory as specified.

It sounds like what we want here is a `:dir` template variable which would be just `/mydocs`.

Would you be willing to create a PR? This will modify the `URLDrop` class to add a `dir` method. It may require a `dir` method to be added to `Document`.

rlue:
Thanks for helping get to the bottom of this. To be clear, I don't personally have much need for a `:dir` template variable. My issue was merely that `:path`'s behavior was different from what is described in the docs, and I wasn't sure which was “correct”. 

I will be submitting a PR for the docs instead, if that's peachy by you.

rlue:
[...]

PR #5912, opened 2017-02-28
Revamp Permalink section

rlue:
The previous version contained errors regarding the definition and usage of the `:path` permalink template variable (see Issue #5901). This version corrects those errors, removes redundant language where possible, and restructures the section to place examples before template variable definitions.

The reasoning behind the third change is that it's generally easier for learners to understand a generalized or abstracted definition after having seen concrete examples in action first.

### DIFF
--- a/docs/_docs/collections.md
+++ b/docs/_docs/collections.md
@@ -89,56 +89,67 @@ choice and written out to `<dest>/my_collection/some_subdir/some_doc.html`.
 
 ## Configuring permalinks for collections {#permalinks}
 
-You can specify a pattern for the URLs where your Collection documents will reside with the [`permalink` property](../permalinks/):
+If you wish to specify a custom pattern for the URLs where your Collection pages
+will reside, you may do so with the [`permalink` property](../permalinks/):
 
 ```yaml
 collections:
   categories:
-    output: true                  # without this flag, this Collection's documents will not be generated at build time
-    permalink: /category/:name    # custom permalink definition
+    output: true
+    permalink: /category/:name
 ```
 
 ### Examples
 
-For a collection named `apidocs` with the following source file structure,
+For a collection with the following source file structure,
 
 ```
-├── _apidocs
-│   └── archive
-│       └── deprecated.md
+_my_collection/
+└── some_subdir
+    └── some_doc.md
 ```
 
-the examples below present various `permalink` configurations and the document structures they produce under the `_site` directory.
+each of the following `permalink` configurations will produce the document structure shown below it.
 
 * **Default**  
   Same as `permalink: /:collection/:path`.
   ```
-  ├── apidocs
-  │   └── archive
-  │       └── deprecated.html
+  _site/
+  ├── my_collection
+  │   └── some_subdir
+  │       └── some_doc.html
+  ...
   ```
 * `permalink: pretty`  
   Same as `permalink: /:collection/:path/`.
   ```
-  ├── apidocs
-  │   └── archive
-  │       └── deprecated
+  _site/
+  ├── my_collection
+  │   └── some_subdir
+  │       └── some_doc
   │           └── index.html
+  ...
   ```
 * `permalink: /doc/:path`
   ```
+  _site/
   ├── doc
-  │   └── archive
-  │       └── deprecated.html
+  │   └── some_subdir
+  │       └── some_doc.html
+  ...
   ```
 * `permalink: /doc/:name`
   ```
+  _site/
   ├── doc
-  │   └── deprecated.html
+  │   └── some_doc.html
+  ...
   ```
 * `permalink: /:name`
   ```
-  ├── deprecated.html
+  _site/
+  ├── some_doc.html
+  ...
   ```
 
 ### Template Variables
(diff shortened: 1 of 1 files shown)


docs/_docs/collections.md
```diff
@@ -89,18 +89,59 @@ choice and written out to `<dest>/my_collection/some_subdir/some_doc.html`.
...
 
-In this example, the collection documents will the have the URL of `awesome` followed by the path to the document and its file extension.
+### Examples
+
+For a collection named `apidocs` with the following source file structure,
+
+```
+├── _apidocs
+│   └── archive
+│       └── deprecated.md
```
tomjohnson1492: Why do you have "deprecated.md" as an example of the file here? What's this about?

rlue: Ha, I was worried that might raise an eyebrow. 

I figured that a generic-but-realistic use case would make for a more descriptive example than `apidocs/mydocs/doc1.md`; _i.e.,_ that it would be easier to imagine why you might choose to set your permalinks up as, say, `doc/deprecated.html` than `awesome/doc1.html`. But I also wrote it in a hurry, and I recognize that maybe `apidocs/archive/deprecated.md` is not as realistic as it could be.

If it creates more confusion than it resolves, I'm happy to change it back to `apidocs/mydocs/doc1.md` and resubmit the PR.

tomjohnson1492: I suggest using a generic file name that doesn't call attention to itself like "deprecated" does. Maybe myfile.my or something. I have some more comments that I'll insert inline.

docs/_docs/collections.md
```diff
@@ -89,18 +89,59 @@ choice and written out to `<dest>/my_collection/some_subdir/some_doc.html`.
 
 ## Configuring permalinks for collections {#permalinks}
 
-You can customize the [Permalinks](../permalinks/) for your collection's documents by setting `permalink` property in the collection's configuration as follows:
+You can specify a pattern for the URLs where your Collection documents will reside with the [`permalink` property](../permalinks/):
```
tomjohnson1492: There's a modifier problem with the last "with the ..." clause. Maybe recast as "Through the [`permalink` property](../permalinks/), you can specify a pattern for the URLs for your Collection documents. The permalink also defines the structure for the files in the output."

tomjohnson1492:
In the Examples section, the tree formatting is messed up. This may be intentional, but it looks awkward like this to me. Can you preserve the tree formatting in the examples in a way that leaves the tree diagrams complete?

tomjohnson1492:
If you don't want to use the complete tree diagrams there, you could use forward slashes instead.

docs/_docs/collections.md
DirtyF: We *could* use a more generic file name, but I don't mind the current example here, seems pretty explanatory to me.

parkr:
@DirtyF Where are you with this? Need more info?

rlue:
Sorry to ghost you guys, I'm happy to make the changes @tomjohnson1492 requested, and will get to work on it this weekend.

rlue:
Okay folks, I've made the changes requested.

@tomjohnson1492:

* In the interest of consistency, I abandoned the `apidocs` example Collection altogether in favor of `_my_collection/some_subdir/some_doc.md` (which appears in [the section immediately preceding Permalinks](https://jekyllrb.com/docs/collections/#step3))
* Thanks for pointing out the misplaced modifier. Hopefully the new version is to your liking?
* Not sure exactly what was wrong with the tree diagrams in my last commit; I simply copied the format as it was [before my edits](https://jekyllrb.com/docs/collections/#permalink-examples-for-collections). I suspect you either meant that root directories for the tree diagrams were absent, or that in the first column, vertical lines descended down from the top and didn't connect to anything underneath.

  To address the first point, I added root directories to each tree diagram. To address the second, I added ellipses to the bottom of some tree diagrams to indicate that there were irrelevant portions omitted.

Let me know if there's anything else I should do!

docs/_docs/collections.md
```diff
@@ -89,18 +89,70 @@ choice and written out to `<dest>/my_collection/some_subdir/some_doc.html`.
...
 ## Configuring permalinks for collections {#permalinks}
 
-You can customize the [Permalinks](../permalinks/) for your collection's documents by setting `permalink` property in the collection's configuration as follows:
+If you wish to specify a custom pattern for the URLs where your Collection pages
+will reside, you may do so with the [`permalink` property](../permalinks/):
 
 ```yaml
 collections:
-  my_collection:
+  categories:
```
DirtyF: Did you mean to replace `my_collection` with `categories`here?

docs/_docs/collections.md
```diff
@@ -89,18 +89,70 @@ choice and written out to `<dest>/my_collection/some_subdir/some_doc.html`.
...
+└── some_subdir
+    └── some_doc.md
+```
 
-Collections have the following template variables available for permalinks:
+each of the following `permalink` configurations will produce the document structure shown below it.
+
+* **Default**  
+  Same as `permalink: /:collection/:path`.
+  ```
```
DirtyF: Please insert carriage returns to create a paragraph before each code sample

```
*  **Default**

Same as `permalink: /:collection/:path`.

```

rlue:
Good looking out!

DirtyF:
@rlue Thanks for making Jekyll's documentation better 📗 

@jekyllbot: merge +docs

PR #5914, opened 2017-02-28
Require `runtime_dependencies` of a Gem-based theme from its `.gemspec` file

ashmaroli:
Have Jekyll require the `runtime_dependencies` of a theme-gem from its `.gemspec` file so that the theme's consumers need not have to edit their `gems` array to include plugins required by their theme.

--
/cc @jekyll/ecosystem

### DIFF
--- a/docs/_docs/themes.md
+++ b/docs/_docs/themes.md
@@ -206,6 +206,12 @@ Your theme's styles can be included in the user's stylesheet using the `@import`
 {% raw %}@import "{{ site.theme }}";{% endraw %}
 ```
 
+### Theme-gem dependencies
+
+From `v3.5`, Jekyll will automatically require all `whitelist`-ed `runtime_dependencies` of your theme-gem even if they're not explicitly included under the `gems` array in the site's config file. (NOTE: `whitelist`-ing is only required when `build`-ing or `serve`-ing with the `--safe` option.)
+
+With this, the end-user need not keep track of the plugins required to be included in their config file for their theme-gem to work as intended.
+
 ### Documenting your theme
 
 Your theme should include a `/README.md` file, which explains how site authors can install and use your theme. What layouts are included? What includes? Do they need to add anything special to their site's configuration file?
--- a/docs/_docs/themes.md
+++ b/docs/_docs/themes.md
@@ -208,7 +208,7 @@ Your theme's styles can be included in the user's stylesheet using the `@import`
 
 ### Theme-gem dependencies
 
-From `v3.5`, Jekyll will automatically require all `whitelist`-ed `runtime_dependencies` of your theme-gem even if they're not explicitly included under the `gems` array in the site's config file. (NOTE: `whitelist`-ing is only required when `build`-ing or `serve`-ing with the `--safe` option.)
+From `v3.5`, Jekyll will automatically require all whitelisted `runtime_dependencies` of your theme-gem even if they're not explicitly included under the `gems` array in the site's config file. (NOTE: whitelisting is only required when building or serving with the `--safe` option.)
 
 With this, the end-user need not keep track of the plugins required to be included in their config file for their theme-gem to work as intended.
 
--- a/lib/jekyll/plugin_manager.rb
+++ b/lib/jekyll/plugin_manager.rb
@@ -32,11 +32,9 @@ module Jekyll
 
     def require_theme_deps
       if site.theme.runtime_dependencies
-        dependencies = site.theme.runtime_dependencies.select do |dep|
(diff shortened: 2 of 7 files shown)


ashmaroli:
Requesting a restart of the Travis build for this PR..
Thanks :smiley:

parkr:
I couldn't load the page to trigger a restart... Try pushing another commit?

ashmaroli:
> Try pushing another commit?

okay..

ashmaroli:
label: `themes`

lib/jekyll/plugin_manager.rb
```diff
@@ -29,6 +30,17 @@ def require_gems
...
     end
 
+    def require_theme_deps
+      if site.theme.runtime_dependencies
+        dependencies = site.theme.runtime_dependencies.select do |dep|
+          plugin_allowed?(dep.name)
+        end
+        dependencies.each do |dep|
+          External.require_with_graceful_fail(dep.name) unless dep.name == "jekyll"
+        end
```
parkr: You could simplify this to

```ruby
site.theme.runtime_dependencies.each do |dep|
  next if dep.name == "jekyll"
  External.require_with_graceful_fail(dep.name) if plugin_allowed?(dep.name)
end
```

ashmaroli:
refactored..
/cc @parkr

lib/jekyll/plugin_manager.rb
```diff
@@ -29,6 +30,15 @@ def require_gems
       )
     end
 
+    def require_theme_deps
+      if site.theme.runtime_dependencies
+        site.theme.runtime_dependencies.each do |dep|
```
parkr: We tend to prefer guard block style, which is:

```ruby
def require_theme_deps
  return false unless site.theme && site.theme.runtime_dependencies
  # ... do the actual work, expecting the above variables to exist
end
```

You could alternatively use `#Array()`, which will create an empty array if the value is `nil`:

```ruby
def require_theme_deps
  Array(site.theme.runtime_dependencies).each do |dep|
    # ...
  end
end
```

parkr:
@pathawks Any objections to the idea here?

lib/jekyll/plugin_manager.rb
ashmaroli: I gez I'll go with the explicit `return` style instead of the alternative wherein a block is called even on an empty array.

docs/_docs/themes.md
```diff
@@ -206,6 +206,12 @@ Your theme's styles can be included in the user's stylesheet using the `@import`
 {% raw %}@import "{{ site.theme }}";{% endraw %}
 ```
 
+### Theme-gem dependencies
+
+From `v3.5`, Jekyll will automatically require all `whitelist`-ed `runtime_dependencies` of your theme-gem even if they're not explicitly included under the `gems` array in the site's config file. (NOTE: `whitelist`-ing is only required when `build`-ing or `serve`-ing with the `--safe` option.)
+
+With this, the end-user need not keep track of the plugins required to be included in their config file for their theme-gem to work as intended.
+
```
ashmaroli: Added a piece of documentation for this particular feature.
/cc @DirtyF

DirtyF: @ashmaroli Could we please stick to plain english instead of mixing commands and dashes to ease readability?

ashmaroli: Sure.
Something like this: ?
```
From `v3.5`, Jekyll will automatically require all whitelisted `runtime_dependencies` of
your theme-gem even if they're not explicitly included under the `gems` array in the
site's config file. (NOTE: whitelisting is only required when building or serving with the
`--safe` option.)

With this, the end-user need not keep track of the plugins required to be included in their
config file for their theme-gem to work as intended.
```

ashmaroli:
I just realized that this may have an unintended side-effect:
Users won't be able to disable a certain plugin if they wish to. Say, a theme listed `jekyll-feed` as a dependency. But the user would like to use another plugin to generate their feeds. There's now no way to do so..

pathawks:
Themes should only require what they require.

If a user wants something lighter weight, they can either use a lighter theme, or fork a theme and remove the cruft.

parkr:
@jekyllbot: merge +minor

ashmaroli:
Hi @tlnagy the consensus is that a theme should henceforth list only the very essential plugins (i.e. those without which, your theme would simply fail) as `runtime_dependencies`. The non-essential ones can be brought to the end-user's notice with effective documentation.
They're then free to add the non-essential plugins to their config file as required.

PR #5920, opened 2017-03-02
Allow abbreviated post dates

### LINKED ISSUE #5879: Bug: Jekyll post loop not returning anything on GitLab Pages, but returning posts on GitHub pages
I've run into something I can't seem to figure out. I push my Jekyll theme to GitLab, which I have set up to mirror the repository to GitHub. Because of this, the two repos (should be) exactly the same. However, GitLab Pages, which uses Jekyll, is producing an issue, while GitHub Pages is not.

I wanted to remove the `url` variable from the `_config.yml` file so that I can have the demo from GitLab run on GitLab Pages and the demo from GitHub run on GitHub Pages. I used the new-ish `relative_url` filter on some paths and everything seems to be working, except for [a posts loop](https://gitlab.com/andrewbanchich/forty-jekyll-theme/blob/master/_includes/tiles.html). You can see the output of this code on the homepage in [the GitHub Pages demo](https://andrewbanchich.github.io/forty-jekyll-theme/) but, as you can see in [the GitLab Pages demo](https://andrewbanchich.gitlab.io/forty-jekyll-theme/), the tiles are not displaying.

I'm not sure what is causing this, but if the two code bases are the same and producing different results on different sites, wouldn't that imply there is an issue with them being processed?

- [Link to GitLab repo](https://gitlab.com/andrewbanchich/forty-jekyll-theme)
- [Link to GitHub repo](https://github.com/andrewbanchich/forty-jekyll-theme)

parkr:
In https://github.com/jekyll/jekyll/pull/5609, we made the date matching much stricter: it had to be in `YYYY-MM-DD`. In a brief test on GitHub Pages, we noticed this issue.

This PR loosens this to allow 2-digit years, and 1-digit months and days. That makes this a valid filename:

```text
_posts/12-2-5-my-post.md
```

I'm iffy on allowing 2-digit years, but the previous regexp allowed it (just three instances of `\d+`).

Fixes #5879 as well, @andrewbanchich. Would you please test this out?

I'm hoping to get this out as a v3.4.1 and include this patch in v3.5 as well.

/cc @jekyll/stability

### DIFF
--- a/test/test_generated_site.rb
+++ b/test/test_generated_site.rb
@@ -49,7 +49,9 @@ class TestGeneratedSite < JekyllUnitTest
     end
 
     should "include a post with a abbreviated dates" do
-      refute_nil -1, @site.posts.index { |post| post.relative_path == "_posts/2017-2-5-i-dont-like-zeroes.md" }
+      refute_nil @site.posts.index { |post|
+        post.relative_path == "_posts/2017-2-5-i-dont-like-zeroes.md"
+      }
       assert_exist dest_dir("2017", "02", "05", "i-dont-like-zeroes.html")
     end
 
--- a/test/test_generated_site.rb
+++ b/test/test_generated_site.rb
@@ -11,7 +11,7 @@ class TestGeneratedSite < JekyllUnitTest
     end
 
     should "ensure post count is as expected" do
-      assert_equal 51, @site.posts.size
+      assert_equal 52, @site.posts.size
     end
 
     should "insert site.posts into the index" do
@@ -46,6 +46,11 @@ class TestGeneratedSite < JekyllUnitTest
     should "process other static files and generate correct permalinks" do
       assert_exist dest_dir("contacts.html")
       assert_exist dest_dir("dynamic_file.php")
+    end
+
+    should "include a post with a abbreviated dates" do
+      refute_nil -1, @site.posts.index { |post| post.relative_path == "_posts/2017-2-5-i-dont-like-zeroes.md" }
+      assert_exist dest_dir("2017", "02", "05", "i-dont-like-zeroes.html")
     end
 
     should "print a nice list of static files" do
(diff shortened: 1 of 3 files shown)


ashmaroli:
is this going to be merged into `master`?
I think **`3.4.1`** should branch off the `3.4-stable`. `master` already contains enhancements that bumps to `3.5`
Thoughts?

parkr:
> is this going to be merged into master?
> I think 3.4.1 should branch off the 3.4-stable. master already contains enhancements that bumps to 3.5
> Thoughts?

You're correct! This is still an issue on master, though. Once this is merged, this will  be backported.

test/test_generated_site.rb
```diff
@@ -49,7 +49,9 @@ class TestGeneratedSite < JekyllUnitTest
     end
 
     should "include a post with a abbreviated dates" do
-      refute_nil -1, @site.posts.index { |post| post.relative_path == "_posts/2017-2-5-i-dont-like-zeroes.md" }
+      refute_nil @site.posts.index { |post|
+        post.relative_path == "_posts/2017-2-5-i-dont-like-zeroes.md"
+      }
```
ashmaroli: any reason you decided to depart from the convention of using `do..end` for code blocks?
```ruby
      refute_nil @site.posts.index do |post|
        post.relative_path == "_posts/2017-2-5-i-dont-like-zeroes.md"
     end
```

parkr: In this case, I like to imaging parentheses around the argument, like `refute_nil(@site.posts.index...)`. The use of the curly braces indicate to me that this is used as an argument to `refute_nil`, whereas your example above doesn't read that directly. By diverging from the standard `do...end`, I call to attention that this statement is part of an argument.

parkr:
@jekyllbot: merge +bug

parkr:
v3.4.1 released! Working on some release notes.

PR #5926, opened 2017-03-02
Allow releases & development on *-stable branches

parkr:
This should make backporting easier as we'll have CI running on those PR's and allow releasing from the *-stable branches.


parkr:
@jekyllbot: merge +dev

Issue #5929, opened 2017-03-03
Default layout overwriting local layout if layout == null


davebowker:
Hey,

Quick issue I've run into. I have a default scope applying a layout to a particular folder. I want to put a search.json file into that folder. To do that without having jekyll generate the page I set the `permalink: /folder/search.json` in the front matter and the `layout: null`

What's happening is, if the front matter value for layout is null, then the global default overwrites it. (FYI, if the front matter value for layout is something else, eg 'another-layout' then the local variable applies)

I think this may be a bug, but is there an alternate way to do what I want? I tried `output:false` but no joy. 

Thanks,
Dave

ashmaroli:
if `search.json` has YAML FrontMatter and should not be output to the destination `_site` folder, remove the `layout: null` and add `published: false`

pathawks:
Do you not want Jekyll to include it in the generated site, or do you want Jekyll to copy it without applying a layout?

davebowker:
What @pathawks said. I want it to be output, but I don't want the json file to have a layout.

ashmaroli:
have you tried removing the FrontMatter block entirely?
Without the FrontMatter dashes, a file will be seen as a `static_file` which will be simply copied to `destination` without any processing..

davebowker:
I have, but the liquid tags that loop through the posts and outputs the json doesn't get processed. It does treat it as a static file, basically copying verbatim what I had into the `serve` directory.

pathawks:
The easiest fix for this may be to create a new layout that only contains `{{ content }}` and use that as your layout for JSON files

ashmaroli:
Ah! so you need the file to be processed, yet should not have the default layout set for the directory. The way out is to actually use an empty layout, say for example:
```
---
# _layouts/json.html
---
{{ content }}
```

ashmaroli:
@pathawks beat me to it by a few seconds.. :laughing:

davebowker:
Yeah, I actually had the same fix applied already. I still think this is a bug though, as a local scope should override a global scope (and does for any other layout, just not for `null`)

ashmaroli:
its not a bug.. it is by design..
`layout: null` is the same as no layouts specified..

davebowker:
Yeah, and that's what it should be, no...?

- Global scope tells jekyll to use layout-X on all files in a particular folder
- Local scope tells jekyll to use layout-Y on a specific file, whether it has a global scope applied to it or not

[...]

PR #5933, opened 2017-03-04
Disable default layouts for documents with a `layout: none` declaration

ashmaroli:
Inspired by discussion in #5929 

While `layout: null` will still continue to be overridden by the `defaults`, with this pull in place, the user will be able to process and output a `Jekyll::Document` object without being placed into a layout by 
simply declaring `layout: none` in the Front Matter

/cc @jekyll/ecosystem 

Let me know if this results in any unintended behavior..  I'll update as required.

### DIFF
--- a/features/rendering.feature
+++ b/features/rendering.feature
@@ -40,6 +40,44 @@ Feature: Rendering
     And I should not see "Ahoy, indeed!" in "_site/index.css"
     And I should not see "Ahoy, indeed!" in "_site/index.js"
 
+  Scenario: Ignore defaults and don't place documents with layout set to 'none'
+    Given I have a "index.md" page with layout "none" that contains "Hi there, {{ site.author }}!"
+    And I have a _trials directory
+    And I have a "_trials/no-layout.md" page with layout "none" that contains "Hi there, {{ site.author }}!"
+    And I have a "_trials/test.md" page with layout "null" that contains "Hi there, {{ site.author }}!"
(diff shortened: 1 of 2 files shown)


parkr:
This should be `layout: nil`.

ashmaroli:
> This should be layout: null. I don't know if I like us fiddling with the way YAML works.

IMO, the proposed approach is better because it introduces a new route and only affects `Jekyll::Document` objects..
Modifying `layout: null` will require modification of `#deep_merge_hashes` which is used by various other objects

parkr:
Ugh, I don't love this. If I had a `_layouts/none.html` file and `layout: none` today, then after this PR, I wouldn't have a layout on my page anymore. That seems like a breaking change. 

A [brief search shows that it's probably safe...](https://github.com/search?q=%22layout%3A+none%22&type=Code&utf8=%E2%9C%93).

@benbalter, I think this could be an acceptable breaking change. Thoughts?

mattr-:
IMHO, this is definitely a breaking change. I'm also not sure what the use case for this would be when we have the standard YAML version already in `layout: null`.

pathawks:
> we have the standard YAML version already in `layout: null`.

That's the problem: we don't. When we merge the defaults into a document’s front matter, we will override `layout: null` the same way we would if no layout were specified at all.

parkr:
@jekyllbot: merge +minor

Issue #5944, opened 2017-03-10
feat(404): customized 404 page


MiniJerome:
- [ ] I believe this to be a bug, not a question about using Jekyll.
- [x] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [x] This is a feature request.

---

- [x] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux
- [x] I am on (or have tested on) ***Docker*** image

---

## The Awesome Feature I dream of

Hey, 

I would like to customize a 404 page. I have insert the following code at the line 332 of the file `/usr/lib/ruby/2.3.0/webrick/httpresponse.rb `: 

```
    def create_error_page                                                                                        
      @body = ''                                                  
      @body << <<-_end_of_html_                                   
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0//EN">                                                                                       
<HTML>                                                                    
  <HEAD><TITLE>#{HTMLUtils::escape(@reason_phrase)}</TITLE></HEAD>
  <BODY>                                                  
    <H1>#{HTMLUtils::escape(@reason_phrase)}</H1>         
    <HR>                                                  
    <P>Custom error page!</P>                                    
  </BODY>                                                                   
</HTML>                                                        
      _end_of_html_                                            
    end    
```

Then I have adding this example page (404.md) in the root dir of my website: 

```
---
layout: post
title: Lost in Internet
permalink: /404.html
comments: false
---

Welcome nowhere!
``` 

Then every requests to anything that doesn't exists are redirected to this page instead of : 

```
Not Found

`/pouet' not found.
WEBrick/1.3.1 (Ruby/2.3.3/2016-11-21) at your-domain-name:port
```

Anyway, let's me know if this is possible to add this feature. 

Thanks!

pathawks:
I don't think we intend for users to use Webrick in production. Can you provide any detail regarding _why_ you need this?

ashmaroli:
I don't see the need of patching WEBrick. The best solution would be to install the `404.md` via `jekyll new`. You're welcome to create a pull implementing that.

MiniJerome:
Well, I'm confusing this issue seems to be useless.

PR #5945, opened 2017-03-11
Add a template for custom 404 page

ashmaroli:
Ref: #5944 

Provide a Custom 404 Error Page with `jekyll new`

Personally, not very happy about including internal CSS. However, this is preferable until Minima decides to ship a layout dedicated to this page.

/cc @pathawks @benbalter

### DIFF
--- a/lib/site_template/404.html
+++ b/lib/site_template/404.html
@@ -1,0 +1,29 @@
+---
+layout: default
+---
+
+<style type="text/css" media="screen">
+  .container {
+    margin: 10px auto;
+    width: 600px;
+    text-align: center;
+  }
+  h1 {
+    width: 800px;
+    position: relative;
+    left: -100px;
+    margin: 30px 0;
+    font-size: 60px;
+    line-height: 1;
+    letter-spacing: -1px;
+  }
+</style>
+
+<div class="container">
+  <h1>404</h1>
+
+  <p><strong>Page not found :(</strong></p>
+  <p>The requested page could not be found.</p>
+
+  <p>Edit the <code>404.html</code> file in your site to show a custom 404 Error Message.</p>
+</div>
--- a/lib/site_template/404.html
+++ b/lib/site_template/404.html
@@ -5,15 +5,12 @@ layout: default
 <style type="text/css" media="screen">
   .container {
     margin: 10px auto;
-    width: 600px;
+    max-width: 600px;
     text-align: center;
   }
   h1 {
-    width: 800px;
-    position: relative;
-    left: -100px;
     margin: 30px 0;
-    font-size: 60px;
+    font-size: 4em;
     line-height: 1;
     letter-spacing: -1px;
   }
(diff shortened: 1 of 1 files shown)


crispgm:
It's a good idea. However, it may hardly take effect because we haven't taken control of webserver.

ashmaroli:
This works on `localhost:4000`
This should work on GH Pages as per [this Help Article](https://help.github.com/articles/creating-a-custom-404-page-for-your-github-pages-site/)

For Apache webservers, users may need to modify their `.htaccess` but that's a different story.

ashmaroli:
A Custom 404 page on the WEBrick development server

ashmaroli:
It does work on GH Pages. [Sample this](https://ashmaroli.github.io/test-site/aloha)

pathawks:
Thanks @ashmaroli! Is this documented in _our_ docs? If not, it should be :beers:

ashmaroli:
> Is this documented..

Pardon me, but what exactly were you pointing at?

crispgm:
@ashmaroli Looks cool! I think we should add a new section on docs to show how to make it work with webservers like `nginx`.

ashmaroli:
> I think we should add a new section on docs to show how to make it work with webservers like nginx.

Since this works out-of-the-box on GitHub Pages, I feel *a how to..* documentation for `nginx` and `Apache` softwares should be on a separate PR to the `tutorials` collection.

lib/site_template/404.html
```diff
@@ -0,0 +1,29 @@
...
+    margin: 10px auto;
+    width: 600px;
+    text-align: center;
+  }
+  h1 {
+    width: 800px;
+    position: relative;
+    left: -100px;
+    margin: 30px 0;
+    font-size: 60px;
```
DirtyF: One commonly used blog template should be mobile-first and responsive, even here.

ashmaroli: Thnx @DirtyF I had totally ignored mobile-friendliness.. 
Will update.

ashmaroli: updated.

404m

lib/site_template/404.html
```diff
@@ -0,0 +1,26 @@
...
+  }
+</style>
+
+<div class="container">
+  <h1>404</h1>
+
+  <p><strong>Page not found :(</strong></p>
+  <p>The requested page could not be found.</p>
+
+  <p>Edit the <code>404.html</code> file in your site to show a custom 404 Error Message.</p>
```
benbalter: Perhaps, "If you're the maintainer of the site, edit the..."

ashmaroli: I intended it to be similar to the imperative voice in default `{{ site.description }}`

benbalter: My thinking was that it's almost assumed that you're going to change the default description, but that we shouldn't *require* a site publisher to customize the 404 (either due to the imperative voice or to remove something that looks potentially embarrassing).

ashmaroli: Valid point. I'll just remove that line entirely.

parkr:
Neat!

@jekyllbot: merge +minor

PR #5958, opened 2017-03-16
A command to generate a manifest file for the theme-gem currently in use

ashmaroli:
Introducing `manifest-theme`: a Jekyll subcommand to output a theme-manifest that lists the contents of the theme-gem currently in use.

Example manifest:
```
---------------------------------------------------------------------
THEME NAME:      minima

THEME VERSION:   2.1.0

THEME DIRECTORY: [redacted]/gems/minima-2.1.0
---------------------------------------------------------------------

GEM CONTENTS:

  ASSETS:
    * main.scss

  INCLUDES:
    * disqus_comments.html
    * footer.html
    * google-analytics.html
    * head.html
    * header.html
    * icon-github.html
    * icon-github.svg
    * icon-twitter.html
    * icon-twitter.svg

  LAYOUTS:
    * default.html
    * home.html
    * page.html
    * post.html

  SASS:
    * minima/_base.scss
    * minima/_layout.scss
    * minima/_syntax-highlighting.scss
    * minima.scss

  FILES AT ROOT:
    * LICENSE.txt
    * README.md
```

--
/cc @jekyll/ecosystem


pathawks:
@ashmaroli I'm probably not the right guy, since I don't use themes.

It's not obvious to me what the use case for this command would be.

ashmaroli:
Ah! alright then.
Since you mentioned it, this command will be helpful to theme-gem consumers to have a readily accessible means to know whats included and where in the theme-gem, they're currently using.

parkr:
I like:

- showing users what they can use in a gem
- categorization by type, that's nice 👌 

I don't like:

- the fact that it saves to a file. just pipe to stdout and I can write it to a file like `> .jtheme` if I want it there. no writing to files
- `FILES AT ROOT` section seem useless. you really just want to show what the user can use in their site, so there's no point in showing these other files which are useless to that cause.
- the subcommand name of `manifest-theme` is fine, but we have no coherence for our commands and that bugs me.

ashmaroli:
> I don't like:
> 
> the fact that it saves to a file. just pipe to stdout and I can write it to a file like `> .jtheme` if I want it there. no writing to files

Okay, I had it in reverse actually: Write to `jtheme` at all times but additionally pipe to `STDOUT` with the 
`--console` switch.  
I opted for this because, `stdout` is volatile and the user would have to run the command again for every new development session.

Nevertheless, I'll make the changes as requested.

--

> FILES AT ROOT section seem useless. you really just want to show what the user can use in their site, so there's no point in showing these other files which are useless to that cause.

This is included for completion. Its possible that a theme-gem has also bundled a `release-notes.txt` for example, and I feel a `manifest` file should make a note of all contents in the gem.

**Note: If a developer includes a `_data/` or `demo/` or `example/` or any other directory, they will also be included in the manifest-file**

--

> the subcommand name of `manifest-theme` is fine, but we have no coherence for our commands and that bugs me.

I chose `manifest-theme` in line with existing `new-theme` and the proposed `extract-theme` subcommands.

PR #5959, opened 2017-03-17
Document `--unpublished` flag introduced in 91e9ecf

zyv:


zyv:
AppVeyor failure is not my fault ^__^
```
features/site_configuration.feature:160  Scenario: Generate proper dates with explicitly set timezone (same as posts' time) ✔ ✔ ✔ ✔ ✔ ✔ ✔ ✔ ✔ ✔ ✔ ✘ 
  expected "Post Layout: <p>content for entry1.</p>\n built at 2013-04-09T23:22:00-04:00" not to match /Post Layout: <p>content for entry1.<\/p>\n built at 2013-04-09T23:22:00-04:00/m
  Diff:
  @@ -1,2 +1,3 @@
  -/Post Layout: <p>content for entry1.<\/p>\n built at 2013-04-09T23:22:00-04:00/m
  +Post Layout: <p>content for entry1.</p>
  + built at 2013-04-09T23:22:00-04:00
   (RSpec::Expectations::ExpectationNotMetError)
  C:/projects/jekyll/features/step_definitions.rb:240:in `/^I should (not )?see "(.*)" in "(.*)" unless Windows$/'
  features/site_configuration.feature:177:in `And I should see "Post Layout: <p>content for entry1.</p>\n built at 2013-04-09T23:22:00-04:00" in "_site/2013/04/09/entry1.html" unless Windows'
```

ashmaroli:
> AppVeyor failure is not my fault

working on a fix..

zyv:
@parkr anything else I can do?

parkr:
Why did you choose the frontmatter page?

DirtyF:
@zyv What do you say about mention this in https://jekyllrb.com/docs/drafts/ instead ?

zyv:
> Why did you choose the frontmatter page?

My reasoning was as follows:

1. `published` is one of the few pre-defined front matter parameters, and it is introduced on the front matter page; it only makes sense to hint at a directly related option right next to it.
2. It fits nicely into the narrative and allows to mention related draft functionality for posts which will be described later on.
3. I was not able to think of a better place for it.

zyv:
> What do you say about mention this in https://jekyllrb.com/docs/drafts/ instead ?

I don't like this idea for the following reasons:

1. It is my understanding that setting `published` for **posts** is deprecated in favor of newer drafts functionality which arguably is better suited for this task, so I feel that mentioning it on the drafts page would be confusing, unless, instead of adding a ProTip, one expands it into a longer discussion of what's the difference between the two, and which one should be used for which purpose. 
2. I would expect to find a mention of `--unpublished` flag next to the place where `published` front matter variable for pages is introduced in the first place. It wouldn't occur to me, that I should be looking for it on a page describing drafts functionality for posts.

zyv:
Having that said, I'm just trying to clarify my line of thinking to you guys. If you are still unconvinced and would like to have it documented differently, that's certainly fine with me, as long as it gets documented at all. Thank you for considering my patch!

DirtyF:
> a longer discussion of what's the difference between the two, and which one should be used for which purpose.

@zyv AFAIC I use YFM `published: false` to unpublish pages or posts **already** published whether I use drafts **before** I publish a post for instance.

What's your use?

zyv:
@DirtyF yes, that's they way I see it too.

> What's your use?

In our case, we use Jekyll as a CMS, and in addition to what you already described, the `published` flag is used as a drafts system for **pages**, plus to mark 'special' pages like documentation showcasing various components available to the editors. For development builds that go out to the staging system, I set `--drafts --unpublished`, so that editors can preview how pages currently in progress are really going to look like after the go through Jekyll, and for release builds I remove these flags, and the draft pages & posts are dropped from the live version accordingly.

parkr:
@jekyllbot: merge +docs

PR #5960, opened 2017-03-17
Mention Bash on Ubuntu on Windows

jakubklimek:
With Bash on Ubuntu on Windows, jekyll can run as with normal Ubuntu.

### DIFF
--- a/docs/_docs/windows.md
+++ b/docs/_docs/windows.md
@@ -4,8 +4,9 @@ permalink: /docs/windows/
 ---
 
 While Windows is not an officially-supported platform, it can be used to run
-Jekyll with the proper tweaks. This page aims to collect some of the general
-knowledge and lessons that have been unearthed by Windows users.
+Jekyll with the proper tweaks. If you are using Windows 10 Anniversary Update, 
+the easiest way to run Jekyll is to use the new [Bash on Ubuntu on Windows](https://msdn.microsoft.com/en-us/commandline/wsl/about?f=255&MSPPError=-2147217396).
+For older installations, this page aims to collect some of the general knowledge and lessons that have been unearthed by Windows users.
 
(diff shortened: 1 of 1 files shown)


DirtyF:
@jakubklimek good to know. If by any chance you know any step-by-step procedure, we'll glady point to it. I'm pretty sure this would be useful for Windows users who are not necessarily familiar with bash commands.

DirtyF:
Anyone from @jekyll/windows care to review this?

jakubklimek:
@DirtyF IMHO no need for step-by-step procedure. It is
1) Run Bash on Ubuntu on Windows
2) Follow the quick-start instructions
3) If not working = missing prerequisites in Ubuntu => install them

Of course if you think a real step-by-step tutorial including prerequisites installation would be beneficial, I can create it, but it is not there even for regular Ubuntu...

XhmikosR:
Personally, I don't use Bash on Windows, but if the OP has tested this and works it's OK.

crispgm:
Just setup WSL last week. Basically Jekyll runs the same as standard Ubuntu. But I have problems with live reload, will inspect today.

jakubklimek:
@crispgm Do you have W10 Anniversary Update or do you run on the latest Insider Preview, soon to become Creators Update? There were issues with live reload in the AU, which are already resolved in insider preview (at least works for me), so it should definitely work with the upcoming Creators Update

envygeeks:
The insider preview changes you are playing with won't land in stable within a few months, Bash on Windows does not currently have inotify so it should be mentioned to poll.  And Bash on Windows behaves nowhere close to Ubuntu so it should be noted too, that way users don't assume it does.

parkr:
Thank you!

@jekyllbot: merge +docs

PR #5961, opened 2017-03-17
Fix Appveyor with DST-aware cucumber steps

ashmaroli:
Fix recent failures in AppVeyor that occurred due to `ENV["TZ"]` being set on Windows with `tzinfo` gem  with DST-aware cucumber steps.

### DIFF
--- a/features/site_configuration.feature
+++ b/features/site_configuration.feature
@@ -174,6 +174,8 @@ Feature: Site configuration
       Then I should get a zero exit status
       And the _site directory should exist
       And I should see "Page Layout: 2" in "_site/index.html"
+      And I should see "Post Layout: <p>content for entry1.</p>\n built at" in "_site/2013/04/09/entry1.html"
+      And I should see "Post Layout: <p>content for entry2.</p>\n built at" in "_site/2013/04/10/entry2.html"
       And I should see date "2013-04-09T23:22:00-04:00" in "_site/2013/04/09/entry1.html" unless Windows
       And I should see date "2013-04-09T22:22:00-05:00" in "_site/2013/04/09/entry1.html" if on Windows
       And I should see date "2013-04-10T03:14:00-04:00" in "_site/2013/04/10/entry2.html" unless Windows
--- a/features/site_configuration.feature
+++ b/features/site_configuration.feature
@@ -172,12 +172,12 @@ Feature: Site configuration
         | entry2    | 2013-04-10 03:14 -0400 | post    | content for entry2. |
       When I run jekyll build
       Then I should get a zero exit status
-    And the _site directory should exist
+      And the _site directory should exist
       And I should see "Page Layout: 2" in "_site/index.html"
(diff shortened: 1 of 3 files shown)


features/site_configuration.feature
```diff
@@ -172,12 +172,12 @@ Feature: Site configuration
         | entry2    | 2013-04-10 03:14 -0400 | post    | content for entry2. |
       When I run jekyll build
       Then I should get a zero exit status
-    And the _site directory should exist
+      And the _site directory should exist
       And I should see "Page Layout: 2" in "_site/index.html"
-      And I should see "Post Layout: <p>content for entry1.</p>\n built at 2013-04-09T23:22:00-04:00" in "_site/2013/04/09/entry1.html" unless Windows
-      And I should see "Post Layout: <p>content for entry1.</p>\n built at 2013-04-09T22:22:00-05:00" in "_site/2013/04/09/entry1.html" if on Windows
-      And I should see "Post Layout: <p>content for entry2.</p>\n built at 2013-04-10T03:14:00-04:00" in "_site/2013/04/10/entry2.html" unless Windows
-      And I should see "Post Layout: <p>content for entry2.</p>\n built at 2013-04-10T02:14:00-05:00" in "_site/2013/04/10/entry2.html" if on Windows
```
parkr: We also want to see the `Post Layout: <p> content for entry...` here. Feel free to add that assertion above each of these to ensure we don't accidentally cause regressions.

features/step_definitions.rb
```diff
@@ -246,6 +246,30 @@
 
 #
 
+Then(%r!^I should see date "(.*)" in "(.*)" unless Windows$!) do |text, file|
+  step %(the "#{file}" file should exist)
+  regexp = Regexp.new(text)
+  if Jekyll::Utils::Platforms.really_windows? && !dst_active?
+    expect(file_contents(file)).not_to match regexp
```
parkr: Why are you using the negative here?

ashmaroli: Because the regex `2013-04-10T03:14:00-04:00` **will match** `actual value` when:
  - The Platform is Windows and **in an active DST Zone**.
  - The Platform is not Windows.

So this step is:
```ruby
if !Jekyll::Utils::Platforms.really_windows?
  expect(file_contents(file)).to match regexp
elsif Jekyll::Utils::Platforms.really_windows? && dst_active?
  expect(file_contents(file)).to match regexp
else
  expect(file_contents(file)).not_to match regexp
end
```
which could be shortened to ( but may hamper readability ):
```ruby
if (!Jekyll::Utils::Platforms.really_windows?) ||
  (Jekyll::Utils::Platforms.really_windows? && dst_active?)
  expect(file_contents(file)).to match regexp
else
  expect(file_contents(file)).not_to match regexp
end
```
which could be alternatively written in the negative as:
```ruby
if Jekyll::Utils::Platforms.really_windows?) && !dst_active?
  expect(file_contents(file)).not_to match regexp
else
  expect(file_contents(file)).to match regexp
end
```

features/site_configuration.feature
ashmaroli: okay, will do.

parkr:
@jekyllbot: merge +dev

PR #5969, opened 2017-03-21
Release post for v3.4.3

pathawks:
Feel free to make any changes you think necessary.

Our theme does not seem to expect using monospace text as a link:
preview

Should I use `<code class="filter">` for my example Liquid, so that it is green like our filter documentation?

### DIFF
--- a/docs/_posts/2017-03-21-jekyll-3-4-3-released.markdown
+++ b/docs/_posts/2017-03-21-jekyll-3-4-3-released.markdown
@@ -1,0 +1,49 @@
+---
+title: 'Jekyll 3.4.3 Released'
+date: 2017-03-21 08:52:53 -0500
+author: pathawks
+version: 3.4.3
+categories: [release]
+---
+
+Another one-PR patch update as we continue our quest to destroy all bugs. A
+fairly technical debriefing follows, but the TLDR is that we have updated the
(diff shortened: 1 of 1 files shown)


pathawks:
Writing this release post, I have questions about the difference between `cgi_escape` and `uri_escape`, and if it is really _correct_ for `uri_escape` to take some input that is _not_ a URI.

Maybe there is a discussion here for Jekyll 4.0...

parkr:
> Writing this release post, I have questions about the difference between cgi_escape and uri_escape, and if it is really correct for uri_escape to take some input that is not a URI.

If there isn't a divergence in behaviour then we could just alias one to the other. I imagine they are used to escape/encode different pieces of a URL, but I can't remember why. I'm sure conversation in the PR's that created them would describe their differences. Too bad our comments don't reflect that necessarily. 😦

pathawks:
Both filters are quite old.

`cgi_escape` was added in 69683467031b3a34c64fc160ff6ea8fd18aa0466
`uri_escape` was added in #234

Only difference is encoding spaces:

```liquid
{{ "Hello World" | cgi_escape }} => "Hello+World"
{{ "Hello World" | uri_escape }} => "Hello%20World"
```

We should make a note of this.

PR #5975, opened 2017-03-23
Re-surface missing public methods in `Jekyll::Document`

ashmaroli:
Numerous `public` methods for an instance of `Jekyll::Document` got removed (converted to `private` methods) in [`158e026`](https://github.com/jekyll/jekyll/pull/5045/commits/158e02623a4b0207878dfd01fda7a9bd666937fc) which should've been put off till `v4.0`

This came to my attention via [this thread](https://talk.jekyllrb.com/t/serve-error-undefined-method-post-read-for-class-jekyll-document/353/3) at Jekyll Talk, which reported a broken plugin `jekyll-multiple-languages`

Now that we have a system to easily release backport-hotfixes, this should probably be released as `v3.3.2` (IMO).

--
/cc @ayastreb, @parkr

### DIFF
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -3,9 +3,12 @@
 module Jekyll
   class Document
     include Comparable
+    extend Forwardable
 
     attr_reader :path, :site, :extname, :collection
     attr_accessor :content, :output
+
+    def_delegator :self, :read_post_data, :post_read
 
     YAML_FRONT_MATTER_REGEXP = %r!\A(---\s*\n.*?\n?)^((---|\.\.\.)\s*$\n?)!m
     DATELESS_FILENAME_MATCHER = %r!^(?:.+/)*(.*)(\.[^.]+)$!
@@ -255,7 +258,7 @@ module Jekyll
         begin
           merge_defaults
           read_content(opts)
-          post_read
+          read_post_data
         rescue SyntaxError => e
           Jekyll.logger.error "Error:", "YAML Exception reading #{path}: #{e.message}"
         rescue => e
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -255,7 +255,7 @@ module Jekyll
         begin
           merge_defaults
           read_content(opts)
-          read_post_data
+          post_read
         rescue SyntaxError => e
           Jekyll.logger.error "Error:", "YAML Exception reading #{path}: #{e.message}"
         rescue => e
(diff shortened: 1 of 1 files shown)


lib/jekyll/document.rb
```diff
@@ -265,6 +265,42 @@ def read(opts = {})
...
+        .split(File::SEPARATOR)
+        .reject do |c|
+        c.empty? || c == special_dir || c == basename
+      end
+      merge_data!({ "categories" => superdirs }, :source => "file path")
+    end
+
+    def post_read
+      read_post_data
+    end
```
parkr: This should use `alias_method`.

lib/jekyll/document.rb
```diff
@@ -255,7 +255,7 @@ def read(opts = {})
         begin
           merge_defaults
           read_content(opts)
-          read_post_data
+          post_read
```
parkr: This should remain unchanged as one is an alias to another.

lib/jekyll/document.rb
```diff
@@ -265,6 +265,42 @@ def read(opts = {})
...
+        )
+        ).map(&:to_s).flatten.uniq,
+      })
+    end
+
+    def populate_tags
+      merge_data!({
+        "tags" => Utils.pluralized_array_from_hash(data, "tag", "tags").flatten,
+      })
+    end
```
parkr: Can these be marked as `public` without moving them around?

ashmaroli: Since all methods after the first declaration of `private` are marked as `private_methods` internally, these methods need to be 'moved up` to before the declaration, at least in this case, right after `:respond_to_missing?`.

lib/jekyll/document.rb
ashmaroli: @parkr since `:read_post_data` is a `private_method`, what would be the best way to alias to `post_read`? `def_delegator`?

pathawks:
Any way we could add a note that these methods should be made `private` in 4.0?

ashmaroli:
> Any way we could add a note that these methods should be made private in 4.0?

@pathawks Think the following will do for now. (Even though it'll be included in the auto-generated documentation):
```
# DEPRECATION: This method will be made 'private' in the next major release.
```
Then leave a `TODO:` comment to our `Jekyll 4.0 Wishlist` as a reminder for future. As we start developing `4.0`, we should have `Jekyll::Deprecator` warn/inform the user about its removal in the near future.

parkr:
@jekyllbot: merge +bug

PR #5977, opened 2017-03-23
Fix alignment inconsistencies in features

ashmaroli:
Found a few alignment inconsistencies in various cucumber features.
This PR attempts to address those inconsistencies.


parkr:
I appreciate the eye for detail but I'd rather not blurry any git history by having substance-less whitespace changes in the logs. As people change these tests, we can ask them to update the indentation. Thanks, though!

Issue #4381, opened 2016-01-20
Remove support for Ruby 2.0.x


parkr:
[Ruby 2.0.x is presently in maintenance mode](https://www.ruby-lang.org/en/news/2015/02/25/ruby-2-0-0-p643-is-released/). Support for it will end on February 24, 2016. Once it's EOL'd, we should stop supporting it. I'm thinking Jekyll v3.2.

We should document this on the website under "Requirements" and enforce in the gemspec.

/cc @jekyll/core

mattr-:
Is removing support for a ruby version a backwards incompatible change according to SemVer? I know one of the reasons cited for Rails going to 5.0 instead of 4.3 was because they wanted to remove compatibility with older Ruby versions.

Since we can enforce this in the gemspec and `rubygems` and `bundler`, it's probably a moot point, but I figured I'd ask.

parkr:
> Is removing support for a ruby version a backwards incompatible change according to SemVer?

That was a concern of mine. I don't think the spec says anything about it. I feel like dropping 1.9 support was a bigger thing than dropping 2.0, because Ruby 1.9 and Ruby 2.0 weren't necessarily compatible (thus making it difficult to upgrade) but 2.0 and 2.1 _are_ supposed to be compatible so the upgrade path should be easier for users.

Notably, this breaks support for native installation on Mac, which, I believe, runs 2.0.

mattr-:
It does break native installation on Mac. I'm on El Capitan and still only have 2.0.0p645. :neutral_face:

envygeeks:
The only notable incompability (which isn't even listed on the Ruby site) is keywords.  We can enforce them in 2.0... we can in 2.1. This is what prompted me to get angry at 2.0.... because when building Pathutil to encapsulate safety and speed into Pathname for Jekyll and Docker-Template and my apps, I couldn't enforce keywords, I had to loosen my restrictions and add some logic I wouldn't normally add.

I think the drop should only be superficial until 4.0 though, don't test for it, don't care about it, but if there is a problem take it on a case-by-case until 4.0 and announce the change as official.

mattr-:
>I think the drop should only be superficial until 4.0 though, don't test for it, don't care about it, but if there is a problem take it on a case-by-case until 4.0 and announce the change as official.

:+1: from me on this.

envygeeks:
:tada:

PR #5983, opened 2017-03-26
Require Ruby > 2.1.0

DirtyF:
Support for Ruby 2.0 has been dropped since https://github.com/jekyll/jekyll/issues/4381

This PR aims to state the current Ruby depency in the docs and on Rubygems.

~~Some core plugins could be impacted, for example[ jekyll/import still requires Ruby 1.9.x ](https://github.com/jekyll/jekyll-import/blob/master/jekyll-import.gemspec#L10
)~~~

/cc @jekyll/plugin-core

### DIFF
--- a/docs/_docs/troubleshooting.md
+++ b/docs/_docs/troubleshooting.md
@@ -17,7 +17,7 @@ that might be of help. If the problem you’re experiencing isn’t covered belo
 ## Installation Problems
 
 If you encounter errors during gem installation, you may need to install
-the header files for compiling extension modules for Ruby 2.0.0. This
+the header files for compiling extension modules for Ruby 2.x This
 can be done on Ubuntu or Debian by running:
 
 ```sh
--- a/docs/_docs/upgrading/2-to-3.md
+++ b/docs/_docs/upgrading/2-to-3.md
@@ -12,7 +12,7 @@ Before we dive in, go ahead and fetch the latest version of Jekyll:
 $ gem update jekyll
 ```
 
-Please note: Jekyll 3 requires Ruby version >= 2.0.0.
+Please note: Jekyll 3.2 requires Ruby version >= 2.1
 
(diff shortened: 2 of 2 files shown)


jekyll.gemspec
```diff
@@ -7,7 +7,7 @@ Gem::Specification.new do |s|
   s.specification_version = 2 if s.respond_to? :specification_version=
   s.required_rubygems_version = Gem::Requirement.new(">= 0") if s.respond_to? :required_rubygems_version=
   s.rubygems_version = "2.2.2"
-  s.required_ruby_version = ">= 2.0.0"
+  s.required_ruby_version = ">= 2.1.0"
```
parkr: I'm fine with the documentation updates, but I'm 👎 on this change before Jekyll 4. Our policy on this has been:

- Each major Jekyll version cannot remove support for old Ruby verisons by requiring newer ones in the gemspec. This is considered a breaking change.
- Each minor Jekyll version can deprecate support for an old Ruby version by removing that Ruby version from CI and stating that support has been deprecated in the release blog post.

DirtyF: OK so Jekyll 4 can bump Ruby dependencies from 2.0 to 2.4 for instance, but Jekyll 3.5 cannot bump from 2.0 to 2.1, even if [2.0 is not tested anymore on CI](https://github.com/jekyll/jekyll/blob/master/.travis.yml#L8-L10).

PR updated accordingly.

parkr:
Yeah, updating the `required_ruby_version` in the gemspec is a big deal and causes tons of headaches for people. Ruby 2.0 and 2.1 are basically compatible so it makes way more sense to me to just silently drop full support but still allow people to live dangerously and run Ruby 2.0 if they want.

DirtyF:
@jekyllbot: merge +docs

Issue #5771, opened 2017-01-16
Move plugins listing to new plugins directory


pathawks:
Right now, we have [a list of community contributed plugins in the documentation](https://jekyllrb.com/docs/plugins/#available-plugins).

It would be cool if we could remove this list from the docs, and migrate this functionality over to the [plugins directory](https://github.com/jekyll/plugins).

That directory still is not totally polished, so I'd like to really define a path forward. What would it take to move everything over to that new site?

(On my personal wishlist: I would love to also list themes there, or at least create a comparable directory for themes)

parkr:
> That directory still is not totally polished, so I'd like to really define a path forward. What would it take to move everything over to that new site?

I kind of did this before to make a scaffold for the original site. I grabbed all the links in the plugins listing, and put the metadata where it needed to go (title, url, type of plugin, etc). I think we should migrate the data.

I think a key piece still missing is search. We should add a search capability to the plugins website – right now folks can do Cmd-F and find what they need, but an equivalent functionality isn't yet there AFAIK.

> I would love to also list themes there, or at least create a comparable directory for themes

Now that we have gem-based themes, I'd love to see this as well. Perhaps it can be our "marketplace" or "directory" for all plugins/themes made for Jekyll.

DirtyF:
As far as search is concerned, I'd love users to benefit from Algolia search user experience. If you are ok, I can submit a PR for this.

parkr:
Please do this soon! Algolia, Lunr.js, etc are all fine with me. Key is that they work and that they can do fuzzy finding on multiple input fields (i.e. not just name, but also things like URL and description).

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>3.3-stable</code> or <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, please consider building it first as a plugin. Jekyll 3 introduced [hooks](http://jekyllrb.com/docs/plugins/#hooks) which provide convenient access points throughout the Jekyll build pipeline whereby most needs can be fulfilled. If this is something that cannot be built as a plugin, then please provide more information about why in order to keep this issue open.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

jekyllbot:
[...]

PR #5995, opened 2017-03-30
Default options for script/proof

DirtyF:
- As http://jekyllrb.com/tutorials/navigation/ uses `<a href="#">` links, `allow-hash-href` is now set to true by default
- Set `disable-external-url` by default to avoid checking more than 2500 external URLs **everytime** 
- remove the lasting 404 errors

```shell
$ script/proof

Running ["ScriptCheck", "ImageCheck", "LinkCheck"] on ["./_site"] on *.html...
Ran on 122 files!
HTML-Proofer finished successfully.
```

/cc @jekyll/documentation

### DIFF
--- a/script/proof
+++ b/script/proof
@@ -9,7 +9,7 @@ function msg {
   printf "\e[0;37m==> $1\e[0m\n"
 }
 
-INGORE_HREFS=$(ruby -e 'puts %w{
+IGNORE_HREFS=$(ruby -e 'puts %w{
   Chrononaut
   twitter.com
   nearlyfreespeech.net
@@ -32,4 +32,4 @@ bundle exec jekyll build -s $SOURCE -d $DESTINATION --trace
 
 # 3.
 msg "Proofing..."
-time bundle exec htmlproofer ./$DESTINATION --url-ignore $INGORE_HREFS $@
+time bundle exec htmlproofer ./$DESTINATION --allow-hash-href --url-ignore $IGNORE_HREFS $@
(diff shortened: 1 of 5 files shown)


pathawks:
Could @jekyllbot periodically check for 404s?

script/proof
```diff
@@ -32,4 +32,4 @@ bundle exec jekyll build -s $SOURCE -d $DESTINATION --trace
 
 # 3.
 msg "Proofing..."
-time bundle exec htmlproofer ./$DESTINATION --url-ignore $INGORE_HREFS $@
+time bundle exec htmlproofer ./$DESTINATION --allow-hash-href --disable-external --url-ignore $IGNORE_HREFS $@
```
parkr: I'm not sure about `--disable-external` – I'd like to know when we're pointing to resources that are no longer there. What is failing? Can we periodically nix them?

DirtyF: Right, let's deal with this in another PR, the primary goal was to fix the current errors.
It should be less painful to run the script when #5771 will be done anyway.

DirtyF:
@pathawks you mean for example `+docs` would trigger `script/proof`before merging and @jekyllbot would post a comment if a 404 is present?

pathawks:
@dirtyf That would indeed be cool, but I was thinking more about just periodically (daily?) making sure everything we link to still exists. I imagine link rot is the problem we are trying to target with this, right?

It's important to make sure the external pages we link to still exist, but it's frustrating when that causes an unrelated PR to fail.

script/proof
parkr: `script/proof` is not like our normal CI that needs to pass no matter what, so it's ok to leave it half broken. I'd say remove `--disable-external` from this PR.

DirtyF:
@jekyllboy: merge +docs

script/proof
DirtyF: done

PR #5996, opened 2017-03-31
Filters#time helper: Duplicate time before calling #localtime.

parkr:
Failure noticed on `master`: https://travis-ci.org/jekyll/jekyll/jobs/217231764

In Ruby 2.4, Time#localtime modifies the Time instance.
In Ruby 2.3, Time#localtime appears to leave them alone.

Either that, or `#to_time` used to duplicate and now it doesn't.


parkr:
@jekyllbot: merge +fix

PR #5999, opened 2017-04-02
Updated upgrading.md

moizsajid:


DirtyF:
Thanks. Why would you rather not use bundler to upgrade?

We invite users to use `bundler`on purpose. Our docs don't necessarily reflect that and commands are not always prefixed with `bundle`.

moizsajid:
Got it. Thank you!

PR #6000, opened 2017-04-02
Updated installation.md

moizsajid:


DirtyF:
Thanks for the PR, I would tend to keep the double dash syntax `--version`, if you wanna make it shorter, use `- v`

PR #6002, opened 2017-04-03
Fix missing quotation mark

admhlt:
I was reading the docs tonight and noticed this quotation mark was missing, which also threw off the highlighting in this block.

### DIFF
--- a/docs/_docs/configuration.md
+++ b/docs/_docs/configuration.md
@@ -603,7 +603,7 @@ collections:
 # Handling Reading
 safe:         false
 include:      [".htaccess"]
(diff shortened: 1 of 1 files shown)


ashmaroli:
@admhlt Thanks for fixing the error (which was my fault.)

DirtyF:
Thanks @admhlt  

@jekyllbot merge +docs

@ashmaroli Don't beat yourself, If reviewers don't catch it, they're equally responsible 😕

DirtyF:
@jekyllbot: merge +docs

PR #6008, opened 2017-04-05
Update normalize.css to v6.0.0

chrisfinazzo:


DirtyF:
@pathawks  We don't use style:compressed option right now, I don't know why, I just tested and website renders fine. I suggest we deal with this in a dedicated PR 😉

DirtyF:
Thanks @chrisfinazzo ⛑  ⚙️  📏  

@jekyllbot: merge +site

PR #6009, opened 2017-04-05
Compress Website Sass output

DirtyF:
Suggested by @pathawks in #6008

### DIFF
--- a/rake/site.rake
+++ b/rake/site.rake
@@ -43,12 +43,10 @@ namespace :site do
   end
   task :build => :generate
 
-  desc "Update normalize.css library to the latest version and minify"
+  desc "Update normalize.css library to the latest version"
   task :update_normalize_css do
     Dir.chdir("#{docs_folder}/_sass") do
-      sh 'curl "https://necolas.github.io/normalize.css/latest/normalize.css" -o "normalize.scss"'
-      sh 'sass "normalize.scss":"_normalize.scss"'
-      rm ["normalize.scss", Dir.glob("*.map")].flatten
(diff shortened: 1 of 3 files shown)


rake/site.rake
```diff
@@ -47,7 +47,7 @@ namespace :site do
   task :update_normalize_css do
     Dir.chdir("#{docs_folder}/_sass") do
       sh 'curl "https://necolas.github.io/normalize.css/latest/normalize.css" -o "normalize.scss"'
-      sh 'sass "normalize.scss":"_normalize.scss" --style compressed'
+      sh 'sass "normalize.scss":"_normalize.scss"'
```
pathawks: What is the purpose of this?

Why not:

```sh
sh 'curl "https://necolas.github.io/normalize.css/latest/normalize.css" -o "_normalize.scss"'
```

DirtyF: It seems that previously we used a Sass workflow, and now we rely entirely on GitHub Pages pipeline. We therefore can simplify here to the bare minimum, you're right. 

Suggestion applied. 🖖

pathawks:
…as long as there isn't some reason I'm not thinking of to _not_ do this

parkr:
This is automatically done in production - are you seeing different results on jekyllrb.com?

DirtyF:
@parkr https://jekyllrb.com/css/screen.css does not look compressed to me

parkr:
You're absolutely right! It defaults to `:compact`, which isn't the same.

parkr:
@jekyllbot: merge +docs

PR #6006, opened 2017-04-05
Fixup tutorial on creating theme from existing HTML templates

ashmaroli:
Found a few errors in the latest tutorial some of which, we missed while reviewing, the others became evident on reading the generated output.

/cc @jekyll/documentation @tomjohnson1492

PR #6010, opened 2017-04-06
(Another) tutorial to convert existing site to Jekyll

ashmaroli:
Hello guys
While I was working on #6006 I realized I wanted to make a lot more changes to the tutorial to make it seem like a tutorial to convert an existing site to Jekyll.

So I moved all those changes to a new branch.

I totally understand if this doesn't get merged, but I request a read-through when possible and leave your feedback in the comments.

Thanks

/cc @DirtyF @tomjohnson1492

### DIFF
--- a/docs/_data/tutorials.yml
+++ b/docs/_data/tutorials.yml
@@ -5,6 +5,7 @@
   - orderofinterpretation
   - custom-404-page
   - create-jekyll-theme
+  - convert-existing-site-to-jekyll
 
 #- title: Another section
 #  tutorials:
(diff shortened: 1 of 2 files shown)


tomjohnson1492:
This is interesting. I like your approach in "reverse engineering" a site and walking through a specific example. It's a little odd to start jekyllizing a site that is already a jekyll site, though. But that's okay I guess. Maybe call out that you're treating the site as a static HTML output, similar to an HTML template one might want to convert into a Jekyll site.

With site code samples like this, will they be stored in the Jekyll project or just referenced in their own repos from contributors?

Overall, I like your approach. Could we publish *both* tutorials? One long and general, the other more specific and focused on a specific example. I think it's okay to include both, even with overlap like this. You really dig into post and permalink structures in ways the original didn't. I would lop off the last part about layouts, though. That seemed appended on as an afterthought.

ashmaroli:
Thanks for the feedback @tomjohnson1492 :smiley:

ashmaroli:
@DirtyF If you have the time please go through this one and let me know if this is worth a merge.. or close..
Thanks

PR #6013, opened 2017-04-08
(Minor) Update MathJax CDN

brentyi:
Hi there!

The MathJax CDN is [shutting down at the end of the month](https://www.mathjax.org/cdn-shutting-down/) -- I just updated the documentation to suggest using the cdnjs one instead.

/cc @jekyll/documentation


DirtyF:
@jekyllbot: merge +docs

PR #6014, opened 2017-04-08
Add a blacklist to restrict gems when not in safe mode

ashmaroli:
## Why?
 users load plugins via
   - `:jekyll_plugins` in `Gemfile` (or)
   - `gems` array in `_config.yml` when they don't have a `Gemfile` (or)
   - both of the above

and since recently, `runtime_dependencies` of a `theme` will be automatically loaded by Jekyll. Restricting a plugin requires *adding every other plugin* to the `whitelist` and building the site with `--safe`.

So it would be better if there exists a single point of control to not allow a plugin when the site is not built with `--safe` switch: **simply add the concerned plugin to a `blacklist`.**


pathawks:
Can you explain a bit more why a user would want to blacklist a required plugin?

ashmaroli:
> Can you explain a bit more why a user would want to blacklist a required plugin?

- to use another alternative. Its possible that another plugin exists that provides similar functionality but with additional features.
- to disable a plugin with one setting. I see that jekyll-admin#74 is contemplating this feature

parkr:
Having a blacklist and a whitelist is going to be very confusing. If a theme wants a plugin and we require it but the user blacklists it, that makes for a pretty crappy user experience. 💥

DirtyF:
That does not look like a 90% use case to me. Let's not forget https://jekyllrb.com/philosophy 😇

ashmaroli:
> If a theme wants a plugin and we require it but the user blacklists it, that makes for a pretty crappy user experience.

@parkr When a user **blacklists** a required plugin, the onus is on the user to shoulder consequences. `blacklist` is not even present in the `DEFAULT Configuration`.

> That does not look like a 90% use case to me.

@DirtyF True. But if you guys feel this doesn't belong in the Core, can you please create and attach new label named **`plugin-case`** to this before closing? That should be a cue to the minority to look for a plugin that provides this functionality, and to the author to proceed accordingly. :smiley:

pathawks:
> can you please create and attach new label named plugin-case

That seems presumptuous. It seems `wont-fix` is more general.

parkr:
We went down this road with `include`/`exclude` for files and it's a bit of a mess since one overrides the other but some users want that overriding to work in one direction and some want it to work in the other direction. It's a tough spot to be in.

A simple whitelist is what we'd like to encourage. We want plugin writers to think: "When my plugin is added, what should the default behaviour be? Should it automatically kick in, or should I ask the user to configure something first?" Additionally, theme writers should think carefully before plugging tons of plugins into their theme: does it add value to the core of my theme, or is it something I'd rather document to allow users to do themselves?

At any rate, a blacklist and a whitelist is confusing (especially to new users) so I think we should try to solve the underlying problems as to why the whitelist doesn't work.

Thanks!

ashmaroli:
Okay.. no probs.. I'll release this as a plugin if there's enough demand for it.. Thank you for the inputs everyone :smiley:

PR #6024, opened 2017-04-12
Allow index.json to be served from /myuri/

cameronmcefee:
Disclaimer: I have no idea what I'm doing. I don't know how to test this or even if this is the right idea, but I figured a little action was better than no action. I won't be hurt if you nuke this.

From time to time I like to include static APIs in my jekyll sites that for logistical reasons, have to mirror production API url structure. For example `site.com/api/myapi/` would return json data. However, it's not currently possible to do something like:

api/myapi.json:

```
---
permalink: /api/myapi/
---

{
...
}
```

This will create /api/myapi/index.json, which won't get served from /api/myapi/

The desired result can be achieved by giving the file a .html extension, but that just leaves me feeling dirty.

If this PR does actually do what I want, hell yeah, look ma I write software. If not, maybe I can perturb someone enough to do it the right way 😄

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -88,6 +88,7 @@ module Jekyll
               index.rhtml
               index.cgi
               index.xml
+              index.json
             ),
           }
 


benbalter:
@cameronmcefee See http://ben.balter.com/2012/12/27/introducing-jekyllbot/ which may be what you're looking for. The code's old, but may be relevant. Ideally, I'd love to break the API components out of https://github.com/jekyll/jekyll-admin to make a static API exactly as you describe, but in a bit more robust way.

parkr:
Hey @cameronmcefee! Thanks for the pull.

We have done our best to match GitHub Pages's serving logic here, so ideally we'd have a very strong reason for divergence. As a user browsing the web, I would be very surprised if I went to a webpage that looks like it would serve me HTML and I got JSON.

One thing I'd be down to see is this: if the client sends a `Accept: application/json` header, then find the `index` file with the proper extension and serve that up. So most browsing would get `index.html` but if you sent that `Accept` header, you'd get `index.json`.

cameronmcefee:
> One thing I'd be down to see is this: if the client sends a Accept: application/json header, then find the index file with the proper extension and serve that up. So most browsing would get index.html but if you sent that Accept header, you'd get index.json.

I think this is more what I'm after (note the "I don't know what I'm doing" disclaimer). It would just be nice to be able to not be forced to only use .json in the url.

I'll close this for now since it looks like the path forward is something I don't know how to implement myself, but I consider the awareness raised if anyone else finds it valuable to explore.

PR #6026, opened 2017-04-13
fix typo using past participle

jpuck:


DirtyF:
Thanks @jpuck ❤️ 

@jekyllbot: merge +docs

PR #6028, opened 2017-04-14
doc fix typo word usage

jpuck:


DirtyF:
Good catch!

@jekyllbot: merge +docs

PR #6031, opened 2017-04-17
doc use example.com

jpuck:
unless you're explicitly trying to advertise for domain.com
then you should use the official example domain setup by IETF

https://tools.ietf.org/html/rfc2606
https://tools.ietf.org/html/rfc6761


DirtyF:
@jekyllbot: merge +dev

PR #6032, opened 2017-04-17
corrected reference to layout in index.md

meyer-sven:
The file index.md referenced the layout "default.html" in its frontmatter. As far as I know and according to my tests, a layout has to be referenced without its file extension.

This is my first proposal for a file change in github - sorry, if I did something wrong. Your tutorial was very helpful to me for starting my first Jekyll-blog.


DirtyF:
@jekyllbot: merge +docs

PR #6035, opened 2017-04-18
Add MvvmCross to samples

MarcBruins:
We have recently build a Jekyll website for the MvvmCross open source project(https://github.com/MvvmCross/MvvmCross) which we are really proud of. It will make maintaining documentation a lot easier. I've put it in the samples area on the website in the hope that it can inspire other open source projects.


pnn:
@jekyllbot: merge +docs

Issue #6038, opened 2017-04-19
Since 3.2, where-filter incorrectly handles missing/null properties


chuckhoupt:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I updated to the latest Jekyll
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [ ] This is a feature request.

---

- [x] I am on (or have tested on) ***macOS*** 10+

---

- [x] I was trying to build.

## Description

Before Jekyll 3.2, the `where` filter had the useful ability to select objects based on a missing or empty keys (`where: "key", nil`). The new array-support feature (pull #4555) changed this behavior causing problems for sites upgrading from 3.1.

Pull request fix and test: #6039

## My Reproduction Steps

0. Using Jekyll 3.2+, create a test site: `jekyll new testsite`
0.  In testsite, add `test.md`:
    ```
    ---
    objects: [{}, {a: null}, {a: ""}, {a: "text"}]
    ---
    
    Count of Empty As: {{ page.objects | where: "a", nil | size }}
    ```
0. Run `jekyll build` and examine `_site/test.html`
    
    Three of the objects have empty A properties, but Jekyll 3.2+ reports only one:
    
    ```
    Count of Empty As: 1
    ```

## The Output I Wanted

With versions before Jekyll 3.2, the output is:

```
Count of Empty As: 3
```

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>3.3-stable</code> or <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, please consider building it first as a plugin. Jekyll 3 introduced [hooks](http://jekyllrb.com/docs/plugins/#hooks) which provide convenient access points throughout the Jekyll build pipeline whereby most needs can be fulfilled. If this is something that cannot be built as a plugin, then please provide more information about why in order to keep this issue open.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

PR #6039, opened 2017-04-19
Restore where-filter null handling to match earlier versions (<3.2)

chuckhoupt:
Summary of full bug report #6038:

> Before Jekyll 3.2, the `where` filter had the useful ability to select objects based on a missing or empty keys (`where: "key", nil`). The new array-support feature (pull #4555) changed this behavior causing problems for sites upgrading from 3.1.

This pull request tests and fixes the bug introduced in pull #4555. The core problem was that `Array(nil)` returns `[]`, rather then `[nil]`, so the fix uses an immediate array constructor to preserve nils.

### DIFF
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -221,7 +221,9 @@ module Jekyll
       return input unless input.respond_to?(:select)
       input = input.values if input.is_a?(Hash)
       input.select do |object|
-        Array(item_property(object, property)).map(&:to_s).include?(value.to_s)
+        item_value = item_property(object, property)
+        item_value = [item_value] unless item_value.respond_to?("map")
+        item_value.map(&:to_s).include?(value.to_s)
       end || []
     end
 
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -222,7 +222,7 @@ module Jekyll
       input = input.values if input.is_a?(Hash)
       input.select do |object|
         item_value = item_property(object, property)
-        item_value = [item_value] unless item_value.respond_to?("map")
+        item_value = [item_value] unless item_value.respond_to?(:map)
         item_value.map(&:to_s).include?(value.to_s)
       end || []
     end
--- a/test/test_filters.rb
+++ b/test/test_filters.rb
@@ -655,6 +655,17 @@ class TestFilters < JekyllUnitTest
         assert_equal 2, @filter.where(@array_of_objects, "color", "red").length
       end
 
+      should "filter properties with nil values and undefined properties appropriately" do
+        objects = [
+          {},
+          {"a" => nil},
+          {"a" => ""}
+        ]
+        assert_equal 3, @filter.where(objects, "a", nil).length
+        assert_equal 3, @filter.where(objects, "a", "").length
+        assert_equal 0, @filter.where(objects, "b", "something").length
+      end
+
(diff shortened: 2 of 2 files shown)


pathawks:
> fixes the bug 

Just so we're clear, was this previous behavior documented, or is this just a case where the undefined behavior changed?

Will this PR break somebody's site that depends on the _new_ behavior?

chuckhoupt:
The previous behavior was undocumented/untested, so it is a change to undefined behavior. It will break sites that depend on the new behavior (often by subtly changing output). So I see that my argument for restoring the old behavior is a weak appeal to "tradition", and maybe it is too late to restore it after 10 months.

Possibly a more conservative alternative would be to simply make the undefined areas illegal, so an error or warning is thrown? That would give a clear signal that the old behavior has changed and new sites shouldn't rely on it.

I only noticed the change by running a diff of a site while upgrading from 3.1 to 3.4 -- it could easily have been missed. A warning would have helped.

test/test_filters.rb
```diff
@@ -655,6 +655,17 @@ def to_liquid
...
       end
 
+      should "filter properties with nil values and undefined properties appropriately" do
+        objects = [
+          {},
+          { "a" => nil },
+          { "a" => "" },
+        ]
+        assert_equal 3, @filter.where(objects, "a", nil).length
+        assert_equal 3, @filter.where(objects, "a", "").length
```
parkr: ??

lib/jekyll/filters.rb
```diff
@@ -221,7 +221,9 @@ def where(input, property, value)
       return input unless input.respond_to?(:select)
       input = input.values if input.is_a?(Hash)
       input.select do |object|
-        Array(item_property(object, property)).map(&:to_s).include?(value.to_s)
+        item_value = item_property(object, property)
+        item_value = [item_value] unless item_value.respond_to?("map")
```
parkr: This should be `:map` instead of `"map"`.

lib/jekyll/filters.rb
```diff
@@ -221,7 +221,9 @@ def where(input, property, value)
...
+        item_value.map(&:to_s).include?(value.to_s)
```
parkr: Could we special-case `value`? Then we could do

```ruby
if value.nil?
  item_value.any?(&:nil?)
else
  item_value.map(&:to_s).include?(value.to_s)
end
```

lib/jekyll/filters.rb
chuckhoupt: Fixed.

test/test_filters.rb
chuckhoupt: I improved the test description.

lib/jekyll/filters.rb
chuckhoupt: The problem with special-casing nil is that it would break the existing behavior. `where` has always treated nil and empty-string (as well as nonexistent properties) as identical because they all stringify to the empty-string. The array-support added by #4555 continued this tradition.

I've added some more tests to make the current handling of nil/empty-string in list properties explicit. A special case for nil will break these tests.

chuckhoupt:
I realize that `where`'s identical handling of nil and empty-string are a bit odd, but I think changing it may break existing sites.

pathawks:
How do we want to approach this in `4.0`?

PR #6040, opened 2017-04-19
fix typo

danielyang1009:
should be `{% post_url %}` not `{{ post url }}`

PR #6044, opened 2017-04-22
Fix CI 

DirtyF:
Fix the corresponding test after #6040


DirtyF:
@jekyllbot: merge +dev

PR #6045, opened 2017-04-22
site template: Use plugins key instead of gems 

DirtyF:
Since #5130 we are encouraged to use `plugins` key rather than `gems` in `config.yml`

This should be merged only when GitHub Pages has updated to next release.

### DIFF
--- a/lib/site_template/_config.yml
+++ b/lib/site_template/_config.yml
@@ -27,7 +27,7 @@ github_username:  jekyll
 # Build settings
 markdown: kramdown
 theme: minima
-gems:
+plugins:
   - jekyll-feed
 
 # Exclude from processing.


ashmaroli:
Shouldn't this be added to Milestone `3.5.1` instead?

DirtyF:
Waiting for https://github.com/github/pages-gem/pull/449

parkr:
If you separated this into two PR's – one for the example and one for the docs site – then we could merge one and wait on the other, rather than holding both up due to the latter's dependency on Pages. Should I merge #6169 and this can be just for the docs site?

DirtyF:
@parker OK, you can merge this one, I'll open another PR for Jekyll's website.

benbalter:
@jekyllbot: merge +minor.

benbalter:
🎉 💥 🌮 @DirtyF!

PR #6046, opened 2017-04-22
allow check if site has a specific plugin enabled

### LINKED ISSUE #6020: Ability to conditionally call a plugin if it is enabled
Via https://github.com/jekyll/minima/pull/109, it's not currently possible to conditionally call a plugin tag if it exists (without reverting to a weired include work around). Or if it is possible, it's not currently documented.

Example:

```liquid
{% if site.plugins contains "jekyll-seo-tag" %}
    {% seo %}
{% else %}
    <title>{{ page.title }}</title>
{% endif %}
```

With the above, or something similar, themes could call a plugin, conditioned on it being installed, without breaking the build with an opaque "tag not found" error.

Crunch09:
A plugin is considered as enabled when it is listed in the `plugins`
config and in the `whitelist` config in case the site is build in
safe mode.

Example:
```
{% if site.plugins contains "jekyll-seo-tag" %}
    {% seo %}
{% else %}
    <title>{{ page.title }}</title>
{% endif %}
```

This fixes #6020.

### DIFF
--- a/features/plugins.feature
+++ b/features/plugins.feature
@@ -35,3 +35,22 @@ Feature: Configuring and using plugins
     And I should see "Whatever" in "_site/index.html"
     And the "_site/test.txt" file should exist
     And I should see "this is a test" in "_site/test.txt"
+
+  Scenario: Plugin is enabled
+    Given I have an "index.html" page that contains "{% if site.plugins contains "jekyll_test_plugin" %} plugin loaded {% endif %}"
+    And I have a configuration file with "gems" set to "[jekyll_test_plugin]"
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "plugin loaded" in "_site/index.html"
+
+  Scenario: Plugin is not enabled
+    Given I have an "index.html" page that contains "{% if site.plugins contains "jekyll_test_plugin" %} plugin loaded {% endif %}"
+    And I have a configuration file with:
+    | key       | value                |
+    | gems      | [jekyll_test_plugin] |
+    | whitelist | []                   |
+    When I run jekyll build --safe
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should not see "plugin loaded" in "_site/index.html"
--- a/lib/jekyll/drops/site_drop.rb
+++ b/lib/jekyll/drops/site_drop.rb
@@ -37,6 +37,10 @@ module Jekyll
         @site_collections ||= @obj.collections.values.sort_by(&:label).map(&:to_liquid)
       end
 
+      def plugins
+        @obj.plugin_manager.enabled_plugins
+      end
+
       private
       def_delegator :@obj, :config, :fallback_data
     end
--- a/lib/jekyll/plugin_manager.rb
+++ b/lib/jekyll/plugin_manager.rb
@@ -25,9 +25,7 @@ module Jekyll
     #
     # Returns nothing.
     def require_gems
-      Jekyll::External.require_with_graceful_fail(
-        site.gems.select { |plugin| plugin_allowed?(plugin) }
-      )
+      Jekyll::External.require_with_graceful_fail(enabled_plugins)
     end
 
     # Require each of the runtime_dependencies specified by the theme's gemspec.
@@ -55,6 +53,18 @@ module Jekyll
       else
         false
       end
+    end
+
+    # Build an array of enabled plugin names.
+    #
+    # Returns an array of strings, each string being the name of a plugin
+    #   that is specified (and whitelisted in case the site is in safe mode).
+    def enabled_plugins
+      @enabled_plugins ||= if site.safe
+                             site.gems & whitelist
+                           else
+                             site.gems
+                           end
     end
 
     # Check whether a gem plugin is allowed to be used during this build.
--- a/test/test_plugin_manager.rb
+++ b/test/test_plugin_manager.rb
@@ -47,11 +47,8 @@ class TestPluginManager < JekyllUnitTest
       expect(Jekyll::External).to(
         receive(:require_with_graceful_fail).with(gems).and_return(nil)
       )
-      site = double(:gems => gems)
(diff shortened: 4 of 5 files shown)


lib/jekyll/plugin_manager.rb
```diff
@@ -57,6 +55,18 @@ def self.require_from_bundler
       end
     end
 
+    # Build an array of enabled plugin names.
+    #
+    # Returns an array of strings, each string being the name of a plugin
+    #   that is specified (and whitelisted in case the site is in safe mode).
+    def enabled_plugins
+      @enabled_plugins ||= if site.safe
+                             site.gems & whitelist
```
ashmaroli: think this should be `&&` instead

Crunch09: I assumed a plugin is only enabled if it is both listed in the `site.gems` (`plugins` config key) and in the `whitelist`? Isn't the whitelist not only a filter for this list for safe plugins?
```ruby
a = ["foo", "bar"]
b = ["bar", "baz"]
a & b  # => ["bar"]
a && b # => ["bar", "baz"]

**Comment by Crunch09** [comment]
@ashmaroli Nope, internally we still [kept it to being `gems`](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/site.rb#L53-L54) to avoid a breaking change.

File: lib/jekyll/plugin_manager.rb (hunk 1 above)
Comment (ashmaroli) [inline lib/jekyll/plugin_manager.rb]: Thanks for the example. I did not know this

**Comment by ashmaroli** [comment]
> Nope, internally we still kept it to being `gems`

Yes, I know that. And that's why its confusing..:
Any key in the `_config.yml` is available via  `{{ site }}` namespace. 

So if a user has `plugins: ["foo", "bar", "baz"]` in their config, wouldn't this be equal to `{{ site.plugins }}` irrespective of `--safe` switch?
Oh! wait! That's probably why you're overriding it with this PR .......

**Comment by Crunch09** [comment]
@ashmaroli I agree it's confusing 🙂  Unfortunately IMHO we can only fix it in the next major version update. As it is right now (on master) basically:
```
config "plugins" == Site#gems
config "plugins_path" == Site#plugins
{{ site.plugins }} == SiteDrop#plugins
```
Hope this makes it clearer!

File: lib/jekyll/drops/site_drop.rb [hunk 2]
```diff
@@ -37,6 +37,10 @@ def collections
         @site_collections ||= @obj.collections.values.sort_by(&:label).map(&:to_liquid)
       end
 
+      def plugins
```
Crunch09: @ashmaroli 's comment made me think about that. Maybe we should call this something like `enabled_plugins` instead? Otherwise this might be considered a breaking change as `{{ site.plugins }}` before this PR was referencing the `plugins` config key.

features/plugins.feature
```diff
@@ -35,3 +35,22 @@ Feature: Configuring and using plugins
     And I should see "Whatever" in "_site/index.html"
     And the "_site/test.txt" file should exist
     And I should see "this is a test" in "_site/test.txt"
+
+  Scenario: Plugin is enabled
+    Given I have an "index.html" page that contains "{% if site.plugins contains "jekyll_test_plugin" %} plugin loaded {% endif %}"
```
benbalter: Did you test that this works if the tag itself is called (e.g., `{% seo %}`)? IIRC the issue was with liquid parsing the entire file in one pass, rather than respecting `if` logic.

lib/jekyll/plugin_manager.rb
benbalter: Why not use `site.gems.select { |plugin| plugin_allowed?(plugin) }`?

lib/jekyll/plugin_manager.rb
```diff
@@ -25,9 +25,7 @@ def conscientious_require
     #
     # Returns nothing.
     def require_gems
-      Jekyll::External.require_with_graceful_fail(
-        site.gems.select { |plugin| plugin_allowed?(plugin) }
-      )
+      Jekyll::External.require_with_graceful_fail(enabled_plugins)
     end
```
benbalter: What happens if a plugin is in the plugins array but not in the Gemfile?

features/plugins.feature
Crunch09: Apologies, i thought i did test that manually, but turns out i didn't... This is indeed not working. The only solution i'd see is to override [`Liquid::Block#unknown_tag`](https://github.com/Shopify/liquid/blob/1370a102c95b83aeb03d9b7290b269435bc66753/lib/liquid/block.rb#L26-L38) but i'm afraid there is no logic for that which would make sense?

lib/jekyll/plugin_manager.rb
Crunch09: I Just checked, same behavior as before. It throws a `Jekyll::Errors::MissingDependencyException` for that plugin.

ashmaroli:
This feature may get side-stepped entirely by a site not in `--safe` mode, when a plugin is `enabled` or `required` directly without it being added to the config file:
  * enabled by including under `:jekyll_plugins` group in the Gemfile.
  * listed as `runtime_dependencies` of a theme-gem in conjunction with `Jekyll - v3.5`

ashmaroli:
@Crunch09 IMO, instead of checking if a plugin is listed in a particular list, it'd be awesome if we could check if a particular plugin has been loaded into memory. ( Similar to how Ruby plugins can be checked  with a `:respond_to?` call. )

Crunch09:
@ashmaroli Not sure about that, because we would need to run through to those "lists" (e.g. `Site#gems`) anyway to figure out if a given plugin name is actually loaded as a jekyll plugin and not just a random ruby gem.
That said i don't know if merging this PR is even still an option after the most important usecase for this feature IMHO can only be fixed if liquid would change its parsing.

Crunch09:
Closing this, because as @benbalter pointed out, liquid is parsing the entire file in one go and therefore using tags from unavailable plugins would still throw an error 😞

PR #6048, opened 2017-04-24
fix sentence in documentation

ashmaroli:
--
/cc @jekyll/documentation


pnn:
while we're on it, we might as well remove `on that` in this sentence. it sounds redundant

ashmaroli:
I agree :+1:

DirtyF:
@jekyllbot: merge +docs

PR #6053, opened 2017-04-27
adjusted date-format in sitemap

meyer-sven:
According to the linked specification https://www.sitemaps.org/protocol.html the date-format is https://www.w3.org/TR/NOTE-datetime, e. g. "2017-04-27". The current sitemap produces dates like "2017-April-27" and fails the Google test for sitemaps in the Google Search Console.

### DIFF
--- a/docs/_tutorials/convert-existing-site-to-jekyll.md
+++ b/docs/_tutorials/convert-existing-site-to-jekyll.md
@@ -457,7 +457,7 @@ search: exclude
     {% for page in site.pages %}
     <url>
         <loc>{{page.url}}</loc>
-        <lastmod>{{site.time | date: '%Y-%B-%d' }}</lastmod>
+        <lastmod>{{site.time | date: '%Y-%m-%d' }}</lastmod>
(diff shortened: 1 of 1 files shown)


DirtyF:
Thanks @meyer-sven for helping improving the docs!
 📅  🌐  🔍  

@jekyllbot: merge +docs

DirtyF:
@pathawks Both plugins are mentioned at the end of each section.

> You can also auto-generate your sitemap by adding a gem called jekyll-sitemap. This gem will also work on GitHub Pages.

> You can also auto-generate your posts feed by adding a gem called jekyll-feed. This gem will also work on GitHub Pages.

pathawks:
I wonder if it would be better to make that more prominent?
The vast majority of users would be better off _not_ writing their own XML templates.

DirtyF:
@pathawks if you ask me plugins should be the default way of pointing new users to getting a feed and a sitemap with Jekyll, and then you mention you can also create custom templates if you need more advances options.

PR #6054, opened 2017-04-28
Ignore final newline in folded YAML string

ashmaroli:
**before:** `{{ site.description }}` =>
```
"Write an awesome description for your new site here. You can edit this line in _config.yml. It
will appear in your document head meta (for Google search results) and in your feed.xml site
description./n"
```

   **after:** `{{ site.description }}` =>
```
"Write an awesome description for your new site here. You can edit this line in _config.yml. It
will appear in your document head meta (for Google search results) and in your feed.xml site
description."
```

### DIFF
--- a/lib/site_template/_config.yml
+++ b/lib/site_template/_config.yml
@@ -15,7 +15,7 @@
 # in the templates via {{ site.myvariable }}.
 title: Your awesome title
 email: your-email@example.com
-description: > # this means to ignore newlines until "baseurl:"
+description: >- # this means to ignore newlines until "baseurl:"
   Write an awesome description for your new site here. You can edit this
   line in _config.yml. It will appear in your document head meta (for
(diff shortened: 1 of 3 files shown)


ashmaroli:
> Should we also test that the new site's site.config["description"] doesn't have a trailing newline?

I think its implied..

parkr:
> I think its implied..

That's the point... I don't want it to be implicit, I think if we're making this change we should *explicitly* test it.

DirtyF:
@ashmaroli do you intend to test for a traling newline here?

ashmaroli:
> do you intend to test for a trailing newline here?

Sorry, I forgot all about this. Thanks for reminding.. Will push a commit shortly.

ashmaroli:
Can this be included in `v3.6`?

DirtyF:
@jekyllbot: merge +minor

PR #6058, opened 2017-04-29
patch URLFilters to prevent `//`

### LINKED ISSUE #6057: Handle '//' (double slash) caused by baseurl: '/'

---

## My Reproduction Steps

To start off with, I realise that setting baseurl to `/` is the wrong way to do it, but in reality, it seems it is often done, and sometimes seems to be the only way to workaround other bugs/things:
* https://github.com/jekyll/jekyll-seo-tag/issues/200
* https://github.com/github/pages-gem/issues/359#issuecomment-263375327
* https://github.com/jekyll/github-metadata/issues/87
* etc

I figured that with the [relative/absolute filters added in 3.3](https://jekyllrb.com/news/#2-relative_url-and-absolute_url-filters) there was now a standard/universal place to fix this in a way that would make the end user experience easier, while possibly also hinting them towards not using the 'bad method'

```
# _config.yml
baseurl: "/"

# default.html
<h1>{{ "/docs/assets/" | relative_url }}</h1>
<h1>{{ "/docs/assets/" | absolute_url }}</h1>
```

Outputs
```
<h1>//docs/assets/</h1>
<h1>http://localhost:4000//docs/assets/</h1>
```

## The Output I Wanted

```
<h1>/docs/assets/</h1>
<h1>http://localhost:4000/docs/assets/</h1>
```

This could be achieved by using a regex/similar to strip off any trailing/leading slashes before concatenation.

For what it's worth, the hacky filter I used in my theme to work around this was:

```
module Jekyll
  module ChompSlashFilter
    def chomp_slash(input)
      input.chomp('/')
    end
  end
end

Liquid::Template.register_filter(Jekyll::ChompSlashFilter)
```

But this obviously doesn't let me work around it when it comes up in other plugins.

It would also be reasonable to expect a warning from Jekyll about how using `/` in baseurl is a bad idea/why, and how to do it better/properly (maybe a link to a blog post about it or similar).

ashmaroli:
Fix #6057 

/cc @jekyll/ecosystem 
requesting additional review from @parkr as well.

### DIFF
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -35,8 +35,7 @@ module Jekyll
       end
 
       def sanitized_baseurl
-        baseurl = site.config["baseurl"]
-        return baseurl unless baseurl.nil? || (baseurl == "/")
+        site.config["baseurl"].chomp("/")
       end
 
       def ensure_leading_slash(input)
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -11,7 +11,6 @@ module Jekyll
       def absolute_url(input)
         return if input.nil?
         return input if Addressable::URI.parse(input).absolute?
-        site = @context.registers[:site]
         return relative_url(input).to_s if site.config["url"].nil?
         Addressable::URI.parse(site.config["url"] + relative_url(input)).normalize.to_s
       end
@@ -23,14 +22,23 @@ module Jekyll
       # Returns a URL relative to the domain root as a String.
       def relative_url(input)
         return if input.nil?
-        site = @context.registers[:site]
-        return ensure_leading_slash(input.to_s) if site.config["baseurl"].nil?
+        return ensure_leading_slash(input.to_s) if sanitized_baseurl.nil?
         Addressable::URI.parse(
-          ensure_leading_slash(site.config["baseurl"]) + ensure_leading_slash(input.to_s)
+          ensure_leading_slash(sanitized_baseurl) + ensure_leading_slash(input.to_s)
         ).normalize.to_s
       end
 
       private
+
+      def site
+        @context.registers[:site]
+      end
+
+      def sanitized_baseurl
+        baseurl = site.config["baseurl"]
+        return baseurl unless baseurl.nil? || (baseurl == "/")
+      end
+
       def ensure_leading_slash(input)
         return input if input.nil? || input.empty? || input.start_with?("/")
         "/#{input}"
--- a/test/test_filters.rb
+++ b/test/test_filters.rb
@@ -458,6 +458,24 @@ class TestFilters < JekyllUnitTest
         assert_equal "/base", filter.relative_url(page_url)
       end
 
+      should "not prepend a forward slash if baseurl ends with a single '/'" do
+        page_url = "/css/main.css"
+        filter = make_filter_mock({
+          "url"     => "http://example.com",
+          "baseurl" => "/base/",
+        })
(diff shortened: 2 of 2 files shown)


lib/jekyll/filters/url_filters.rb
```diff
@@ -23,14 +22,23 @@ def absolute_url(input)
...
 
       private
+
+      def site
+        @context.registers[:site]
+      end
+
+      def sanitized_baseurl
+        baseurl = site.config["baseurl"]
+        return baseurl unless baseurl.nil? || (baseurl == "/")
```
alias1: Will this fail and not return if baseurl is nil? (Unsure of ruby specifics for unless)

What happens if baseurl is set to /foo/bar/ or even something really silly like "///////" or /foo/bar//" etc?

Can we compress trailing slashes to a single one, or removed entirely, as required by the concatenation?

test/test_filters.rb
```diff
@@ -448,6 +457,15 @@ def select; end
         })
         assert_equal "/base", filter.relative_url(page_url)
       end
+
+      should "not prepend a forward slash if both input and baseurl are simply '/'" do
```
alias1: Should this be "append"?

lib/jekyll/filters/url_filters.rb
alias1: And I guess I haven't really checked what currently happens, but what if similarly there is some silliness going on in the site URL part with slashes. Can we robustly handle this in both potential locations?

pathawks: > What happens if baseurl is set to /foo/bar/ or even something really silly like "///////" or /foo/bar//" etc?

Pray, Mr. Babbage, if you put into the machine wrong figures, will the right answers come out?

alias1:
A fair point, but I suppose my use cases there were exaggerated examples to highlight a point. Should we be catering to a specific instance of a problem (e.g. Baseurl = /), or the more general pattern of the problem (ending in 1 or more slashes, that when concatenated cause duplication)

My personal view leans towards fixing the pattern rather than the specific instance.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

pathawks:
But multiple slashes are perfectly legal in a URL. Do we really want to stop supporting that just so setting  `baseurl: "////////"` has no effect?

Let's try to figure out what the root problem is and try to fix that with as few unintended consequences as possible, rather than trying to decide how hypothetical situations should be magically changed without the users permission.

lib/jekyll/filters/url_filters.rb
ashmaroli: > Will this fail and not return if baseurl is nil? 

Ruby will `return nil` if `baseurl` equals `nil` or `/`

> Can we compress trailing slashes to a single one, or removed entirely, as required by the concatenation?

IMO substituting a regex match with something is not necessary as it is known to affect performance.  But again, there is no issue in employing that method if there's a strong use case. Setting `baseurl` with multiple slashes is wrong.
The least appropriate measure would be to output a warning message that the `baseurl` has been configured incorrectly.

ashmaroli:
updated to return valid url if `baseurl == /base/`

parkr:
@jekyllbot: merge +bug

ashmaroli:
Would it make sense to partially revert this (remove the private method `:site` introduced here) instead of fixing plugins downstream?

/cc @parkr @benbalter @pathawks

parkr:
> Would it make sense to partially revert this (remove the private method :site introduced here) instead of fixing plugins downstream?

Making this more backwards-compatible sounds like a great idea!

PR #6059, opened 2017-04-30
add formester to the list of saas form backend

ankitsinghaniyaz:
I saw that Jekyll shares list of all SAAS form backend I would like to add one I am building to the list. I hope this will give users more options to choose from.


mattr-:
@jekyllbot: merge +doc

PR #6070, opened 2017-05-09
Typo fix (welcomed change -> welcome change).

relsqui:
This is admittedly ambiguous--it would be grammatical either way--but a quick corpus search confirmed my intuition that "welcome change" is the normal wording (by about 40:1).


pnn:
@jekyllbot: merge +docs

pnn:
@relsqui nice catch!

PR #6073, opened 2017-05-10
Add top-level `layout` liquid variable to Documents

### LINKED ISSUE #6071: layout variable not available to index.html
- [X] I am on (or have tested on) ***macOS*** 10+

jekyll 3.4.3

I have index.html in my working directory. The frontmatter for this file is:

`---`
`layout: default`
`title: homepage`
`---`

in the _layouts directory, I have a file named default.html - the frontmatter for this file is:

`---`
`nav:`
` - text: Home`
`   url: /`
` - text: Second Page`
`   url: /second-page/`
` - text: Third Page`
`   url: /third-page/`
`---`

Inside index.html, I am including a file from the _includes directory.

`{% include nav.html %}`

The contents of nav.html contains my navigation aid which incorporates a for loop. Lets just say that if we put the following in here, it should output onto the page.

`{{ layout.nav }}`

For second page, and third page (they contain the same frontmatter as index) it works. The issue is that for index.html **it doesn't.**

Furthermore, index.html doesn't appear to inherit **anything** at all from the layout variable.

In my _config.yml - I have **include: ['_pages']**
The second and third pages are located in the _pages directory.

To further test, I disabled this line and reloaded the site. index.html still doesn't show anything from the layout variable.

I re-enabled **include: ['_pages']** and duplicated index.html into the _pages directory. Surprisely it now works. But the same file index.html has to be in both locations.

This seems strange and very much looks like a bug.

ayastreb:
Fixes #6071 

we add "layout" key to payload when we place doc into layout (see `def render_layout`), so when first doc is rendered - layout data is not present in payload and only added there after, so it is available for next documents.

In this PR I add layout data to payload before rendering the document.

### DIFF
--- a/lib/jekyll/renderer.rb
+++ b/lib/jekyll/renderer.rb
@@ -54,6 +54,7 @@ module Jekyll
       assign_pages!
       assign_related_posts!
       assign_highlighter_options!
+      assign_layout_data!
 
       Jekyll.logger.debug "Pre-Render Hooks:", document.relative_path
       document.trigger_hooks(:pre_render, payload)
@@ -231,6 +232,14 @@ module Jekyll
     def assign_highlighter_options!
       payload["highlighter_prefix"] = converters.first.highlighter_prefix
       payload["highlighter_suffix"] = converters.first.highlighter_suffix
+    end
+
+    private
+    def assign_layout_data!
+      layout = layouts[document.data["layout"]]
+      if layout
+        payload["layout"] = Utils.deep_merge_hashes(layout.data, payload["layout"] || {})
+      end
     end
 
     private
--- a/test/source/_layouts/default.html
+++ b/test/source/_layouts/default.html
@@ -1,3 +1,6 @@
+---
+front_matter_var: variable from layout
+---
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
     "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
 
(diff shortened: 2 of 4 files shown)


lib/jekyll/renderer.rb
```diff
@@ -234,6 +235,14 @@ def assign_highlighter_options!
     end
 
     private
+    def assign_layout_data!
+      layout = layouts[document.data["layout"]]
```
parkr: What do you do if that layout has layouts?

lib/jekyll/renderer.rb
```diff
@@ -54,6 +54,7 @@ def run
       assign_pages!
       assign_related_posts!
       assign_highlighter_options!
+      assign_layout_data!
```
parkr: What if a plugin modifies the layout data?

lib/jekyll/renderer.rb
parkr: What if the document doesn't have a layout parameter? `layouts[nil]`?

ayastreb: Yes, it's the same line as in https://github.com/jekyll/jekyll/blob/master/lib/jekyll/renderer.rb#L145

ayastreb: Not sure... basically the problem is that we don't have layout data in `payload` when we render document [here](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/renderer.rb#L75). I've tried to move the  `place_in_layouts` call before rendering the document (because we add layout data to `payload` in that method), but that breaks a lot of tests.

Not sure how to fix this without too much changes :)

ayastreb:
hi @parkr !
how should we proceed with this PR? Do you see a different approach to fixing the bug or is this fine?

parkr:
@jekyllbot: merge +minor

PR #6075, opened 2017-05-15
Add strip_index filter

benbalter:
https://github.com/jekyll/jekyll-sitemap/pull/170 made me realize that in several of Jekyll's core plugins we use the pattern `{{ file.path | replace:'/index.html','/' }}`.

That filter works, but is (A) easy to forget, and (B) is a blind replace, not a replace of the a trailing `/index.html`. 

This PR adds a quick `strip_index` filter (very up for a better name, maybe `pretty_permalink`?), to implement the above logic as a native URL filter.

Thoughts?

### DIFF
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -37,7 +37,7 @@ module Jekyll
       # Returns a URL with the trailing `/index.html` removed
       def strip_index(input)
         return if input.nil?
-        input.sub(/\/\.index\.html?$/, '/')
+        input.sub(%r!/\.index\.html?$!, "/")
       end
 
       private
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -36,7 +36,7 @@ module Jekyll
       #
       # Returns a URL with the trailing `/index.html` removed
       def strip_index(input)
-        return if input.nil?
+        return if input.nil? || input.to_s.empty?
         input.sub(%r!/index\.html?$!, "/")
       end
 
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -37,7 +37,7 @@ module Jekyll
       # Returns a URL with the trailing `/index.html` removed
       def strip_index(input)
         return if input.nil?
-        input.sub(%r!/\.index\.html?$!, "/")
+        input.sub(%r!/index\.html?$!, "/")
       end
 
       private
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -30,6 +30,16 @@ module Jekyll
         ).normalize.to_s
       end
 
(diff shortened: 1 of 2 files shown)


test/test_filters.rb
```diff
@@ -450,6 +450,28 @@ def select; end
       end
     end
 
+    context "strip_index filter" do
+      should "strip trailing /index.html" do
+        assert_equal `/foo/`, @filter.strip_index('/foo/index.html')
+      end
```
ashmaroli: Why are we asserting `/foo/` enclosed in back ticks instead of `quotes`?

pathawks:
Is there a time when “we” (plugins or themes) would want to use the URL filters _without_ stripping `index.html`?

lib/jekyll/filters/url_filters.rb
```diff
@@ -30,6 +30,16 @@ def relative_url(input)
...
       end
 
+      # Strips trailing `/index.html` from URLs to create pretty permalinks
+      #
+      # input - the URL with a possible `/index.html`
+      #
+      # Returns a URL with the trailing `/index.html` removed
+      def strip_index(input)
+        return if input.nil?
+        input.sub(%r!/\.index\.html?$!, "/")
```
ashmaroli: Error in the regex here:
```diff
- input.sub(%r!/\.index\.html?$!, "/")
+ input.sub(%r!/index\.html?$!, "/")
```

ashmaroli:
> Is there a time when “we” (plugins or themes) would want to use the URL filters without stripping `index.html`?

From the test case included, I'm guessing:
```
/foo/index.html#bar
```

parkr:
LGTM once the tests are passing!

benbalter:
> Is there a time when “we” (plugins or themes) would want to use the URL filters _without_ stripping `index.html`?

I'm thinking that may be a better approach. Any reason not to? 

@parkr mentioned the edge case where servers don't support indexes, but this should be the default in 99% of environments.

pup:
`strip_index` sounds fine to me, it describes what it does pretty well

benbalter:
Do we want to go this route or the route of automatically doing this in the existing URL filters?

pup:
@benbalter i think the latter would be a major change? considering that we want to get this in 3.5, let's go with this for now

pathawks:
> considering that we want to get this in 3.5, let's go with this for now

So plugins + themes + users should change all of their layout code to use this filter, and then change back at some point in the future?

😓

pup:
you're right.... is this big enough of a usecase to warrant automatically doing it?

parkr:
Well if they had this filter then we automatically applied it, then this filter would just no-op. Making it opt-in by just supplying the filter gives us space to get it right and make it fast before pushing it into the process for everyone. Thoughts?

ashmaroli:
> We shouldn't let perfect be the enemy of good.

:+1: quote :clap: :smiley:

pathawks:
@jekyllbot: merge +minor

parkr:
@jekyllbot: merge +minor

parkr:
LOL jinx.

PR #6078, opened 2017-05-18
Use more precise language when explaining links

szafranek:
The original wording implies that the links *is* `page_a.md`. Instead, the link is *in* or *from* that page.


DirtyF:
@jekyllbot: merge +docs

PR #6081, opened 2017-05-22
Document difference between cgi_escape and uri_escape #5970

crispgm:
/cc: @pathawks

PR #6086, opened 2017-05-24
Add test for uri_escape on reserved characters

crispgm:
#6081 docs the difference between `uri_escape` and `cgi_escape` on reserved characters. This PR is to add test which covers all the reserved characters.


parkr:
@jekyllbot: merge +dev

PR #6090, opened 2017-05-25
Update plugins example with new name

missmatsuko:
I had this console message when I tried to use the example:
"Deprecation: The 'plugins' configuration option has been renamed to 'plugins_dir'. Please update your config file accordingly."


DirtyF:
@missmatsuko We renamed the `gems` key to `plugins` in #5130 but didn't do a proper release since 😕

Crunch09:
@DirtyF Sorry, i totally missed that you requested a review from me on this :(
@missmatsuko Thanks for your PR but as @DirtyF mentioned this is correct and it will hopefully be more obvious as soon as there is a new release.

Issue #5953, opened 2017-03-14
Collection documents with future date and output set to false are not ignored


BrandonDusseau:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [ ] This is a feature request.

---

- [ ] I am on (or have tested on) ***macOS*** 10+
- [x] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux (Linux Mint)
- [ ] I am on (or have tested on) ***Fedora*** GNU/Linux
- [ ] I am on (or have tested on) ***Arch*** GNU/Linux
- [x] I am on (or have tested on) ***Other*** GNU/Linux
- [ ] I am on (or have tested on) ***Windows*** 10+

---

- [ ] I was trying to install.
- [ ] There is a broken Plugin API.
- [ ] I had an error on GitHub Pages, and I have reproduced it locally.
- [ ] I had an error on GitHub Pages, and GitHub Support said it was a Jekyll Bug.
- [ ] I had an error on GitHub Pages and I did not test it locally.
- [x] I was trying to build.
- [x] It was another bug.

## My Reproduction Steps

1. I created a collection, we'll call it "mycoll", with the following configuration in `_config.yml`:

```
collections:
  mycoll:
    output: false
```

2. I added a few files to the `_mycoll` directory; one in the past, one in the present, and one in the future. Each one contains front matter similar to the following:

```
date: 2017-03-16 00:00:00-04
link: http://example.com/place
preview: example.png
```

3. I used the following to fetch the latest item on my page:

```
{% assign item = site.mycoll.last %}
```

This is where the problem begins. For posts and collections with `output` set to `true`, posts and documents with the `date` field set to a time in the future are not published by default, unless built with `--future` or `future: true` in the config file; this is documented behavior. However, if `output` is set to `false`, objects with future dates are still displayed. This is not documented behavior.

This behavior is reproducible on both Jekyll 3.4.2 and 3.3.0.

## The Output I Wanted

What I would expect to happen is that ALL collection documents with future dates would be ignored, not just collections with `output: true`. Either that, or the documentation should be updated to reflect the current behavior.

Build output:

```
Configuration file: _config.yml
Configuration file: _config_beta.yml
            Source: _site
       Destination: public
 Incremental build: disabled. Enable with --incremental
[...]

PR #6103, opened 2017-05-31
Test if hidden collections output a document with a future date

ashmaroli:
Tests #5953 

/cc @jekyll/build

### DIFF
--- a/features/collections.feature
+++ b/features/collections.feature
@@ -111,6 +111,10 @@ Feature: Collections
     And the _site directory should exist
     And I should see "content for Rover" in "_site/puppies/rover.html"
     And the "_site/puppies/fido.html" file should not exist
+    When I run jekyll build --future
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And the "_site/puppies/fido.html" file should exist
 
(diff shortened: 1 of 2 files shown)


features/collections.feature
```diff
@@ -94,6 +94,51 @@ Feature: Collections
...
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And the "_site/puppies/fido.html" file should exist
+
+  Scenario: Hidden collection with document with future date
+    Given I have a _puppies directory
+    And I have the following documents under the puppies collection:
+      | title  | date       | content             |
+      | Rover  | 2007-12-31 | content for Rover.  |
+      | Fido   | 2020-12-31 | content for Fido.   |
```
crispgm: IMO, `2020` is not "future" enough.

ashmaroli: no? In theory, even `tomorrow` is future..

crispgm: @ashmaroli Right! But 3 years is not far. And this test case will be outdated.

parkr:
@jekyllbot: merge +dev

Issue #6105, opened 2017-05-31
Use a source folder by default


noraj1337:
- [X] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [X] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [X] This is a feature request.
- [X] I am on (or have tested on) ***Arch*** GNU/Linux (Manjaro in fact)

---

It will be cleaner to a default source folder by using something like `source: _sources` in `_config.yml` **by default**.

This will allow to keep a clean repository tree view. For a website with a lot of page, not using a source folder make the root directory full of file and that mix config files, gem fiels and source files so this is dirty.

Also if user create some other directories that he don't want to include in the source but forget to add in the `exclude` of `_config.yml` these folders will be added/copied to the build. Using a source folder by default will prevent that (unwanted behaviour) to happen.

So what is ask is already possible but this will be nice if it is default. That's really not a lot of works, only the documentation need to be updated.

noraj1337:
I created a PR https://github.com/jekyll/jekyll/pull/6106

Note: maybe documentation need more changes

parkr:
Please see https://github.com/jekyll/jekyll/pull/6106#issuecomment-308547870 for explanation why this isn't necessary. To prevent a breaking change, we will not change this default.

PR #6106, opened 2017-05-31
default source folder

noraj1337:
- add a default source folder
- modify the documentation to take into account the change

### DIFF
--- a/docs/_docs/structure.md
+++ b/docs/_docs/structure.md
@@ -18,24 +18,25 @@ A basic Jekyll site usually looks something like this:
 ├── _config.yml
 ├── _data
 |   └── members.yml
-├── _drafts
-|   ├── begin-with-the-crazy-ideas.md
-|   └── on-simplicity-in-technology.md
 ├── _includes
 |   ├── footer.html
 |   └── header.html
 ├── _layouts
 |   ├── default.html
 |   └── post.html
-├── _posts
-|   ├── 2007-10-29-why-every-programmer-should-play-nethack.md
-|   └── 2009-04-26-barcamp-boston-4-roundup.md
 ├── _sass
 |   ├── _base.scss
 |   └── _layout.scss
 ├── _site
 ├── .jekyll-metadata
-└── index.html # can also be an 'index.md' with valid YAML Frontmatter
+└── _sources
+    ├── about.md
+    ├── index.md # can also be an 'index.html' with valid YAML Frontmatter
+    ├── _drafts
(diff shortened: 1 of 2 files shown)


ashmaroli:
sorry mate, simply adding a line to the config file does not change a core functionality that has been around since inception. This PR in its current state will be closed as it simply creates confusion..even though the intent has been to illustrate an idea..

noraj1337:
@ashmaroli There is no core fonctionnality to change, they mecanism already exist, we just need to change the default value.

DirtyF:
I get the idea but this is a breaking change for Github Pages while it adds little value.

/cc @jekyll/stability

benbalter:
@noraj1337 Thanks for the pull request! :smile: 

Can you share a bit more about why you think we should make the change?

noraj1337:
@DirtyF Are you sure about that? I didn't see [anything](https://help.github.com/articles/using-jekyll-as-a-static-site-generator-with-github-pages/) and you can still build Jekyll locally and deploy the `_site` folder on gh-pages only (that's why I'm doing with Hexo).

@benbalter I explained it on the issue https://github.com/jekyll/jekyll/issues/6105

benbalter:
> I explained it on the issue #6105

To confirm, the motivation is not having e.g., your Gemfile or README in the same folder as markdown files?

noraj1337:
@benbalter This should answer your question:

```
$ tree                                                                      noraj@manji
.
├── _config.yml
├── Gemfile
├── Gemfile.lock
├── _public
│   ├── about
│   │   └── index.html
│   ├── assets
│   │   └── main.css
│   ├── feed.xml
│   ├── index.html
│   └── jekyll
│       └── update
│           └── 2017
│               └── 05
│                   └── 31
│                       └── welcome-to-jekyll.html
└── _sources
    ├── about.md
    ├── index.md
    └── _posts
        ├── 2017-05-31-welcome-to-jekyll.markdown
        └── test.md

10 directories, 12 files
```

pathawks:
@noraj1337 The question is with regard to motivation, not a particular directory hierarchy.

noraj1337:
@pathawks I don't agree, putting all source files in the root directory is a mess.

parkr:
This is a backwards-incompatible change and will not be accepted. You have the opportunity to use the `--source` flag for the CLI and `source` setting in your `_config.yml` and that is its purpose: to set apart your source from other files.

noraj1337:
@parkr I know the option but this should have been default...
No need for me anywayn using middleman now, this is more hackable and cleaner.

Issue #5704, opened 2016-12-29
Overriding theme sass partials


ashmaroli:
- I think this is a bug with Jekyll
- I'm using Jekyll 3.3.1 on Windows 7

At present, if a user needed to override one of minima's sass partials, they have to:
  - have a copy of the concerned sass partial at `source_dir` (expected, documented)
  - have a copy of the **main sass file (`_sass/minima.scss`) at `source_dir` as well.** (unexpected, undocumented)

#### Inference
  - Jekyll loads sass partial(s) at `source_dir` only if the main sass file importing the partial(s) is also present at `source_dir`

parkr:
I think overriding sass partials with jekyll is a much too complicated procedure. CSS naturally cascades so add the styles you need and CSS will do its thing. I think we should be sticking to output CSS files instead of getting into the business of dealing with overriding sass partials.

That said, we do set the Sass LOAD_PATH to the site's source then to the theme so this should already be possible.

ashmaroli:
hmm.. 
> I think overriding sass partials with jekyll is a much too complicated procedure... I think we should be sticking to output CSS files instead of getting into the business of dealing with overriding sass partials.

Our documentation does not say that though.. hence users tend to use the same formula for overriding any other theme file:
- copy partial to `<source>/_sass`
- make changes
- run `jekyll serve`

> so this should already be possible.

I'll try to come up with a cucumber feature to demonstrate this on my fork.

ashmaroli:
> I'll try to come up with a cucumber feature to demonstrate this on my fork.

@parkr : 
branch compare: [review-me](https://github.com/jekyll/jekyll/compare/master...ashmaroli:theme-sass)
travis-log: [review-me](https://travis-ci.org/ashmaroli/jekyll/jobs/187614602#L448)

PR #5690, opened 2016-12-26
Improve theme docs

tomjohnson1492:
See https://github.com/jekyll/jekyll/pull/5630 for more details on the update. 

@jekyll/documentation

PR #6107, opened 2017-05-31
Explain how to override theme styles

Strangehill:
#5704 
#5690 https://github.com/jekyll/jekyll/pull/5690#issuecomment-269599039

The theme documentation is frustratingly incomplete when it comes to overriding theme styles. This, or something similar, seems long overdue.

### DIFF
--- a/docs/_docs/themes.md
+++ b/docs/_docs/themes.md
@@ -88,7 +88,7 @@ With a clear understanding of the theme's files, you can now override any theme 
 
 Let's say, for a second example, you want to override Minima's footer. In your Jekyll site, create an `_includes` folder and add a file in it called `footer.html`. Jekyll will now use your site's `footer.html` file instead of the `footer.html` file from the Minima theme gem.
 
-To modify any stylesheet you must take the extra step of also copying the main sass file (`_sass/minima.scss`) into your source directory. 
+To modify any stylesheet you must take the extra step of also copying the main sass file (`_sass/minima.scss` in the Minima theme) into your source directory. 
 
 Jekyll will look first to your site's content before looking to the theme's defaults for any requested file in the following folders:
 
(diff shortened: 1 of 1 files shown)


tomjoht:
I'm glad to see this update. Lots of people have trouble trying to override Sass styles in Minima, and the process wasn't well documented. Thanks! 

Re "An alternative, to continue getting theme updates on all stylesheets, is to use higher specificity CSS selectors in your own additional, originally named CSS files." Can you add some more detail here, including an example?  This is an important detail that keeps tripping people up. I have seen this issue pop up half a dozen times in the forums.

Can a user add a _sass folder in their project, with a new partial and reference it in the main.css file? If so, would a user would also create an assets folder with main.scss that would be something like this? 

```
@import "minima";
@import "mycustomstyles";
```

I'm really not sure what the process is or what works. But I do feel that we need more detail around this point.

Also, should we add similar info into the [themes page](http://jekyllrb.com/docs/themes/) that explains how overriding Sass styles isn't done the same as overriding files in assets, layouts, or includes? This probably important information for theme developers to know.

tomjohnson1492:
I'm glad to see this update. Lots of people have trouble trying to override Sass styles in Minima, and the process wasn't well documented. Thanks! 

Re "An alternative, to continue getting theme updates on all stylesheets, is to use higher specificity CSS selectors in your own additional, originally named CSS files." Can you add some more detail here, including an example?  This is an important detail that keeps tripping people up. I have seen this issue pop up half a dozen times in the forums.

Can a user add a _sass folder in their project, with a new partial and reference it in the main.css file? If so, would a user would also create an assets folder with main.scss that would be something like this? 

```
@import "minima";
@import "mycustomstyles";
```

I'm really not sure what the process is or what works. But I do feel that we need more detail around this point.

Also, should we add similar info into the [themes page](http://jekyllrb.com/docs/themes/) that explains how overriding Sass styles isn't done the same as overriding files in assets, layouts, or includes? This probably important information for theme developers to know.

Strangehill:
I think that's exactly how it would work and didn't go into detail because I haven't totally figured it out myself yet and I'm hoping we can shed shed some light here. 

On instructions for theme creators: if they were "to ship a default /assets/styles.scss" as the instructions on creating a theme put it, it would seem wise for the creator to update theme styles everywhere but in this file to allow users to modify it freely.

DirtyF:
Thank you all!

@Strangehill I hope the override process is clearer for our users with this addition. 🎨 

@tomjohnson1492 good question,  I'm gonna merge this as is for now, and we can go further with multiple Sass files import later. (I don't develop themes myself right now).

@jekyllbot: merge +docs

PR #6108, opened 2017-05-31
Failing plugin feature test

agius:
I tried creating a plugin for `jekyll mything` in the `_plugins` directory for my site, and it doesn't seem like that's possible at the moment. Here's a scenario demonstrating the behavior. Am I missing something obvious?

I took a crack at getting it working, but it seems I'd have to instantiate a  `Jekyll::Site` to set up paths and get the command plugins. This goes against the current pattern in `exe/jekyll` and has all sorts of side effects - the configuration, plugin loading, and site building objects are pretty tightly coupled.

1. Is this a problem worth addressing at all? I can extract my command to a gem and use that publicly or privately, nbd
2. How much refactoring is that worth? I'd want to pick apart the path & plugin management from `Jekyll::Site` to `Jekyll::Configuration` ... or something. Seems a bit big for a first-time contributor

If this is not worth doing, feel free to close the PR. I can submit a different one that will update the docs and make this limitation clear.

If this is a good refactor opportunity, any guidance or suggestions before I dive into it would be welcome.

Cheers,

### DIFF
--- a/features/plugins.feature
+++ b/features/plugins.feature
@@ -11,6 +11,10 @@ Feature: Configuring and using plugins
     And I should see "Whatever" in "_site/index.html"
     And I should see "this is a test" in "_site/test.txt"
 
+  Scenario: Add a local command plugin
+    When I run jekyll stub
+    Then I should get a zero exit status
+
   Scenario: Add an empty whitelist to restrict all gems
     Given I have an "index.html" file that contains "Whatever"
     And I have a configuration file with:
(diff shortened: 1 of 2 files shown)


ashmaroli:
Hello @agius.
Some tips:
  - If you check existing commands, you'll see that they inherit from `Jekyll::Commands::Command` instead of `Jekyll::Command`.
  - I have never tried to run a custom command from within the `_plugins` directory. I always create `gems` out of them and add them to my `Gemfile` like any other gem-based plugin. (`.rb` file in `_plugins` may not do the trick due to the time they're pulled into memory.)
  - no refactoring required at all for a new command (from my own experience)

parkr:
It's explicitly not allowed in our flow to write a subcommand in the `_plugins` directory – it must be a separate RubyGem. I believe we have stated this before, but if it's not in the docs then it should be.

PR #6100, opened 2017-05-29
windows 10 tutorial

HenryKobin:
Hey guys! I got jekyll running flawlessly on my copy of windows 10 creators update with some pretty basic steps taken. I was curious if I could contribute this guide to the docs. Thanks for your time, let me know if you want me to edit anything / reformat the markdown.

/cc @parkr  @jekyllbot

PR #6109, opened 2017-06-01
Added windows to docs.yml

HenryKobin:
Added windows to list of docs in docs.yml after adding updated tutorial to the windows page (#6100)

/cc @jekyll/documentation

### DIFF
--- a/docs/_data/docs.yml
+++ b/docs/_data/docs.yml
@@ -6,6 +6,7 @@
   - usage
   - structure
   - configuration
+  - windows
 
 - title: Your Content
   docs:
--- a/docs/_data/docs.yml
+++ b/docs/_data/docs.yml
@@ -3,10 +3,10 @@
   - home
   - quickstart
   - installation
+  - windows
   - usage
   - structure
   - configuration
-  - windows
 
 - title: Your Content
   docs:


docs/_data/docs.yml
```diff
@@ -6,6 +6,7 @@
   - usage
   - structure
   - configuration
+  - windows
```
DirtyF: We could keep the `Jekyll on Windows` title and put it under the installation entry

HenryKobin:
@DirtyF I think it would be fine under Installation instead of Windows, because I'm running it with Ruby 2.3 / Jekyll and it's basically running as if I had it natively on Linux. I feel like Bash on windows is reaching the point where we could start allowing Windows users to feel more like they are an officially supported language, so they can at least get involved and start troubleshooting for themselves / contributing more to the jekyll ecosystem.
cc/ @jekyll/documentation @pathawks  @jekyllbot

docs/_data/docs.yml
DirtyF: this entry should be next to installation

DirtyF:
@HenryKobin I was only talking about the order of the links in the navigation menu. Let's keep it as a separate page for now. If @jekyll/windows says it's bug free, then we'll think about merging all installation instructions on the very same page.

HenryKobin:
@DirtyF corrected.

DirtyF:
@jekyllbot: merge +docs

PR #6082, opened 2017-05-22
Docs: rename `gems` to `plugins`

crispgm:
When I run the docs locally, Jekyll said:
```
Deprecation: The 'gems' configuration option has been renamed to 'plugins'. Please update your config file accordingly.
```

PR #6110, opened 2017-06-01
plugins -> gems

DirtyF:
Revert #6082 to prevent GitHub Pages automatic build from failing.


DirtyF:
@jekyllbot: merge +site

PR #6111, opened 2017-06-01
updated Bash on Ubuntu on Windows link in tutorial

HenryKobin:
Changed the link to Microsoft's install guide for Bash on Ubuntu on Windows to go directly to the tutorial instead of a different page. (pull #6100)
Also, we are considering moving the windows guide from its own page to the Installation directions.(Pull #6109).

/cc @jekyll/documentation


DirtyF:
@jekyllbot: merge +docs

PR #6126, opened 2017-06-09
Remove Blank Newlines from "Jekyll on Windows" Page

ericleong:
I was following the new instructions introduced by #6100 

Some of the bash commands have an extra newline, so they look like this:

jekyll

This pull request removes them.


pathawks:
Good catch! Thanks for noticing, and for taking the time to submit a PR :beers::+1:

/cc: @jekyll/documentation

DirtyF:
@jekyllbot: merge +docs

Issue #6127, opened 2017-06-09
Can't build new site.


Kaligule:
Hi, this is my first try using jekyll.

I installed jekyll and generated a new project in an empty directory. Then, since there already was an example blogpost in ```_posts``` I tried to build it (This is supposed to work right? Thats why the ```new``` command exists in the first place, isn't it?).

I got the following error:

```
Configuration file: ~/myblog/_config.yml
            Source: ~/myblog
       Destination: ~/myblog/_site
 Incremental build: disabled. Enable with --incremental
      Generating... 
             ERROR: YOUR SITE COULD NOT BE BUILT:
                    ------------------------------------
                    Invalid date '<%= Time.now.strftime('%Y-%m-%d %H:%M:%S %z') %>': Document 'vendor/bundle/gems/jekyll-3.4.3/lib/site_template/_posts/0000-00-00-welcome-to-jekyll.markdown.erb' does not have a valid date in the YAML front matter.
```

I found Issue #2938, which seems to be the same problem. I tried the suggested solution with ```echo "exclude: [vendor]" >> _config.yml``` and the problem was gone. #2938 is closed at this moment.

I guess this problem still exists for some reason, even though it is an old one and seems easy to fix. At least as a beginner I shouldn't have this problem, since it makes it harder to start using jekyll.

- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [ ] This is a feature request.

---

- [ ] I am on (or have tested on) ***macOS*** 10+
- [ ] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux
- [ ] I am on (or have tested on) ***Fedora*** GNU/Linux
- [ ] I am on (or have tested on) ***Arch*** GNU/Linux
- [x] I am on (or have tested on) ***Other*** GNU/Linux : CentOS
- [ ] I am on (or have tested on) ***Windows*** 10+

---

- [ ] I was trying to install.
- [ ] There is a broken Plugin API.
- [ ] I had an error on GitHub Pages, and I have reproduced it locally.
- [ ] I had an error on GitHub Pages, and GitHub Support said it was a Jekyll Bug.
- [ ] I had an error on GitHub Pages and I did not test it locally.
- [x] I was trying to build.
- [ ] It was another bug.

## My Reproduction Steps

```
[...]

PR #5361, opened 2016-09-16
Exclude vendor by default

DirtyF:
Fix #5359 

- [x] Update documentation

cc/ @jekyll/core

Issue #2938, opened 2014-09-19
invalid date '0000-00-00' from template


jnozsc:
Hi, 

I tried to upgrade my site from jekyll 1.x to 2.x on heroku.
I meet this issue.
<b>
2014-09-19T20:34:36.760150+00:00 app[web.1]: /app/vendor/bundle/ruby/2.1.0/gems/jekyll-2.4.0/lib/jekyll/utils.rb:95:in `rescue in parse_date': Invalid date '0000-00-00': Post '/vendor/bundle/ruby/2.1.0/gems/jekyll-2.4.0/lib/site_template/_posts/0000-00-00-welcome-to-jekyll.markdown.erb' does not have a valid date in the filename. (Jekyll::Errors::FatalException)
2014-09-19T20:34:36.760217+00:00 app[web.1]:    from /app/vendor/bundle/ruby/2.1.0/gems/jekyll-2.4.0/lib/jekyll/site.rb:143:in`each'
</b>

```
2014-09-19T20:34:31.414716+00:00 app[web.1]:    from /app/vendor/bundle/ruby/2.1.0/gems/jekyll-2.4.0/lib/jekyll/site.rb:147:in `block in read_directories'
2014-09-19T20:34:31.414735+00:00 app[web.1]:    from /app/vendor/bundle/ruby/2.1.0/gems/jekyll-2.4.0/lib/jekyll/site.rb:143:in `read_directories'
2014-09-19T20:34:31.414749+00:00 app[web.1]:    from /app/vendor/bundle/ruby/2.1.0/gems/jekyll-2.4.0/lib/jekyll/site.rb:44:in `process'
2014-09-19T20:34:31.414763+00:00 app[web.1]:    from /app/vendor/bundle/ruby/2.1.0/gems/rack-1.5.2/lib/rack/builder.rb:40:in `parse_file'
2014-09-19T20:34:31.414708+00:00 app[web.1]:    from /app/vendor/bundle/ruby/2.1.0/gems/jekyll-2.4.0/lib/jekyll/site.rb:143:in `read_directories'
2014-09-19T20:34:31.414727+00:00 app[web.1]:    from /app/vendor/bundle/ruby/2.1.0/gems/jekyll-2.4.0/lib/jekyll/site.rb:143:in `read_directories'
[...]

PR #6128, opened 2017-06-09
Document a reproducable error and its solution

Kaligule:
The problem is well known and was dicussed at  #2938 #5564 #5361.

See also [github/pages-gem#352](https://github.com/github/pages-gem/pull/352)

In #6127 @DirtyF suggested a pullrequest if I thought there was a better way of warning users. This is the requested pullrequest.

This would have helped me when I tried Jekyll the first time and used the quickstarting guide.

### DIFF
--- a/docs/_docs/quickstart.md
+++ b/docs/_docs/quickstart.md
@@ -25,7 +25,7 @@ If you already have a full [Ruby](https://www.ruby-lang.org/en/downloads/) devel
 If you encounter an error in building the site, with the error message
 "'0000-00-00-welcome-to-jekyll.markdown.erb' does not have a valid date in the
 YAML front matter." try including the line `exclude: [vendor]`
-in `_config.yml` (This is also documented at [troubleshooting](/docs/troubleshooting/#configuration-problems)).
+in `_config.yml` (This is also documented on the [troubleshooting](/docs/troubleshooting/#configuration-problems) page).
 For any other unexpected errors during the above, please refer to the already-mentioned [requirements](/docs/installation/#requirements) page, as you might be missing development headers or other prerequisites.
 
 ## About Bundler
--- a/docs/_docs/quickstart.md
+++ b/docs/_docs/quickstart.md
@@ -22,7 +22,11 @@ If you already have a full [Ruby](https://www.ruby-lang.org/en/downloads/) devel
 # Now browse to http://localhost:4000
 ```
 
-If you encounter any unexpected errors during the above, please refer to the already-mentioned [requirements](/docs/installation/#requirements) page, as you might be missing development headers or other prerequisites.
+If you encounter an error in building the site, with the error message
+"'0000-00-00-welcome-to-jekyll.markdown.erb' does not have a valid date in the
+YAML front matter." try including the line `exclude: [vendor]`
+in `_config.yml` (This is also documented at [troubleshooting](/docs/troubleshooting/#configuration-problems)).
+For any other unexpected errors during the above, please refer to the already-mentioned [requirements](/docs/installation/#requirements) page, as you might be missing development headers or other prerequisites.
 
 ## About Bundler
 


Kaligule:
Not sure how a test can fail if all I added was 2 sentences in the documentation. Any suggestions?

pathawks:
Failure has nothing to do with you :+1:

/cc: @jekyll/documentation

docs/_docs/quickstart.md
```diff
@@ -22,7 +22,11 @@ If you already have a full [Ruby](https://www.ruby-lang.org/en/downloads/) devel
 # Now browse to http://localhost:4000
 ```
 
-If you encounter any unexpected errors during the above, please refer to the already-mentioned [requirements](/docs/installation/#requirements) page, as you might be missing development headers or other prerequisites.
+If you encounter an error in building the site, with the error message
+"'0000-00-00-welcome-to-jekyll.markdown.erb' does not have a valid date in the
+YAML front matter." try including the line `exclude: [vendor]`
+in `_config.yml` (This is also documented at [troubleshooting](/docs/troubleshooting/#configuration-problems)).
```
pup: not sure about the wording of `at [page name]`. maybe `on the [page name] page`?

Kaligule: You are right. I am not a nativ english speaker, so this stuff happens. I corrected the wording.

docs/_docs/quickstart.md
```diff
@@ -22,7 +22,11 @@ If you already have a full [Ruby](https://www.ruby-lang.org/en/downloads/) devel
 # Now browse to http://localhost:4000
 ```
 
-If you encounter any unexpected errors during the above, please refer to the already-mentioned [requirements](/docs/installation/#requirements) page, as you might be missing development headers or other prerequisites.
+If you encounter an error in building the site, with the error message
+"'0000-00-00-welcome-to-jekyll.markdown.erb' does not have a valid date in the
```
DirtyF: Please include the complete error message here for users who copy the message in a search engine and remove the `.erb` extension on the post file.

```sh
Invalid date '<%= Time.now.strftime('%Y-%m-%d %H:%M:%S %z') %>': Document '/_posts/0000-00-00-welcome-to-jekyll.markdown' does not have a valid date in the YAML front matter.
```

Kaligule: I don't think the full errormessage has a place in the quickstarting guide. I think it should be documented on the troubleshooting page.

I don't know what the `.erb` means, but it was in original errormessage I got, so I think it should stay there.

Kaligule: @DirtyF Any thoughts on that?

DirtyF: 👍  about including the error message in the troubleshooting page

👎  about keeping the `.erb` extension, as Jekyll does not support ERB templating. It's rather something odd about your setup.

pathawks: @dirtyf That's the point. Some how, the Jekyll source code is being included in his `vendor/` directory. That's where the `.erb` comes from, and is the source of all this trouble.

ashmaroli: [This](https://github.com/jekyll/jekyll/tree/v3.4.3/lib/site_template/_posts) is the source of the offending ERB file in @Kaligule's `/vendor/` directory just in case you need additional corroboration..

docs/_docs/quickstart.md
```diff
@@ -22,7 +22,11 @@ If you already have a full [Ruby](https://www.ruby-lang.org/en/downloads/) devel
...
+YAML front matter." try including the line `exclude: [vendor]`
+in `_config.yml` (This is also documented on the [troubleshooting](/docs/troubleshooting/#configuration-problems) page).
```
ashmaroli: This should ideally be `in the [troubleshooting]` IMO..

docs/_docs/quickstart.md
DirtyF: @pathawks oh I see, didn't think of that 😲  thanks for the clarification

Kaligule: At this point I don't understand what you are talking about anymore. Should I do remove the `.erb` extension?

docs/_docs/quickstart.md
Kaligule: It is from the troubleshooting page. But following the steps from the "quickstart" manual leads to this error reproducibly, so either the steps at "quickstart" should be changed or it should be documented there.

Otherwise we can get rid of the "quickstart" page all together, since following it reqires bughunting, googling and reading the rest of the manual anyways.

ashmaroli:
@Kaligule I'll try my best to make things clear for you:

When a User runs `jekyll new [name]`, Jekyll creates a ready-to-publish site at `[name]` using a few template files. One of the template file is `_posts/0000-00-00-welcome-to-jekyll.markdown.erb` which gets converted into `_posts/2017-06-01-welcome-to-jekyll.markdown` if the current date is `Jun 01 2017`.

Till v3.4.3 the template used to create the `_config.yml` for the new site contained the following:
```yaml
exclude:
  - Gemfile
  - Gemfile.lock
```
which effectively overrode our default internal setting to ignore `[vendor/*]`.

So when a user has a Jekyll gem in his `vendor` directory, Jekyll tries to process the `_posts/2017-06-01-welcome-to-jekyll.markdown` **"template"** within the Jekyll gem and raises the issue you are facing.

This has been fixed properly on the `master` branch now and from the next release onward this issue wont occur out-of-the-box.

Regarding this PR, there are two things to note:
  - The Quickstart Page should remain as it is. Additional link to the section in the Troubleshooting page is fine. But the prime focus in this page is the steps that document the installation of gems `Jekyll`, `Bundler`, generating a new Jekyll Site and then previewing it locally.
  - The section in the Troubleshooting page can be expanded slightly to explain the scenario and necessary solution: **The `exclude:` array in the configuration should contain `vendor` and/or its subdirs.**
i.e.,
if the current `exclude:` array in the users config file is as below,
```yaml
exclude:
  - Gemfile
  - Gemfile.lock
```
they should update it to:
```yaml
exclude:
  - Gemfile
  - Gemfile.lock
  - vendor
```
or if there is no `exclude:` key in their `_config.yml` at all, (possibly generated ages ago), then they need to simply add the array to their config file viz. either `exclude: [vendor]` or
```yaml
exclude:
  - vendor
```

Kaligule:
This was helpfull, thank you.
I am going to close this pullreqest then. A new one will be created where the troubleshooting page is simply linked.

Kaligule:
Created pullrequest #6134 like said above. Was merged. Thanks for all the explaining, everyone.

PR #6129, opened 2017-06-09
Start from user's home, not a system directory

diev:
When this manual used to install Jekyll on Bash run in *Windows PowerShell (Administrator)*, the current directory is `Windows\System32`. It is not a proper place to create `my_project` in. So please be sure you start from your home, not a system directory.

### DIFF
--- a/docs/_docs/windows.md
+++ b/docs/_docs/windows.md
@@ -4,7 +4,7 @@ permalink: /docs/windows/
 ---
 
 While Windows is not an officially-supported platform, it can be used to run
-Jekyll with the proper tweaks. If you are using Windows 10 Anniversary Update, 
+Jekyll with the proper tweaks. If you are using Windows 10 Anniversary Update (1607) or newer, 
 the easiest way to run Jekyll is to use the new [Bash on Ubuntu on Windows](https://msdn.microsoft.com/en-us/commandline/wsl/install_guide).
 For older installations, this page aims to collect some of the general knowledge and lessons that have been unearthed by Windows users.
 
@@ -45,6 +45,7 @@ sudo gem install jekyll bundler
 You can test by running:
 
 ```
+cd ~
 jekyll new my_project
 ```
 


docs/_docs/windows.md
```diff
@@ -4,7 +4,7 @@ permalink: /docs/windows/
 ---
 
 While Windows is not an officially-supported platform, it can be used to run
-Jekyll with the proper tweaks. If you are using Windows 10 Anniversary Update, 
+Jekyll with the proper tweaks. If you are using Windows 10 Anniversary Update (1607) or newer,
```
pup: is this relevant? it's another (albeit optional) step and besides, wouldn't this be irrelevant if you `cd ~` anyways?

XhmikosR:
All this seems futile to me; it's like telling a user how to use cmd...

pup:
@XhmikosR i would argue that it's very relevant, seeing as this is a frequent problem new users come across. not everyone knows how to use the command line, you know :wink:

parkr:
Ideally we wouldn't actually teach people how to use the command line. It's an expected prerequisite. I'd be happy to link out to external resources, but I wouldn't explain it ourselves.

docs/_docs/windows.md
```diff
@@ -45,6 +45,7 @@ sudo gem install jekyll bundler
 You can test by running:
 
 ```
+cd ~
```
parkr: Would it be easier to simply add a comment here saying that you should choose which directory you run the command in?

docs/_docs/windows.md
diev: When I used the previous Win10 AU (1607), it used the well-known `cmd`. Now Win10 CU (1703) uses `powershell` instead. And when I was installing *as Administrator*, I had found myself to be in the system directory during my performing of those steps.
Again, just mentioning *AU* only makes a future problem for users to understand that it is at all. So my collegues are not smart to know about the number of releases of Win10 for the moment.

docs/_docs/windows.md
diev: New users (as I have seeing them) want to perform the steps of some manual to get a just working environment. A professional one will pass `cd ~` as a known step (or choose the directory as he prefers). But a novice will pass that comment text just going step by step in the `code` frames only. Of course the good knowledge of the English language is also an expected prerequisite for you. Please sorry for mine.

DirtyF:
Documentation has been updated since, closing this. Thanks for your effort.

PR #6134, opened 2017-06-14
Link the troubleshooting page in the quickstart page

Kaligule:
This is a follow up to pull request #6128. The original problem is said to be fixed on master and there is no point in explaining and solving the same bug twice. A link for those who have problems should be nice, though.


pup:
@jekyllbot: merge +docs

PR #6136, opened 2017-06-14
Elaborate on excluding items from processing

ashmaroli:
In light of recent discussions on the `exclude:` configuration, this PR elaborates the troubleshooting tip on that matter

/cc @jekyll/documentation

### DIFF
--- a/docs/_docs/troubleshooting.md
+++ b/docs/_docs/troubleshooting.md
@@ -195,10 +195,37 @@ That is: defaults are overridden by options specified in `_config.yml`,
 and flags specified at the command-line will override all other settings
 specified elsewhere.
 
-If you encounter an error in building the site, with the error message
-"'0000-00-00-welcome-to-jekyll.markdown.erb' does not have a valid date in the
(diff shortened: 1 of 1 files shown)


ashmaroli:
not very happy with the flow of the text here.. feel free to edit as required.. :wink:

DirtyF:
IMHO It's typically the kind of explanation that belongs to a release post.

ashmaroli:
> It's typically the kind of explanation that belongs to a release post.

I see.. but an average user won't consult a release-post when trying to solve a particular problem..

DirtyF:
>  an average user won't consult a release-post when trying to solve a particular problem

Chances are people will use a search engine when they encounter a problem, and then hopefully read the release post who explains the changes as it's one of the fe documents indexed.

My point is we can not keep _all_ changes related to a previous version in the docs, it could become unreadable. AFAIC documentation should reflect the current state of the software. That's why I'd rather point to changelog or release posts to explain minor changes.

DirtyF:
Let's hope this saves some headache to users.

@jekyllbot: merge +docs

PR #6137, opened 2017-06-14
Default `baseurl` to `nil` instead of empty string

parkr:
Using an empty string as the default value causes all sorts of problems where you don't know if this is user-specified or not. Using an explicitly "illegal" value, like `nil`, resolves this issue.

See https://github.com/jekyll/github-metadata/pull/97 for more info.

/cc @jekyll/build @benbalter


parkr:
@jekyllbot: merge +bug

PR #6140, opened 2017-06-15
Lock nokogiri to 1.7.x for Ruby 2.1

parkr:
Fails our AppVeyor tests: https://ci.appveyor.com/project/jekyll/jekyll/build/1722/job/3vlp2o2i69uw6n2p

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,11 +22,8 @@ group :test do
   gem "cucumber", "~> 2.1"
   gem "jekyll_test_plugin"
   gem "jekyll_test_plugin_malicious"
-  if RUBY_VERSION > "2.1"
-    gem "nokogiri"
-  else
-    gem "nokogiri", "~> 1.7.0"
-  end
+  # nokogiri v1.8 does not work with ruby 2.1 and below
+  gem "nokogiri", RUBY_VERSION > "2.1" ? "~> 1.7" : "~> 1.7.0"
   gem "rspec"
   gem "rspec-mocks"
   gem "rubocop", "~> 0.48.1"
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -22,7 +22,8 @@ group :test do
   gem "cucumber", "~> 2.1"
   gem "jekyll_test_plugin"
   gem "jekyll_test_plugin_malicious"
-  gem "nokogiri"
+  # nokogiri v1.8 does not work with ruby 2.1 and below
+  gem "nokogiri", RUBY_VERSION > "2.1" ? "~> 1.7" : "~> 1.7.0"
```
DirtyF: Shouldn't it be something like:
`gem "nokogiri",  RUBY_VERSION < "2.1" ? "1.6" : "1.7"`

ashmaroli: no @DirtyF , it should instead be:
```ruby
gem "nokogiri", RUBY_VERSION >= "2.2" ? "~> 1.7" : "~> 1.7.0"
```
`> 2.1` will include `2.1.1`, `2.1.2` etc

pathawks: `> "2.1"` is equivalent to `>= "2.2"`

?

ashmaroli: no it isn't

ashmaroli: `> 2.1` will include `2.1.1`, `2.1.2`... `2.2.x`, `2.3.x`, `2.4.x`
`>= 2.2` will include only `2.2.0`, `2.2.1`,..., `2.4.x`

pathawks: Is it different practically or only semantically? 😕

ashmaroli: both.. Appveyor still fails with the last commit because, its still installing `1.8.0` instead..

pathawks: TIL :beers:

parkr:
💚 

@jekyllbot: merge +dev

PR #6141, opened 2017-06-15
docs(JekyllOnWindows): Add a new Installation way

KeJunMao:
Installation via RubyInstaller

### DIFF
--- a/docs/_docs/windows.md
+++ b/docs/_docs/windows.md
@@ -191,13 +191,6 @@ Please add the following to your Gemfile to avoid polling for changes:
     gem 'wdm', '>= 0.1.0' if Gem.win_platform?
 ```
 
-But auto-regeneration still works fine.Its an outdated warning IMO.
+Auto-regeneration will work fine without including `gem 'wdm'`
 
-### time-zone
-
-Requires one gem:
-
-```shell
-gem install tzinfo-data
-```
-
+### [time-zone](/docs/windows/#timezone-management)
--- a/docs/_docs/windows.md
+++ b/docs/_docs/windows.md
@@ -177,7 +177,27 @@ In the future the installation process of the github-pages should be as simple a
 RubyInstaller is a self-contained Windows-based installer that includes the Ruby language, an execution environment, important documentation, and more.
 
 1. Install a package manager for Windows called [RubyInstaller](https://rubyinstaller.org/).
-2. Install Jekyll via RubyInstaller: `gem install bundler jekyll`
-3. Reopen a command prompt and run Jekyll: `jekyll -v`
+2. Install Jekyll and Bundler via a command prompt window: `gem install jekyll bundler`
+3. Check if the installation is accessible: `jekyll -v`
 
 see [Autoinstall Jekyll for windows](https://github.com/KeJunMao/fastjekyll#autoinstall-jekyll-for-windows)
+
+### auto-regeneration
+
+Although jekyll would suggest:
+
+```
+Please add the following to your Gemfile to avoid polling for changes:
+    gem 'wdm', '>= 0.1.0' if Gem.win_platform?
+```
+
+But auto-regeneration still works fine.Its an outdated warning IMO.
+
+### time-zone
+
+Requires one gem:
+
+```shell
+gem install tzinfo-data
+```
+
--- a/docs/_docs/windows.md
+++ b/docs/_docs/windows.md
@@ -171,3 +171,13 @@ In the future the installation process of the github-pages should be as simple a
 [Bundler]: http://bundler.io/ "Ruby Dependencie Manager"
 [nokogiriReleases]: https://github.com/sparklemotion/nokogiri/releases "Nokogiri Releases"
 [nokogiriFails]: https://github.com/sparklemotion/nokogiri/issues/1456#issuecomment-206481794 "Nokogiri fails to install on Ruby 2.3 for Windows"
+
+## Installation via RubyInstaller
+
+RubyInstaller is a self-contained Windows-based installer that includes the Ruby language, an execution environment, important documentation, and more.
+
+1. Install a package manager for Windows called [RubyInstaller](https://rubyinstaller.org/).
+2. Install Jekyll via RubyInstaller: `gem install bundler jekyll`
+3. Reopen a command prompt and run Jekyll: `jekyll -v`
+
+see [Autoinstall Jekyll for windows](https://github.com/KeJunMao/fastjekyll#autoinstall-jekyll-for-windows)
--- a/docs/_docs/windows.md
+++ b/docs/_docs/windows.md
@@ -180,9 +180,9 @@ RubyInstaller is a self-contained Windows-based installer that includes the Ruby
 2. Install Jekyll and Bundler via a command prompt window: `gem install jekyll bundler`
 3. Check if the installation is accessible: `jekyll -v`
 
-see [Autoinstall Jekyll for windows](https://github.com/KeJunMao/fastjekyll#autoinstall-jekyll-for-windows)
+See [Autoinstall Jekyll for Windows](https://github.com/KeJunMao/fastjekyll#autoinstall-jekyll-for-windows)
 
-### auto-regeneration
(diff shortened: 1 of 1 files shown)


HenryKobin:
/cc: @jekyllbot @jekyll/documentation

By chance have you tested auto-regeneration or time-zone management to see if they are functional if you install via this method?

docs/_docs/windows.md
```diff
@@ -171,3 +171,13 @@ In the future the installation process of the github-pages should be as simple a
 [Bundler]: http://bundler.io/ "Ruby Dependencie Manager"
 [nokogiriReleases]: https://github.com/sparklemotion/nokogiri/releases "Nokogiri Releases"
 [nokogiriFails]: https://github.com/sparklemotion/nokogiri/issues/1456#issuecomment-206481794 "Nokogiri fails to install on Ruby 2.3 for Windows"
+
+## Installation via RubyInstaller
+
+RubyInstaller is a self-contained Windows-based installer that includes the Ruby language, an execution environment, important documentation, and more.
+
+1. Install a package manager for Windows called [RubyInstaller](https://rubyinstaller.org/).
+2. Install Jekyll via RubyInstaller: `gem install bundler jekyll`
```
ashmaroli: I haven't tried the latest version from RubyInstaller2, but does the Installer itself have an environment to accept shell commands?
I feel L#180 and L#181 should rather be:
```
2. Install Jekyll and Bundler via a command prompt window: `gem install jekyll bundler`
3. Check if the installation is accessible: `jekyll -v`
```

KeJunMao:
@HenryKobin 

Yes!

# auto-regeneration

Although jekyll would suggest:

```
Please add the following to your Gemfile to avoid polling for changes:
    gem 'wdm', '>= 0.1.0' if Gem.win_platform?
```

But auto-regeneration still works fine.

# timezone
Requires two gems:
```
gem install tzinfo
gem install tzinfo-data
```

ashmaroli:
> Requires two gems:

No, actually, only requires `tzinfo-data`. `tzinfo` is automatically included with the former.

docs/_docs/windows.md
KeJunMao: Well, it makes sense.

HenryKobin:
@KeJunMao Could you add that part about time-zone management and auto-regeneration to the end of your tutorial, please? Thank you, by the way! 👍

ashmaroli:
Note: `gem 'wdm'` is not actually required for regeneration to work on Windows. Its an outdated warning IMO,

KeJunMao:
@HenryKobin Thanks!

docs/_docs/windows.md
```diff
@@ -177,7 +177,27 @@ In the future the installation process of the github-pages should be as simple a
...
+
+Although jekyll would suggest:
+
+```
+Please add the following to your Gemfile to avoid polling for changes:
+    gem 'wdm', '>= 0.1.0' if Gem.win_platform?
+```
+
+But auto-regeneration still works fine.Its an outdated warning IMO.
+
```
ashmaroli: Although and but don't play well together. Please update L#194 to:
```
Auto-regeneration will work fine without including `gem 'wdm'`
```

docs/_docs/windows.md
```diff
@@ -177,7 +177,27 @@ In the future the installation process of the github-pages should be as simple a
...
+### time-zone
+
+Requires one gem:
+
+```shell
+gem install tzinfo-data
```
ashmaroli: Though `tzinfo-data` is required, you don't have to install it separately.
It is already included in a new Gemfile generated by `jekyll new` from `v3.4.0` For sites generated with older versions, they **should** add it to their Gemfile.
```
# Windows does not include zoneinfo files, so bundle the tzinfo-data gem
gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
```
It was documented under `https://jekyllrb.com/docs/windows/#timezone-management` but seems to have lost the ref tag..

ashmaroli: > It was documented under https://jekyllrb.com/docs/windows/#timezone-management but seems to have lost the ref tag..

/cc @DirtyF Please fix on `master`

KeJunMao: Change to: ?

### auto-regeneration

Although jekyll would suggest:

```
Please add the following to your Gemfile to avoid polling for changes:
    gem 'wdm', '>= 0.1.0' if Gem.win_platform?
```

Auto-regeneration will work fine without including `gem 'wdm'`

### time-zone

Though `tzinfo-data` is required, you don't have to install it separately.
It is already included in a new Gemfile generated by `jekyll new` from `v3.4.0` For sites generated with older versions, they **should** add it to their Gemfile.

```
# Windows does not include zoneinfo files, so bundle the tzinfo-data gem
gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
```

ashmaroli: I missed to point out that the user will have to run `bundle install` after they update their Gemfile.
But since this entire section on Timezone Management is already documented under `/docs/windows/#timezone-management`, simply link to that section.

(The link is currently broken. Will be fixed by maintainers separately)

KeJunMao: @ashmaroli 

Change to: ?

### auto-regeneration

Although jekyll would suggest:

```
Please add the following to your Gemfile to avoid polling for changes:
    gem 'wdm', '>= 0.1.0' if Gem.win_platform?
```

Auto-regeneration will work fine without including `gem 'wdm'`

### time-zone

Though `tzinfo-data` is required, you don't have to install it separately.
It is already included in a new Gemfile generated by `jekyll new` from `v3.4.0` For sites generated with older versions, they **should** add it to their Gemfile,Then run the `bundle install`.

```
# Windows does not include zoneinfo files, so bundle the tzinfo-data gem
gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
```

More about timezone-management see [here](/docs/windows/#timezone-management).

ashmaroli: I would leave out the entire section, but let others weigh in on this..

docs/_docs/windows.md
```diff
@@ -171,3 +171,26 @@ In the future the installation process of the github-pages should be as simple a
...
+
+## Installation via RubyInstaller
+
+RubyInstaller is a self-contained Windows-based installer that includes the Ruby language, an execution environment, important documentation, and more.
+
+1. Install a package manager for Windows called [RubyInstaller](https://rubyinstaller.org/).
+2. Install Jekyll and Bundler via a command prompt window: `gem install jekyll bundler`
+3. Check if the installation is accessible: `jekyll -v`
+
+see [Autoinstall Jekyll for windows](https://github.com/KeJunMao/fastjekyll#autoinstall-jekyll-for-windows)
```
parkr: Capital S here, please, i.e. `See`. Same for the `W` in `Autoinstall Jekyll for Windows`.

docs/_docs/windows.md
```diff
@@ -171,3 +171,26 @@ In the future the installation process of the github-pages should be as simple a
...
+
+### auto-regeneration
```
parkr: Capital A here, please.

docs/_docs/windows.md
```diff
@@ -171,3 +171,26 @@ In the future the installation process of the github-pages should be as simple a
...
+Although jekyll would suggest:
+
+```
+Please add the following to your Gemfile to avoid polling for changes:
+    gem 'wdm', '>= 0.1.0' if Gem.win_platform?
+```
+
+Auto-regeneration will work fine without including `gem 'wdm'`
+
+### [time-zone](/docs/windows/#timezone-management)
```
parkr: Did you want to elaborate on this at all or just head a dangling header?

KeJunMao: I just want to dangling a header.

KeJunMao: or :

pup:
@jekyllbot: merge +docs

PR #6145, opened 2017-06-15
Backport #6137 for v3.4.x: Default `baseurl` to `nil` instead of empty string

parkr:
This backports #6137.

Conflicting files:


parkr:
Hm.

PR #6147, opened 2017-06-15
add documentation about the "pinned" label

pup:
i noticed that documentation about this label was missing. are there any other undocumented special labels?


pup:
cc @jekyll/documentation

DirtyF:
@pup according to the source code `security`and `has-pull-request` also prevent an issue from being marked as `stale`

https://github.com/parkr/auto-reply/blob/f1933491fbda8e521572d7695dbe2bbc623e111c/cmd/mark-and-sweep-stale-issues/mark_and_sweep_stale_issues.go#L22-L25

pup:
@DirtyF huh! the question is, do we want to list those as viable alternatives that could be used instead of `pinned`? i think it might cause some confusion, especially with `security`

pup:
i'll add a note that `has-pull-request` has the same effect, but shouldn't be used singularly to achieve it

DirtyF:
@pup `has-pull-request`is often set automatically when mentioning `fix`and isssue number in a PR comment, so it's not a label I use very much.

DirtyF:
@jekyllbot: merge +docs

PR #6148, opened 2017-06-16
added header tags to Chocolatey tutorial

HenryKobin:
Added header tags to Time-Zone management and Auto Regeneration in the Installing Jekyll via Chocolatey on Windows tutorial.

cc/ @jekyllbot @jekyll/documentation

### DIFF
--- a/docs/_docs/windows.md
+++ b/docs/_docs/windows.md
@@ -81,6 +81,7 @@ the site generation process. It can be done with the following command:
 ```sh
 $ chcp 65001
 ```
+### Time-Zone Management
 Since Windows doesn't have a native source of zoneinfo data, the Ruby Interpreter would not understand IANA Timezones and hence using them had the `TZ` environment variable default to UTC/GMT 00:00.
 Though Windows users could alternatively define their blog's timezone by setting the key to use POSIX format of defining timezones, it wasn't as user-friendly when it came to having the clock altered to changing DST-rules.
 
@@ -91,7 +92,7 @@ While 'new' blogs created with Jekyll v3.4 and greater, will have the following 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
 ```
-
+### Auto Regeneration
 As of v1.3.0, Jekyll uses the `listen` gem to watch for changes when the
 `--watch` switch is specified during a build or serve. While `listen` has
 built-in support for UNIX systems, it requires an extra gem for compatibility


ashmaroli:
@HenryKobin Those sections apply to all forms of installation on Windows and not just via Chocolatey..

DirtyF:
@ashmaroli If it's the case then we should move both at the end after the Chocolatey section.

HenryKobin:
@dirtyf that isn't the case. At least not my bash on windows method. But do what you guys think is right.

ashmaroli:
> my bash on windows method

Awesome! I did not know that Windows ship timezone info with the BASH shell as well.. Great to know!!
In that case the section needs a little tweaking....

HenryKobin:
@ashmaroli  I just want to confirm about this part in the auto-regeneration section under the Chocolatey tutorial 
`As of v1.3.0, Jekyll uses the `listen` gem to watch for changes when the
`--watch` switch is specified during a build or serve. While `listen` has
built-in support for UNIX systems, it requires an extra gem for compatibility
with Windows. Add the following to the Gemfile for your site:
`
is this saying that it does work out of the box if you have v 1.3.0+ of Jekyll?
Just wanted to make sure. Because if you install via Bash on Windows, it just works out of the box. You don't need to add any gemfiles. I might be misunderstanding, though.

ashmaroli:
> that it does work out of the box if you have v 1.3.0+ of Jekyll?

What it means is Jekyll started "watching" the file tree for changes from v1.3.0 of Jekyll and according to the authors of the concerned gem, `listen', it will work as advertised out-of-the box for UNIX-like systems but Windows (at that time) needed an additional gem.
But I currently run Jekyll on Windows 7 `cmd.exe` with Ruby installed via RubyInstaller and have no issues without including `gem 'wdm'`. So, my best guess is, its no longer needed for any method of Ruby installation.. but I could be wrong.. I even opened a ticket at the `listen` repo to confirm.. got no response though..

HenryKobin:
@ashmaroli  Interesting. Yeah i'm not too sure what to do in this case.  Would we perhaps want to omit these sections entirely, or just elaborate on each install method what has to be done to make the features work (since they may just work out of the box with any windows tutorial now).
@dirtyF it seem's like we may no longer need either of these sections. What do? Maybe just wait until the `listen` ticket gets responded to?

ashmaroli:
> Would we perhaps want to omit these sections entirely

No. The sections need to remain but be updated based on relevance. The `wdm` gem *may be* necessary for an environment neither of us have tested on..
The Timezone section definitely needs to stay because apart from `Bash on Windows` no other method will include the required timezone data for the `module` to work properly.

ashmaroli:
> Would we perhaps want to omit these sections entirely

No. The sections need to remain but be updated based on relevance. The `wdm` gem `may be` necessary for an environment neither of us have tested on..
The Timezone section definitely needs to stay because apart from `Bash on Windows` no other method will include the required timezone data for the `module` to work properly.

HenryKobin:
@ashmaroli I just saw this AFTER I saw your comment about saying "Why? didn't we say this applies to all of them." my bad!

PR #6150, opened 2017-06-18
[Rouge 2] use `Rouge::Formatters::HTMLLegacy` and fix tests

Crunch09:
Hey,
i changed `Formatters::HTML` for `Formatters::HTMLLegacy` based on @jneen 's suggestion (https://github.com/jekyll/jekyll/pull/5919#issuecomment-308536849) and because `Formatters::HTML` (the rouge 2 version) doesn't print line numbers (unless i missed something).

**Note**: This targets #5919 instead of `master`. Will close #5230 in favor of this.

### DIFF
--- a/jekyll.gemspec
+++ b/jekyll.gemspec
@@ -38,6 +38,6 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency("liquid",                "~> 4.0")
   s.add_runtime_dependency("mercenary",             "~> 0.3.3")
   s.add_runtime_dependency("pathutil",              "~> 0.9")
-  s.add_runtime_dependency("rouge",                 "~> #{ENV["ROUGE_VERSION"] || "2.1.0"}")
+  s.add_runtime_dependency("rouge",                 "~> #{ENV["ROUGE_VERSION"] || "2.1"}")
   s.add_runtime_dependency("safe_yaml",             "~> 1.0")
 end
--- a/jekyll.gemspec
+++ b/jekyll.gemspec
@@ -38,6 +38,6 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency("liquid",                "~> 4.0")
   s.add_runtime_dependency("mercenary",             "~> 0.3.3")
   s.add_runtime_dependency("pathutil",              "~> 0.9")
-  s.add_runtime_dependency("rouge",                 "~> #{ENV["ROUGE_VERSION"] || "1.7"}")
+  s.add_runtime_dependency("rouge",                 "~> #{ENV["ROUGE_VERSION"] || "2.1.0"}")
   s.add_runtime_dependency("safe_yaml",             "~> 1.0")
 end
--- a/lib/jekyll/utils/rouge.rb
+++ b/lib/jekyll/utils/rouge.rb
@@ -4,16 +4,16 @@ module Jekyll
 
       def self.html_formatter(*args)
         Jekyll::External.require_with_graceful_fail("rouge")
-        html_formatter = ::Rouge::Formatters::HTML.new(*args)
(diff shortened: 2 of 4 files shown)


jekyll.gemspec
```diff
@@ -38,6 +38,6 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency("liquid",                "~> 4.0")
   s.add_runtime_dependency("mercenary",             "~> 0.3.3")
   s.add_runtime_dependency("pathutil",              "~> 0.9")
-  s.add_runtime_dependency("rouge",                 "~> #{ENV["ROUGE_VERSION"] || "1.7"}")
+  s.add_runtime_dependency("rouge",                 "~> #{ENV["ROUGE_VERSION"] || "2.1.0"}")
```
parkr: Does this work with 2.0 & 2.2, or is only 2.1.x compatible? Usually we'd just use 2 digits here, e.g. 2.0.

lib/jekyll/utils/rouge.rb
```diff
@@ -4,16 +4,16 @@ module Rouge
...
       def self.html_formatter(*args)
         Jekyll::External.require_with_graceful_fail("rouge")
-        html_formatter = ::Rouge::Formatters::HTML.new(*args)
-        return html_formatter if old_api?
-
-        ::Rouge::Formatters::HTMLPygments.new(html_formatter)
+        if old_api?
+          ::Rouge::Formatters::HTML.new(*args)
+        else
+          ::Rouge::Formatters::HTMLLegacy.new(*args)
```
parkr: What does the new API look like? is this just a wrapper around the new functionality with the old API?

jekyll.gemspec
Crunch09: 👍 It should also work with any 2.x version, i'll update!

lib/jekyll/utils/rouge.rb
Crunch09: Yes, exactly. It produces (almost) the same output as rouge 1.11 while receiving the same arguments.

Crunch09: [Here](https://github.com/jneen/rouge#formatters) are all default formatters listed.  IMHO line numbers (which we rely on in a couple of tests) would also be present in `Rouge::Formatters::HTMLTable` and `Rouge::Formatters::HTMLLinewise` but `HTMLLegacy` should be the easiest update.

parkr: Does HTMLLegacy not have those features?

Crunch09: Oh, sorry, this was unclear: HTMLLegacy also has line numbers and all features we currently use.
The 2.x version of the standard `HTML` formatter doesn't have line numbers.

jekyll.gemspec
```diff
@@ -38,6 +38,6 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency("liquid",                "~> 4.0")
   s.add_runtime_dependency("mercenary",             "~> 0.3.3")
   s.add_runtime_dependency("pathutil",              "~> 0.9")
-  s.add_runtime_dependency("rouge",                 "~> #{ENV["ROUGE_VERSION"] || "1.7"}")
+  s.add_runtime_dependency("rouge",                 "~> #{ENV["ROUGE_VERSION"] || "2.1"}")
```
envygeeks: This is a big wonky IMO.  You could probably do something like:

```ruby
s.add_runtime_dependency "rogue", ">= 1.7", "< 2.2"
```

And have a much easier time.

parkr:
I'll fix up the dependency in the parent PR. Thank you!!!

PR #6151, opened 2017-06-18
Added direct collection access to future collection item feature test

BrandonDusseau:
Further tests #5953 
/cc @jekyll/build

### DIFF
--- a/features/collections.feature
+++ b/features/collections.feature
@@ -128,16 +128,18 @@ Feature: Collections
       puppies:
         output: false
     """
-    And I have a "foo.txt" file that contains "random static file"
+    And I have a "index.html" page that contains "Newest puppy: {% assign puppy = site.puppies.last %}{{ puppy.title }}"
     When I run jekyll build
     Then I should get a zero exit status
     And the _site directory should exist
     And the "_site/puppies/rover.html" file should not exist
     And the "_site/puppies/fido.html" file should not exist
+    And I should see "Newest puppy: Rover" in "_site/index.html"
     When I run jekyll build --future
     Then I should get a zero exit status
     And the _site directory should exist
     And the "_site/puppies/fido.html" file should not exist
+    And I should see "Newest puppy: Fido" in "_site/index.html"
 
   Scenario: All the documents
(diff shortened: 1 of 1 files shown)


BrandonDusseau:
@parkr It would have effectively made a duplicate of the test I modified. Since the scenario described matched the situation I was testing, I modified it. I can split it out into another test if you'd like.

features/collections.feature
```diff
@@ -128,16 +128,18 @@ Feature: Collections
       puppies:
         output: false
     """
-    And I have a "foo.txt" file that contains "random static file"
+    And I have a "index.html" page that contains "Newest puppy: {% assign puppy = site.puppies.last %}{{ puppy.title }}"
```
parkr: I really thought `site.puppies.first` would be the newest.

BrandonDusseau: From my experience, apparently that is not the case. The documentation doesn't seem to specify.

BrandonDusseau:
Split this into a separate test as previously suggested.

features/collections.feature
```diff
@@ -139,6 +139,28 @@ Feature: Collections
     And the _site directory should exist
     And the "_site/puppies/fido.html" file should not exist
 
+  Scenario: Hidden collection with document with future date, accessed via Liquid
```
mattr-: This is nitpicky, but could you change this to `Hidden collection has document with future date, accessed via Liquid`

I had trouble parsing the scenario description until I read the full test.

BrandonDusseau: Sure, no problem. I was just keeping with the phrasing on the prior test. That'll be fixed in just a sec.

mattr-:
Thanks! Appreciate you adding this test! 👍

mattr-:
@jekyllbot: merge +dev

parkr:
@mattr- I believe this is a failing test showing a behavior that isn’t working right. I see master as currently failing cucumber now on all ruby versions. Would you mind sending a patch to fix master and retry this patch with the proper fix to make the test pass?

mattr-:
Oh! My bad! Why did I think this was passing earlier? 😞

ashmaroli:
@mattr- do you have the time right now, or should I send the PR..?

PR #6156, opened 2017-06-19
Fix release date in History

DirtyF:
@parkr Is it a human error or a script to update somewhere?


ashmaroli:
> or a script to update somewhere?

The same error occurred  once before.. when `3.4.0` was released.. If its an error in a script or rake file, it occurred only with minor-version  bumps....

parkr:
Proooooobably just me! I add those manually.

crispgm:
Should we add some tests for this? 😄 @parkr

PR #6158, opened 2017-06-19
update config.yml with new plugins key

marcanuy:
As of Jekyll 3.5 the `gems` key in `_config.yml` is `plugins`, make `new` command to use the newer form.


DirtyF:
duplicate #6045

PR #6162, opened 2017-06-20
Static files contain front matter default keys when to_liquid'd 

benbalter:
Trying to track down https://github.com/jekyll/jekyll-sitemap/issues/176, it seems in 3.5, when you call `to_liquid` on a convertible (page, static_file), the resulting hash doesn't contain front matter default keys, although calling a specific key returns the expected value. This means that things like `to_json` or anything that looks for `.key?` will not act as expected.

This PR currently contains a failing test in hopes of finding the root cause (and adding tests for front matter defaults going forward).

/cc @parkr

### DIFF
--- a/lib/jekyll/static_file.rb
+++ b/lib/jekyll/static_file.rb
@@ -28,10 +28,6 @@ module Jekyll
       @collection = collection
       @relative_path = File.join(*[@dir, @name].compact)
       @extname = File.extname(@name)
-
-      data.default_proc = proc do |_, key|
-        site.frontmatter_defaults.find(relative_path, type, key)
-      end
     end
     # rubocop: enable ParameterLists
 
@@ -104,7 +100,7 @@ module Jekyll
     end
 
     def data
-      @data ||= {}
+      @data ||= @site.frontmatter_defaults.all(relative_path, type)
     end
 
     def basename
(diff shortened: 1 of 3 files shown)


parkr:
> the resulting hash doesn't contain front matter default keys, although calling a specific key returns the expected value. This means that things like to_json or anything that looks for .key? will not act as expected.

The key resolves because we inherit the `default_proc`. It makes more sense to get all frontmatter defaults for the given page we're processing and do `Utils.deep_merge_hashes(defaults, frontmatter)`.

benbalter:
> It makes more sense to get all frontmatter defaults for the given page we're processing and do Utils.deep_merge_hashes(defaults, frontmatter).

I thought we were doing that via https://github.com/jekyll/jekyll/blob/master/lib/jekyll/convertible.rb#L122-L123?

parkr:
That looks like it should work for `Page` and `Layout`, but `Document` and `StaticFile` don't use `Convertible`.

lib/jekyll/static_file.rb
```diff
@@ -104,7 +100,7 @@ def to_liquid
     end
 
     def data
-      @data ||= {}
+      @data ||= @site.frontmatter_defaults.all(relative_path, type)
```
parkr: This isn't guaranteed to run if `@data` is set by some other process. Should we just do this in `#initialize`?

benbalter: Good catch. Implemented via e9c691e.

parkr:
@jekyllbot: merge +bug

Issue #4392, opened 2016-01-22
Jekyll 3.0.2 cannot call a collection by name like site.collections.mycollection


djacquel:
In Jekyll 2, you can call a collection with `site.collections["myCollection"]` or `site.collections.myCollection`, and it returns the collection with all attributes (label, docs, ...).

With Jekyll 3.0.2, doing `site.collections["myCollection"]` or `site.collections.myCollection` returns nothing.

Am I missing something ?

parkr:
Hey, @djacquel! Do you happen to have an example? I can't seem to replicate this.

djacquel:
I reproduce it :

  1. On Github Pages with Jekyll 2.4.0 [repo](https://github.com/djacquel/jekyll-collections-test) and [preview](http://pygmeeweb.com/jekyll-collections-test/).
  `{{ site.collections.col }}` returns a collection with all attributes.

  2. locally using ruby 2.1.8 and Jekyll 3.0.2

```
git clone git@github.com:djacquel/jekyll-collections-test.git
cd jekyll-collections-test/
bundle install
bundle exec jekyll serve
```

We clearly see that `site.collections.col` returns nothing.

envygeeks:
Hi @parkr and @djacquel; Hulk here:

My owner has helped and allowed me to pull and debug your Git repository on a secure Docker instance. I have taken your steps, striped them and done as you requested, here are your results:

```
Jekyll version : 3.0.2

site.collections.col

site.collections.col.label

site.collections.col.docs

loop in site.collections.col.docs
```

---

During my trace I noticed that `site.collections` is an `Array` (or behaves like it) and you are trying to access it like a `Hash`.  Here are the relevant sources (I may be wrong):

- https://github.com/jekyll/jekyll/blob/3.0-stable/lib/jekyll/site.rb#L276
- https://github.com/jekyll/jekyll/blob/3.0-stable/lib/jekyll/collection.rb#L168

I noticed that you are using `collections.values`, this will result in an `Array` with Ruby if used on a `Hash`, I am unreliably assuming that this is not an intended behavior based on keywords.

---

Since my owner is constantly banging on about references here are some details:

* My Host: Ubuntu 14.04
* My Kernel: 4.2.0-25-generic
* I used the Docker image `jekyll/jekyll:3.0.2`
  * It uses Alpine Linux 3.3 from envygeeks/alpine

```
[...]

PR #6165, opened 2017-06-21
[docs] Add information about finding a collection in `site.collections`

adamvoss:
Add information about finding a collection in `site.collections`.  Information taken from the "Upgrading from 2.x to 3.x" page which was discovered by reading issue #4392.

### DIFF
--- a/docs/_docs/collections.md
+++ b/docs/_docs/collections.md
@@ -309,6 +309,8 @@ you specified in your `_config.yml` (if present) and the following information:
   you have a <code>_posts</code> directory or not. This is something to note 
   when iterating through <code>site.collections</code> as you may need to 
   filter it out.</p>
+  <p>You may wish to use filters to find your collection:
+  <code>{{ site.collections | where: "label", "myCollection" | first }}</code></p>
 </div>
 
 


docs/_docs/collections.md
```diff
@@ -309,6 +309,8 @@ you specified in your `_config.yml` (if present) and the following information:
   you have a <code>_posts</code> directory or not. This is something to note 
   when iterating through <code>site.collections</code> as you may need to 
   filter it out.</p>
+  <p>You may wish to use filters to find your collection:
+  <code>site.collections | where: "label", "myCollection" | first</code></p>
```
ashmaroli: You forgot the braces..
```diff
- <code>site.collections | where: "label", "myCollection" | first</code>
+ <code>{{ site.collections | where: "label", "myCollection" | first }}</code>
```

adamvoss: Added.  I hadn't been sure if they were desired since the other nearby code blocks (that were just a variable or member access) didn't have them.

DirtyF:
Thanks for the reminder!

@jekyllbot: merge +docs

annegentle:
I can see this addition in the source file, but the example within the `<code>` tags doesn't render properly. Here's a side-by-side screenshot on Chrome Mac OS X.

screen shot 2017-06-24 at 3 59 01 pm

ashmaroli:
> but the example within the `<code>` tags doesn't render properly.

Ah yes, my apologies @annegentle and @adamvoss I should've given a proper suggestion. The line should've actually been:
```diff
- <code>site.collections | where: "label", "myCollection" | first</code>
+ <code>{% raw %}{{ site.collections | where: "label", "myCollection" | first }}{% endraw %}</code>
```
it didnt render on site because Liquid parsed the code above and returned a `falsy`. The `raw` tags tell Liquid to ignore this code.

/cc @DirtyF

adamvoss:
@ashmaroli I probably should have realized that switching it to full Liquid tags would need to be escaped.

I just made the change locally, and seeing it rendered, I am [wondering again](https://github.com/jekyll/jekyll/pull/6165#discussion_r123270228) about whether the code block should be in braces.

image

Putting it in braces, to me, suggests usage advice.  However, outputting the collection is not probably the typical usage.  I suspect it more likely that it is be used with assign, so then I wonder if it should be:

```
{% assign myCollection = site.collections | where: "label", "myCollection" | first %}
```

Where if it is a fragment, not in braces, it is up to the user to decide how they want to use it.

pathawks:
Looks like a new PR/issue needs to be opened for this.

PR #6168, opened 2017-06-22
Allow default data access fron StaticFile

pathawks:
This is probably not the most efficient method, if there are lots of keys set that are never read.
Then again, I don't know why that would happen.

~~Note that this is _not_ a PR to `master`, but to #6162~~

/cc: @jekyll/ecosystem


lib/jekyll/static_file.rb
```diff
@@ -104,7 +100,7 @@ def to_liquid
     end
 
     def data
-      @data ||= {}
+      @data ||= @site.frontmatter_defaults.all(relative_path, type)
```
ashmaroli: Why not resort to existing method `:defaults` in [`StaticFile`](https://github.com/jekyll/jekyll/blob/v3.5.0/lib/jekyll/static_file.rb#L144-L148)

```ruby
def data
  @data ||= defaults
end
```

benbalter:
@pathawks I added a test for document front matter keys via https://github.com/jekyll/jekyll/pull/6162/commits/b139ebf4e6d8b21a7877ce69a582d008f0e9a534, which passes.

I believe Documents are okay, in that [`merge_defaults` is called when the document is read](https://github.com/jekyll/jekyll/blob/19a84584c99bfaaea5c4b33a78a485bf979d9eba/lib/jekyll/document.rb#L438-L444).

Static files aren't read, so they don't have the same pattern, but I believe your approach of making the `data` method the front matter defaults (since static files can't have other data) is the correct approach.

TL;DR: If you merge in b139ebf4e6d8b21a7877ce69a582d008f0e9a534 and revert the changes to `Document` I believe this PR should pass.

parkr:
Can you fix the conflicts with `master` and change your base branch back to `master`?

pathawks:
> Can you fix the conflicts with `master` and change your base branch back to `master`?

Done and done.
I have also reverted changes to Document.

parkr:
Doesn't line 31 do this already for StaticFile?

benbalter:
I believe this can be closed in favor of https://github.com/jekyll/jekyll/pull/6162, although the advantage here (if we removed line 31) would be that defaults are read in when data is first called, rather than when the object is initialized, which may have a performance improvement (lazy loading defaults) on larger sites or sites with many defaults.

PR #6170, opened 2017-06-23
Disable build warning when documents contain a `layout: none` declaration

jekylltools:
https://github.com/jekyll/jekyll/pull/5933 introduces `layout: none` however this triggers a build warning so check for `!document.no_layout?` before throwing the warning. I think this will fix it.

### DIFF
--- a/lib/jekyll/renderer.rb
+++ b/lib/jekyll/renderer.rb
@@ -135,7 +135,7 @@ module Jekyll
     #
     # Returns Boolean true if the layout is invalid, false if otherwise
     def invalid_layout?(layout)
-      !document.data["layout"].nil? && layout.nil? && !(document.is_a? Jekyll::Excerpt)
+      !document.data["layout"].nil? && layout.nil? && !(document.is_a? Jekyll::Excerpt) && !document.no_layout?
     end
 
     # Render layouts and place document content inside.


ashmaroli:
@jekylltools Thank you for opening this PR. I concede there is indeed a bug in my implementation.
The current implementation only works for `Jekyll::Document` objects leaving `Jekyll::Page` objects unchecked.

The solution is **to disable the concerned file being rendered into a layout completely** instead of just suppressing the warning.

You're welcome to alter the course of this PR towards the above goal or start anew.

jekylltools:
> The current implementation only works for Jekyll::Document objects leaving Jekyll::Page objects unchecked.

ah ok, I understand. I am seeing build warnings on what I think are Jekyll::Document objects

> The solution is to disable the concerned file being rendered into a layout completely instead of just suppressing the warning.

I'm confused. I thought that https://github.com/jekyll/jekyll/pull/5933 disables a `Jekyll::Document` object from being rendered into a layout? But I am still seeing `Build Warning: Layout 'none' requested in assets/tipuesearch/tipuesearch_content.js does not exist.` and I assume [`tipuesearch_content.js`](https://github.com/jekylltools/jekyll-tipue-search/blob/master/assets/tipuesearch/tipuesearch_content.js) is Jekyll::Document object?

ashmaroli:
the file `tipuesearch_content.js`  ismost likely a `Jekyll::Page` object. Documents reside in a collection.
The simplest fix for this would be render.rb:L#83
```ruby

      if document.place_in_layout? && document.data["layout"] != "none
        Jekyll.logger.debug "Rendering Layout:", document.relative_path
        output = place_in_layouts(output, payload, info)
      end
```

jekylltools:
aha!!! That makes sense now. Very cool. And for some reason I thoughts Pages were a sub-class of Documents (if sub-class is the correct terminology?) --  I'll close this.

PR #6174, opened 2017-06-24
Added improved Pug plugin - removed 404 Jade plugin

DougBeney:
I replaced the current Pug plugin, which had major flaws, with my new plugin that fixes those flaws and more. The previous Pug plugin didn't support Pug includes, Jekyll layouts, and at times was buggy with Jekyll includes. 

I also removed the Jade plugin, as it was a 404 link.

### DIFF
--- a/docs/_docs/plugins.md
+++ b/docs/_docs/plugins.md
@@ -758,10 +758,9 @@ LESS.js files during generation.
 
 #### Converters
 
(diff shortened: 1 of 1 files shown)


pathawks:
I don't know if we should just be removing other's plugins.

/cc: @jekyll/documentation

DougBeney:
Jade plugin was removed because the link was broken, thus not providing value to reader. 
The Pug plugin was replaced because it was unmaintained and lacked many useful features.

DirtyF:
@jekyllbot: merge+docs

PR #6176, opened 2017-06-25
Test with Ruby 2.4.1-1 on AppVeyor

ashmaroli:
AppVeyor added Ruby 2.4.1-1 to their list of [supported Ruby versions](https://www.appveyor.com/docs/build-environment/#ruby) close to a fortnight ago..

So let us test with this by default then :smiley:


pup:
@jekyllbot: merge +dev

PR #6179, opened 2017-06-25
Add {%raw%} to Liquid example on site

adamvoss:
Fix for https://github.com/jekyll/jekyll/pull/6165#issuecomment-310865627

Discussion: https://github.com/jekyll/jekyll/pull/6165#issuecomment-310906562
I can change it to escaping the braces if that is what is ultimately wanted.

/cc @DirtyF @ashmaroli @jekyll/documentation

### DIFF
--- a/docs/_docs/collections.md
+++ b/docs/_docs/collections.md
@@ -310,7 +310,7 @@ you specified in your `_config.yml` (if present) and the following information:
   when iterating through <code>site.collections</code> as you may need to 
   filter it out.</p>
   <p>You may wish to use filters to find your collection:
-  <code>{{ site.collections | where: "label", "myCollection" | first }}</code></p>
+  <code>{% raw %}{{ site.collections | where: "label", "myCollection" | first }}{% endraw %}</code></p>
 </div>
 
 


pathawks:
I worry this makes it less obvious that the code is Liquid.

I understand that `{{ site.collections | where: "label", "myCollection" | first }}` is not the most useful code, but if I copy-paste it into my site, I will get something that works, and it should be fairly simple to bend it to whatever task I am trying to accomplish.

I think throughout the site we use `{{ }}` and `{% %}` to denote Liquid syntax.
I think (hope) we use `{{ }}` consistently when describing filters and `{% %}` when describing tags.

docs/_docs/collections.md
```diff
@@ -310,7 +310,7 @@ you specified in your `_config.yml` (if present) and the following information:
   when iterating through <code>site.collections</code> as you may need to 
   filter it out.</p>
   <p>You may wish to use filters to find your collection:
-  <code>{{ site.collections | where: "label", "myCollection" | first }}</code></p>
+  <code>site.collections | where: "label", "myCollection" | first</code></p>
```
ashmaroli: nope this should be
```
<code>{% raw %}{{ site.collections | where: "label", "myCollection" | first }}{% endraw %}</code>
```
That's how it is elsewhere in our docs.
The priority is for how it renders at `jekyllrb.com/docs` rather than `github.com`

ashmaroli:
..and the PR title needs to edited as well. We're no longer removing the braces..

pathawks:
@jekyllbot: merge +docs

PR #6172, opened 2017-06-23
Prevent rendering Pages into layout if `layout: none` front-matter variable is set

jekylltools:
Prevent rendering pages into layout if `layout: none` front-matter variable is set. Thanks to @ashmaroli https://github.com/jekyll/jekyll/pull/6170#issuecomment-310591588

PR #6182, opened 2017-06-26
Disable default layouts for Pages with a `layout: none` declaration

ashmaroli:
Continuing from #6172 
Resolves confusion reported in PR #6170 where-in users thought the feature added in #5933 applied to regular `Jekyll::Page` objects as well.

/cc @jekyll/ecosystem

### DIFF
--- a/features/rendering.feature
+++ b/features/rendering.feature
@@ -40,7 +40,7 @@ Feature: Rendering
     And I should not see "Ahoy, indeed!" in "_site/index.css"
     And I should not see "Ahoy, indeed!" in "_site/index.js"
 
-  Scenario: Ignore defaults and don't place documents with layout set to 'none'
+  Scenario: Ignore defaults and don't place pages and documents with layout set to 'none'
     Given I have a "index.md" page with layout "none" that contains "Hi there, {{ site.author }}!"
     And I have a _trials directory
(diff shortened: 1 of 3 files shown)


pathawks:
Is there anything else we need to modify so that `layout: none` works consistently everywhere?

ashmaroli:
> Is there anything else we need to modify so that `layout: none` works consistently everywhere?

Should I move this patch and the released public_method `:no_layout?` from #5933  into `Jekyll::Convertible` instead?

parkr:
Just modify :place_in_layout? Please! You can create a helper private method but place_in_layout? Should return false if layout:none.

ashmaroli:
> because everything breaks somebody's workflow.

This was noted by Parker [in this comment](https://github.com/jekyll/jekyll/pull/5933#issuecomment-290621036) and since nobody reported a break so far, the safe assumption continues..

ashmaroli:
@parkr I feel this can be included in `v3.6.0`..

parkr:
You could also call this a bugfix and release in 3.6.1 since it could be considered a bug that there's no way to exclude these files from layouts.

pathawks:
@jekyllbot: merge +minor

PR #6183, opened 2017-06-26
bump kramdown to 1.14.0 for rouge 2 support

Crunch09:
see https://kramdown.gettalong.org/news.html

cc: @parkr

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -70,7 +70,7 @@ group :jekyll_optional_dependencies do
   gem "jekyll-gist"
   gem "jekyll-paginate"
   gem "jekyll-redirect-from"
-  gem "kramdown", "~> 1.9"
+  gem "kramdown", "~> 1.14"
   gem "mime-types", "~> 3.0"
   gem "rdoc", "~> 5.0"
   gem "toml", "~> 0.1.0"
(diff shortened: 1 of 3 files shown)


Crunch09:
This fixes the rendering issue of `raw` tags mentioned by @DirtyF in #5919

jekyll.gemspec
```diff
@@ -34,7 +34,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency("colorator",             "~> 1.0")
   s.add_runtime_dependency("jekyll-sass-converter", "~> 1.0")
   s.add_runtime_dependency("jekyll-watch",          "~> 1.1")
-  s.add_runtime_dependency("kramdown",              "~> 1.3")
+  s.add_runtime_dependency("kramdown",              "~> 1.14", ">= 1.14.0")
```
Crunch09: I wasn't sure if we should still allow versions < 1.14 when using rouge 1 but it would make things quite complicated and a bump should still work with rouge 1.

Gemfile
```diff
@@ -70,7 +70,7 @@ group :jekyll_optional_dependencies do
   gem "jekyll-gist"
   gem "jekyll-paginate"
   gem "jekyll-redirect-from"
-  gem "kramdown", "~> 1.9"
+  gem "kramdown", "~> 1.14", ">= 1.14.0"
```
parkr: This is unnecessary – `~> 1.14` is sufficient. 😄

jekyll.gemspec
parkr: `~> 1.14` is the same as `>= 1.14.0, < 2.0`.

Gemfile
Crunch09: Thank you! I always mix up these things :(

Crunch09:
Yes, it seemed to work find at first but e.g. `raw` tags were not properly highlighted before gettalong/kramdown#413.

PR #6184, opened 2017-06-26
set minimum requirement for jekyll-feed

Crunch09:
Since jekyll/jekyll#6137 `baseurl` can be `nil`. Until jekyll-feed 0.8
the `baseurl` has been used as argument to `URI.join`. When `baseurl`
is `nil` this throws an exeption:

> URI.join('http://example.com', nil)
> ArgumentError: bad argument (expected URI object or URI string)

The usage of `baseurl` within jekyll-feed has been removed in
jekyll/jekyll-feed@bf728c321d715f8bb1673e43fb3c848fe736fa6d

/cc @jekyll/ecosystem

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -65,7 +65,7 @@ group :jekyll_optional_dependencies do
   gem "coderay", "~> 1.1.0"
   gem "jekyll-coffeescript"
   gem "jekyll-docs", :path => "../docs" if Dir.exist?("../docs") && ENV["JEKYLL_VERSION"]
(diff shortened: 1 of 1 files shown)


pathawks:
[There are probably children out there holding down spacebar to stay warm in the winter! YOUR UPDATE MURDERS CHILDREN.](https://xkcd.com/1172/)

pathawks:
@jekyllbot: merge +dev

pathawks:
Thanks @Crunch09!

PR #6186, opened 2017-06-28
Always normalize the result of the relative_url filter (3.4.x backport)

benbalter:
Backport of https://github.com/jekyll/jekyll/pull/6185 for the 3.4.x branch since https://github.com/jekyll/jekyll/pull/6137 was backported and introduced the bug.

### DIFF
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -23,9 +23,9 @@ module Jekyll
       def relative_url(input)
         return if input.nil?
         site = @context.registers[:site]
-        return ensure_leading_slash(input.to_s) if site.config["baseurl"].nil?
+        parts = [site.config["baseurl"], input]
         Addressable::URI.parse(
-          ensure_leading_slash(site.config["baseurl"].to_s) + ensure_leading_slash(input.to_s)
+          parts.compact.map { |part| ensure_leading_slash(part.to_s) }.join
         ).normalize.to_s
       end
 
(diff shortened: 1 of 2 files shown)


benbalter:
Added a test via dc6e5bc and confirmed it fails as expected on `3.5-stable`.

lib/jekyll/filters/url_filters.rb
```diff
@@ -23,9 +23,9 @@ def absolute_url(input)
       def relative_url(input)
         return if input.nil?
         site = @context.registers[:site]
-        return ensure_leading_slash(input.to_s) if site.config["baseurl"].nil?
+        parts = [site.config["baseurl"], input]
```
parkr: The other method in `master`, [`sanitized_baseurl`](https://github.com/jekyll/jekyll/blob/b0136fccfab51fd7885207fa79bc3dbf3f64c6f4/lib/jekyll/filters/url_filters.rb#L44-L47), always returns a string, which means this code is not valid on `master`. Here, `site.config["baseurl"]` _can_ be `nil`, but on master, [`parts.compact`](https://github.com/jekyll/jekyll/blob/b0136fccfab51fd7885207fa79bc3dbf3f64c6f4/lib/jekyll/filters/url_filters.rb#L26-L29) is always a no-op. @benbalter, does that invalidate the PR on master?

parkr:
All tests pass for me locally.

parkr:
3.4.5 is released. https://rubygems.org/gems/jekyll/versions/3.4.5

PR #6192, opened 2017-07-01
Use Warn for deprecation messages

DirtyF:
Following discussion in #5130  

@crunch09 and @envygeeks suggested to use `warn` instead of `error` for deprecation message.

/cc @jekyll/stability


DirtyF:
@jekyllbot: merge +minor

PR #6193, opened 2017-07-01
Small correction in documentation for includes

kplattret:
When passing references to YAML files as parameter values the references shouldn't be added as an object, i.e. it shouldn't be wrapped in double curly brackets.


parkr:
This seems right to me. Great catch.

I'll try to test this out and ensure it works as intended before merging. If someone else wants to test this, feel free to add an approval to this PR. 

Thank you!

parkr:
@jekyllbot: merge +site

PR #6196, opened 2017-07-03
Correct deprecated configuration.

envygeeks:
This pull-request is correcting a flaw in your `_config.yml` for Jekyll.  The current Jekyll `_config.yml` causes a deprecation message out of the gate.  This is unexpected behavior.


DirtyF:
Duplicates #6045 but you're right nevertheless

PR #6197, opened 2017-07-03
Github-Pages is not a plugin, it's a replacement.

envygeeks:
Putting Github-Pages inside of Jekyll Plugins is bad form, it's also misleading to users who often assume it's a plugin, it's not... it's a replacement for Jekyll in it's entirety, with it's own patch set.


benbalter:
> Putting Github-Pages inside of Jekyll Plugins is bad form

It's the recommended way to install the GitHub Pages gem. It allows the Gem to bootstrap earlier in the build process and enforce local defaults/configurations so that the local development environment more closely replicates production.

DirtyF:
While Jordon is technically right, as Ben mentions it's an official recommendation from GitHub, we better leave this as is.

PR #6198, opened 2017-07-03
Users should be installing patch versions.

envygeeks:
Setting Jekyll to only explicitly install the latest version at runtime is bad form.


DirtyF:
@jekyllbot: merge +dev

PR #6200, opened 2017-07-03
Access custom config array throughout session

### LINKED ISSUE #6180: [Discussion] Custom config file array
As of `v3.5.0`, **`Jekyll::Command.configuration_from_options`** strips out the array of override config file via
[this line from `comfiguration.rb`](https://github.com/jekyll/jekyll/blob/v3.5.0/lib/jekyll/configuration.rb#L157-L158) which was added way back before v1.0
and this discussion ticket is to unearth the rationale behind supporting the above deletion.

From my understanding of the code, *the presence of key `"config"` doesn't seem to cause any trouble.* Hence the confusion why it was decided to be removed..?
If the presence of key `"config"` indeed has no serious consequences, I'd then like to send a PR to patch that line to obtain the array via:
```diff
- config_files = override.delete("config")
+ config_files = override["config"]
```

This patch would then let me use that array in a plugin like `jekyll-admin` (possibly allow resolution of jekyll/jekyll-admin#321)

--
Requesting inputs from @parkr, @mattr-, @benbalter, @pathawks, @Crunch09

ashmaroli:
Resolves #6180 
Instead of having the array of custom configs removed, let it be available throughout current session.

#### Specific Use Case:
Allow Jekyll Admin to determine which config file to re-read if not the default `_config.yml` or `_config.yaml`

/cc @jekyll/ecosystem @parkr

### DIFF
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -155,7 +155,7 @@ module Jekyll
       )
 
       # Get configuration from <source>/_config.yml or <source>/<config_file>
-      config_files = override.delete("config")
+      config_files = override["config"]
       if config_files.to_s.empty?
         default = %w(yml yaml).find(-> { "yml" }) do |ext|
           File.exist?(Jekyll.sanitized_path(source(override), "_config.#{ext}"))
--- a/lib/jekyll/drops/site_drop.rb
+++ b/lib/jekyll/drops/site_drop.rb
@@ -37,6 +37,9 @@ module Jekyll
         @site_collections ||= @obj.collections.values.sort_by(&:label).map(&:to_liquid)
       end
 
+      # return nil for `{{ site.config }}` even if --config was passed via CLI
+      def config; end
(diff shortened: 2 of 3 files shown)


ashmaroli:
> Three things: (1) when config is empty/has an empty string like above, as checked in the line directly after the one you change, let's fill that in with whatever file we read by default; (2) let's add a test; (3) let's not expose this in liquid yet, so create a no-op method in SiteDrop called config.

(1) Isn't that how it is already? `If config_files is an empty array, proceed to look for default config file(s)
(2) Fixing failing tests do produce enough results asserting the changed behavior, no clear on what else needs to be tested..
(3) Implemented with an empty `#config`

@parkr

ashmaroli:
Hi, Any other changes to be made?

ashmaroli:
:+1:  Thanks Olivia

parkr:
👏 

@jekyllbot: merge +minor

PR #6208, opened 2017-07-07
Use HTTPS protocol in "./test/source/_slides"

SuriyaaKudoIsc:
Change protocol from `http://` to `https://`.

*Cheers,*
[**Suriyaa Kudo**](https://github.com/SuriyaaKudoIsc) :octocat:

### DIFF
--- a/test/source/_slides/example-slide-4.html
+++ b/test/source/_slides/example-slide-4.html
@@ -5,7 +5,7 @@
 
 Jekyll is a simple, blog-aware, static site generator. It takes a template
 directory containing raw text files in various formats, runs it through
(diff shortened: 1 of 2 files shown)


SuriyaaKudoIsc:
TODO:
- [ ] Rebase commits

parkr:
Hey! Thanks for the pull request. As these are just test fixtures, updating to use of HTTPS doesn't provide value. These files are never rendered and served to Internet users, so using HTTPS instead of HTTP here doesn't matter. It's just for testing, so I'd rather not create unnecessary churn on these files. Thanks, though!

PR #6210, opened 2017-07-08
Linking the link

phocks:
Helpful if people don't like copying and pasting URLs :)


DirtyF:
Thanks @phocks for being helpful 😄  

@jekyllbot: merge +docs

Issue #4097, opened 2015-11-02
Jekyll 3.0.0 | Error:  undefined method `gsub' for #<Hash....


Jepz:
Just updated my Jekyll  2.5.3 to 3.0, tried to run "jekyll serve", under "Generating..."-phase i'm getting the error: jekyll 3.0.0 | Error:  undefined method `gsub' for #<Hash:0x007fa363880460>. FYI, the numbers after Hash is new everytime.

Care to help me out? =)

alfredxing:
Can you run Jekyll with the `--trace` flag? It'll help us pinpoint the issue. Also, if you're able to, the source code of the site you're trying to build would be super useful too.

Jepz:
Terminal gives me this:
  Deprecation: Jekyll now uses subcommands instead of just switches. Run `jekyll --help` to find out more.
..when I run jekyll --trace.

I guess something must be broken... :-/

Jepz:
Sorry about the previous post, changed it to the log information you were asking for.
When you say source code, what do you need? Would it be good enough with a lin to the repo?

alfredxing:
Found the issue! If you're specifying `output` or `permalink` for a collection, you'll need to specify the collection itself as a hash and not part of an array.

So in your `_config.yml`, use this instead:
```yaml
collections:
  projects:
    output: true
    permalink: /projects/:path/
```

Let me know if that works!

Jepz:
Thanks! it helped!
I actually hade that already in my config, but hade the hyphen before projects. Removing the hyphen made it work =)

Thanks for helping out!

ron4stoppable:
thanks @alfredxing that worked for me too

PR #5518, opened 2016-10-29
YAML fix

dret:
assuming that none of the hyphens in there were intentional. even if they were, the documentation page as published is inconsistent in the examples and should be cleaned up one way or another.

PR #6212, opened 2017-07-08
Update collections.md

Jared37:
removed the opening hyphen from line 27. The hyphen created the error described here:
https://github.com/jekyll/jekyll/issues/4097

### DIFF
--- a/docs/_docs/collections.md
+++ b/docs/_docs/collections.md
@@ -24,7 +24,7 @@ with the name of your collection:
 
 ```yaml
 collections:
-- my_collection
+  my_collection
 ```
 
 You can optionally specify metadata for your collection in the configuration:


pathawks:
Shouldn't `collections` be an array?

Jared37:
Perhaps, I'm not versed on the technical elements. What I know is that the remainder of the references do no include the hyphen, and when i do include the hyphen in jekyll 3.5, entering a collection like so:
> 234  collections:
> 235    - teaching:
> 236     output: true

it produces this error:
>jekyll 3.5.0 | Error:  (/home/jared/mmj/_config.yml): did not find expected '-' indicator while parsing a block collection at line 235 column 3
It's also worth noting that the hyphen isn't used in the remainder of the example snippets in the documentation.

ashmaroli:
This change is incorrect. `collections` **need to be** an array **or** an object (Hash) like so..
```
collections:
  - members
  - methods
```
```
collections:
  members:
    output: false
  methods:
    output: false
```

ashmaroli:
Please refer #5518 for further info

parkr:
The hyphen is needed for it to be an array with the default of no options. I'd be down to call out the fact that we changed the data structure but removing the hyphen just makes the YAML invalid.

PR #6214, opened 2017-07-09
Fix docs site page margin

cobyism:
Having not looked at this site in a while, I noticed it looks like the page `body` has lost its margin reset. Note the space around the gold line at the top of the page in the screenshots below:

2017-07-09 at 16 34

2017-07-09 at 16 44

This PR resets the `margin` of the `body` element to zero, and also adds a little bit of horizontal padding for the mobile version of the site-nav, which otherwise wouldn’t have much space—especially when the leftmost item is the currently selected menu item:

2017-07-09 at 16 45

2017-07-09 at 16 45

/cc @parkr @mattr-

### DIFF
--- a/docs/_sass/_style.scss
+++ b/docs/_sass/_style.scss
@@ -22,6 +22,7 @@ body {
   -o-font-feature-settings: "kern" 1;
   font-feature-settings: "kern" 1;
   font-kerning: normal;
+  margin: 0;
 }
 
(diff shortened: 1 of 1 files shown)


parkr:
Thank you so much, @cobyism! Your design endures to this day — thank you for building it!

@jekyllbot: merge +site

Issue #6216, opened 2017-07-10
Two problems when installing Jekyll on macOS 


doekman:
I tried to install Jekyll on a brand new Sierra machine. I found two issues, which prevented me to be "Get up and running in seconds"... This can be a bit discouraging when trying to install Jekyll, and the [web-page](http://jekyllrb.com) doesn't provide any other hints how to install on macOS. 

1. Permission problems

`gem install jekyll bundler` results in `ERROR:  While executing gem ... (Gem::FilePermissionError)
    You don't have write permissions for the /Library/Ruby/Gems/2.0.0 directory.`.

I solved this by running sudo: `sudo gem install jekyll bundler`

2. Wrong ruby version

When running the above install command, I got the error `ERROR:  Error installing jekyll:
	liquid requires Ruby version >= 2.1.0.` (default ruby install on Sierra: `ruby 2.0.0p648 (2015-12-16 revision 53162) [universal.x86_64-darwin16]`)

I fixed this by install the latest Ruby with brew: `brew install ruby`.

jmknoll:
Not sure if this is exactly the issue that you're dealing with, but its important to use something like [rbenv](https://github.com/rbenv/rbenv) or [rvm](https://rvm.io/) on MacOS, as the operating system uses ruby for a lot of things under the hood. By default, you can't access the system ruby. You can force it with sudo, but this isn't recommended, as you can end up with a mess of dependencies (more dealing with Rails, but additional info [here](http://railsapps.github.io/installrubyonrails-mac.html). That should also solve your second problem.

I think this isn't an issue particular to Jekyll, but more to the workings of MacOS.

pup:
i believe that it's listed on the installation page that ruby 2.1+ is required for jekyll to run, which in itself forces users to not use system ruby (aside from the fact that it's usually inaccessible). the method using homebrew is [described here](http://jekyllrb.com/docs/troubleshooting/#jekyll--mac-os-x-1011).

so yeah, as @jmknoll said, this is more of a MacOS thing, rather than jekyll thing

doekman:
Perhaps not a Jekyll (as in Ruby code) issue, but I think it sure is a problem... The [installation page](https://jekyllrb.com/docs/installation/) has some issues installing jekyll on macOS. However, I don't know enough about ruby/gem to address this.

(just a Ruby-Noob wanting to help)

Crunch09:
[...]

PR #6220, opened 2017-07-12
Bump supported Ruby version to >= 2.1.0

ashmaroli:
I realized from #6216 that our `runtime_dependency` `Liquid 4.0` has already [dropped support](https://rubygems.org/gems/liquid/versions/4.0.0) for Ruby Versions older than `2.1.0`. So there's no point in not following suit ourselves.


parkr:
This is a major breaking change and will have to wait for v4.0 which we have no plans for at the moment. Thanks!

parkr:
Ah, I see, Liquid has already done this.

parkr:
@jekyllbot: merge +minor

Issue #5963, opened 2017-03-17
Jekyll with Ruby 2.4 ignores timezone for date field of posts


stomar:
We have posts with a date field in the YAML front matter that includes a timezone offset. The site is generated with UTC as local timezone.

The expected and established behavior in this setting is that post URL's are generated based on the date converted to UTC, such that e.g. a post with date `2017-01-01 06:00:00 +1200` appears under the URL `.../2016/12/31/...`.

This ensures for example that posts authored in different timezones will still be sorted and displayed in the proper chronological order.

Strangely, this behavior changes when switching from Ruby 2.3 to Ruby 2.4: with 2.4, the timezone offset seems to be ignored, and the post appears under the URL `.../2017/01/01/...`.
This occurs without changing the Jekyll version; here I tested with Jekyll 3.3.1, but observed the same behavior also for earlier Jekyll releases.

For Ruby 2.4:
``` sh
$ ls -R
.:
_config.yml _posts

./_posts:
2000-01-01-post-1.md
$ cat _posts/2000-01-01-post-1.md
---
date: 2017-01-01 06:00:00 +1200
---
$ cat _config.yml
timezone: UTC
$ ruby -v
ruby 2.4.0p0 (2016-12-24 revision 57164) [i686-linux]
$ jekyll build
Configuration file: none
[snip]
$ ls _site/*/*/*/post-1.html
_site/2017/01/01/post-1.html
```

For Ruby 2.3 the URL is as expected:

``` sh
$ ruby -v
ruby 2.3.3p222 (2016-11-21 revision 56859) [i686-linux]
$ jekyll build
Configuration file: none
[snip]
$ ls _site/*/*/*/post-1.html
_site/2016/12/31/post-1.html
```

stomar:
I found out that SafeYAML.load changed the way how dates are handled. I don't know whether this change is intended or a bug.

The behavior until 2.3 was that the parsed date is converted to the local timezone (that's also what Ruby's YAML.load does):

``` ruby
RUBY_VERSION  # => "2.3.3"

ENV["TZ"] = "UTC"  
data = "date: 2017-01-01 06:00:00 +1200"

require "yaml"
YAML.load data      # => {"date"=>2016-12-31 18:00:00 +0000}

require "safe_yaml"
SafeYAML.load data  # => {"date"=>2016-12-31 18:00:00 +0000}

require "safe_yaml/version"
SafeYAML::VERSION   # => "1.0.4"
```
[...]

Issue #6033, opened 2017-04-17
Respecting the post's "local" timezone (revisiting #4096)


t-brink:
I would like to revisit issue #4096. Summary: I record the timezone offset for every post in the front matter: `date: 2017-04-17 23:00:00 +0200`. Jekyll always converts that to the local timezone, a fact that will not change to avoid regressions, I understand that.

Sadly, this is not entirely satisfactory. One may want to change the timezone for the blog temporarily (holidays, staying abroad for a few months) or permanently (moving from one timezone to another inside the US should be rather common). This can have very bad side effects: When using URL schemes including the date, the URLs might change! The data in any RSS feed would also change. I would also prefer that the timestamp of any blog post retains the local time from when it was written. This last item may not be everyone's taste, though, and they may be served by always using UTC.

So the question is what can be done about this? Currently, you still have to either monkey patch `Document` (see link in #4096) or keep another, redundant field in the front matter. From my point of view, the optimal solution would be some sort of global option to keep the blog post date in the same timezone. Alternatively, at least storing the original string somewhere or providing a hook to get that data before conversion would be acceptable. Although, then it is still unclear to me how to generate the correct URLs.

I would be willing to try to submit a patch, but I would like some input on what should be done before I get my hands dirty, it is after all a user-facing issue. Any thoughts about the best way to do this?

pathawks:
I would love to have an option to preserve the timezone anywhere one is provided. This is certainly what I expect to happen as a user.

t-brink:
OK, looking into this further, the problem starts with SafeYAML, see <https://github.com/dtao/safe_yaml/blob/master/lib/safe_yaml/parse/date.rb>. It also converts dates to local time, either explicitly or by calling DateTime.to_time, which converts to local time, too. That means the behavior in jekyll would differ between dates given with or without quotes. The patch for SafeYAML would be trivial, the question is more one of API.

As there seems to be at least some interest, I will try engage this project about the issue first.

pathawks:
> That means the behavior in jekyll would differ between dates given with or without quotes.

That's a non-starter. I wish that everybody would just use YAML native types, but that is not the case.

t-brink:
I agree that "date:" should behave the same in all cases. Looking at the commit history and the lack of comments on issues, it seems that SafeYAML development is currently inactive. What is the jekyll project's policy in this case? Should we monkey patch SafeYAML for the time being? Fork it?

t-brink:
[...]

PR #6224, opened 2017-07-15
Optionally preserve timezones of time stamps, see issue #6033

t-brink:
Time stamps in jekyll are converted to the local time zone (either of the OS or via the "timezone" setting in _config.yaml). This is not the best way to do it, since changing the time zone changes (amongst other things) the posts' URLs. This pull request aims to fix that, see discussion in #6033.

The problem is that SafeYAML (used to parse the front matter) does discard the timezone data. I proposed a patch to its upstream (https://github.com/dtao/safe_yaml/pull/87), but upstream seems to be inactive. Therefore, I had to monkey patch!

This pull request works for me, but misses documentation and unit tests. Since it is still undecided if the monkey patching is acceptable (see #6033), I would first like to get feedback on the code. If the code can be accepted, I will update the patch to include docs and tests.

### DIFF
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -54,6 +54,7 @@ module Jekyll
       "permalink"           => "date",
       "paginate_path"       => "/page:num",
       "timezone"            => nil, # use the local timezone
+      "preserve_timezones"  => false, # convert to the above timezone
 
       "quiet"               => false,
       "verbose"             => false,
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -33,6 +33,22 @@ require "colorator"
 
 SafeYAML::OPTIONS[:suppress_warnings] = true
 
+# Monkey patch safe_yaml to preserve the time zone of dates.
+module SafeYAML
+  class Parse
+    class Date
+      def self.value(value)
+        d = DateTime.parse(value)
+        if Jekyll.preserve_timezones
+          d.to_time.localtime(d.zone)
+        else
+          d.to_time
+        end
+      end
+    end
+  end
+end
+
 module Jekyll
   # internal requires
   autoload :Cleaner,             "jekyll/cleaner"
@@ -90,6 +106,12 @@ module Jekyll
       ENV["JEKYLL_ENV"] || "development"
     end
 
+    # Public: Tells if timezones in dates should be preserved (true) or
+    # reset to the local timezone.
+    def preserve_timezones
+      @preserve_timezones
+    end
+
     # Public: Generate a Jekyll configuration Hash by merging the default
     # options with anything in _config.yml, and adding the given options on top.
     #
@@ -108,6 +130,9 @@ module Jekyll
 
       # Merge DEFAULTS < _config.yml < override
       Configuration.from(Utils.deep_merge_hashes(config, override)).tap do |obj|
+        if obj.key?("preserve_timezones")
+          @preserve_timezones = obj["preserve_timezones"] ? true : false
+        end
         set_timezone(obj["timezone"]) if obj["timezone"]
       end
     end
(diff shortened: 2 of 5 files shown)


parkr:
Hey, @t-brink! Inconsistent behaviour across versions is a bug to me. What do you think about a patch which just applies this change for v2.4.0 and above automatically? If I have a site and I run it through 2.3.3, I should get the same thing when I subsequently run it through 2.4.0. What do you think?

t-brink:
Hi! I am not sure what you mean. The change is only enabled if the config setting "preserve_timezones" is set to true in _config.yml. It defaults to false. The only thing I did is to add "preserve_timezones: true" to the default config for new sites, as I think this behavior is more sane. Therefore, behavior for existing sites should not change, everything else would be a bug, I agree.

Could you clarify your comment please? Maybe I'm missing something.

Crunch09:
@t-brink I think what Parker meant is that the timezone has been preserved by SafeYAML until Ruby 2.3.3 and it should be the same default behavior when the user is using Ruby 2.4. The issue is described in #5963.

t-brink:
Hmm, I think I see what you mean. I have Ruby 2.3.3, so I did not encounter bug #5963, yet.

So the behavior should be: convert to the blog's timezone (either system timezone or whatever the user specified in _config.yml) by default. With my option, keep the timezone given in the blog post. This should be easy to implement and fix #5963 in one go.

The only question is, if the timezone preservation should be added to the _config.yml of new sites or not. For existing sites, we would preserve the old behavior of converting to the blog's timezone, i.e., what happened with ruby <= 2.3.

Did I understand this correctly? I am traveling a lot with limited internet access currently, so I will be slow to implement these fixes, but I can do it over the next few weeks, I think. If somebody else wants to take my patch and run with it, just mention it and feel free to do it.

ashmaroli:
@t-brink What Parker means can be summarized as following:
```ruby
if RUBY_VERSION >= "2.4.0"
  apply_timezone_patch
end
```
no extra config file entries or modifications.. the bug is patched internally without any user-intervention

t-brink:
This is a misunderstanding. I find the behavior of Jekyll with Ruby 2.3.3 wrong, see discussion in #6033! This is not about fixing #5963, although that can be achieved, too, in this patch!

What I want is to preserve the timezone of the blog post, which is _converted_ to the blog's timezone setting in Ruby 2.3.3. Basically the behavior in #5963 is what I _want_, but it breaks compatibility. Thus I want to enable it for all ruby versions optionally, while keeping the old behavior for all ruby versions by default.

The word "preserve" in the context of bug #5963 is misleading, since timezone data is not preserved, but discarded by converting to another timezone. What I mean by preserving is that I set "date: 2017-01-01 23:18:33 +0600" in my blog post and the timezone "+0600" is preserved independently of the system time zone and any setting in _config.yml! I hope this is clearer now.

pathawks:
@Crunch09 @ashmaroli Jekyll (SafeYAML) has never respected timezones of posts.

This is not about fixing something that has changed with Ruby 2.4.0; this is about adding a new feature.

ashmaroli:
> this is about adding a new feature

In that case the `fix` label needs to be replaced with the `enhancement` label

lib/jekyll/configuration.rb
```diff
@@ -54,6 +54,7 @@ class Configuration < Hash
       "permalink"           => "date",
       "paginate_path"       => "/page:num",
       "timezone"            => nil, # use the local timezone
+      "preserve_timezones"  => false, # convert to the above timezone
```
pathawks: I suppose defaulting this to `true` could change permalinks of sites that relied on the old behavior.

t-brink: Yes. People were saying in some older bug reports that they rely on the old behavior. That's why I set the default to `false` and added a line to the default config (`lib/site_template/_config.yml`) to set it to `true` for new sites. Although that is a policy decision; the project might prefer to not do the latter. I don't care either way.

pathawks: I agree with you that this conservative approach is best :+1:

PR #6225, opened 2017-07-16
Adding DevKit helpers

phocks:
wdm requires the Ruby DevKit to install native extensions

### DIFF
--- a/docs/_docs/windows.md
+++ b/docs/_docs/windows.md
@@ -212,5 +212,7 @@ As of v1.3.0, Jekyll uses the `listen` gem to watch for changes when the `--watc
 Add the following to the Gemfile for your site if you have issues with auto-regeneration on Windows alone:
 
 ```ruby
-gem 'wdm', '~> 0.1.0' if Gem.win_platform?
+gem 'wdm', '~> 0.1.1' if Gem.win_platform?
 ```
+
+You may first have to download and install the [Ruby DevKit](https://rubyinstaller.org/downloads/) by following [the instructions here](https://github.com/oneclick/rubyinstaller/wiki/Development-Kit).


docs/_docs/windows.md
```diff
@@ -212,5 +212,7 @@ As of v1.3.0, Jekyll uses the `listen` gem to watch for changes when the `--watc
 Add the following to the Gemfile for your site if you have issues with auto-regeneration on Windows alone:
 
 ```ruby
-gem 'wdm', '~> 0.1.0' if Gem.win_platform?
+gem 'wdm', '~> 0.1.1' if Gem.win_platform?
```
ashmaroli: `gem 'wdm', '~> 0.1.0' if Gem.win_platform?` will automatically install the most recent version in `0.1.x` series. So this change is unnecessary.

docs/_docs/windows.md
```diff
@@ -212,5 +212,7 @@ As of v1.3.0, Jekyll uses the `listen` gem to watch for changes when the `--watc
...
 ```
+
+You may first have to download and install the [Ruby DevKit](https://rubyinstaller.org/downloads/) by following [the instructions here](https://github.com/oneclick/rubyinstaller/wiki/Development-Kit).
```
ashmaroli: RubyInstaller is only for Ruby Versions upto those in v2.3. Those using Ruby 2.4.x and above will need to check out RubyInstaller2

docs/_docs/windows.md
Crunch09: It makes sense though if at least version `0.1.1` is needed. Given that in this version the installation with Ruby 2.2 apparently has been fixed ([see wdm changelog](https://github.com/Maher4Ever/wdm/releases/tag/v0.1.1)) i think this update to the docs makes sense.

ashmaroli: Ah! I see.. Thanks Florian :smiley: :+1:

parkr:
@jekyllbot: merge +site

PR #6226, opened 2017-07-17
Reader#read_directories: guard against an entry not being a directory

parkr:
Fixes the following error:

```text
Not a directory @ dir_chdir - /home/jekyll/src/_posts
```

/cc @jekyll/stability


parkr:
@jekyllbot: merge +bug

PR #6228, opened 2017-07-18
Strip unnecessary leading whitespace in template

ashmaroli:
:scissors:


pup:
@jekyllbot: merge +dev

pup:
wow, was that the quickest PR we've ever had?

ashmaroli:
Thanks Olivia! That was super-quick! :smiley:

PR #6231, opened 2017-07-18
Fix serving files that clash with directories (#6222)

### LINKED ISSUE #6222: WEBrick directory listings take priority over permalinks
Prefixing the default post permalink and setting page permalink to the same value doesn't work.

Seems to only affect the local `jekyll serve` WEBrick server. GitHub Pages, for example, isn't affected.

---------

---

- [x] I am on (or have tested on) ***Arch*** GNU/Linux

---

- [ ] I was trying to build.
- [x] It was another bug.

## My Reproduction Steps

- in `_config.yml`, set permalink to `/blog/:title`
- create a page and set its permalink to `/blog`
- run `jekyll serve`
- visit [`localhost:4000/blog`](http://localhost:4000/blog)
  - Note that you are seeing the WEBrick server's directory index, not your page
- visit [`localhost:4000/blog.html`](http://localhost:4000/blog.html)
  - Note that the page _does_ exist
- in `_config.yml`, set permalink to `/notblog/:title`
- rerun `jekyll serve`
- visit [`localhost:4000/blog`](http://localhost:4000/blog)
  - Note that it works correctly now

## The Output I Wanted

I expected that visiting the permalink defined in the page's front matter would display that page.

Related to #3452

MattSturgeon:
## Fix #6222.

To avoid having to re-implement the entire set_filename method, I instead extended `search_index_file` and temporarily mutate `res.filename`. Kinda hacky, so I tried to leave plenty of inline comments. (Also first time using ruby, so beware!)

For reference, here's the base class `search_index_file` [implementation](https://github.com/ruby/webrick/blob/3fb6a011/lib/webrick/httpservlet/filehandler.rb#L356-L363).

Feel free to checkout the commit message too.

<!--
`script/cibuild` passed ok, though I did [disable](/jekyll/jekyll/blob/16897d3b/lib/jekyll/commands/serve/servlet.rb#L28) `Lint/AssignmentInCondition` so that I could do things like:

```ruby
  while basename = path_arr.pop
    break unless basename == "/"
  end
```-->

### DIFF
--- a/lib/jekyll/commands/serve/servlet.rb
+++ b/lib/jekyll/commands/serve/servlet.rb
@@ -22,7 +22,7 @@ module Jekyll
 
         def search_file(req, res, basename)
           # /file.* > /file/index.html > /file.html
-          super || super(req, res, "#{basename}.html")
+          super || super(req, res, ".html") || super(req, res, "#{basename}.html")
         end
 
         # rubocop:disable Style/MethodName

### CODE REVIEW

**Comment by pathawks** [comment]
Thank you @MattSturgeon for learning Ruby just to fix this! :beers:

**Comment by MattSturgeon** [comment]
No problem, feel free to recommend any changes if required.

Just noticed the first mistake, I linked to the wrong super method in the commit message! Should be [this one](https://github.com/ruby/webrick/blob/3fb6a011/lib/webrick/httpservlet/filehandler.rb#L356-L363) instead, so I'll amend the commit and we can wait for the CI builds to start from scratch!

File: lib/jekyll/commands/serve/servlet.rb [hunk 1]
```diff
@@ -21,10 +21,43 @@ def initialize(server, root, callbacks)
...
         end
 
+        # rubocop:disable Lint/AssignmentInCondition
+        def search_index_file(req, res)
+          # /file/index.html -> /file.html
+
+          # First, let's see if the default implementation can figure it out.
+          # (Check for index files in the res.filename directory)
+          if file = super
+            return file
```
ashmaroli: arguments for `if`, `unless`, `while` etc check for equality via (`==` or `===`) and the body *`assigns`* a value to a variable via `=`

MattSturgeon: I wasn't checking for equality. I was checking if the assigned value was truthy. if `file = super` evaluates to `nul` we continue, otherwise we return the value assigned to `file`.

MattSturgeon: Appreciate you mentioning it though, since I did say I was new to ruby. (Not new to programming though).

ashmaroli: I see.. that wasn't readily apparent to me from reading the code.
Learnt something new today then.. :beers:

lib/jekyll/commands/serve/servlet.rb
```diff
@@ -29,31 +29,38 @@ def search_file(req, res, basename)
...
+            # basename.
+            #
+            # We use a while loop just in case res.filename has trailing slashes.
+            #
+            # Once we have popped off the basename, we can join up what's left and use it
+            # as the new res.filename.
+            path_arr = res.filename.scan(%r!/[^/]*!)
+            while basename = path_arr.pop
+              break unless basename == "/"
+            end
```
ashmaroli: I feel this `while..break unless --` can be replaced by an alternative..

MattSturgeon:
I've tried to make the code more readable, following @ashmaroli's feedback.

In [20a333f](/jekyll/jekyll/pull/6231/commits/20a333fe11518570081de67f698a1759f41c3f15?w=1) I replaced a conditional return with an `unless` branch and removed two out of three cases of assignment within conditions. I also re-write most of the comments.

It's a fixup commit, so intended to be rebased/autosquashed into the original commit, just thought I'd keep it separate for now in case anyone wanted to compare it with the original PR _([`?w=1`](https://github.com/blog/967-github-secrets) is your friend here)_.

lib/jekyll/commands/serve/servlet.rb
MattSturgeon: It can probably just be `basename = path_arr.pop`, but to cover edge cases I wanted to loop until basename wasn't empty (well, "/") since there may be (a?) trailing slash.

ashmaroli: okay, can you list 3-5 such edge-cases of `path` and `basename`? Will help you write tests later..

MattSturgeon: The normal code flow would be

```ruby
basename = arr.pop
if basename == "/" then basename = arr.pop # I can't imagine this happening
                                           # more than once, but imho it
                                           # makes more sense to allow for
                                           # the possibility.
done
```

To me, `while ... break unless ...` is actually quite an elegant solution to popping off an unknown number of elements and assigning the final pop.

MattSturgeon: Sorry, didn't see your comment before replying. 3-5? Not sure.

Edge case may not be the best choice of words.

Basically our code doesn't know if `res.filename` will have a trailing `/` or not. We could go with whether it does or doesn't right now, but it may change in the future or (without analysing the WEBrick code) be inconsistent.

Essentially I'm treating it as unsanitised input.

ashmaroli: or `until`
```ruby
basename == arr.pop until basename == "/"
```

ashmaroli: have you tried `until`?

```ruby
basename == arr.pop until basename == "/"
```

MattSturgeon: `hostname == arr.pop until hostname == "/"` doesn't assign the value of `arr.pop` to `hostname`. Also `hostname` will be undefined when evaluating the `until` condition.

Looking at the `set_filename` implementation, (which sets up the `res.filename` and calls `search_index_file`), `res.filename` is fairly sane and unlikely to change.

It looks like the filename shouldn't have any trailing slashes (so `basename = arr.pop` without testing for trailing slashes may work) when we get in `search_index_file`, but I haven't tested.

Still, the `while` method isn't complex or slow and it makes our method more robust and resilient to change.

MattSturgeon: I could use index instead of arrays, something like [this](/MattSturgeon/jekyll/commit/8f52910e):

```ruby
# Index of final / that isn't followed by another / or EOL
# Could just be rindex('/') if we don't care about trailing slashes
i = res.filename.rindex %r!/(?!/|$)!
basename = res.filename[i..-1]
res.filename = res.filename[0, i]
```

ashmaroli:
FYI, Currently, Jekyll's merge bot will squash all commits irrespective of the `--fixup` flag

lib/jekyll/commands/serve/servlet.rb
```diff
@@ -21,10 +21,37 @@ def initialize(server, root, callbacks)
...
+            res.filename = res.filename[0, index]
+
+            # Try and find a file named dirname.html in the parent directory.
+            file = search_file(req, res, basename + ".html")
+
+            # If we didn't find a file, revert our changes to res.filename.
+            res.filename << basename unless file
+          end
+
+          return file
```
pup: you technically don't need the `return` statement, since ruby returns the last statement in a function anyways

MattSturgeon:
Ok, good to know. I think this is probably ready now, I've pushed up a slightly re-worded commit with all changes including switching to index based splitting:

```ruby
i = res.filename.rindex "\/"
basename = res.filename[i..-1]
res.filename = res.filename[0, i]
```

And simply appending `basename` back onto `filename` instead of copying a backup.

From what I could tell, trailing slashes are unlikely to ever be an issue with WEBrick so I shifted to simply splitting on the index of the final `/` char without messing around with arrays or loops.

lib/jekyll/commands/serve/servlet.rb
MattSturgeon: But the last statement is `res.filename << basename`, `unless file` is evaluated first, right?

ashmaroli:
What we need now is tests.. :smiling_imp:

MattSturgeon:
Where are the existing tests for the `serve` command?

ashmaroli:
https://github.com/jekyll/jekyll/blob/master/test/test_commands_serve.rb

MattSturgeon:
How did I miss that? I did look, I promise!

lib/jekyll/commands/serve/servlet.rb
pup: i agree!

lib/jekyll/commands/serve/servlet.rb
```diff
@@ -21,10 +21,36 @@ def initialize(server, root, callbacks)
         # up with a different preference on which comes first.
 
         def search_file(req, res, basename)
-          # /file.* > /file/index.html > /file.html
+          # /file.* -> /file.html
```
ashmaroli: please undo this change.

MattSturgeon:
I'm probably missing something, but I'm not sure there's any testable changes here. Short of running the server, and making a mini client within the test suit to assert that HTTP requests return expected results.

The tests for command serve seem to be mostly sanity checking configuration options to make sure config files and command line switches have taken affect.

I'd need to run the command asynchronously, then fire off a HTTP request to a few URLs that I know are either directories, files or both and assert that the result matches the local file I wanted.

Is this within scope of the tests? It sounds more complex than the entire feature to me. It doesn't seem to be something that was added for #3452

What kind of things should I be testing for this change?

lib/jekyll/commands/serve/servlet.rb
```diff
@@ -21,10 +21,36 @@ def initialize(server, root, callbacks)
...
           super || super(req, res, "#{basename}.html")
         end
 
+        def search_index_file(req, res)
+          # /file/index.html -> /file.html
+
+          # First, let's see if the our super method can figure it out.
```
ashmaroli: small typo => `..if the our..`

lib/jekyll/commands/serve/servlet.rb
```diff
@@ -21,10 +21,36 @@ def initialize(server, root, callbacks)
...
+            # Ok, I guess that didn't work, I guess there's no basename/index.html
+            # Let's look for basename.html instead...
+
+            # Split the basename from the path (res.filename)
+            i = res.filename.rindex "\/"
+            basename = res.filename[i..-1]
+            res.filename = res.filename[0, i]
+
+            # Try and find a file named dirname.html in the parent directory.
+            file = search_file(req, res, basename + ".html")
```
ashmaroli: in the really rare chance that `basename` is not a `String`, this will fail. Instead interpolate:
```diff
- file = search_file(req, res, basename + ".html")
+ file = search_file(req, res, "#{basename}.html")
```

lib/jekyll/commands/serve/servlet.rb
MattSturgeon: Any reason why? It was an intended change, since `search_file` _doesn't_ change "/file" into "/file/index.html" (that's `search_index_file`'s job)

ashmaroli:
> What kind of things should I be testing for this change?

Mainly, if the original issue you reported has been fixed and stays fixed..

MattSturgeon:
> >What kind of things should I be testing for this change?
>
>Mainly, if the original issue you reported has been fixed and stays fixed..

IMO that's not practical to test. Unless you have some test suite that can compare files served over HTTP to files on the filesystem?

lib/jekyll/commands/serve/servlet.rb
ashmaroli: I don't think the comment *meant* changing `/file.*` into `/file.index.html` and then into `/file.html` either.
Its *probably* documenting the order of traversal.

MattSturgeon: Ah, in that case I'll clarify and move it above `search_file` since it applies to both.

MattSturgeon: How's this?

```ruby
# Order of precedence:
# /file -> /file/index.html -> /file.html -> fancy-indexing -> 404.html
```

ashmaroli: I'll defer this to the original author for his input.
/cc @envygeeks

MattSturgeon: For now I have this:

```ruby
# Add the ability to tap file.html the same way that Nginx does on our
# Docker images (or on GitHub Pages.) The difference is that we might end
# up with a different preference on which comes first.
#
# Order of precedence:
# /file -> /file/index.html -> /file.html -> fancy-indexing -> 404.html

def search_file
  # First see if we can find /file normally, if not try /file.html
end

def search_index_file
  # First, let's see if our super method can find an index file.
  unless
    # Ok, looks like there's no index file in the directory.
    # Let's look for directory.html instead...

    # Try and find a file named dirname.html in the parent directory.
  end
end
```

_actual code removed for brevity_

/cc @envygeeks

envygeeks:
This is a lot of complicated code for such a simple problem.

MattSturgeon:
Well the problem isn't _as_ simple as it sounds, but I wouldn't say the solution was complex or even large (10 lines).

`set_filename` passes us a directory path (`res.filename`) and it expects us to return a file from within that directory to use as an index file (usually `index.html`).

We want to abuse this slightly and return a file from the parent directory of the one we have, so we have to modify `res.filename` and look for files in the modified directory path.

If we fail, we undo our modification so that we don't break WEBrick's Fancy Indexing.

Are you saying that it isn't worth this amount of complexity? Also, are you happy with the modification to your `/file.* > /file/index.html > /file.html` comment?

envygeeks:
No, I'm not happy with the comment change, because it's arbitrary.  I'm happy you learned Ruby and welcome to the Ruby community! This can be fixed by changing search file to `super(req, res, ".html") || super || super(req, res, "#{basename}.html")` without the extra 9 lines of code.

Sorry it took me so long to respond.

MattSturgeon:
Ah, much simpler. Probably should be ordered `super || super(req, res, ".html") || super(req, res, "#{basename}.html")` to prioritise the default.

As soon as I noticed `search_index_file` was used when searching a directory, I focused on that and didn't notice that you could just return `".html"` from `search_file` and it would concatenate into a sane filename.

I've added your changes.

envygeeks:
No, it should be in the order I put it in, as that was my intention, and it was my intention with this source to begin with.  You should also rebase, squash, and only have commits in your name, it is bad form to commit in another's name, it should instead be credited in the extended message (if you wish,) as committing in my name forces me to have to say "that commit is not mine, it's not signed" in the future, should a problem arise, and it also alters the blame in a confusing way on Github.  Sorry to be weird.

Thanks!

envygeeks:
I guess what I am saying is, I do wish to take blame if something goes wrong (and I'll happily help fix it,) but the commit should not be in my name, you should instead opt to link to my comment in the extended commit, so that people know that the source was originally my idea, and it's my fault if something is wrong.

MattSturgeon:
I didn't commit in your name, I listed you as the author of the changes, that's what `--author` is for, committing other's changes. It still list's me as the committer, so I still get the blame for any issues with the commit.

```
commit 1dc2bc74580060d4f00dd4a45e02a70574a56672
tree 4207e3ea2b7b52733fb6d3e3a85d5cc00d6b6522
parent 56546a28fda95098c44b31090050db0f415be574
author Jordon Bedwell <jordon@********> 1500420691 +0100
committer Matt Sturgeon <matt@********> 1500521221 +0100

    Fix serving files that clash with directories
```

---

Why did you want it to check ".html" before the default check of `basename`? Surely `basename` -> ".html" -> "`basename`.html" or `basename` -> "`basename`.html" -> ".html" makes more sense?

Either way since it is your change now, it is probably easiest if you make a separate PR.

parkr:
Thank you, all! I'm guessing the comment directly above this line is fine?

parkr:
Additionally, a test would be nice.

pathawks:
> Additionally, a test would be nice.

Suggestions would be appreciated. See above comment:

> I'm probably missing something, but I'm not sure there's any testable changes here. Short of running the server, and making a mini client within the test suit to assert that HTTP requests return expected results.
> ...
> What kind of things should I be testing for this change?

envygeeks:
This source was never really tested to begin with, I don't know if I would force that constraint on the new author, I'll take responsibility outside of this pull request if necessary, it shouldn't really fall on him.

parkr:
Ok! I'd love to get that module tested so we don't break this behaviour in the future, but it's not a super high priority. Thanks, all!

@jekyllbot: merge +minor

envygeeks:
This commit is not what my code originally was, what it intended, and it still has my name stamped all over it as the author even though the person who used my name falsely refused to fix the problem, it is merged now and my name is associated with something I didn't write, please fix this problem.

MattSturgeon:
@envygeeks your name is not on the merged commit (ec84bec6d60ce2a6d475054395fce2b1253dbff3)

image

PR #6236, opened 2017-07-21
Small markdown lint (mdl) fixes for about.md

antgel:


mattr-:
Thanks for the contribution! Jekyll doesn't conform to strict markdown, but rather GitHub Flavored Markdown (GFM), which allows for longer than 80 column lines. Does mdl have a GFM mode that you could use instead?

antgel:
Got it. mdl can be configured to ignore specific rules, but I wouldn't ignore the long line rule, precisely because long lines are painful to work with (for me). I guess you'll be closing this. :)

ashmaroli:
@antgel some powerful text-editors (e.g Sublime Text) allow you to configure word-wrap-column width which will softly wrap long lines to your liking.

antgel:
Of course, I use vim - the pain is not the display issue caused by potential horizontal scrolling. It's that with a really long line, if I want to get to the middle, the up and down keys skip over the line and it is like skipping a paragraph.

mattr-:
Vim has sentence motions that you can use to navigate between sentences which make long lines easier to deal with. They are `(` and `)` for backwards and forwards respectively. Hope this helps and thanks again for your contribution.

PR #6239, opened 2017-07-25
exit site.process sooner

ashmaroli:
```ruby
  # site.rb
    def print_stats
      if @config["profile"]
        puts @liquid_renderer.stats_table
      end
    end
```
Since `site.print_stats` *further runs* only if `@config["profile"]`, *I feel* it'd be better if the method is **called only in the special scenario.**

The drawback here is the duplicate `if` guard unchanged to prevent breaking downstream use of `site.print_stats` if any

/cc @jekyll/stability

### DIFF
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -73,10 +73,11 @@ module Jekyll
       render
       cleanup
       write
-      print_stats
+      print_stats if @config["profile"]
     end
 
     def print_stats
+      # TODO: remove duplicate guard statement in next major version change
       if @config["profile"]
         puts @liquid_renderer.stats_table
       end
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -73,12 +73,12 @@ module Jekyll
       render
       cleanup
       write
-      print_stats if @config["profile"]
+      print_stats if config["profile"]
     end
 
     def print_stats
       # TODO: remove duplicate guard statement in next major version change
-      if @config["profile"]
+      if config["profile"]
         puts @liquid_renderer.stats_table
       end
     end
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -77,10 +77,7 @@ module Jekyll
     end
 
     def print_stats
-      # TODO: remove duplicate guard statement in next major version change
-      if config["profile"]
-        puts @liquid_renderer.stats_table
-      end
+      puts @liquid_renderer.stats_table
     end
 
     # Reset Site details.


ashmaroli:
Ci failures unrelated.. addressed in #6240

pup:
@ashmaroli wanna merge from master now that #6240 is in?

pup:
that being said, i'm not sure whether this is useful. doesn't it make stack traces harder to parse since it's less specific?

ashmaroli:
> doesn't it make stack traces harder to parse

not sure if its going to be so..
@fw42 can I have your inputs on this PR? Thanks :slightly_smiling_face:

lib/jekyll/site.rb
```diff
@@ -73,10 +73,11 @@ def process
       render
       cleanup
       write
-      print_stats
+      print_stats if @config["profile"]
     end
 
     def print_stats
+      # TODO: remove duplicate guard statement in next major version change
       if @config["profile"]
```
parkr: If you're to move the `if` above, then just remove it here. That means that this method will always print stats if called.

lib/jekyll/site.rb
```diff
@@ -73,10 +73,11 @@ def process
       render
       cleanup
       write
-      print_stats
+      print_stats if @config["profile"]
```
parkr: We should try not to use `@config` – isn't there an accessor for that?

lib/jekyll/site.rb
ashmaroli: Yes, I didn't want the method to change behavior in a minor release-cycle. Hence the `TODO` comment.

lib/jekyll/site.rb
ashmaroli: Ah yes.. there's an `attr_reader` ... that should do just fine.. I'll update.

fw42:
No objections from me 👍 

I don't have a strong preference for where the `if` should be. Either works for me and I don't think this really matters too much (I would say you can probably remove the old one without a major version bump but I'll leave that up to Parker to decide).

lib/jekyll/site.rb
```diff
@@ -73,11 +73,12 @@ def process
...
       cleanup
       write
-      print_stats
+      print_stats if config["profile"]
     end
 
     def print_stats
-      if @config["profile"]
+      # TODO: remove duplicate guard statement in next major version change
+      if config["profile"]
```
parkr: Let's remove this line – I will tell anyone who complains that it was a bug to have this check in this method; if I call print_stats, it had better print stats! 😉

parkr:
Thank you!!

@jekyllbot: merge +dev

PR #6244, opened 2017-07-26
Doc y_day in docs/permalinks

osteele:
The permalink documentation references `:y_day` in the `ordinal` built-in permalink style, but does not document it. This change adds documentation for `y_day`.


DirtyF:
@jekyllbot: merge +docs

PR #6245, opened 2017-07-26
Allow placing liquid helper tags for JekyllFeed and JekyllSEOTag at all times

ashmaroli:
@pathawks Came up with a hack to ship theme templates with liquid tags required by official plugins (e.g. `jekyll-feed`, `jekyll-seo-tag`, etc)

If the plugin is not included in the build, the tag renders an empty line.
If the plugin is included, the `<link -- />` is rendered..

tested manually locally..


lib/jekyll/tags/plugin_tags.rb
```diff
@@ -0,0 +1,31 @@
+module Jekyll
+  module PluginTags
+    def render(context); end
+  end
+
+  class FeedMetaTag < Liquid::Tag
```
envygeeks: `FeedMetaTag = Class.new(Liquid::Tag)`

lib/jekyll/tags/plugin_tags.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  class SeoTag < Liquid::Tag
+  end
+end
+
+#
+
+plugin_tags = [
```
envygeeks: Maybe this should be a frozen constant on PluginTags?

lib/jekyll/tags/plugin_tags.rb
```diff
@@ -0,0 +1,31 @@
...
+    "tag_name"  => "seo",
+    "tag_class" => Jekyll::SeoTag,
+  },
+]
+
+#
+
+plugin_tags.each do |tag|
+  tag["tag_class"].include Jekyll::PluginTags
+  Liquid::Template.register_tag(tag["tag_name"], tag["tag_class"])
```
envygeeks: `*tag.values_at("tag_name", "tag_class")`  I normally use this because it will throw early with a missing key, rather than throwing later with a `nil` value, in Ruby `nil` is treated too loosely most of the time, mistakes happen!

pathawks:
I think if we are going to do this, it needs to be a much more general solution.

Jekyll could expose a Liquid tag called `head_hook` or something. Templates could include `{% head_hook %}` inside `<head>`. Plugins would notify Jekyll that they want to expose something to the `head_hook`. When the tag is called, Jekyll would then call the each of the registered plugins.

This is something that should be very carefully designed, otherwise it will be far more complicated/limited than it is worth.

parkr:
I don't get the motivation for this... Can you link me to your discussion, please? Or give me a quick summary here. Thanks!

ashmaroli:
This was the result of a sudden "Eureka!" moment and decided to send in a PR to gauge response.
I no longer feel the need to have this merged.

My current thoughts being:
```
If a theme needs to use a certain custom-tag from a plugin, then that plugin should be listed
as a `runtime_dependency` which will always be 'required' by Jekyll 3.5 and above.
```

ashmaroli:
Thank you Jordon, for your comments. I learnt something new from them..

PR #6246, opened 2017-07-26
Fix 1dc2bc74580060d4f00dd4a45e02a70574a56672

envygeeks:
That code violated the intention of this source, the intention of this source was to always serve html files before directories and only allow indexing in the case there is no clash.

### DIFF
--- a/lib/jekyll/commands/serve/servlet.rb
+++ b/lib/jekyll/commands/serve/servlet.rb
@@ -22,7 +22,7 @@ module Jekyll
 
         def search_file(req, res, basename)
           # /file.* > /file/index.html > /file.html
-          super || super(req, res, ".html") || super(req, res, "#{basename}.html")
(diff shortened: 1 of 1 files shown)


lib/jekyll/commands/serve/servlet.rb
```diff
@@ -22,7 +22,7 @@ def initialize(server, root, callbacks)
 
         def search_file(req, res, basename)
           # /file.* > /file/index.html > /file.html
```
parkr: Can you illustrate what this change does by modifying this comment?

`/.html > /file > /file.html`?

parkr:
Trying to grok the new behaviour but don't quite understand the intentioned difference here. Could you please provide an example?

/cc the other PR which @envygeeks is fixing here https://github.com/jekyll/jekyll/pull/6231

parkr:
Closing due to no response. Feel free to respond here and we can re-open.

Issue #5980, opened 2017-03-25
kramdown GFM is not enable?


garin:
I'm setting kramdown's GFM support

_config.yml

```
markdown: kramdown
kramdown:
  input: GFM
```

but unenable GFM format

### pry-byebug
https://github.com/jekyll/jekyll/blob/27ed81547b12d28a60c51961b82a5723981feb7d/lib/jekyll/converters/markdown/kramdown_parser.rb#L41
```ruby
    42: def convert(content)
    43:    binding.pry
 => 44:   Kramdown::Document.new(content, @config).to_html
    45: end

[1] pry(#<Jekyll::Converters::Markdown::KramdownParser>)> p @config
=> {"auto_ids"=>true,
 "toc_levels"=>"1..6",
 "entity_output"=>"as_char",
 "smart_quotes"=>"lsquo,rsquo,ldquo,rdquo",
 "input"=>"GFM",                             # GFM enable ? 
 "hard_wrap"=>false,
 "footnote_nr"=>1,
 "syntax_highlighter"=>"rouge",
 "syntax_highlighter_opts"=>{},
 "coderay"=>{}}

with all @config optoins
> Kramdown::Document.new("aaa\nbbb", @config).to_html
=> "<p>aaa\n" + "bbb</p>\n"           # not with linebreak

input GMF option only 
> Kramdown::Document.new("aaa\nbbb", input: 'GFM').to_html
=> "<p>aaa<br />\n" + "bbb</p>\n"     # with linebreak. (I want this behavior)
```

### environment
* debian-9
* ruby-2.4.1
* jekylly-3.4.3
* kramdown-1.13.2

parkr:
Hm, this might be a question for @gettalong. It looks like we're using kramdown properly, based on your pry output.

gettalong:
@parkr @garin Looks like the config uses the string key "input" instead of the symbol :input.

parkr:
That would do it! Thank you, @gettalong.

I believe our code turns these keys into symbols... If not, we should add a test & symbolize the keys!

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>3.3-stable</code> or <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

[...]

PR #6247, opened 2017-07-26
kramdown: symbolize keys in-place

parkr:
Currently-released version of Jekyll uses default_proc to assign values for the kramdown config. `Kramdown::Document#initialize` calls `Options.merge` which does `Hash#each` over our hash which doesn't have symbolized keys. To fix this, we need to add the symbolized keys into the hash itself.

Fixes https://github.com/jekyll/jekyll/issues/5980.

/cc @jekyll/stability

### DIFF
--- a/lib/jekyll/converters/markdown/kramdown_parser.rb
+++ b/lib/jekyll/converters/markdown/kramdown_parser.rb
@@ -43,13 +43,9 @@ module Jekyll
 
         private
         def make_accessible(hash = @config)
-          proc_ = proc { |hash_, key| hash_[key.to_s] if key.is_a?(Symbol) }
-          hash.default_proc = proc_
-
-          hash.each do |_, val|
-            make_accessible val if val.is_a?(
-              Hash
-            )
+          hash.keys.each do |key|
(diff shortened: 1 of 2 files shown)


lib/jekyll/converters/markdown/kramdown_parser.rb
```diff
@@ -43,13 +43,9 @@ def convert(content)
...
         def make_accessible(hash = @config)
-          proc_ = proc { |hash_, key| hash_[key.to_s] if key.is_a?(Symbol) }
-          hash.default_proc = proc_
-
-          hash.each do |_, val|
-            make_accessible val if val.is_a?(
-              Hash
-            )
+          hash.keys.each do |key|
+            hash[key.to_sym] = hash[key]
```
pathawks:
```ruby
hash[key.to_sym] = hash[key] unless key.is_a?(Symbol)
```

?

parkr: It should be a no-op since Symbol#to_sym returns self, right?

parkr:
@jekyllbot: merge +fix

PR #6253, opened 2017-07-27
Call to_s on site.url before attempting to concatenate strings

benbalter:
We call `to_s` before we call `site.baseurl`, but we don't call `to_s` before we call `site.url`, meaning if `site.url` is something other than a string, things blow up.

The specific use case for the `to_s` here, other than users that set `url` to something else, is the ability to stub `site.url` with a Proc that would lazily calculate the URL when called.

### DIFF
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -13,7 +13,7 @@ module Jekyll
         return input if Addressable::URI.parse(input).absolute?
         site = @context.registers[:site]
         return relative_url(input).to_s if site.config["url"].nil?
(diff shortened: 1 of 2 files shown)


parkr:
In `test/test_filters.rb` there should be a section for URL-related filters. Throw in something that would break Addressable::URI.parse and just make sure it ends up as a string. I would recommend using URI or some class that has a built-in `to_s`, or you could build a tiny custom class for this purpose. If you're not able to get to writing a test, I can see about adding one for this myself.

Also a quick fmt error. 😸

parkr:
Added 2 tests: one for `site.url` and one to ensure the same behaviour for `site.baseurl`.

parkr:
@jekyllbot: merge +fix

PR #6255, opened 2017-07-29
Docs: fix links to issues in History.markdown

### LINKED ISSUE #6254: Links to repository on history page are broken
[History](https://jekyllrb.com/docs/history/) links are in this mangled format:

https://jekyllrb.com/docs/history/%7B%7B%20site.repository%20%7D%7D/issues/...

/cc @jekyll/documentation

DirtyF:
fix #6254 

/cc @jekyll/documentation

### DIFF
--- a/History.markdown
+++ b/History.markdown
@@ -52,7 +52,7 @@
   * add plugins for multiple page pagination (#6055)
   * Update minimum Ruby version in installation.md (#6164)
   * [docs] Add information about finding a collection in `site.collections` (#6165)
-  * Add {%raw%} to Liquid example on site (#6179)
+  * Add `{% raw %}` to Liquid example on site (#6179)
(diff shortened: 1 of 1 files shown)


ashmaroli:
@DirtyF I just realized, Liquid doesn't care about backticks..
The solution here is to remove / replace the `{% raw %} tag or even try enclosing it in a *raw* block..

DirtyF:
@ashmaroli Thanks for your concern, but I did test this change locally before committing and it worked as expected.

It's because the source file `History.markdown` goes through a [liquid_escape](https://github.com/jekyll/jekyll/blob/master/Rakefile#L62-L64) function when it generates the output file in docs.

parkr:
@jekyllbot: merge +docs

PR #6256, opened 2017-07-31
Accept unicode characters in data subfolder filenames

aivuk:
Without this change, if you have a subfolder in _data/ that contains an unicode character in its name, e.g. _data/example/prüfung/content.yml, sanitize_filename() would just remove this character from the object key site.data.example['prfung']. With the change in this pull request, you can access the subfolder correctly with site.data.example['prüfung'].


pup:
is there a test for this? if not, there should be one

parkr:
Hm... it looks like we don't have tests in `test/` for this module. 🤦‍♂️ They're all in `features/`!

Let's go ahead and add a `test/test_data_reader.rb` with:

```ruby
# frozen_string_literal: true

require "helper"

class TestDataReader < JekyllUnitTest
  context "#sanitize_filename" do
    setup do
      @reader = DataReader.new(fixture_site)
    end

    should "remove evil characters" do
      assert_equal "helpwhathaveIdone", @reader.sanitize_filename(
        "help/what^&$^#*(!^%*!#haveId&&&&&&&&&one"
      )
    end

    should "allow unicode alphabet characters" do
      assert_equal "prüfung.content", @reader.sanitize_filename("prüfung.content")
    end
  end
end
```

For this to work as well, update `sanitized_filename` to include some handling of frozen strings:

```ruby
def sanitize_filename(name)
  name = name.dup if name.frozen?
  # ... the gsub!'s go here
end
```

pathawks:
@DirtyF What happened? Zero files changed, or merge gone wrong?

DirtyF:
@pathawks I merged the changes introduced by @parkr with a first test for `DataReader`. That means that @aivuk has now to add another test for the proposed change, then modify `sanitize_file` to allow unicode chars and submit his changes. Sorry for any inconvenience.

ashmaroli:
@DirtyF IMO, since you attempted to resolve the merge conflicts in this branch, you should have resolved it by incorporating the author's original changes instead of re-setting the branch to Jekyll's `master`

lib/jekyll/readers/data_reader.rb
```diff
@@ -70,8 +70,8 @@ def read_data_file(path)
     end
 
     def sanitize_filename(name)
-      name.gsub!(%r![^\p{Word}\s-]+|(?<=^|\b\s)\s+(?=$|\s?\b)!, "".freeze)
```
parkr: The problem is here: this merge commit should not have changed `\p{Word}`

pathawks:
Can we either fix this bad merge and add tests, or close this PR and try again?

PR #6260, opened 2017-08-02
Gemfile and Gemfile.lock needed to run jekyll serve

moxiegirl:
I was able to run `jekyll serve .` with these files in place. If I hide them out by renaming them *then* the command fails.


moxiegirl:
The sentence reads:

```
 If your site doesn’t have these Gemfiles, you can omit bundle exec and just run jekyll serve.
```

This is equivalent to saying

```
If you site has the Gemfiles, you cannot omit the bundle exec and just run jekyll serve . 
```

It appears from your comment that the documentation should read:

**If you omit the `bundle exec` prefix and  *the Gemfiles* are present, Jekyll ignores these files when building the local site.** 

Would that be correct? I haven't done exhaustive testing...just started here but the original sentence was so wack to parse it caused a full stop when I was reading it.

pathawks:
If you do not have a Gemfile, but have installed Jekyll with `gem install jekyll`, then it is *possible* though not recommended to run Jekyll directly instead of through Bundler.

I think using Bundler is definitely the recommended way to install/update Jekyll, so we should for sure encourage that. Any wording that makes that easier for users is 👌🏼

moxiegirl:
Ty.  I'll tweak to make this clear and update this PR.

ashmaroli:
@moxiegirl this change is wrong.
If you have a `Gemfile` or a `Gemfile.lock` then Bundler helps to `build` or `serve` your site within the context of the versions specified in those files, otherwise Jekyll uses the latest version of the gems installed on your machine which may or may not be compatible with Jekyll or its plugins

ashmaroli:
Additionally, `jekyll serve` doesn't take arguments.. just options (a.k.a switches. e.g. `--port 8080`). So your premise regarding `jekyll serve .` is basically flawed..

DirtyF:
What's the intention here? We don't want to advise to run jekyll without bundle exec

We can redirect to bundler documentation to learn about [Gemfile](https://bundler.io/v1.16/man/gemfile.5.html) and [bundle exec](https://bundler.io/v1.16/man/bundle-exec.1.html).

bundler is shipped by default since Ruby 2.5, so this procedure will become standard eventually.

parkr:
I think the problem here is this code: 

https://github.com/jekyll/jekyll/blob/7e31e274fbb1d1367908bf0203c68a396f855809/lib/jekyll/plugin_manager.rb#L46-L60

We only check for a `Gemfile`, but not a `Gemfile.lock`. I believe `Bundler.setup` explodes if a `Gemfile.lock` isn't present.

ashmaroli:
Closing as the change here is fundamentally wrong, or quite the opposite of what's actually observed..

PR #6261, opened 2017-08-02
Enhance 'link' tag to include variables support

alexey-pelykh:
This would allow to use
```
{% assign path = 'some/variable/path.md' %}
{% link {{path}} %}
```
in addition to
```
{% link some/hardcoded/path.md %}
```
(similar to `include` tag)


parkr:
What do you think about

```text
{% assign path = 'some/variable/path.md' %}
{% link path %}
```

?

alexey-pelykh:
@parkr I think it's better to leave a marker of a variable to avoid misreading + is there a chance that I'll have an extensionless file `path`? Maybe it would be great to support something like
```
{% assign path = 'some/variable/path.md' %}
{% link '{{site.locale}}{{path}}' %}
```
?

And how to treat spaces between marker `{{` and first character of the variable?
Yesterday I got a bit confused about jekyll-assets plugin since `{% include {{ variable }} %}`(note the spaces) is ok, yet `{% image {{\ variable\ }} %}` (note the escaping). That inconsistency broke my mind for a couple of minutes :)

parkr:
Ah, yes! Also, I realized that this functionality is basically the same as for the `include` tag (you can do `{% include {{path}} %}` so the syntax lines up OK. I really think we should just run the whole input string (e.g. `{{path}}`) through Liquid 😉  At any rate, I wonder if we can reuse some of that code here?

alexey-pelykh:
Since I've used `include` tag as a reference :) You mean parse path like in #6269?

alexey-pelykh:
@parkr what's next?

alexey-pelykh:
@parkr rebased

pathawks:
Closing in favor of #6269

PR #6263, opened 2017-08-03
Replace complex if-block in `Jekyll::Filters`

ashmaroli:
Refactoring a private method `as_liquid` in `Jekyll::Filters` module with guard clauses.


ashmaroli:
@pathawks  I didn't intend to change behaviors. Tests passed locally.
Any particular case you want me to test?

lib/jekyll/filters.rb
```diff
@@ -382,17 +382,10 @@ def as_liquid(item)
...
-          else
-            as_liquid(liquidated)
-          end
-        else
-          item
-        end
+        # prevent infinite recursion for simple types (which return `self`) when
+        #   `item` responds to :to_liquid
+        return item if !item.respond_to?(:to_liquid) || (item.to_liquid == item)
+        as_liquid(item.to_liquid)
```
pathawks: It looks like this is calling `item.to_liquid()` twice, instead of caching the results.

I agree that the two line refactor _looks_ nicer, but it is not so good for performance.

DirtyF: Is there a benchmark we could rely on to measure such a change?

pathawks: I’m not sure

Crunch09: Created a small benchmark for it: https://gist.github.com/Crunch09/b5d96d730829c6baf395739349ac1c93 (test case is mostly copied form `test_filter.rb`). @pathawks is right, master is clearly quicker.

DirtyF:
Based on the feedback, I'm gonna close this. Thanks for the effort though @ashmaroli ❤️

lib/jekyll/filters.rb
ashmaroli: Thank you for the benchmark code, @Crunch09! Always wondered if one could easily measure the impact of a refactor on performance..

ashmaroli:
> Thanks for the effort though

You're welcome!
If one were to meet halfway from the two sides (cleaner yet fast), then I think using `ternary operator` is a nice bet
```ruby
          liquidated = item.to_liquid
          # prevent infinite recursion for simple types (which return `self`)
          liquidated == item ? item : as_liquid(liquidated)
        else
          item
        end
```

But I won't be proceeding with it..

PR #6265, opened 2017-08-03
Enforce Style/FrozenStringLiteralComment

parkr:
The frozen_string_literal: true magic comment in Ruby can help
dramatically decrease memory allocations for new strings and can thusly
speed up your program. The intent here is for Jekyll to use less memory
and make fewer memory allocations (which must later be GC'd).

Using the output of `GC.stats` from `script/stackprof`:

Before:

```text
total_allocated_pages: 67284
total_allocated_objects: 27434203
total_freed_objects: 23567
```

After:

```text
total_allocated_pages: 63008
total_allocated_objects: 25690804
total_freed_objects: 22907
```

With this patch, we allocate 4,276 fewer [pages](https://www.speedshop.co/2017/03/09/a-guide-to-gc-stat.html), and 1,743,399 fewer objects. Holy smokes! That's _huge_ savings.

/cc @jekyll/stability


docs/_docs/history.md
```diff
@@ -27,7 +27,7 @@ note: This file is autogenerated. Edit /History.markdown instead.
 - add plugins for multiple page pagination ([#6055]({{ site.repository }}/issues/6055))
 - Update minimum Ruby version in installation.md ([#6164]({{ site.repository }}/issues/6164))
 - [docs] Add information about finding a collection in `site.collections` ([#6165]({{ site.repository }}/issues/6165))
-- Add {%raw%} to Liquid example on site ([#6179]({{ site.repository }}/issues/6179))
+- Add {% raw %}`{% raw %}`{% endraw %} to Liquid example on site ([#6179]({{ site.repository }}/issues/6179))
```
pathawks: This probably doesn't belong in this PR
?

exe/jekyll
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env ruby
+# frozen_string_literal: true
+#
```
pathawks: This should probably be an empty line

lib/jekyll/mime.types
```diff
@@ -744,7 +759,7 @@ text/x-fortran                                                            f for
 text/x-handlebars-template                                                hbs
 text/x-java-source                                                        java
 text/x-lua                                                                lua
-text/x-markdown                                                           markdown md mkd
+text/x-markdown                                                           mkd
```
pathawks: Mime type stuff probably doesn't belong in this PR
?

parkr:
Updated, thanks for the review @pathawks!

@jekyllbot: merge +fix

PR #6266, opened 2017-08-04
Memoize the return value of Document#url

parkr:
Running `bundle exec rake site:generate` locally, this brings generation time from 16 seconds down to 8 seconds.

```text
done in 16.421 seconds.
```

```text
done in 8.685 seconds.
```

### DIFF
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -204,7 +204,7 @@ module Jekyll
     #
     # Returns the computed URL for the document.
     def url
-      @url = URL.new({
+      @url ||= URL.new({
         :template     => url_template,
         :placeholders => url_placeholders,
         :permalink    => permalink,


parkr:
Lol @ ruby. Also lol @ us not catching this! If you haven't checked out stackprof, it is _so_ cool.

```console
$ export BENCHMARK=1
$ bundle install
$ script/stackprof # measures cpu
$ script/stackprof object # measures memory
```

parkr:
@jekyllbot:  merge +fix

pup:
@jekyllbot: merge +fix

pup:
@parkr could it be that jekyllbot doesn't accept comments that have multiple spaces in them (like yours did)?

parkr:
Perhaps! I'm not sure. Thanks for submitting the correction for me :)

ianthehenry:
This breaks, e.g., generators that programatically calculate permalinks -- you must now explicitly un-set the `url` property after setting the permalink to forget the memoized value.

I feel that as it used to be a dependent property, changing any of the values it depends on should invalidate the cache. But I don't know how to do that when it depends on values stored in hashes...

parkr:
Hey @ianthehenry! Can you give a simple example? We could set the `@url` value to `nil` if we can catch the changes. This change is important to keep, as it is a significant performance gain.

ianthehenry:
Here was a plugin that I had ~~written~~ copied almost verbatim from StackOverflow:

```ruby
# Because we put our posts in their own directories:
#
# _posts/2017-05-11-foo/index.md
#
# Instead of their own files:
#
# _posts/2017-05-11-foo.md
#
# The default slug is "foo/index". But this is annoying, so we strip it.

module Jekyll
  class PermalinkRewriter < Generator
    safe true
    priority :low

    def generate(site)
      site.posts.docs.each do |post|
        unless post.data.member?('permalink')
          post.data['permalink'] = post.data['slug'].sub(/\/index$/, '') + '/'
        end
      end
    end
  end
end
```

After upgrading to 3.5.2, I had to make this tweak:

```ruby
module Jekyll
  class Document
    def invalidate_url
      @url = nil
    end
  end

  class PermalinkRewriter < Generator
    safe true
    priority :low

    def generate(site)
      site.posts.docs.each do |post|
        unless post.data.member?('permalink')
          post.data['permalink'] = post.data['slug'].sub(/[\/-]index$/, '') + '/'
          # As of 3.5.2, post.url is memoized, and not recalculated when one of
          # the properties it's based on changes.
          #
          # https://github.com/jekyll/jekyll/pull/6266
          post.invalidate_url
        end
      end
    end
  end
end
```

Which, you know, was after a bit of head-scratching and then looking for probable commit messages.

I suppose I could also just override the `url` getter on a post. That feels less hacky than this, I suppose. But yeah; it was a thing that used to work that stopped working.

PR #5854, opened 2017-02-03
Support wildcards in scope => path

alexey-pelykh:
When this may be useful: when you have two different set of files (e.g. *.doc & *.pdf) in the same folder that requires different values used by Jekyll plugins or Jekyll itself.
Examples:
```
defaults:
  # Content
  -
    scope:
      path: "content/*.pdf"
    values:
      sitemap: false
  -
    scope:
      path: "content/*.html"
    values:
      sitemap: true
```
or
```
defaults:
  # Offers
  -
    scope:
      path: "offers/*/index.html"
    values:
      layout: offer
  -
    scope:
      path: "offers/*/thank-you.html"
    values:
      sitemap: false
```

PR #6268, opened 2017-08-04
Scope path glob

alexey-pelykh:
Same as #5854 

When this may be useful: when you have two different set of files (e.g. *.doc & *.pdf) in the same folder that requires different values used by Jekyll plugins or Jekyll itself.
Examples:
```
defaults:
  # Content
  -
    scope:
      path: "content/*.pdf"
    values:
      sitemap: false
  -
    scope:
      path: "content/*.html"
    values:
      sitemap: true
```
or
```
defaults:
  # Offers
  -
    scope:
      path: "offers/*/index.html"
    values:
      layout: offer
  -
    scope:
      path: "offers/*/thank-you.html"
    values:
      sitemap: false
```

### DIFF
--- a/lib/jekyll/frontmatter_defaults.rb
+++ b/lib/jekyll/frontmatter_defaults.rb
@@ -100,12 +100,18 @@ module Jekyll
     def applies_path?(scope, path)
       return true if !scope.key?("path") || scope["path"].empty?
 
-      scope_path = Pathname.new(scope["path"])
-      Pathname.new(sanitize_path(path)).ascend do |ascended_path|
-        if ascended_path.to_s == scope_path.to_s
-          return true
+      site_path = Pathname.new(@site.source)
+      rel_scope_path = Pathname.new(scope["path"])
+      abs_scope_path = File.join(@site.source, rel_scope_path)
+      Dir.glob(abs_scope_path).each do |scope_path|
+        scope_path = Pathname.new(scope_path).relative_path_from site_path
+        Pathname.new(sanitize_path(path)).ascend do |ascended_path|
+          if ascended_path.to_s == scope_path.to_s
+            return true
+          end
         end
       end
+      return false
     end
 
     # Determines whether the scope applies to type.
--- a/lib/jekyll/frontmatter_defaults.rb
+++ b/lib/jekyll/frontmatter_defaults.rb
@@ -100,17 +100,26 @@ module Jekyll
     def applies_path?(scope, path)
       return true if !scope.key?("path") || scope["path"].empty?
 
+      sanitized_path = sanitize_path(path)
+
       site_path = Pathname.new(@site.source)
       rel_scope_path = Pathname.new(scope["path"])
       abs_scope_path = File.join(@site.source, rel_scope_path)
       Dir.glob(abs_scope_path).each do |scope_path|
         scope_path = Pathname.new(scope_path).relative_path_from site_path
-        Pathname.new(sanitize_path(path)).ascend do |ascended_path|
+        Pathname.new(sanitized_path).ascend do |ascended_path|
           if ascended_path.to_s == scope_path.to_s
             return true
           end
         end
       end
+
+      Pathname.new(sanitized_path).ascend do |ascended_path|
+        if ascended_path.to_s == rel_scope_path.to_s
+          return true
+        end
+      end
+
       return false
     end
 
--- a/lib/jekyll/frontmatter_defaults.rb
+++ b/lib/jekyll/frontmatter_defaults.rb
@@ -100,27 +100,27 @@ module Jekyll
     def applies_path?(scope, path)
       return true if !scope.key?("path") || scope["path"].empty?
 
-      sanitized_path = sanitize_path(path)
+      sanitized_path = Pathname.new(sanitize_path(path))
 
       site_path = Pathname.new(@site.source)
       rel_scope_path = Pathname.new(scope["path"])
       abs_scope_path = File.join(@site.source, rel_scope_path)
       Dir.glob(abs_scope_path).each do |scope_path|
         scope_path = Pathname.new(scope_path).relative_path_from site_path
-        Pathname.new(sanitized_path).ascend do |ascended_path|
-          if ascended_path.to_s == scope_path.to_s
-            return true
-          end
-        end
+        return true if path_is_subpath?(sanitized_path, scope_path)
       end
 
-      Pathname.new(sanitized_path).ascend do |ascended_path|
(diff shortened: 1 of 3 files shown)


test/test_front_matter_defaults.rb
```diff
@@ -27,6 +27,30 @@ class TestFrontMatterDefaults < JekyllUnitTest
     end
   end
 
+  context "A site with full front matter defaults (glob)" do
+    setup do
+      @site = fixture_site({
+        "defaults" => [{
+          "scope"  => {
+            "path" => "contacts/*.html",
```
parkr: What happens if I have `*.html` here? Should it apply to `about.html` *and* `contacts/index.html`?

lib/jekyll/frontmatter_defaults.rb
```diff
@@ -100,12 +100,27 @@ def applies?(scope, path, type)
...
 
-      scope_path = Pathname.new(scope["path"])
-      Pathname.new(sanitize_path(path)).ascend do |ascended_path|
-        if ascended_path.to_s == scope_path.to_s
+      sanitized_path = sanitize_path(path)
+
+      site_path = Pathname.new(@site.source)
+      rel_scope_path = Pathname.new(scope["path"])
+      abs_scope_path = File.join(@site.source, rel_scope_path)
+      Dir.glob(abs_scope_path).each do |scope_path|
```
parkr: We have some code in `EntryFilter#glob_include?` we might be able to use here.

test/test_front_matter_defaults.rb
alexey-pelykh: Since we're using `Dir.glob`, according to [docs](https://ruby-doc.org/core-2.2.0/Dir.html#method-c-glob), `*.html` will match only files on top directory, to match _any_ HTML file, a `**/*.html` pattern should we used

parkr: That's what would happen with the current code, yeah. I'm wondering if it makes sense to have `*.html` apply to all files in my entire source that end in `.html`. If I were using globs, that's what I would expect.

alexey-pelykh: Seems to me that `EntryFilter#glob_include` behaves in a similar way since `File.fnmatch` uses same rules of matching. Making `*.html` to match all files recursively will limit usability of the feature since if I want to set values only for root html files for whatever reason and not for any other - it will be impossible to achieve. Anyhow, this behaviour should be documented somewhere?

parkr: It should definitely be documented! If you wanted to apply to just root, would `/*.html` work?

lib/jekyll/frontmatter_defaults.rb
alexey-pelykh: Looked through that code, it relies on `File#fnmatch?` that uses the same rules as `Dir#glob`, though `EntryFilter#glob_include?` itself is not super-useful in this code, it would take more lines of code to reuse `glob_include?`

test/test_front_matter_defaults.rb
alexey-pelykh: Well, it would, though it would definitely take more custom logic that using `Dir#glob` and then there would be question why `EntryFilter` treats `*.html` differently

alexey-pelykh: Regarding the documentation - where should I write what?

alexey-pelykh:
@parkr what's next?

alexey-pelykh:
@parkr rebased

DirtyF:
@alexey-pelykh That's great, could you add a usage example in our documentation?

ashmaroli:
> If you can refactor it's great, if not just add the offended file in .rubocop.yml

Do not add the file to `.rubocop.yml`.
Instead refactor the two `Pathname.new(sanitized_path).ascend { |path| [...] }` into a private method..

That should appease Rubocop..
Also the last `return` isn't necessary.. Ruby always returns the last statement evaluated..

alexey-pelykh:
@DirtyF to which section should I add that example?
@ashmaroli hope it will work better now

lib/jekyll/frontmatter_defaults.rb
```diff
@@ -100,12 +100,27 @@ def applies?(scope, path, type)
...
+        scope_path = Pathname.new(scope_path).relative_path_from site_path
+        return true if path_is_subpath?(sanitized_path, scope_path)
+      end
+
+      path_is_subpath?(sanitized_path, rel_scope_path)
+    end
+
+    def path_is_subpath?(path, parent_path)
```
ashmaroli: do you want to expose this method as a `public` method? or would it rather be a `private` helper method?

lib/jekyll/frontmatter_defaults.rb
```diff
@@ -100,12 +100,27 @@ def applies?(scope, path, type)
...
+      path.ascend do |ascended_path|
+        if ascended_path.to_s == parent_path.to_s
           return true
         end
```
ashmaroli: you made a mistake while committing this patch..
its now a method within a method..

lib/jekyll/frontmatter_defaults.rb
alexey-pelykh: @ashmaroli Isn't there a private scope starting right before `applies?`

lib/jekyll/frontmatter_defaults.rb
alexey-pelykh: @ashmaroli are you sure?

lib/jekyll/frontmatter_defaults.rb
ashmaroli: right.. my bad.. ignore..

ashmaroli: how does the following read?
```ruby
    def path_is_subpath?(path, parent_path)
      path.ascend do |ascended_path|
        ascended_path.to_s == parent_path.to_s
      end
    end
```

alexey-pelykh:
@DirtyF docs and example added

DirtyF:
@jekyllbot: merge +minor

letrastudio:
Is it possible to attach this to a release milestone yet?

Found it mentioned in the documentation (it was exactly what I was looking for!), sad to see it's unreleased. I can provide my specific usecase if it helps.

ashmaroli:
@letrastudio this has already been merged to the `master` branch. It'll be included in the next release. If you want, you can use it right now by pointing your Gemfile to the `master` branch:
```ruby
gem "jekyll", :git => "https://github.com/jekyll/jekyll.git"
```

PR #6269, opened 2017-08-04
Liquefied link tag

alexey-pelykh:
This would allow to use

{% assign path = 'some/variable/path' %}
{% assign ext = 'md' %}
{% link {{path}}.{{ext}} %}
in addition to

{% link some/hardcoded/path.md %}


danpalmer:
Some feedback on this, I think this could be confusing and might not be the best choice for implementation.

The use-case here (please correct me if I'm wrong) is to enable linking to dynamic files. Currently, `link` can only take a hard-coded relative path, but we want to be able to pass something else in.

This implementation re-uses liquid templating, nested within the process of templating another file, in a way which I think could get confusing and be a maintenance burden in the future.

I'd suggest that instead of templating the data passed to `link`, we should allow looking it up as a variable name in the current context. If we did this, your example could be achieved in a similar way with the following code:

```
{% assign path = 'some/variable/path' %}
{% assign fullpath = path|append: ".md" %}
{% link fullpath %}

{% link some/hardcoded/path.md %}
```

This re-uses existing functionality in the templating system to achieve the string building, rather than introducing another layer of templating, and I think this will be simpler to document and maintain.

I'd be interested to hear what you and the maintainers think of these different solutions, and let me know if you want to do this, otherwise I might open a PR with an implementation.

alexey-pelykh:
This implementation is based on #6261, which itself uses the same approach as `include` tag. Current one is more similar to jekyll-assets tags. Anyhow, writing one line instead of three seems more convenient, even if that requires extra depth of templating.

alexey-pelykh:
@danpalmer updated

DirtyF:
@alexey-pelykh Could you document a use case for this? Thanks.

alexey-pelykh:
@DirtyF sure, to which page that should go?

DirtyF:
@alexey-pelykh https://jekyllrb.com/docs/templates/#links please.

alexey-pelykh:
@DirtyF done

lib/jekyll/tags/link.rb
```diff
@@ -18,14 +18,17 @@ def initialize(tag_name, relative_path, tokens)
       def render(context)
         site = context.registers[:site]
 
+        liquid = site.liquid_renderer.file("(jekyll:link)")
```
ashmaroli: @alexey-pelykh can you please explain what `"(jekyll:link)"` is supposed to mean here? That is beyond my Ruby knowledge..

pathawks:
Is this a duplicate of #6261? Seems like one or the other should be closed.

alexey-pelykh:
@pathawks #6261 is super-simple, while this one is "as it should be"

pathawks:
@alexey-pelykh Are you saying _that_ should be closed in favor of _this_?

alexey-pelykh:
Which-ever implementation suits project vision better! I prefer #6269 over #6261 - check the test cases, this one supports more sophisticated usage

pathawks:
Could somebody from @jekyll/windows kick AppVeyor? It seems to have choked.

Otherwise this LGTM :+1:

ashmaroli:
> Could somebody from @jekyll/windows kick AppVeyor? It seems to have choked.

AppVeyor builds cannot be restarted.. The best solution here is pushing the `merge 'master` branch into..` commit.

pathawks:
Thanks for all the work you put into this. Sorry it took so long to get this merged, but this will be a great feature in Jekyll 4.0 :+1::beers:

@jekyllbot: :shipit: +minor

PR #6274, opened 2017-08-04
Execute build command using bundle.

Ryuno-Ki:
Hi,

I just created my first automatic deployment using Jekyll and Git post-receive hook.
Thereby I spotted an error, which leads to situation like described [on StackOverflow](https://stackoverflow.com/q/44836330) (I had the same situation on a CentOS box).

Luckily the fix is quite easy, so I wrote a PR to save others research of an evening ^^


pup:
@jekyllbot: merge +docs

Issue #6232, opened 2017-07-19
[minor][tests] static file last modified time test is minorly flawed


MattSturgeon:
**This almost certainly doesn't matter**, but I noticed a test incorrectly fail due to this earlier.

I had the following error from `cibuild`: static file should know its last modification time (Expected: 1500423023 Actual: 1500423022). Notice that the test asserted that time should be 1 millisecond later than the actual time.

The [test code](https://github.com/jekyll/jekyll/blob/f91b61/test/test_static_file.rb#L134-L136) asserts if the current time equals the file's modification time.

So random chance caused the millisecond to tick over between the test setup and running the assertion.

Possible solutions may be to use a coarser time (seconds instead of milliseconds?) or to cache the time from immediately after creating the file.

You could even save the time from before creating the file and assert that the modification time is _between_ the saved time and the assertion time to be completely safe.

Anyway, now that I've reported it, **feel free to close as WONTFIX** if you feel the added complexity isn't worth it. After all the test only exists to highlight potential errors (and it is easy to spot false negatives like this).

pathawks:
I hate tests that sometimes fail for no good reason. We should fix this.

Good catch.

ashmaroli:
A PR to fix this has been merged. Please close.

PR #6277, opened 2017-08-06
make flakey test more robust

Crunch09:
Sometimes `Time.new` is not the same time as the creation of `@static_file` anymore. See #6232 for more information. Thanks to @MattSturgeon for noticing it!

cc: @jekyll/build


mattr-:
@jekyllbot: merge +dev

PR #6278, opened 2017-08-07
docs: Jekyll 3 paginate instructions update

nikolas:
* Changed the unmaintained `jekyll-paginate` gem to `jekyll-paginate-v2` - https://github.com/sverrirs/jekyll-paginate-v2
* Added explicit example of how to add this to your Gemfile

### DIFF
--- a/docs/_docs/pagination.md
+++ b/docs/_docs/pagination.md
@@ -8,8 +8,14 @@ break the main listing of posts up into smaller lists and display them over
 multiple pages. Jekyll offers a pagination plugin, so you can automatically
 generate the appropriate files and folders you need for paginated listings.
 
-For Jekyll 3, include the `jekyll-paginate` plugin in your Gemfile and in
+For Jekyll 3, include the `jekyll-paginate-v2` plugin in your Gemfile and in
 your `_config.yml` under `plugins`. For Jekyll 2, this is standard.
+
+```yaml
+group :jekyll_plugins do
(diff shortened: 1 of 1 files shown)


ashmaroli:
Do note that GitHub Pages do not support the `v2` gem.. only the older `jekyll-paginate`

/cc @jekyll/documentation

nikolas:
Ah okay, then we should note that in the docs and have examples for both `jekyll-paginate` and `jekyll-paginate-v2`.

pathawks:
@nikolas Thanks for the PR! :beers:

AFAIK We don't own or control **jekyll-paginate-v2**. I'm not sure that we are responsible for its documentation.

DirtyF:
@nikolas What we could do is mention @sverrirs plugin and explain why one might want to use `jekyll-paginate-v2`, i.e. for handling categories and tags pagination. For now this is not an official plugin and it's not supported by GitHub Pages, so we can't just make it the defaults in our docs.

If you want to take a shot at it, we'll happily review your submission. Thanks for bringing this to our attention. 👈

sverrirs:
Thanks for the mention. 

Just a FYI, I would be more than happy include more people on the jekyll-paginate-v2 plugin project if you think including it would make a positive contribution to the official distribution. 

I'm also willing to do the work necessary to bring it up to any standards you might require of official plugins.

DirtyF:
https://jekyllrb.com/docs/pagination/#enable-pagination now mentions @sverris `jekyll-paginate-v2`.

PR #6281, opened 2017-08-08
Fix Drop#key? so it can handle a nil argument

parkr:
In Liquid 4.0, `Liquid::VariableLookup#evaulate` [checks `input.key?` when it's evaluating](https://github.com/Shopify/liquid/blob/v4.0.0/lib/liquid/variable_lookup.rb#L44). If the `key` is `nil`, it throws the following error:

```text
  Liquid Exception: nil is not a symbol nor a string in <template>
```

We should simply return `false` for `Drop#key?(nil)`.

/cc @jekyll/stability


parkr:
@jekyllbot: merge +fix

PR #6284, opened 2017-08-10
Add a quick test for DataReader

parkr:
In #6256, I noticed that we had no tests whatsoever for DataReader.

We should definitely go back and ensure all our classes/modules are tested. Perhaps we can reorganize the test directory too, so the paths/filenames are predictable.

/cc @jekyll/stability


parkr:
We should add more tests but at least this gets the job started! 

@jekyllbot: merge +dev

PR #6285, opened 2017-08-10
Backport #6280 for v3.5.x: Guard against type error in absolute url

parkr:
This backports #6280.

Conflicting files:


parkr:
Our script is broken.

PR #6299, opened 2017-08-12
make the sort filter stable

### LINKED ISSUE #6275: Seemingly inconsistent sorting

---

---

- [x] It was another bug.

## My Reproduction Steps

I can't seem to really reproduce this.

Refer to these two websites: http://bi.tardis.ed.ac.uk/inf2 and https://betterinformatics.com/inf2

The latter URL has the correct sorting, whereas the first one (bi.tardis.ed.ac.uk) **appears** to sort sections (courses) randomly.

The code sorting these "sections" is handled by this line [`{% assign sections = site.sections | where: "year", name | sort: "title" | sort: "ordering" %}
`](https://github.com/compsoc-edinburgh/betterinformatics/blob/master/_layouts/frontpage.html#L19)

On betterinformatics.com, it filters sections based on the year (`where: "year", name`), sorts those by title (`sort: "title"`) and then sorts the remaining by priority (`sort: "ordering"`).

On bi.tardis.ed.ac.uk, it appears to sort by ordering, and completely garbage the sort by title, messing up the original title sorting. When removing the sort by title, the ordering doesn't change at all, so it appears that only the last sort is being applied.

- bi.tardis.ed.ac.uk is running jekyll 3.5.1 (issue on this machine), running in docker, ruby 2.3.3p222
- betterinformatics.com is running jekyll 3.5.1 , ruby 2.3.1-2~16.04

In general, say you have a list of items

- Bob
- Alex
- Christina (ordering: -5)
- Dani

The above snippet for sorting should form:

- Christina (ordering: -5)
- Alex
- Bob
- Dani 

And that sorting works fine on most platforms. However, on bi.tardis.ed.ac.uk, only the last sort is applied, mixing up the rest of the items:

- Christina (ordering: -5)
- ???
- ???
- ???

I also give all sections a default ordering value of 0. https://github.com/compsoc-edinburgh/betterinformatics/blob/master/_config.yml#L22

Both websites are running the same the same github.com/compsoc-edinburgh/betterinformatics code, but perform sorting differently..

/cc @jekyll/build

Crunch09:
It's hard to write a failing test case for the old behavior, as `Array#sort` doesn't always move equal elements around. Also happy to write a benchmark for it, i just wasn't sure what would be a proper test case for the benchmark.
(I hope) this fixes #6275 

cc: @jekyll/stability

### DIFF
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -336,7 +336,7 @@ module Jekyll
 
     private
     def sort_input(input, property, order)
-      input.sort do |apple, orange|
+      input.each_with_index.sort do |(apple, apple_ix), (orange, orange_ix)|
         apple_property = item_property(apple, property)
         orange_property = item_property(orange, property)
 
@@ -344,10 +344,12 @@ module Jekyll
           - order
         elsif apple_property.nil? && !orange_property.nil?
           + order
+        elsif apple_property == orange_property
+          apple_ix <=> orange_ix
         else
           apple_property <=> orange_property
         end
-      end
+      end.map(&:first)
     end
 
     private
--- a/test/test_filters.rb
+++ b/test/test_filters.rb
@@ -1019,6 +1019,15 @@ class TestFilters < JekyllUnitTest
           @filter.sort([{ "a" => { "b" => 2 } }, { "a" => { "b" => 1 } },
                         { "a" => { "b" => 3 } }, ], "a.b")
       end
+      should "return sorted by property array with stable order" do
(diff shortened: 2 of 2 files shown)


parkr:
Thanks for this! We're currently revamping how we sort a bit so we may have to revisit this in a bit to handle the Schwartzian transform.

mattr-:
Thanks for your work @Crunch09!

With the schwartzian transform work done now, I'm going to close this as it needs to be revamped to work with that updated sort handling and I would prefer to see that done in a separate PR.

PR #6305, opened 2017-08-12
Loggers should accept both numbers and symbols.

envygeeks:
Jekyll's logger should start trying to conform to standard logging practices in Ruby.  That said, the logger level should accept both a `:symbol` (if you really want,) and a number, this way users aren't taken aback when they know what log level they want.  This commit makes it so that the logger level can be set normally and within Jekyll's own constraints.

```ruby
Jekyll.logger.log_level = 0
Jekyll.logger.log_level = Logger::DEBUG
Jekyll.logger.log_level = :debug
```

### DIFF
--- a/lib/jekyll/log_adapter.rb
+++ b/lib/jekyll/log_adapter.rb
@@ -29,7 +29,9 @@ module Jekyll
     #
     # Returns nothing
     def log_level=(level)
-      writer.level = LOG_LEVELS.fetch(level)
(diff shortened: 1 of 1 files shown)


mattr-:
:thumbsup: I agree that we should be doing this.

@envygeeks Would you mind rebasing this to remove the conflicts?

pathawks:
Moving to #6967

PR #6306, opened 2017-08-12
Add the ability to `yield` a message.

envygeeks:
Standard loggers in Ruby mostly `yield` their messages (this is the default behavior of Ruby's own logger.) Or if there is a block then the message is the topic, a colon is automatically added, and the message is shipped with the topic.  Jekyll should conform to this behavior so that it's logger can be passed to libraries that expect this standard behavior.

```ruby
[1] pry(main)> Jekyll.logger.info("Just so you know") # => Just so you know 
[2] pry(main)> Jekyll.logger.info("Just so you know", "this is a message") # => Just so you know this is a message
[3] pry(main)> Jekyll.logger.info("Just so you know") { "this is a message" }
  # => Just so you know: this is a message
```

### DIFF
--- a/lib/jekyll/log_adapter.rb
+++ b/lib/jekyll/log_adapter.rb
@@ -48,8 +48,8 @@ module Jekyll
     # message - the message detail
     #
     # Returns nothing
-    def debug(topic, message = nil)
-      writer.debug(message(topic, message))
+    def debug(topic, message = nil, &block)
+      writer.debug(message(topic, message, &block))
     end
 
     # Public: Print a message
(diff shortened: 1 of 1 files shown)


lib/jekyll/log_adapter.rb
```diff
@@ -79,7 +79,7 @@ def warn(topic, message = nil)
     #
     # Returns nothing
     def error(topic, message = nil)
-      writer.error(message(topic, message))
+      writer.error(message(topic, message, &block))
```
parkr: Need to add this to the method signature on line 81

lib/jekyll/log_adapter.rb
```diff
@@ -69,7 +69,7 @@ def info(topic, message = nil)
     #
     # Returns nothing
     def warn(topic, message = nil)
-      writer.warn(message(topic, message))
+      writer.warn(message(topic, message, &block))
```
parkr: Need to add this to the method signature on line 71 (&block)

parkr:
The original repository is gone so no edits can be made. Rolled this change into https://github.com/jekyll/jekyll/pull/6315.

PR #6311, opened 2017-08-14
trim and relocate plugin info across docs

### LINKED ISSUE #6312: revisit templating page's info on tags
The templates page talks about  "a few handy filters and tags" specific of Jekyll. This suggests that there are more than are listed on the page, but does not point in the direction of an authoritative reference.
In #6311, I've tried phrasing things a little differently, but I might have stopped short, not knowing with certainty whether the page is already the comprehensive listing I think it should be.
I also moved the explanation for the jekyll-gist plugin from this page to the plugins where it makes more sense to have it, even though at the moment, it is the only specific plugin that gets more than a line of explanation. So although it's not the perfect fit on that page, it seems closer to home.

/cc @jekyll/documentation

Strangehill:
fix #6312

### DIFF
--- a/docs/_docs/plugins.md
+++ b/docs/_docs/plugins.md
@@ -31,21 +31,26 @@ Jekyll generates your site.
 values of the gem names of the plugins you'd like to use. An example:
 
 
-        plugins: [jekyll-coffeescript, jekyll-watch, jekyll-assets]
+        plugins:
+          - jekyll-gist
+          - jekyll-coffeescript
+          - jekyll-assets
+          - another-jekyll-plugin
         # This will require each of these plugins automatically.
 
-    Then install your plugins using `gem install jekyll-coffeescript jekyll-watch jekyll-assets`
+    Then install your plugins using `gem install jekyll-gist jekyll-coffeescript jekyll-assets another-jekyll-plugin`
 
 3. Add the relevant plugins to a Bundler group in your `Gemfile`. An
     example:
 
         group :jekyll_plugins do
-          gem "my-jekyll-plugin"
+          gem "jekyll-gist"
+          gem "jekyll-coffeescript"
+          gem "jekyll-assets"
           gem "another-jekyll-plugin"
         end
 
-    Now you need to install all plugins from your Bundler group by running single command `bundle install`
-
+    Now you need to install all plugins from your Bundler group by running single command `bundle install`.
 
 <div class="note info">
   <h5>
@@ -59,13 +64,15 @@ values of the gem names of the plugins you'd like to use. An example:
   </p>
 </div>
 
-In general, plugins you make will fall into one of five categories:
+In general, plugins you make will fall broadly into one of five categories:
 
 1. [Generators](#generators)
 2. [Converters](#converters)
 3. [Commands](#commands)
 4. [Tags](#tags)
 5. [Hooks](#hooks)
+
+See the bottom of the page for a [list of available plugins](#available-plugins)
 
 ## Generators
 
@@ -810,6 +817,9 @@ LESS.js files during generation.
 
 #### Tags
 
+You can find a few useful plugins at the following locations:
+
+- [Jekyll-gist](https://github.com/jekyll/jekyll-gist): Use the `gist` tag to easily embed a GitHub Gist onto your site. This works with public or secret gists.
 - [Asset Path Tag](https://github.com/samrayner/jekyll-asset-path-plugin) by [Sam Rayner](http://www.samrayner.com/): Allows organisation of assets into subdirectories by outputting a path for a given file relative to the current post or page.
 - [Delicious Plugin by Christian Hellsten](https://github.com/christianhellsten/jekyll-plugins): Fetches and renders bookmarks from delicious.com.
 - [Ultraviolet Plugin by Steve Alex](https://gist.github.com/480380): Jekyll tag for the [Ultraviolet](https://github.com/grosser/ultraviolet) code highligher.
@@ -929,14 +939,6 @@ LESS.js files during generation.
 - [Jekyll-Post](https://github.com/robcrocombe/jekyll-post): A CLI tool to easily draft, edit, and publish Jekyll posts.
 - [jekyll-numbered-headings](https://github.com/muratayusuke/jekyll-numbered-headings): Adds ordered number to headings.
 - [jekyll-pre-commit](https://github.com/mpchadwick/jekyll-pre-commit): A framework for running checks against your posts using a git pre-commit hook before you publish them.
-
-#### Editors
-
-- [sublime-jekyll](https://github.com/23maverick23/sublime-jekyll): A Sublime Text package for Jekyll static sites. This package should help creating Jekyll sites and posts easier by providing access to key template tags and filters, as well as common completions and a current date/datetime command (for dating posts). You can install this package manually via GitHub, or via [Package Control](https://packagecontrol.io/packages/Jekyll).
-- [vim-jekyll](https://github.com/parkr/vim-jekyll): A vim plugin to generate
-  new posts and run `jekyll build` all without leaving vim.
-- [markdown-writer](https://atom.io/packages/markdown-writer): An Atom package for Jekyll. It can create new posts/drafts, manage tags/categories, insert link/images and add many useful key mappings.
-- [Wordpress2Jekyll](https://wordpress.org/plugins/wp2jekyll/): A Wordpress plugin that allows you to use Wordpress as your editor and (automatically) export content in to Jekyll. WordPress2Jekyll attempts to marry these two systems together in order to make a site that can be easily managed from all devices.
 
 <div class="note info">
   <h5>Jekyll Plugins Wanted</h5>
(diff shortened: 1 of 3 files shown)


docs/_docs/templates.md
```diff
@@ -416,9 +416,14 @@ The default is `default`. They are as follows (with what they filter):
 - `default`: spaces and non-alphanumeric characters
 - `pretty`: spaces and non-alphanumeric characters except for `._~!$&'()+,;=@`
 
-## Tags
+## Jekyll Tags
```
ashmaroli: changing this heading alone will break any url bookmarked to former `/templates/#tags`

docs/_docs/templates.md
```diff
@@ -416,9 +416,14 @@ The default is `default`. They are as follows (with what they filter):
...
 
-### Includes
+* [Includes](#include)
+* [Code snippet highlighting](#highlight)
+* [Linking to pages, collections and posts (the new and improved way)](#link)
+* [Linking to posts (the old way)](#post_url)
+
+### Includes {#include}
```
ashmaroli: this `id` should be `#includes` for the same reason as above.

docs/_docs/templates.md
Strangehill: aah, I see. Best to change this back.

docs/_docs/templates.md
Strangehill: and changes these back, and I'm not sure I'd like to introduce new anchors, so I'm getting rid of any I introduced.

docs/_docs/templates.md
```diff
@@ -6,8 +6,8 @@ permalink: /docs/templates/
 Jekyll uses the [Liquid](https://shopify.github.io/liquid/) templating language to
 process templates. All of the standard Liquid [tags](https://shopify.github.io/liquid/tags/control-flow/) and
 [filters](https://shopify.github.io/liquid/filters/abs/) are
-supported. Jekyll even adds a few handy filters and tags of its own to make
-common tasks easier.
+supported. Jekyll even adds four new tags and even more handy filters of its own to make
```
DirtyF: No need to mention they are four tags. This number could change, and we won't remember to update this, we could link to the aforementioned tags if we really wanna be helpful.

docs/_docs/plugins.md
```diff
@@ -722,6 +725,28 @@ The complete list of available hooks is below:
 
 ## Available Plugins
 
+### Gist
+
+The `gist` plugin is an example of tag plugin. To use it you'll need to add the
```
DirtyF: I would just drop this. There's already an example of a tag plugin here: https://jekyllrb.com/docs/plugins/#tags

If people want to know how to use a plugin, they can go check the plugin's README.

docs/_docs/plugins.md
```diff
@@ -59,14 +60,16 @@ values of the gem names of the plugins you'd like to use. An example:
...
-In general, plugins you make will fall into one of five categories:
+In general, plugins you make will fall broadly into one of five categories:
 
 1. [Generators](#generators)
 2. [Converters](#converters)
 3. [Commands](#commands)
 4. [Tags](#tags)
 5. [Hooks](#hooks)
 
+See the bottom of the page for available examples of each.
```
DirtyF: This feels unnecessary to me.

docs/_docs/plugins.md
```diff
@@ -31,21 +31,22 @@ Jekyll generates your site.
 values of the gem names of the plugins you'd like to use. An example:
 
 
-        plugins: [jekyll-coffeescript, jekyll-watch, jekyll-assets]
+        plugins: [jekyll-gist, jekyll-coffeescript, jekyll-assets, another-jekyll-plugin]
         # This will require each of these plugins automatically.
 
-    Then install your plugins using `gem install jekyll-coffeescript jekyll-watch jekyll-assets`
+    Then install your plugins using `gem install jekyll-gist jekyll-coffeescript jekyll-assets another-jekyll-plugin`
```
DirtyF: Let's follow http://ben.balter.com/jekyll-style-guide/config/#arrays here.

docs/_docs/plugins.md
```diff
@@ -31,21 +31,22 @@ Jekyll generates your site.
...
-          gem "my-jekyll-plugin"
+          gem "jekyll-gist"
+          gem "jekyll-coffeescript"
+          gem "jekyll-assets"
           gem "another-jekyll-plugin"
         end
 
-    Now you need to install all plugins from your Bundler group by running single command `bundle install`
-
+    Now you need to install all plugins from your Bundler group by running single command `bundle install`. See below to find out what different plugins, like "jekyll-gist" can do.
```
DirtyF:
```diff
-- See below to find out what different plugins, like "jekyll-gist" can do.
```

docs/_docs/templates.md
Strangehill: I meant it to be a hint for the reader that this list holds all jekyll tags and filters. I see how the four could easily become outdated when new tags come along. I'm removing the four and rephrasing again to make clearer that this list if 'the' list of jekyll specific tags and filters.

docs/_docs/plugins.md
Strangehill: much better. done.

docs/_docs/plugins.md
Strangehill: yes. since it's not in the available plugin list at the bottom, I'm making it a one liner and add it there instead.

docs/_docs/plugins.md
Strangehill: What I'm hoping for is some indication to the reader of what can be found on this page. Perhaps not this solution, but I think it's a good idea to way to somehow inform the reader, near the top of the page before going deep, about the contents of the page, especially so far below the fold. [Alternatively...](https://github.com/jekyll/jekyll/issues/6312#issuecomment-322353194).

DirtyF: So let's add the missing navigation item then:

```diff 
++ 6. List of available plugins(#available-plugins)
```

FYI This section is due to move to a specific website (see https://dirtyf.github.io/directory/ for the current WIP)

docs/_docs/templates.md
```diff
@@ -418,6 +419,11 @@ The default is `default`. They are as follows (with what they filter):
 
 ## Tags
 
+* Includes
+* Code snippet highlighting
+* Linking to pages, collections and posts (the new and improved way)
+* Linking to posts (the old way)
+
```
DirtyF: DId you mean to add a toc here? Because links to anchors are missing.

Strangehill: Yes, but I had changed several of the sections' headers not realizing it would break links, so I took those changes back and didn't add links to this TOC. I'm adding them now.

docs/_docs/plugins.md
```diff
@@ -810,6 +814,9 @@ LESS.js files during generation.
 
 #### Tags
 
+You can find a few useful plugins at the following locations:
+
+- [Jekyll-gist by Parkr](https://github.com/jekyll/jekyll-gist): Use the `gist` tag to easily embed a GitHub Gist onto your site. This works with public or secret gists.
```
parkr: Just use `jekyll-gist` here – it's a community project!

docs/_docs/plugins.md
```diff
@@ -59,13 +64,14 @@ values of the gem names of the plugins you'd like to use. An example:
...
 
-In general, plugins you make will fall into one of five categories:
+In general, plugins you make will fall broadly into one of five categories:
 
 1. [Generators](#generators)
 2. [Converters](#converters)
 3. [Commands](#commands)
 4. [Tags](#tags)
 5. [Hooks](#hooks)
+6. [List of available plugins](#available-plugins)
```
parkr: This isn't a category, so this link should go elsewhere. Additionally, we should really move this list out of this page – it's clutter! /cc #5771

docs/_docs/plugins.md
```diff
@@ -722,8 +728,6 @@ The complete list of available hooks is below:
 
 ## Available Plugins
 
-You can find a few useful plugins at the following locations:
```
parkr: I liked this sentence. Any strong reason it can't be there?

Strangehill: A mistake! Glad you caught that.  I'm putting it back.

docs/_docs/plugins.md
Strangehill: that makes sense... but where should the link go?
I added it as a sentence at first, but it had no link, maybe putting it back with the link makes it work? 
```
 1. [Generators](#generators)
 2. [Converters](#converters)
 3. [Commands](#commands)
 4. [Tags](#tags)
 5. [Hooks](#hooks)

See the bottom of the page for a [List of available plugins](#available-plugins).
```
/cc @DirtyF

docs/_docs/plugins.md
Strangehill: move the list of available plugins to a gist, then embed that here? sounds good. a temporary move till plugins directory is ready?

DirtyF: What Parrkr meant was simply to remove "by Parkr" from the link to `jekyll-gist`

Strangehill: Ah. Of course. I'll do that and squash commits.

parkr:
@jekyllbot: merge +docs

PR #6315, opened 2017-08-17
Allow `yield` to logger methods & bail early on no-op messages

### LINKED ISSUE #6313: Jekyll storing messages in Logger, it's not using, slowing down builds, +mem
https://github.com/jekyll/jekyll/blob/6b8de2a75794be30d9816cf44e914088fca9a31e/lib/jekyll/log_adapter.rb#L102-L106

When we patch out `messages << msg`
```
mem%s: 145576
mem%e: 167840
real	0m32.676s
user	0m6.324s
sys	0m0.400s
```

Without the patch
```
mem%s: 151000
mem%e: 175708
real	0m44.494s
user	0m7.072s
sys	0m0.548s
```

This test is very low quality, considering we disabled 11/12 of our logging.  This problem was discovered when our builds were consuming upwards of 500mb capping out our memory limits for Jekyll builds, that's a lot of memory for something doing such a simple task.

parkr:
This subsumes #6306 and fixes the bugs with that PR.

It also bails early for `debug`, `info`, `warn`, and `error` if the logger's log level is higher than the message's log level. In this case the logging writer will skip it, so it's best to skip it here. It reduces the number of strings we create and greatly reduces the size of the `messages` array. This fixes #6313.

### DIFF
--- a/lib/jekyll/log_adapter.rb
+++ b/lib/jekyll/log_adapter.rb
@@ -2,7 +2,7 @@
 
 module Jekyll
   class LogAdapter
-    attr_reader :writer, :messages
+    attr_reader :writer, :messages, :level
 
     LOG_LEVELS = {
       :debug => ::Logger::DEBUG,
@@ -30,6 +30,7 @@ module Jekyll
     # Returns nothing
     def log_level=(level)
       writer.level = LOG_LEVELS.fetch(level)
+      @level = level
     end
 
     def adjust_verbosity(options = {})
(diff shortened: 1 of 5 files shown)


DirtyF:
Maybe @envygeeks wants to do a proper benchmark to see how this affect its build performance?

With Jekyll's `script/test`

Before:
```sh
real	2m24.422s
user	1m56.664s
sys	0m11.537s
```

After: 
```
real	2m6.318s
user	1m42.051s
sys	0m9.632s
```

parkr:
Whoa, who knew logging could be so expensive! I just remembered that we can push this even further by setting the log level to `:error` by default in the tests.

parkr:
@jekyllbot: merge +minor

PR #5722, opened 2017-01-05
Date filters should never raise an exception

jekyllbot:
PR automatically created for @pathawks.

```text
Failing test: date filter should allow empty input to pass through
```

PR #6316, opened 2017-08-18
Date filters should pass through empty strings with whitespace

sparanoid:
Getting `Invalid Date: '"\n \n \n\n \n \n \n"' is not a valid datetime` exception after upgrading from Jekyll 3.4.3 to 3.5.2, this PR should fix this problem. 

Related PR https://github.com/jekyll/jekyll/pull/5722

### DIFF
--- a/test/test_filters.rb
+++ b/test/test_filters.rb
@@ -295,6 +295,13 @@ class TestFilters < JekyllUnitTest
         end
       end
 
+      context "with whitespace" do
+        should "return input" do
+          assert_nil(@filter.date_to_xmlschema(nil))
+          assert_equal("  ", @filter.date_to_xmlschema(""))
+        end
+      end
+
       context "without input" do
         should "return input" do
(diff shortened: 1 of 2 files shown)


pathawks:
It seems like the real problem is trying to use a Date filter with something that is not a Date.

test/test_filters.rb
```diff
@@ -295,6 +295,13 @@ def select; end
         end
       end
 
+      context "with whitespace" do
+        should "return input" do
+          assert_nil(@filter.date_to_xmlschema(nil))
+          assert_equal("  ", @filter.date_to_xmlschema(""))
```
parkr: It appears this assertion is invalid. Check out the CI output to see more. Looks like it returns `""`. Did you mean to pass `"  "` to `date_to_xmlschema`?

We should test all 3 methods you touched, too.

pathawks:
@sparanoid Are you able to add tests for all 3 methods and fix CI?

PR #6326, opened 2017-08-19
Plugins is the new gems

DirtyF:
We can now use this by default.


DirtyF:
@jekyllbot: merge+minor

parkr:
This is a site/docs change, right?

DirtyF:
doh, right.

PR #6331, opened 2017-08-23
Allow the user to set collections_dir to put all collections under one subdirectory

parkr:
Replaces https://github.com/jekyll/jekyll/pull/3723 for Jekyll v3.6 if we want.

User configures as:

```yaml
collections_dir: my_collections
```

Then we look in `my_collections/_pizza` for the `pizza` collection, and `my_collections/_lasagna` for the `lasagna` collection.

/cc @jekyll/build

### DIFF
--- a/docs/_docs/configuration.md
+++ b/docs/_docs/configuration.md
@@ -602,12 +602,13 @@ file or on the command-line.
 
 ```yaml
 # Where things are
-source:       .
-destination:  ./_site
-plugins_dir:  _plugins
-layouts_dir:  _layouts
-data_dir:     _data
-includes_dir: _includes
+source:          .
+destination:     ./_site
+collections_dir: .
+plugins_dir:     _plugins
+layouts_dir:     _layouts
+data_dir:        _data
+includes_dir:    _includes
 collections:
   posts:
     output:   true
--- a/lib/jekyll/collection.rb
+++ b/lib/jekyll/collection.rb
@@ -100,7 +100,7 @@ module Jekyll
     # Returns a String containing the directory name where the collection
     #   is stored on the filesystem.
     def relative_directory
-      @relative_directory ||= "_#{label}"
+      @relative_directory ||= Pathname.new(directory).relative_path_from(Pathname.new(site.source)).to_s
     end
 
     # The full path to the directory containing the collection.
@@ -108,7 +108,7 @@ module Jekyll
     # Returns a String containing th directory name where the collection
     #   is stored on the filesystem.
     def directory
-      @directory ||= site.in_source_dir(relative_directory)
+      @directory ||= site.in_source_dir(File.join(site.config["collections_dir"], "_#{label}"))
     end
 
     # The full path to the directory containing the collection, with
--- a/lib/jekyll/collection.rb
+++ b/lib/jekyll/collection.rb
@@ -100,7 +100,9 @@ module Jekyll
     # Returns a String containing the directory name where the collection
     #   is stored on the filesystem.
     def relative_directory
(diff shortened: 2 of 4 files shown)


lib/jekyll/collection.rb
```diff
@@ -100,15 +100,15 @@ def filtered_entries
     # Returns a String containing the directory name where the collection
     #   is stored on the filesystem.
     def relative_directory
-      @relative_directory ||= "_#{label}"
+      @relative_directory ||= Pathname.new(directory).relative_path_from(Pathname.new(site.source)).to_s
```
ashmaroli: is this a breaking change?

lib/jekyll/configuration.rb
```diff
@@ -9,6 +9,7 @@ class Configuration < Hash
       # Where things are
       "source"              => Dir.pwd,
       "destination"         => File.join(Dir.pwd, "_site"),
+      "collections_dir"     => "",
```
ashmaroli: in `docs/_docs/configuration.md` **`collections_dir`** has a value of `.` (interpreted as current location) but its empty here.

lib/jekyll/collection.rb
mattr-: This isn't a breaking change because the functionality is now in the `directory` method below.

lib/jekyll/configuration.rb
mattr-: an empty string here is fine. If `site.config["collections_dir"]` is empty, then `File.join` will do the right things with the paths.

mattr-:
You do have a couple of long line violations from Rubocop but I'm willing to ignore those. ¯\\\_(ツ)\_/¯

DirtyF:
@mattr- shouldn't we add here `# rubocop:disable LineLength`then?

mattr-:
@DirtyF That would work.

lib/jekyll/configuration.rb
brandon1780: 4861789-118978

lib/jekyll/collection.rb
parkr: It'd be a breaking change if plugins did `instance_variable_set(:@relative_directory, "blah")` and hoped it would affect `directory`. It reverses their role for sure, but instance variable manipulation like that is 💯 % not support.

docs/_docs/configuration.md
```diff
@@ -602,12 +602,13 @@ file or on the command-line.
...
 # Where things are
-source:       .
-destination:  ./_site
-plugins_dir:  _plugins
-layouts_dir:  _layouts
-data_dir:     _data
-includes_dir: _includes
+source:          .
+destination:     ./_site
+collections_dir: .
```
ashmaroli: this should be `""` instead of a `.`

ccamacho:
+1

docs/_docs/configuration.md
pathawks: @ashmaroli Isn't it the current directory? (ie: same as source?)

ashmaroli: I was pointing at a discrepancy in `lib/jekyll/configuration.rb` above and the documentation here..

In `lib/jekyll/configuration.rb` above, its `{"source" => Dir.pwd, "collections_dir" => ""}` but here its `source:  .` and `collections_dir:  .`

DirtyF:
@jekyllbot: merge +minor

hosnas:
Can we have multiple directories for collections?

Something like this:

`collections_dir: ['coldir/dir1' , 'coldir/dir2']`

Issue #6177, opened 2017-06-25
Permalink date variables do not behave as described


ticky:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [x] I ran `jekyll doctor` to check my configuration
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/

---

- [x] I am on (or have tested on) ***macOS*** 10+
- [ ] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux
- [ ] I am on (or have tested on) ***Fedora*** GNU/Linux
- [ ] I am on (or have tested on) ***Arch*** GNU/Linux
- [ ] I am on (or have tested on) ***Other*** GNU/Linux
- [ ] I am on (or have tested on) ***Windows*** 10+

---

- [ ] I was trying to install.
- [ ] There is a broken Plugin API.
- [ ] I had an error on GitHub Pages, and I have reproduced it locally.
- [ ] I had an error on GitHub Pages, and GitHub Support said it was a Jekyll Bug.
- [ ] I had an error on GitHub Pages and I did not test it locally.
- [ ] I was trying to build.
- [x] It was another bug.

## My Reproduction Steps

1. Create a Jekyll site
2. Add a post as `2017-06-24-blog-post-one.md`, with no `date` property in the front-matter
3. Add a post as `2017-06-24-blog-post-two.md`, with a `date` front-matter value of `2017-06-24 22:52:00 -0700`
4. Set the time zone in `config.yml` to “Australia/Melbourne”
5. Build the site

Note that only the post with no `date` property is listed as being on the 24th of June.

## The Output I Wanted

Jekyll’s docs state that permalinks [use the date “from the post’s filename”](https://jekyllrb.com/docs/permalinks/#template-variables).

This is false in the case where the date has been set within the post.

/cc @jekyll/documentation

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>3.3-stable</code> or <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

[...]

PR #6334, opened 2017-08-26
Docs: post's date can be overriden in YAML front matter

DirtyF:
From https://github.com/jekyll/jekyll/issues/6177#issuecomment-325160280 

Here's the updated version:

> VARIABLE | DESCRIPTION
> -- | --
> year | Year from the post's filename. May be overridden via the document’s date YAML front matter
> month | Month from the post's filename. May be overridden via the document’s date YAML front matter
> i_month | Month without leading zeros from the post's filename. May be overridden via the document’s date YAML front matter
> day | Day from the post's filename. May be overridden via the document’s date YAML front matter
> i_day | Day without leading zeros from the post's filename. May be overridden via the document’s date YAML front matter
> short_year | Year without the century from the post's filename. May be overridden via the document’s date YAML front matter

### DIFF
--- a/docs/_docs/permalinks.md
+++ b/docs/_docs/permalinks.md
@@ -66,7 +66,10 @@ The following table lists the template variables available for permalinks. You c
         <p><code>year</code></p>
       </td>
       <td>
-        <p>Year from the post's filename</p>
+        <p>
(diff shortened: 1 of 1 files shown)


pathawks:
@jekyllbot: merge +docs

PR #6335, opened 2017-08-28
Docs: Remove prompt from code examples

spacecowboy1985:
I think the "~ $" that starts each line is unnecessary and on fist glance a potential (yet minor) point of confusion. Everyones terminals  come in different shapes and sizes, so I think sticking to the commands is often the best bet. Especially with such a basic and brief, section.

~ /myblog $ bundle exec jekyll serve in particular can be confusing for people following along. Given it's a quick start, many users will just be doing a copy and paste, or just typing out what they see. On a glance you could be easily mistaken in thinking "~/myblog" is there so run the command from this location and is not in fact the current directory with the command following after the "$".

It's a very minor change, but I think it will read better.

### DIFF
--- a/docs/_docs/contributing.md
+++ b/docs/_docs/contributing.md
@@ -124,19 +124,27 @@ If your contribution changes any Jekyll behavior, make sure to update the docume
 
 To run the test suite and build the gem you'll need to install Jekyll's dependencies by running the following command:
 
-<pre class="highlight"><code>$ script/bootstrap</code></pre>
+```sh
+script/bootstrap
+```
 
 Before you make any changes, run the tests and make sure that they pass (to confirm your environment is configured properly):
 
-<pre class="highlight"><code>$ script/cibuild</code></pre>
+```sh
+script/cibuild
+```
 
 If you are only updating a file in `test/`, you can use the command:
 
-<pre class="highlight"><code>$ script/test test/blah_test.rb</code></pre>
+```sh
+script/test test/blah_test.rb
+```
 
 If you are only updating a `.feature` file, you can use the command:
 
-<pre class="highlight"><code>$ script/cucumber features/blah.feature</code></pre>
+```sh
+script/cucumber features/blah.feature
+```
 
 Both `script/test` and `script/cucumber` can be run without arguments to
 run its entire respective suite.
--- a/docs/_docs/installation.md
+++ b/docs/_docs/installation.md
@@ -49,7 +49,7 @@ The best way to install Jekyll is via
 simply run the following command to install Jekyll:
 
 ```sh
-$ gem install jekyll
+gem install jekyll
 ```
 
 All of Jekyll’s gem dependencies are automatically installed by the above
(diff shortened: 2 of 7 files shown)


pathawks:
I think it's important to show *something* as the prompt, but I agree that it would be easier to understand if it were a different color. I have seen some sites where the prompt is not selectable, so it does not get in the way of copy-paste.

/cc: @jekyll/documentation

DirtyF:
Thanks @spacecowboy1985 for raising this issue. But is it a real one?

I don't mind getting rid of the prompt as long as we do it everywhere, not just here (i.e the homepage and other commands in the docs).

I guess that while adding a prompt and the current directory may feel more _real_ - like a screenshot of what you should see on your own screen - it's not necessary useful when you just want to follow a quick start guide and copy-paste commands. I've checked on few other popular software (like [Hugo](https://gohugo.io/getting-started/quick-start/), [React](https://facebook.github.io/react/docs/installation.html) and they don't add the prompt. 

Both works I guess as in my experience nobody ever complained about it and told us the prompt fooled them.

spacecowboy1985:
@DirtyF I wouldn't say this is an issue, more a potential improvement and a minor one at that. It's not a taxing one though should it be determined to have enough merit. As you say consistency is important and should extend to all other pages.

With that said I've noticed that these code examples are generally inconsistent across the site. In some doc pages you don't have the prompt, in other pages some blocks have the prompt and others do not. When a directory is changed in an example, this is in some cases reflected in the prompt, in others it is not.

For ease, I'm going to whiz through and just remove all the prompts so you can see the differences on the review.

spacecowboy1985:
On the deployment methods page, there is the server and laptops prompts which I have left in place. If all prompts are removed, perhaps remove these also and split the server and client parts and making the text more descriptive?

amzn-tomjoht:
I agree that the `$` prompt isn't necessary. Also agree that the inconsistency potentially raises confusion among users.

pathawks:
> That should be fairly simple, no?

It's just a CSS property, but it means the prompt will need to be HTML rather than MD.

parkr:
> It's just a CSS property, but it means the prompt will need to be HTML rather than MD.

That's fine. We can take the HTML that is generated by Rouge and plop it into the source instead.

DirtyF:
I noticed `user-select: none` is already in use on the frontpage with a custom HTML code, but when you triple click to select the line, spaces between the path, the prompt and the command get also selected. 

select-command

And here is what you get when you paste: 

```
 $ gem install jekyll bundler
~
```

On a rouge highlight, without any specific CSS, as we add prompt and path, they get selected:

select-rouge

So here you get the whole line selected:

```sh
~ $ gem install jekyll bundler
```

Also on pages like https://jekyllrb.com/docs/installation/ there are promptless commands, as s result it's way easier to copy commands. 

Let's not forget the users here.

DirtyF:
See also: http://www.cirosantilli.com/markdown-style-guide/#dollar-signs-in-shell-code

spacecowboy1985:
My bad guys, didn't notice that that I had created a new branch per page that had the prompt or were inconsistent. (My fault for going via the site and the improve this page links). I think I'be merged them all into branch that this pull request is on.

I know your still deciding on what you want to do with this, but this is every occurrence I can find and I hope it will be of some help.

DirtyF:
Sorry @spacecowboy1985, as you mentioned it's merely an issue, so I'm gonna close this for now.

I would like to see a PR that allows for users to be able to copy-paste a command from the docs without embedding the fake prompt though.

DirtyF:
cleaned up in #6398

spacecowboy1985:
@DirtyF, glad to see the changes went through, looks much better.

PR #6336, opened 2017-08-29
Update mime-types.

bmeck:
ran `script/vendor-mimes`

### DIFF
--- a/lib/jekyll/mime.types
+++ b/lib/jekyll/mime.types
@@ -14,7 +14,7 @@ application/cdmi-domain                                                   cdmid
 application/cdmi-object                                                   cdmio
 application/cdmi-queue                                                    cdmiq
 application/cu-seeme                                                      cu
-application/dash+xml                                                      mdp
+application/dash+xml                                                      mpd
 application/davmount+xml                                                  davmount
 application/docbook+xml                                                   dbk
 application/dssc+der                                                      dssc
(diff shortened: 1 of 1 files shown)


bmeck:
I'm not 100% sure how versioning works here, but want to make a follow up PR to https://github.com/github/pages-gem whenever this gets merged. It looks like the versions currently used are at https://pages.github.com/versions/ , and just want to be sure I make the PR to the right version if/when this gets lands in a version.

ashmaroli:
`jekyll/jekyll` follows [Semantic Versioning](http://semver.org/)

bmeck:
@ashmaroli understood, but those rules of what a breaking change are vague. I know in nodejs/node almost any changes to existing behavior are breaking, so this would be considered breaking/major?

ashmaroli:
`mime-types` is not part of the the Public API per se.. it seems to come into play only while running the default WEBrick server (`$ jekyll serve`)

But I'll defer to the word from the maintainers on this one..

parkr:
@bmeck The GitHub team updates these mime types independently - they're not pulled in from the Jekyll gem. Which mime type are you looking to update?

bmeck:
@parkr `.mjs` files to be `application/javascript`. Matches https://github.com/github/linguist/pull/3783

parkr:
@jekyllbot: merge +minor

parkr:
@bmeck Ok, I'll see if we can get an update going on Pages. No promises. The best thing to do in the future is to email GitHub Support asking us to update our MIME types for GitHub Pages. They pass it along to the appropriate team.

PR #6338, opened 2017-08-30
Alias Drop#invoke_drop to Drop#[]

benbalter:
@DirtyF discovered an interesting behavior via https://github.com/jekyll/github-metadata/pull/108#issuecomment-325817689, which I had attempted to resolve (unsuccessfully) via https://github.com/jekyll/github-metadata/pull/110.

In the particular instance, `Jekyll::Utils.deep_merge_hashes(drop, hash)` was returning a drop as expected, but without the hash values overriding the drop values. After some digging, I discovered the merge logic was correct (and calling `[key]=value`, which gets stored in `@mutations`), but that when a drop method was called, it was returning the drop's value, not the mutation.

If a drop is mutable, `#[]` [prefers the user-set value](https://github.com/jekyll/jekyll/blob/6a2a33d09d614377e62eba804496ad37ab63f427/lib/jekyll/drops/drop.rb#L41-L57) over the default (a Jekyll construct). 

`#invoke_drop(key)` which Liquid uses, however, doesn't understand the concept of mutations as it's using [`Liquid::Drop`'s native implementation](https://github.com/Shopify/liquid/blob/master/lib/liquid/drop.rb#L32-L39). This leads to scenarios where `drop.key`, `drop.invoke_drop('key')` and `drop["key"]` return different values.

Normally, `Drop#[]` is aliased to `Drop#invoke_drop` ensuring the values remain in sync. Since they both implement the same core logic `public_send key if self.class.invokable? key`, by aliasing `#invoke_drop` to `#[]`, both will respect mutability and return the same value regardless of `drop.invoke_drop('key')` or `drop['key]` is called (and allowing `Jekyll::Utils.deep_merge_hashes` to perform the expected operation).

Given the following test case:

```ruby
require 'jekyll'

class TestDrop < Jekyll::Drops::Drop
  mutable true

  def fallback_data
    @fallback_data ||= {}
  end

  def value
    "value from drop method"
  end

  alias_method :invoke_drop, :[]
end

obj = { "value" => "value from drop object" }
drop = TestDrop.new(obj)

puts "\ndrop.to_h after construction:"
puts drop.to_h

hash = { "value" => "value from hash" }
drop = drop.merge(hash)

puts "\ndrop.to_h after merge:"
puts drop.to_h

puts "\ndrop.invoke_drop('value') after merge:"
puts drop.invoke_drop('value')

puts "\ndrop['value'] after merge:"
puts drop['value']
```

Output before change:

```
drop.to_h after construction:
{"value"=>"value from drop method"}

drop.to_h after merge:
{"value"=>"value from hash"}

drop.invoke_drop('value') after merge:
value from drop method

drop['value'] after merge:
value from hash
```

Output after change:

```
drop.to_h after construction:
{"value"=>"value from drop method"}

drop.to_h after merge:
{"value"=>"value from hash"}

drop.invoke_drop('value') after merge:
value from hash

drop['value'] after merge:
value from hash
```

The one gotcha, is with this implementation is we lose the [`liquid_method_missing`](https://github.com/Shopify/liquid/blob/27c91203ab4adfde2cc5e1142bce2a267634ef8e/lib/liquid/drop.rb#L26-L30) logic, which we could encorporate into `#[]` if we want, but isn't strictly necessary in our implementation since we never raise for missing methods.

### DIFF
--- a/lib/jekyll/drops/drop.rb
+++ b/lib/jekyll/drops/drop.rb
@@ -104,7 +104,7 @@ module Jekyll
       #
       # Returns true if the given key is present
       def key?(key)
-        if self.class.mutable
+        if self.class.mutable?
           @mutations.key?(key)
         else
           !key.nil? && (respond_to?(key) || fallback_data.key?(key))
--- a/lib/jekyll/drops/drop.rb
+++ b/lib/jekyll/drops/drop.rb
@@ -212,12 +212,6 @@ module Jekyll
         return yield(key) unless block.nil?
         return default unless default.nil?
       end
-
-      private
-
-      def fallback_data
-        @fallback_data ||= {}
-      end
     end
   end
 end
--- a/lib/jekyll/drops/drop.rb
+++ b/lib/jekyll/drops/drop.rb
@@ -55,6 +55,7 @@ module Jekyll
           fallback_data[key]
         end
       end
+      alias_method :invoke_drop, :[]
 
       # Set a field in the Drop. If mutable, sets in the mutations and
       # returns. If not mutable, checks first if it's trying to override a
--- a/lib/jekyll/drops/drop.rb
+++ b/lib/jekyll/drops/drop.rb
@@ -213,6 +213,8 @@ module Jekyll
         return default unless default.nil?
       end
 
+      private
+
       def fallback_data
         @fallback_data ||= {}
       end
--- a/lib/jekyll/drops/drop.rb
+++ b/lib/jekyll/drops/drop.rb
@@ -212,6 +212,10 @@ module Jekyll
         return yield(key) unless block.nil?
         return default unless default.nil?
       end
+
+      def fallback_data
+        @fallback_data ||= {}
+      end
     end
   end
 end
--- a/lib/jekyll/drops/unified_payload_drop.rb
+++ b/lib/jekyll/drops/unified_payload_drop.rb
@@ -16,11 +16,6 @@ module Jekyll
       def site
         @site_drop ||= SiteDrop.new(@obj)
       end
-
-      private
-      def fallback_data
-        @fallback_data ||= {}
-      end
     end
   end
 end
--- a/lib/jekyll/drops/unified_payload_drop.rb
+++ b/lib/jekyll/drops/unified_payload_drop.rb
@@ -16,6 +16,11 @@ module Jekyll
       def site
         @site_drop ||= SiteDrop.new(@obj)
       end
+
+      private
+      def fallback_data
+        @fallback_data ||= {}
+      end
     end
   end
 end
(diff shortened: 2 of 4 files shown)


DirtyF:
@benbalter Great explanation, I'm a big fan of how you document issues and allow other to learn.  👏

parkr:
Whoops – `test/test_drop.rb` would be a more appropriate place to add a test for this.

benbalter:
> could you add a test 

Implemented via a9489988.

lib/jekyll/drops/drop.rb
```diff
@@ -211,6 +212,10 @@ def fetch(key, default = nil, &block)
         return yield(key) unless block.nil?
         return default unless default.nil?
       end
+
+      def fallback_data
+        @fallback_data ||= {}
+      end
```
benbalter: The base drop class assumes fallback data to exist and return a hash. Stubbed out the default value so that child classes don't each have to implement it if they don't have fallback data (including our test fixture).

lib/jekyll/drops/drop.rb
```diff
@@ -211,6 +212,12 @@ def fetch(key, default = nil, &block)
         return yield(key) unless block.nil?
         return default unless default.nil?
       end
+
+      private
+
+      def fallback_data
+        @fallback_data ||= {}
+      end
```
benbalter: The base drop class (in `#[]`, `#keys` and `#key?`) assumes fallback data to exist and return a hash. Stubbed out the default value so that child classes don't each have to implement it if they don't have fallback data (including our test fixture).

parkr: I intentionally left this method out so all drops _had_ to implement this – it was part of the design of Jekyll Drops, that some fallback data had to be provided by the developer of a drop. I'm not sure I like that going away. :/

benbalter: > it was part of the design of Jekyll Drops, that some fallback data had to be provided by the developer of a drop

Admittedly out of scope, so glad to remove from this PR. Curious if that rationale still holds true when Jekyll internally often implements it as an empty hash?

benbalter: Removed via 847ed7f

benbalter:
Fixing this one problem revealed another blocker to mutability taking precedence over methods.

Take a look at 3776c44 (failing test) and c998c59 (fix).

In short, `key?` called `mutable` not `mutable?` meaning calls to `fetch` or Liquid's own use of `key?` at render time would call `class.mutable` without any arguments. Doing so would (A) always return false, but more importantly (B) set `@is_mutable` to false, meaning fetching a mutable key would zero out all mutability (even if accessed using another method like `#[]`.

This now PR simply changes `self.class.mutable` to `self.class.mutable?` which is the getter, not the setter.

parkr:
Sounds like we should backport this to 3.5.x once it ships. 

👍 to merge once CI is passing.

benbalter:
@jekyllbot: merge +minor.

PR #6339, opened 2017-08-30
add failing test for non-utf8 encoding

Crunch09:
Based on the files provided by @DirtyF . At least the UTF-16 one did fail for me locally so hopefully it will also fail on CI. Happy to try to fix the issue but won't have time before the weekend.

cc: @jekyll/build


test/test_document.rb
```diff
@@ -7,6 +7,14 @@ def assert_equal_value(key, one, other)
     assert_equal(one[key], other[key])
   end
 
+  def setup_encoded_document(filename)
+    site = fixture_site
+    Document.new(site.in_source_dir(File.join("_encodings", filename)), {
+      :site       => site,
+      :collection => site.posts,
```
parkr: This should be an encodings collection.

Crunch09:
Thanks, good point 👍 Fixed that

DirtyF:
@jekyllbot: merge +dev

PR #6344, opened 2017-09-03
skip adding binary files as posts

### LINKED ISSUE #5181: "Liquid Exception: invalid byte sequence in UTF-8..."  with binaries
- [x] I am on (or have tested on) **_macOS**_ 10+
- [x] I was trying to build.

## My Reproduction Steps

1. Create a subdirectory under the `_posts` folder called `2016-08-02-example`.
2. Add a file `2016-08-02-example.md` to this folder. (With proper frontmatter, etc.)
3. `bundle exec jekyll serve` and all is well.
4. Add an image file (ie. `08-02-16-image.png`) to this folder.
5. `bundle exec jekyll serve` now yields an error.

With both `JEYKLL_LOG_LEVEL=debug` and the `-t` switch output is:

```
...
 Rendering Markup: _posts/2016-08-02-example/2016-08-02-example.md
         Rendering: _posts/2016-08-02-example/2016-08-02-image.png
  Pre-Render Hooks: _posts/2016-08-02-example/2016-08-02-image.png
$USER_DIR$/blog/_posts/2016-08-02-example/2016-08-02-image.png render_with_liquid?  false 
  Rendering Liquid: _posts/2016-08-02-example/2016-08-02-image.png
  Liquid Exception: invalid byte sequence in UTF-8 in $USER_DIR$/blog/_posts/2016-08-02-example/2016-08-02-image.png
$USER_DIR$/.gem/ruby/2.2.3/gems/liquid-3.0.6/lib/liquid/template.rb:232:in `split': invalid byte sequence in UTF-8 (ArgumentError)
  from $USER_DIR$/.gem/ruby/2.2.3/gems/liquid-3.0.6/lib/liquid/template.rb:232:in `tokenize'
  from $USER_DIR$/.gem/ruby/2.2.3/gems/liquid-3.0.6/lib/liquid/template.rb:122:in `parse'
  from $USER_DIR$/.gem/ruby/2.2.3/gems/liquid-3.0.6/lib/liquid/template.rb:108:in `parse'
  from $USER_DIR$/.gem/ruby/2.2.3/bundler/gems/jekyll-6d2b344c0e24/lib/jekyll/liquid_renderer/file.rb:11:in `block in parse'
  from $USER_DIR$/.gem/ruby/2.2.3/bundler/gems/jekyll-6d2b344c0e24/lib/jekyll/liquid_renderer/file.rb:47:in `measure_time'
  from $USER_DIR$/.gem/ruby/2.2.3/bundler/gems/jekyll-6d2b344c0e24/lib/jekyll/liquid_renderer/file.rb:10:in `parse'
  from $USER_DIR$/.gem/ruby/2.2.3/bundler/gems/jekyll-6d2b344c0e24/lib/jekyll/renderer.rb:109:in `render_liquid'
  from $USER_DIR$/.gem/ruby/2.2.3/bundler/gems/jekyll-6d2b344c0e24/lib/jekyll/renderer.rb:62:in `run'
  from $USER_DIR$/.gem/ruby/2.2.3/bundler/gems/jekyll-6d2b344c0e24/lib/jekyll/site.rb:447:in `block (2 levels) in render_docs'
...
```
## The Output I Wanted

I would expect binary files (images, etc.) to simply be ignored and copied through to the output directory unmodified.
## Previous issues

This issue seems to have come up in past issues which may relate, including: #2592, #4276, #2262, and #2228.
## Root cause

This appears to be due to the logic of [`Jekyll::Document.render_with_liquid?`](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/document.rb#L168-L170)

``` ruby
    def render_with_liquid?
      !(coffeescript_file? || yaml_file?)
    end
```

As you can see, it returns true if the source file is not coffeescript or yaml, which a `.png` isn't.  It's not suitable for liquid either though, hence the problem.

As a quick and dirty test, I added a case for `.png` extensions:

``` ruby
    def render_with_liquid?
      !(coffeescript_file? || yaml_file? || %w(.png .jpg .bin).include?(extname)
    end
```

This has the desired effect, and seems to confirm I'm on the right track.  Obviously it's not a good solution though (hence no PR).  Assuming that storing non-liquid files in the same directory is supported (I believe it is?) this probably needs to ensure the content is suitable (ie. text, likely UTF-8) before passing it to Liquid.  Maybe something along the lines of checking against config_yml's existing `markdown_ext`?

I'll see if I can get a test together to send as a PR, while I'm not quite sure of the proper solution, I think I've got a solid feel for the issue behind this behavior at least.

Crunch09:
Instead of letting Liquid crash later in the build process IMHO we should just skip these files and inform the user.
 
**Before**:
bildschirmfoto 2017-09-03 um 14 48 26

**After**:
bildschirmfoto 2017-09-03 um 14 48 42

This fixes #5181

### DIFF
--- a/lib/jekyll/readers/post_reader.rb
+++ b/lib/jekyll/readers/post_reader.rb
@@ -36,10 +36,15 @@ module Jekyll
     def read_publishable(dir, magic_dir, matcher)
       read_content(dir, magic_dir, matcher).tap { |docs| docs.each(&:read) }
         .select do |doc|
-          site.publisher.publish?(doc).tap do |will_publish|
-            if !will_publish && site.publisher.hidden_in_the_future?(doc)
-              Jekyll.logger.debug "Skipping:", "#{doc.relative_path} has a future date"
+          if doc.content.valid_encoding?
+            site.publisher.publish?(doc).tap do |will_publish|
+              if !will_publish && site.publisher.hidden_in_the_future?(doc)
+                Jekyll.logger.debug "Skipping:", "#{doc.relative_path} has a future date"
+              end
             end
+          else
+            Jekyll.logger.debug "Skipping:", "#{doc.relative_path} is no valid UTF-8"
+            false
           end
         end
     end
--- a/test/test_site.rb
+++ b/test/test_site.rb
@@ -3,6 +3,22 @@
 require "helper"
 
 class TestSite < JekyllUnitTest
+  def with_image_as_post
+    tmp_image_path = File.join(source_dir, "_posts", "2017-09-01-jekyll-sticker.jpg")
+    FileUtils.cp File.join(Dir.pwd, "docs", "img", "jekyll-sticker.jpg"), tmp_image_path
+    yield
+  ensure
+    FileUtils.rm tmp_image_path
+  end
+
+  def read_posts
+    @site.posts.docs.concat(PostReader.new(@site).read_posts(""))
+    posts = Dir[source_dir("_posts", "**", "*")]
+    posts.delete_if do |post|
+      File.directory?(post) && !(post =~ Document::DATE_FILENAME_MATCHER)
+    end
+  end
+
   context "configuring sites" do
     should "have an array for plugins by default" do
       site = Site.new default_configuration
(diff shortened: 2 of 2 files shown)


Crunch09:
cc: @jekyll/build

lib/jekyll/readers/post_reader.rb
```diff
@@ -36,10 +36,15 @@ def read_posts(dir)
...
-            if !will_publish && site.publisher.hidden_in_the_future?(doc)
-              Jekyll.logger.debug "Skipping:", "#{doc.relative_path} has a future date"
+          if doc.content.valid_encoding?
+            site.publisher.publish?(doc).tap do |will_publish|
+              if !will_publish && site.publisher.hidden_in_the_future?(doc)
+                Jekyll.logger.debug "Skipping:", "#{doc.relative_path} has a future date"
+              end
             end
+          else
+            Jekyll.logger.debug "Skipping:", "#{doc.relative_path} is no valid UTF-8"
```
Crunch09: I wasn't sure if this should rather be a warning? But given that we already have an error message when this file is read i though it's good enough.

DirtyF: From a user perspective, a comprehensive warning would be easier to grasp:

```sh
Warning;  Found a binary file in your '_posts' directory, skipping: /path/to/image.jpg
```

parkr: Why is it a warning? Is something bad happening?

Crunch09: I guess @DirtyF changed his suggestion in the meantime to an `info` message? I like it, makes it easier to detect then a `debug` message and the existing `error` message does not include the information that a file has been skipped.

parkr:
Thank you!

@jekyllbot: merge +fix

PR #6345, opened 2017-09-03
require `pry` for easier debugging

Crunch09:
Tiny change but when debugging jekyll i often find it useful to set breakpoints with `binding.pry` as jekyll is already including `pry` as a [development dependency](https://github.com/jekyll/jekyll/blob/b6853bf9389261d25b410d86a41d2b8f06f63a41/Gemfile#L15#L15).
Without requiring `pry` first though, this throws an error:
```ruby
NoMethodError: undefined method `pry' for #<Binding:0x007fe67b5a2ad0>
``` 
This PR tries to make this easier for contributors.

### DIFF
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -33,6 +33,12 @@ require "liquid"
 require "kramdown"
 require "colorator"
 
+# development
+begin
+  require "pry"
+rescue LoadError => _
+end
+
 SafeYAML::OPTIONS[:suppress_warnings] = true
 
 module Jekyll


Crunch09:
I'm not sure which team to tag for this 🤔

parkr:
Pry is a great debugger! I don't think putting it in code we ship to the user is a good idea, though. I think we might have a solution for you actually. Have you used `script/console` (In this repo) before? It boots pry but also has some helpful methods specific to that debugging environment like `fixture_site` to make it easier to create a test site without typing as much.

DirtyF:
Note: we should mention this in https://jekyllrb.com/docs/contributing/#code-contributions

Crunch09:
Thanks @parkr ! I wasn't aware, will check it out 👍

Crunch09:
@DirtyF Do you want me to add this information from Parker to the contribution guide with this PR? Or should i rather close this one and open a new one?

DirtyF:
@Crunch09 Es is mir egal 😄

Crunch09:
@DirtyF Thanks, will do so later! 👏 Props to your german 😄

Crunch09:
Closing this in favor of #6349 .
@parkr `script/console` is really neat 👌  The only thing i still miss is being able to run tests and break in it without manually requiring `pry` first but i understand your point 👍

PR #6346, opened 2017-09-04
Fix precedence docs

ohadschn:
Before the fix no precedence took place because one configuration was for pages and the other was for posts


pup:
@jekyllbot: merge +docs

PR #6350, opened 2017-09-07
Mutable drops should fallback to their own methods when a mutation isn't present

benbalter:
A follow up to https://github.com/jekyll/jekyll/pull/6338 and discovered via https://github.com/jekyll/github-metadata/pull/112, the current logic for `Jekyll::Drops::Drop#key?` for mutable drops is to look for a mutated key, but then not to look for a native method or fallback data. Instead, Drops should look for a key in the mutated data, in native methods, AND in fallback data.

This PR updates the logic of `Drop#key?` and includes tests to ensure the method does what we expect it to going forward.

### DIFF
--- a/lib/jekyll/drops/drop.rb
+++ b/lib/jekyll/drops/drop.rb
@@ -105,7 +105,8 @@ module Jekyll
       # Returns true if the given key is present
       def key?(key)
         return false if key.nil?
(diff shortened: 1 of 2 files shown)


parkr:
@benbalter Should this be backported and released as v3.5.3?

parkr:
@jekyllbot: merge +bug

Issue #6354, opened 2017-09-11
`{{page.url}}` does not include `.html`, while the actual page URL really does


kvz:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [x] I ran `jekyll doctor` to check my configuration
- [ ] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [ ] This is a feature request.

---

- [x] I am on (or have tested on) ***macOS*** 10+
- [ ] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux
- [ ] I am on (or have tested on) ***Fedora*** GNU/Linux
- [ ] I am on (or have tested on) ***Arch*** GNU/Linux
- [ ] I am on (or have tested on) ***Other*** GNU/Linux
- [ ] I am on (or have tested on) ***Windows*** 10+

---

- [ ] I was trying to install.
- [ ] There is a broken Plugin API.
- [ ] I had an error on GitHub Pages, and I have reproduced it locally.
- [ ] I had an error on GitHub Pages, and GitHub Support said it was a Jekyll Bug.
- [ ] I had an error on GitHub Pages and I did not test it locally.
- [x] I was trying to build.
- [ ] It was another bug.

## My Reproduction Steps

In `_config.yml`:

```yaml
permalink: "/blog/:year/:month/:day/:title.html"
```

In my terminal, checking if my page has `.html` in its URL:

```bash
$ curl -I http://localhost:3000/implementations/
HTTP/1.1 404 Not Found

$ curl -I http://localhost:3000/implementations.html
HTTP/1.1 200 OK
```

In `implementations.md`:

```html
{{page.url}}
```

## The Output I Wanted

`/implementations.html`

## The Output I Got

`/implementations`

DirtyF:
Didn't reproduce this behavior, make sure you don't have a `permalink` setting in your markdown page.

kvz:
Full markdown frontmatter:

```
---
layout: default
title: Implementations
comments: true
---
```

full _config.yml:

```
permalink: "/blog/:year/:month/:day/:title.html"
tusd_endpoint: "http://master.tus.io"
commenting_baseurl: https://tus.io

kramdown:
  input: GFM
  hard_wrap: false
  enable_coderay: false
  auto_ids: true
  footnote_nr: 1
  entity_output: as_char
  toc_levels: 2..6

[...]

PR #6355, opened 2017-09-11
Test page URL with permalink: "/:categories/:year/:month/:day/:title.html"

DirtyF:
Add failing test for #6354

### DIFF
--- a/test/test_page.rb
+++ b/test/test_page.rb
@@ -206,6 +206,19 @@ class TestPage < JekyllUnitTest
         end
       end
 
+      context "with custom permalink style with html extension" do
+        setup do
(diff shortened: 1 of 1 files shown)


Crunch09:
@DirtyF Hey, thanks for providing the test! I just tested this and it passes if you change the permalink style to
```
:categories/:year/:month/:day/:title.:output_ext
```
To me this makes sense because `:output_ext` will be replace with the actual extension of your document and that's more dynamic as setting it to a hardcoded extension. Maybe we should update the documentation instead?
@kvz Would that work for you?

kvz:
> @kvz Would that work for you?

Sure does! Thanks for looking into this!

Issue #6360, opened 2017-09-12
Generating relative URL (and I mean *actually* relative)


aioobe:
I just discovered `relative_url` but it doesn't do what I expect it to.

In my `default.html` template, I have `<link rel="stylesheet" href="css/style.css" type="text/css">` This works fine as long as I create pages in the root, but not if I have a page (based on the `default.html` template) in a subdirectory.

I can obviously add a leading `/` i.e. change from `href="css/style.css"` to `href="/css/style.css"` and it will work locally and when I've uploaded to my server, but only if I'm serving the site from the root of my domain.

I thought `{{ "/css/style.css" | relative_url }}` would turn into `css/style.css` for pages in the root, and `../css/style.css` for pages in a subdirectory, but no, unless I mess around with the `baseurl` settings, it turns into `/css/style.css` for all pages. This is what I call an **absolute url**. It starts with a `/`!

I would like a way to turn `"/css/style.css"` into `"css/style.css"` for pages in the root, and `"../css/style.css"` in subdirectories. (And `relative_url` is a very misleading name, but maybe that's too late to do something about?)

ashmaroli:
Hello @aioobe you seem to be confused with the two terminologies here.
`relative_url` does not mean relative to the current file or directory but rather *relative* to the site's location.
And `absolute_url` means a fully resolved path to a resource.

If your site is at `http://www.somewhere.com/my-site` then the absolute URL for a `styles.css` in the `css` subdir in an `assets` dir would be `http://www.somewhere.com/my-site/assets/css/styles.css` whereas the relative url for the same file is only `/assets/css/styles.css`

ashmaroli:
Another point to note: the relative path is only resolved by an active server. Simply opening a generated `*.html` in the browser won't load the css. You need to run `jekyll serve` or `bundle exec jekyll serve` first..

aioobe:
Ok, I see the train of thought that led up to the name `relative_url`. It's still confusing imo. The first few hits that come up when googling `relative url` shows examples such as `"sibling-page.html"` and `"../other-page.html"`. Not `"/some/other/page.html"`.

My question remains though: How do I produce links such as `"../css/style.css"`?

I would like to view the HTML on disk (`file://`) and upload the content of `_site` to a static web-server without bothering about how it's served or from which root directory. After all, that's one of the big benefits with working with static HTML-pages, no?

ashmaroli:
> After all, that's one of the big benefits with working with static HTML-pages, no?

[...]

PR #6362, opened 2017-09-13
Add relativize_url

aioobe:
Added a filter called `relativize_url` which generates URLs relative to the current page.

See discussion here: https://github.com/jekyll/jekyll/issues/6360

I'd be happy to add tests and docs, but would first like to hear if maintainers find this a worthwhile addition. 

Personally I find it incredibly useful as I no longer have to care about the path of the directory from which the content of `_site` is served (as is the case when using for instance `relative_url`). Also, I don't need to mess with different `baseUrl` in development and production. Resulting HTML is "self-contained" and can even be viewed directly on disk (`file://...`). Granted, it's easy for users to put this filter in the `_plugins` directory manually.


mattr-:
I'm not sure there's a strong enough use case for this in Jekyll's core. As mentioned in #6360, there's `baseurl` that can be used to counteract some of this. Maybe I'm just missing the point of this?

parkr:
I think it has a place in our section of filters. People have always wanted a way to take one URL and turn it into a URL relative to the current page. It makes the HTML files more navigable since you don't need to host them anywhere. That's my person take, anyway.

lib/jekyll/filters/url_filters.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "addressable/uri"
+require "pathname"
```
ashmaroli: this line is not required as its already loaded by [`jekyll.rb`](https://github.com/jekyll/jekyll/blob/v3.5.2/lib/jekyll.rb#L24)

lib/jekyll/filters/url_filters.rb
```diff
@@ -33,6 +34,20 @@ def relative_url(input)
...
+      # page lives in `/images`, `/` or `/sub/dir/` respectively.
+      #
+      # input - a path relative to the project root
+      #
+      # Returns the supplied path relativized to the current page.
+      def relativize_url(input)
+        pageUrl = @context.registers[:page]["url"]
+        pageDir = Pathname(pageUrl).parent
+        Pathname(input).relative_path_from(pageDir).to_s
+      end
```
ashmaroli: if you check the existing URLFilters, you'll see that there's a guard against `nil` inputs..
`Pathname(input).relative_path_from(pageDir).to_s` will raise if `input` is `nil`

Perhaps alternatively stringify the input beforehand?
```diff
- `Pathname(input).relative_path_from(pageDir).to_s`
+ Pathname(input.to_s).relative_path_from(pageDir).to_s
```

ashmaroli:
I too think this filter will find acceptance if it'll allow one to view the entire generated site on the `file://` protocol

lib/jekyll/filters/url_filters.rb
```diff
@@ -33,6 +33,21 @@ def relative_url(input)
...
+      # page lives in `/images`, `/` or `/sub/dir/` respectively.
+      #
+      # input - a path relative to the project root
+      #
+      # Returns the supplied path relativized to the current page.
+      def relativize_url(input)
+        return if input.nil?
+        page_url = @context.registers[:page]["url"]
+        page_dir = Pathname(page_url).parent
+        Pathname(input).relative_path_from(page_dir).to_s
```
parkr: Are there any error scenarios we should be worried about here? What if the file doesn't exist?

We should also ensure the `input` is inside of the `destination` directory or else this could be used for arbitrary path traversal which would be no fun.

aioobe: I don't think we should worry about whether the file exists or not. In the end `{{ "/non/existent.file" | relativize_url }}` is the same as just writing `non/existent.file` directly in markup.

As it happens, `Pathname.relative_path_from` ensures that the result doesn't go outside the root. Added tests for it.

lib/jekyll/filters/url_filters.rb
```diff
@@ -33,6 +33,22 @@ def relative_url(input)
...
+      # input - a path relative to the project root
+      #
+      # Returns the supplied path relativized to the current page.
+      def relativize_url(input)
+        return if input.nil?
+        input = ensure_leading_slash(input)
+        page_url = @context.registers[:page]["url"]
+        page_dir = Pathname(page_url).parent
+        Pathname(input).relative_path_from(page_dir).to_s
+      end
```
ashmaroli: some of the assignments can be done away:
```ruby
      def relativize_url(input)
        return if input.nil?
        page_dir = Pathname(@context.registers[:page]["url"]).parent
        Pathname(ensure_leading_slash(input)).relative_path_from(page_dir).to_s
      end
```
But I'll leave it to you to decide if the above hampers readability..

aioobe:
I think I have addressed all feedback I've received so far. Please let me know if there's anything I could improve.

aioobe:
@parkr Any further thoughts on this?

pathawks:
I'm not sure that we want to introduce a new filter with behavior so similar to an existing filter. When would I use `relative_url` instead of `relativize_url`?

aioobe:
@pathawks I gave the name some thought but failed to come up with something that's not confusing but still distinguishes itself from `relative_url`.

`relative_path`? `relativize`?

I'm up for suggestions, but I think it would be a shame if one confusingly named filter (`relative_url` which produces an absolute path) is the reason for making other filter names confusing. One confusing filter is better than two so to speak.

I can't tell why you'd want to use `relative_url` tbh. I guess you'd have to ask the guy that introduced that filter.

DirtyF:
Well technically a URI starts with a protocol, so I guess `relative_path` is a good fit here.

pathawks:
We can't do it in a minor version, but I would much rather just replace `relative_url` with the behavior of this filter. There shouldn't be multiple ways to do the same thing.

aioobe:
When I suggested `relative_path` I actually wasn't aware of the `page.relative_path` variable. Name clash too confusing? How about `relativize_path`?

parkr:
> We can't do it in a minor version, but I would much rather just replace relative_url with the behavior of this filter. There shouldn't be multiple ways to do the same thing.

@pathawks This feels like an overloading if the term relative: domain-relative link vs page-relative link. They’re both perfectly valid relatives, but the naming vagueness is confusing. Without breaking the old way (it’s used everywhere!), is there a way to make this filter more specific for the expected outcome of page-specificness?

DirtyF:
Why not `local_path`or something equivalent that conveys the idea we do no want to generate an URL for the path provided?

aioobe:
> Why not local_path, no_url or something equivalent that conveys the idea that we do no want to generate an URL for the path provided?

@DirtyF Neither `relative_url` nor `absolute_url` generate URLs either. They both result in paths, so I'm not sure why this proposed filter is more "local" than those.

DirtyF:
@aioobe Naming things is hard.

`relativize_path `might be a good fit.

pathawks:
`relative_path`?

aioobe:
Summarizing suggestions so far:

- `relative_url` – already in use
- `relativize_url` – too similar, confusing
- `relative_path` – clashes with the `page.relative_path` variable. Too confusing, or ok?
- `local_path` – strange, as `relative_url` and `absolute_url` also generate "local" paths
- `no_url` – not particularly descriptive
- `relativize_path` – no objections so far

A new one:

- `page_relative` – Descriptive. Can't find any clashes.

Personally I'd be fine with `relative_path`, `relativize_path` or `page_relative`.

I'm new around here... Who are the maintainers? Someone authoritative needs to just make a decision :-)

ashmaroli:
:+1: to `page_relative`

DirtyF:
Let's narrow this down to `relative_path` or `page_relative`. Both are equally fine to me.

@benbalter @parkr which one has your preference?

parkr:
Is it weird to use `page_relative_url`? It is specific and helps you see what the output will contain (a url).

benbalter:
Taking a step back, what problem are we trying to solve here? Making it easier for users to browse their site via the `file://` protocol? 

Do we want to encourage two different ways to preview the site locally (build + navigate directly versus build + serve)? If that's the case, what's the value of the server then? Can we just remove the existing `jekyll serve` code entirely and tell users to just browse via the `file://` protocol? 

Would an easier workaround not be to set the local url to `file://[source]`? To make that the default instead of `localhost:4000`? Are their not [same-origin issues](https://bugs.chromium.org/p/chromium/issues/detail?id=40787) with using the `file://` protocol (so a user builds their site using this new filter only to realize they can't actually preview it the way this filter was intended to allow)? 

Reading https://github.com/jekyll/jekyll/issues/6360 it sounds like the problem we're trying to solve is being able to serve via the `file://` protocol, which itself requires all links to be relative to the linking file, rather than the site root, but I don't see any discussion if formal `file://` protocol support is something we *should* do (or if relative-to-the-linking-file links is a best practice we want to encourage new developers to adopt as their default versus say relative to the domain root or the absolute URL (personally, I don't think moving a file should require you to rewrite all the links in it)). 

Given the discussion above and in #6360 around the confusion our existing filter names cause, it's clear there's a non-negligible cost to adding another URL related filter with slightly different behavior, and I want to make sure we're weighing that tradeoff before landing on a name (which if we did, something super explicit like `path_relative_to_current_page` may be better).

aioobe:
> Taking a step back, what problem are we trying to solve here? Making it easier for users to browse their site via the `file://` protocol?

@benbalter Absolutely not.

I simply like my resulting HTML to be oblivious to which directory it's served from. If I move my site from `/var/www/blog` to `/var/www/old_blog` all my links break, and `relative_url` is of no use here since it generates paths starting with `/`. The only way forward would be to mess with `baseurl` and recompile my site to make it compatible with the new path. To me it just seems like a silly constraint.

A *side effect* of the HTML being "self contained" is that simple sites can be browsed through `file://`. By all means, keep the `jekyll serve` story if it simplifies same-origin constraint.

benbalter:
> I simply like my resulting HTML to be oblivious to which directory it's served from. If I move my site from `/var/www/blog` to `/var/www/old_blog` all my links break, and `relative_url` is of no use here since it generates paths starting with `/`.

@aioobe First, thank you for taking the time to talk through your pull request. It's discussions like these, that show such a commitment to the end-user experience that make Jekyll such a great project.

I'm confused why the `relative_url` filter wouldn't work if you were using `jekyll serve` to preview the site? All URLs would be relative to the domain's root, which in this case would be `localhost:4000` by default. If you move the site to another directory, neither the compiled site nor the URL would need to change. Unless I'm misunderstanding the problem you're trying to solve?

aioobe:
Suppose I log into my webserver and do

`$ mv /var/www/blog /var/www/old_blog`

A page previously located at `http://example.com/blog/about.html` is now located at `http://example.com/old_blog/about.html`. Since `relative_url` always generates paths starting with `/`, all links will break.

Sure, I can change the `baseurl`, recompile my site and upload the new `_site` to `/var/www/old_blog`. But for a tool that targets static HTML I find this a bit silly. I mean, why should the path on _the server_ (`/blog`, then `/old_blog`) necessarily be hard coded in the source? To me it's like writing a stand alone application that assumes it lives in `C:\Program Files`.

aioobe:
These seem to be the most popular solutions:

- https://ricostacruz.com/til/relative-paths-in-jekyll ([395 hits on Github](https://github.com/search?q=%22{%25+elsif+depth+%3D%3D+2+%25}{%25+assign+base+%3D+%27..%27+%25}%22&type=Code&utf8=%E2%9C%93))
- https://stackoverflow.com/a/14254928/276052 ([267 hits on Github](https://github.com/search?q=%22%7B%25+assign+lvl+%3D+page.url+%7C+append%3A%27X%27+%7C+split%3A%27%2F%27+%7C+size+%25%7D%22&type=Code&utf8=%E2%9C%93))

I consider both of them to be ugly hacks. For starters I believe they need to be included in each page that requires the functionality.

DirtyF:
@aioobe AFAIC I'm 👍 on this, I see no reason not to allow domainless page relative relative links in Jekyll. It does not change the actual default behavior. But it should be well explained and documented.
If it does not make his way in the core, 🙏  please release it as a plugin , it's also this modular approach that makes Jekyll so great. 

Reminder:

- [ ] Choose and change the filter name (`page_relative`, `relative_page` have my vote, I prefer the latter because it is mirroring `relative_url`) 🤙 
- [ ] Document the filter usage on the website 📖 
- [ ] Getting approval by two other core members 🙈

aioobe:
@DirtyF glad to hear you're :+1: with this.

If two more core members approve the filter, I'd be more than happy to change filter name and contribute with documentation.

ashmaroli:
@aioobe say this filter has been shipped and a user replaces all use of `relative_url` with this filter.
What would you expect to happen when the user
  - runs `jekyll build`, or
  - runs `jekyll serve`
?

aioobe:
@ashmaroli the links would change from being relative to the root of the webserver, to being relative to the current page.

Suppose we're editing `/some/page.html`. Changing from `relative_url` to this new filter we'd get the following:

| Filter input        | `relative_url`   | This new filter |
| ------------- |-------------| -----|
| <sup>`"/some/other/article.html"`</sup> | <sup>`/base_url/some/other/article.html`</sup>      | <sup>`other/article.html`</sup> |
| <sup>`"/some/sibling.html"`</sup> | <sup>`/base_url/some/sibling.html`</sup>  | <sup>`sibling.html`</sup> |
| <sup>`"/home.html"</sup>` | <sup>`/base_url/home.html`</sup>      | <sup>`../home.html`</sup> |

Nothing would break. The site as a whole could be moved, say `/var/www/blog` to `/var/www/old_blog` without causing all links to break.

aioobe:
@benbalter you never replied after https://github.com/jekyll/jekyll/pull/6362#issuecomment-332417981.

Any thoughts?

ashmaroli:
@aioobe I sort of tested this locally with a new setup.. and I came across a bug:
**pages that are configured to `:pretty` permalinks do not resolve the links properly,**

i.e. `./about.md` configured to output to `./_site/about/index.html` remains unstyled because the resulting href to the stylesheet is `"assets/main.css"` instead of expected `"../assets/main.css"`

aioobe:
@ashmaroli  I'm not familiar with `:pretty`. I have that precise situation in my page however, and I get `../assets/main.css` as the resulting URL for the style.

ashmaroli:
> I'm not familiar with `:pretty`

Sorry, I should've used a more general lingo..
`:pretty` is the [built-in permalink style](https://jekyllrb.com/docs/permalinks/#builtinpermalinkstyles) where an `./about.md` gets output to `./_site/about/index.html` instead of the default `./_site/about.html`

ashmaroli:
@DirtyF may i request you to test what I encountered, at your end..? Thanks..

  1. checkout a local copy of this branch
  2. Switch the names for `relative_url` and `relativize_url`
  3. Correct references to `relative_url` in `absolute_url` as well
  4. Run `script/default-site` to create a dummy-site that points to the *checked-out version of Jekyll*
  5. `build` / `serve` the dummy-site.
  6. Check if the `href` to the stylesheet in `./tmp/default-site/_site/about/index.html` is `../assets/main.css`

aioobe:
Ok, I can see how `/about.md` ending up in `/about/index.html` could be an issue. Maybe there's an alternative to `@context.registers[:page]["url"]` which refers to the target file (`/about/index.html`) rather than the source file (`/about.md`). I'm very new to Jekyll. I have no idea about this.

ashmaroli:
`@context.registers[:page]["url"]` **does refer to the target file** instead of the source path.
I came up with a workaround for the bug above.. (a hack at most..)

```ruby
      def relativize_url(input)
        return if input.nil?
        input    = ensure_leading_slash(input)
        page_url = @context.registers[:page]["url"]
        page_dir = Pathname(page_url).parent

        target   = page_url.end_with?("/") ? Pathname(page_url) : page_dir
        Pathname(input).relative_path_from(target).to_s
      end
```

It nukes the above bug.. not sure if there still lurks other hidden critters..

aioobe:
Ah, seeing the solution makes it clear why the original code was buggy. Your solution makes sense.

ashmaroli:
what's the status on this @aioobe ?

pathawks:
I don’t think this is necessary. I would rather not have two different ways to create URLs without clear use cases for each.

aioobe:
@pathawks, well I (and as comment https://github.com/jekyll/jekyll/pull/6362#issuecomment-333292878 indicate, quite a few other) disagree. 

- `relative_url` is broken since it generates an absolute path(!)
- Forcing users to rebuild their site with an updated `baseUrl` in order to move the site from one directory to another is silly at best.

With that said, I'm stepping out of this discussion. This has taken way more time than I initially set out to spend on this. This is not the first time the lack of this type of filter has been brought up, and I doubt it will be the last. The current patch is quite polished and includes tests. Should some one want to pick up where this was left off, it's not far to the finish line.

pathawks:
@aioobe You are correct, the code is solid and is the best possible implementation of this feature. If I could go back and reinvent the `relative_url` filter, I would be tempted to do it this way rather than the current behavior.

I fully support making this available as a plugin for users looking for this behavior. Is that something you would be willing to support?

ashmaroli:
@aioobe
Unrelated to this proposal, but I just came across a point of conflict that would arise when you try to view a Jekyll site on the `file://` protocol..

Say, I have generated a site by using this implementation instead of existing `relative_url`..
The site would break at `:pretty` permalinks (e.g. `about.md` would generate an internal link `/about/` and writes to `./_site/about/index.html`)
When a user clicks on a link pointing to `/about/`, then on the `file://` protocol the user would only see the directory structure for `about`.

My point being, Jekyll is currently not ready to generate sites compatible both on the `file://` and `http(s)://` protocol..

pathawks:
@ashmaroli Good point.

I think for now we should close this as plugin territory. The advantage of that being the ability to override the existing `relative_url` filter instead of adding one more filter.

kastaneda:
Very sad. Seems that someone had completely missed the point about plugins and Github pages.

Mayeu:
I'd like to point out another usage: generating website compatible [with IPFS](https://github.com/ipfs/notes/issues/66) and possibly future other decentralise technology.

Also, [Hugo has that](https://gohugo.io/content-management/urls/#relative-urls) :)

mdavidn:
I'd like to point out another usage: Easily sharing work-in-progress with clients who can't run Jekyll on their local computer.

Perhaps the client doesn't know how to install Ruby, or their local IT department prohibits it. These reasons also eliminate Python's `SimpleHTTPServer` and various Node modules as options.

DirtyF:
@mdavidn I'd rather recommend services like Netlify or Surge for that.

mdavidn:
If I could deploy these sites to the public internet, I would. That is unfortunately not an option.

obfischer:
Could you please consider to reopen this PR? It would be nice to have this functionality. At least I need it, but it seems to that I am not the only one.

ashmaroli:
@obfischer May I ask what exactly is your use-case? This could be released as a third-party plugin and developed independently of Jekyll Core since its mainly intended to navigate a site without a server (and therefore, of not much interest to GitHub Pages)

obfischer:
@ashmaroli Sometimes I need to deploy my pages to an subdirectory of the used webserver as `/preview` or I would like to archive old versions of the homepage my moving them from `/` to `/archive/2017`. This would be much easier to do if we could use relative paths instead of absolute ones.

I like your suggestion to release this functionality as plugin. I hope @aioobe is still aware of this PR and has time to build the plugin.

WyseNynja:
This looks like exactly I want. I'm not trying to serve my site out of file:// (although I think that would be nice to have). Instead, I am trying to serve my site from IPFS where the baseurl is something like /ipfs/XXXX and it changes every time the site is published.

PR #6364, opened 2017-09-14
Add sort_by_post_count Feature

mpchadwick:
Currently there's no built in way to sort `post_attr_hash`es (e.g. `site.tags`) by post count.

This has lead to some <strike>solutions</strike> hacks proposed on various blog posts and Stack Overflow threads

- https://stackoverflow.com/questions/24700749/how-do-you-sort-site-tags-by-post-count-in-jekyll
- http://blog.jupo.org/2013/05/05/sandboxed-jekyll-hacks/#frequency-tag-sort

This adds a new filter for sorting `post_attr_hash`es by post count.

Usage example...

```
{% assign sorted_tags = (site.tags | sort_by_post_count) %}
{% for tag in sorted_tags %}
    <h2 id="{{ tag[0] | slugify }}">{{ tag[0] }}</h2>
    <ul>
    {% for post in tag[1] %}
        <li><a href="{{ post.url }}">{{ post.title }}</a> <span class="date">{{ post.date | date: "%B %-d, %Y"  }}</span></li>
    {% endfor %}
    </ul>
{% endfor %}
```


mpchadwick:
Right after submitting this I realized there may be way to add the functionality to the current `sort` filter without breaking backwards compatibility.

For example, if `property` was passed as "by_post_count" or "post_count" it could execute the required sorting. Alternately, `sort` could accept a fourth `by_post_count` parameter which defaults to false. Then the method would first check if it was set to true and execute the sort by post count logic in that case. 

Looking forward to any feedback on this. Currently in a world of pain with [this](http://blog.jupo.org/2013/05/05/sandboxed-jekyll-hacks/#frequency-tag-sort) because it's really the only feasible solution on GitHub pages.

mpchadwick:
Not sure if any of the maintainers will get a notification from this comment but I'm wondering what the timeline is for getting some feedback on this. This was submitted nearly 2 weeks ago and hasn't received any feedback at all.

pathawks:
I would rather see this integrated into the existing `sort` filter (as you suggest).

Maybe something like:

```text
{% assign sorted_tags = site.tags | sort: 'length' %}
```
or
```text
{% assign sorted_tags = site.tags | sort: 'size' %}
```

mattr-:
It feels like we're missing some additional functionality around handling tags that this issue exposes. I understand the use case for what's being asked but I'm not convinced that any of the solutions thus far are actually the way that we need to be solving the problem. What about adding more functionality to our `tags` object that we send into the liquid in order to do things like sort by most posts (or least posts). What if we bring some tag cloud functionality into Jekyll core itself?

mattr-:
For example, I'd love to be able to see people write `site.tags.sorted_by_post_count` or something similar in liquid rather than what we've got going on today.

DirtyF:
> What if we bring some tag cloud functionality into Jekyll core itself?

Tag clouds in core? It doesn't look like a 80% users feature.
```

With what we have today I feel like  `{{ site.posts | sort: 'parameter' }}` is the logical path to follow for adding new filters. `parameter` could take different values like`author`, `title`,  `tags`, `random` . ideally we could pipe this to a `reverse` filter that would sort the other way around.

**Comment by mpchadwick** [comment]
Seems like this is at a stalement.

For my use case, I don't think that `site.tags` needs to define additional behavior and the `sort` filter can be enhanced to support what I'm looking for. Not sure if there's other things one might want to do tags, though.

[Currently, `site.tags` is a `Hash`](https://github.com/jekyll/jekyll/blob/72debf854b57bf13d7bd3ec4a6dfe7147909be72/lib/jekyll/site.rb#L220-L246). I'm not super familiar with Ruby but I guess to support @mattr-'s idea, it would become a class which defined methods to do things like sorting?

As a contributor, it would definitely be much more straight forward to make the change to `sort` filter to support this feature request rather than implementing a new class for tags.

**Comment by parkr** [comment]
As I understand it, your use-case is: I want `site.tags` to be sorted by the number of elements in the value. The `sort` filter does support sorting by a value of the item, in this case you'd want `.length`, right? But it doesn't naturally allow you to dip into elements.

I found a minor solution here which isn't super ideal, but allows you to do this:

```liquid
{% assign sorted_tags = site.tags | sort: "last.length" %}
```

During sorting, Hashes turn into Arrays. The first item in the array is the hash key, which is the tag name in this case. The second item in the array is the hash value, the array of posts in this case. `last` here means we call `Array#last`, which gets the array of posts, then we chain `length` onto that. It works in local tests:

```text
input => {"government"=>["one post", "two post"], "tech"=>["three post"], "lol"=>[1, 2, 3]}
interim => [[2, ["government", ["one post", "two post"]]], [1, ["tech", ["three post"]]], [3, ["lol", [1, 2, 3]]]]
output => [["tech", ["three post"]], ["government", ["one post", "two post"]], ["lol", [1, 2, 3]]]
```

You can then use the `reverse` liquid filter to turn the array around such that the tag with 3 posts shows up first, then the tag with 2 posts, etc.

Here's the code: https://gist.github.com/parkr/a9ee3d3a4eceb74a99bfc4957c52a446

mpchadwick:
@parkr

Yes, you are understanding my use case and the issue correctly.

In my opinion, the syntax proposed by @pathawks is cleaner. 

```
{% assign sorted_tags = site.tags | sort: 'count' %}
```

vs.

```
{% assign sorted_tags = site.tags | sort: "last.length" | reverse %}
```

It could be achieved by updating the sort method with different mechanics in  the case where `input` is a `Hash` and `property` is the 
 agreed upon reserved keyword, e.g. "count" (or "length" or "size")

```ruby
def sort(input, property = nil, nils = "first")
  if input.nil?
    raise ArgumentError, "Cannot sort a null object."
  end
  if property.nil?
    input.sort
  elsif input.is_a?(Hash) && property == "count"
    if nils == "last"
      new_hash = Hash[input.sort_by { |tag, posts| posts.join.length}]
    else
      new_hash = Hash[input.sort_by { |tag, posts| -posts.join.length}]
    end
    new_hash
  else
    if nils == "first"
      order = - 1
    elsif nils == "last"
      order = + 1
    else
      raise ArgumentError, "Invalid nils order: " \
        "'#{nils}' is not a valid nils order. It must be 'first' or 'last'."
    end

    sort_input(input, property, order)
  end
end
```

In the example above it will be sorted by tag with MOST posts by default, but if "last" is passed for `nils` the tag with LEAST posts will show first

user414:
I found this feature request and for some unknown reason instead of using the suggested method in the OP I spent sometime getting a one liner to work. Just sharing in case it might be useful to someone else. 

```{% assign set_of_tags = site.tags | group_by_exp: 'item', 'item[1] | size' | sort: 'name' | reverse %}```

The reverse is just for my purpose since I need them in descending order. I have to admit I was even surprised by the automatic resolving of hash/array just as I needed it. Sometime I run into a lack of function to manipulate hash but this time worked like a charm. Thanks for the great work.

DirtyF:
@ashmaroli What's your take on this request? Plugin or core?

ashmaroli:
As it stands, the proposed filter doesn't have universal application &mdash; it is designed specifically for a datatype made up of two elements where the second element responds to `:join`.

So, my take is that this implementation does not make the cut to be merged into Core.

PR #6368, opened 2017-09-15
Bump rubocop to use `v0.50.x`

### LINKED ISSUE #6367: Update dependency constraint to allow for rubocop v0.50.0
Hey there! :wave:

The constraint I found was `~> 0.49.1`, and the latest version available is `0.50.0`.

ashmaroli:
Fixes #6367

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -88,8 +88,6 @@ Metrics/PerceivedComplexity:
   Max: 8
 Naming/FileName:
   Enabled: false
-Naming/HeredocDelimiterCase:
-  Enabled: false
 Naming/HeredocDelimiterNaming:
   Enabled: false
 Security/MarshalLoad:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -21,7 +21,7 @@ Layout/EmptyLinesAroundAccessModifier:
 Layout/EmptyLinesAroundModuleBody:
   Enabled: false
 Layout/EndOfLine:
-  EnforcedStyle: lf
+  EnforcedStyle: native
 Layout/ExtraSpacing:
   AllowForAlignment: true
 Layout/FirstParameterIndentation:
@@ -44,9 +44,13 @@ Layout/SpaceInsideBrackets:
   Enabled: false
 Lint/EndAlignment:
   Severity: error
+Lint/RescueWithoutErrorClass:
+  Enabled: false
 Lint/UnreachableCode:
   Severity: error
 Lint/UselessAccessModifier:
+  Enabled: false
+Lint/Void:
   Enabled: false
 Metrics/AbcSize:
   Max: 21
@@ -82,6 +86,12 @@ Metrics/ParameterLists:
   Max: 4
 Metrics/PerceivedComplexity:
   Max: 8
+Naming/FileName:
+  Enabled: false
+Naming/HeredocDelimiterCase:
+  Enabled: false
+Naming/HeredocDelimiterNaming:
+  Enabled: false
 Security/MarshalLoad:
   Exclude:
     - !ruby/regexp /test\/.*.rb$/
@@ -109,7 +119,7 @@ Style/Documentation:
     - !ruby/regexp /features\/.*.rb$/
 Style/DoubleNegation:
   Enabled: false
-Style/FileName:
+Style/Encoding:
   Enabled: false
 Style/GuardClause:
   Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -118,7 +118,7 @@ Style/Documentation:
 Style/DoubleNegation:
   Enabled: false
 Style/Encoding:
-  Enabled: false
+  EnforcedStyle: when_needed
 Style/GuardClause:
   Enabled: false
 Style/HashSyntax:
--- a/lib/jekyll/commands/doctor.rb
+++ b/lib/jekyll/commands/doctor.rb
@@ -86,7 +86,7 @@ module Jekyll
         def urls_only_differ_by_case(site)
           urls_only_differ_by_case = false
           urls = case_insensitive_urls(site.pages + site.docs_to_write, site.dest)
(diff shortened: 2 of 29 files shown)


lib/jekyll/commands/doctor.rb
```diff
@@ -86,7 +86,7 @@ def fsnotify_buggy?(_site)
         def urls_only_differ_by_case(site)
           urls_only_differ_by_case = false
           urls = case_insensitive_urls(site.pages + site.docs_to_write, site.dest)
-          urls.each do |_case_insensitive_url, real_urls|
+          urls.each_value do |real_urls|
```
pathawks: Cool!

lib/jekyll/converters/markdown/kramdown_parser.rb
```diff
@@ -42,12 +42,14 @@ def convert(content)
         end
 
         private
+        # rubocop:disable Performance/HashEachMethods
         def make_accessible(hash = @config)
           hash.keys.each do |key|
```
parkr: Can we use each_key here?

.rubocop.yml
```diff
@@ -82,6 +86,12 @@ Metrics/ParameterLists:
   Max: 4
 Metrics/PerceivedComplexity:
   Max: 8
+Naming/FileName:
+  Enabled: false
```
parkr: What fails for this one?

.rubocop.yml
```diff
@@ -82,6 +86,12 @@ Metrics/ParameterLists:
...
+Naming/HeredocDelimiterCase:
+  Enabled: false
```
parkr: We should enable this!

.rubocop.yml
```diff
@@ -82,6 +86,12 @@ Metrics/ParameterLists:
...
+Naming/HeredocDelimiterNaming:
+  Enabled: false
```
parkr: What fails for this?

.rubocop.yml
```diff
@@ -109,7 +119,7 @@ Style/Documentation:
     - !ruby/regexp /features\/.*.rb$/
 Style/DoubleNegation:
   Enabled: false
-Style/FileName:
+Style/Encoding:
```
parkr: Can we just exclude problematic files?

.rubocop.yml
```diff
@@ -44,10 +44,14 @@ Layout/SpaceInsideBrackets:
...
 Lint/EndAlignment:
   Severity: error
+Lint/RescueWithoutErrorClass:
+  Enabled: false
 Lint/UnreachableCode:
   Severity: error
 Lint/UselessAccessModifier:
   Enabled: false
+Lint/Void:
+  Enabled: false
```
parkr: What even is this? :joy:

.rubocop.yml
```diff
@@ -44,10 +44,14 @@ Layout/SpaceInsideBrackets:
   Enabled: false
 Lint/EndAlignment:
   Severity: error
+Lint/RescueWithoutErrorClass:
+  Enabled: false
```
parkr: Oooo we should enable this.

.rubocop.yml
```diff
@@ -21,7 +21,7 @@ Layout/EmptyLinesAroundAccessModifier:
 Layout/EmptyLinesAroundModuleBody:
   Enabled: false
 Layout/EndOfLine:
-  EnforcedStyle: lf
+  EnforcedStyle: native
```
parkr: Done we want lf here?

.rubocop.yml
ashmaroli: From their docs,
> This cop makes sure that Ruby source files have snake_case names. Ruby scripts (i.e. source files with a shebang in the first line) are ignored.

and flags us for the following:
    - 'Gemfile'
    - 'Rakefile'
    - 'lib/theme_template/Gemfile'
    - 'test/fixtures/test-dependency-theme/test-dependency-theme.gemspec'
    - 'test/fixtures/test-theme/test-theme.gemspec'

which IMO is a bug because this cop existed as an `enabled-by-default-cop` earlier known as `Style/FileName`

.rubocop.yml
ashmaroli: okay.. I'll go with enforcing `uppercase` (default)

.rubocop.yml
ashmaroli: - 'features/support/helpers.rb'
 - 'test/test_redcarpet.rb'
 - 'test/test_tags.rb'

are flagged for presence of `EOS` as the delimiter

.rubocop.yml
ashmaroli: This cop flags files that contain the following magic comment because from Ruby 2.0+, UTF-8 is the default source file encoding:
```ruby
# encoding: UTF-8
```
So, IMO, we switch to `when_needed` if we're to enable the cop -- only enforce an encoding comment if there are non-ASCII chars

lib/jekyll/converters/markdown/kramdown_parser.rb
ashmaroli: `each_key` broke our test-suite..

.rubocop.yml
parkr: 👍

.rubocop.yml
parkr: Yeah, but it's always better to rescue explicitly what you want to rescue instead of everything. We can do this in a subsequent pass.

.rubocop.yml
parkr: Oh we must return some value and it doesn't like that! We can enable it in a subsequent pass.

.rubocop.yml
parkr: Yeah... that looks like a bug. Weird.

.rubocop.yml
parkr: Let's do `when_needed`, then!

pathawks:
@jekyllbot: merge +dev

PR #6371, opened 2017-09-18
Update frontmatter.md

i-give-up:
Mention that you can set the front-matter variable `layout` to `null` if you don't want to use any layout file.

### DIFF
--- a/docs/_docs/frontmatter.md
+++ b/docs/_docs/frontmatter.md
@@ -65,10 +65,23 @@ front matter of a page or post.
 
           If set, this specifies the layout file to use. Use the layout file
           name without the file extension. Layout files must be placed in the
-          <code>_layouts</code> directory. Set this to <code>null</code> if you
-          don't want to use any layout file.
+          <code>_layouts</code> directory. 
 
         </p>
+        <ul>
+          <li>
+            Using <code>null</code> will produce a file without using a layout 
+            file. However this is overridden if the file is a post/document and has a 
+            layout defined in the <a href="../configuration/#front-matter-defaults">
+            frontmatter defaults</a>.
+          </li>
+          <li>
+            Starting from version 3.5.0, using <code>none</code> in a post/document will
+            produce a file without using a layout file regardless of frontmatter defaults.
+            Using <code>none</code> in a page, however, will cause Jekyll to attempt to
+            use a layout named "none".
+          </li>
+        </ul>
       </td>
     </tr>
     <tr>
--- a/docs/_docs/frontmatter.md
+++ b/docs/_docs/frontmatter.md
@@ -65,7 +65,8 @@ front matter of a page or post.
 
           If set, this specifies the layout file to use. Use the layout file
           name without the file extension. Layout files must be placed in the
-          <code>_layouts</code> directory.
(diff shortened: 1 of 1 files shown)


DirtyF:
@i-give-up Thanks for the add, FWIW since 3.5.0 you can also use `layout: none` with a slightly difference:

> Using `layout: none` will now produce a file with no layout. Equivalent to `layout: null`, with the exception that `none` is a truthy value and won't be overwritten by front matter defaults.

Maybe we could also mention this?

docs/_docs/frontmatter.md
```diff
@@ -65,7 +65,8 @@ front matter of a page or post.
 
           If set, this specifies the layout file to use. Use the layout file
           name without the file extension. Layout files must be placed in the
-          <code>_layouts</code> directory.
+          <code>_layouts</code> directory. Set this to <code>null</code> if you
+          don't want to use any layout file.
```
ashmaroli: The downside to the proposition is that `layout: null` will be overridden by `frontmatter_defaults` defined in for the document (?) in the config file. As of Jekyll 3.5.x, this is overcome by setting `layout: none` in the document. (Works only for `Jekyll:Document` objects. For regular `Jekyll::Page` objects (like `./about.md`), there's a proposed port via an open PR.)

i-give-up: So, currently to prevent override by `defaults` in `_config.yml`, you have to use?

- `layout: none` for documents
- `layout: null` for pages

Seems a bit messy.

ashmaroli: nope.. setting `layout: none` means it will not be overridden by `_config.yml`

i-give-up: I know. That's why I said currently you have to use `layout: null` for pages. If you use `layout: none` for pages it'll try to use a layout named `none`.

So to summarize

- Using `layout: null` will produce a file without layout file, but this will be overridden if the file is a Document and has a layout defined in front matter defaults.
- (version 3.5.0 and above) Using `layout: none` in a Document produce a file without a layout file. If used in a Page, it will try to use a layout named `none`.

ashmaroli: yes, that is 99.99% correct.. :smiley:  If things go as I hope, `v3.6.0` might land with `layout: none` applicable to Pages as well..

i-give-up:
Ok, made some changes. It's not as pretty as before, but should be a bit more accurate in describing `layout: null` and `layout: none`.

docs/_docs/frontmatter.md
```diff
@@ -65,9 +65,23 @@ front matter of a page or post.
 
           If set, this specifies the layout file to use. Use the layout file
           name without the file extension. Layout files must be placed in the
-          <code>_layouts</code> directory.
+          <code>_layouts</code> directory. 
 
         </p>
+        <ul>
```
DirtyF: Note: `tbody td ul ` font-size is currently not the same as `tbody td p` elements.  This should be taken care of in another PR. (or just limit ourselves to `p` tags in tables.)

DirtyF:
@jekylldocs: merge +docs

PR #6372, opened 2017-09-18
Docs: Lint Sass files

DirtyF:
Like we use Rubocop to lint and fix Ruby, we could also lint the website's Sass files with [Stylelint](https://stylelint.io/). We could also lint our markdown with `remark` but that's for another PR.

For the configuration, I went with a slightly less strict version of https://github.com/primer/stylelint-config-primer. No strong feelings here, I just wanted to adopt an existing configuration, there are other presets.

Please note that browser support is [defined in the upstream config](https://github.com/primer/stylelint-config-primer/blob/dce5f59b568de3d6abc87099577cf390a33b3fdd/index.js#L272-L279).

`npm run lint:sass` will make sure the main style file respect our conventions.

I wonder what @jekyll/core thinks of this approach, that's why I didn't add this check to the Rakefile or to CI for now.


pathawks:
Is this better than just using Sass itself to rewrite our Scss files and diffing the output?

DirtyF:
@pathawks I know Sass can check syntax, but that's pretty much it no? 

Stylelint allows you to define your _own_ rules (like Rubocop), so e.g. here the order of the properties is pre-defined, the max-nesting level allowed, the specificity or the duplicity of selectors, etc. 

Thanks to the numerous plugins Stylelint is very powerful (PostCSS is used under the hood).
Like for Rubopcop, thanks to Atom plugins, you have instant feedback if you don't follow the convention. I don't think they're any convention now for Sass files.

 That might be a bit overkill for this project, you tell me. It's well suited for teams who wanna stick to conventions.  I do like conventions, through guiding them,  it makes newcomers contribution easier.

PR #6377, opened 2017-09-21
Don't break if bundler is not installed

ashmaroli:
ref: #5660 [comment by @pathawks ](https://github.com/jekyll/jekyll/pull/5660#issuecomment-331191402)

/cc @pathawks

### DIFF
--- a/lib/jekyll/commands/new.rb
+++ b/lib/jekyll/commands/new.rb
@@ -132,6 +132,7 @@ RUBY
               require "bundler"
               bundle_install path
             rescue LoadError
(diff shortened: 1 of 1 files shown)


parkr:
> `raise SystemExit unless process.success?`

We should let the user know that they should try again with `--skip-bundle` or run `gem install bundler`.

In Ruby 2.5, Bundler will be installed by default. 🎉

parkr:
@jekyllbot: merge +fix

pathawks:
Thanks!

PR #6379, opened 2017-09-22
Docs: Style lists in tables

DirtyF:
Fix an edge case when a list is used in a table:

### Before

unstyled-ul

### After

styled-ul


ashmaroli:
funny that `<code/>` elements have a larger font-size.. (according to the pasted screenshot)

DirtyF:
@jekyllbot: merge +docs

PR #6388, opened 2017-09-27
Fix diff syntax-highlighting

ashmaroli:
Since this is subjective, its going to take some inputs from multiple perspective..

Current:
image

### PR Initial Commit:
diff-syntax

### DIFF
--- a/docs/_sass/_pygments.scss
+++ b/docs/_sass/_pygments.scss
@@ -13,11 +13,11 @@
   .cp { color: #cd5c5c} /* Comment.Preproc */
   .c1 { color: #87ceeb} /* Comment.Single */
   .cs { color: #87ceeb} /* Comment.Special */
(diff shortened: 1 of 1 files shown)


DirtyF:
@jekyllbot: merge +docs

DirtyF:
We should have kept the bold here but it's already more readable, that's what matters the most. Thanks.

ashmaroli:
Bold + text-shadow seemed a li'l too much to me..

ashmaroli:
@DirtyF see the extra borders around the `code` blocks in the pics above..? I'm guessing it has something to do with the change in Rouge versions.. (GitHub Pages is still on the older version..)
Is that worth addressing..?

DirtyF:
Yeah it's clearly a visual regression there's an extra `div` with `highlight` class that wasn't here before. We should remove it.

ashmaroli:
handled in #6389 ...

PR #6389, opened 2017-09-27
Add special styling for code-blocks run in shell

ashmaroli:
Proposing to add additional styling to indicate code-blocks with `sh` syntax-highlighting for added discernibility..

### Sample preview:

shell-syntax

Additional benefits:
  - This will allow us to easily detect which of the blocks have incorrectly highlighted as shell-syntax
  - consistently use `sh` for shell-highlighting..

/cc @jekyll/documentation

### DIFF
--- a/docs/_sass/_style.scss
+++ b/docs/_sass/_style.scss
@@ -682,15 +682,15 @@ h5 > code,
 
 .highlight {
   margin: 1em 0;
-  padding: 10px 0;
   width: 100%;
   overflow: auto;
 }
 
+pre.highlight { padding: 10px 0.5em; }
+
 .highlighter-rouge .highlight {
   @extend .highlight;
   margin: 0;
-  padding: 10px 0.5em;
 }
 
 /* HTML Elements */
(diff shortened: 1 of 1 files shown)


DirtyF:
I like this, as it resolves the problem of prefixing with having to add a `$` sign for shell commands and thus allow easier copy-pasting for users.

ProTip™: Remark presets are our friend to spot to ensure consistence:
- https://www.npmjs.com/package/remark-lint-fenced-code-flag
- https://www.npmjs.com/package/remark-lint-no-shell-dollars

ashmaroli:
Nice.. is there any related Linters in Ruby world..? If yes, we could add that to our Travis matrix as well..

DirtyF:
There is https://github.com/markdownlint/markdownlint

mattr-:
Is there a way to make the color of the `TERMINAL` bit the yellow color we use throughout the rest of the site? If so, I'd like to see a screenshot of what that looks like for comparision purposes. It feels like the black text on the gray background washes out the border a bit. Just my 2¢. :smiley:

ashmaroli:
> It feels like the black text on the gray background washes out the border a bit

I was shooting for a subtle-indication actually.. IMO, the yellow would only be visual noise..

#### Black Terminal
terminal

#### Yellow Terminal
yellow-terminal

DirtyF:
@jekyllbot: merge +site

PR #6390, opened 2017-09-27
Fix the code snippet for pagination.

miayam:
I copy-paste the code to my blog and something weird happened. The first page of pagination always redirect to previous page. Let's say if the current pagination is page number 4, the first pagination will be page number 3. It must redirect to root (home page).


DirtyF:
Just tested this and could not reproduce this issue. 
Default code snippet for posts pagination works fine.

Closing this issue,feel free to reopen with your detailed use case example in a public repository.

Issue #6394, opened 2017-09-28
Update dependency constraint to allow for cucumber v3.0.0


jekyllbot:
Hey there! :wave:

I noticed that the constraint you have for cucumber doesn't allow for the latest version to be used.

The constraint I found was `~> 2.1`, and the latest version available is `3.0.0`.

Can you look into updating that constraint so our users can use the latest and greatest version? Thanks! :revolving_hearts:

pathawks:
Blog post: https://cucumber.io/blog/2017/09/27/announcing-cucumber-ruby-3-0-0

Only breaking change with 3.0.0 is dropping support for Ruby 1.9 & 2.0

DirtyF:
As we don't support these versions anymore, the upgrade should be fine.

```sh
$ script/cucumber
WARNING: The formatter Jekyll::Cucumber::Formatter is using the deprecated formatter API which will be removed in v4.0 of Cucumber.
...
```

Very nice to warn from next breaking changes.

PR #6395, opened 2017-09-28
Upgrade to Cucumber 3.0

DirtyF:
Tests are passing but the `print_stats` method is now undefined.

```sh
undefined method `print_stats' for #<Jekyll::Cucumber::Formatter:0x007fa12cbeba48> (NoMethodError)
/Users/frank/code/jekyll/jekyll/features/support/formatter.rb:210:in `print_summary'
/Users/frank/code/jekyll/jekyll/features/support/formatter.rb:49:in `after_features'
```

Breaking change introduced in https://github.com/cucumber/cucumber-ruby/pull/999

### DIFF
--- a/script/cucumber
+++ b/script/cucumber
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
 time ruby -S bundle exec cucumber \
-  -f Jekyll::Cucumber::Formatter "$@"
+-f summary "$@"
--- a/script/cucumber
+++ b/script/cucumber
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
 time ruby -S bundle exec cucumber \
--f summary "$@"
+  --format progress "$@"
--- a/Gemfile
+++ b/Gemfile
@@ -23,7 +23,7 @@ end
 
 group :test do
   gem "codeclimate-test-reporter", "~> 1.0.5"
-  gem "cucumber", "~> 2.1"
+  gem "cucumber", "~> 3.0"
   gem "jekyll_test_plugin"
   gem "jekyll_test_plugin_malicious"
   # nokogiri v1.8 does not work with ruby 2.1 and below
--- a/features/support/formatter.rb
+++ b/features/support/formatter.rb
@@ -59,7 +59,7 @@ module Jekyll
       #
 
       def feature_element_timing_key(feature_element)
-        "\"#{feature_element.name.to_s.sub("Scenario: ", "")}\" (#{feature_element.location})"
+        "\"#{feature_element.name}\" (#{feature_element.location})"
       end
 
       #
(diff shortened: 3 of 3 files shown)


pathawks:
https://github.com/cucumber/cucumber-ruby/pull/999#discussion-diff-71759600

`print_stats` is now `print_statistics`

pathawks:
Fixes #6394

DirtyF:
@pathawks I thought we could go with the new `summary` format for now. 
The only difference is that it doesn't print the time for each test.

Plus the following warning message implies that if we want to continue to customize the cucumber tests outputs, some refactoring will be necessary one day or another:

```
WARNING: The formatter Jekyll::Cucumber::Formatter is using the deprecated formatter API which will be removed in v4.0 of Cucumber.
```

Could we tackle this in another PR?

pathawks:
> Could we tackle this in another PR?

Yeah, let’s do. :+1:

script/cucumber
```diff
@@ -1,4 +1,3 @@
 #!/usr/bin/env bash
 
-time ruby -S bundle exec cucumber \
-  -f Jekyll::Cucumber::Formatter "$@"
+time ruby -S bundle exec cucumber --format summary
```
pathawks: Just for Git History, could we leave line 3 as is?

```sh
time ruby -S bundle exec cucumber \
  --format summary
```

pathawks:
Thanks @DirtyF.

I'm looking at using the new events API to get our timing information :+1:

script/cucumber
DirtyF: Done.

DirtyF:
@pathawks  ❤️ 

Here is the summary display (very readable if you ask me)

cucumber-summary-jekyll

and the final summary:

```sh
212 scenarios (212 passed)
1763 steps (1763 passed)
4m16.614s
```

ashmaroli:
Wow.. nice..
Can you post a view for a failing test too..? (easily done locally..)

ashmaroli:
also I think this leaves `/features/support/formatter.rb` reduntant..

script/cucumber
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
 time ruby -S bundle exec cucumber \
-  -f Jekyll::Cucumber::Formatter "$@"
+-f summary "$@"
```
ashmaroli: It'd be nice to leave the indentation as ln#4 is a continuation of ln#3.. instead of a new expression..

pathawks:
> also I think this leaves `/features/support/formatter.rb` reduntant..

No, I'm working on adding back some of that functionality. Let's keep it there.

pathawks:
cuc

Abbreviated to only show one group, but you can get the gist.

pathawks:
I added a commit that will more closely resemble the previous output of Cucumber, but without using our own formatter.

This is just a first pass, and I tried to change as little as reasonable. Much of `features/support/formatter.rb` really could probably be removed, and I'm not sure if just sticking `AfterConfiguration` at the end is a good idea.

@DirtyF Feel free to change anything I added or just remove my commit altogether. Either way, I'm good upgrading to Cucumber 3.0 and we can tweak things as we go if need be.

pathawks:
ac61f65cb7363a6a46067f86316c38fd2159f807
cuc

DirtyF:
@pathawks We'll discuss formatting in another PR. Thanks for your help on this. 🤝 
 
@jekyllbot: merge +dev

PR #6396, opened 2017-09-29
Docs: Update instructions

DirtyF:
Promoting `bundler update` and adding the upgrade to minor versions.


mattr-:
@jekyllbot: merge +site

PR #6398, opened 2017-10-01
Fix code-block highlighting in docs

ashmaroli:
- Remove `$` from code-blocks for easier copy-pasting
- Add correct syntax-highlighting where previously not present
- Correct erroneous syntax-highlighting markers based on context

/cc @jekyll/documentation


DirtyF:
@jekyllbot: merge +docs

Issue #6003, opened 2017-04-03
Categories are poorly documented


duckinator:
The most common question I've seen around the internet when it comes to Jekyll is how to group posts in the filesystem. I'd always assumed &mdash; based on none of those posts I've seen ever coming to an answer &mdash; that Jekyll didn't have this feature, but eventually someone off-handedly mentioned categories, so I looked it up.

I found this comment, from 2015: https://github.com/jekyll/jekyll/issues/1819#issuecomment-89003858

That comment points to 3 sentences on http://jekyllrb.com/docs/variables/#page-variables

I also found a brief mention of it here: https://jekyllrb.com/docs/frontmatter/#predefined-variables-for-posts

But it's never actually properly documented on its own anywhere, and I feel like it should be.

pathawks:
What sort of information would you like to see?

Out of the box, Jekyll doesn't _do_ much with categories; a plugin is needed to enable category archive pages.

I agree we could do better, just trying to get the discussion going about specific areas of improvement.

duckinator:
Honestly, I can't find much information at all about it, so I'm not sure where to begin.

A good starting point would probably be:

1. what it can do without a plugin.
2. what the plugin adds.
3. an example setup or two.

I'm rather limited in terms of how much I can contribute currently due to health issues, so any contributions I make will be pretty sporadic and will probably take a while. But if someone else dumps any relevant information or links in comments on here, I can look into organizing it when I'm up for it.

DirtyF:
It's true that there are no dedicated categories or tags pages in the core documentation yet, I guess you are looking for something like :

### Working with Categories

- Setting up your directory structure
- Defining categories in the YAML Front Matter
- Customizing permalinks
- Categories pagination
- Categories archives 
- Using categories in Collections

That might be a good fit for a [tuturial](https://jekyllrb.com/tutorials/home/).

duckinator:
@DirtyF yeah, a tutorial sounds like a good fit for this.

zhuqingguang:
I have seen the code like this  ``{{ category | last | size }}``，but I can't find anything about the constructure about category, if there are some document about this ?
Thank you for your reply.

adamvoss:
I found @benbalter's [blog post about Jekyll Collections](http://ben.balter.com/2015/02/20/jekyll-collections/) quite informative.  It convinced me I should be using collections after I had decided based on the Jekyll documentation that I didn't need to use collections in my use case.
__
[...]

PR #6399, opened 2017-10-01
Docs: Filtering Posts with categories, tags, or other variables

kentonh:
I swear there is an archived issue that I found last week about this, but for the life of me, I can't find it now. However, I still think this is valuable. I put this together to show what's possible with just basic Jekyll.

Ninja edit:
This solves issue #6003

### DIFF
--- a/docs/_docs/posts.md
+++ b/docs/_docs/posts.md
@@ -78,7 +78,7 @@ digital assets along with your text content. While the syntax for linking to
 these resources differs between Markdown and Textile, the problem of working
 out where to store these files in your site is something everyone will face.
 
-There are a number of ways to include digital assets in Jekyll. 
+There are a number of ways to include digital assets in Jekyll.
 One common solution is to create a folder in the root of the project directory
 called something like `assets` or `downloads`, into which any images, downloads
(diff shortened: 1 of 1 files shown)


docs/_docs/posts.md
```diff
@@ -155,6 +155,51 @@ you wish to access the currently-rendering page/posts's variables (the
...
+or a template displaying just the posts in those categories you specify. Here's
+a basic example of how to create a list of posts from a specific category.
+
+First, in the `_layouts` directory create a new file called `category.html` - in
+that file put (at least) the following:
+```html
+---
+layout: page
+---
+{% for post in site.categories.[page.category] %}
```
parkr: I have never seen the `.[]` syntax before. Is that new?

I believe `site.categories[page.category]` will work here – please correct me if I'm wrong! This approach, if it's valid, will make this look a bit more like other programming languages, which can help folks new to Liquid understand what's going on here.

kentonh: It might be less "new" and more "what has worked for me."

I'll edit to be standardized.

DirtyF:
@jekyllbot: merge +docs

Issue #6397, opened 2017-09-30
Cannot override JEKYLL_ENV in `jekyll build`


JasonYao:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [x] I ran `jekyll doctor` to check my configuration
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [ ] This is a feature request.
---

- [x] I am on (or have tested on) ***macOS*** 10+
- [ ] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux
- [ ] I am on (or have tested on) ***Fedora*** GNU/Linux
- [ ] I am on (or have tested on) ***Arch*** GNU/Linux
- [ ] I am on (or have tested on) ***Other*** GNU/Linux
- [ ] I am on (or have tested on) ***Windows*** 10+

---

- [ ] I was trying to install.
- [ ] There is a broken Plugin API.
- [ ] I had an error on GitHub Pages, and I have reproduced it locally.
- [ ] I had an error on GitHub Pages, and GitHub Support said it was a Jekyll Bug.
- [ ] I had an error on GitHub Pages and I did not test it locally.
- [ ] I was trying to build.
- [x] It was another bug.

## My Reproduction Steps
```sh
bundle exec jekyll build JEKYLL_ENV=development
```

## The Output I Wanted
Expected result:
- `{{ site.url }}`s should resolve to `http://localhost:4000`

Actual result:
- `{{ site.url }}`s resolved to the production value in `_config.yml`

Additional testing:
- `bundle exec jekyll build JEKYLL_ENV=dev` (same result)
- `bundle exec jekyll build JEKYLL_ENV=develop` (same result)

## Questions
- Does `jekyll serve` set more settings than just the `JEKYLL_ENV`?
- What are the settings required for me to change `jekyll build` to produce
a development build and not a production build?

DirtyF:
What Jekyll version are you running? http://localhost:4000 is already the default in development environment since Jekyll 3.3.

`JEKYLL_ENV=development bundle exec jekyll build` 

You can pass host and port as options.

```shell
-H, --host [HOST]  Host to bind to
-P, --port [PORT]  Port to listen on
```

I use `JEKYLL_ENV development bundle exec jekyll serve --host 0.0.0.0` for instance when I want to access the local server through my mobile device.

[...]

PR #6400, opened 2017-10-01
Bugfixed `jekyll build` and moved environment check from `jekyll serve` to `jekyll build`

JasonYao:
## Patch Information
- Issue related to: [6397](https://github.com/jekyll/jekyll/issues/6397)

## What Does This Patch Do
- Enabled `jekyll build` to be able to set its `url` configuration parameter through the commandline via `JEKYLL_ENV`, specifically to the default value (`localhost:4000`) if `JEKYLL_ENV=development`

## Current Master Branch Status
- `jekyll serve` will automatically resolve `url` to `jekyll serve`'s default (`localhost:4000`)
- `jekyll build` will automatically resolve `url` to a production build (whatever the value in `_config.yml` is) <-- ANOTHER BUG, since per the [docs](https://jekyllrb.com/docs/configuration/#specifying-a-jekyll-environment-at-build-time), the default for `jekyll build` should have been `JEKYLL_ENV=development` in the first place
- `JEKYLL_ENV=development jekyll build` is bugged, and still resolves `url` to a production build (whatever the value in `_config.yml` is)

## Wanted Status Changes with this patch
- [ ] `jekyll serve` will automatically resolve `url` to `jekyll build`'s default (`http://localhost:4000`)
- [ ] `jekyll build` will automatically resolve `url` to `jekyll build`'s default (`http://localhost:4000`)
- [ ] `JEKYLL_ENV=production jekyll build` will automatically resolve `url` to a production build (whatever the value in `_config.yml` is)
- [ ] `JEKYLL_ENV=some_other_value` will automatically resolve `url` to a development build (`http://localhost:4000`)

### DIFF
--- a/lib/jekyll/commands/build.rb
+++ b/lib/jekyll/commands/build.rb
@@ -27,6 +27,12 @@ module Jekyll
           Jekyll.logger.adjust_verbosity(options)
 
           options = configuration_from_options(options)
+
+          # Enables environment setting
+          if Jekyll.env == "development"
+            options["url"] = default_url(config)
+          end
+
           site = Jekyll::Site.new(options)
 
           if options.fetch("skip_initial_build", false)
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -36,9 +36,7 @@ module Jekyll
               opts["watch"  ] = true unless opts.key?("watch")
 
               config = configuration_from_options(opts)
-              if Jekyll.env == "development"
-                config["url"] = default_url(config)
-              end
+
               [Build, Serve].each { |klass| klass.process(config) }
             end
           end


pathawks:
> Fixes the `jekyll build` command to actually have `JEKYLL_ENV=development` be the default, and not the production build, per the docs

This is unacceptable for Jekyll 3.x

When in `development` mode, `site.url` is overridden. This is fine when using `jekyll serve`, but would break a lot of sites for users that have never used `jekyll serve`, but just use Jekyll to build a site and then deploy.

JasonYao:
@pathawks Gotcha, in that case the documentation for https://jekyllrb.com/docs/configuration/#specifying-a-jekyll-environment-at-build-time will need to be updated as well in this PR to reflect the fact that `jekyll build`'s default environment is `production`

lib/jekyll/commands/build.rb
```diff
@@ -27,6 +27,12 @@ def process(options)
           Jekyll.logger.adjust_verbosity(options)
 
           options = configuration_from_options(options)
+
+          # Enables environment setting
+          if Jekyll.env == "development"
+            options["url"] = default_url(config)
```
parkr: `s/config/options`

pathawks:
@JasonYao Yeah, clearly the documentation is just wrong.

I wouldn’t mind supporting `JEKYLL_ENV` for `build`, but it would need to default to `production` to maintain compatibility.

parkr:
> I wouldn’t mind supporting JEKYLL_ENV for build, but it would need to default to production to maintain compatibility.

Hm, can you clarify?

```diff
diff --git a/lib/jekyll/commands/build.rb b/lib/jekyll/commands/build.rb
index 1bd176b8..0f3b7455 100644
--- a/lib/jekyll/commands/build.rb
+++ b/lib/jekyll/commands/build.rb
@@ -29,6 +29,8 @@ module Jekyll
           options = configuration_from_options(options)
           site = Jekyll::Site.new(options)

+          p Jekyll.env
+
           if options.fetch("skip_initial_build", false)
             Jekyll.logger.warn "Build Warning:", "Skipping the initial build." \
                         " This may result in an out-of-date site."
```

outputs `"development"`. Jekyll defaults to `development`, not `production`.

pathawks:
But previously this had no impact when running `jekyll build`

If we change this, anybody who runs `jekyll build` and then uploads the generated site to a server is going to be in for a nasty suprise when all their URLs point to `localhost`

JasonYao:
@parkr Looks like we're conflating the two- IIUC @pathawks and I are assuming:

| JEKYLL_ENV value | Expected behavior regarding `{{ site.url }}` |
| --- | --- |
| production | `{{ site.url }}` resolves to whatever value is set in `_config.yml` |
| development | `{{ site.url }}` resolves to `http://localhost:4000` |

Assuming that is correct, it looks like `jekyll build` is currently resolving to the incorrect value. There's 2 ways to proceed:
- Either we set `JEKYLL_ENV` to be set to `production` (to match what the current functionality is) and we update the documentation to say that `production` is the default
- Or we leave `JEKYL_ENV` to be `development` (breaking old functionality), and make the expected behavior match what the docs say [here](https://jekyllrb.com/docs/configuration/#specifying-a-jekyll-environment-at-build-time)



that when `JEKYLL_ENV=production`, that expected behavior

oe:
Looks to me like there's no clear consensus on whether this goes into 3.8 or 4.x. Does anyone have more arguments that could help us reach a consensus here?

ashmaroli:
I'm sticking with `4.0` on this one..
Since `jekyll build` had generated production-ready URLs, till date, we shouldn't just change this and update the docs hoping that users will eventually read the updated version..

We broke a workflow once, when we automatically assigned `site.url = http://localhost:4000` by default on `jekyll serve`. Users complained, but forgave us because they could just run `jekyll build` and get production-ready URLs by default..

We shouldn't test our users once again..

ashmaroli:
That said, if we do want to release this functionality in `3.x`, then we should make sure that the URLs from `jekyll build` does not have `http://localhost:4000` in them, ever...!

lib/jekyll/commands/build.rb
ashmaroli: you can't just move a block of code from one class to another without looking for collateral damage..
I don't think `Commands::Build` even has a `#default_url` method..

pathawks:
I _really_ don't like the idea of Jekyll changing my `site.url` without my permission. I understand that some users would benefit from setting `site.url = localhost` during `jekyll build`, but I don't think it should _ever_ be the default.

For Jekyll 4.0, we could set `site.url = localhost` for both `build` and `serve` when `JEKYLL_ENV = development`. Consistency is good.
?

PR #6401, opened 2017-10-02
Front matter defaults without default_proc

parkr:
**DO NOT MERGE**.

This is a PoC of a system which shims accessing data in front of front
matter defaults to eliminate the need for default_proc.

Use of Hash#default_proc has caused some confusion and frustration,
especially when trying to marshal and unmarshal Documents, or Pages, or
any other Hash which has an attached default_proc. Unfortunately, procs
cannot be marshaled.

All the default_proc was doing was asking the site's frontmatter
defaults which values applied to the given object and passing them
along. Here, we place a custom class in front of the Hash data and reach
out to the front matter defaults when applicable for the given object.

An alternative method would be to strip and re-add the default_proc
before and after marshaling, respectively.

@envygeeks, I know you were interested in us getting rid of
default_proc, so let me know if this is an acceptable solution for your
needs. If we can hook into the marshaling & unmarshaling processes, then
we can remove & re-add the default_proc quite easily there as well.

### DIFF
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -48,6 +48,7 @@ module Jekyll
   autoload :Excerpt,             "jekyll/excerpt"
   autoload :External,            "jekyll/external"
   autoload :FrontmatterDefaults, "jekyll/frontmatter_defaults"
+  autoload :FrontmatterWithDefaults, "jekyll/frontmatter_with_defaults"
   autoload :Hooks,               "jekyll/hooks"
   autoload :Layout,              "jekyll/layout"
   autoload :CollectionReader,    "jekyll/readers/collection_reader"
--- a/lib/jekyll/convertible.rb
+++ b/lib/jekyll/convertible.rb
@@ -23,6 +23,17 @@ module Jekyll
       content || ""
     end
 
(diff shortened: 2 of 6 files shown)


envygeeks:
Why wouldn't you just merge the user config into your defaults as a copy and then cache it on `@config`?

envygeeks:
The only problems I see here is the lack of enumerators, thus far I see no problem, though I think you would have to provide a marshaler for your class (can't remember if it's still *explicitly required*, I don't think it is if it can detect `#to_h` and others.  The question really then becomes when you do a `#to_h` should you add the front-matter defaults to that, or should you just add a custom marshal and unmarshal to the class that spits that out in a way that can be restored quickly.

Soon I plan on looking into  using Concurrent-Ruby with Jekyll, I don't remember if it will hang up on `default_proc` (I don't normally use them) but I think the removal of default procs would be better all around as my old solution can be cleaned up significantly and reshipped with half the code at that point.

parkr:
I'm not going to continue down this road. I think wrapping the Marshaler interface where we remove, marshal, then re-add the default_proc is best.

PR #6405, opened 2017-10-02
Fixes formatting on pre-formatted text.

kentonh:
Pre-formatted text was marked as html, and so the important example was left out.


DirtyF:
@jekyllbot: merge +docs

PR #6407, opened 2017-10-02
Docs: updates

DirtyF:
A mass review of the documentation addressing:

- Code blocks / raw tags uniformisation for better source readability
- highlight liquid rather than html in code blocks
- Tip: `jekyll-paginate-v2` for pagination
- Tip: `jekyll-livereload` for serve
- update Shopify Liquid Help URL
- remove obsolete `site.help_url`

### DIFF
--- a/docs/_docs/contributing.md
+++ b/docs/_docs/contributing.md
@@ -125,19 +125,27 @@ site or configuration. [Feel free to check it out!](https://github.com/jekyll/je
 
 To run the test suite and build the gem you'll need to install Jekyll's dependencies by running the following command:
 
-<pre class="highlight"><code>$ script/bootstrap</code></pre>
+```sh
+script/bootstrap
+```
 
 Before you make any changes, run the tests and make sure that they pass (to confirm your environment is configured properly):
 
-<pre class="highlight"><code>$ script/cibuild</code></pre>
+```sh
+script/cibuild
+```
 
 If you are only updating a file in `test/`, you can use the command:
 
-<pre class="highlight"><code>$ script/test test/blah_test.rb</code></pre>
+```sh
+script/test test/blah_test.rb
+```
 
 If you are only updating a `.feature` file, you can use the command:
 
-<pre class="highlight"><code>$ script/cucumber features/blah.feature</code></pre>
+```sh
+script/cucumber features/blah.feature
+```
 
 Both `script/test` and `script/cucumber` can be run without arguments to
 run its entire respective suite.
--- a/docs/_docs/posts.md
+++ b/docs/_docs/posts.md
@@ -80,53 +80,52 @@ out where to store these files in your site is something everyone will face.
 
 There are a number of ways to include digital assets in Jekyll.
 One common solution is to create a folder in the root of the project directory
-called something like `assets` or `downloads`, into which any images, downloads
+called something like `assets`, into which any images, files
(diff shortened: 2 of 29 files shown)


docs/_docs/contributing.md
```diff
@@ -125,19 +125,27 @@ site or configuration. [Feel free to check it out!](https://github.com/jekyll/je
 
 To run the test suite and build the gem you'll need to install Jekyll's dependencies by running the following command:
 
-<pre class="highlight"><code>$ script/bootstrap</code></pre>
+```sh
+script/bootstrap
+```
```
parkr: I believe this page is generated, so this will have to be updated in `.github/CONTRIBUTING.md`. I may be wrong though!

docs/_docs/posts.md
```diff
@@ -170,24 +171,27 @@ a basic example of how to create a list of posts from a specific category.
...
+
+{% raw %}
+```liquid
 ---
 layout: page
 ---
+
 {% for post in site.categories[page.category] %}
-    <a href="{{ post.url | prepend: site.baseurl }}">
+    <a href="{{ post.url | absolute_url }}">
```
parkr: This makes it `| prepend: site.baseurl | prepend: site.url` which is technically a change in behaviour. Just noting it. 🤷‍♂️

docs/_docs/themes.md
```diff
@@ -218,13 +218,15 @@ _sass
...
 ```css
-{% raw %}@import "{{ site.theme }}";{% endraw %}
+@import "{{ site.theme }}";
 ```
+{% endraw %}
 
 ### Theme-gem dependencies
 
-From `v3.5`, Jekyll will automatically require all whitelisted `runtime_dependencies` of your theme-gem even if they're not explicitly included under the `gems` array in the site's config file. (NOTE: whitelisting is only required when building or serving with the `--safe` option.)
+From `v3.5`, Jekyll will automatically require all whitelisted `runtime_dependencies` of your theme-gem even if they're not explicitly included under the `gems` array in the site's config file. (Note: whitelisting is only required when building or serving with the `--safe` option.)
```
ashmaroli: `gems` array and/or `plugins` arrays..

docs/_docs/windows.md
```diff
@@ -17,45 +17,45 @@ If you are using Windows 10 Anniversary Update, the easiest way to run Jekyll is
 
 First let's make sure all our packages / repositories are up to date. Open a new Command Prompt instance, and type the following:
 
-```
+```shell
```
ashmaroli: can these be changed to <code>\`\`\`sh</code> please..? I don't think the new `TERMINAL` styles get applied to `shell` ...

DirtyF: Now it does.  Stuck to ````sh` for consistency, but we're bulletproof.

docs/_docs/themes.md
DirtyF: Thanks for noticing, as we mention v3.5, let's just use `plugins` here as we did in #6082

docs/_docs/posts.md
DirtyF: True, I thought we should encourage the use of the filter in our docs.  Should we?
There is no URLs good practices in http://ben.balter.com/jekyll-style-guide/ yet but it might be a good candidate.

DirtyF:
@jekyllbot: merge +docs

PR #6409, opened 2017-10-03
Docs: Fix Ruby 2.4 path for Windows

rashevskyv:
Wrong ruby 2.4 path

### DIFF
--- a/docs/_docs/windows.md
+++ b/docs/_docs/windows.md
@@ -112,7 +112,7 @@ The development kit did not set the environment path for Ruby so we need to do i
 
  * Open command prompt in `C:\tools\DevKit2`
  * Execute `ruby dk.rb init` to create a file called `config.yml`
- * Edit the `config.yml` file and include the path to Ruby `- C:/tools/ruby22`
+ * Edit the `config.yml` file and include the path to Ruby `- C:/tools/ruby24`
  * Execute the following command to set the path: `ruby dk.rb install`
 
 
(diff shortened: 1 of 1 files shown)


docs/_docs/windows.md
```diff
@@ -112,7 +112,7 @@ The development kit did not set the environment path for Ruby so we need to do i
 
  * Open command prompt in `C:\tools\DevKit2`
  * Execute `ruby dk.rb init` to create a file called `config.yml`
- * Edit the `config.yml` file and include the path to Ruby `- C:/tools/ruby22`
+ * Edit the `config.yml` file and include the path to Ruby `- C:/tools/ruby24`
```
XhmikosR: I don't think this is valid judging by the Ruby version mentioned above (2.2.4).

docs/_docs/windows.md
```diff
@@ -129,13 +129,12 @@ choco install libxslt -Source "https://www.nuget.org/api/v2/"
...
 choco install libiconv -Source "https://www.nuget.org/api/v2/
 
-gem install nokogiri --^
-   --with-xml2-include=C:\Chocolatey\lib\libxml2.2.7.8.7\build\native\include^
-   --with-xml2-lib=C:\Chocolatey\lib\libxml2.redist.2.7.8.7\build\native\bin\v110\x64\Release\dynamic\cdecl^
-   --with-iconv-include=C:\Chocolatey\lib\libiconv.1.14.0.11\build\native\include^
-   --with-iconv-lib=C:\Chocolatey\lib\libiconv.redist.1.14.0.11\build\native\bin\v110\x64\Release\dynamic\cdecl^
-   --with-xslt-include=C:\Chocolatey\lib\libxslt.1.1.28.0\build\native\include^
-   --with-xslt-lib=C:\Chocolatey\lib\libxslt.redist.1.1.28.0\build\native\bin\v110\x64\Release\dynamic
+gem install nokogiri -- --with-xml2-include=C:\Chocolatey\lib\libxml2.2.7.8.7\build\native\include
```
XhmikosR: These are nice but could be split to a separate PR. Or you could change the snippet's highlighting to `bat` and leave the lines as they where. `^` is the right way to continue a line in Windows batch files.

i-give-up:
It's also worth noting that RubyInstaller version 2.4 onwards uses MSYS2 for native compilation of gems instead of a separate Devkit (see [here](https://rubyinstaller.org/add-ons/devkit.html)) so those parts in this document would have to be updated as well.

DirtyF:
Closing as Windows docs now recommend using RubyInstaller.

PR #6408, opened 2017-10-03
Docs: Fix `collections_dir` example

ashmaroli:
- `collections_dir` is a top-level config key
  - use example collections referred to just immediately

Thanks to @lsrdg and @hnaseri for trying the unreleased feature out and consequently opening tickets at Jekyll Talk:

  - Ticket 1: http://talk.jekyllrb.com/t/solved-listing-pages-of-a-collection/1011
  - Ticket 2: http://talk.jekyllrb.com/t/how-all-collections-inside-one-subdirectory-works/1034

PR #6412, opened 2017-10-04
Docs: Mark `collection_dir` as unreleased

DirtyF:
Following #6408 

Our Style Guide provide has a specific note type for unreleased features, let us make use of it.

collection_dir-unreleased

Unreleased notes are not meant to embed entire code snippets, but that's OK the feature is still understandable.

### DIFF
--- a/docs/_docs/collections.md
+++ b/docs/_docs/collections.md
@@ -46,22 +46,14 @@ defaults:
       layout: page
 ```
 
-**From `v3.7.0`**: You can optionally specify a directory if you want to store all your collections
-in the same place:
+<div class="note unreleased">
+  <h5>Gather your collections</h5>
 
-```yaml
-collections_dir: my_collections
-collections:
-  books:
-    foo: bar
-    output: true
-  recipes:
-    foo: baz
-    output: true
-```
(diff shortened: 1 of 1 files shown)


ashmaroli:
Cool! Always wondered if that style was ever used..
What happens when `v3.7.0` is released? Does JekyllBot remind you to make the switch once again or do you keep tabs manually?

pathawks:
> What happens when v3.7.0 is released? 

Users running older versions of Jekyll will see that they beed to upgrade to use this feature:+1:

DirtyF:
> Does JekyllBot remind you to make the switch once again or do you keep tabs manually?

It's all manual for now.

When we create the issue for a release, we should always check that new features are properly documented. Ideally, we specifically ask for it in the CONTRIBUTING template, the same way we ask to add tests. At least we make sure all new features are documented by their authors. 

I'm not sure what to automate here, but it should be a blocker before merging a feature.

ashmaroli:
> I'm not sure what to automate here

The "golden_box" represents a note on an unreleased feature. When the said version is released, the "golden-box" becomes a misnomer..
So in an ideal universe, JekyllBot steps in and creates an issue reminding the maintainers about the discrepancy..

ashmaroli:
I'm not saying it needs to be so.. was only wondering aloud.. :wink:

DirtyF:
@jekyllbot: merge +docs

Issue #6241, opened 2017-07-25
For gem-based themes, an incremental build run always as FULL


jueadams:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [x] I ran `jekyll doctor` to check my configuration

---

- [x] I am on (or have tested on) ***macOS*** 10+
- [x] I am on (or have tested on) ***Ubuntu*** GNU/Linux
- [x] I am on (or have tested on) ***Windows*** 10+

---

- [x] I was trying to build.

## My Reproduction Steps

```
jekyll --version
jekyll 3.5.1
```
1. jekyll new minima
2. jekyll doctor
```
Configuration file: /Users/jadams/minima/_config.yml
  Your test results are in. Everything looks fine.
```
3. jekyll s -i --profile

```
  Configuration file: /Users/jadams/minima/_config.yml
           Cleaner: Nothing to do for /Users/jadams/minima/_site.
           Cleaner: Nothing to do for /Users/jadams/minima/.jekyll-metadata.
           Cleaner: Nothing to do for /Users/jadams/minima/.sass-cache. 
  Configuration file: /Users/jadams/minima/_config.yml
            Source: /Users/jadams/minima
       Destination: /Users/jadams/minima/_site
   Incremental build: enabled
      Generating...

  Filename                                                                    | Count |  Bytes |  Time
  ----------------------------------------------------------------------------+-------+--------+------
_posts/2017-07-25-welcome-to-jekyll.markdown                                |     1 |  1.52K | 0.135
_layouts/default.html                                                       |     4 | 23.82K | 0.030
usr/local/lib/ruby/gems/2.3.0/gems/minima-2.1.1/_includes/head.html         |     4 |  2.44K | 0.012
feed.xml                                                                    |     1 |  3.75K | 0.010
usr/local/lib/ruby/gems/2.3.0/gems/minima-2.1.1/_includes/header.html       |     4 |  5.34K | 0.009
usr/local/lib/ruby/gems/2.3.0/gems/minima-2.1.1/_includes/footer.html       |     4 | 11.78K | 0.007
_layouts/post.html                                                          |     1 |  2.07K | 0.001
_layouts/home.html                                                          |     1 |  0.40K | 0.001
usr/local/lib/ruby/gems/2.3.0/gems/minima-2.1.1/_includes/icon-github.html  |     4 |  4.18K | 0.001
[...]

PR #6417, opened 2017-10-07
Provide a better default hash for tracking liquid stats

mattr-:
This avoids doing the `||=` check for an existing value when setting values for the per-file stats hash in the liquid renderer. It also gives us the flexibility to add new numerical values to this hash later without having to do the same `||=` check as we can just know that the starting value will be zero.

Discovered while I was investigating #6241


mattr-:
@jekyllbot: merge +dev

pathawks:
Thanks @mattr- :beers:🎊

PR #6322, opened 2017-08-18
Problematic UTF+bom files

DirtyF:
add test files for #2853

PR #6433, opened 2017-10-17
Fix BOM-related read problems

parkr:
Utils.merged_file_read_opts was attempting to shim in `bom|<encoding>` to a Hash key that didn't exist...

/cc @jekyll/build https://github.com/jekyll/jekyll/pull/6322


Crunch09:
@parkr Do you need to change the [failing test](https://github.com/jekyll/jekyll/blob/69241f2707f3955690ff1612a6a8bb838946bf0f/test/test_utils.rb#L393) to also use the symbol instead of the string?

parkr:
@Crunch09 Yes!

parkr:
CI passes on my machine.

Issue #123, opened 2010-01-31
pygment highlighting broken in ruby 1.9


icco:
in ruby 1.9 you get an array of strings printed out, instead of the actual strings, when highlighting code.

PR #6468, opened 2017-10-23
Do not linkify escaped characters as PRs in History

### LINKED ISSUE #6461: History escaping characters improperly
From https://jekyllrb.com/docs/history/#documentation:

> Remove duplicate &amp;&#35;34;available&amp;&#35;34; (#6380)

It looks like this is being escaped twice somewhere. That should be fixed.

jekyllbot:

```text
Do not linkify escaped characters as PRs in History
```

Fixes #6461


pathawks:
The regular expression that we were using to match PRs in the format `#123` was also matching escaped characters in the format `&#123;` which was causing weird things to happen in the generated history on the website.

Thanks to @ashmaroli for pointing out this problem.

Rakefile
```diff
@@ -46,7 +46,7 @@ def normalize_bullets(markdown)
 end
 
 def linkify_prs(markdown)
-  markdown.gsub(%r!#(\d+)!) do |word|
+  markdown.gsub(%r!(?<=[^&])#(\d+)!) do |word|
```
pathawks: This should really be a *negative lookbehind*, but I couldn't get that to work. Instead, this is a *positive lookbehind* that asserts that the preceding character was anything other than a `&`.

If anybody with some RegExFoo can make this better, that would be lovely.

pathawks:
Fixes #6461

mattr-:
>oh lord! this code is not tested... 😆

:joy::joy::joy::joy:

@jekyllbot: merge +dev

PR #6469, opened 2017-10-23
Rely on jekyll-mentions for linking usernames

jekyllbot:

```text
Rely on jekyll-mentions for linking usernames

This is more robust than our regex as it will not link anything that is
inside of backticks, like `@config` or `@options`
```


pathawks:
And [**jekyll-mentions**](https://github.com/jekyll/jekyll-mentions) _is_ included in our site config already. :+1:

https://github.com/jekyll/jekyll/blob/c0e4ce13b5e8439eb7037e8bc16fd2fe907afcd6/docs/_config.yml#L48-L55

docs/_docs/history.md
```diff
@@ -1436,7 +1436,7 @@ note: This file is autogenerated. Edit /History.markdown instead.
 ### Site Enhancements
 {: #site-enhancements-v2-5-3}
 
-- Add `[@alfredxing](https://github.com/alfredxing)` to the `[@jekyll](https://github.com/jekyll)/core` team. :tada: ([#3218]({{ site.repository }}/issues/3218))
+- Add `@alfredxing` to the `@jekyll/core` team. :tada: ([#3218]({{ site.repository }}/issues/3218))
```
pathawks: Should probably remove the backticks around the name here, so it links properly.

Rakefile
```diff
@@ -51,14 +51,8 @@ def linkify_prs(markdown)
...
-def linkify_users(markdown)
-  markdown.gsub(%r!(@\w+)!) do |username|
-    "[#{username}](https://github.com/#{username.delete("@")})"
-  end
-end
-
 def linkify(markdown)
-  linkify_users(linkify_prs(markdown))
+  linkify_prs(markdown)
 end
```
DirtyF: So this function is useless (for) now, we can directly call `linkify_prs` in `converted_history`

DirtyF:
@jekyllbot: merge +dev

PR #5997, opened 2017-04-01
Bump Rubocop to v0.48

ashmaroli:
Fixes #5992, fixes #6005
~~WIP~~

Rubocop `0.48` introduced some new cops that broke our tests. While we temporarily locked to the previous version `0.47.1`, this pull will remove the lock to bump Rubocop to `v0.48` and beyond.

Lets discuss if the fixes proposed here are acceptable.

The remaining catches appear to me as `bugs` in Rubocop-0.48

PR #6470, opened 2017-10-23
Rubocop: Enable more cops

jekyllbot:

This was disabled in #5997 because it was new to Rubocop. I think it can
be enabled now.


ashmaroli:
I think the following can be `enabled` with, minimal-to-no-collateral damages, for consistency..
```yaml
Layout/SpaceInsideBrackets
Style/Attr
Style/IfUnlessModifier
Style/RedundantReturn
Style/UnneededCapitalW
```

ashmaroli:
@pathawks What's the status of this..? If you don't have the time, I can contribute to this branch by opening another PR to this branch (`pull/InverseMethods`)

pathawks:
This is pretty low priority; I’ll have some time in the next few days to return to this.

I certainly wouldn’t be offended if you did some work on this, but I can’t help but wonder if there are better uses of our efforts.

pathawks:
@ashmaroli If you wanted, you could look into only rescuing from exceptions that might actually be thrown.

I have disabled this cop selectively in 1b62cb51e75ae468a10d2698bfa3728109834142. I would _like_ to amend these methods so they use `rescue` properly, but in order to do it without breaking anything, we do need to make sure that we are rescuing _anything_ that could be raised.

I won't have time to do this--probably ever--but it would be nice.

ashmaroli:
> I would *like* to amend these methods so they use `rescue` properly

Okay, I'll work on that..

pathawks:
I worry that this has become a monster PR that is impossible to review :rage3:

ashmaroli:
> that is impossible to review

Nothing is impossible..
Except for `Enable most cops that aren't disabled by GHP`, all changes are in consolidated commits.. easy to grok..

pathawks:
Two more cops that will take a bit of work to disable:
 * `Style/ModuleFunction`
 * `Layout/IndentHeredoc`

For `Layout/IndentHeredoc`, we [have a utility method](https://github.com/jekyll/jekyll/blob/3715633aaecd15d87a1d3336d3390f41cb417226/lib/jekyll/utils.rb#L22-L24) that will take care of the whitespace for us, we just aren't using it very often.

pathawks:
If and when we decide to merge this, at the very lease _please_ let me cleanup the history. It might be best just to squash everything, once we decide what to keep.

ashmaroli:
> *please* let me cleanup the history

@pathawks If I were you, I'd split these into multiple PRs.. a PR each for enabling a certain cop.. it would open up for easier review, push fixes and merge sooner..

pathawks:
The goal here is to get this done without having a dozen pull requests.

There shouldn't really be much here to discuss. As new cops have been added to Rubocop, we have just disabled them so we could upgrade to the latest Rubocop as quickly as possible. This is merely going back and enabling those new cops.

I don't think Jekyll is a big enough project to necessitate our own style guide, so I don't think any detailed discussions of "this vs that" would be productive.

DirtyF:
> I don't think Jekyll is a big enough project to necessitate our own style guide, so I don't think any detailed discussions of "this vs that" would be productive.

Do we follow an existing Style Guide ? Should we ?

pathawks:
> Do we follow an existing Style Guide ?

Our de facto style guide has kind of been Rubocop with some overrides from the GitHub Pages gem. I think just deferring to Rubocop would make things easiest for us to review.

envygeeks:
Deferring to Rubocop alone results in some ugly and confusing source. Jekyll-Assets was asked if we wanted to inherit from Jekyll but we override quite a bit to get clean source: https://github.com/jekyll/jekyll-assets/blob/master/.rubocop.yml#L5

## Let Me Elaborate

https://github.com/jekyll/jekyll-assets/blob/17ae16f00f1f49ded0b81285951907865eb70a02/.rubocop.yml#L3

This is arbitrary, arbitrary rules are annoying.

---

https://github.com/jekyll/jekyll-assets/blob/17ae16f00f1f49ded0b81285951907865eb70a02/.rubocop.yml#L4

Prevents

```ruby
def this
  if that
     return other
  end
end
```

Which is much more readable.

---

https://github.com/jekyll/jekyll-assets/blob/17ae16f00f1f49ded0b81285951907865eb70a02/.rubocop.yml#L5

```ruby
private
def this
  return that
end
```

Is much more explicit, here we put it above ***every*** method.  Not just at the first method, or the way rails does it (really not readable at all.) Explicit source is much better than implicit source, even Rails is better than the Rubocop default.

---

mattr-:
I'm also not a fan of the default Rubocop styles. Most of the other projects I've started or worked on have defaulted to the GitHub Ruby Styleguide.

Either way, I feel like a better approach would be to open a PR per Cop that we're thinking about enabling and then discuss each one in turn. A little bit more work on our part as maintainers but makes for a much cleaner and contained discussion, IMHO

PR #6478, opened 2017-10-24
Allow URL filters to work directly with documents

pathawks:
I may have proposed this before, but I'm curious if anybody else would be interested in URL filters that work directly with documents.

Instead of `{{ page.url | absolute_url }}` this would allow `{{ page | absolute_url }}` which feels just a little nicer to me.  Curious to hear other's thoughts.

/cc: @jekyll/build @jekyll/ecosystem

### DIFF
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -12,6 +12,7 @@ module Jekyll
       # Returns the absolute URL as a String.
       def absolute_url(input)
         return if input.nil?
+        input = input.url if input.respond_to?(:url)
         return input if Addressable::URI.parse(input.to_s).absolute?
         site = @context.registers[:site]
         return relative_url(input) if site.config["url"].nil?
@@ -28,6 +29,7 @@ module Jekyll
       # Returns a URL relative to the domain root as a String.
       def relative_url(input)
         return if input.nil?
+        input = input.url if input.respond_to?(:url)
         return input if Addressable::URI.parse(input.to_s).absolute?
 
         parts = [sanitized_baseurl, input]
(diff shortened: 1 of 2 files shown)


parkr:
What's the use-case for it? It seems fine, but it might confuse folks as being a bit magical. The primary requirement for the url filters is that they're chainable: input must allow URL, output must be URL. If this is just allowing an extra possible input, then I'm open to it.

ashmaroli:
> If this is just allowing an **extra possible input**, then I'm open to it.

:+1:  from me too, if that's the case..
`{{ page | absoulte_url }}` would then read as "*Extract the absolute-url of the given page and have it inserted here.*"

Understandably, the associated PR would need to cover:
- [ ] Implementation
- [ ] Documenting the additional behavior (*Extract the absolute-url of the given **`page` object** and have it inserted here.*)
- [ ] Having similar for `relative_url` too (?)
- [ ] Wide-range of tests to ensure nothing breaks

lib/jekyll/filters/url_filters.rb
```diff
@@ -12,6 +12,7 @@ module URLFilters
       # Returns the absolute URL as a String.
       def absolute_url(input)
         return if input.nil?
+        input = input.url if input.respond_to?(:url)
         return input if Addressable::URI.parse(input.to_s).absolute?
         site = @context.registers[:site]
         return relative_url(input) if site.config["url"].nil?
```
ashmaroli: in the case where `site.config["url"].nil?`, I think the `respond_to?(:url)` will be called again and therefore `input.url` will be called once again.. and consequently, a **re-assignment** of `input` unnecessarily

pathawks:
You’re right, this probably isn’t worth the added complexity.

parkr:
New users would find this useful (those who don't know you need to call `page.url`), so that's why I approved it.

ashmaroli:
@pathawks Please appease RuboCop here :smiley:

pathawks:
@ashmaroli ✅

DirtyF:
@jekyllbot: merge +minor

PR #6486, opened 2017-10-27
Remove `sudo` from macOS troubleshooting instructions

andrewda:
It should be made very clear that `sudo` is not recommended on macOS, so the examples should show the commands without `sudo`. If the user knows what they're doing, they can prepend `sudo` themselves as the note suggests.

/cc @jekyll/documentation

### DIFF
--- a/docs/_docs/troubleshooting.md
+++ b/docs/_docs/troubleshooting.md
@@ -74,7 +74,7 @@ apt update && apt install libffi-dev clang ruby-dev make
 On macOS, you may need to update RubyGems (using `sudo` only if necessary):
 
 ```sh
-sudo gem update --system
+gem update --system
 ```
 
 If you still have issues, you can download and install new Command Line
@@ -84,11 +84,11 @@ Tools (such as `gcc`) using the following command:
 xcode-select --install
 ```
 
-which may allow you to install native gems using this command (again using
+which may allow you to install native gems using this command (again, using
 `sudo` only if necessary):
 
 ```sh
-sudo gem install jekyll
+gem install jekyll
 ```
 
 Note that upgrading macOS does not automatically upgrade Xcode itself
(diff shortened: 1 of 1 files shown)


parkr:
Definitely agree with this.

For the record, we have gone back and forth with this one line a few times. Folks who can’t install without sudo insist it should be included. Then someone comes in later and insists sudo should never be used. I think High Sierra makes it moot on Mac but folks running system installations of Ruby may still need sudo.

Just raising awareness of the fact that this is not a “no brainer,” and that we have changed this line once a year for the last 4 years, back and forth.

DirtyF:
On macOS High Sierra, with `rbenv` I never have to use `sudo` for those commands:

```sh
$ ruby -v
ruby 2.4.2p198 (2017-09-14 revision 59899) [x86_64-darwin16]

$ gem update --system
Latest version currently installed. Aborting.

$ gem install jekyll
Successfully installed jekyll-3.6.2
1 gem installed
```

I'll happily add a section about installing Jekyll with `brew` and `rbenv` in the docs.

chrisfinazzo:
@DirtyF, I've since forgotten - this isn't global, correct? I'm so tied into the Bundler way of thinking that I would be concerned about doing that. The exception I suppose being if you have many sites to manage and need the `jekyll` commands available from anywhere.

@andrewda, The whole "Bundler as a requirement" seems to be tied up in this discussion as well. This specific issue was first raised in #3999 and later #4368. I am of the opinion that we really should be encouraging people to adopt it.

@parkr Do we have any sense about how common this behavior is? It seems like Apple has clued into the fact that they were behind on shipping newer Ruby versions as the default (my system, on 10.12.6 has 2.4.x), but I would be concerned about upgrades if the environment changes as part of an upgrade.

DirtyF:
>  I am of the opinion that we really should be encouraging people to adopt it.

[Bundle has been merged into standard Library in Ruby 2.5 preview 1](https://www.ruby-lang.org/en/news/2017/10/10/ruby-2-5-0-preview1-released/).

mattr-:
@jekyllbot: merge +docs

mattr-:
@DirtyF whoops! I just noticed this was assigned to you. I didn't see any reason to not merge it, so I just went ahead and did it. Sorry. 🙈

DirtyF:
@mattr- I only assigned myself documentation issues to prevent stale, thanks for steppin' in ❤️

Issue #6348, opened 2017-09-06
Support index.json in DirectoryIndex


adborden:
Hey there, I'm using Jekyll collections to generate JSON APIs. In order to have more flexible, prettier URLs, it would be nice if Jekyll supported `index.json` in the Webrick DirectoryIndex. This would allow you to have API URLs like `/api/agency/gsa/` instead of `/api/agency/gsa.json`. Currently, `jekyll serve` just shows the directory listing because it doesn't know to look for `index.json` in directories. I've got my production config setup this way, so I'm just looking for the development server to also have this behavior. I'm happy to submit a pull request if you're OK with this addition.

- [ ] I believe this to be a bug, not a question about using Jekyll.
- [ ] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [ ] I ran `jekyll doctor` to check my configuration
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [x] This is a feature request.

parkr:
We like to keep `jekyll serve` matching GitHub Pages, but I'd be open to making index page extensions configurable.

parkr:
Alternatively, you could use a separate server locally.

adborden:
I'd prefer if there was Jekyll solution. It's easier for collaborators to run when it's just `bundle install; bundle exec jekyll serve`.

If we added this as a config option, are you thinking something like a list of files in `_config.yml`?
```yml
directory_index:
- index.html
- index.htm
- index.xml
- index.json
```

Or like a boolean just for json?

parkr:
Yeah, a listing like that, or something we have for markdown extensions:

```yaml
directory_index_exts: "html,htm,xml,json"
```

More explicit is probably nicer here. I'd shy away from the boolean option. Which would you prefer here? Feel free to submit a PR.

parkr:
[Here's where we keep the list currently.](https://github.com/jekyll/jekyll/blob/ac3e6b384f80c4726f0eac19a6f1a73ee7376d62/lib/jekyll/commands/serve.rb#L88-L94)

adborden:
I went with your suggestion of following the markdown_ext pattern. I think it's good to be consistent. It also allows you to _reduce_ the list if you'd like.

PR #6487, opened 2017-10-27
Add configuration for DirectoryIndex setting

adborden:
Fixes https://github.com/jekyll/jekyll/issues/6348

### DIFF
--- a/docs/_docs/configuration.md
+++ b/docs/_docs/configuration.md
@@ -660,6 +660,7 @@ port:    4000
 host:    127.0.0.1
 baseurl: "" # does not include hostname
 show_dir_listing: false
+directory_index_ext: "htm,html,rhtml,cgi,xml"
 
 # Outputting
 permalink:     date
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -77,7 +77,7 @@ module Jekyll
 
         private
         def webrick_opts(opts)
-          directory_index = opts["directory_index_ext"].split(",").map do |e|
+          directory_index = (opts["directory_index_ext"] || "").split(",").map do |e|
             "index.#{e.downcase}"
           end
 
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -77,6 +77,10 @@ module Jekyll
 
         private
         def webrick_opts(opts)
+          directory_index = opts["directory_index_ext"].split(",").map do |e|
+            "index.#{e.downcase}"
+          end
+
           opts = {
             :JekyllOptions      => opts,
             :DoNotReverseLookup => true,
@@ -85,13 +89,7 @@ module Jekyll
             :StartCallback      => start_callback(opts["detach"]),
             :BindAddress        => opts["host"],
             :Port               => opts["port"],
-            :DirectoryIndex     => %W(
-              index.htm
-              index.html
-              index.rhtml
-              index.cgi
-              index.xml
-            ),
+            :DirectoryIndex     => directory_index,
           }
 
           opts[:DirectoryIndex] = [] if opts[:JekyllOptions]["show_dir_listing"]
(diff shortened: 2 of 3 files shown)


ashmaroli:
@adborden This is nice..!! But, I'd like to take this a bit further.
Currently, if you wanted to add `.json` to the list, you'd have to **list all the required extensions**, i.e. `"htm","html","rhtml","cgi","xml","json"` to restore original functionality.

So, I feel we can reduce the overhead by having the array in the config file **appended** to the default array already supported by Jekyll

DirtyF:
Should `json` extension be also added to the default options?

adborden:
Yes, you're right, but I think it's okay for users to enter the entire list, after all, they know what index files they're using in the project. Plus the default list is documented.

Consider the alternative though, what if you wanted to reduce the default list to only "html"? There wouldn't be a way to do it if the config only appended. What do you think?

ashmaroli:
> what if you wanted to reduce the default list to only "html"?

I don't see the benefit of it..
Either ways, having such an option could be a double-edged sword.. (just my 2cents..)

Also, I understand that you used a comma-separated-string in sync with existing configuration for Markdown extensions, but wouldn't it be better to pass in an array of extensions directly, instead of having Jekyll *split* it out of a csv?

pathawks:
> wouldn't it be better to pass in an array of extensions directly, instead of having Jekyll split it out of a csv?

👍

parkr:
@adborden I'm thinking about the user experience here, and I think being exclusive is troublesome.

Imagine a user wants to do what you're doing: add "index.json" to the list. They go to set `directory_index_exts: "json"` and run `jekyll serve`. Dang! Their site's broken and they have no idea why. They do some research online, maybe find this PR, and discover only there that it overrides the defaults.

I think that's a bad user experience because it doesn't feel expected. A web server should always serve at _least_ `index.html`, right? Isn't it a 20-year-old convention? I can see not wanting `.htm` or some of the others here, but we should at least show `index.html`. That seems like a requirement for calling something a web server.

If we did that, we'd take `Set.new(["html"]) | Set.new(site.config["directory_index_exts"])` instead of just the latter. We'd always guarantee `html` is in there. `index.cgi`? `index.xml`? Yeah, I don't think those are nearly as important as `index.html` here.

What do you think of always guaranteeing `index.html`? 

Regarding the configuration setting, we actually do funky things where we allow both strings and arrays, so `directory_index_exts: json,xml,erb` and `directory_index_exts: [json, xml, erb]` are identical (YAML reads the first as a String, the second as an Array of Strings). We could standardize the reading of those options so they work both ways. 🤷‍♂️ I think Arrays of Strings makes more sense because that's what the program wants to use, but I'm open to arguments about consistency (e.g. `markdown_exts` as a comma-separated string) and ease-of-use (e.g. users don't understand YAML so we should handle anything that looks like an Array, such as a comma-separated string).

mattr-:
If the goal is just to be able to serve `index.json` along with other the index pages we already serve, could we go with the simpler option of just adding `index.json` to Jekyll's WEBrick configuration and skip adding the option for configuring new ones? 

This would require work from the GitHub Pages crew to add the same support but from the strictly Jekyll point of view, this satisfies the request for the functionality and is much simpler to implement.

Just my two cents. :smile:

adborden:
If Github would be open to adding index.json to their configuration, I'm totally down for the simpler solution of just adding it to the WEBrick config.

Otherwise, I'm happy to force `html` to be included as per @parkr 's recommendation. I can also try adding dual support for arrays/strings.

Let me know if there's consensus and I'll push an update :)

parkr:
@adborden Let's add `index.json` to the list here and I'll work on getting it supported in Pages. If we come up with a really good reason it shouldn't be supported, then we can at least have it here in dev environments and for folks running custom server setups.

👍 on adding `index.json` to the default list in lieu of customization.

adborden:
Closing in favor of https://github.com/jekyll/jekyll/pull/6550

Thanks for the input, all!

Issue #6358, opened 2017-09-12
Remove /assets/javascript/anchor-js


laukstein:
Even when no JavaScript, no plugin or template is used on Jekyll project, `jekyll serve` creates **_site/assets/javascript/anchor-js** with bunch of irrelevant staff.

![grabilla g13256](https://user-images.githubusercontent.com/183966/30325522-bfeeb4a8-97cd-11e7-8146-ef7c24095a57.png)

few of the files
https://jekyllrb.com/assets/javascript/anchor-js/anchor.js
https://jekyllrb.com/assets/javascript/anchor-js/banner.js
https://services.github.com/assets/javascript/anchor-js/banner.js

How to avoid/remove this **anchor-js**? It seems all Jekyll projects are affected by this issue. I hope it doesn't contain harmful JavaScript files.

DirtyF:
[anchorjs](https://github.com/bryanbraun/anchorjs) is an open source project used on many website to create anchor links. It's not embedded in Jekyll by default, nor in minima the default theme.

It is used on Jekyll's website and on many other sites powered by Jekyll, but it's up to you to add it or not.

laukstein:
@DirtyF, you are wrong. Please re-open this task. This is Jekyll bug - every Jekyll installation has this package without any option to remove it. It is not in **gem**, even not **_config.yml** and every `jekyll serve` it injects it again and again under **_site/assets**. When Jekyll root has no **assets** directory, it is not expected to even generate **_site/assets/***. @DirtyF, do you understand what I mean? I suggest you to run clean `jekyll new` >  `jekyll serve` and you will see the bug.

ashmaroli:
@laukstein This is not a Jekyll bug, like DirtyF said. Its not included by the Jekyll gem or its default theme Minima.

If it gets injected into **_site/assets** it's only because of **your theme**

laukstein:
After testing empty Jekyll folder with only **_config.yml** and **Gemfile**, I find out that this issue is caused by Gemfile:

    gem "github-pages", group: :jekyll_plugins

No **_site/assets** when removed Gemfile or used in it `gem "github-pages"`.

But `gem "github-pages", group: :jekyll_plugins` is the recommended setup for GitHub Pages, see https://github.com/github/pages-gem#usage

@ashmaroli, is it GitHub bug?

nickmccurdy:
[...]

PR #6488, opened 2017-10-27
Add a note on `:jekyll_plugins` group in the docs

ashmaroli:
Ref: https://github.com/jekyll/jekyll/issues/6358#issuecomment-339859914

gem-group

/cc @jekyll/documentation

### DIFF
--- a/docs/_docs/plugins.md
+++ b/docs/_docs/plugins.md
@@ -58,21 +58,6 @@ You have 3 options for installing plugins:
 
 <div class="note info">
   <h5>
-    The <code>:jekyll_plugins</code> group in a <code>Gemfile</code>
-  </h5>
-  <p>
-    Jekyll treats this particular group of gems(s) with a special love. Any gem
-    included in this group is given a high priority and "loaded" way before
-    Jekyll starts processing the rest of your source directory.
-    <br/>
-    Because of the priority, a gem included here will be activated (even under
-    <code>--safe</code> mode), even if its not explicitly listed under the
-    <code>plugins:</code> key in your site's config file.
-  </p>
-</div>
-
-<div class="note info">
-  <h5>
     <code>_plugins</code>, <code>_config.yml</code> and <code>Gemfile</code>
     can be used simultaneously
   </h5>
@@ -82,6 +67,24 @@ You have 3 options for installing plugins:
     others.
   </p>
 </div>
+
+### The jekyll-plugins group
+
+Jekyll gives this particular group of gems in your `Gemfile`, a different
+treatment. Any gem included in this group is loaded before Jekyll starts
+processing the rest of your source directory.
+
+A gem included here will be activated even if its not explicitly listed under
+the `plugins:` key in your site's config file.
+
+<div class="note warning">
+  <p>
+    Gems included in the <code>:jekyll-plugins</code> group are activated
+    regardless of the <code>--safe</code> mode setting. Be careful of what
+    gems are included under this group!
+  </p>
+</div>
+
 
 In general, plugins you make will fall broadly into one of five categories:
 
--- a/docs/_docs/plugins.md
+++ b/docs/_docs/plugins.md
@@ -70,7 +70,7 @@ You have 3 options for installing plugins:
 
 ### The jekyll-plugins group
 
-Jekyll gives this particular group of gems in your `Gemfile`, a different
+Jekyll gives this particular group of gems in your `Gemfile` a different
 treatment. Any gem included in this group is loaded before Jekyll starts
 processing the rest of your source directory.
 
--- a/docs/_docs/plugins.md
+++ b/docs/_docs/plugins.md
@@ -58,6 +58,21 @@ You have 3 options for installing plugins:
 
 <div class="note info">
   <h5>
+    The <code>:jekyll_plugins</code> group in a <code>Gemfile</code>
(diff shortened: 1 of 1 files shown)


DirtyF:
Could we recommend this as the default ? I mean you just list your gems in :jekyll_plugins and you're done. IIUC only GitHub Pages users have to rely on the `plugins` key in the `_config.yml`

ashmaroli:
> should we recommend this as the default

Since the documentation already states that these three options are equally potent and intermixable, I don't think *one of those options* can be a default..

docs/_docs/plugins.md
```diff
@@ -58,6 +58,21 @@ You have 3 options for installing plugins:
 
 <div class="note info">
   <h5>
+    The <code>:jekyll_plugins</code> group in a <code>Gemfile</code>
+  </h5>
+  <p>
+    Jekyll treats this particular group of gems(s) with a special love. Any gem
```
mattr-:
```
Jekyll gives this particular group of gems different treatment.
```

instead of
```
Jekyll treats this particular group of gems(s) with a special love.
```

docs/_docs/plugins.md
```diff
@@ -58,6 +58,21 @@ You have 3 options for installing plugins:
...
+    included in this group is given a high priority and "loaded" way before
```
mattr-: ~given a high priority and~ 

remove the quotes around `loaded`.

Remove the word `way` from this line.

docs/_docs/plugins.md
```diff
@@ -58,6 +58,21 @@ You have 3 options for installing plugins:
...
+    Jekyll starts processing the rest of your source directory.
+    <br/>
+    Because of the priority, a gem included here will be activated (even under
```
mattr-: ~Because of the priority,~ 

Capitalize the a after removing the priority bit.

docs/_docs/plugins.md
```diff
@@ -58,6 +58,21 @@ You have 3 options for installing plugins:
...
+    <code>--safe</code> mode), even if its not explicitly listed under the
```
mattr-: This `--safe` note is important. It should get at least it's own sentence and perhaps even a warning box wrapper as well. Something like

```
Gems included in the `jekyll-plugins` group are activated regardless of the `--safe` mode setting. Be careful what gems are included here!
```

docs/_docs/plugins.md
```diff
@@ -58,6 +58,21 @@ You have 3 options for installing plugins:
 
 <div class="note info">
   <h5>
+    The <code>:jekyll_plugins</code> group in a <code>Gemfile</code>
```
mattr-: I think I'd prefer if this is not a note and just an added section to the documentation.

mattr-:
> As we require Bundler, should we recommend this as the default?

We should probably document the order of precedence each one has rather than recommending a default and just leave it at that.

docs/_docs/plugins.md
ashmaroli: or how about
```
Jekyll treats this particular group of gems(s) differently
```
?

ashmaroli:
> I have a few things I'd like to see addressed

...done.

docs/_docs/plugins.md
mattr-: That's fine too. Just need one `s` on `gem(s)` though. 😄

docs/_docs/plugins.md
```diff
@@ -83,6 +68,24 @@ You have 3 options for installing plugins:
   </p>
 </div>
 
+### The jekyll-plugins group
+
+Jekyll gives this particular group of gems in your `Gemfile`, a different
```
mattr-: no comma necessary.

ashmaroli:
Thanks Matt :smiley:

docs/_docs/plugins.md
```diff
@@ -68,6 +68,24 @@ You have 3 options for installing plugins:
   </p>
 </div>
 
+### The jekyll-plugins group
```
DirtyF: s/jekyll_plugins

DirtyF:
@jekyllbot: merge +docs

docs/_docs/plugins.md
```diff
@@ -68,6 +68,24 @@ You have 3 options for installing plugins:
...
+processing the rest of your source directory.
+
+A gem included here will be activated even if its not explicitly listed under
+the `plugins:` key in your site's config file.
+
+<div class="note warning">
+  <p>
+    Gems included in the <code>:jekyll-plugins</code> group are activated
+    regardless of the <code>--safe</code> mode setting. Be careful of what
+    gems are included under this group!
```
parkr: I'm a little weary of the FUD (fear, uncertainty, and doubt) that "Be careful!" leads to. Could we rephrase this to be less fear-mongery? "Be aware" perhaps?

DirtyF: oops didn't see you comment, but totally agree with you. Will fix this right away

DirtyF: fixed in https://github.com/jekyll/jekyll/commit/52c34060e3311c1d2a369f02b4d2319c1b283f7b

mattr-: I didn't even notice this. 🙈  Thanks for pointing it out @parkr.

PR #6490, opened 2017-10-27
filter relative_url should keep absolute urls with scheme/authority

straight-shoota:
Currently, `relative_url` converts all uris to an absolute path even absolute uris with authority.
This can lead to unexpected results. For example `{{ "https://example.com/" | relative_url }}` becomes `/https://example.com/` (with `baseurl = "/"`).

This is an issue when you want to define for example a navigation structure with internal and external links.
```liquid
{% for link in nav %}
  <a href="{{ link.url | relative_url }}">{{ link.label }}</a>
{% endfor %}
```
The `relative_url` converts all links to absolute paths, even external links. Leaving it out would mean that internal links don't get the baseurl prepended.

This PR adds a simple check if the input is an absolute url with authority and doesn't modify it (except normalization).

As a minor issue, `https://example.com/` *could* also be a valid relative path and `/https://example.com/` the expected result. But this is not typically desired and can easily be mitigated by escaping this path as `https%3A//example.com/` to avoid misinterpration.

### DIFF
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -20,13 +20,17 @@ module Jekyll
         ).normalize.to_s
       end
 
-      # Produces a URL relative to the domain root based on site.baseurl.
+      # Produces a URL relative to the domain root based on site.baseurl
+      # unless it is already an absolute url with an authority (host).
       #
       # input - the URL to make relative to the domain root
       #
       # Returns a URL relative to the domain root as a String.
       def relative_url(input)
         return if input.nil?
+        uri = Addressable::URI.parse(input.to_s)
+        return uri.normalize.to_s if uri.absolute?
+
         parts = [sanitized_baseurl, input]
         Addressable::URI.parse(
           parts.compact.map { |part| ensure_leading_slash(part.to_s) }.join
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -28,8 +28,7 @@ module Jekyll
       # Returns a URL relative to the domain root as a String.
       def relative_url(input)
         return if input.nil?
-        uri = Addressable::URI.parse(input.to_s)
-        return uri.normalize.to_s if uri.absolute?
+        return input if Addressable::URI.parse(input.to_s).absolute?
 
         parts = [sanitized_baseurl, input]
         Addressable::URI.parse(
--- a/test/test_filters.rb
+++ b/test/test_filters.rb
@@ -526,6 +526,21 @@ class TestFilters < JekyllUnitTest
         filter = make_filter_mock({ "baseurl" => Value.new(proc { "/baseurl/" }) })
         assert_equal "/baseurl#{page_url}", filter.relative_url(page_url)
       end
+
+      should "transform protocol-relative url" do
+        url = "//example.com/"
+        assert_equal "/base//example.com/", @filter.relative_url(url)
+      end
+
+      should "not modify an absolute url with scheme" do
+        url = "file:///file.html"
+        assert_equal url, @filter.relative_url(url)
+      end
+
+      should "normalize international URLs with scheme" do
+        url = "https://example.com/错误"
+        assert_equal "https://example.com/%E9%94%99%E8%AF%AF", @filter.relative_url(url)
+      end
     end
 
     context "strip_index filter" do
--- a/test/test_filters.rb
+++ b/test/test_filters.rb
@@ -537,9 +537,9 @@ class TestFilters < JekyllUnitTest
         assert_equal url, @filter.relative_url(url)
       end
 
-      should "normalize international URLs with scheme" do
+      should "not normalize absolute international URLs" do
         url = "https://example.com/错误"
-        assert_equal "https://example.com/%E9%94%99%E8%AF%AF", @filter.relative_url(url)
+        assert_equal "https://example.com/错误", @filter.relative_url(url)
       end
     end
 


ashmaroli:
> `/https://example.com/` the expected result.

When is that the expected result?

lib/jekyll/filters/url_filters.rb
```diff
@@ -20,13 +20,17 @@ def absolute_url(input)
...
-      # Produces a URL relative to the domain root based on site.baseurl.
+      # Produces a URL relative to the domain root based on site.baseurl
+      # unless it is already an absolute url with an authority (host).
       #
       # input - the URL to make relative to the domain root
       #
       # Returns a URL relative to the domain root as a String.
       def relative_url(input)
         return if input.nil?
+        uri = Addressable::URI.parse(input.to_s)
```
ashmaroli: we are now calling `Addressable::URI.parse()` twice.. consider refactoring to keep it DRY?

pathawks:
I’m in the fence about this.

On the one hand, it makes sense for the filter to have a pass through mechanism rather than returning something nonsensical.

On the other hand, even with these changes the filter returns garbage if called twice:

```
{{ page.url | relative_url | relative_url }}
```

There’s no way to prevent this from returning garbage.

What is the use case here? Shouldn’t the site know if a URL will be absolute already?

pathawks:
@ashmaroli Stating that `/http://example.com` is a perfectly valid URL and _could_ be the expected result.

straight-shoota:
@pathawks Yes, calling it twice makes rubbish, but that's also the status quo. That shouldn't matter too much I think, because this filter is typically used directly in a print expression, so no further modifications.

A usecase for this feature is a navigation list including internal and external URLs. The internal ones need to be prefixed with `baseurl` the external should be unmodified.
```yaml
nav:
- label: home
   url: /home
- label: jekyll
   url: https://jekyllrb.com
```

```liquid
{% for link in site.nav %}
  <a href="{{ link.url | relative_url }}">{{ link.label }}</a>
{% endfor %}
```

lib/jekyll/filters/url_filters.rb
straight-shoota: The second one parses the input with baseurl prepended and you need to know if the original input has a host before that. So this can't really be optimized.

pathawks:
Ok, I’m on board conceptually.

lib/jekyll/filters/url_filters.rb
straight-shoota: Only the `normalize.to_s` could be refactored, but I'm not sure this is better:

```ruby
uri = Addressable::URI.parse(input.to_s)
unless uri.host
  parts = [sanitized_baseurl, input]
  uri = Addressable::URI.parse(
    parts.compact.map { |part| ensure_leading_slash(part.to_s) }.join
  )
end
uri.normalize.to_s
```

It avoids duplication but there isn't much duplicated in the first place and I really like to return early for special cases.

pathawks: In the `absolute_url` filter, we have a one-line check to see if the URL is already absolute.

Let’s replace these two lines with that one.

straight-shoota: An absolute URI according to RFC 3986 includes a scheme. This is also what `Addressable::URI.absolute?` checks.
This would URIs like `//example.com/` slip through  - this is technically a URI with host `example.com` and empty path (`/`). 

It might be arguable if these should rather be interpreted as an (absolute) path beginning with two path separators.

But protocol-relative URIs are also valid and even with https-only becoming more and more standard they might still prove useful in some situations.

But you're certainly right that this should be handled similarly in both methods.

pathawks:
> But protocol-relative URIs are also valid and even with https-only becoming more and more standard they might still prove useful in some situations.

Protocol-relative URLs are an antipattern we do not support. The two URL filters should behave consistently.

straight-shoota:
@pathawks What do you think about normalizing absolute URIs? In both methods.

lib/jekyll/filters/url_filters.rb
```diff
@@ -20,13 +20,17 @@ def absolute_url(input)
         ).normalize.to_s
       end
 
-      # Produces a URL relative to the domain root based on site.baseurl.
+      # Produces a URL relative to the domain root based on site.baseurl
+      # unless it is already an absolute url with an authority (host).
```
ashmaroli: small correction here. **`authority != host`**
```ruby
def uri(input = "http://localhost:4000")
  Addressable::URI.parse(input.to_s)
end

print uri.absolute? # => true
print uri.host      # => "localhost"
print uri.authority # => "localhost:4000"
```

straight-shoota: It's not supposed to mean authority == host but host is the integral component of an authority. No host, no authority. It may also have userinfo and port, but they're not the key.

ashmaroli: okay then..

straight-shoota:
Does this need any more work?

lib/jekyll/filters/url_filters.rb
```diff
@@ -20,13 +20,17 @@ def absolute_url(input)
...
       #
       # input - the URL to make relative to the domain root
       #
       # Returns a URL relative to the domain root as a String.
       def relative_url(input)
         return if input.nil?
+        uri = Addressable::URI.parse(input.to_s)
+        return uri.normalize.to_s if uri.absolute?
```
pathawks: What does `normalize` actually do? Do we need that?

pathawks: By which I mean: can we instead `return input` instead?

straight-shoota: Oh, my bad. I thought `absolute_url` was using it for already absolute URIs. But in fact it normalizes only those generated by that method. So, `relative_url` should behave the same way.

pathawks:
@jekyllbot: merge +minor

PR #6494, opened 2017-10-28
Fix a few minor issues in the docs

j-f1:
- Correct a couple syntax highlighting declarations
- Change “our” to “their” in the docs for deploying with Aerobatic


mattr-:
Thanks for updating the syntax highlighting and correcting that text! 🎊 🥇 

@jekyllbot: merge +docs

PR #6499, opened 2017-10-29
Fix #6498: Use Gem to discover the location of bundler.

### LINKED ISSUE #6498: Jekyll should be using Gem bins, not system.
The system can wrap in on itself.  Jekyll should be using Ruby to find Ruby stuff.

envygeeks:


mattr-:
@envygeeks it looks like this needs a rebase but is good to go after that.

DirtyF:
Isn't the error on Appveyor related?

mattr-:
> Isn't the error on Appveyor related?

Looks that way. I'm going to hold off on this then until that's fixed. I know Windows isn't officially supported but I don't see a reason to merge something that we know is going to regress that platform.

ashmaroli:
just confirmed.. I was **unable to run `script/default-site`** with this branch on my Windows machine..

ashmaroli:
@envygeeks The issue here is that Windows doesn't parse the shebang `#!/usr/bin/env ruby` like Unix does.. so to run a ruby executable on Windows, the program creates two helper files in the `bin/` directory in the `%PATH%` (in this case, `RUBY_VERSION/bin/bundler.bat` and `RUBY_VERSION/bin/bundler`)

But in the context of this PR, you can achieve the same by passing `"ruby"` to the utility function
```diff
- process, output = Jekyll::Utils::Exec.run(exe, "install")
+ process, output = Jekyll::Utils::Exec.run("ruby", exe, "install")
```

envygeeks:
I'll get it fixed.

mattr-:
@jekyllbot: merge +dev

PR #6500, opened 2017-10-30
.sass-cache doesn't *always* land in options['source']

envygeeks:
https://github.com/sass/sass/blob/f31605339b4f4dd3098b010186db481b7563a634/lib/sass/plugin/configuration.rb#L18

### DIFF
--- a/lib/jekyll/commands/clean.rb
+++ b/lib/jekyll/commands/clean.rb
@@ -22,7 +22,7 @@ module Jekyll
           options = configuration_from_options(options)
           destination = options["destination"]
           metadata_file = File.join(options["source"], ".jekyll-metadata")
-          sass_cache = File.join(options["source"], ".sass-cache")
+          sass_cache = ".sass-cache"
 
           remove(destination, :checker_func => :directory?)
           remove(metadata_file, :checker_func => :file?)


parkr:
I think it's a bug that we don't write into the site source. I think that bug should be fixed in jekyll-sass-converter: https://github.com/jekyll/jekyll-sass-converter/pull/64.

envygeeks:
TBQF, I would consider what you want a bug, my source is my source, if I change my source to say "source" and it's not the current directory, I would be pretty mad that source is being written to by Jekyll instead of the current directory, because that source directory was created to reduce the amount of clutter that Jekyll likes to create.

This is also part of a larger long-standing issue where Jekyll hasn't created a caching directory that can be customized, and is available to everyone, including plugins.

pathawks:
> This is also part of a larger long-standing issue where Jekyll hasn't created a caching directory that can be customized

Absolutely, this is ultimately the path forward. For now though, I think it's best that we keep the SASS cache inside the source directory, as that is where we write all other cache files (`.jekyll-metadata`) and where I as a user would expect the cache to be.

Also, it seems like the SASS generator should be responsible for cleaning its own cache. Maybe we need to add a hook to make that possible?

envygeeks:
> we keep the SASS cache inside the source directory

The SASS cache directory *isn't* and never does land within the source directory so changing that behavior now breaks well established behavior.  Behavior I was chastised for just recently.  It lands within the current working directory and has since inception, it was just never noticed because people rarely adjust the source directory (apparently?)

As far as leaving the cache to SASS, I don't know that it should be that way.  They don't necessarily have supporting functions to do that (I could be wrong, I've not browsed the source all that closely)  I think Jekyll should still handle it since it's the one that wants the cache (but that's not true either technically since it's enabled by default.)  I've stated my opinion but really I'm neutral on leaving this part of the topic.

pathawks:
Maybe we should focus instead on a centralized cache that is available to all plugins, which Jekyll can destroy during `jekyll clean`

mattr-:
> Maybe we should focus instead on a centralized cache that is available to all plugins, which Jekyll can destroy during `jekyll clean`

I like this idea.

ashmaroli:
> never does land within the source directory so changing that behavior now breaks well established behavior.

Can we safely assume that no one (users and plugin developers alike) interacts with a `.sass-cache` manually or programmatically and hence its location can be *silently* switched to the intended place at `source_dir`? Or is that going overboard?

parkr:
We probably cannot safely assume that, so let’s accept this PR and move forward with a cache folder. Is there a proposal for that?

ashmaroli:
> so let’s accept this PR and move forward

:+1:

DirtyF:
@jekyllbot: merge +minor

PR #6501, opened 2017-10-30
Remove unnecessary variables, and messages.

envygeeks:


lib/jekyll/commands/clean.rb
```diff
@@ -20,17 +20,13 @@ def init_with_program(prog)
...
-
-          remove(destination, :checker_func => :directory?)
-          remove(metadata_file, :checker_func => :file?)
-          remove(sass_cache, :checker_func => :directory?)
+          remove(File.join(options["source"], ".sass-cache"))
+          remove(File.join(options["source"], ".jekyll-metadata"))
+          remove(options["destination"])
         end
 
-        def remove(filename, checker_func: :file?)
```
ashmaroli: *Mon ami*, is changing the signature of a public method in a minor release cycle, good practice?
Removing that parameter is going to break all those code that has passed a second argument to the `Jekyll::Commands::Clean.remove()` method..

This is best earmarked for `4.0`

envygeeks: If Jekyll cared about that they wouldn't have upgraded to Liquid 4, and removed support for Liquid 3, in a breaking way... in the middle of the 3.x cycle.  Your argument has no standing.

ashmaroli: > Your argument has no standing.

But can you deny that this is a breaking change, either ways?

envygeeks: Yes, I can deny it. Should I? Probably.  It's a command, specifically for Jekyll... its internal, if you are tapping into it, even though it's defacto private... since it's specifically for Jekyll and internal, it's not considered a semantically breaking change. Internal stuff does not fall under that category.  Only public facing API's that are meant to be publicly used.  Like Hooks.

ashmaroli: Lets meet half-way here and just print a deprecation message when the `arity > 1` and proceed using your refinements.. does that make sense?

lib/jekyll/commands/clean.rb
```diff
@@ -20,17 +20,13 @@ def init_with_program(prog)
...
-          if File.public_send(checker_func, filename)
+        def remove(filename)
+          if File.exists?(filename)
```
pup: please use `File.exist?` here instead!

envygeeks:
We've changed this in our fork, I don't need it to be public.

On the `exist?` vs `exists?` bit, you don't lock to 2.3+ so that change is arbitrary and not relevant.  You still support EOL Ruby's that even enterprises don't use anymore, and argue against `exists?` like it's a big deal.

pup:
@envygeeks sure, but it makes our rubocop fail

PR #6502, opened 2017-10-30
Allow plugins to modify the obsolete files.

envygeeks:


DirtyF:
What is the use case here?

envygeeks:
@DirtyF https://github.com/jekyll/jekyll-assets/blob/master/lib/jekyll/assets/patches/obsolete_files.rb should be unnecessary, people shouldn't have to hack Jekyll to prevent clobbering if their files at-large.

DirtyF:
@envygeeks got you 👌

DirtyF:
@jekyllbot: merge +minor

DirtyF:
Thanks @envygeeks. 👍

parkr:
This should be documented on the plugins page.

PR #6506, opened 2017-10-31
Ignore future dated documents in hidden collection

### LINKED ISSUE #5953: Collection documents with future date and output set to false are not ignored

ashmaroli:
Resolves #5953 

It may not be the most elegant approach, but it resolves the issue while preserving other existing functionality..

/cc @jekyll/build

### DIFF
--- a/lib/jekyll/collection.rb
+++ b/lib/jekyll/collection.rb
@@ -208,7 +208,7 @@ module Jekyll
       doc = Jekyll::Document.new(full_path, :site => site, :collection => self)
       doc.read
       if site.publisher.publish?(doc) || !write?
-        docs << doc
+        docs << doc unless site.publisher.hidden_in_the_future?(doc)
       else
         Jekyll.logger.debug "Skipped From Publishing:", doc.relative_path
       end
--- a/features/collections.feature
+++ b/features/collections.feature
@@ -116,7 +116,30 @@ Feature: Collections
     And the _site directory should exist
     And the "_site/puppies/fido.html" file should exist
 
-  Scenario: Hidden collection with future dated document
+  Scenario: Access rendered collection with future dated document via Liquid
+    Given I have a _puppies directory
+    And I have the following documents under the puppies collection:
+      | title  | date       | content             |
+      | Rover  | 2007-12-31 | content for Rover.  |
+      | Fido   | 2120-12-31 | content for Fido.   |
+    And I have a "_config.yml" file with content:
+    """
+    collections:
+      puppies:
+        output: true
+    """
+    And I have a "index.html" page that contains "Newest puppy: {% assign puppy = site.puppies.last %}{{ puppy.title }}"
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should not see "Newest puppy: Fido" in "_site/index.html"
+    But I should see "Newest puppy: Rover" in "_site/index.html"
+    When I run jekyll build --future
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "Newest puppy: Fido" in "_site/index.html"
+
+  Scenario: Unrendered collection with future dated document
     Given I have a _puppies directory
     And I have the following documents under the puppies collection:
       | title  | date       | content             |
@@ -139,7 +162,7 @@ Feature: Collections
     And the _site directory should exist
     And the "_site/puppies/fido.html" file should not exist
 
-  Scenario: Access hidden collection with future dated document via Liquid
+  Scenario: Access unrendered collection with future dated document via Liquid
     Given I have a _puppies directory
     And I have the following documents under the puppies collection:
       | title  | date       | content             |
--- a/features/collections.feature
+++ b/features/collections.feature
@@ -116,7 +116,7 @@ Feature: Collections
     And the _site directory should exist
     And the "_site/puppies/fido.html" file should exist
 
-  Scenario: Hidden collection with document with future date
+  Scenario: Hidden collection with future dated document
     Given I have a _puppies directory
     And I have the following documents under the puppies collection:
       | title  | date       | content             |
@@ -138,6 +138,29 @@ Feature: Collections
     Then I should get a zero exit status
     And the _site directory should exist
     And the "_site/puppies/fido.html" file should not exist
+
+  Scenario: Access hidden collection with future dated document via Liquid
+    Given I have a _puppies directory
+    And I have the following documents under the puppies collection:
+      | title  | date       | content             |
+      | Rover  | 2007-12-31 | content for Rover.  |
+      | Fido   | 2120-12-31 | content for Fido.   |
+    And I have a "_config.yml" file with content:
+    """
+    collections:
+      puppies:
+        output: false
+    """
+    And I have a "index.html" page that contains "Newest puppy: {% assign puppy = site.puppies.last %}{{ puppy.title }}"
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should not see "Newest puppy: Fido" in "_site/index.html"
+    But I should see "Newest puppy: Rover" in "_site/index.html"
+    When I run jekyll build --future
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "Newest puppy: Fido" in "_site/index.html"
 
   Scenario: All the documents
     Given I have an "index.html" page that contains "All documents: {% for doc in site.documents %}{{ doc.relative_path }} {% endfor %}"


ashmaroli:
As of `v3.6.2`, documents in an unrendered collection (`output: false`) can be iterated through in Liquid (including future dated documents). Based on how you perceive that situation, this can be considered as either an `enhancement` or a `bug-fix`

pathawks:
So, if I have a collection of upcoming events, and I am displaying them in a sidebar on my site, this enhancement will break that?

ashmaroli:
> So, if I have a collection of upcoming events, and I am displaying them in a sidebar on my site, this enhancement will break that?

When you say `I am displaying` that means you probably built/served your site with `--future` which is then fine as demonstrated by the cucumber test..

pathawks:
Iterating through them in a sidebar does not necessitate each item being output to its own page.

I agree that the expexted behavior is currently ambiguous and needs to be fixed, but we should probably call this a major change so users can be aware that the behavior has fundamentally changed.

ashmaroli:
> So, if I have a collection of upcoming events, and I am displaying them in a sidebar on my site, this enhancement will break that?

I just tested.. this is indeed breaking that behavior.. :disappointed:

pathawks:
I think we might consider making `future` available as a per-collection option.

Perhaps I don’t want my site built with future blog posts, but _do_ want future events.

I think the current `future` option was added when posts were the only dated content. That is no longer the case.

ashmaroli:
> but we should probably call this a major change so users can be aware that the behavior has fundamentally changed.

But according to the scenario you described, people would probably **love** being able to do that.. I think the sane way out here is to apply a `wont-fix` label to the associated issue and document things to remove ambiguity..

ashmaroli:
Though its frustrating that behaviors are different for future-dated docs in a Rendered collection (`output; true`) vs Unrendered collection (`output: false`)

ashmaroli:
### Behavioral pattern for dated documents in a Rendered Collection vs Unrendered Collection

#### Default Jekyll build

<table>
<tr>
<th></th>
<th>Configuration</th>
<th colspan="2">Physical files at URL</th>
<th colspan="2">Access via liquid</th>
</tr>
<tr>
<td></td>
<td></td>
<td>Future dated</td>
<td>Non-future dated</td>
<td>Future dated</td>
<td>Non-future dated</td>
</tr>
<tr>
<td>Rendered Collection</td>
<td><code>output: true</code></td>
<td>:x:</td>
<td>:white_check_mark:</td>
<td>:x:</td>
<td>:white_check_mark:</td>
</tr>
<tr>
<td>Unrendered Collection</td>
<td><code>output: false</code></td>
<td>:x:</td>
<td>:x:</td>
<td>:white_check_mark:</td>
<td>:white_check_mark:</td>
</tr>
</table>

---

#### Jekyll Build with `--future`

<table>
<tr>
<th></th>
<th>Configuration</th>
<th colspan="2">Physical files at URL</th>
<th colspan="2">Access via liquid</th>
</tr>
<tr>
<td></td>
<td></td>
<td>Future dated</td>
<td>Non-future dated</td>
<td>Future dated</td>
<td>Non-future dated</td>
</tr>
<tr>
<td>Rendered Collection</td>
<td><code>output: true</code></td>
<td>:white_check_mark:</td>
<td>:white_check_mark:</td>
<td>:white_check_mark:</td>
<td>:white_check_mark:</td>
</tr>
<tr>
<td>Unrendered Collection</td>
<td><code>output: false</code></td>
<td>:x:</td>
<td>:x:</td>
<td>:white_check_mark:</td>
<td>:white_check_mark:</td>
</tr>
</table>

ashmaroli:
With this PR, the above table becomes..:


#### Default Jekyll build

<table>
<tr>
<th></th>
<th>Configuration</th>
<th colspan="2">Physical files at URL</th>
<th colspan="2">Access via liquid</th>
</tr>
<tr>
<td></td>
<td></td>
<td>Future dated</td>
<td>Non-future dated</td>
<td>Future dated</td>
<td>Non-future dated</td>
</tr>
<tr>
<td>Rendered Collection</td>
<td><code>output: true</code></td>
<td>:x:</td>
<td>:white_check_mark:</td>
<td>:x:</td>
<td>:white_check_mark:</td>
</tr>
<tr>
<td>Unrendered Collection</td>
<td><code>output: false</code></td>
<td>:x:</td>
<td>:x:</td>
<td>:x:</td>
<td>:white_check_mark:</td>
</tr>
</table>

---

#### Jekyll Build with `--future`

<table>
<tr>
<th></th>
<th>Configuration</th>
<th colspan="2">Physical files at URL</th>
<th colspan="2">Access via liquid</th>
</tr>
<tr>
<td></td>
<td></td>
<td>Future dated</td>
<td>Non-future dated</td>
<td>Future dated</td>
<td>Non-future dated</td>
</tr>
<tr>
<td>Rendered Collection</td>
<td><code>output: true</code></td>
<td>:white_check_mark:</td>
<td>:white_check_mark:</td>
<td>:white_check_mark:</td>
<td>:white_check_mark:</td>
</tr>
<tr>
<td>Unrendered Collection</td>
<td><code>output: false</code></td>
<td>:x:</td>
<td>:x:</td>
<td>:white_check_mark:</td>
<td>:white_check_mark:</td>
</tr>
</table>

lib/jekyll/collection.rb
```diff
@@ -208,7 +208,7 @@ def read_document(full_path)
       doc = Jekyll::Document.new(full_path, :site => site, :collection => self)
       doc.read
       if site.publisher.publish?(doc) || !write?
```
parkr: Why doesn't `publish?` check `hidden_in_the_future?`?

pathawks:
@ashmaroli What are your thoughts regarding making `future` a per-collection option?

lib/jekyll/collection.rb
ashmaroli: > Why doesn't `publish?` check `hidden_in_the_future?` ?

@parkr, `:publish?` does `check hidden_in_the_future?` but it gets overridden by the `|| !write?`
`publisher == false` **but** `!write? == true` and therefore, the expression returns `true`

I initially thought about removing the `|| !write?` but changed it for some reason I do not remember now..

ashmaroli:
> What are your thoughts regarding making `future` a per-collection option?

Isn't that too a `breaking-change`? Sites already built with `--future` or `future: true` suddenly won't be able to access future dated docs in their collection(s)..

mattr-:
If it's a breaking change, it's a breaking change and we can do the Ruby things and this thing and release Jekyll 4.0. :smiley:

The topic of Jekyll 4.0 has come up a lot recently. My position is that if there's a breaking change that we want to make, we shouldn't be shy about making it and bumping the major version number. The `github-pages` gem, for example, is up to v167. The bumping of the version number communicates intent with the release, per SemVer, and IMHO, that's more valuable than dancing around releasing certain changes because they'd require a major version bump. The only caveat is that we need to be **very** clear about any potential breakages we'd cause as part of our release notes and our messaging around the major version release. I don't see that as being a large problem though.

ashmaroli:
> we can do the Ruby things and this thing and release Jekyll 4.0.

Parker had mentioned long ago that he doesn't plan on shipping a 4.0 in the near future.. And if it changes now, we need a proper plan.. a roadmap.. to ensure things are well laid-out, tested and documented..
For that we have an open `4.0 milestone`

PR #6509, opened 2017-11-01
Add latin mode to slugify

alextsui05:
This is in response to #4623 

To summarize the discussion in the above issue, trying to slugify strings with Swedish letters in them (accented characters in general) produced unexpected output. `ascii` mode does the job but drops accented characters, which is probably undesirable to some users. So we add `latin` mode, which will transliterate accented characters to the ASCII equivalent.

### Example: 
```
{{ 'kvalité, då, äta, öl' | slugify: 'ascii' }} # -> kvalit-d-ta-l
{{ 'kvalité, då, äta, öl' | slugify: 'latin' }} # -> kvalite-da-ata-ol
```

### Changes

* Add a new `latin` mode to be used as an option for `#slugify`
* Add runtime dependency on `i18n` gem to perform the transliteration
* Update the documentation to demonstrate `latin` mode, and also `ascii` mode, which existed but was not documented

### Documentation screenshots

vmplayer_2017-11-02_18-04-38

vmplayer_2017-11-02_18-05-00

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,8 @@ gem "rake", "~> 12.0"
 
 gem "rouge", ENV["ROUGE"] if ENV["ROUGE"]
 
-gem 'activesupport', '~> 4.2'
+# Dependency of jekyll-mentions. RubyGems in Ruby 2.1 doesn't shield us from this.
+gem 'activesupport', '~> 4.2', :groups => [:test_legacy, :site] if RUBY_VERSION < "2.2.2"
 
 group :development do
   gem "launchy", "~> 2.3"
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ gem "rake", "~> 12.0"
 gem "rouge", ENV["ROUGE"] if ENV["ROUGE"]
 
 # Dependency of jekyll-mentions. RubyGems in Ruby 2.1 doesn't shield us from this.
-gem 'activesupport', '~> 4.2', :groups => [:test_legacy, :site] if RUBY_VERSION < "2.2.2"
+gem "activesupport", "~> 4.2", :groups => [:test_legacy, :site] if RUBY_VERSION < "2.2.2"
 
 group :development do
   gem "launchy", "~> 2.3"
--- a/Gemfile
+++ b/Gemfile
@@ -7,8 +7,7 @@ gem "rake", "~> 12.0"
 
 gem "rouge", ENV["ROUGE"] if ENV["ROUGE"]
 
-# Dependency of jekyll-mentions. RubyGems in Ruby 2.1 doesn't shield us from this.
-gem "activesupport", "~> 4.2", :groups => [:test_legacy, :site] if RUBY_VERSION < "2.2.2"
+gem 'activesupport', '~> 4.2'
 
 group :development do
   gem "launchy", "~> 2.3"
--- a/lib/jekyll/utils.rb
+++ b/lib/jekyll/utils.rb
@@ -332,13 +332,11 @@ module Jekyll
     # See Utils#slugify for a description of the character sequence specified
     # by each mode.
     private
-    def replace_character_sequence_with_hyphen(string, mode:)
-      re =
+    def replace_character_sequence_with_hyphen(string, mode: "default")
+      replaceable_char =
         case mode
         when "raw"
           SLUGIFY_RAW_REGEXP
-        when "default"
-          SLUGIFY_DEFAULT_REGEXP
         when "pretty"
           # "._~!$&'()+,;=@" is human readable (not URI-escaped) in URL
           # and is allowed in both extN and NTFS.
@@ -348,12 +346,12 @@ module Jekyll
           # method is to ditch anything else but latin letters and numeric
           # digits.
           SLUGIFY_ASCII_REGEXP
-        when "latin"
+        else
           SLUGIFY_DEFAULT_REGEXP
         end
 
       # Strip according to the mode
-      string.gsub(re, "-")
+      string.gsub(replaceable_char, "-")
     end
   end
 end
--- a/lib/jekyll/utils.rb
+++ b/lib/jekyll/utils.rb
@@ -186,10 +186,10 @@ module Jekyll
     #   # => "The-_config.yml file"
     #
     #   slugify("The _config.yml file", "ascii")
-    #   # => "the-config.yml-file"
+    #   # => "the-config-yml-file"
     #
     #   slugify("The _config.yml file", "latin")
-    #   # => "the-config.yml-file"
+    #   # => "the-config-yml-file"
     #
     # Returns the slugified string.
     def slugify(string, mode: nil, cased: false)
--- a/lib/jekyll/utils.rb
+++ b/lib/jekyll/utils.rb
@@ -188,7 +188,7 @@ module Jekyll
     #   slugify("The _config.yml file", "ascii")
     #   # => "the-config.yml-file"
     #
-    #   slugify("The _cönfig.yml file", "latin")
+    #   slugify("The _config.yml file", "latin")
     #   # => "the-config.yml-file"
     #
     # Returns the slugified string.
@@ -203,7 +203,7 @@ module Jekyll
       # Drop accent marks from latin characters. Everything else turns to ?
       string = ::I18n.transliterate(string) if mode == "latin"
 
-      slug = replace_character_sequence_with_hyphen(string, mode: mode)
+      slug = replace_character_sequence_with_hyphen(string, :mode => mode)
 
       # Remove leading/trailing hyphen
       slug.gsub!(%r!^\-|\-$!i, "")
--- a/lib/jekyll/utils.rb
+++ b/lib/jekyll/utils.rb
@@ -1,4 +1,3 @@
-
 # frozen_string_literal: true
 
 module Jekyll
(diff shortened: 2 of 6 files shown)


ashmaroli:
Hello @alextsui05 from a first-pass, it looks like you're calling `I18n::transliterate` instead of `ActiveSupport::Inflector::transliterate`. Do correct me if I'm wrong, but I like using `I18n` directly so that we need not pull in the entire ActiveSupport library but can instead depend on `gem "i18n"` directly.

That said, from the failing CI logs, it looks like the intended results are being delivered, but for the maintainers to accept this, they'd like to have the tests passing and have this change documented because suddenly users of the `:ascii` mode are going to find their slugs different.. ergo, a broken-link and the 404 fiasco..

pathawks:
> because suddenly users of the :ascii mode are going to find their slugs different.. ergo, a broken-link and the 404 fiasco..

Which makes this a major change rather than an enhancement, correct?

ashmaroli:
> Which makes this a major change rather than an enhancement, correct?

As of now, yes, but if the author proposes a new mode altogether, then it can still be an `enhancement`..

ashmaroli:
@alextsui05 You'll need to make a couple of changes:
- Jekyll uses `double-quotes` *everywhere*
- You may want to consider introducing a *new `slugify` mode* so that existing `:ascii` mode slugs do not break. (I'm thinking maybe, `:ascii_plus`, is nice, but its open to your creativity..)

Gemfile
```diff
@@ -8,7 +8,7 @@ gem "rake", "~> 12.0"
 gem "rouge", ENV["ROUGE"] if ENV["ROUGE"]
 
 # Dependency of jekyll-mentions. RubyGems in Ruby 2.1 doesn't shield us from this.
-gem "activesupport", "~> 4.2", :groups => [:test_legacy, :site] if RUBY_VERSION < "2.2.2"
+gem 'activesupport', '~> 4.2', :groups => [:test_legacy, :site] if RUBY_VERSION < "2.2.2"
```
DirtyF: Our Rubocop config is set to double quotes

alextsui05:
I realized `active_support` is a heavy requirement, so I took @ashmaroli 's suggestion and updated the code to use `i18n` (+ updated the test).

The more I think about it, maybe it would be better for compatibility + clearer to define a new mode that is more explicit about what it does -- maybe `slugify: 'latin'` gives a good hint to the user.

```
I18n.transliterate('á, à, ç, é, è, í, ï, ó, ò, ú, ü, n·h, s·h.')
 => "a, a, c, e, e, i, i, o, o, u, u, n?h, s?h."
```

I'll make another pass at the code at the end of the day to address the style issues too. Thanks for the help.

alextsui05:
I updated the PR to make it clear that we are adding a new `latin` mode rather than modifying the `ascii` mode. The CI tests are still winding down but I believe style issues are all ironed out.

lib/jekyll/utils.rb
```diff
@@ -336,5 +326,34 @@ def duplicate_frozen_values(target)
         target[key] = val.dup if val.frozen? && duplicable?(val)
       end
     end
+
+    # Replace each character sequence with a hyphen.
+    #
+    # See Utils#slugify for a description of the character sequence specified
+    # by each mode.
+    private
+    def replace_character_sequence_with_hyphen(string, mode:)
```
ashmaroli: not a big fan of the long name `:replace_character_sequence_with_hyphen` but its a `private` method and the name expresses the method's intention.. so its okay I guess..

But, how about making the `:default` mode the *default* parameter?
```ruby
private
def replace_character_sequence_with_hyphen(string, mode: "default")
  replaceable_char =
    case mode
    when "raw"
      SLUGIFY_RAW_REGEXP
    when "pretty"
      # "._~!$&'()+,;=@" is human readable (not URI-escaped) in URL
      # and is allowed in both extN and NTFS.
      SLUGIFY_PRETTY_REGEXP
    when "ascii"
      # For web servers not being able to handle Unicode, the safe
      # method is to ditch anything else but latin letters and numeric
      # digits.
      SLUGIFY_ASCII_REGEXP
    else
      SLUGIFY_DEFAULT_REGEXP
    end

  string.gsub(replaceable_char, "-")
end
```
(and effectively reduce the total no.of `case` branches`)

alextsui05:
I took @ashmaroli 's advice and simplified the switch statement.

Thanks all for the reviews and help!

ashmaroli:
@alextsui05 Thank you very much! This is already an awesome `First-Contribution` to Jekyll! :tada:
Congratulations in advance! :smiley: 

Wanna take it to a higher level?
If yes, you could document this new mode by editing [this section](https://github.com/jekyll/jekyll/blob/master/docs/_docs/templates.md#options-for-the-slugify-filter) and table right above

---

On a side-note, if you've the time, you can **open a new PR** documenting the `ascii` mode

lib/jekyll/utils.rb
```diff
@@ -185,6 +188,9 @@ def has_yaml_header?(file)
     #   slugify("The _config.yml file", "ascii")
     #   # => "the-config.yml-file"
     #
+    #   slugify("The _config.yml file", "latin")
```
mattr-: I think it would be better to see the example here demonstrate the difference between the ascii mode and the latin mode. This will make it clearer as to why the latin mode exists.

lib/jekyll/utils.rb
```diff
@@ -194,26 +200,10 @@ def slugify(string, mode: nil, cased: false)
...
-          # method is to ditch anything else but latin letters and numeric
-          # digits.
-          SLUGIFY_ASCII_REGEXP
-        end
+      # Drop accent marks from latin characters. Everything else turns to ?
+      string = ::I18n.transliterate(string) if mode == "latin"
 
-      # Strip according to the mode
-      slug = string.gsub(re, "-")
+      slug = replace_character_sequence_with_hyphen(string, :mode => mode)
```
mattr-: Using the symbol syntax would be better here, so that you're consistent with the rest of the codebase.
```ruby
mode: mode
```

mattr-:
Related to my above review comment, I would prefer that the documentation work for the `latin` filter be done as part of this PR as well. The documentation for the `ascii` mode isn't required since it's an existing part of the codebase.

lib/jekyll/utils.rb
alextsui05: @mattr- Yeah, I thought so, too. To really demonstrate the difference, the input string would need an accented character, but when I tried that, one of the builds failed with a note about non-ASCII characters disallowed in comments. I would point out the failed build but I'm not sure how to go through the build history for this PR in TravisCI.

lib/jekyll/utils.rb
alextsui05: @mattr- Another style error in the build came up here as well, something about using hash rocket syntax, so I wrote it this way.

alextsui05:
@mattr- Thanks, I'll update the documentation in the `docs/` subdirectory after the work day is over today. Also, I responded to the review comments about what I ran into, so if you could advise me on what to do there, I'll also take care of those changes.

mattr-:
@alextsui05 If making those changes would cause the build to fail, then let's not do those right now. I think that's on us as maintainers to fix those. :smiley:

Looking forward to seeing the new documentation and getting this merged soon! Thanks!
bitmoji

ashmaroli:
I think its best that you make only changes related to the `latin` mode in this PR.. everything else in a separate PR..

alextsui05:
@mattr- I just pushed a documentation update that includes a description of the new `latin` mode. While I was at it, I also added a description for `ascii` mode that @ashmaroli mentioned -- it was existing but was not actually in the documentation.

I posted screenshots on the main body of the PR for the sections that changed. Once that's been finalized, I'll also update the table in the link @ashmaroli posted in an earlier comment.

ashmaroli:
> I'll also update the table in the link

@alextsui05 Don't worry about it. The "table" has [already been updated](https://github.com/alextsui05/jekyll/blob/f057c14ace22f64a08fb14498cc11cb2fbc6a8ec/docs/_docs/templates.md) with your latest commit..

ashmaroli:
> it was existing but was not actually in the documentation.

again, that should be handled in a separate PR..

alextsui05:
Okay, let me split out the `ascii` part of the documentation into a separate commit and put it in a new PR.

mattr-:
>Okay, let me split out the ascii part of the documentation into a separate commit and put it in a new PR.

Please don't. That's just churn for no good reason.

ashmaroli:
:smile: @mattr-  I apologize if i were too picky..
since you made the `merge` manually, you'll have to **manually make an entry in `History.markdown` now..

ashmaroli:
> That's just churn for no good reason.

```
¯\_(ツ)_/¯
```
:laughing:

alextsui05:
> On a side-note, if you've the time, you can open a new PR documenting the ascii mode

@ashmaroli Sorry, I remembered the part about documenting ascii mode but forgot the part about in a new PR (this gets me at work, too). I'll keep it in mind for the next PR!

mattr-:
I see no reason for the appveyor build to fail since all I did was merge in master so that we can merge this, so away we go!

@jekyllbot: merge +minor

@alextsui05 Thank you again for making the changes we've requested and congratulations on your first contribution to Jekyll!
bitmoji

PR #5063, opened 2016-07-07
Extract tag name to class method

ayastreb:
@envygeeks , @parkr is this what you meant in #4878 ?

Issue #4878, opened 2016-05-09
Some tags are not friendly to inheritance.


envygeeks:
https://github.com/jekyll/jekyll/blob/df2a92cb39418e0eb457d34b1c53fe70bbda975d/lib/jekyll/tags/link.rb#L4

This should be a method that takes the class name and turns it into a tag name so that if a plugin author wishes to sub-class and adjust something cleanly they can with little to know care about superficial stuff like that.

parkr:
Ideally it'd be even simpler: `self.class.name.downcase`. Subclassing a tag shouldn't get you the same tag name as the superclass.

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least
one month.

The resources of the Jekyll team are limited, and so we are asking for your help.

If you can still reproduce this error on the <code>3.1-stable</code> or <code>master</code> branch,
please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider building it first as a plugin. Jekyll 3 introduced
[hooks](http://jekyllrb.com/docs/plugins/#hooks) which provide convenient access points throughout
the Jekyll build pipeline whereby most needs can be fulfilled. If this is something that cannot be
built as a plugin, then please provide more information about why in order to keep this issue open.

Thank you for all your contributions.

ayastreb:
Can this issue be closed?

parkr:
👍 Thanks, @ayastreb!

PR #6512, opened 2017-11-02
Refactor `Tags::Link`

ashmaroli:
`Jekyll::Tags::Link` looked like it could use some tidying up..
Just internal changes. No change in API

### DIFF
--- a/lib/jekyll/tags/link.rb
+++ b/lib/jekyll/tags/link.rb
@@ -21,8 +21,9 @@ module Jekyll
         site = context.registers[:site]
 
         site.each_site_file do |item|
-          rel_path = item.relative_path
-          return item.url if valid_paths.include?(rel_path)
+          # return `item.url` if `item.relative_path` equals either
+          # @relative path or @relative_path with a leading slash.
+          return item.url if valid_paths.include?(item.relative_path)
         end
 
         raise ArgumentError, <<-MSG
--- a/lib/jekyll/tags/link.rb
+++ b/lib/jekyll/tags/link.rb
@@ -3,6 +3,7 @@
 module Jekyll
   module Tags
     class Link < Liquid::Tag
+      # TODO: Remove singleton class in v4.0
       class << self
         def tag_name
           self.name.split("::").last.downcase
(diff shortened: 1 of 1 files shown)


lib/jekyll/tags/link.rb
```diff
@@ -34,4 +35,4 @@ def render(context)
   end
 end
 
-Liquid::Template.register_tag(Jekyll::Tags::Link.tag_name, Jekyll::Tags::Link)
```
ashmaroli: this is a method call just to output a simple string.. and that method actually chains 4 other methods to `self` (another call)..
```ruby
def tag_name
  self.name.split("::").last.downcase
end
```

lib/jekyll/tags/link.rb
```diff
@@ -11,21 +12,21 @@ def tag_name
...
         site.each_site_file do |item|
-          return item.url if item.relative_path == @relative_path
-          # This takes care of the case for static files that have a leading /
-          return item.url if item.relative_path == "/#{@relative_path}"
+          rel_path = item.relative_path
+          return item.url if valid_paths.include?(rel_path)
         end
 
         raise ArgumentError, <<-MSG
-Could not find document '#{@relative_path}' in tag '#{self.class.tag_name}'.
```
ashmaroli: multiple method calls to output a simple string (`link`)

lib/jekyll/tags/link.rb
```diff
@@ -11,21 +12,21 @@ def tag_name
...
 
       def render(context)
+        # additionally handle static files, that have a leading `/`
+        valid_paths = [@relative_path, "/#{@relative_path}"]
         site = context.registers[:site]
 
         site.each_site_file do |item|
-          return item.url if item.relative_path == @relative_path
-          # This takes care of the case for static files that have a leading /
-          return item.url if item.relative_path == "/#{@relative_path}"
```
ashmaroli: `item.relative_path` is going to be called twice for static files

ashmaroli:
> but I don't see any added value here.

@DirtyF Left a few PR comments. There may not be much to gain with this, but in theory, the reduced no.of method calls should benefit performance

lib/jekyll/tags/link.rb
```diff
@@ -11,21 +12,21 @@ def tag_name
...
+          rel_path = item.relative_path
+          return item.url if valid_paths.include?(rel_path)
```
pathawks: The ambiguity here worries me. This risks false positives.

ashmaroli: please elaborate.

pathawks: Sorry, I misread the code. Your code is correct...
...but more difficult to read. Why create the variable `rel_path`?

ashmaroli: Valid question. I had created `rel_path` in an earlier draft the line was
```ruby
return item.url if rel_path == @relative_path || rel_path == "/#{@relative_path}"
```
The above was flagged by rubocop which led to the current code..

I'll edit to make it simply
```ruby
return item.url if valid_paths.include?(item.relative_path)
```

DirtyF:
> the reduced no.of method calls should benefit performance

Can we benchmark this?

pathawks:
> There may not be much to gain with this, but in theory, the reduced no.of method calls should benefit performance

At the expense of inheritance

#4878 #5063

DirtyF:
Based on previous discussions on that topic pointed by @pathawks, I feel like we can close this.

PR #6513, opened 2017-11-03
Remove unnecessary encoding comment

ashmaroli:
Rubocop did not catch this superfluous magic comment because it was on the second line which was not preceded by a line with a comment starting with a shebang

### DIFF
--- a/lib/jekyll/converters/markdown/kramdown_parser.rb
+++ b/lib/jekyll/converters/markdown/kramdown_parser.rb
@@ -1,5 +1,4 @@
 # Frozen-string-literal: true
-# Encoding: utf-8
 
 module Jekyll
   module Converters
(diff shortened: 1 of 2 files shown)


DirtyF:
@ashmaroli There might be a last encoding comment to delete:

```sh
ag "Encoding: utf-8"
lib/jekyll/utils/ansi.rb
3:# Encoding: utf-8
```

ashmaroli:
> There might be a last encoding comment to delete:

Right. That was not detected by Rubocop either.. :smirk: 
Thanks for catching that.. will push a commit shortly..

DirtyF:
@jekyllbot: merge +dev

pathawks:
What’s with the red X? ❌

pup:
@pathawks somehow got fixed in 42c21ab

DirtyF:
@pathawks fixed in https://github.com/jekyll/jekyll/commit/94dc9265cbdc0677bd2bba49afe24359373de66a

PR #5828, opened 2017-01-28
Rescue from Psych::SyntaxError instead of SyntaxError after parsing YAML

jmhooper:
`SafeYAML` was raising a [`Psych::SynaxError`](https://github.com/ruby/psych/blob/master/lib/psych/syntax_error.rb), but the `Convertible` module and `Document` class were rescuing from a `SyntaxError`. This made the conditional branch for reporting that an error was a result of problems parsing the YAML unreachable and led to some more cryptic error messages.

```shell
Error reading file /Users/jonathanhooper/Developer/jekyll-test-site/pages/error.html: (<unknown>): did not find expected key while parsing a block mapping at line 2 column 1
```

versus

```shell
YAML Exception reading /Users/jonathanhooper/Developer/jekyll-test-site/pages/error.html: (<unknown>): did not find expected key while parsing a block mapping at line 2 column 1
```

This commit tells the `Document` class and `Convertible` module to rescue from the correct error class.

PR #6520, opened 2017-11-05
Assert raising Psych::SyntaxError when `"strict_front_matter"=>true`

### LINKED ISSUE #6515: JRuby build is failing
All PRs are currently failing CI because the JRuby build is failing 😭😭😭

ashmaroli:
Resolves #6515 (as well..)

### DIFF
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -266,7 +266,7 @@ module Jekyll
           merge_defaults
           read_content(opts)
           read_post_data
-        rescue StandardError => e
+        rescue SyntaxError, StandardError, Errors::FatalException => e
           handle_read_error(e)
         end
       end
--- a/test/test_convertible.rb
+++ b/test/test_convertible.rb
@@ -38,7 +38,7 @@ class TestConvertible < JekyllUnitTest
     should "raise for broken front matter with `strict_front_matter` set" do
       name = "broken_front_matter2.erb"
       @convertible.site.config["strict_front_matter"] = true
-      assert_raises do
+      assert_raises(Psych::SyntaxError) do
         @convertible.read_yaml(@base, name)
       end
     end
--- a/test/test_site.rb
+++ b/test/test_site.rb
@@ -308,7 +308,7 @@ class TestSite < JekyllUnitTest
           "collections"         => ["broken"],
           "strict_front_matter" => true
         ))
-        assert_raises do
+        assert_raises(Psych::SyntaxError) do
           site.process
         end
       end


lib/jekyll/document.rb
```diff
@@ -266,7 +266,7 @@ def read(opts = {})
           merge_defaults
           read_content(opts)
           read_post_data
-        rescue StandardError => e
+        rescue SyntaxError, StandardError, Errors::FatalException => e
```
pathawks: Can you explain what's going on here? Why are we rescuing `FatalException`?

ashmaroli: Because `:handle_read_error` is:
```ruby
    private
    def handle_read_error(error)
      if error.is_a? SyntaxError
        Jekyll.logger.error "Error:", "YAML Exception reading #{path}: #{error.message}"
      else
        Jekyll.logger.error "Error:", "could not read file #{path}: #{error.message}"
      end

      if site.config["strict_front_matter"] || error.is_a?(Jekyll::Errors::FatalException)
        raise error
      end
    end
```
My understanding is that unless `Errors::FatalException` is rescued here, the variable `e` is not going be a subclass of our `Errors::FatalException`

What say, @parkr ?

pathawks: That doesn't make any sense. We rescue `FatalException` only so that we can rethrow it?

pathawks: Oh, because we need to print the error message about not reading the file. Gotcha 👌

ashmaroli: Right.. and `Errors::FatalException` is only a subclass of Ruby's `RuntimeError` (which is the default for a bare `raise`), and not a true `fatal` which is documented as `"impossible to rescue"`

ashmaroli: @pathawks  this change is not actually required for this PR.. do you want me to remove this change so the tests can patched faster?

pathawks:
@jekyllbot: merge +dev

parkr:
This changes “production” code so I’d call it a bug fix if anything. I’m concerned as to why the ::SyntaxError is caught (a syntax error in ruby comes to mind) instead of a ::Psych::SyntaxError but I trust y’all :)

ashmaroli:
> I’m concerned as to why the ::SyntaxError is caught (a syntax error in ruby comes to mind) instead of a ::Psych::SyntaxError

You're rightly concerned. There was [already a PR](https://github.com/jekyll/jekyll/pull/5828) proposing to `rescue` from `::Psych::SyntaxError`. I thought since he brought the matter to light first, he should get a fair chance to contribute.

lib/jekyll/document.rb
jmhooper: Since `SyntaxError` and `Jekyll::Errors::FatalException` are subclasses of `StandardError`, rescuing from `StandardError` should collect them as well.

For example, in this code:

```ruby
class ASDFError < StandardError; end

def raise_error
  raise ASDFError
end

begin
  raise_error
rescue StandardError => e
  if e.is_a? ASDFError
    puts 'ASDF Error occured'
  else
    puts "Another error occured"
  end
end
```

...rescuing from `StandardError` still collects the `ASDFError`, and logs that an `ASDFError` occurs.

PR #6522, opened 2017-11-05
Log kramdown warnings if log level is WARN

### LINKED ISSUE #6516: Show kramdown's warning messages

kramdown generates warning messages during the parsing and conversion phases and these warning are available in the `converter.warnings` array. [Currently](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/converters/markdown/kramdown_parser.rb#L40) these warnings are just ignored. However, it might be useful to users to allow them to enable them, e.g. by a configuration option.

For example, kramdown warns when it finds duplicate reference links being declared or when a reference-style inline link refers to a missing link ID.

You can use [webgen's implementation](https://github.com/gettalong/webgen/blob/master/lib/webgen/content_processor/kramdown.rb#L58) as reference on how this integration might be done.

Crunch09:
This fixes #6516 . Thank you @gettalong for making us aware of this option! 👍 

cc: @jekyll/build

### DIFF
--- a/lib/jekyll/converters/smartypants.rb
+++ b/lib/jekyll/converters/smartypants.rb
@@ -29,7 +29,12 @@ module Jekyll
       end
 
       def convert(content)
-        Kramdown::Document.new(content, @config).to_html.chomp
+        document = Kramdown::Document.new(content, @config)
+        html_output = document.to_html.chomp
+        document.warnings.each do |warning|
+          Jekyll.logger.warn "Kramdown warning:", warning
+        end
+        html_output
       end
     end
   end
--- a/lib/jekyll/converters/markdown/kramdown_parser.rb
+++ b/lib/jekyll/converters/markdown/kramdown_parser.rb
@@ -37,7 +37,12 @@ module Jekyll
         end
 
         def convert(content)
-          Kramdown::Document.new(content, @config).to_html
+          document = Kramdown::Document.new(content, @config)
+          html_output = document.to_html
+          document.warnings.each do |warning|
+            Jekyll.logger.warn "Kramdown warning:", warning
+          end
+          html_output
         end
 
         private
--- a/test/test_kramdown.rb
+++ b/test/test_kramdown.rb
@@ -56,6 +56,12 @@ class TestKramdown < JekyllUnitTest
       assert_equal "<h1>Some Header</h1>", @markdown.convert("# Some Header #").strip
     end
 
+    should "should log kramdown warnings" do
+      allow_any_instance_of(Kramdown::Document).to receive(:warnings).and_return(["foo"])
+      expect(Jekyll.logger).to receive(:warn).with("Kramdown warning:", "foo")
+      @markdown.convert("Something")
+    end
+
     context "when asked to convert smart quotes" do
       should "convert" do
         assert_match(


lib/jekyll/converters/smartypants.rb
```diff
@@ -29,7 +29,12 @@ def output_ext(_)
       end
 
       def convert(content)
-        Kramdown::Document.new(content, @config).to_html.chomp
+        document = Kramdown::Document.new(content, @config)
+        html_output = document.to_html.chomp
```
Crunch09: Is there any reason why we're doing `chomp` here but not in `KramdownParser` above?

Crunch09:
> Why not use Jekyll.logger.warn here?

Uhm, good point 😆 There's no real reason for choosing `debug`

lib/jekyll/converters/smartypants.rb
pathawks: My thinking was that SmartyPants only deals with inline elements (and not block elements), so white space is insignificant.

ashmaroli:
@Crunch09, in #6516 Gettalong mentioned activating the warning channel with a configuration option.. IMO, that is a nice idea, because, then the user can be issued `Jekyll.logger.warn` messages which will be output without all the clutter from the `--verbose` option..

Crunch09:
@ashmaroli @DirtyF I agree with you, changed to `#warn` now.

ashmaroli:
> changed to `#warn` now.

and the configuration option..? Is that for a future PR?

DirtyF:
> For example, kramdown warns when it finds duplicate reference links being declared or when a reference-style inline link refers to a missing link ID.

These are useful warnings. They should be displayed by default.

People can still  use the `--quiet` option if they don't want any output.

ashmaroli:
> These are useful warnings. They should be displayed by default.

Fair enough. :+1:

pathawks:
It would also be great to see if this could be done for other converters.

@jekyllbot: merge +minor

DirtyF:
Here are the errors on Jekyll's website, the path of the file is missing:

```
 Kramdown warning: No link definition for link ID '#1339' found on line 2
  Kramdown warning: No link definition for link ID '' found on line 2
  Kramdown warning: No link definition for link ID '#1338' found on line 3
  Kramdown warning: No link definition for link ID '' found on line 3
  Kramdown warning: No link definition for link ID 'docs' found on line 150
  Kramdown warning: No link definition for link ID 'rubocop' found on line 260
  Kramdown warning: No link definition for link ID '"permalink"' found on line 299
  Kramdown warning: No link definition for link ID 'site' found on line 375
  Kramdown warning: No link definition for link ID 'docs' found on line 437
  Kramdown warning: No link definition for link ID 'docs' found on line 580
  Kramdown warning: No link definition for link ID 'docs' found on line 583
  Kramdown warning: No link definition for link ID 'docs' found on line 592
  Kramdown warning: No link definition for link ID 'add note' found on line 926
  Kramdown warning: No link definition for link ID '' found on line 1002
  Kramdown warning: Warning: this should not occur - no block parser handled the line
  Kramdown warning: Warning: this should not occur - no block parser handled the line
  Kramdown warning: Warning: this should not occur - no block parser handled the line
  Kramdown warning: Warning: this should not occur - no block parser handled the line
  Kramdown warning: Warning: this should not occur - no block parser handled the line
  Kramdown warning: Warning: this should not occur - no block parser handled the line
  Kramdown warning: Warning: this should not occur - no block parser handled the line
  Kramdown warning: Warning: this should not occur - no block parser handled the line
  Kramdown warning: Warning: this should not occur - no block parser handled the line
  Kramdown warning: Warning: this should not occur - no block parser handled the line
  Kramdown warning: Warning: this should not occur - no block parser handled the line
```

pathawks:
@DirtyF Could you open a new issue, that we should add the file path to these warnings? This feature will be super helpful!

lib/jekyll/converters/smartypants.rb
```diff
@@ -29,7 +29,12 @@ def output_ext(_)
...
+        document.warnings.each do |warning|
+          Jekyll.logger.warn "Kramdown warning:", warning
```
pathawks: It looks like there are currently no (legitimate) warnings raised by any of the parsers SmartyPants is using in kramdown. Perhaps we just rip this out?

Crunch09: Apologies for the late response. No objections to remove it, if it doesn't help in any way

PR #6526, opened 2017-11-06
[RuboCop] Enable `Style/UnneededCapitalW` cop

ashmaroli:
```
Do not use %W unless interpolation is needed. If not, use %w.
```


DirtyF:
@jekyllbot: merge +dev

PR #6529, opened 2017-11-07
Refactor method to reduce ABC Metric size

ashmaroli:
Assignment Branch Condition Size for the method on repo branches (as reported by RuboCop) :
  - `master` : **29.7**
  - `PR-branch`: ~~20.1~~ **13.64**

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -206,23 +206,20 @@ module Jekyll
 
         private
         def enable_ssl(opts)
-          jekyll_options = opts[:JekyllOptions]
-          ssl_cert       = jekyll_options["ssl_cert"]
-          ssl_key        = jekyll_options["ssl_key"]
-          return if !ssl_cert && !ssl_key
+          cert, key, src =
+            opts[:JekyllOptions].values_at("ssl_cert", "ssl_key", "source")
 
-          # rubocop:disable Style/RedundantException
-          raise RuntimeError, "--ssl-cert or --ssl-key missing." if !ssl_cert || !ssl_key
+          return if cert.nil? && key.nil?
+          unless cert && key
+            # rubocop:disable Style/RedundantException
+            raise RuntimeError, "Missing --ssl_cert or --ssl_key. Both are required."
+          end
 
           require "openssl"
           require "webrick/https"
 
-          source_key = Jekyll.sanitized_path(jekyll_options["source"], ssl_key)
-          source_certificate = Jekyll.sanitized_path(jekyll_options["source"], ssl_cert)
-
-          opts[:SSLCertificate] =
-            OpenSSL::X509::Certificate.new(File.read(source_certificate))
-          opts[:SSLPrivateKey ] = OpenSSL::PKey::RSA.new(File.read(source_key))
+          opts[:SSLCertificate] = OpenSSL::X509::Certificate.new(read_file(src, cert))
+          opts[:SSLPrivateKey ] = OpenSSL::PKey::RSA.new(read_file(src, key))
           opts[:SSLEnable] = true
         end
 
@@ -239,6 +236,11 @@ module Jekyll
         def mime_types
           file = File.expand_path("../mime.types", __dir__)
           WEBrick::HTTPUtils.load_mime_types(file)
+        end
+
+        private
+        def read_file(source_dir, file_path)
+          File.read(Jekyll.sanitized_path(source_dir, file_path))
         end
       end
     end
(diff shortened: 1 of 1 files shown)


lib/jekyll/commands/serve.rb
```diff
@@ -205,27 +205,28 @@ def enable_logging(opts)
...
         def enable_ssl(opts)
-          return if !opts[:JekyllOptions]["ssl_cert"] && !opts[:JekyllOptions]["ssl_key"]
-          if !opts[:JekyllOptions]["ssl_cert"] || !opts[:JekyllOptions]["ssl_key"]
-            # rubocop:disable Style/RedundantException
-            raise RuntimeError, "--ssl-cert or --ssl-key missing."
-          end
+          jekyll_options = opts[:JekyllOptions]
+          ssl_cert       = jekyll_options["ssl_cert"]
+          ssl_key        = jekyll_options["ssl_key"]
+          return if !ssl_cert && !ssl_key
```
pathawks:
```
return unless ssl_cert || ssl_key
```

ashmaroli: the original line defined it along similar lines..
```ruby
return if !opts[:JekyllOptions]["ssl_cert"] && !opts[:JekyllOptions]["ssl_key"]
```
I just replaced with local variables..

pathawks: As long as it’s being refactored, it should be refactored to be more clear :+1:

envygeeks: You didn't make it more clear, you changed it's meaning.  It should be

```ruby
return unless ssl_cert && ssl_key
```

envygeeks: If you want to clean it up this is what it ***should*** be

```ruby
return unless ssl_cert || ssl_key
ssl_key, ssl_cert = opts[:JekyllOptions].values_at("ssl_cert", "ssl_key")
raise RuntimeError, "missing something" unless ssl_cert && ssl_key
```

ashmaroli: @pathawks local testing showed me that the two are functionally different even though they appear same..

envygeeks:
```ruby
def enable_ssl(opts)
  return unless ssl_cert || ssl_key
  key, cert = opts[:JekyllOptions].values_at("ssl_cert", "ssl_key")
  raise RuntimeError, "missing something" unless cert && key

  src = Pathutil.new(opts[:JekyllOptions]["source"])
  opts[:SSLPrivateKey ] = OpenSSL::PKey::RSA.new(src.join(key).cleanpath.read)
  opts[:SSLCertificate] = OpenSSL::X509::Certificate.new(src.join(cert).cleanpath.read)
  opts[:SSLEnable] = true
end
```

ashmaroli: ### sample file:
```ruby
def iflog(input="foo", alpha=false, beta=false)
  return if !alpha && !beta
  print "INPUT: "
  puts input
end

def unlog(input="foo", alpha=false, beta=false)
  return unless alpha && beta
  print "INPUT: "
  puts input
end
```
- load *sample* in `irb`
- exec following couplets:
  ```sh
  iflog # => nil
  unlog # => nil

  iflog("bar", true) # => INPUT: "bar"\n nil
  unlog("bar", true) # => nil

  iflog("bar", false, true) # => INPUT: "bar"\n nil
  unlog("bar", false, true) # => nil

  iflog("bar", true, true) # => INPUT: "bar"\n nil
  unlog("bar", true, true) # => INPUT: "bar"\n nil
  ```

pathawks: I don’t get it.
By De Morgan’s Law, if we want to negate the statement, we have to negate the whole thing:

```
if !a && !b
if !!(!a && !b)
unless !(!a && !b)
unless !!a || !!b
unless a || b
```

envygeeks: Guys, the first one checks to see if neither exists and then leaves, the second one checks to see if either one doesn't exist and then throws.  The first says if both don't exist then cool, lets bail out, the second says that if both don't exist then this is garbage.

So the first needs to be `unless a || b` and the second can be `if !a || !b` or `unless a && b`

envygeeks: Writing it as so should make life much easier and avoid this debate later.

```
return if a.nil? && b.nil?
raise RuntimeError, "need both" unless a && b
```

envygeeks: Who would have thought `nil?` would solve a semantics problem. :man_shrugging:

DirtyF:
@jekyllbot: merge +dev

ashmaroli:
Thanks for your inputs on this one @envygeeks

PR #6537, opened 2017-11-10
Request for Jekyll version on GitHub issue tickets

ashmaroli:
It helps having the Jekyll version included by the user by default, instead of maintainers having to deduce it from the ticket body or via a future explicit question.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -76,3 +76,11 @@
   The minimum should be personal information. Though we normally don't log
   anything like that so there should be no need to alter it.
 -->
+
+## Jekyll version
+
+<!--
+  Insert your gem version.
+  Run `bundle exec jekyll -v` or `jekyll -v` to output the Jekyll version your
+  site is using.
+-->


pathawks:
Doesn’t one of the checkboxes ask users to use the latest version?

ashmaroli:
> Doesn’t one of the checkboxes ask users to use the latest version?

Its a checkbox. The answers are always `yes` or `no`
What if the user installed the latest version (via `gem install`..) but `bundle exec jekyll -v` is still something else? Surely, their `bundle exec jekyll build` is still referring to the older version..

ashmaroli:
> I wish we could add a command like “jekyll bug” that filed this on behalf of the user

But it would fail if `Utils::Internet.connected?` returns `false` at that moment of time

parkr:
@ashmaroli Yes, it would fail! But it could print a template to copy and paste if the user wishes. Just dreaming... no need to add here.

DirtyF:
Can we simply add this to the first checklist?

...
- [ ] If not, what Jekyll version are you using:

ashmaroli:
> Can we simply add this to the first checklist?

Not in favor of the `if not,` clause (based on my earlier comment)
> What if the user installed the latest version (via `gem install`..) but `bundle exec jekyll -v` is still something else? Surely, their `bundle exec jekyll build` is still referring to the older version..

pathawks:
I have already seen a number of users who either ignore or outright delete the whole issue template. Let's try to reduce complexity rather than adding to it.

A `jekyll bug` command might be a step in the right direction.

PR #6546, opened 2017-11-12
Lock to cucumber 3.0.1 on Ruby 2.1

ashmaroli:
Cucumber 3.0.2 has dropped support for Ruby 2.1

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,7 +23,7 @@ end
 
 group :test do
   gem "codeclimate-test-reporter", "~> 1.0.5"
-  gem "cucumber", "~> 3.0"
+  gem "cucumber", RUBY_VERSION >= "2.2" ? "~> 3.0" : "3.0.1"
   gem "jekyll_test_plugin"
   gem "jekyll_test_plugin_malicious"
   # nokogiri v1.8 does not work with ruby 2.1 and below


DirtyF:
I"m all with @mattr and @pathawks we shouldn't patch to support a EOL Ruby version. Let's bump Jekyll to a major version and remove EOL Ruby dependencies.

ashmaroli:
By the end of March 2018, even 2.2 will go EOL.. so I thinks its best we put off `4.0` till dependencies start to drop support for Ruby 2.2 as well.. or latest by May 2018..

ashmaroli:
Cucumber is just a dev dependency.. much like `nokogiri`
https://github.com/jekyll/jekyll/blob/9c7ef8f14ecab04611517651936ad0d9fab65c6f/Gemfile#L30

DirtyF:
> Cucumber is just a dev dependency.. much like nokogiri

So why do we need to lock the dev dependencies to Ruby < 2.2 exactly? (pardon my ignorance)

ashmaroli:
> So why do we need to lock the dev dependencies to Ruby < 2.2 exactly? (pardon my ignorance)

so that our CI continues as usual, till the library code drops support for EOL ruby versions..

pathawks:
> By the end of March 2018, even 2.2 will go EOL

Let's just say that 4.0 only supports >= 2.3, no matter when we release it. Problem solved.

DirtyF:
@jekyllbot: merge +dev

PR #6551, opened 2017-11-13
Suggest re-running command with --trace on fail

### LINKED ISSUE #6359: Suggest re-running command with --trace in limited error message
If the user did not use `--trace` and receives an error message, we should suggest that they re-run the command using that trace switch.

ashmaroli:
Closes #6359 

```
                    ------------------------------------------------
      Jekyll 3.6.0   Please append `--trace` to the `build` command
                     for any additional information or backtrace.
                    ------------------------------------------------
```


lib/jekyll/commands/serve.rb
```diff
@@ -94,7 +90,7 @@ def start(opts)
           if Jekyll.env == "development"
             config["url"] = default_url(config)
           end
-          [Build, Serve].each { |klass| klass.process(config) }
+          process_with_graceful_fail("serve", config, Build, Serve)
```
ashmaroli: The first argument expected here is not a `String` but the `cmd` object from `prog.command(:serve) do |cmd|`
This is the reason I added a `friction` label to this PR..

pathawks: I see. So, we need to pass `cmd` from `init_with_program()` to `start()`?

ashmaroli: Either that or do away with `start` entirely.. i.e. merge `start` into `init_with..`

pathawks: I imagine it was added to reduce perceived complexity with regard to Rubocop. `start()` should be a private method, and I don't see any problem passing `cmd` to it.

ashmaroli: Yeah, we have to decide between _changing method signature_ vs _doing away with unnecessary method_

pathawks: Oh, I'm not worried about backwards compatibility on this one. I can't imagine anybody else making use of `start()`

If you think it makes more sense to inline it, and Rubocop doesn't complain, I'm :+1::+1:

ashmaroli:
@pathawks Do you think I should move the `Do away with the 'start' method` to a separate PR to keep the diff here simpler?

pathawks:
Nah, we’re just sorting out merge conflicts. Let’s keep it all together 👍🏼

ashmaroli:
@oe would like you to re-review as this branch has changed considerably since your last review..
Thanks.

lib/jekyll/command.rb
```diff
@@ -73,6 +73,29 @@ def add_build_options(cmd)
...
+
+      # Run ::process method in a given set of Jekyll::Command subclasses and suggest
+      # re-running the associated command with --trace switch to obtain any additional
+      # information or backtrace regarding the encountered Exception.
+      #
+      # cmd     - the Jekyll::Command to be handled
+      # options - configuration overrides
+      # klass   - an array of Jekyll::Command subclasses associated with the command
+      #
+      # rubocop: disable RescueException
```
pathawks: I wonder if we want to add a comment explaining that we definitely _do_ want to `rescue Exception` here.

I'm just worried that a year from now I might be trying to clean up the code for a new version of Rubocop and forget that this is intentional and not just a bandage.

DirtyF:
@jekyllbot: merge +minor

PR #6552, opened 2017-11-13
Update Jekyll::VERSION to the latest released version

ashmaroli:
I wonder how this remained unchanged. Fortunately, the last released gem has the correct version specified..

### DIFF
--- a/lib/jekyll/version.rb
+++ b/lib/jekyll/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module Jekyll
-  VERSION = "3.6.0".freeze
+  VERSION = "3.6.2".freeze
 end


pathawks:
I think we usually increment this when we release.

The reason this hasn’t changed is because `3.6.2` was released from the `3.6-stable` branch.

parkr:
We definitely want to make sure 3.6.1 and 3.6.2 are present in `master`'s `History.markdown`.

pathawks:
@parkr That much has been taken care of :+1:

https://github.com/jekyll/jekyll/blob/525efbc2f89c4972e5defda74b268222c08fa1f4/History.markdown#L86-L111

parkr:
This itself doesn't need a note in History.markdown so I manually merged.

PR #6553, opened 2017-11-13
Include document path in kramdown warnings

ashmaroli:
Agreed its not the most elegant solution, but it gets the job done..
I'm staying away from `SmartyPants` in the first-pass.

### DIFF
--- a/lib/jekyll/converters/markdown.rb
+++ b/lib/jekyll/converters/markdown.rb
@@ -72,9 +72,14 @@ module Jekyll
         ".html"
       end
 
-      def convert(content)
+      def convert(content, path = "")
         setup
-        @parser.convert(content)
+        convert_arity = @parser.method(:convert).arity
+        if convert_arity == 2
+          @parser.convert(content, path)
+        else
+          @parser.convert(content)
+        end
       end
 
       private
--- a/lib/jekyll/converters/markdown/kramdown_parser.rb
+++ b/lib/jekyll/converters/markdown/kramdown_parser.rb
@@ -40,7 +40,8 @@ module Jekyll
           document = Kramdown::Document.new(content, @config)
           html_output = document.to_html
           document.warnings.each do |warning|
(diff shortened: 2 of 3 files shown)


pathawks:
Regarding `smartify`: https://github.com/gettalong/kramdown/issues/473

ashmaroli:
any word on the patch here? :smiley:

pathawks:
I'm not a fan of the way this is done, but I'm not sure what the better way is.

I don't think the converter needs to know the document path. This feels like a leaky abstraction.

I wonder if Jekyll::Converter could have an array where warnings could be added, and then printed via renderer?

ashmaroli:
> I wonder if Jekyll::Converter could have an array where warnings could be added, and then printed via renderer?

Ah! I like that idea..

pathawks:
Maybe each `warning` could include a `message` and a `lineno`. This way, if we got ambitious, we could adjust the line number before it is printed to account for things like front-matter. That way, the line number that we print is the *actual* line number in the file, rather than the line number in the MarkDown block.

renderer could print each warning in a standard format, and do things like use a different color for the filename or whatnot, and each converter will never need to know or worry about those implementation details.

ashmaroli:
That's a compound list of enhancements.. I like it but prefer to have the enhancement added bit-by-bit over time..

pathawks:
What I am suggesting is that we build this with flexibility in mind, so we don’t trap ourselves in a corner.

ashmaroli:
> What I am suggesting is that we build this with flexibility in mind

I got that.. I'm on board.. thinking of a way to implement this with minimal collaterals.. :thinking:

pathawks:
We should keep thinking about this, but this PR in its current state is not the direction we want to go.

Closing for now.

pathawks:
> Any news on this problem?

I don't believe so. Care to open a PR?

Xeverous:
#7149

PR #6554, opened 2017-11-13
Add an option to configure kramdown warning output

ashmaroli:
With this, warnings from kramdown are swallowed by default.
Add `log_warnings: true` under `kramdown:` options to enable them

### DIFF
--- a/docs/_docs/configuration.md
+++ b/docs/_docs/configuration.md
@@ -688,6 +688,7 @@ kramdown:
   input:          GFM
   hard_wrap:      false
   footnote_nr:    1
+  log_warnings:   false
 ```
 
 ## Liquid Options
--- a/docs/_docs/configuration.md
+++ b/docs/_docs/configuration.md
@@ -688,7 +688,7 @@ kramdown:
   input:          GFM
   hard_wrap:      false
   footnote_nr:    1
-  log_warnings:   false
+  show_warnings:  false
 ```
 
 ## Liquid Options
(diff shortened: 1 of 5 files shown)


pathawks:
I’m wondering if this should be a more general option (rather than kramdown specific) so that other converters can opt-in as well.

ashmaroli:
the warnings are from Kramdown.. why would other converters need to chimy in?

DirtyF:
@pathawks you're thinking of something like:

```yaml
markdown: commonmark
  - warnings: true
```
?

pathawks:
What I mean is an option that allows _converters_ to output warnings, rather than sn option that allows _kramdown_ to output warnings.

@DirtyF Yes, I’m thinking about CommonMark, but maybe Textile or SmartyPants also.

ashmaroli:
changing `config["markdown"]` from a `String` to a `Hash` and then adding in code for validating and backwards-compatibility seems unnecessary complexity to me..

parkr:
We have `show_drafts: true` and `—drafts`. I’d be down to see `show_warnings: true` and `—warnings`.

ashmaroli:
> I’d be down to see `show_warnings: true` and `--warnings`.

hmm.. can I change them to `show_converter_warnings: true` and `--converter-warnings`? I agree the strings are longer but unlike `verbose: true` and `--verbose` that affects verbose output throughout the build, the `show_warnings: false` would only turn off warnings from converters. Other instances of `Jekyll.logger.warn` will still be fired unchecked..

DirtyF:
Converter is a generic term that could refer to e.g. Liquid.
We're only talking about *markdown* converters here (kramdown, commonmark, redcarpet) right?

So the option would be part of the `kramdown` hash (see https://github.com/jekyll/jekyll/pull/6554#issuecomment-343985665).

ashmaroli:
Why was this merged manually..? This needs to be registered in our `History.markdown`..

PR #6555, opened 2017-11-13
Updated Jekyll-Pug listing to include official website

DougBeney:
The official website includes user-friendly instructions, easy for a Jekyll beginner. Soon, there will even be a section for Jekyll-Pug themes.

I also improved the description. Jekyll-Pug is no longer a basic converter plugin. It now supports Jekyll includes, layouts, a custom caching engine, and much more.

An upcoming feature is full Pug support, where you can use native Pug variables, for loops, and if statements, rather than using the Liquid equivalent.

### DIFF
--- a/docs/_docs/plugins.md
+++ b/docs/_docs/plugins.md
@@ -785,7 +785,7 @@ LESS.js files during generation.
 
 #### Converters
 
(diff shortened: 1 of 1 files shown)


DirtyF:
@jekyllbot: merge +docs

PR #6556, opened 2017-11-14
Add PageWithoutAFile class from jekyll plugins

ashmaroli:
Official plugins `jekyll-sitemap`, `jekyll-feed` and a couple more add a `Page` subclass to generate a file (usually from Liquid templates not present in the user's source directory).

Advantages
  - Users adding both plugins need not allocate memory for the twin subclasses.
  - Also allows other third-party plugins that follows the above implementation.

---

Plugins known to implement this subclass or similar:
  - [`jekyll-sitemap`](https://github.com/jekyll/jekyll-sitemap/blob/7c944d0cfd151a51f210ac4eb0ceeb56f3a74031/lib/jekyll/page_without_a_file.rb#L3-L9)
  - [`jekyll-feed`](https://github.com/jekyll/jekyll-feed/blob/2ff6fdcb77cea99ed67b6be0378ccc3ae60417d8/lib/jekyll-feed/page-without-a-file.rb#L3-L9)
  - [`jekyll-admin`](https://github.com/jekyll/jekyll-admin/blob/3415b5a15d4b98d0b41211f00a6a9ae13654a5b4/lib/jekyll-admin/page_without_a_file.rb#L1-L7)
  - [`jekyll-redirect-from`](https://github.com/jekyll/jekyll-redirect-from/blob/37eed64edc1cb9e21f6a151064beef059d2e31ea/lib/jekyll-redirect-from/page_without_a_file.rb#L3-L9)

### DIFF
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -184,4 +184,14 @@ module Jekyll
       true
     end
   end
+
+  class PageWithoutAFile < Page
+    # ---------------------------------------------------------------
+    # Extracted from official plugins, jekyll-feed and jekyll-sitemap
+    # ---------------------------------------------------------------
+
+    def read_yaml(*)
+      @data ||= {}
+    end
+  end
 end
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -185,11 +185,12 @@ module Jekyll
     end
   end
 
+  # A Jekyll::Page subclass to handle processing files that exist outside the
+  # site's configured source directory.
+  # The resulting object will be processed by Liquid and rendered directly at
+  # the destination, without an intermediate reading to determine page-data
+  # and page-content based on Front Matter delimiters.
   class PageWithoutAFile < Page
-    # ---------------------------------------------------------------
-    # Extracted from official plugins, jekyll-feed and jekyll-sitemap
-    # ---------------------------------------------------------------
-
     def read_yaml(*)
       @data ||= {}
     end
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -184,15 +184,4 @@ module Jekyll
       true
     end
   end
-
-  # A Jekyll::Page subclass to handle processing files that exist outside the
-  # site's configured source directory.
-  # The resulting object will be processed by Liquid and rendered directly at
-  # the destination, without an intermediate reading to determine page-data
-  # and page-content based on Front Matter delimiters.
-  class PageWithoutAFile < Page
-    def read_yaml(*)
-      @data ||= {}
-    end
-  end
 end
--- a/test/test_page_without_a_file.rb
+++ b/test/test_page_without_a_file.rb
@@ -3,13 +3,13 @@
 require "helper"
 
 class TestPageWithoutAFile < JekyllUnitTest
-  def setup_page(*args)
+  def setup_page(*args, base: source_dir, klass: PageWithoutAFile)
     dir, file = args
     if file.nil?
       file = dir
       dir = ""
     end
-    @page = PageWithoutAFile.new(@site, source_dir, dir, file)
+    klass.new(@site, base, dir, file)
   end
 
   context "A PageWithoutAFile" do
@@ -22,7 +22,7 @@ class TestPageWithoutAFile < JekyllUnitTest
       }))
     end
 
-    context "on intialized" do
+    context "with default site configuration" do
       setup do
         @page = setup_page("properties.html")
       end
@@ -34,7 +34,7 @@ class TestPageWithoutAFile < JekyllUnitTest
       end
 
       should "have basic attributes defined in it" do
-        regular_page = Page.new(@site, source_dir, "", "properties.html")
+        regular_page = setup_page("properties.html", :klass => Page)
         basic_attrs = %w(dir name path url)
         attrs = {
           "content"   => "All the properties.\n",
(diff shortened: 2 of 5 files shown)


pathawks:
I like this idea.
I’m not sure if this is the best place to put this subclass, but it is so small that I’m ok with it.

pathawks:
Plugins wouldn’t be able to take advantage of this _right_ away, as we tend to support older versions of Jekyll.

I wonder if, as long as we’re consolidating this, we could provide a mechanism to let the user know where a `PageWithoutAFile` originated, since the path is useless.

ashmaroli:
> take advantage of this *right away*, as we tend to support older versions of Jekyll.

I understand.. but it'll be easier once GitHub Pages switches onto a release with this patch.

> I wonder if, as long as we’re consolidating this, we could provide a mechanism to let the user know where a PageWithoutAFile originated, since the path is useless.

A log message would be great. But, its better if its implemented by the concerned plugin itself..
```
           JekyllFeed: feed.xml generated
        JekyllSitemap: sitemap.xml generated
```

ashmaroli:
@benbalter I'd like your thoughts on this.. Thanks.

lib/jekyll/page.rb
```diff
@@ -184,4 +184,14 @@ def write?
       true
     end
   end
+
+  class PageWithoutAFile < Page
```
mattr-: I would prefer a different name such as `BasePage` or `BasicPage` here. The fact that it doesn't have a file is irrelevant, IMHO

lib/jekyll/page.rb
```diff
@@ -184,4 +184,14 @@ def write?
...
+    # ---------------------------------------------------------------
```
mattr-: This documentation here isn't enough for justifying and explaining the existence of this class. Could we get some proper class level documentation?

mattr-:
> Users adding both plugins need not allocate memory for the twin subclasses.

Do you have hard numbers for this?

I do like this idea. Something that multiple plugins are using should be in the core. With just a few small tweaks, I think this will be ready.

ashmaroli:
> Do you have hard numbers for this?

Sorry, no. just an assumption..

lib/jekyll/page.rb
ashmaroli: Naming has been a rough sea here at Jekyll AFAIR.. so I'll change it to whatever the majority zeros in..

test/test_page_without_a_file.rb
```diff
@@ -0,0 +1,123 @@
...
+    setup do
+      clear_dest
+      @site = Site.new(Jekyll.configuration({
+        "source"            => source_dir,
+        "destination"       => dest_dir,
+        "skip_config_files" => true,
+      }))
+    end
+
+    context "on intialized" do
```
Crunch09: `on intialize` ? Or maybe even `with default config` ? What do you think?

test/test_page_without_a_file.rb
```diff
@@ -0,0 +1,123 @@
+# frozen_string_literal: true
+
+require "helper"
+
+class TestPageWithoutAFile < JekyllUnitTest
+  def setup_page(*args)
```
Crunch09: tiny thing: Instead of overriding this entire method to have the only difference in having `PageWithoutAFile.new` instead of `Page.new` ,what do you think of changing the method signature to something like:
```ruby
def setup_page(*args, klass: Page)
```

ashmaroli: Thank you for the suggestion. I had considered using a similar signature (`klass=Page, *args`), but opted out since I'm creating a `Jekyll::Page` instance only once in this new unit-test-class.

That said, I like your syntax with Keyword Arguments. Improves future tests-scalability. :+1:

test/test_page_without_a_file.rb
ashmaroli: I disagree with `with default config`.. `PageWithoutAFile` instances do not change on passing different parameters. For a given site, its always `nil` content and `empty` data hash.

test/test_page_without_a_file.rb
```diff
@@ -3,13 +3,13 @@
 require "helper"
 
 class TestPageWithoutAFile < JekyllUnitTest
-  def setup_page(*args)
+  def setup_page(*args, base: source_dir, klass: PageWithoutAFile)
```
Crunch09: @ashmaroli Hey, sorry my first suggestion was unclear: I tought about moving `setup_page`, which is also defined in [TestPage](https://github.com/jekyll/jekyll/blob/73419cb374be1b8f45818a23116cf71db93549ce/test/test_page.rb#L6-L13), to [JekyllUnitTest](https://github.com/jekyll/jekyll/blob/master/test/helper.rb#L90) and give it then a `klass` option because that's the only difference between those two methods and then we would only define it once :)

ashmaroli: I see.. I should've understood your obvious intention.. my bad..

Crunch09: No worries at all! It's already better 😉

ashmaroli: I don't want to change `test_page.rb` in this PR (since its an addition-only patch, right now..). Perhaps, a separate future PR would be best..

Crunch09: Fair enough 👍

DirtyF:
@jekyllbot: merge +dev

PR #6559, opened 2017-11-15
Symbolize kramdown config keys

### LINKED ISSUE #6558: Jekyll messes up the kramdown hash, creating symbols and string.
Blame: https://github.com/jekyll/jekyll/commit/143367c5ca801a152315ee2de6f4766c6ae1df55

screenshot-2017-11-14_13-43-00

ashmaroli:
Resolves #6558 

/cc @envygeeks

### DIFF
--- a/lib/jekyll/converters/markdown/kramdown_parser.rb
+++ b/lib/jekyll/converters/markdown/kramdown_parser.rb
@@ -36,7 +36,7 @@ module Jekyll
         end
 
         def convert(content)
-          document = Kramdown::Document.new(content, symbolize_keys(@config))
+          document = Kramdown::Document.new(content, symbolize_config)
           html_output = document.to_html
           document.warnings.each do |warning|
             Jekyll.logger.warn "Kramdown warning:", warning
@@ -45,8 +45,10 @@ module Jekyll
         end
 
         private
-        def symbolize_keys(hash)
-          hash.reduce({}) { |hsh, (k, v)| hsh.merge(k.to_sym => v) }
+        def symbolize_config
+          @config.keys.each_with_object({}) do |key, obj|
+            obj[key.to_sym] = @config[key]
+          end
         end
 
         private
--- a/lib/jekyll/converters/markdown/kramdown_parser.rb
+++ b/lib/jekyll/converters/markdown/kramdown_parser.rb
@@ -33,16 +33,20 @@ module Jekyll
           @config["syntax_highlighter_opts"] ||= {}
           @config["coderay"] ||= {} # XXX: Legacy.
           modernize_coderay_config
-          make_accessible
         end
 
         def convert(content)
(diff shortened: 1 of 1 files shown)


pathawks:
This reintroduces issue #5980

envygeeks:
@pathawks how does it re-introduce that problem? It should have symbol keys.

lib/jekyll/converters/markdown/kramdown_parser.rb
```diff
@@ -46,6 +45,11 @@ def convert(content)
         end
 
         private
+        def symbolize_keys(hash)
+          hash.reduce({}) { |hsh, (k, v)| hsh.merge(k.to_sym => v) }
```
envygeeks: Readability wise this should be be a `do...end` block, not a `{}` block.

In theory `Hash[hash.keys.map(&:to_sym).zip(hash.values)]` will be faster, since it should be a single depth hash, and you aren't really nesting it, that said, merge is slow, if you really want it to be a block, it should be `each_with_object` which is way faster than reduce when you want to work the way you want.  Reduce is only marginally faster than `each_with_object` and only when you are using fast methods like `:<<` and others, but when you are merging (which is expensive to begin with) it's hella slower.

ashmaroli: This method is based on https://github.com/jekyll/jekyll/blob/621df9db8e72cfa9706ccd52b86a590bcd5d7cc3/lib/jekyll/configuration.rb#L103-L108

Can that method be optimized too, considering the depth of nesting involved..?

envygeeks: @ashmaroli yessum, in a scenario (like a full-stack application) I would never reduce on a hash like that, because it's just to slow and TTFB matters a lot to me, and `reduce`+`#merge` is slow compared to `each_with_object`+`#[]=`.  It could provide *some* perf to Jekyll but probably not enough to matter unless that method is used everywhere.

envygeeks: In general though, I would have just accepted the performance penalty of being indifferent with ActiveSupport though, because at least then I don't have to deal with all this logic non-sense, conversion non-sense, and all I need to worry about is shipping source.

ashmaroli:
yep, that's the reason I added this to the `4.0` milestone.. With this PR, the config is not going to have *both string keys and symbolized keys..* just symbolized keys..

pathawks:
I don't understand why we need to call `symbolize_config` each time a new document is converted.

ashmaroli:
> why we need to call `symbolize_config` each time

Very good question, @pathawks
I now see how this was a bad idea conceptually..

Still, the fact remains that we should simply symbolize the keys of kramdown-config-hash in `4.0` instead of including both string-keys and their symbolized counterparts in the hash..

PR #6560, opened 2017-11-15
Drop support for Ruby 2.1 and 2.2

ashmaroli:
Ruby 2.1 reached EOL on Mar 31, 2017 and Ruby 2.2 is slated for EOL on Mar 31, 2018
Drop support for above versions and allow incorporating new features introduced since Ruby 2.3


ashmaroli:
> I’m fine with this. We can just bump Jekyll’s major version and let everyone know.

Or when we release `v3.7.0`, we can just state:
  - that `v3.7.0` will be the last minor release in the `v3.x` series 
  - and only bug-fixes will be back-ported to `v3.x` in subsequent releases.

/cc @jekyll/core

pathawks:
So, merge this right after `3.7.0` drops? :+1:

I don’t think we need to commit to back porting anything. This ain’t WordPress; there is no server component. If somebody wants to keep using an old version, there is absolutely no reason they shouldn’t, but our energy should be focused on moving forward.

ashmaroli:
> So, merge this right after `3.7.0` drops? :+1: 

I'm in favor of starting 4.0 development within the near future.
Once `4.0` dev starts, contributors can start submitting PRs for possible changes to Jekyll.

> I don’t think we need to commit to back porting anything.

IMO, we should commit to release `v3.7.x` as necessary for users not jumping onto the `4.0` bandwagon right away, because irrespective of their reason to not migrate to `4.0`, they still matter to us.. :smiley:

DirtyF:
Any reason not to merge this now that Ruby 2.2 is officially EOL?

oe:
@DirtyF I'd rather not do this so close to the 3.8 release, especially since we've already released RC versions. IMO it'd be better to wait until 4.0.

pathawks:
Are we ready to merge this?

pathawks:
Need to resolve conflicts before merge.

ashmaroli:
> Need to resolve conflicts before merge.

@pathawks I do not see any conflicts..

pathawks:
Let's see if @jekyllbot sees any conflicts.

@jekyllbot: merge +major

pathawks:
Weird. Web interface is lying to me 😝

PR #6567, opened 2017-11-19
Site header redesign

ashmaroli:
- render site header at full-width using Flexbox
  - use site color scheme for search input

### Desktop resolution sample

localhost_4000_docs_home_

### Sample at 800px viewport-width

localhost_4000_docs_home_800px

### Sample at 480px viewport-width

localhost_4000_docs_home_ 480px

### DIFF
--- a/docs/_includes/header.html
+++ b/docs/_includes/header.html
@@ -16,8 +16,9 @@
     </nav>
     <div class="search hide-on-mobiles">
       {% include search/input.html %}
-
-      <ul class="meta">
+    </div>
+    <div class="meta hide-on-mobiles">
+      <ul>
         <li>
           <a href="{{ site.repository }}/releases/tag/v{{ jekyll.version }}">v{{ jekyll.version }}</a>
         </li>
--- a/docs/_includes/header.html
+++ b/docs/_includes/header.html
@@ -19,7 +19,7 @@
 
       <ul class="meta">
         <li>
-          <a href="#">v{{ jekyll.version }}</a>
+          <a href="{{ site.repository }}/releases/tag/v{{ jekyll.version }}">v{{ jekyll.version }}</a>
         </li>
         <li>
           <a href="{{ site.repository }}">GitHub</a>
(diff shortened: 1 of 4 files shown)


DirtyF:
Nicely done. 👏 

On full width on desktop we could:

* add jekyll latest version and get GitHub link out of the main navigation (it's an external link).
* increase search font-size a bit so that input height equals current navigation item.

jekyll-header

Now that we use flexbox, I'd like to also display search for lower resolutions and get a full width search on mobile.

ashmaroli:
Regarding the version, would it be static or dynamic via `{{ jekyll.version }}. And if its dynamic, would the site show the verrsion on `master` or the one on `github-pages`?

with this, search has been opened up for 800px wide and above.. was planning to raise the accessibility in future PRs..

DirtyF:
For now I guess we could go the easy way with `{{ jekyll.version }}` as GitHub Pages usually updates quickly.

DirtyF:
Looks like the search needs a bit of margin to be perfectly vertically aligned. 

If I get `.meta`out of the search, navigation, search and meta are more equally distributed:

header-nav

ashmaroli:
> If I get .metaout of the search, navigation, search and meta are more equally distributed:

True. I had modeled it based on your [earlier suggestion](https://user-images.githubusercontent.com/103008/32991456-514eb1fc-cd3c-11e7-8ec9-e953c287bbb9.png). But the current suggestion will allow us little more freedom with the search input.. so :+1: 

The search looks aligned properly to me.. :nerd_face: :mag:

DirtyF:
Thanks @ashmaroli, looks great 🙌 

Maybe we should move away from gridism at one point if we want more control over the grid?

DirtyF:
@jekyllbot: merge +site

ashmaroli:
> Maybe we should move away from gridism at one point if we want more control over the grid?

Yes, like you said a few days ago, it was designed 5 years ago.
Browsers have improved quite a lot in the interim.

PR #6569, opened 2017-11-19
return correct file in dir if dir has same name as file

### LINKED ISSUE #6475: If `bar.html` exists, then `jekyll s` will show `bar.html` for `bar/.*` unless something else explicitly (but not implicitly) overrides it
If a file and a folder have the same "permalink" name, `jekyll s` shows wrong content at files in the folder.

Windows Linux Subsystem, `bash` via Windows 10 Powershell

- [x] It was another bug.

## My Reproduction Steps

`jekyll s` does strange things when a file has the same name as a folder:

- Site is configured as `permalink: /:path`
- `bar.md`, should render at `/bar`
- `bar/baz.md`, should render at `/bar/baz`

Unfortunately:

On accessing `/bar/baz`:
- In 3.5.2 everything was okay.
- Since 3.6.0 it shows the content of `bar.md`

I think this is a serious bug. It probably is a regression of #6231, which tried to fix some related problems.

In `_site` everything is generated as it should:
- `/bar.html` with content from `bar.md`
- `/bar/baz.html` with content from `bar/baz.md`

So the problem is in the way `jekyll s` decides which file to serve.

### Example repo

Here is a repo that shows this exact scenario:
https://github.com/janpio/jekyll-path-test
Use the site navigation for the three example pages. 

If you use `jekyll s`, note how you see the wrong content for "Baz" (only 3.6.x).

(The project also includes 3 collections that test further iterations of this problem:, _test1 and _test2 with `.html` ending are fine, but _test3 has the exact same problem.)

## The Output I Wanted

You can see how everything should work at Github Pages:
https://janpio.github.io/jekyll-path-test/

Crunch09:
- bar.html
- bar
  |- baz.html

_GET /bar/_
_Before:_ Returns bar.html
_After:_  Returns bar.html

_GET /bar/baz_
_Before:_ Returns bar.html
_After:_  Returns bar/baz.html

_GET /bar/whatever_
_Before:_ Returns bar.html
_After:_  Returns 404

This fixes #6475. Thanks @janpio for the great reproduction script! 👍 

cc: @jekyll/build

### DIFF
--- a/lib/jekyll/commands/serve/servlet.rb
+++ b/lib/jekyll/commands/serve/servlet.rb
@@ -18,13 +18,17 @@ module Jekyll
           super
         end
 
+        def search_index_file(req, res)
+          super || search_file(req, res, ".html")
+        end
+
         # Add the ability to tap file.html the same way that Nginx does on our
         # Docker images (or on GitHub Pages.) The difference is that we might end
         # up with a different preference on which comes first.
 
         def search_file(req, res, basename)
           # /file.* > /file/index.html > /file.html
-          super || super(req, res, ".html") || super(req, res, "#{basename}.html")
+          super || super(req, res, "#{basename}.html")
         end
 
         # rubocop:disable Naming/MethodName
--- a/test/fixtures/webrick/bar.html
+++ b/test/fixtures/webrick/bar.html
@@ -1,0 +1,1 @@
+Content of bar.html
--- a/test/fixtures/webrick/bar/baz.html
+++ b/test/fixtures/webrick/bar/baz.html
@@ -1,0 +1,1 @@
+Content of baz.html
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -44,6 +44,8 @@ require "shoulda"
 
 include Jekyll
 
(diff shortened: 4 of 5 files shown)


ashmaroli:
@Crunch09 Does this address #6222 as well (resolved via  #6231). I think you should test for that case as well, if this doesn't already..

Crunch09:
@ashmaroli Yes, this is checked by [this](https://github.com/Crunch09/jekyll/blob/4a769ec5cb32db65aa90fb3c68c255fc6b23a50b/test/test_commands_serve_servlet.rb#L18-L23) test.

parkr:
Have you verified this against GitHub Pages? Ideally we'd keep this server behaviour the same, especially around an edge case like this which is confusing to users.

Crunch09:
@parkr Checked the current Github-Pages behavior with https://janpio.github.io/jekyll-path-test/ (provided by @janpio ). After this PR, webrick should have the same behavior.

DirtyF:
@jekyllbot: merge +bug

PR #6570, opened 2017-11-20
Move logo above site navigation on small screens

ashmaroli:
#### Current layout

localhost_4000_ logo-below

#### Proposed layout

localhost_4000_logo-above

### DIFF
--- a/docs/_includes/header.html
+++ b/docs/_includes/header.html
@@ -1,7 +1,4 @@
 <header>
-  <nav class="mobile-nav show-on-mobiles">
-    {% include mobile-nav-items.html %}
-  </nav>
   <div class="flexbox">
     <div class="center-on-mobiles">
       <h1>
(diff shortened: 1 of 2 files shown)


DirtyF:
The most significant update here would be to stick the navigation on the bottom on mobile, because we use our thumbs to tap and it's hard to reach the top of the screen.

image

ashmaroli:
> update here would be to stick the navigation on the bottom on mobile

I agree on this, even though the examples you [posted above](https://user-images.githubusercontent.com/103008/33010762-f98f0c82-cddb-11e7-9d65-cc00295d23cd.png) are of native apps and ours is a website.. :smiley:

DirtyF:
There's no such things as native apps or websites for our thumbs, I should have posted black screen instead

image


source: https://www.lukew.com/ff/entry.asp?1944

DirtyF:
@jekyllbot: merge +site

pathawks:
I don’t think the nav bar should be always visible at the bottom of the screen. Those buttons will not be frequently used–maybe once or twice each session. Making them always visible seems to give them too much prominence.

ashmaroli:
we can always have a copy of the navbar at the bottom of the page.. then its upto the user to either scroll up or down to a navbar, whichever is nearer..

PR #6575, opened 2017-11-24
Use version-badge on an existing feature intro

ashmaroli:
I'm opting to showcase this feature with the version-badge instead of editing the `From 'v3.5'..` part out entirely, just in case end-users / developers would need a quick reference as to what version introduced the concerned feature.


DirtyF:
@jekyllbot: merge +site

PR #6584, opened 2017-11-29
Cleanup test_redcarpet.rb

ashmaroli:
- replace multiple calls to creating heredocs containing the same content, with an instance variable
- use `%()` to wrap strings containing double quotes to avoid having to escape internal double-quotes, thereby improving readability at the minor expense of multiple string concatenation.


pup:
@jekyllbot: merge +dev

PR #6589, opened 2017-12-02
Dependency: Bump jekyll-watch to 2.0

DirtyF:
See https://github.com/jekyll/jekyll-watch/releases/tag/v2.0.0


DirtyF:
I'll go with 3.7.0 then so that the majority can have the full path output right away.

@jekyllbot: merge +minor

PR #6593, opened 2017-12-04
move duplicate code to a single private method

ashmaroli:
Just keeping things DRY..

### DIFF
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -449,7 +449,7 @@ module Jekyll
     def render_docs(payload)
       collections.each_value do |collection|
         collection.docs.each do |document|
-          render_this(document, payload)
+          render_regenerated(document, payload)
         end
       end
     end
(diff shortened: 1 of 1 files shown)


ashmaroli:
> What do you think of maybe `regenerate_file` instead?

I don't like that since the primary calling is to render.. How about `render_regenerated`..?

Crunch09:
@ashmaroli 👍

lib/jekyll/site.rb
```diff
@@ -449,21 +449,23 @@ def configure_file_read_opts
...
-          page.output = Jekyll::Renderer.new(self, page, payload).run
-          page.trigger_hooks(:post_render)
-        end
+        render_regenerated(page, payload)
+      end
+    end
+
+    private
+    def render_regenerated(document, payload)
+      if regenerator.regenerate?(document)
```
parkr: Let's use a guard clause here: 

```ruby
return unless regenerator.regenerate?(document)

document.output = ...
````

oe:
@jekyllbot: merge +dev

PR #6596, opened 2017-12-04
Require external library only if necessary

ashmaroli:
Even though the Ruby interpreter does not require a library more than once, there's scope for optimization by not initiating `Jekyll::External.require_with_graceful_fail` unless necessary within current build session

### DIFF
--- a/lib/jekyll/converters/markdown/kramdown_parser.rb
+++ b/lib/jekyll/converters/markdown/kramdown_parser.rb
@@ -14,7 +14,9 @@ module Jekyll
         }.freeze
 
         def initialize(config)
-          Jekyll::External.require_with_graceful_fail "kramdown"
+          unless defined?(Kramdown)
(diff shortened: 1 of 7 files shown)


parkr:
@jekyllbot: merge +minor

ashmaroli:
@parkr this should've been marked as a `bug-fix` change instead of `minor-enhancement`

parkr:
@ashmaroli ruby’s `require` method is highly cached so there wasn’t really a bug or performance problem. We can reclassify if you want.

pathawks:
> ruby’s require method is highly cached so there wasn’t really a bug or performance problem.

In that case, what advantage does this change provide?

ashmaroli:
> so there wasn’t really a bug or performance problem.

oi.. so this was an unnecessary change..?

> We can reclassify if you want.

Depends on whether the next release is a `v3.7.1` or `v3.8.0`.. :grin:

PR #6597, opened 2017-12-04
Excerpt relative-path should match its path

ashmaroli:
Currently a document's excerpt's relative path equals the doc's relative path instead of the excerpt-path relative to the site's source.

### Before
```
             doc.path: my_blog/_posts/2017-12-04-some-doc.md
         excerpt.path: my_blog/_posts/2017-12-04-some-doc.md/#excerpt
    doc.relative_path: _posts/2017-12-04-some-doc.md
excerpt.relative_path: _posts/2017-12-04-some-doc.md
```

### After
```
             doc.path: my_blog/_posts/2017-12-04-some-doc.md
         excerpt.path: my_blog/_posts/2017-12-04-some-doc.md/#excerpt
    doc.relative_path: _posts/2017-12-04-some-doc.md
excerpt.relative_path: _posts/2017-12-04-some-doc.md/#excerpt
```

### DIFF
--- a/lib/jekyll/excerpt.rb
+++ b/lib/jekyll/excerpt.rb
@@ -8,7 +8,7 @@ module Jekyll
     attr_accessor :content, :ext
     attr_writer   :output
 
-    def_delegators :@doc, :site, :name, :ext, :relative_path, :extname,
+    def_delegators :@doc, :site, :name, :ext, :extname,
                           :render_with_liquid?, :collection, :related_posts,
                           :url, :next_doc, :previous_doc
 
@@ -39,6 +39,13 @@ module Jekyll
     # Returns the path for the doc this excerpt belongs to with #excerpt appended
     def path
       File.join(doc.path, "#excerpt")
+    end
+
+    # 'Relative Path' of the excerpt.
+    #
+    # Returns the relative_path for the doc this excerpt belongs to with #excerpt appended
+    def relative_path
+      File.join(doc.relative_path, "#excerpt")
     end
 
     # Check if excerpt includes a string
--- a/test/test_excerpt.rb
+++ b/test/test_excerpt.rb
@@ -78,6 +78,15 @@ class TestExcerpt < JekyllUnitTest
       end
     end
 
+    context "#relative_path" do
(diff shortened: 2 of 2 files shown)


mattr-:
Can we do this for 3.7.0 since the excerpt's file path is made available to liquid? It would depend on whether or not we'd consider anything we export to liquid as part of a public facing API that we would need to keep backwards compatible. I tend to think that it is, but I'll let others weigh in.

ashmaroli:
@mattr- I didn't think this would've been a breaking change till I saw the Travis failures.. This cannot be shipped until `4.0` since the tests clearly document the current expected behavior even on `test_document.rb`

ashmaroli:
related:
https://github.com/jekyll/jekyll/blob/9ec9273ed9f04a666e6bc434baa2790321675c8f/lib/jekyll/drops/document_drop.rb#L14

DirtyF:
> I didn't think this would've been a breaking

What would it break?

ashmaroli:
For me to fix the failing test in `test_excerpt.rb`, the **best approach would be to alter the following** 
https://github.com/jekyll/jekyll/blob/9ec9273ed9f04a666e6bc434baa2790321675c8f/lib/jekyll/drops/document_drop.rb#L14 (because the test is testing `@excerpt.to_liquid["path"]`)

the side-effect of the change being no longer having `doc.to_liquid["path"]` mirror `doc.to_liquid["relative_path"]`  &mdash;  potentially a breaking-change for users that depend on serialized / jsonified Jekyll objects (like `jekyll-admin`,etc) and perhaps others I'm not sure of at the moment.

---

That said, if this should really be shipped via `3.7.0`, we'll have to manually ensure `excerpt.to_liquid["path"] = excerpt.path` and leave `document_drop.rb` untouched..

ashmaroli:
Designated as a `4.0` change because this alters the outcome of an existing public_method `:relative_path`

ashmaroli:
> it's just a log output

@DirtyF Please check #6621

DirtyF:
I have (hence my above comment)

DirtyF:
@jekyllbot: :shipit: +minor

PR #6600, opened 2017-12-05
Bump rdoc to 6.0

### LINKED ISSUE #6599: Update dependency constraint to allow for rdoc v6.0.0
Hey there! :wave:

I noticed that the constraint you have for rdoc doesn't allow for the latest version to be used.

The constraint I found was `~> 5.0`, and the latest version available is `6.0.0`.

jekyllbot:
Fix #6599


DirtyF:
image

DirtyF:
Note:  `rdoc` is an optional dependency and now requires Ruby >= 2.2.2

Gemfile
```diff
@@ -75,7 +75,7 @@ group :jekyll_optional_dependencies do
   gem "jekyll-redirect-from"
   gem "kramdown", "~> 1.14"
   gem "mime-types", "~> 3.0"
-  gem "rdoc", "~> 5.0"
+  gem "rdoc", "~> 6.0"
```
pathawks: Maybe still allow `5.x` for Ruby 2.1?

```
gem "rdoc", [">= 5.0", "< 7.0"]

Comment (ashmaroli) [reply Gemfile]: like we have for other gems in similar situations.
```ruby
gem "rdoc", RUBY_VERSION >= "2.2.2" ? "~> 6.0" : "~> 5.1"
```

parkr:
We've dropped Ruby 2.1, if that changes anything.

parkr:
@jekyllbot: merge +dev

PR #6602, opened 2017-12-06
Memoize path to metadata file

ashmaroli:
The path to *metadata_file* need not be generated each time the file is referenced..


DirtyF:
@jekyllbot: merge +fix

PR #6605, opened 2017-12-07
Register reload hooks in Server#process

ashmaroli:
Resolves issue mention in https://github.com/jekyll/jekyll/pull/5142#issuecomment-349786108

Calling `:register_reload_hooks` in `#process` after the site starts to `generate...` instead of much earlier in the build-process..

/cc @awood

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -86,7 +86,6 @@ module Jekyll
           # a reactor created by a previous test when our test might not
           @reload_reactor = nil
 
-          register_reload_hooks(opts) if opts["livereload"]
           config = configuration_from_options(opts)
           if Jekyll.env == "development"
             config["url"] = default_url(config)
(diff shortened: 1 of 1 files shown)


DirtyF:
It works 👍 

port is displayed only if I had `"livereload_port" => 35_729,`, shouldn't it by the default?

ashmaroli:
> port is displayed only if..

I see the following with just `bundle exec jekyll s --livereload` and I feel its adequate.. though it would be better if the `http://` protocol was included as well..
```
LiveReload address: 127.0.0.1:35729
```

DirtyF:
@ashmaroli The port is displayed if you pass the --livereload option on the CLI but not if you only pass the livereload option to the options hash and then call `Jekyll::Commands::Serve.process(options)` (in a rake task for example).

ashmaroli:
I see.. I was running `bundle exec jekyll s` from within the `docs/` folder..
FWIW, the `LiveReload address: 127.0.0.1:35729` message appears when I serve the `docs/` folder with `livereload: true` in `./docs/_config.yml`

awood:
@ashmaroli Thanks for fixing this so quickly.

ashmaroli:
@awood You're welcome.
A question:  If a user opts to run the server using the `--ssl-key` and `--ssl-cert` switches, will the LiveReload Address run on the `https://` address?

DirtyF:
@jekyllbot: merge +fix

PR #6607, opened 2017-12-07
Cleanup LiveReloadReactor

ashmaroli:
some changes to the livereload reactor I withheld requesting to made before this class was merged to avoid noise in that PR..

### DIFF
--- a/lib/jekyll/commands/serve/live_reload_reactor.rb
+++ b/lib/jekyll/commands/serve/live_reload_reactor.rb
@@ -38,7 +38,6 @@ module Jekyll
           ws.onerror { |error| log_error(error) }
         end
 
-        # rubocop:disable Metrics/MethodLength
         def start(opts)
           @thread = Thread.new do
             # Use epoll if the kernel supports it
@@ -60,7 +59,7 @@ module Jekyll
               EM.add_shutdown_hook { @stopped_event.set }
 
               Jekyll.logger.info "LiveReload address:",
-                "#{opts["host"]}:#{opts["livereload_port"]}"
+                "http://#{opts["host"]}:#{opts["livereload_port"]}"
             end
           end
           @thread.abort_on_exception = true
@@ -118,8 +117,9 @@ module Jekyll
 
         private
         def log_error(e)
-          raise e, "LiveReload experienced an error. " \
+          Jekyll.logger.error "LiveReload experienced an error. " \
             "Run with --trace for more information."
+          raise e
         end
       end
     end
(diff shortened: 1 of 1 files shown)


ashmaroli:
Guys, I want to draw your attention towards the following code which is in our `master` now..
https://github.com/jekyll/jekyll/blob/3efb1ad85794bdd4e58ac5f2774b83ea34ada18e/lib/jekyll/commands/serve/live_reload_reactor.rb#L141-L149
It doesn't seem right to me that we are going to printing the backtrace in the debug channel when it already crowded with build-related info..

lib/jekyll/commands/serve/live_reload_reactor.rb
```diff
@@ -25,29 +24,18 @@ def stop
...
-          ws.onerror do |error|
-            log_error(error)
-          end
+          ws.onopen { |handshake| connect(ws, handshake) }
+          ws.onclose { disconnect(ws) }
+          ws.onmessage { |msg| print_message(msg) }
+          ws.onerror { |error| log_error(error) }
         end
 
         # rubocop:disable Metrics/MethodLength
```
DirtyF: We can remove this comment:

```
lib/jekyll/commands/serve/live_reload_reactor.rb:41:9: W: [Corrected] Unnecessary disabling of Metrics/MethodLength.
        # rubocop:disable Metrics/MethodLength
```

lib/jekyll/commands/serve/live_reload_reactor.rb
```diff
@@ -96,11 +76,9 @@ def reload(pages)
...
             }
 
-            Jekyll.logger.debug("LiveReload:", "Reloading #{p.url}")
-            Jekyll.logger.debug(JSON.dump(msg))
-            @websockets.each do |ws|
-              ws.send(JSON.dump(msg))
-            end
+            Jekyll.logger.debug "LiveReload:", "Reloading #{p.url}"
+            Jekyll.logger.debug JSON.dump(msg)
+            @websockets.each { |ws| ws.send(JSON.dump(msg)) }
```
ashmaroli: @awood does `JSON.dump(msg)` need to be called separately for each websocket or can I store the value of the first dump to a local variable and then use that variable in the block..?

awood: You should be able to store the serialized JSON and re-use it.  Not sure why I didn't do that in the first place.  I believe `@websockets` will generally just have one item in it anyway (each item in the list represents a browser that's connected IIRC) unless someone is really excited about livereload.  But it would be better just to dump it to JSON once per `msg`

mattr-:
@jekyllbot: merge +dev

Issue #6297, opened 2017-08-11
Unexpected sorting result


ashmaroli:
This ticket refers to a situation brought to my notice via [this question at StackOverflow](https://stackoverflow.com/questions/45612870/how-do-you-sort-csv-data-in-jekyll-by-a-number)

Sorting an array of hashes (generated by `DataReader`), by an attribute whose value is nominally an integer returns incorrect results because the values are stringified.
i.e. 
With given datafile `datafile.csv` from the StackOverflow question:
```
type,icon,title,description,weight
URL,Weblink,Title,Description,10
URL,Weblink,Title 2,Description,20
URL,Weblink,Title 3,Description,100
```
the resulting array in `site.data["datafile"]` roughly contains:
```diff
--- expected 
+++ actual

--- "weight"=>10
+++ "weight"=>"10"
```

So, if the above is expected from `DataReader`, then we'll have to patch our `sort` filter, otherwise, do we patch `DataReader`?

/cc @jekyll/stability

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>3.3-stable</code> or <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, please consider building it first as a plugin. Jekyll 3 introduced [hooks](http://jekyllrb.com/docs/plugins/#hooks) which provide convenient access points throughout the Jekyll build pipeline whereby most needs can be fulfilled. If this is something that cannot be built as a plugin, then please provide more information about why in order to keep this issue open.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

Crunch09:
still open

parkr:
This is the `"10" < "2"` error? Ugh. We blanket `to_s` things in the `sort` filter, I think. We should fix that to not do that to Integers.

pathawks:
https://github.com/jekyll/jekyll/blob/4865b902c1ee105b2e214399f95baa410f6c9e82/lib/jekyll/filters.rb#L343-L358

I don’t see it.

pathawks:
Oh. Found it.

https://github.com/jekyll/jekyll/blob/4865b902c1ee105b2e214399f95baa410f6c9e82/lib/jekyll/filters.rb#L371-L381

greenkg:
[...]

PR #6608, opened 2017-12-07
update item_property to return numbers as numbers instead of strings

greenkg:
I added method "parse_sort_input" to check if the item properties were integers or floats and converted to integer or float if so. If not, the item_property method should still return a string.

This should allow properties to sort in correct order if properties are numeric.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,7 @@ source "https://rubygems.org"
 gemspec :name => "jekyll"
 
 gem "rake", "~> 12.0"
+gem "require_all"
 
 gem "rouge", ENV["ROUGE"] if ENV["ROUGE"]
 
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -371,13 +371,24 @@ module Jekyll
     def item_property(item, property)
       if item.respond_to?(:to_liquid)
         property.to_s.split(".").reduce(item.to_liquid) do |subvalue, attribute|
-          subvalue[attribute]
+          parse_sort_input(subvalue[attribute])
         end
       elsif item.respond_to?(:data)
-        item.data[property.to_s]
+        parse_sort_input(item.data[property.to_s])
       else
-        item[property.to_s]
+        parse_sort_input(item[property.to_s])
       end
+    end
+
+    private
+    # return numeric values as numbers for proper sorting
+    def parse_sort_input(property)
+      if (Integer(property) && true rescue false)
+        property.to_i
+      elsif (Float(result) && true rescue false)
+        property.to_f
+      end
+      return property
     end
 
     private


Gemfile
```diff
@@ -4,6 +4,7 @@ source "https://rubygems.org"
 gemspec :name => "jekyll"
 
 gem "rake", "~> 12.0"
+gem "require_all"
```
ashmaroli: What is the role of this gem here..?

lib/jekyll/filters.rb
```diff
@@ -371,15 +371,26 @@ def time(input)
...
+        parse_sort_input(item[property.to_s])
       end
     end
 
+    private
+    # return numeric values as numbers for proper sorting
+    def parse_sort_input(property)
+      if (Integer(property) && true rescue false)
+        property.to_i
+      elsif (Float(result) && true rescue false)
```
ashmaroli: No doubt, this appears elegant. But I'm on the fence regarding the use of inline `rescue` especially since its going to `rescue` everything to `false`

Is there another way to go about this, that doesn't involve coupling `sorting` with `rescue`-ing..?

Gemfile
greenkg: Shoot. The role of that gem there is to teach me to be more mindful of what files I've changed!

That is an error, and should not have been saved in there.

My understanding is, to update this pull request, I should make the change on my branch, repeat git add and git commit, and when I git push the branch, this pull request should also be updated. I gave that a shot.

ashmaroli: Your understanding is correct. Though you need to be careful of what gets checked in when you run `git add`

lib/jekyll/filters.rb
greenkg: The first way I went about it was using regex to check if the string was a number, which would avoid the "rescue" but the Integer("string") and Float("string") with rescue seemed to be more popular and considered more "readable". I have no preference for either, although it seems from further reading that the regex approach is faster.

The regex approach would look as below. I believe it can be reduced to just one "if" since both integers and floats will evaluate as "true" and can both be converted to floats and still sort properly.

```ruby
def parse_sort_input(property)
  numeric = /\A-?(\d+)?\.?\d+\Z/
  if (property =~ numeric)
    return property.to_f
  end
  return property
end
```

Did you have another approach in mind?

ashmaroli: I like this approach.. the method's intention is straightforward and transparent.
Unless someone else comes up and tell us this method is slower than the previous, I'd go with this approach, with some minor modifications..:

like, why are we capturing just the digit before the decimal if we intend it to be a float?
```
%r!\A(-?\d+?\.?\d?)\Z!
```
would be better **IMO**.. it captures the whole numeric..
now the refactoring..

```ruby
def parse_sort_input(property)
  last_match = property.match(%r!\A(-?\d+?\.?\d?)\Z!)[0]
  return property if last_match.nil?
  last_match.to_f
end
```

And lastly, wrap this contribution with a test that confirms your code actually does what it advertises.. :smiley:

greenkg: Thanks! I am new to testing, so it might take me a little time to work through that.

In the meantime, a couple questions:
1. Should I wait to submit a new pull request when I have both the code updated and the test written?
2. I'm also new to regex (I mean, I'm new to all of this, haha) and I see that while I was returning the whole numeric by doing property.to_f, I was only checking the pattern up to the decimal.
3. In your improved regex, should the second d? still be d+? I run into a problem where 12.3 matches, but 12.34 does not.
3. With the refactored code, if 
```property.match(%r!\A(-?\d+?\.?\d?)\Z!) is nil (i.e. there is no match), then
```property.match(%r!\A(-?\d+?\.?\d?)\Z!)[0]
returns the error "undefined method `[]' for nil:NilClass"
Am I missing something, or should we check if property.match(regex) is nil before assigning property.match(regex)[0] to last_match?

This works for me:
```ruby
def parse_sort_inputb(property)
  last_match = property.match(%r!\A(-?\d+?\.?\d+?)\Z!)
  return property if last_match.nil?
  return last_match[0].to_f
end

Comment (ashmaroli) [reply lib/jekyll/filters.rb]: @greenkg Yes, you're right. There was a mistake in both my suggested regex and Ruby code.. and the fact that you were returning the entire float anyways..
Besides, matching a regex and then converting the match to a float was not a good suggestion, either.. :face_with_head_bandage: 

With the regex, my intention was to match a larger set (that includes `1`, `1.`, `1.0`, `1.01`, `1.001`..) but in my haste I used the wrong quantifier. the second `\d` should've been `\d*` or `\d*?`

Regarding the Ruby aspect, your code right above :point_up_2:  is fine..(the last `return` isn't necessary in Ruby). But the following is better..

```ruby
def parse_sort_input(property)
  return property unless property =~ %r!\A-?\d+?\.?\d*\Z!
  property.to_f
end
```

pathawks: `\d+?` is the same as `\d*`
Make sure the RegEx *does not* match the empty string, as the example above would.

greenkg: @pathawks When I test empty strings, using d+? does not match an empty string but using d* in both locations does. So @ashmaroli 's modified regex does not match "" when I test it for that, at least. (Although I wasn't thinking to check for that before you pointed it out!)

pathawks: I see: `\d+?` is the same as `\d+`, it seems the `?` is just ignored?

In this case, this RegEx does _not_ match the empty string (yay) but neither does it match `.5` (oops).

```irb
 2.4.0 :001 > "" =~ %r!\A-?\d+?\.?\d*\Z!
 => nil 
 2.4.0 :002 > ".5" =~ %r!\A-?\d+?\.?\d*\Z!
 => nil 
 2.4.0 :003 > "0.5" =~ %r!\A-?\d+?\.?\d*\Z!
 => 0 
```

pathawks:
Fixes #6297

lib/jekyll/filters.rb
greenkg: @pathawks Ah okay, I see. Shoot. Now I've got the following
```ruby
return property unless (property =~ %r!\A-?(\d*\.?\d*)(\.?\d+)\Z!)
```

greenkg: @ashmaroli When you have a moment, I'd love some input on adding the test. With the last code revision in my branch, I'm now failing a sort test but I *think* it *should* be failing this test now.

It is failing at test_filters.rb line 1039 for the following test:
```ruby
assert_equal(
          [{ "a" => "10" }, { "a" => "2" }],
          @filter.sort([{ "a" => "10" }, { "a" => "2" }], "a")
        )
```
I *think* this test is expecting the very result that this change is supposed to avoid. So "10" and "2" are now getting sorted as 10 and 2.

Am I understanding that right and should this test be removed (and then replaced with a new test showing the opposite, essentially?)

Thank you for all the help!

pathawks: I would suggest:

```ruby
NUMBER_LIKE = %r!\A\s*(?:\d+\.?\d*|\.?\d+)\s*\Z!
```

We don't care if the String has leading or trailing spaces. If there is a decimal point, there _must_ be either a numeral preceding or (if there is not a preceding numeral) there must be a following numeral.

This RegEx matches `"1"`, `" 1 "`, `".1"`, `"1."`, and `"1.1"`. It does _not_ match `"1 1"`, `""`, or `"."`.

By making this a named constant, we can make the code read a bit more easily:

```ruby
return Float(property) if property =~ NUMBER_LIKE
return property
```

pathawks: We can also clean the RegEx up once merged. I don't mean for us to get stuck on that :+1:

lib/jekyll/filters.rb
```diff
@@ -371,15 +371,23 @@ def time(input)
...
-        item[property.to_s]
+        parse_sort_input(item[property.to_s])
       end
     end
 
+    private
+    # return numeric values as numbers for proper sorting
+    def parse_sort_input(property)
+      number_like = %r!\A\s*-?(?:\d+\.?\d*|\.\d+)\s*\Z!
+      return property.to_f if property =~ number_like
```
pathawks: 😍

greenkg:
@pathawks I appreciate all the feedback though!

I have updated, with the following changes:

* I get an error "dynamic constant assignment" for using NUMBER_LIKE = %r!\A\s*-?(?:\d+\.?\d*|\.\d+)\s*\Z! inside of the method, so I changed to non-constant number_like.
* The input "4." matches the regex but raises an exception with Float("4.") So I am still using property.to_f still instead of Float(property)

pathawks:
> I get an error "dynamic constant assignment"

We can make this static constant by hoisting it out of the method, but we can do that in a separate PR.

> So I am still using property.to_f

_Sweet!_ Thanks for catching that.

DirtyF:
@greenkg When you feel you're done and this PR is ready to be reviewed, please update the title of this PR and remore `WIP: `. Thanks.

greenkg:
@ashmaroli @DirtyF @pathawks Thank you all for making it a really great experience :clinking_glasses:

pathawks:
What will happen if some items are numbers and some items are strings?

greenkg:
Shoot, well if I write a test for it, it raises an error comparison of Array with Array failed.

Looks like it runs into trouble in sort_input here
```ruby
apple_property <=> orange_property
```
Let me know if I should just go ahead and write a fix as I've thought of it, and then discuss issues after. But what I'm thinking of is as follows:

Replace the <=> line with
```ruby
compare(apple_property, orange_property)
```

Write an additional method, compare, to check if their classes are different before applying <=>.

As of right now, I think we know that the inputs will always be either floats or strings, therefore if they're different, the float should go first. It feels a bit wrong to me though to write code that says something like "if x.class == Float, put it first"?

pathawks:
```ruby
apple_property <=> orange_property || apple_property.to_s <=> orange_property.to_s
```

?

pathawks:
We should definitely have a test for the mixed strings and numbers case.

greenkg:
Ah, I tried that  (although not so succinctly) but ran into this result:
"1" "1abc" "2"
rather than
"1" "2" "1abc" which would be what I would consider the "right" sorted order

Although trying to think about use cases... if someone has a property of "weight" and they've got "1" "1lb" and "2" then that could be okay . . . but would really depend on their data

pathawks:
To be honest, I’m not sure what the proper behavior is in the case of `"1"` `"1abc"` `"2"`. This seems like an editorial decision, and not everybody will agree.

Given that we are making this the default, and only, sorting behavior, I think it is very important that we do what users will expect.

`"1"` `"1abc"` `"2"` is the order that is produced currently. If we are going to change that for all users, I think we need a very clear and compelling reason to do so.

It feels intuitive that `"9"` come before `"10"`, which is what this PR aims to fix.

pathawks:
Also note that, either way, `"100 lbs"` will be sorted _before_ `"99 lbs"`.

greenkg:
Okay--I really dropped the ball here (moved countries and started a new job!)
Jumping back in to make this change, but getting rubocop error that the line:

```ruby
apple_property <=> orange_property || apple_property.to_s <=> orange_property.to_s
```

Is too long (by 4 characters).

My first thought would be to change the variable names from "apple" and "orange" to something shorter like "a" and "b". Is that an acceptable solution or is there a more elegant ruby way?

mattr-:
I feel like switching it to `a` and `b` would be fine.

PR #6611, opened 2017-12-08
Disallow '#p' and '#puts'

blueseagles2002:
Disallow '#p' and '#puts'


ashmaroli:
I don't think @parkr ever intended to use this code.. it was committed over a month ago..

parkr:
Oh I definitely did want to see this merged at some point! But yeah it was more for debugging at the time — I had an errant call to p.

ashmaroli:
> see this merged at some point!

Is that time now..?

PR #6615, opened 2017-12-08
Write a Rubocop Cop to ensure no `#p` or `#puts` calls get committed to master.

parkr:
Using `puts` or `p` should be disallowed, in favor of always using `Jekyll.logger`. `puts` and `p` are useful for debugging, but should never land on the `master` branch.

In order to merge this, we need to fix a few instances of `puts` and `p`!


pathawks:
I love this conceptually, and presumably this would also become enabled in plugins that inherit Jekyll’s Rubocop config.

(Won’t have time to do an in-depth code review today, but I am a +1)

parkr:
> I love this conceptually, and presumably this would also become enabled in plugins that inherit Jekyll’s Rubocop config.

Oh, unclear actually. I think we'd have to add this directory to `s.files` in the gemspec!

ashmaroli:
> I think we'd have to add this directory to s.files in the gemspec!

:+1:

DirtyF:
Replaced remaining `p` and `puts' with `jekyll.logger`

lib/jekyll/commands/help.rb
```diff
@@ -12,9 +12,9 @@ def init_with_program(prog)
             c.action do |args, _|
               cmd = (args.first || "").to_sym
               if args.empty?
-                puts prog
+                Jekyll.logger.info prog
```
parkr: We might have to call `prog.to_s` here.

lib/jekyll/commands/help.rb
```diff
@@ -12,9 +12,9 @@ def init_with_program(prog)
...
               elsif prog.has_command? cmd
-                puts prog.commands[cmd]
+                Jekyll.logger.info prog.commands[cmd]
```
parkr: We might have to call `to_s` on `prog.commands[cmd]` here.

rubocop/jekyll/no_p_allowed.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require "rubocop"
+
+module RuboCop
+  module Cop
+    module Jekyll
+      class NoPAllowed < Cop
+        MSG = "Avoid using p to print things".freeze
```
parkr: Can you use the same message as in `no_puts_allowed` that suggests Jekyll.logger instead?

test/test_filters.rb
```diff
@@ -779,7 +779,7 @@ def to_liquid
       should "include the size of each grouping" do
         grouping = @filter.group_by(@filter.site.pages, "layout")
         grouping.each do |g|
-          p g
+          Jekyll.logger.info g
```
parkr: This should be just removed. It's debugging.

DirtyF:
added a `puts` in a plugin and tested this branch

output:

```
lib/jekyll-seo-tag/json_ld.rb:26:9: C: Jekyll/NoPutsAllowed: Avoid using puts to print things. Use Jekyll.logger instead.
        puts "Jekyll::SeoTag::JSONLD is deprecated"
        ^^^^
```

works fine!

lib/jekyll/commands/help.rb
DirtyF: done

lib/jekyll/commands/help.rb
DirtyF: done

lib/jekyll/commands/help.rb
ashmaroli: LogAdapter converts all parameters to strings..
https://github.com/jekyll/jekyll/blob/1691685c775f2721c1984f6632dee0fc971dfcf9/lib/jekyll/log_adapter.rb#L114-L122

ashmaroli:
this needs some tests.. especially one for the following scenario:
```ruby
File.open("foo", "wb") do |f|
  f.puts heredoc
end
```

DirtyF:
@jekyllbot: merge +dev

PR #6617, opened 2017-12-09
Default time zone depends upon server

mittalyashu:
Here's the discussion about the time zone: https://talk.jekyllrb.com/t/post-doesnt-show-up-according-to-theirs-date

### DIFF
--- a/docs/_docs/configuration.md
+++ b/docs/_docs/configuration.md
@@ -108,7 +108,7 @@ class="flag">flags</code> (specified on the command-line) that control them.
             <a href="https://en.wikipedia.org/wiki/Tz_database">IANA Time Zone
             Database</a> is valid, e.g. <code>America/New_York</code>. A list of all
             available values can be found <a href="https://en.wikipedia.org/wiki/List_of_tz_database_time_zones">
-            here</a>. The default is the server time zone, where your Jekyll website is hosted.
+            here</a>. When serving on a local machine, the default time zone is set by your operating system. But when served on a remote host/server, the default time zone depends on the server's setting or location.
         </p>
       </td>
       <td class="align-center">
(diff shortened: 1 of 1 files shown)


docs/_docs/configuration.md
```diff
@@ -108,7 +108,7 @@ class="flag">flags</code> (specified on the command-line) that control them.
             <a href="https://en.wikipedia.org/wiki/Tz_database">IANA Time Zone
             Database</a> is valid, e.g. <code>America/New_York</code>. A list of all
             available values can be found <a href="https://en.wikipedia.org/wiki/List_of_tz_database_time_zones">
-            here</a>. The default is the local time zone, as set by your operating system.
```
ashmaroli: This information is not wrong. So you should not be *changing* it.. instead:
```
When serving locally, the default is the local time zone, as set by your operating system.
But when served on a remote host, the default time zone depends on the server's setting
or location.
```

mittalyashu: Yup, you are right @ashmaroli. What about this?

> When serving on a local machine, the default time zone is set by your operating system. But when served on a remote host/server, the default time zone depends on the server's setting or location.

ashmaroli: fine as well..

mittalyashu:
@DirtyF. Now you can merge the pull request.

docs/_docs/configuration.md
mittalyashu: Great

DirtyF:
We should explain the potential impact of a different location:

`The local time zone is set by the operating system and can differ between your local machine and the remote server depending on its location.`

mittalyashu:
What do you mean by potential impact?

and you have mentioned this 
> time zone is set by the operating system and can differ between your local machine and the remote server depending on its location.

Timezone cannot always be different from the operating system and remote server. Maybe a person is living in the same location where server is location *(this could be a possibility)*.

But this line explains
> When serving on a local machine, the default time zone is set by your operating system. But when served on a remote host/server, the default time zone depends on the server's setting or location.

If you have not set any timezone then what will be the timezone on your local machine and on the server.

DirtyF:
@jekyllbot: merge +docs

DirtyF:
@mittalyashu Thanks for this first contribution.

PR #6621, opened 2017-12-10
Output correct relative paths to Jekyll's logger

ashmaroli:
Fix to ensure that `Jekyll.logger` outputs proper relative paths to documents and their excerpts.

Changes only logger messages.
No API changes. Backwards-compatible alternative to #6597

### DIFF
--- a/lib/jekyll/renderer.rb
+++ b/lib/jekyll/renderer.rb
@@ -49,14 +49,14 @@ module Jekyll
     #
     # Returns String rendered document output
     def run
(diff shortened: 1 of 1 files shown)


ashmaroli:
> Let’s just define `Excerpt#relative_path` like we do `Excerpt#path`.

That's what I proposed in #6597 but it fails our test suite and changes a public_method so its slated for `4.0`. This is meant for `3.x` series..

parkr:
I don't like this at all... I think `relative_path` can change its output just fine without a breaking change release.

PR #6623, opened 2017-12-11
Update "Requirements" for Ruby version

pnsk:
When I tried to install Jekyll into my local environment using the `gem install jekyll bundler` command, the error was shown and the install failed:

```
ERROR:  Error installing jekyll:
	ruby_dep requires Ruby version >= 2.2.5, ~> 2.2.
```

So the requirement for the Ruby version is 2.2.5 or above.


pnsk:
@DirtyF Thanks ❤️

DirtyF:
@jekyllbot: merge +docs

PR #6627, opened 2017-12-13
Use the first entry in a list of output extensions instead of conditionally selecting the last / second-last entry

ashmaroli:
- always select the first entry in the array of output_exts
  - eliminates an if block => minor optimization

### DIFF
--- a/lib/jekyll/renderer.rb
+++ b/lib/jekyll/renderer.rb
@@ -42,7 +42,7 @@ module Jekyll
     #
     # Returns String the output extname including the leading period.
     def output_ext
-      @output_ext ||= (permalink_ext || converter_output_ext)
+      @output_ext ||= (permalink_ext || output_exts[0])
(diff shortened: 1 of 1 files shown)


DirtyF:
What problem does this solve?

ashmaroli:
> What problem does this solve?

Its not a `fix` per se.. Just saw this and wondered why not omit the conditional and simply select the first extension since the `output_exts` array has just either one or two entries..
```
[".html", ".markdown"]
[".html", ".md"]
[".css", ".scss"]
```
The existing tests did not resolve my confusion either..
I figured opening a PR would either have the change accepted.., or rejected with appropriate logical reasons (which would answer my doubts regarding the code flow..)

parkr:
I'm surprised this doesn't break a test! This was part of `Convertible` since the dawn of time. I worry about changing it!

PR #6630, opened 2017-12-14
Add a script to profile methods in Jekyll::Site & Jekyll::Renderer

ashmaroli:
Introduces a BASH script to profile various methods in `Jekyll::Site` and `Jekyll::Renderer` via the [`method_profiler`](https://github.com/change/method_profiler) gem, using the official docs site folder as test source.


DirtyF:
1. Please explain _why_ you want to add that script (we already have a benchmark suite, what does this bring to the table?
2. Please document _how_ it's supposed to be used, when, etc.

Thanks.

pathawks:
Wouldn’t it be better to add this as a Rake task, instead of a Bash script that evals a string of Ruby code?

ashmaroli:
> Please explain *why* you want to add that script..

First, a brief intro:
This script profiles all methods (class methods and instance methods) in the said classes (`Jekyll::Site` and `Jekyll::Renderer`) during a Jekyll build session, and prints the collected data as a table.
Why I want to add this?
This is pure-Ruby and **runs fine on Windows with BASH support**.
I haven't been able to run `stackprof` and `rubyprof` without encountering issues with native extensions. So *I can't say how inferior this script is in comparison..*


> Wouldn’t it be better to add this as a Rake task..?

I saw a similar script ([`rubyprof`](https://github.com/jekyll/jekyll/blob/d1440b9738eedb11208d3e66a128263e705b6fc9/script/rubyprof)) and thought that's the preferred way..
I like the idea of this being a Rake Task.. :+1:

Crunch09:
`method_profiler` hasn't had a release for almost 5 years. Are there any alternatives which are still maintained?
Regarding `stackprof` and `ruby-prof`: At least `ruby-prof` should be able to be also run with Windows afaik.

ashmaroli:
yea.. I managed to get the `ruby-prof` gem install successfully on Windows by directly installing it
```
gem install ruby-prof
```
... and consequently ran the `rubyprof` script as well.. But had various issues..

PR #6635, opened 2017-12-16
Correct custom inspect string for Jekyll::Page

ashmaroli:
Scope Resolution Operator should be `::` instead of `:` which is for Symbols..


DirtyF:
@jekyllbot: merge +bug

PR #6640, opened 2017-12-19
Allow Doctor to check for outdated dependencies [WIP]

ashmaroli:
Let users know that their site's dependencies are up-to-date or not when they're connected to the Internet and runs `jekyll doctor`

### DIFF
--- a/lib/jekyll/commands/doctor.rb
+++ b/lib/jekyll/commands/doctor.rb
@@ -46,15 +46,20 @@ module Jekyll
         end
 
         def outdated_dependency_check(site)
-          return true unless testable?(site)
+          return true unless can_test_dependency_health?(site)
 
           Dir.chdir(site.source) do
             exe = Gem.bin_path("bundler", "bundle")
             output = Jekyll::Utils::Exec.run("ruby", exe, "outdated", "--parseable")[-1]
             output.to_s.each_line do |line|
-              Jekyll.logger.warn "Outdated:", line
+              Jekyll.logger.warn "Outdated gem:", line
             end
-            output.empty?
+
+            return true if output.empty?
+            Jekyll.logger.info "", "Found outdated site dependencies!".magenta
+            Jekyll.logger.info "", "Run `bundle update` to use latest " \
+                                   "versions of those gems in your site".magenta
+            false
           end
         end
 
@@ -164,7 +169,7 @@ module Jekyll
           false
         end
 
-        def testable?(site)
+        def can_test_dependency_health?(site)
           Jekyll::Utils::Internet.connected? &&
             File.exist?(site.in_source_dir("Gemfile")) &&
             bundler_installed?
--- a/lib/jekyll/commands/doctor.rb
+++ b/lib/jekyll/commands/doctor.rb
@@ -48,16 +48,13 @@ module Jekyll
         def outdated_dependency_check(site)
           return true unless testable?(site)
 
-          oudated_dep_regex = %r!\A\s*\*\s*!
-
           Dir.chdir(site.source) do
             exe = Gem.bin_path("bundler", "bundle")
-            _process, output = Jekyll::Utils::Exec.run("ruby", exe, "outdated")
+            output = Jekyll::Utils::Exec.run("ruby", exe, "outdated", "--parseable")[-1]
             output.to_s.each_line do |line|
-              next unless line =~ oudated_dep_regex
-              Jekyll.logger.warn "Outdated:", line.sub(oudated_dep_regex, "")
+              Jekyll.logger.warn "Outdated:", line
             end
-            output.end_with? "Bundle up to date!"
+            output.empty?
           end
         end
 
--- a/lib/jekyll/commands/doctor.rb
+++ b/lib/jekyll/commands/doctor.rb
@@ -50,12 +50,17 @@ module Jekyll
 
           Dir.chdir(site.source) do
             exe = Gem.bin_path("bundler", "bundle")
-            output = Jekyll::Utils::Exec.run("ruby", exe, "outdated", "--parseable")[-1]
+            process, output = Jekyll::Utils::Exec.run(
+              "ruby", exe, "outdated", "--parseable"
+            )
+
+            raise SystemExit unless process.success?
+            return true if output.empty?
+
             output.to_s.each_line do |line|
               Jekyll.logger.warn "Outdated gem:", line
             end
 
-            return true if output.empty?
             Jekyll.logger.info "", "Found outdated site dependencies!".magenta
             Jekyll.logger.info "", "Run `bundle update` to use latest " \
                                    "versions of those gems in your site".magenta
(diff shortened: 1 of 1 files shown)


lib/jekyll/commands/doctor.rb
```diff
@@ -149,6 +166,19 @@ def url_absolute(url)
               "check the value of `url` in your config file."
           false
         end
+
+        def testable?(site)
```
parkr: This is only used by `outdated_dependency_check` but the name indicates that it's globally applicable. Perhaps `can_test_plugin_health?`?

lib/jekyll/commands/doctor.rb
```diff
@@ -44,6 +45,22 @@ def healthy?(site)
...
+          return true unless testable?(site)
+
+          oudated_dep_regex = %r!\A\s*\*\s*!
+
+          Dir.chdir(site.source) do
+            exe = Gem.bin_path("bundler", "bundle")
+            _process, output = Jekyll::Utils::Exec.run("ruby", exe, "outdated")
+            output.to_s.each_line do |line|
+              next unless line =~ oudated_dep_regex
+              Jekyll.logger.warn "Outdated:", line.sub(oudated_dep_regex, "")
```
parkr: `Outdated plugin:`, perhaps?

ashmaroli: The patch is not testing just *plugins* per se.. it tests every *gem* the current site **depends** on..
*plugins / theme-gem / dependencies-of-jekyll / dependencies-of-plugins-or-theme-gem*

parkr: Hm. I’m not sure that’s useful here. I think just plugins is better, since Jekyll shouldn’t care about rake or simplecov or any other dependency. Filtering with `site.theme` and `site.plugins` would be preferable I think because it’s more relevant to Jekyll’s interests.

ashmaroli: By dependencies.., I was referring to the [runtime dependencies](https://github.com/jekyll/jekyll/blob/a3351cff4f7756251f7670de6cb781c99b25fde4/jekyll.gemspec#L33-L45).. *stuff that matter when our users run `bundle exec jekyll --`*..

lib/jekyll/commands/doctor.rb
```diff
@@ -44,6 +45,24 @@ def healthy?(site)
           ].all?
         end
 
+        def outdated_dependency_check(site)
+          return true unless can_test_dependency_health?(site)
+
+          Dir.chdir(site.source) do
+            exe = Gem.bin_path("bundler", "bundle")
+            output = Jekyll::Utils::Exec.run("ruby", exe, "outdated", "--parseable")[-1]
```
parkr: This [-1] is confusing. Can we just do what you did before with the splat and the underscore prefixed variable?

You should be checking whether bundler successfully exited anyway, I would presume.

lib/jekyll/commands/doctor.rb
```diff
@@ -44,6 +45,24 @@ def healthy?(site)
...
+            output.to_s.each_line do |line|
+              Jekyll.logger.warn "Outdated gem:", line
+            end
+
+            return true if output.empty?
```
parkr: Why not do this first, before doing output.to_s.each_line?

lib/jekyll/commands/doctor.rb
```diff
@@ -50,12 +50,17 @@ def outdated_dependency_check(site)
 
           Dir.chdir(site.source) do
             exe = Gem.bin_path("bundler", "bundle")
-            output = Jekyll::Utils::Exec.run("ruby", exe, "outdated", "--parseable")[-1]
+            process, output = Jekyll::Utils::Exec.run(
+              "ruby", exe, "outdated", "--parseable"
+            )
+
+            raise SystemExit unless process.success?
```
parkr: We should show an error...

ashmaroli: hmm.. what would the message be?
```
Jekyll 3.8.0 | SystemExit: Dependency health check failed
```
?

localheinz:
What do you think about adding options to suppress / enable different checks?

Let's assume at least the following (hypothetical) scenario:

- I have my site deployed via Travis CI
- I run `bundle exec jekyll doctor` as part of the build

Then

- I plan to go on a trip around the world
- I have a ton of articles in the pipeline with future date set, and then use the https://docs.travis-ci.com/user/cron-jobs/ feature to build and deploy the site regularly
- dependencies become outdated, and the site fails to deploy

It's not quite the same, but in the PHP world, when using `composer validate` to validate `composer.json` (the equivalent of `Gemfile`, if you will), there are options to enable/suppress different checks, see https://getcomposer.org/doc/03-cli.md#validate.

ashmaroli:
If you're deploying via Travis, simply add an intelligent `bundle update` rescue to the script sequence.. That will eliminate a chance for failed builds due to outdated dependencies..

lib/jekyll/commands/doctor.rb
```diff
@@ -44,6 +45,29 @@ def healthy?(site)
...
+            return true if output.empty?
+
+            output.to_s.each_line do |line|
+              Jekyll.logger.warn "Outdated gem:", line
+            end
+
+            Jekyll.logger.info "", "Found outdated site dependencies!".magenta
+            Jekyll.logger.info "", "Run `bundle update` to use latest " \
+                                   "versions of those gems in your site".magenta
```
DirtyF: I like the idea of encouraging users to upgrade.

Now if I run `bundle update` I don't necessarily the latest versions installed, depends also on the version specified in the `Gemfile`, and `.gemspec`, right?

Here's Jekyll current output if I check for outdated gems:

```sh
bundle outdated --parseable

activesupport (newest 5.1.4, installed 4.2.10)
coffee-script-source (newest 1.12.2, installed 1.11.1)
rainbow (newest 3.0.0, installed 2.2.2)
rdoc (newest 6.0.1, installed 5.1.0, requested ~> 5.0)
rubocop (newest 0.52.1, installed 0.51.0, requested ~> 0.51.0)
shoulda-matchers (newest 3.1.2, installed 2.8.0)
simplecov (newest 0.15.1, installed 0.13.0)
```

So here even after running `bundler update`, `jekyll doctor`would still encourage to run `bundle update`, it could be confusing don't you think? 😕

ashmaroli: ohh.. I did not expect Bundler to output when the `Gemfile` *locks* to a certain version series..
hmm.. thanks for pointing this out.. :+1:

DirtyF: `jekyll doctor` could compare the result of running `bundle update && bundle outdated`--parseable` vs `bundle outdated --parseable` and only warn users if the output differs?

pathawks: But don't run `bundle update` because that might result unintentionally upgrading

pathawks:
Aside from conflicts, is this good to merge?

DirtyF:
closing this as this does not resolve an issue, and bundler has already a command for that. it's not jekyll's job to manage dependencies.

PR #6641, opened 2017-12-20
Improve docs styling for code to be run in shell

ashmaroli:
### Current design:
jekyllrb com_docs_quickstart

### With this patch..:
localhost_4000_docs_quickstart

### DIFF
--- a/docs/_sass/_style.scss
+++ b/docs/_sass/_style.scss
@@ -1131,27 +1131,23 @@ code.output {
     display: table;
     padding: 8px;
     width: 100%;
-    color: #888;
+    padding: 5px 0;
+    font: 400 16px/24px 'Helvetica Neue', Helvetica, Arial, sans-serif;
+    color: #444;
     text-shadow: 0 1px 0 rgba(255,255,255,.5);
-    font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
-    font-size: 15px;
-    font-weight: 600;
-    content: "TERMINAL";
+    background-color: #f7f7f7;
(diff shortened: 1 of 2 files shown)


docs/_sass/_style.scss
```diff
@@ -1136,11 +1137,26 @@ code.output {
     line-height: 1.25;
     font-weight: 700;
     content: "TERMINAL";
-    background: #7a7a7a;
+    text-align: center;
+    background: linear-gradient(#b3b3b3, #ededed);
```
pup: imo this could be a solid color, to be consistent with the current MacOS styling

ashmaroli: Agreed, but I wasn't shooting for an exact macOS style.. just a close replica..
Moreover we do use gradients elsewhere on `jekyllrb.com`.. ..

mattr-: Could we see it with a solid color instead? I'm leaning that way as well.

mattr-:
Nice work on this. I like that the terminal background is a bit darker which gives more contrast to the white text. I like that the title bar is a lighter color too.

I don't agree with the color change to the comments because they serve as instructions for what to do and now they're in danger of getting lost due to the lack of contrast. Could you tweak it such that they're still gray but just a lighter shade? I don't think it's necessary to go back to the blue but we definitely don't want to lose them.

ashmaroli:
Looking around.., we already have a style similar to an OS chrome..

jekyllrb com_

pup:
@ashmaroli okay, that's fair then

DirtyF:
> Looking around.., we already have a style similar to an OS chrome..

👍 Let's just use the existing style then, we don't need the extra macOS buttons here.

ashmaroli:
To use the existing style would need to modify generated markup as well.. instead incorporated that style into this patch..
...kept the mac buttons since it immediately signals an app window chrome..
...also used a lighter-gray for the comment strings..

localhost_4000_docs_quickstart_mod

DirtyF:
I tested  and we can match the existing style without modifying the markup:

terminal

ashmaroli:
@DirtyF  Feel free to update the branch as you see fit.. :slightly_smiling_face:

DirtyF:
@jekyllbot: merge +site

PR #6541, opened 2017-11-11
Dependencies: upgrade to toml 0.2.0

jekyllbot:
Fix #6540

```text
Dependencies: upgrade to toml 0.2.0
```

PR #6652, opened 2017-12-27
Switch to an actively-maintained TOML parser.

jashank:
We have an ancient version of the `toml` gem which appears to have been abandoned for years and isn't compliant with TOML 0.4.0, the current state of the art.

The `tomlrb` and `toml-rb` gems are both (more) actively maintained and kept up-to-date; switching to either gives us TOML 0.4.0 support.  This patch switches us to `tomlrb`.

(See also: #6540 #6541)


jashank:
Whoops! I didn't notice the tests; amended.

pup:
@parkr okay!

@jekyllbot: merge +minor

PR #6654, opened 2017-12-27
Add `jekyll-html` to plugins.

kacperduras:


DirtyF:
You know you can add HTML in Markdown right? (I must be missing the use case here)

DirtyF:
@jekyllbot: merge +docs

PR #6676, opened 2018-01-06
Display latest version in header

jekyllbot:
PR automatically created for @DirtyF.

```text
Display latest version in header
```

PR #6677, opened 2018-01-06
Update version in `config.yml` via YAML load / dump

Crunch09:
As discussed in #6676 , I've already ran the command once, so we can see what changes in the `config.yml` file when running this command (as it normalizes the YAML output)

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -1,12 +1,25 @@
 ---
-sass:
-  style: compressed
 version: 3.7.0
+name: Jekyll • Simple, blog-aware, static sites
+description: Transform your plain text into static websites and blogs
+url: https://jekyllrb.com
+repository: https://github.com/jekyll/jekyll
+timezone: America/Los_Angeles
+twitter:
+  username: jekyllrb
+logo: "/img/logo-2x.png"
 gauges_id: 503c5af6613f5d0f19000027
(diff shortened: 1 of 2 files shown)


DirtyF:
@Crunch09 nice extra feature 😃

ashmaroli:
@Crunch09 anyway to get the indentation back to two spaces?

Crunch09:
@ashmaroli Good point, there should be an option i think! I'll check

DirtyF:
It also remove any comments left in the file 😕

ashmaroli:
The docs built fine locally so i guess the indentation is not that much of an issue for the list items (**T.I.L.**)

Crunch09:
Yeah, if we want to have comments in there we need to stick with the regex i'm afraid :(

@ashmaroli Indentation is already set to 2 by default. I assume you meant the `plugins` and `exclude` arrays? I don't have enough YAML knowledge but i think there is no need for spaces so the YAML implementation doesn't add them.

ashmaroli:
Yes, I was referring to them.. but since everything works fine, guess its okay then.. just that the config file looks a bit crowded to me now..

DirtyF:
This file is rarely edited so I guess it's OK.

@jekyllbot: merge +site

PR #6679, opened 2018-01-07
Add some tests for custom collections directory

ashmaroli:
Replaces #6678 
Cucumber tests for custom `collections_dir` configuration


pathawks:
So, when there is a custom `collections_dir`, the `_posts` directory needs to be moved into that directory instead of in the top level?

ashmaroli:
> So, when there is a custom `collections_dir`, the `_posts` directory needs to be moved into that directory instead of in the top level?

Yes. `_posts` is a collection too.. just that it is a special collection..
  - always rendered &mdash; you can't hide them
  - has a different naming pattern (`YYYY-MM-DD-title.markdown`)
  - has `categories` and `tags` associated with it
  - has `_drafts` associated with it

localheinz:
@ashmaroli 

Ha, then this isn’t actually a bug, but rather a documentation issue? 

> You can optionally specify a directory to store all your collections in the same place with `collections_dir: my_collections`
Then Jekyll will look in `my_collections/_books` for the `books` collection, and in `my_collections/_recipes` for the `recipes` collection.

It wasn’t obvious to me that `_posts` should be moved into the collections directory as well.

ashmaroli:
> but rather a documentation issue?

Yep. You're welcome to submit a PR for improving the documentation..

localheinz:
@ashmaroli 

Can confirm that this works for me!

ashmaroli:
I take back most of what I have said over the span of last 8 hours.. :smile: @localheinz was right! There is a bug in the code in fact..!!

My Apologies @localheinz !!

Documents within `<source>/_posts` do get processed into the `_site`. They get rendered into **`_site/_posts/`** and are accessible via liquid construct `{{ site.posts }}` as well..

localheinz:
@ashmaroli 

Ha, nice, wasn't quite sure what I was doing wrong.

ashmaroli:
We were both right and wrong at the same time.. :smile: 
I kept testing with `_posts` in the `collection_dir` while you kept building with `_posts` at the root.. It was only when I combined both scenarios did I realize the true gravity of the situation.. :wink:

ashmaroli:
closing in favor of #6685

PR #6681, opened 2018-01-07
Allow Jekyll Doctor to detect stray posts dir

ashmaroli:
If one defines a custom location for their collections, `_posts` directory should not exist at the root of the source directory

### DIFF
--- a/lib/jekyll/commands/doctor.rb
+++ b/lib/jekyll/commands/doctor.rb
@@ -49,9 +49,11 @@ module Jekyll
           return true if site.config["collections_dir"].empty?
           posts_at_root = site.in_source_dir("_posts")
           return true unless File.directory?(posts_at_root)
-          Jekyll.logger.warn "Warning:", "Contents of #{posts_at_root} will not be " \
-                             "processed since you have specified a custom directory " \
-                             "to house all collections."
+          Jekyll.logger.warn "Warning:",
+            "Detected '_posts' directory outside custom `collections_dir`!"
+          Jekyll.logger.warn "",
+            "Please move '#{posts_at_root}' into the custom directory at " \
+            "'#{site.in_source_dir(site.config["collections_dir"])}'"
           false
         end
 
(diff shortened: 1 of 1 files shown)


lib/jekyll/commands/doctor.rb
```diff
@@ -41,9 +41,20 @@ def healthy?(site)
...
           ].all?
         end
 
+        def properly_gathered_posts?(site)
+          return true if site.config["collections_dir"].empty?
+          posts_at_root = site.in_source_dir("_posts")
+          return true unless File.directory?(posts_at_root)
+          Jekyll.logger.warn "Warning:", "Contents of #{posts_at_root} will not be " \
+                             "processed since you have specified a custom directory " \
+                             "to house all collections."
```
DirtyF: Could we rater display the awaited action from the user?

Warning: Custom `collection_dir` detected : move the `_posts` directory inside  the `my_collections` directory.

localheinz:
@ashmaroli 

This is awesome, and I'm sure much better suited to save people from headaches, hehe!

ashmaroli:
> I'm sure much better suited to save people from headaches

@localheinz only if they make it a point to run `jekyll doctor` when something's not right.. :smile: 
Besides, this won't reach our users until next minor release.. Documentation updates are go live instantly.. :smiley:

DirtyF:
What about an option  "jekyll build --check" that basically run `jekyll doctor` before build?

ashmaroli:
> What about an option `jekyll build --check` that basically run `jekyll doctor` before build?

Comparing https://github.com/jekyll/jekyll/blob/f77d704a1b97ac43684df85e92a8738412d1725a/lib/jekyll/commands/doctor.rb#L24-L28
and https://github.com/jekyll/jekyll/blob/f77d704a1b97ac43684df85e92a8738412d1725a/lib/jekyll/site.rb#L69-L77
the first three steps of the "process" is the same..

So it has to be more of a split-action procedure instead of a before-and-after..
Nice idea nevertheless.. :+1:

oe:
@jekyllbot: merge +minor

PR #6685, opened 2018-01-08
Refactor collections_dir feature for consistency

ashmaroli:
I believe this removes bugs with the recently released `collection_dir` feature reported via jekyll/jekyll-feed#202 and confirmed via #6679 

Basically with this patch,
  - Site runs as usual if the `collections_dir` setting is not customized.
  - When the `collections_dir` setting is customized, Jekyll looks for all collections *within* that directory only:
    - Posts and drafts at the root will no longer be read / processed &mdash; you need to move them inside the `collections_dir`
    - Relative path of all "documents" will be with-respect-to that directory. e.g. If the directory is named `my_collections`, then the relative_path for `./my_collections/_puppies/rover.md` is simply `_puppies/rover.md`
    - Ergo, **relative_path for a document does not change whether the collection is at the site-root or at the root of the `collection_dir`**

### DIFF
--- a/features/collections_dir.feature
+++ b/features/collections_dir.feature
@@ -1,0 +1,157 @@
+Feature: Collections Directory
+  As a hacker who likes to structure content without clutter
+  I want to be able to organize my collections under a single directory
+  And render them from there
+
+  Scenario: Custom collections_dir containing only posts
+    And I have a collections/_posts directory
+    And I have the following post within the "collections" directory:
+      | title         | date       | content         |
+      | Gathered Post | 2009-03-27 | Random Content. |
+    And I have a "_config.yml" file with content:
+    """
+    collections_dir: collections
+    """
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "Random Content." in "_site/2009/03/27/gathered-post.html"
+
+  Scenario: Rendered collection in custom collections_dir also containing posts
+    Given I have a collections/_puppies directory
+    And I have the following document under the "puppies" collection within the "collections" directory:
+      | title  | date       | content            |
+      | Rover  | 2007-12-31 | content for Rover. |
+    And I have a collections/_posts directory
+    And I have the following post within the "collections" directory:
+      | title         | date       | content         |
+      | Gathered Post | 2009-03-27 | Random Content. |
+    And I have a "_config.yml" file with content:
+    """
+    collections:
+      puppies:
+        output: true
+
+    collections_dir: collections
+    """
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And the "_site/puppies/rover.html" file should exist
+    And I should see "Random Content." in "_site/2009/03/27/gathered-post.html"
+
+  Scenario: Rendered collection in custom collections_dir with posts at the site root
+    Given I have a collections/_puppies directory
+    And I have the following document under the "puppies" collection within the "collections" directory:
+      | title  | date       | content            |
+      | Rover  | 2007-12-31 | content for Rover. |
+    And I have a _posts directory
+    And I have the following post:
+      | title        | date       | content         |
+      | Post At Root | 2009-03-27 | Random Content. |
+    And I have a "_config.yml" file with content:
+    """
+    collections:
+      puppies:
+        output: true
+
+    collections_dir: collections
+    """
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And the "_site/puppies/rover.html" file should exist
+    And the "_site/post-at-root.html" file should not exist
+    And the _site/_posts directory should not exist
+
+  Scenario: Rendered collection in custom collections_dir also containing drafts
+    Given I have a collections/_puppies directory
+    And I have the following document under the "puppies" collection within the "collections" directory:
+      | title  | date       | content            |
+      | Rover  | 2007-12-31 | content for Rover. |
+    And I have a collections/_drafts directory
+    And I have the following draft within the "collections" directory:
+      | title          | date       | content         |
+      | Gathered Draft | 2009-03-27 | Random Content. |
+    And I have a "_config.yml" file with content:
+    """
+    collections:
+      puppies:
+        output: true
+
+    collections_dir: collections
+    """
+    When I run jekyll build --drafts
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And the "_site/puppies/rover.html" file should exist
+    And I should see "Random Content." in "_site/2009/03/27/gathered-draft.html"
+    And the _site/collections directory should not exist
+
+  Scenario: Rendered collection in custom collections_dir with drafts at the site root
+    Given I have a collections/_puppies directory
+    And I have the following document under the "puppies" collection within the "collections" directory:
+      | title  | date       | content            |
+      | Rover  | 2007-12-31 | content for Rover. |
+    And I have a _drafts directory
+    And I have the following draft:
+      | title         | date       | content         |
+      | Draft At Root | 2009-03-27 | Random Content. |
+    And I have a "_config.yml" file with content:
+    """
+    collections:
+      puppies:
+        output: true
+
+    collections_dir: collections
+    """
+    When I run jekyll build --drafts
+    Then I should get a zero exit status
+    And the _site directory should exist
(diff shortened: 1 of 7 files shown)


localheinz:
@ashmaroli 

As a note, what I tripped over initially was that I assumed that a collections directory would be special directory (similar to `_data`, `_includes`, etc), so after upgrading from 3.6.2 to 3.7.0, my initial assumption was that with a configuration of

```yml
collections_dir: _collections
```

and a directory layout similar to

```
.
├── _collections
│   ├── books
│   └── puppies
├── _data
├── _includes
└── _layouts
```

everything would be neatly arranged.

Note:

* collections directory prefixed with `_` (so *special* directories are all in order)
* directories for collections within collections directory without `_` prefix

Just felt like throwing this in, as an FYI.

features/collections_dir.feature
```diff
@@ -0,0 +1,157 @@
...
+      puppies:
+        output: true
+
+    collections_dir: collections
+    """
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And the "_site/puppies/rover.html" file should exist
+    And the "_site/post-at-root.html" file should not exist
```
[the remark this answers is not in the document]
ashmaroli: yes.. it should.. we ought to be checking for the expected outcomes.. :+1:

features/step_definitions.rb
```diff
@@ -108,6 +122,16 @@
 
 #
 
+Given(%r!^I have the following documents? under the "(.*)" collection within the "(.*)" directory:$!) do |label, dir, table|
+  table.hashes.each do |input_hash|
+    title = slug(input_hash["title"])
+    path = File.join(dir, "_#{label}", "#{title}.md")
```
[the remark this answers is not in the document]
ashmaroli: only because that's how our default site gets created from `jekyll new` .. and I think, an existing step also has something similar set up.. not sure..

lib/jekyll/reader.rb
```diff
@@ -130,5 +131,20 @@ def get_entries(dir, subfolder)
...
+    #
+    # Determine if the directory is supposed to contain posts and drafts.
+    # If the user has defined a custom collections_dir, then attempt to read
+    # posts and drafts only from within that directory.
+    #
+    # Returns true if a custom collections_dir has been set but current directory lies
+    # outside that directory.
+    def not_within_configured_directory?(dir)
+      collections_dir = site.config["collections_dir"]
+      !collections_dir.empty? && !dir.start_with?("/#{collections_dir}")
```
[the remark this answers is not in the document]
ashmaroli: it can be done.. I kept it negative so that human can readily parse two negative expressions easily in comparison to one negative and the other positive..
```ruby
    def within_configured_directory?(dir)
      collections_dir = site.config["collections_dir"]
      !collections_dir.empty? && dir.start_with?("/#{collections_dir}")
    end
```

ashmaroli:
@localheinz I understand your suggestion completely.
While I agree that the custom collections_dir with a leading underscore would allow greater aesthetics, I'm not in favor of removing the underscore from the directory for collections for reasons below:
  - having the leading underscore will allow greater flexibility when moving the collections *en bulk* into and outside the custom container (during migration and testing across older and newer versions of Jekyll)
  - if we decide to add support for `categories` in collections, similar to hows posts have it with e.g. `movies/_posts/*.markdown`, having the leading underscore in the path would allow writing simpler code and also allow users to mentally scan for their actual collection directory in a file tree

localheinz:
@ashmaroli 

I'm totally fine with `_` and without, so all good. Thanks a lot for your efforts towards fixing these issues! 👍

ashmaroli:
> why do we add a date to the collection documents here?

Its not required at all, but existing scenarios had them, so I included them here as well.. will make it easier if future developers need to test for some other date-related parameter..

features/collections_dir.feature
```diff
@@ -0,0 +1,156 @@
...
+      puppies:
+        output: true
+
+    collections_dir: collections
+    """
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And the "_site/puppies/rover.html" file should exist
+    And the "_site/2009/03/27/post-at-root.html" file should not exist
```
parkr: Woof, this is going to screw people up. We can't enforce that the `_posts` directory is in the `collections_dir` until 4.0, if we even want to.

features/collections_dir.feature
```diff
@@ -0,0 +1,156 @@
...
+      puppies:
+        output: true
+
+    collections_dir: collections
+    """
+    When I run jekyll build --drafts
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And the "_site/puppies/rover.html" file should exist
+    And the "_site/2009/03/27/draft-at-root.html" file should not exist
```
parkr: `_drafts` can't be enforced to be in `collections_dir` until 4.0 either.

lib/jekyll/collection.rb
```diff
@@ -95,14 +95,13 @@ def filtered_entries
...
+    # The directory for this Collection, relative to the site source or the directory
+    # containing the collection.
     #
     # Returns a String containing the directory name where the collection
     #   is stored on the filesystem.
     def relative_directory
-      @relative_directory ||= Pathname.new(directory).relative_path_from(
-        Pathname.new(site.source)
-      ).to_s
+      @relative_directory ||= "_#{label}"
```
parkr: This should be relative to the site source, so if source is `.` then it should be `collections_dir/_label`.

lib/jekyll/document.rb
```diff
@@ -83,7 +83,7 @@ def draft?
     # Returns a String path which represents the relative path
     #   from the site source to this document
     def relative_path
-      @relative_path ||= Pathutil.new(path).relative_path_from(site.source).to_s
+      @relative_path ||= Pathutil.new(path).relative_path_from(focal_point).to_s
```
parkr: Why can't this be the site.source? Are you thinking the `collections_dir` needs to disappear in the output? Hm. Maybe `relative_path_from(collection.path)`?

lib/jekyll/collection.rb
ashmaroli: Things that're going to break **if `@relative_directory ||= collections_dir/_label`**:
https://github.com/jekyll/jekyll/blob/f77d704a1b97ac43684df85e92a8738412d1725a/lib/jekyll/document.rb#L32-L36
https://github.com/jekyll/jekyll/blob/f77d704a1b97ac43684df85e92a8738412d1725a/lib/jekyll/document.rb#L76-L79
https://github.com/jekyll/jekyll/blob/f77d704a1b97ac43684df85e92a8738412d1725a/lib/jekyll/document.rb#L121-L124

lib/jekyll/document.rb
ashmaroli: `collection` does not have an instance method `:path`

features/collections_dir.feature
ashmaroli: its not.. we just released this feature in `3.7.0` .. The sooner we ship this patch, the lesser no.of users gonna have issues..

lib/jekyll/document.rb
parkr: Any reason it _can't_ have that method? I don't like the `focal_point` method – seems like it's capturing some logic that should be handled somewhere else, like `path`.

ashmaroli: > Any reason...?

I did not want to **introduce a new `public_method`** in a patch-release..

DirtyF: Well now that we have introduce a potential breaking changes in a minor version, we should be above that kind of consideration here. The more `collections_dir` behavior is well thought, the better.

ashmaroli: @parkr Your call.. should I **introduce a new public_method in a patch-release..**?

parkr: I think adding Collection#path to fix this bug is a fine solution for a patch release. SemVer makes no assumptions about methods being added, as far as I can tell. Our SemVer is guided by new features (minor bump) and breaking changes (major bump).

ashmaroli: I would call "enhancing" `Collections` class with a new `public_method` as a feature.. just like removing an existing `public_method` is considered a breaking change..

That said, this is not the place for debating SemVer.. your comment above is clear enough an answer for how to proceed..
Will push a commit soon..

pup:
@jekyllbot: merge +bug

PR #6686, opened 2018-01-09
Updated Windows RubyInstaller instructions

Rarst:
1. Added up to date Ruby version information.
2. Added GitHub Pages information.
3. Added possible solutions for MSYS2 and libcurl errors.

This is what worked for me to install GitHub Pages (Jekyll) on Windows 7x64 via RubyInstaller at this time.

Chocolatey instructions are also outdated (recent versions need MSYS2, not DevKit; Nokogiri gem is a non–issue by now), but I  was hesitant to edit that section since I hadn't managed to successfully install start to finish with it.

### DIFF
--- a/docs/_docs/windows.md
+++ b/docs/_docs/windows.md
@@ -72,12 +72,28 @@ You can make sure time management is working properly by inspecting your `_posts
 
 [RubyInstaller][] is a self-contained Windows-based installer that includes the Ruby language, an execution environment, important documentation, and more.
 
-1. Download and Install a package manager version from [RubyInstaller Downloads][RubyInstaller-downloads].
-2. Install Jekyll and Bundler via a command prompt window: `gem install jekyll bundler`
-3. Check if Jekyll installed properly: `jekyll -v`
+1. Download and Install Ruby version 2.4.3 from [RubyInstaller Downloads][RubyInstaller-downloads] (2.5 is not recommended by them at the time of writing and conflicts with dependency later).
+2. Let it install [MSYS2][] or do it yourself.
+3. Install Bundler via a command prompt `gem install bundler`.
+4. Install Jekyll `gem install jekyll` _or_ GitHub Pages `gem install github-pages` if you want it (includes `jekyll`).
+5. Check if Jekyll installed properly: `jekyll -v`.
+6. Create new site: `jekyll new myblog`
+7. If you want GitHub Pages version edit generated `Gemfile` to comment out `jekyll` line and uncomment `github-pages` line.
+8. `bundle install` (in `myblog`)
+9. `bundle exec jekyll serve` (in `myblog`)
+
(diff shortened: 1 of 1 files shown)


DirtyF:
@Rarst Thanks for the feedback.

Any Window user care to review this install process?

XhmikosR:
I can, but I feel all this is futile.

Getting Jekyll to work on Windows is super easy nowadays.

1. Just install Ruby 2.4.x with the default options
2. `gem install bundler`
3. `gem install jekyll`

I can't speak for the GitHub Pages gem, and I don't think that info belongs to this repo. Nor should we cover all the possible installation methods.

Just my 2 cents.

XhmikosR:
Oh, also, I never got any of those errors this PR is supposed to fix and I have been doing this for years with numerous installations.

Rarst:
> Just install Ruby 2.4.x with the default options

Instructions currently don't point out 2.4.x version correctly, they either omit version or reference 2.2 and 2.3. Also Chocolatey installs 2.5 implicitly which conflicted with some gem later for me.

> I can't speak for the GitHub Pages gem, and I don't think that info belongs to this repo.

Instructions already refer to installing `github-pages` and in my own case I needed _that_, not just `jekyll`.  I am fine with removing the reference, I just covered the same material docs already covered.

> Nor should we cover all the possible installation methods.

I would be happy to see someone more closely involved with the project produce unified single installation method for Windows, I had a lot of trouble because docs _already_ cover multiple and partially outdated approaches.

> I never got any of those errors this PR is supposed to fix and I have been doing this for years with numerous installations.

I can drop the fixes part if it's unwanted, though in my opinion those issues aren't unique to me and burnt a lot of my time to look up working fixes.

Overall please let me know if you would like me to make further changes to the PR. I am fine if it's not accepted, just wanted to provide feedback on the state of it. :)

XhmikosR:
That is why I say all other ways shouldn't be listed at all here. You can't cover all possible ways, and keep them up to date at all times.

The simplest way is the way I described above; 3 simple steps without a lot of words and useless stuff for most users.

DirtyF:
👍 Let just leave GitHub Pages out of the default install process, if you have jekyll running, you're already good to go.

👍  to recommend and link to Ruby 2.4 by default.

Rarst:
I've truncated RubyInstaller section to the requested scope.

ashmaroli:
Sorry @Rarst IMO, I don't see the need for updating here.. (based on current diff)
I'm using Jekyll 3.7.0 on Ruby 2.3 and Windows 7.. no issues at all whatsoever..

Jekyll has support for upto Ruby 2.2.. so I don't see the point in documenting that Ruby 2.4.x has to be installed. We can recommend it but there's no need to promote it..
Steps 2 and 3 are basically expanding the valid single command `gem install jekyll bundler`

PR #6692, opened 2018-01-11
Glob scope path only if configured with a pattern

### LINKED ISSUE #6691: Front Matter defaults performance regression

---

- [x] I am on (or have tested on) ***Windows*** 7+

---

- [ ] I was trying to build.
- [x] It was another bug.

## My Reproduction Steps

1. Clone test repo [`https://github.com/mmistakes/front-matter-defaults.git`](https://github.com/mmistakes/front-matter-defaults)
2. Run `bundle update` to install Jekyll 3.6.2
3. Run `bundle exec jekyll build`, and take note of build time `~15 seconds`.
4. Change `gem "jekyll", "~> 3.6.2"` to `3.7.0` in Gemfile.
5. Run `bundle update` to install latest Jekyll version.
6. Run `bundle exec jekyll build` again, take note of increased build time of `~40 seconds`.

## The Output I Wanted

For build time not to double. I noticed this on [my rather large Jekyll site](https://github.com/mmistakes/made-mistakes-jekyll) when updating to `3.7.0`. Build time was usually around two minutes, but now it's unmanageable.

Took me awhile to track down what was the bottleneck. When running Jekyll with `--verbose` I noticed that during "Generating" each line of `Reading: _posts/xxxx-xx-xx-whatever.md` were taking much longer, which led me to [lib/jekyll/frontmatter_defaults.rb](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/frontmatter_defaults.rb).

When I disabled all of my Front Matter defaults in `_config.yml` everything sped right up to pre 3.7.0 speeds. But obviously my pages don't build because I'm assigning layouts using `defaults`.

ashmaroli:
Resolves #6691

Take the less expensive route unless the user configures the `scope["path"]` with a glob pattern. e.g. include
  - `"content/*.pdf"`
  - `"content/**/*"`
  - `"content/config.?"`
  - `"content/config.[^a]"`
  - `"content/config.{yaml, yml}"`

### DIFF
--- a/docs/_docs/configuration.md
+++ b/docs/_docs/configuration.md
@@ -549,7 +549,9 @@ defaults:
 In this example, the `layout` is set to `default` inside the
 [collection](../collections/) with the name `my_collection`.
 
-It is also possible to use glob patterns when matching defaults. For example, it is possible to set specific layout for each `special-page.html` in any subfolder of `section` folder. {%- include docs_version_badge.html version="3.7.0" -%}
+### Glob patterns in Front Matter defaults
+
+It is also possible to use glob patterns (currently limited to patterns that contain `*`) when matching defaults. For example, it is possible to set specific layout for each `special-page.html` in any subfolder of `section` folder. {%- include docs_version_badge.html version="3.7.0" -%}
 
 ```yaml
 collections:
@@ -563,6 +565,17 @@ defaults:
     values:
       layout: "specific-layout"
 ```
+
+<div class="note warning">
+  <h5>Globbing and Performance</h5>
+  <p>
+    Please note that globbing a path is a known performance-poison and is
+    currently not optimized, especially on Windows. Globbing a path will
+    increase your build times in proportion to the size of the associated
+    collection directory.
+  </p>
+</div>
+
 
 ### Precedence
 
--- a/docs/_docs/configuration.md
+++ b/docs/_docs/configuration.md
@@ -569,10 +569,10 @@ defaults:
 <div class="note warning">
   <h5>Globbing and Performance</h5>
   <p>
-    Please note that globbing a path is a known performance-poison and is
-    currently not optimized, especially on Windows. Globbing a path will
-    increase your build times in proportion to the size of the associated
-    collection directory.
+    Please note that globbing a path is known to have a negative effect on
+    performance and is currently not optimized, especially on Windows.
+    Globbing a path will increase your build times in proportion to the size
+    of the associated collection directory.
   </p>
 </div>
 
--- a/lib/jekyll/frontmatter_defaults.rb
+++ b/lib/jekyll/frontmatter_defaults.rb
@@ -97,6 +97,7 @@ module Jekyll
       applies_path?(scope, path) && applies_type?(scope, type)
     end
 
+    # rubocop:disable Metrics/AbcSize
     def applies_path?(scope, path)
       return true if !scope.key?("path") || scope["path"].empty?
 
@@ -108,12 +109,14 @@ module Jekyll
       if glob_pattern?(rel_scope_path)
         Dir.glob(abs_scope_path).each do |scope_path|
           scope_path = Pathname.new(scope_path).relative_path_from site_path
+          Jekyll.logger.debug "Globbed Scope Path:", scope_path
           return true if path_is_subpath?(sanitized_path, scope_path)
         end
       end
 
       path_is_subpath?(sanitized_path, rel_scope_path)
     end
+    # rubocop:enable Metrics/AbcSize
 
     def path_is_subpath?(path, parent_path)
       path.ascend do |ascended_path|
--- a/lib/jekyll/frontmatter_defaults.rb
+++ b/lib/jekyll/frontmatter_defaults.rb
@@ -106,7 +106,7 @@ module Jekyll
       rel_scope_path = Pathname.new(scope["path"])
       abs_scope_path = File.join(@site.source, rel_scope_path)
 
-      if glob_pattern?(rel_scope_path)
+      if scope["path"].to_s.include?("*")
         Dir.glob(abs_scope_path).each do |scope_path|
           scope_path = Pathname.new(scope_path).relative_path_from site_path
           Jekyll.logger.debug "Globbed Scope Path:", scope_path
@@ -125,13 +125,6 @@ module Jekyll
         end
       end
 
-      false
-    end
-
-    def glob_pattern?(path)
-      path.each_filename do |str|
-        return true if str =~ %r!\*+|\?|\[.*\]|{.*}!
-      end
       false
     end
 
--- a/lib/jekyll/frontmatter_defaults.rb
+++ b/lib/jekyll/frontmatter_defaults.rb
@@ -101,13 +101,15 @@ module Jekyll
       return true if !scope.key?("path") || scope["path"].empty?
 
       sanitized_path = Pathname.new(sanitize_path(path))
-
-      site_path = Pathname.new(@site.source)
+      site_path      = Pathname.new(@site.source)
       rel_scope_path = Pathname.new(scope["path"])
       abs_scope_path = File.join(@site.source, rel_scope_path)
-      Dir.glob(abs_scope_path).each do |scope_path|
-        scope_path = Pathname.new(scope_path).relative_path_from site_path
-        return true if path_is_subpath?(sanitized_path, scope_path)
+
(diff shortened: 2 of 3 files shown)


DirtyF:
On latest macOS with my macbook air and @mmistakes test repo:

|  3.7.0     |   Fix       |
|---------|---------|
| 25.321s | 15.913s |
| 30.145s | 15.979s |
| 29.49s  |  16.09s   |

ashmaroli:
> Might be worth adding a test to check that

@Crunch09 That's a good idea to be extra sure.. Will see if I can come up with a simple check..


> On latest macOS with my macbook air

@DirtyF  Thank you very much for running the tests :100:

mmistakes:
Tested the fix on the front-matter-defaults repo and my large personal site... both are back to building at `3.6.2` speeds.

:smile:

lib/jekyll/frontmatter_defaults.rb
```diff
@@ -123,6 +128,13 @@ def path_is_subpath?(path, parent_path)
       false
     end
 
+    def glob_pattern?(path)
+      path.each_filename do |str|
+        return true if str =~ %r!\*+|\?|\[.*\]|{.*}!
```
parkr: Does `path.include?(“*”)` not work for this case?

ashmaroli: I wanted to test the path component string for specific glob patterns. and regex seems better suited for it than multiple `String#include?` calls

parkr: @ashmaroli Why do we need to test each filename separately? AFAIK, the presence of `*` in `scope["path"]` is enough to know it's a glob pattern.

lib/jekyll/frontmatter_defaults.rb
```diff
@@ -97,21 +97,26 @@ def applies?(scope, path, type)
...
-
-      site_path = Pathname.new(@site.source)
+      site_path      = Pathname.new(@site.source)
       rel_scope_path = Pathname.new(scope["path"])
       abs_scope_path = File.join(@site.source, rel_scope_path)
-      Dir.glob(abs_scope_path).each do |scope_path|
-        scope_path = Pathname.new(scope_path).relative_path_from site_path
-        return true if path_is_subpath?(sanitized_path, scope_path)
+
+      if glob_pattern?(rel_scope_path)
```
parkr: So instead of this, we do `if scope["path"].to_s.include?("*")`. Or you can change the implementation of `glob_pattern?` to just do `path.to_s.include?("*")`. What do you think?

lib/jekyll/frontmatter_defaults.rb
ashmaroli: @parkr 
```
defaults:
  - scope:
      path: "_puppies/arrears.?"
      type: "puppies"
    values:
      layout: arrears
``` 
is also a glob pattern to catch all kinds of files with basename `arrears`, if you ask me..

ashmaroli:
> if you don't mind

@parkr I don't mind adding tests at all.. but just that I find it difficult to digest your suggestion above..
The absence of `*` in the `scope["path"]` doesn't mean it's not a glob pattern..

/cc @pathawks  for your inputs.. (Thanks in advance..)

Regarding the unit test,
I won't be able to `assert @site.frontmatter_defaults.glob_include?("content/*.pdf")` because `glob_include?` is a `private` method.. ( yeah, I can use `:send`.. )

DirtyF:
Those use case seems already fine for most users, we never spoke of supporting all possible regexp patterns in path, do we?

```
"content/*.pdf"
"content/**/*"
```

Let's try to keep it simple here for now, and ship a quick fix.

ashmaroli:
> we never spoke of supporting all possible regexp patterns in path, do we?

The `scope["path"]` in @mmistakes test-repo doesn't contain *any* glob-patterns.. If `Dir.glob()` is able to suffocate such a path, what's the point in filtering just one type of glob-pattern..? His site is going to choke when he adds another glob pattern other than the basic `**/*` or `*` patterns..

lib/jekyll/frontmatter_defaults.rb
```diff
@@ -97,21 +97,26 @@ def applies?(scope, path, type)
...
+
+      if scope["path"].to_s.include?("*")
+        Dir.glob(abs_scope_path).each do |scope_path|
+          scope_path = Pathname.new(scope_path).relative_path_from site_path
+          Jekyll.logger.debug "Globbed Scope Path:", scope_path
+          return true if path_is_subpath?(sanitized_path, scope_path)
+        end
       end
 
       path_is_subpath?(sanitized_path, rel_scope_path)
```
pathawks: Could we make this an `else` for the previous block (`if scope["path"].to_s.include?("*")`)?
Multiple exits can get confusing.

ashmaroli: We can't make this an `else` branch because if `diff-ln#113` evaluates to `false` even at the end of looping the globbed array, `diff-ln#117` acts as a fallback..

That said, it does not make sense to me why that fallback's needed..

docs/_docs/configuration.md
```diff
@@ -564,6 +566,17 @@ defaults:
       layout: "specific-layout"
 ```
 
+<div class="note warning">
+  <h5>Globbing and Performance</h5>
+  <p>
+    Please note that globbing a path is a known performance-poison and is
```
parkr: The `poison` sounds really terrifying! Can we rephrase?

> Please note that front matter defaults with glob paths have known performance problems.
> Windows users may notice especially high build times by adding globbing. We are tracking [in this issue](let's open an issue).

ashmaroli: > We are tracking [in this issue](let's open an issue).

I believe this is an issue with Ruby for Windows itself as Ruby is not optimized for Windows.. There's not much one can do here IMO..

So.. I'll just re-phrase the "poison" bit to something more mellow..

DirtyF:
@jekyllbot: merge +bug

ashmaroli:
@mmistakes If possible can you please provide the following feedback on this change using your `front-matter-defaults` repo.. and Jekyll's `master` branch..?
  1. Build times on Windows and Unix with `jekyll build`
  2. Build times on Windows and Unix with `jekyll build --watch` followed by a minor change to `about.md`
  3. Build times on Windows and Unix with `jekyll serve` followed by a minor change to `about.md`

Thanks in advance..

mmistakes:
Sure thing @ashmaroli, here's what I'm getting on Windows.

| **3.6.2** | **3.7.0** | **jekyll/jekyll `master`** |
| --- | --- | --- |
| 20.098s | 48.487s | 20.729s |
| 20.721s | 49.879s | 21.192s |
| 20.977s | 50.313s | 21.518s |

ashmaroli:
> here's what I'm getting

Right.. and the regeneration times..?

mmistakes:
Those are the regeneration times. It wasn't any faster regenerating the page after `about.md` was modified. Still took ~20s... or in the case of `3.7.0` around 50.

ashmaroli:
Okay buddy.. Thanks a lot..
On my system (Windows 7, Ruby 2.3, `jekyll:master` and `minima:master`)
the build times were:
  - initial-build: **45.412 seconds**
  - regeneratn.: **254.069825 seconds**

Something's wrong somewhere!!

mmistakes:
Forgot to mention, I added the `wdm` gem to my `Gemfile` as instructed by this warning:

```
gem 'wdm', '>= 0.1.0' if Gem.win_platform?
```

Without it the regen speeds were much slower for some reason.

## Without `wdm`:

- initial-build: **20.914 seconds**
- regeneration: **37.999 seconds**

## With `wdm`:

- initial-build: **20.624 seconds**
- regeneration:  **21.493 seconds**

ashmaroli:
MIND=BLOWN !!
Two years of Jekyll on Windows and I finally realize the truth behind the suggestion to include `gem 'wdm'`..

### Without `wdm`:
  - initial-build: **47.503 seconds**
  - regeneratn.: **246.402432 seconds**

### With `wdm`:
  - initial-build: **45.334 seconds**
  - regeneratn.: **48.469285 seconds**

:100: :clap:  @mmistakes

mmistakes:
@ashmaroli :smiley: 

I've never really known what adding `wdm` did. I feel like at one point `--watch` stopped working for me reliably so I added it, but didn't know it helped performance until now.

PR #6696, opened 2018-01-15
Debug writing files during the build process

ashmaroli:
We have log outputs for the `Reading` phase upto `Rendering`.. but no outputs when a generated file is being written to the disk.

This prints the destination path of each file being written to disk and the write-path to the metadata file. (all relative to `site.source`)

### DIFF
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -205,7 +205,9 @@ module Jekyll
     # Returns nothing.
     def write
       each_site_file do |item|
-        item.write(dest) if regenerator.regenerate?(item)
+        next unless regenerator.regenerate?(item)
+        Jekyll.logger.debug "Writing:", item.destination(dest).sub("#{source}/", "")
+        item.write(dest)
       end
       regenerator.write_metadata
       Jekyll::Hooks.trigger :site, :post_write, self
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -205,9 +205,7 @@ module Jekyll
     # Returns nothing.
     def write
       each_site_file do |item|
-        next unless regenerator.regenerate?(item)
-        Jekyll.logger.debug "Writing:", item.destination(dest).sub("#{source}/", "")
-        item.write(dest)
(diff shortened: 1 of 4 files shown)


lib/jekyll/site.rb
```diff
@@ -205,7 +205,9 @@ def cleanup
     # Returns nothing.
     def write
       each_site_file do |item|
-        item.write(dest) if regenerator.regenerate?(item)
+        next unless regenerator.regenerate?(item)
+        Jekyll.logger.debug "Writing:", item.destination(dest).sub("#{source}/", "")
```
parkr: Lots of unnecessary string allocation here! I wish there were a good way to only allocate these arguments if you wanted debugging output... :/

pathawks: I don't know anything about what Functional Programming in Ruby... is there some way to pass a closure that returns a string only if logging is enabled?

ashmaroli: > I wish there were a good way to only allocate these arguments if you wanted debugging output... :/

I think that's how it already is..
https://github.com/jekyll/jekyll/blob/bc2c0c4f80e34b78ed1d83b05710e770ffccf728/lib/jekyll/log_adapter.rb#L142-L145

https://github.com/jekyll/jekyll/blob/bc2c0c4f80e34b78ed1d83b05710e770ffccf728/lib/jekyll/log_adapter.rb#L52-L54

DirtyF: @ashmaroli …so we should apply @pathawks 's feedback here?

ashmaroli: > so we should apply..?

IMO, there's no need to.. because as per my understanding of the code linked to above, the `:write` method is going to return `false` instead of strings if the user is not building with `--verbose`

pathawks: The problem is not with the return, but with the parameters being passed.

`item.destination(dest).sub("#{source}/", "")` will be executed even if logging is turned off.

DirtyF:
@jekyllbot: merge +minor

PR #6697, opened 2018-01-16
Fix timezone incosistencies between different ruby version

Crunch09:
**tl;dr** SafeYAML with Ruby < 2.4 parses times in the timezone currently set in the environment  while SafeYAML for Ruby 2.4 keeps times in the timezone in which they were dumped in.

As described in issues like #5963 and #6033 the current `SafeYAML.load` behaviour depends on the ruby version.
```ruby
# ruby -v => ruby 2.3.3p222
ENV['TZ'] = 'Australia/Melbourne'
hash = {date: Time.now}       # => {:date=>2018-01-17 09:25:57 +1100}
SafeYAML.load YAML.dump(hash) # => {:date=>2018-01-17 09:25:57 +1100}
ENV['TZ'] = 'UTC'
SafeYAML.load YAML.dump(hash) # => {:date=>2018-01-16 22:25:57 +0000}
```

```ruby
# ruby -v => ruby 2.4.2p198
ENV['TZ'] = 'Australia/Melbourne'
hash = {date: Time.now}       # => {:date=>2018-01-17 09:25:57 +1100}
SafeYAML.load YAML.dump(hash) # => {:date=>2018-01-17 09:25:57 +1100}
ENV['TZ'] = 'UTC'
SafeYAML.load YAML.dump(hash) # => {:date=>2018-01-17 09:25:57 +1100}
```

This scenario should have caught this, but unfortunately the dates in our scenarios were read as `String` instead of `Time` (which doesn't happen with a real frontmatter, see unit tests) so the issue didn't occur. 
5365355   should fix that and make this scenario fail in Ruby 2.4
https://github.com/jekyll/jekyll/blob/603f2ea52f52cd3a64e37371349d8292f3a83655/features/site_configuration.feature#L184-L204

_I couldn't test so far locally with Ruby 2.5 but think the behaviour should stay the same as it is with Ruby 2.4_

cc: @parkr Made a PR out of this because it might be easier to discuss and also CI runs then.


lib/jekyll/document.rb
```diff
@@ -426,7 +426,7 @@ def merge_categories!(other)
 
     private
     def merge_date!(source)
-      if data.key?("date") && !data["date"].is_a?(Time)
+      if data.key?("date")
```
pathawks: I see the addition of tests :+1:

What’s this line about?

Crunch09: Sorry, didn't had time yesterday to properly update this PR with more information.

This change actually fixes the inconsistent behaviour, because as long as `data['date']` is set it will call [`Utils.parse_date`](https://github.com/jekyll/jekyll/blob/603f2ea52f52cd3a64e37371349d8292f3a83655/lib/jekyll/utils.rb#L135-L139) which transforms the given data into `localtime` (as determined by `ENV['TZ']`). This is exactly the behaviour that SafeYAML was doing automatically before Ruby 2.4.
Because of the old version of this line there were no inconsistencies between the different Ruby versions when `data['date']` was interpreted as a `String` rather than a `Time`.

That said, I am not sure the behaviour of Ruby 2.3 is my preferred one, i would actually prefer the behaviour of newer ruby versions but the most important thing is to fix the incosistency and i would argue that jekyll users are more used to the behaviour of `everything gets converted into whatever ENV['TZ'] is`.

Crunch09:
@stomar @yous @Manishearth @t-brink Would love to hear your thoughts on this as you were quite involved in past discussions of this issue.

cc: @jekyll/build

Crunch09:
@parkr It's not visible in the changed files so i mention it now here: This scenario https://github.com/jekyll/jekyll/blob/603f2ea52f52cd3a64e37371349d8292f3a83655/features/site_configuration.feature#L184 would now fail in newer Ruby versions without the fix. ([Travis log of failure](https://travis-ci.org/jekyll/jekyll/jobs/329682575#L916))

yous:
In my local machine with KST timezone,

``` sh
$ ruby -e 'puts ENV["TZ"].inspect'
nil
$ date +"%Z %z"
KST +0900
```

The test `features/permalinks.feature:73` fails. I added `TEST` step to show the file entries in `_site`.

```
$ bundle exec cucumber features/permalinks.feature:73

features/permalinks.feature:73  Scenario: Use custom permalink schema with date and time Feature: Fancy permalinks
  As a hacker who likes to blog
  I want to be able to set permalinks
  In order to make my blog URLs awesome

  Scenario: Use custom permalink schema with date and time            # features/permalinks.feature:73
    Given I have a _posts directory                                   # features/step_definitions.rb:68
    And I have the following post:                                    # features/step_definitions.rb:76
      | title                   | category | date                | content         |
      | Custom Permalink Schema | stuff    | 2009-03-27 22:31:07 | Totally custom. |
    And I have a configuration file with:                             # features/step_definitions.rb:125
      | key       | value                                      |
      | permalink | "/:year:month:day:hour:minute:second.html" |
      | timezone  | UTC                                        |
    When I run jekyll build                                           # features/step_definitions.rb:157
    Then I should get a zero exit status                              # features/step_definitions.rb:361
    And the _site directory should exist                              # features/step_definitions.rb:215
    And TEST                                                          # features/step_definitions.rb:289
      [".", "..", "20090327133107.html"]
    And I should see "Totally custom." in "_site/20090327223107.html" # features/step_definitions.rb:225
      expected #<Pathname:_site/20090327223107.html> to exist (RSpec::Expectations::ExpectationNotMetError)
      /Users/yous/src/jekyll/features/step_definitions.rb:310:in `/^the "(.*)" file should +(not )?exist$/'
      /Users/yous/src/jekyll/features/step_definitions.rb:226:in `/^I should (not )?see "(.*)" in "(.*)"$/'
      features/permalinks.feature:86:in `And I should see "Totally custom." in "_site/20090327223107.html"'
 (0.580991s)

Worst offenders:
  0.580991s for "Use custom permalink schema with date and time" (features/permalinks.feature:73)

Failing Scenarios:
cucumber features/permalinks.feature:73 # Scenario: Use custom permalink schema with date and time

1 scenario (1 failed)
8 steps (1 failed, 7 passed)
0m0.581s
```

Is this intentional? I didn't set `ENV['TZ']` explicitly and Jekyll's configuration contains `timezone: UTC`, but the permalink is created based on KST.

Crunch09:
@yous Thanks! Looking at that test.

Crunch09:
@yous This issue only occured in a test and wouldn't have happened in a real application. This is because unlike during a real build, the post is read before any configuration has been applied (in this case `timezone`). Even though your `ENV['TZ']` is not set Ruby still recognizes the time / timezone of your computer, you can test this with:
```ruby
Time.new # 2018-01-20 20:58:12 +0900
```
https://github.com/jekyll/jekyll/pull/6697/commits/e3b32f3a28de85e85bb054513d36063f0fe97d0a now makes sure that the timezone of the scenario is set before reading any post frontmatter.
Many thanks again for catching this 👍 I couldn't see this behaviour earlier because i'm living in the UTC timezone and i assume CI tests also run in UTC. Please let me know if this scenario (or any other one) is still failing for you.

yous:
@Crunch09 Now every test passes. Thanks for the quick fix and explanation. 👍

pup:
Additionally, we can probably squeeze this in v3.7.1. What do you think, @jekyll/core?

pup:
Merging this, assuming nobody has any further qualms with it.

@jekyllbot: merge +bug

pup:
@jekyllbot: merge +bug

Crunch09:
Thanks @pup ! 🙇

Issue #6702, opened 2018-01-18
Posts not generated when _posts folder is located within custom collections_dir


paulrobertlloyd:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [x] I ran `jekyll doctor` to check my configuration
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [ ] This is a feature request.

---

- [x] I am on (or have tested on) ***macOS*** 10+
- [ ] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux
- [ ] I am on (or have tested on) ***Fedora*** GNU/Linux
- [ ] I am on (or have tested on) ***Arch*** GNU/Linux
- [ ] I am on (or have tested on) ***Other*** GNU/Linux
- [ ] I am on (or have tested on) ***Windows*** 10+

---

- [ ] I was trying to install.
- [ ] There is a broken Plugin API.
- [ ] I had an error on GitHub Pages, and I have reproduced it locally.
- [ ] I had an error on GitHub Pages, and GitHub Support said it was a Jekyll Bug.
- [ ] I had an error on GitHub Pages and I did not test it locally.
- [x] I was trying to build.
- [ ] It was another bug.

## My Reproduction Steps

Given the following folder setup:

```
├─ _collections
│  ├─ _books
│  ├─ _posts
│  └─ _recipes
└─ _config.yml
```
…and the following configuration:

```yams
collections_dir: _collections
collections:
  books:
    output: true
  posts:
    output: true
  recipes:
    output: true
```

No posts are generated. However, moving the `_posts` folder back into the source directory:

```
├─ _collections
│  ├─ _books
│  └─ _recipes
├─ _posts
└─ _config.yml
```

Generates posts as expected. I am showing representative folder structure and config here, different from my own. But I did the same (moving `_posts` into a custom collections folder and updating the config) after creating a new site (`jekyll new`), and the same behaviour is exhibited.

## The Output I Wanted

The following messages appears on the website:

> **Be sure to move posts into custom collections directory**
> If you specify a directory to store all your collections in the same place with `collections_dir: my_collections`, then you will need to move your `_posts` directory to `my_collections/_posts`.

This was the expected behaviour, however it would appear that the opposite is true.

ashmaroli:
Your `collections_dir` value **should not start with an underscore (`_`)**

[...]

PR #6703, opened 2018-01-18
Add note about naming of collections_dir

paulrobertlloyd:
Clarifying situation in follow up to #6702. /cc @ashmaroli

### DIFF
--- a/docs/_docs/collections.md
+++ b/docs/_docs/collections.md
@@ -58,7 +58,7 @@ defaults:
 <div class="note warning">
   <h5>Be sure to move posts into custom collections directory</h5>
 
-  <p>If you specify a directory to store all your collections in the same place with <code>collections_dir: my_collections</code>, then you will need to move your <code>_posts</code> directory to <code>my_collections/_posts</code>.</p>
+  <p>If you specify a directory to store all your collections in the same place with <code>collections_dir: my_collections</code>, then you will need to move your <code>_posts</code> directory to <code>my_collections/_posts</code>. Note that, the name of your collections directory cannot start with an underscore (`_`).</p>
 </div>
 
 ### Step 2: Add your content {#step2}


docs/_docs/collections.md
```diff
@@ -55,10 +55,10 @@ defaults:
   in <code>my_collections/_recipes</code> for the <code>recipes</code> collection.</p>
 </div>
 
-<div class="note warning">
-  <h5>Be sure to move posts into custom collections directory</h5>
+<div class="note">
+  <h5>Moving posts into the custom collections directory</h5>
 
-  <p>If you specify a directory to store all your collections in the same place with <code>collections_dir: my_collections</code>, then you will need to move your <code>_posts</code> directory to <code>my_collections/_posts</code>.</p>
+  <p>If you want to move your `_posts` folder inside the custom collections directory as well, the name of the `collections_dir` cannot start with an underscore (_).</p>
```
ashmaroli: There is no scope for a *"If you want to move `_posts`.."*
Its always, "If you specify a custom directory.., then `_posts` **has to be moved** into it as well.."
and since `_posts` won't be read inside directories starting with an underscore, it implies:
```
Custom collections_dir cannot start with an underscore
```

paulrobertlloyd: Hmmm, that’s not the behaviour I’m seeing. I can set a custom collections directory, **and** have my `_posts` folder outside of it. The problem only occurs if I move `_posts` into a the `collections_dir` (and it begins with an underscore).

ashmaroli: Please check the generated URL to your posts..
`./_posts/2018-01-18-hello.md` will have generated a `./_site/_posts/2018/01/18/hello.html` .. (Note the `_posts` in your URL..)

ashmaroli:
Therefore, I'm of the opinion, please do not *alter* the existing note. Instead, add to it..
/cc @jekyll/documentation

docs/_docs/collections.md
paulrobertlloyd: Ah yes, I hadn’t spotted that in my test site (and for my own purposes, I am using front matter defaults to explicitly set permalink for posts). Okay, so how about this:

> If you specify a directory to store all your collections in the same place with `collections_dir: my_collections`, then you will need to move your `_posts` directory to `my_collections/_posts`. Note that if you do this, the name of your collections directory cannot start with an underscore (_).

DirtyF: Thanks @paulrobertlloyd, 👍 for your last proposal

paulrobertlloyd: Great. Do you need me to update this PR?

ashmaroli: IMO, there's no need for two if statements..
> **If you specify a directory to store all your collections in the same place with `collections_dir: my_collections`, then you will need to move your` _posts` directory to `my_collections/_posts`. Note that, the name of your collections directory cannot start with an underscore (`_`).**


> Great. Do you need me to update this PR?

Yes. Please..

paulrobertlloyd: @ashmaroli I wasn’t sure about that, but then realised we are adding this note within the context of moving `_posts` into the custom directory. So yes, no need for the extra ‘if’. I have updated this PR.

DirtyF:
@jekyllbot: merge +docs

PR #6711, opened 2018-01-24
Add gem "wdm" to all newly generated Gemfiles

ashmaroli:
I recently found out that adding `gem "wdm"` to the Gemfile actually [boosts performance](https://github.com/jekyll/jekyll/pull/6692#issuecomment-359865460) on Windows while `watch: true`..
So its better we have it included in the newly generated `Gemfile` by default..


DirtyF:
@jekyllbot: merge +dev

parkr:
@dirtyf this is a `+minor` Or `+bug` since it affects the output of the `jekyll new` Command.

DirtyF:
@parkr right, updated https://github.com/jekyll/jekyll/blob/master/History.markdown#bug-fixes

ashmaroli:
This has been classified as a `bug-fix` ?? :confused:

PR #6716, opened 2018-01-25
Update plugins.md

hendrikschneider:
Updated plugin description

### DIFF
--- a/docs/_docs/plugins.md
+++ b/docs/_docs/plugins.md
@@ -920,7 +920,7 @@ You can find a few useful plugins at the following locations:
 
 #### Other
 
-- [Analytics for Jekyll](https://github.com/hendrikschneider/jekyll-analytics) by Hendrik Schneider: An effortless way to add  various trackers like Google Analytics, Piwik, etc. to your site
+- [Analytics for Jekyll](https://github.com/hendrikschneider/jekyll-analytics) by Hendrik Schneider: An effortless way to add  various trackers like Google Analytics, Piwik, Matomo, MPulse, etc. to your site
 - [ditaa-ditaa](https://github.com/tmthrgd/ditaa-ditaa) by Tom Thorogood: a drastic revision of jekyll-ditaa that renders diagrams drawn using ASCII art into PNG images.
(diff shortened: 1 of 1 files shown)


docs/_docs/plugins.md
```diff
@@ -920,7 +920,7 @@ You can find a few useful plugins at the following locations:
 
 #### Other
 
-- [Analytics for Jekyll](https://github.com/hendrikschneider/jekyll-analytics) by Hendrik Schneider: An effortless way to add  various trackers like Google Analytics, Piwik, etc. to your site
+- [Analytics for Jekyll](https://github.com/hendrikschneider/jekyll-analytics) by Hendrik Schneider: An effortless way to add  various trackers like Google Analytics, Piwik, Matomo, MPulse, etc. to your site
```
ashmaroli: - Piwik is now Matomo.. so it should not be Piwik, Matomo
- All docs refer to "MPulse" as "mPulse"

```diff
- trackers like Google Analytics, Piwik, Matomo, MPulse, etc. to your site
+ trackers like Google Analytics, Matomo (Piwik), mPulse, etc. to your site
```

hendrikschneider: I think it is also important to mention Piwik to show backwards compatibility. Would you accept this description?

```
An effortless way to add various trackers like Google Analytics, 
Matomo (former Piwik), mPulse, etc. to your site.
```

ashmaroli: yes.. I'm fine with adding an adjective.. `formerly` sounds more apt though..

docs/_docs/plugins.md
```diff
@@ -920,7 +920,7 @@ You can find a few useful plugins at the following locations:
 
 #### Other
 
-- [Analytics for Jekyll](https://github.com/hendrikschneider/jekyll-analytics) by Hendrik Schneider: An effortless way to add  various trackers like Google Analytics, Piwik, etc. to your site
+- [Analytics for Jekyll](https://github.com/hendrikschneider/jekyll-analytics) by Hendrik Schneider: An effortless way to add various trackers like Google Analytics, Matomo (former Piwik), mPulse, etc. to your site.
```
pup: This should say `formerly` instead of `former`.

pup:
@jekyllbot: merge +docs

Issue #204, opened 2010-08-30
Test/Unit 1.9.2 Compatibility Fixes


jbw:
- Test/Unit suite now runs on 1.9.2 (Minor changes).

I'm looking to add a feature I'd like so thought I would make this work first.

http://github.com/jbw/jekyll/tree/ruby-v1.9

PR #6720, opened 2018-01-25
Remove redgreen gem

yous:
This is a reflect of #204, 093a5ef.

About redgreen:
- https://rubygems.org/gems/redgreen/versions/1.2.2
- http://on-ruby.blogspot.kr/2006/05/red-and-green-for-ruby.html

I found that redgreen is installed by running `rg` command. It stands for ripgrep now, at least for me.


DirtyF:
@jekyllbot: merge +dev

PR #6724, opened 2018-01-28
Append appropriate closing tag to Liquid block in an excerpt

### LINKED ISSUE #5596: Excerpt.extract_excerpt() fails when first paragraph contains part of a multi-line Liquid tag.
- [ ] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`

---

---

- [x] I was trying to build.
- [ ] It was another bug.

## My Reproduction Steps

1. Create a default Jekyll site with `jekyll new jekyll-issue`
2. `cd jekyll-issue`
3. Edit `_posts/YYYY-MM-DD-welcome-to-jekyll.markdown`:
  - NOTE: Leave the YAML front matter untouched
  - Remove all Markdown content before the Ruby code sample
  - Insert an empty line somewhere in the middle of the Ruby code sample
4. Run `jekyll build`
  - `jekyll build` exits with error "Liquid Exception: undefined method `t' for nil:NilClass in _posts/YYYY-MM-DD-welcome-to-jekyll.markdown/#excerpt"

## The Output I Wanted

A **Welcome to Jekyll** post which starts with a Ruby code sample that has empty lines in it for readability.

## Work Around 1

Don't put any empty lines in the code sample.

## Work Around 2

Add some other Markdown content before the before the Ruby code sample (a  [Markdown comment](http://stackoverflow.com/a/20885980) works well for this). Note that adding content after the Ruby code sample does not work.

## My System

- Jekyll 3.0.1
- Ruby 2.3.1p112 (2016-04-26) [x86_64-linux-gnu]
- Ubuntu 16.04.1

/cc include any Jekyll affinity teams here (see https://teams.jekyllrb.com/ for more info)

ashmaroli:
Resolves #5596 

If a `Liquid::Block` tag subclass gets caught between the `excerpt_separator`, then append the concerned block delimiter to the excerpt to avoid a `Liquid Exception due to invalid syntax`

### DIFF
--- a/lib/jekyll/excerpt.rb
+++ b/lib/jekyll/excerpt.rb
@@ -116,11 +116,28 @@ module Jekyll
     def extract_excerpt(doc_content)
       head, _, tail = doc_content.to_s.partition(doc.excerpt_separator)
 
+      # append appropriate closing tag (to a Liquid block), to the "head" if the
+      # partitioning resulted in leaving the closing tag somewhere in the "tail"
+      # partition.
+      if head =~ %r!{%\s*(\w+).+\s*%}!
+        tag_name = Regexp.last_match(1)
+
+        if liquid_block?(tag_name) && head.match(%r!{%\s*end#{tag_name}.+\s*%}!).nil?
+          head << "\n{% end#{tag_name} %}"
+        end
+      end
+
       if tail.empty?
         head
       else
         head.to_s.dup << "\n\n" << tail.scan(%r!^ {0,3}\[[^\]]+\]:.+$!).join("\n")
       end
+    end
+
+    private
+
+    def liquid_block?(tag_name)
+      Liquid::Template.tags[tag_name].superclass == Liquid::Block
     end
   end
 end
--- a/lib/jekyll/excerpt.rb
+++ b/lib/jekyll/excerpt.rb
@@ -119,7 +119,8 @@ module Jekyll
       # append appropriate closing tag (to a Liquid block), to the "head" if the
       # partitioning resulted in leaving the closing tag somewhere in the "tail"
       # partition.
-      if head =~ %r!{%\s*(\w+).+\s*%}!
+      if head.include?("{%")
+        head =~ %r!{%\s*(\w+).+\s*%}!
         tag_name = Regexp.last_match(1)
 
         if liquid_block?(tag_name) && head.match(%r!{%\s*end#{tag_name}.+\s*%}!).nil?
--- a/lib/jekyll/excerpt.rb
+++ b/lib/jekyll/excerpt.rb
@@ -123,7 +123,7 @@ module Jekyll
         head =~ %r!{%\s*(\w+).+\s*%}!
         tag_name = Regexp.last_match(1)
 
-        if liquid_block?(tag_name) && head.match(%r!{%\s*end#{tag_name}.+\s*%}!).nil?
+        if liquid_block?(tag_name) && head.match(%r!{%\s*end#{tag_name}\s*%}!).nil?
           head << "\n{% end#{tag_name} %}"
         end
       end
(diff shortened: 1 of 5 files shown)


oe:
Would it make sense having full Liquid blocks in excerpts in the first place? Assuming you had something like an `if` block in you excerpt, I'm fairly sure that would cause unexpected results for the user if the entire excerpt was suddenly hidden or something. Not sure whether that's intuitive behavior.

lib/jekyll/excerpt.rb
```diff
@@ -116,11 +116,28 @@ def place_in_layout?
     def extract_excerpt(doc_content)
       head, _, tail = doc_content.to_s.partition(doc.excerpt_separator)
 
+      # append appropriate closing tag (to a Liquid block), to the "head" if the
+      # partitioning resulted in leaving the closing tag somewhere in the "tail"
+      # partition.
+      if head =~ %r!{%\s*(\w+).+\s*%}!
```
parkr: Regexp's are _very_ slow! Can we use `head.include?("{%")` before we do regexp? I think this runs on every page generation...

ashmaroli:
> Would it make sense having full Liquid blocks in excerpts in the first place?

@oe IMO, Users are responsible for the "content" within their files.
If a user has multiple documents, each starting with an `{% if %}`, then its definitely something that needs to be moved into a `layout`.
On the other hand, if its one or two documents with a liquid block, the excerpt will render fine (via this patch).. the only case where it would be rendered empty is if `{% capture %}` had been used in which case, they may want to reconsider using the default `excerpt_separator`..

I don't think Jekyll should halt the build process in such cases.. If need be, Jekyll can issue a warning that a liquid block has been handled as required in the excerpt, and may not render as expected.. but that's for another PR..

oe:
@ashmaroli Alright, fair enough.

lib/jekyll/excerpt.rb
```diff
@@ -116,11 +116,29 @@ def place_in_layout?
...
       head, _, tail = doc_content.to_s.partition(doc.excerpt_separator)
 
+      # append appropriate closing tag (to a Liquid block), to the "head" if the
+      # partitioning resulted in leaving the closing tag somewhere in the "tail"
+      # partition.
+      if head.include?("{%")
+        head =~ %r!{%\s*(\w+).+\s*%}!
+        tag_name = Regexp.last_match(1)
+
+        if liquid_block?(tag_name) && head.match(%r!{%\s*end#{tag_name}.+\s*%}!).nil?
```
pathawks: What's with the `.+`?

If is necessary, you can get rid of the `\s*`, since `.+` will also swallow any whitespace.

ashmaroli: nice catch.. `.+` is not needed since end-tags do not take argument(s)..

ashmaroli:
> So this will only close one Liquid tag, is that correct?

yes. That is correct.. I consider excerpts that would contain multiple block-tags or nested block-tags as edge-cases which we shouldn't bother handling.. Let the build fail and leave it to the author to take appropriate action..

pathawks:
I wonder if we could have an _option_ to fail the build when something like this happens.

I would rather the build fail than have a wonky site, but I understand that is maddening when not running Jekyll locally. This change will be an improvement for many users :+1:

ashmaroli:
> I wonder if we could have an *option* to fail the build

I disagree.. IMO, that'd be over-engineering..
Let Liquid decide how to go about based on its `:error_mode`

pathawks:
> Let Liquid decide how to go about…

I mean instead of Jekyll quietly trying to patch things up, we could just let it fail. This is not over-engineering, it is our current behavior.

ashmaroli:
> it is our current behavior

ohh.. I thought you were referring to the above mentioned edge-cases..
yeah.. having a config option say, `smart_excerpts: true # false, by default` would be nice..

oe:
@jekyllbot: merge +bug

DirtyF:
This prevents my blog from building, when I have Liquid in excerpts.

PR #6732, opened 2018-01-31
Replace simple regex with a native Ruby method

ashmaroli:
- Replace `sub(%r!/$!, "")` with `chomp("/")


DirtyF:
@jekyllbot: merge +dev

PR #6735, opened 2018-02-01
Bypass rendering via Liquid unless required

ashmaroli:
Currently, all `Page` objects and certain `Document` objects are always *parsed* and *rendered* via `Liquid` before being passed onto the concerned `Converter`.

With this patch, a simple check is performed after `:pre-render` hooks to determine if the `content` contains any `Liquid Tags / Variables` and then handled accordingly:
  - if the content is simply 100% plaintext (or Markdown), then `Liquid`'s parse-render phase is skipped and directly passed onto the associated `Converter`
  - otherwise, follow current behavior

/cc @jekyll/build

### DIFF
--- a/lib/jekyll/utils.rb
+++ b/lib/jekyll/utils.rb
@@ -147,6 +147,14 @@ module Jekyll
     rescue EOFError
       false
     end
+
+    # Determine whether the given content string contains Liquid Tags or Vaiables
+    #
+    # Returns true is the string contains sequences of `{%` or `{{`
+    def has_liquid_construct?(content)
+      return false if content.nil? || content.empty?
+      content.include?("{%") || content.include?("{{")
+    end
     # rubocop: enable PredicateName
 
     # Slugify a filename or title.
--- a/benchmark/conditional_liquid.rb
+++ b/benchmark/conditional_liquid.rb
@@ -1,0 +1,79 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require "liquid"
+require "benchmark/ips"
+
+# Test if processing content string without any Liquid constructs, via Liquid,
+# is slower than checking whether constructs exist ( using `String#include?` )
+# and return-ing the "plaintext" content string as is..
+#
+# Ref: https://github.com/jekyll/jekyll/pull/6735
+
+# Sample contents
+WITHOUT_LIQUID = <<-TEXT.freeze
+Lorem ipsum dolor sit amet, consectetur adipiscing elit. Fusce auctor libero at
+pharetra tempus. Etiam bibendum magna et metus fermentum, eu cursus lorem
+mattis. Curabitur vel dui et lacus rutrum suscipit et eget neque.
+
+Nullam luctus fermentum est id blandit. Phasellus consectetur ullamcorper
+ligula, at finibus eros laoreet id. Etiam sit amet est in libero efficitur
+tristique. Ut nec magna augue. Quisque ut fringilla lacus, ac dictum enim.
+Aliquam vel ornare mauris. Suspendisse ornare diam tempor nulla facilisis
+aliquet. Sed ultrices placerat ultricies.
+TEXT
+
+WITH_LIQUID = <<-LIQUID.freeze
+Lorem ipsum dolor sit amet, consectetur adipiscing elit. Fusce auctor libero at
+pharetra tempus. {{ author }} et metus fermentum, eu cursus lorem
+mattis. Curabitur vel dui et lacus rutrum suscipit et eget neque.
+
+Nullam luctus fermentum est id blandit. Phasellus consectetur ullamcorper
+ligula, {% if author == "Jane Doe" %} at finibus eros laoreet id. {% else %}
+Etiam sit amet est in libero efficitur.{% endif %}
+tristique. Ut nec magna augue. Quisque ut fringilla lacus, ac dictum enim.
+Aliquam vel ornare mauris. Suspendisse ornare diam tempor nulla facilisis
+aliquet. Sed ultrices placerat ultricies.
+LIQUID
+
+WITH_JUST_LIQUID_VAR = <<-LIQUID.freeze
+Lorem ipsum dolor sit amet, consectetur adipiscing elit. Fusce auctor libero at
+pharetra tempus. et metus fermentum, eu cursus lorem, ac dictum enim.
+mattis. Curabitur vel dui et lacus rutrum suscipit et {{ title }} neque.
+
+Nullam luctus fermentum est id blandit. Phasellus consectetur ullamcorper
+ligula, at finibus eros laoreet id. Etiam sit amet est in libero efficitur.
+tristique. Ut nec magna augue. {{ author }} Quisque ut fringilla lacus
+Aliquam vel ornare mauris. Suspendisse ornare diam tempor nulla facilisis
+aliquet. Sed ultrices placerat ultricies.
+LIQUID
+
+# Mimic how Jekyll's LiquidRenderer would process a non-static file, with
+# some dummy payload
+def always_liquid(content)
+  Liquid::Template.error_mode = :warn
+  Liquid::Template.parse(content, :line_numbers => true).render(
+    "author" => "John Doe",
+    "title"  => "FooBar"
+  )
+end
+
+# Mimic how the proposed change would first execute a couple of checks and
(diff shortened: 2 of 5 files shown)


pathawks:
How expensive are the two `.include?` operations plus an `.empty?` check, vs just letting Liquid perform a noop?

Would really like to see some benchmarks here.

DirtyF:
@pathawks tested it on front-matter-default repo and no significant change (more or less the same)

pathawks:
I would expect that Liquid’s parsing is no more expensive (and possibly less) than a couple RegExes when the content contains no Liquid tags. The extra RegExes will (of course) make rendering take _longer_ when documents _do_ contain Liquid markup.

Because this introduces complexity, I am hesitant to accept unless the benefits are pretty substantial.

lib/jekyll/utils.rb
```diff
@@ -147,6 +147,14 @@ def has_yaml_header?(file)
     rescue EOFError
       false
     end
+
+    # Determine whether the given content string contains Liquid Tags or Vaiables
+    #
+    # Returns true is the string contains sequences of `{%` or `{{`
+    def has_liquid_construct?(content)
+      return false if content.nil? || content.empty?
+      content.include?("{%") || content.include?("{{")
```
pathawks: Are the strings converted to RegEx on the backend? This _might_ be faster?

```ruby
!%r!{[{%]!.match(content).nil?
```

ashmaroli:
@pathawks I've added a rudimentary benchmark to assess the proposed change..

lib/jekyll/utils.rb
ashmaroli: This *might* be faster.. or may not.. Needs to be benchmarked.. :wink:

mattr-:
How big is the front-matter defaults repo? I ask because we'll need to be careful about using microbenchmarks with some of these rendering performance changes as they may introduce real speedups but we won't see it until the size of the site reaches a certain point.

ashmaroli:
@mattr-  The `front-matter-defaults` repo doesn't factor in this PR because this PR deals with the rendering difference between: pages/documents that contain Liquid tags/variables vs pages/documents that are purely Markdown or other type of text..

DirtyF:
I made 10 duplicates of files in the `_docs` folder and ran a _single_ test to check for significant changes: 120s with `master` vs 118s with this branch. More or less the same given build times can differ.

ashmaroli:
> 10 duplicates of files in the `_docs` folder

duplicates of `jekyll/docs/_docs`?

DirtyF:
Nevermind, I'll let you do your own benchmarks, I just wish we could test these kind of optimizations against real cases websites.

DirtyF:
yeah i thought about that, but none of the [websites listed in acceptance tests](https://github.com/jekyll/acceptance/blob/master/script/cibuild) is very large.

parkr:
Do you have the output on these benchmarks? Curious to see the output but away from the computer :)

DirtyF:
@tomjoht blog takes more than 22min to build at that time, we should check all the latest small perf improvements against his website: https://github.com/tomjoht/tomjoht.github.io

DirtyF:
```sh
$ be ruby benchmark/conditional-liquid.rb
         regex-check    71.427k i/100ms
       builtin-check   152.025k i/100ms
         regex-check    961.880k (± 3.5%) i/s -      4.857M in   5.056289s
       builtin-check      3.308M (± 5.1%) i/s -     16.571M in   5.026962s

Comparison:
       builtin-check:  3307537.3 i/s
         regex-check:   961880.1 i/s - 3.44x  slower

         regex-check    55.609k i/100ms
       builtin-check   125.621k i/100ms
         regex-check    691.777k (± 7.6%) i/s -      3.448M in   5.022451s
       builtin-check      2.137M (± 3.3%) i/s -     10.678M in   5.003542s

Comparison:
       builtin-check:  2136608.2 i/s
         regex-check:   691776.7 i/s - 3.09x  slower

         regex-check    53.448k i/100ms
       builtin-check    94.997k i/100ms
         regex-check    617.459k (±12.0%) i/s -      3.047M in   5.017810s
       builtin-check      1.376M (± 5.8%) i/s -      6.935M in   5.059654s

Comparison:
       builtin-check:  1376031.7 i/s
         regex-check:   617459.4 i/s - 2.23x  slower

always thru liquid - plain text
                         4.812k i/100ms
conditional liquid - plain text
                       150.590k i/100ms
always thru liquid - plain text
                         49.822k (± 3.1%) i/s -    250.224k in   5.027488s
conditional liquid - plain text
                          2.931M (± 7.1%) i/s -     14.607M in   5.016229s

Comparison:
conditional liquid - plain text:  2931073.6 i/s
always thru liquid - plain text:    49822.0 i/s - 58.83x  slower

always thru liquid - tags n vars
                       785.000  i/100ms
conditional liquid - tags n vars
                       765.000  i/100ms
always thru liquid - tags n vars
                          8.354k (± 6.1%) i/s -     41.605k in   5.000493s
conditional liquid - tags n vars
                          8.394k (± 5.7%) i/s -     42.075k in   5.030741s

Comparison:
conditional liquid - tags n vars:     8394.2 i/s
always thru liquid - tags n vars:     8354.5 i/s - same-ish: difference falls within error

always thru liquid - just vars
                         1.259k i/100ms
conditional liquid - just vars
                         1.234k i/100ms
always thru liquid - just vars
                         12.765k (± 5.5%) i/s -     64.209k in   5.048326s
conditional liquid - just vars
                         12.867k (± 2.9%) i/s -     65.402k in   5.087556s

Comparison:
conditional liquid - just vars:    12867.0 i/s
always thru liquid - just vars:    12764.9 i/s - same-ish: difference falls within error
```

pathawks:
I changed `check_with_regex` to use `String.match?` instead of `Regexp.match` so that it will stop looking after finding a single match. This is now faster than `check_with_builtin` _if_ the input contains any Liquid.

ashmaroli:
@pathawks Two questions:
- can you post the updated benchmark results?
- can I split the benchmarks into two separate files?

pathawks:
> can I split the benchmarks into two separate files?

Sure :+1:

```text
regex-check   - plain text
                       176.614k i/100ms
builtin-check - plain text
                       229.229k i/100ms
regex-check   - plain text
                          2.897M (± 2.3%) i/s -     14.659M in   5.062371s
builtin-check - plain text
                          4.784M (± 3.9%) i/s -     24.069M in   5.039610s

Comparison:
builtin-check - plain text:  4783868.4 i/s
regex-check   - plain text:  2897317.0 i/s - 1.65x  slower

regex-check   - tags n vars
                       214.528k i/100ms
builtin-check - tags n vars
                       180.318k i/100ms
regex-check   - tags n vars
                          4.204M (± 3.0%) i/s -     21.024M in   5.006021s
builtin-check - tags n vars
                          3.057M (± 2.0%) i/s -     15.327M in   5.016518s

Comparison:
regex-check   - tags n vars:  4203840.7 i/s
builtin-check - tags n vars:  3056602.0 i/s - 1.38x  slower

regex-check   - just vars
                       198.649k i/100ms
builtin-check - just vars
                       134.002k i/100ms
regex-check   - just vars
                          3.649M (± 3.6%) i/s -     18.276M in   5.015865s
builtin-check - just vars
                          1.952M (± 3.3%) i/s -      9.782M in   5.017950s

Comparison:
regex-check   - just vars:  3648780.1 i/s
builtin-check - just vars:  1951683.0 i/s - 1.87x  slower
```

ashmaroli:
hmm.. I'm in a fix now.. not sure what would be the best way to proceed..

ashmaroli:
@pathawks  just realized that `String#match?` was introduced only in Ruby 2.4.0.. So we can't use it now.. :grin:

pathawks:
> `String#match?` was introduced only in Ruby 2.4.0

🤣 That makes things easy then :+1:

DirtyF:
@jekyllbot: merge + minor

pathawks:
The **jekyll-feed** test suite is broken because of this PR.

The failing test: [`it "renders Liquid inside posts"`](https://github.com/jekyll/jekyll-feed/blob/ad043c806089d3a2812c348f71ad7969405ba25b/spec/jekyll-feed_spec.rb#L89-L92)

pathawks:
Turns out, this is already fixed by #6945 :+1:

PR #6737, opened 2018-02-01
Adding the jekyll-algolia plugin to the list of plugins

pixelastic:
I've added the [jekyll-algolia](https://community.algolia.com/jekyll-algolia/) to the list of plugins that can be used with Jekyll. I added them to the "Other" category as this is the one that seemed the more appropriate (the plugin index content to make it available through search).


DirtyF:
@jekyllbot: merge docs

PR #6738, opened 2018-02-01
Memoize parsing the date in document data

ashmaroli:
Parsing a document's `data["date"]` value multiple times across a build session can be avoided especially since the method boils down to calling `Time.parse()` on the same "date"

### DIFF
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -425,12 +425,13 @@ module Jekyll
 
     private
     def merge_date!(source)
-      if data.key?("date")
-        data["date"] = Utils.parse_date(
-          data["date"].to_s,
-          "Document '#{relative_path}' does not have a valid date in the #{source}."
-        )
-      end
+      @doc_date ||=
+        if data.key?("date")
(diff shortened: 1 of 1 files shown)


ashmaroli:
closing as this will lead to more problems than benefits..

pathawks:
What about a hashtable that maps strings to memoized dates? Hashing a string should be cheaper than parsing a date, right?

ashmaroli:
Hashing a date string won't check if the date is valid.. e.g. `2018-13-32` can be hashed but wont pass through `Time.parse()`

..or I missed your point entirely.. please provide a minimal example.. :grin:

pathawks:
Precondition: Empty hashtable

1. Check to see if string is in hashtable. If so, return the date stored with that key.
2. Otherwise, parse time from string. If invalid, complain.
3. If the date _is_ valid, store it in hashtable using the original string as the key. That specific string will never need to be parsed via `Time.parse` again.

ashmaroli:
Interesting.. :heart:

PR #6740, opened 2018-02-04
Access document permalink attribute efficiently

ashmaroli:
Currently `Renderer#permalink_ext` calls `document.permalink` 3 times per document instance which results in accessing a *non-memoized method* involving conditional traversing,,
https://github.com/jekyll/jekyll/blob/0f2c27bcb0a491a1eca0c2f801139992377bd6e2/lib/jekyll/document.rb#L195-L201

While a better approach would be to additionally cache the `permalink` value within the `Document` instance itself, I'm deferring that for cases that may be altering the permalink via a `:post_render` hook using plugins..

### DIFF
--- a/lib/jekyll/renderer.rb
+++ b/lib/jekyll/renderer.rb
@@ -240,8 +240,9 @@ module Jekyll
 
     private
     def permalink_ext
-      if document.permalink && !document.permalink.end_with?("/")
-        permalink_ext = File.extname(document.permalink)
+      document_permalink = document.permalink
+      if document_permalink && !document_permalink.end_with?("/")
+        permalink_ext = File.extname(document_permalink)
         permalink_ext unless permalink_ext.empty?
       end
     end


ashmaroli:
> Why approach this differently with a temporary variable?

I mentioned my doubts in the OP. In case you missed it, I elected for this accommodate for a scenario where a plugin would be altering the document's permalink via a `:post_render` hook..
(Just being cautious..)

pathawks:
I don't understand what this is trying to do. Is `document.permalink` not constant time access? `document` is not a Hash...

I really don't know enough about Ruby to say for sure, but _if_ this offers any improvement, I would be surprised if it was at all significant.

ashmaroli:
@pathawks Since you asked..
every time an instance calls `document.permalink`, the sequence of `data && data.is_a?(Hash) && data["permalink"] ` is executed.. which (according to me) is kinda expensive call because of the checks involved..
Now ideally, a Rubyist aiming for gain in performance would simply address this situation using memoization.. but here, I opted to defer memoization and implement the next best thing being (strangulate before Document instance is invoked..) to cache the value of `document.permalink` in a *local variable*

pathawks:
@ashmaroli I follow.

Sorry, I didn’t thoroughly read your original post. I understand now why this is helpful. :+1:

oe:
@jekyllbot: merge +dev

PR #6741, opened 2018-02-04
Cache the list of documents to be written

ashmaroli:
To avoid multiple traversal of the `Site#documents` array using `Array#select` and `Document#write?`
https://github.com/jekyll/jekyll/blob/0f2c27bcb0a491a1eca0c2f801139992377bd6e2/lib/jekyll/site.rb#L317-L324
https://github.com/jekyll/jekyll/blob/0f2c27bcb0a491a1eca0c2f801139992377bd6e2/lib/jekyll/site.rb#L310-L315

### DIFF
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -311,7 +311,7 @@ module Jekyll
     #
     # Returns an Array of Documents which should be written
     def docs_to_write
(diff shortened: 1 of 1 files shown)


DirtyF:
@jekyllbot: merge +minor

PR #6742, opened 2018-02-04
Cache the value of unescaped document / page url string

ashmaroli:
To avoid unnecessary calls to [`Addressable::URI.unencode()`](https://github.com/sporkmonger/addressable/blob/0a0e96acb17225f9b1c9cab0bad332b448934c9a/lib/addressable/uri.rb#L401-L451) for the same value of `url`, when `Document#destination` is called multiple times

### DIFF
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -188,8 +188,8 @@ module Jekyll
     def unescaped_url
       @url_stash ||= {}
 
-      @unescaped_url = nil unless @url_stash.key?(url)
-      @unescaped_url ||= @url_stash[url] = URL.unescape_path(url)
+      return @url_stash[url] if @url_stash.key?(url)
+      @url_stash[url] = URL.unescape_path(url)
     end
   end
 end
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -186,7 +186,10 @@ module Jekyll
 
     private
     def unescaped_url
-      @unescaped_url ||= URL.unescape_path(url)
+      @url_stash ||= {}
+
+      @unescaped_url = nil unless @url_stash.key?(url)
+      @unescaped_url ||= @url_stash[url] = URL.unescape_path(url)
     end
   end
 end
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -155,7 +155,7 @@ module Jekyll
     #
     # Returns the destination file path String.
     def destination(dest)
-      path = site.in_dest_dir(dest, URL.unescape_path(url))
+      path = site.in_dest_dir(dest, unescaped_url)
       path = File.join(path, "index") if url.end_with?("/")
       path << output_ext unless path.end_with? output_ext
       path
@@ -182,6 +182,11 @@ module Jekyll
 
     def write?
       true
+    end
+
+    private
+    def unescaped_url
+      @unescaped_url ||= URL.unescape_path(url)
     end
   end
 end
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -222,7 +222,7 @@ module Jekyll
     # Returns the full path to the output file of this document.
     def destination(base_directory)
       dest = site.in_dest_dir(base_directory)
-      path = site.in_dest_dir(dest, URL.unescape_path(url))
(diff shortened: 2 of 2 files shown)


ashmaroli:
I now see that we have the same code in `Page` as well..

ashmaroli:
> If url is modified

@parkr how would `url` be modified since it involves a memoized instance variable and there is no `attr_accessor` or `attr_writer` defined for `url`..?

parkr:
@ashmaroli In a hook plugin, I could do `document.data[“permalink”] = “/foo”` or `document.data[“title”] = “foo”’` and that should change the URL.

ashmaroli:
> In a hook plugin, I could do `document.data[“permalink”] = “/foo”`

Would it be irresponsibility on my part if I were to leave this as is by stating that if a plugin plans to change a document's `url`, its on plugin thenceforth to reset `@unescaped_url` to `nil` before altering the document's data..?

mattr-:
> Would it be irresponsibility on my part if I were to leave this as is by stating that if a plugin plans to change a document's url, its on plugin thenceforth to reset @unescaped_url to nil before altering the document's data..?

The plugin shouldn't know about or be messing with the `@unescaped_url` variable at all, so I'm gonna say 'yes' here. 🙂

ashmaroli:
I like some help to implement "nullify the instance_variable if `url` value changes"

ashmaroli:
> In a hook plugin, I could do `document.data[“permalink”] = “/foo”` or `document.data[“title”] = “foo”’` and that should change the URL.

@parkr  This won't redefine the value of `url` even on `master`..

lib/jekyll/page.rb
```diff
@@ -183,5 +183,13 @@ def trigger_hooks(hook_name, *args)
...
       true
     end
+
+    private
+    def unescaped_url
+      @url_stash ||= {}
+
+      @unescaped_url = nil unless @url_stash.key?(url)
+      @unescaped_url ||= @url_stash[url] = URL.unescape_path(url)
+    end
```
pathawks: This method is too tricky.

Can you explain to me in English what this method is doing? There must be a better way to go about this.

ashmaroli: How / why is this method risky?

What this method is doing:
For a straight-forward build session, `url` method returns **a memoized value** computed for the `Page` instance. `unescaped_url` takes that `url` and returns **a memoized unencoded URI** via `Addressable`.
But Parker asked to have the `unescaped_url` *break memoization* when `url` itself breaks memoization via a hook plugin..
So now the `unescaped_url` assigns itself the unencoded uri cached in `url_stash` based on the given `url`..


In English, first check if `url_stash` has a key that equals to the value of `url`. If the key exists, then proceed, otherwise ensure the `unescaped_url` has been reset to `nil`. In the following step, if the value of `unescaped_url` is either `nil` or `false`, then assign it the value from the `url_stash` corresponding to the `url` as key, (the value) being in turn computed by `URL.unescape_path` and the current value of `url`

pathawks: > In English, first check if url_stash has a key that equals to the value of url. If the key exists, then proceed, otherwise ensure the unescaped_url has been reset to nil. In the following step, if the value of unescaped_url is either nil or false, then assign it the value from the url_stash corresponding to the url as key, (the value) being in turn computed by URL.unescape_path and the current value of url

You are speaking Ruby 😏

The method is not _risky_, it is _tricky_; it takes a _lot_ of mental energy to figure out what it is doing.

Is this equivalent?

```ruby
private
def unescaped_url
  return url_stash[url] if url_stash.key?(url)
  url_stash[url] = URL.unescape_path(url)
end

private
def url_stash
  @url_stash ||= {}
end
```

ashmaroli: yep.. its equivalent..
I had initially implemented something similar, but then Rubocop complained of having too many lines in the class `Jekyll::Document` (305 of 300 IIRC)
Perhaps when we decide to remove the "extra `private` keywords, I can split the method as per your suggestion..

ashmaroli: @pathawks I've pushed a commit to improve the readability like you suggested.. but I'm using instance variables as I'm unable to even define a private `attr_reader :url_stash` without having Rubocop flag the `Jekyll::Document` class for having too many lines.. `[301/300]`

PR #6744, opened 2018-02-04
Add 'jekyll-fontello' to plugins

ericcornelissen:
Added a small plugin I made to download a webfont from [Fontello](https://github.com/fontello/fontello) to the list of plugins, generators.


pathawks:
@DirtyF I feel the same way.
Perhaps we should state that explicitly, and remove existing Gists from our listing?

DirtyF:
>  Perhaps we should state that explicitly, and remove existing Gists from our listing?

💯  I'll happily do that in a dedicated PR.

oe:
@ericcornelissen If you want to add this plugin, feel free to copy the changes into a separate repository, release it as a gem, and then open another pull request!

ericcornelissen:
@oe Sure! Just went with a Gist since I saw a couple others do the same 🙂

PR #6748, opened 2018-02-06
WIP: Add basic benchmarking to CI

pathawks:
This will build a basic temp site with the current build (probably a pull request), and again with the latest `master` branch. This will allow us to see if something has an impact on performance.

Ideally it would be nice to actually compare the times between the two, and perhaps fail the build if the second one is slower (allowing for some epsilon).

Only opening this PR so early to get a conversation going about how we'd like this to work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,8 @@ matrix:
       env: TEST_SUITE=default-site
     - rvm: *ruby1
       env: ROUGE_VERSION=1.11.1 # runs everything with this version
+    - rvm: *ruby1
+      env: TEST_SUITE=benchmark
   exclude:
     - rvm: *jruby
       env: TEST_SUITE=cucumber
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ matrix:
       env: ROUGE_VERSION=1.11.1 # runs everything with this version
     - rvm: *ruby1
       env: TEST_SUITE=benchmark
+      install: true
   exclude:
     - rvm: *jruby
       env: TEST_SUITE=cucumber
--- a/script/benchmark
+++ b/script/benchmark
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+
+set -e
+
+function benchmark() {(
+  cd tmp/bench-site
+  bundle update
+  time bundle exec jekyll build
+)}
+
+rm -rf tmp/benchmark tmp/bench-site
+
+if ! [[ -d tmp ]]; then
+  mkdir tmp
+fi
+
+if ! [[ -d tmp/bench-site ]]; then
+  mkdir tmp/bench-site
+fi
+
+git clone --depth 1 https://github.com/jekyll/benchmarking.git tmp/benchmark
+
+tmp/benchmark/generate_site.sh 1000 tmp/bench-site
+
+cat > tmp/bench-site/Gemfile << EOF
+gem "jekyll", :path => "."
+gem "minima"
+EOF
+
+benchmark
+
+cat > tmp/bench-site/Gemfile << EOF
+gem "jekyll", :github => "jekyll/jekyll", branch: "master"
+gem "minima"
+EOF
+
+benchmark
--- a/script/benchmark
+++ b/script/benchmark
@@ -8,15 +8,18 @@ function benchmark() {(
   time bundle exec jekyll build
 )}
 
+echo "travis_fold:start:setup"
 rm -rf tmp/benchmark tmp/bench-site
 mkdir -p tmp/bench-site
-
 git clone --depth 1 https://github.com/jekyll/benchmarking.git tmp/benchmark
+echo "travis_fold:end:setup"
 
+echo "travis_fold:start:generate_site.sh"
 tmp/benchmark/generate_site.sh 1000 tmp/bench-site
+echo "travis_fold:end:generate_site.sh"
 
 cat > tmp/bench-site/Gemfile << EOF
-gem "jekyll", :path => "."
+gem "jekyll", :path => "../.."
 gem "minima"
 EOF
 
--- a/script/benchmark
+++ b/script/benchmark
@@ -9,14 +9,7 @@ function benchmark() {(
 )}
 
 rm -rf tmp/benchmark tmp/bench-site
-
-if ! [[ -d tmp ]]; then
-  mkdir tmp
-fi
-
-if ! [[ -d tmp/bench-site ]]; then
-  mkdir tmp/bench-site
-fi
+mkdir -p tmp/bench-site
 
 git clone --depth 1 https://github.com/jekyll/benchmarking.git tmp/benchmark
 
--- a/script/benchmark
+++ b/script/benchmark
@@ -12,6 +12,14 @@ function benchmark() {(
   time bundle exec jekyll build
 )}
 
+function gemfile() {
+  cat > tmp/bench-site/Gemfile << EOF
+  source "https://rubygems.org"
+  gem "jekyll", $@
+  gem "minima"
+EOF
+}
+
 echo "travis_fold:start:setup"
 rm -rf tmp/benchmark tmp/bench-site
 mkdir -p tmp/bench-site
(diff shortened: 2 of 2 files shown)


script/benchmark
```diff
@@ -0,0 +1,24 @@
...
+function benchmark() {(
+  cd tmp/bench-site
+  bundle update
+  time bundle exec jekyll build
+)}
+
+rm -rf tmp/benchmark tmp/bench-site
+git clone --depth 1 https://github.com/jekyll/benchmarking.git tmp/benchmark
+
+( cd tmp/benchmark; ./generate_site.sh 1000 ../bench-site )
```
DirtyF: ./generate_site.sh: 2: ./generate_site.sh: Syntax error: "(" unexpected

script/benchmark
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -e
+
+function benchmark() {(
+  cd tmp/bench-site
```
ashmaroli: Where are we running a `jekyll new tmp/bench-site` ?

script/benchmark
pathawks: I don’t know what this is about. It works on my local Bash.
I’ll keep playing.

script/benchmark
```diff
@@ -0,0 +1,30 @@
...
+git clone --depth 1 https://github.com/jekyll/benchmarking.git tmp/benchmark
+
+tmp/benchmark/generate_site.sh 1000 tmp/bench-site
+
+cat > tmp/bench-site/Gemfile << EOF
+gem "jekyll", :path => "."
+gem "minima"
+EOF
+
+benchmark
```
DirtyF: why call it here?

script/benchmark
```diff
@@ -0,0 +1,30 @@
...
+
+cat > tmp/bench-site/Gemfile << EOF
+gem "jekyll", :github => "jekyll/jekyll", branch: "master"
```
DirtyF: Shouldn't we add:

```
# frozen_string_literal: true 
 
source "https://rubygems.org" 
```

pathawks: OH!!!! 💡

pathawks: `frozen_string_literal` is not a concern here, but `source` is the reason everything kept failing on Travis but not locally 😝

pathawks: ...at least, that's what I thought.
https://travis-ci.org/jekyll/jekyll/jobs/338255064#L726

script/benchmark
```diff
@@ -0,0 +1,35 @@
...
+)}
+
+echo "travis_fold:start:setup"
+rm -rf tmp/benchmark tmp/bench-site
+mkdir -p tmp/bench-site
+git clone --depth 1 https://github.com/jekyll/benchmarking.git tmp/benchmark
+echo "travis_fold:end:setup"
+
+echo "travis_fold:start:generate_site.sh"
+tmp/benchmark/generate_site.sh 1000 tmp/bench-site
```
Crunch09: I'm wondering if instead of a temp site we should measure the time of the docs site to have a more realistic example?

pathawks: One problem with the docs site is that it's pretty small and always generates pretty quickly.
We need something that takes more than a few seconds to generate so we can really see the difference in build times.

Crunch09: Good point, makes total sense then 👍

script/benchmark
DirtyF: > Gems in the groups benchmark, site and development were not installed.

pathawks:
It actually runs!!! :tada:

It took me a long time to figure out that `BUNDLE_GEMFILE` was being overridden by Travis :rage3:

pathawks:
https://travis-ci.org/jekyll/jekyll/jobs/338287354#L635-L636

> ```text
> this				    master
> 214 seconds			    211 seconds
>```

pathawks:
With 10,000 posts, it takes just over 9 minutes to build the current PR and the master branch.

Travis has a limit of 10 minutes for open source projects. We are very close to timing out.

I could maybe scale back to 5,000 posts?

script/benchmark
```diff
@@ -0,0 +1,59 @@
...
+  echo "100% means no performance change, lower than 100% means this is faster."
+}
+
+fold setup
+fold generate_site
+
+gemfile ":path => '../..'"
+benchmark this
+
+gemfile ":github => 'jekyll/jekyll', branch: 'master'"
```
ashmaroli: I prefer using `:git => 'https://github.com/jekyll/jekyll.git',  branch: 'master' to avoid seeing warnings from Bundler regarding the use of a non-secure protocol to fetch the commit

.travis.yml
```diff
@@ -19,6 +19,9 @@ matrix:
       env: TEST_SUITE=default-site
     - rvm: *ruby1
       env: ROUGE_VERSION=1.11.1 # runs everything with this version
+    - rvm: *ruby1
```
ashmaroli: crude benchmarks run on a test repo has shown that Jekyll's build times are considerably faster on Ruby 2.5 in comparison with older rubies..
The trend observed is..:
```
Ruby 2.5 < Ruby 2.4 < Ruby 2.3 < Ruby 2.2
```

pathawks: But is there a change that would make Jekyll _faster_ on one and _slower_ on another? Seems unlikely.
At least to start, it seems like we should be focusing our limited efforts on optimizing for performance with latest Ruby.

script/benchmark
pathawks: I think I must have turned off that warning on my local machine 😜
Sure, I can change this :+1:

.travis.yml
ashmaroli: The observations are that the time difference is greater on older rubies.. So, at times there is a chance that our decision to ship a change be wrongly influenced by a small improvement in build time on Ruby 2.5..

The observations are at https://travis-ci.org/ashmaroli/related-posts-test if you need a reference

script/benchmark
```diff
@@ -0,0 +1,59 @@
...
+}
+
+function display() {
+  THIS=$(tail -n 1 tmp/this.txt)
+  MASTER=$(tail -n 1 tmp/this.txt)
+  DIFF=$(echo "scale=4;($THIS-$MASTER)/($MASTER/100)" | bc)
+  pr -m -t tmp/this.txt tmp/master.txt
+  echo "This is $DIFF% the speed of master"
+  echo "100% means no performance change, lower than 100% means this is faster."
+}
```
ashmaroli: Can we have the display *beautified* a bit..?
Like, for example:
```
      Generating... 
                    done in 213.682 seconds.

--------------------------
this		|	    master
--------------------------
212			|	    214

This is 0% the speed of master
100% means no performance change, lower than 100% means this is faster.
```
(Optionally, adding color codes to indicate `improvement`, `same-state`, regression` would be awesome as well.. )

pathawks: Added some color just for you :beers:

ashmaroli: Thank you very much, Pat :smiley:

DirtyF:
> This is 3.04% slower than master

This PR isn't supposed to be slower than master 🤔

ashmaroli:
> epsilon should be quiet large :thinking: 

Could it be because the `benchmark()` function includes the `git clone ..` and `bundle update` steps?

pathawks:
@ashmaroli Look again. The timer doesn’t start until _after_ `bundle update`

ashmaroli:
I restarted the build.. it now shows `This is 5.24% slower than master`

pathawks:
I wonder if it would be better to run a number of smaller builds instead of one (of each) big build

ashmaroli:
> I wonder if it would be better to run a number of smaller builds instead

I'd prefer that.. benchmark about 5000 posts on each of the Ruby version in the matrix..

pathawks:
I mean in this same build. Run each one a few times to account for weird context switching weirdness.

Otherwise, it seems maybe Travis isn’t well suited for benchmarking.

DirtyF:
> Run each one a few times to account for weird context switching weirdness.

An average on at least 3 builds would be already more significant I guess. Even when I test locally I get different build times on the same branch, we can't blame Travis for that.

pathawks:
> Travis isn’t well suited for benchmarking.

PR #6750, opened 2018-02-06
Added Premonition plugin to liste of plugins

lazee:
Added Premonition plugin to list of plugins


oe:
@jekyllbot: merge docs

PR #6755, opened 2018-02-08
Avoid creating unnecessary Renderer instances

ashmaroli:
Currently during a build process,
  - for every `Page` object, two `Renderer` instances are created:
    - via `Site#render_regenerated`
    - via `Convertible#_renderer`
  - for every `Document`object, *numerous* `Renderer` instances are created:
    - via `Site#render_regenerated`
    - via multiple calls to `Document#output_ext`

The patch attempts to minimize the Renderer instances created unnecessarily

Note: Has minimal affect on the total build time for a Site even though there is a reduction in net allocated memory

### DIFF
--- a/lib/jekyll/convertible.rb
+++ b/lib/jekyll/convertible.rb
@@ -216,7 +216,7 @@ module Jekyll
       Jekyll.logger.debug "Post-Render Hooks:", self.relative_path
       Jekyll::Hooks.trigger hook_owner, :post_render, self
     ensure
-      renderer = nil # this will allow the modifications above to disappear
+      @renderer = nil # this will allow the modifications above to disappear
     end
 
     # Write the generated page file to the destination directory.
--- a/lib/jekyll/convertible.rb
+++ b/lib/jekyll/convertible.rb
@@ -80,7 +80,7 @@ module Jekyll
     #
     # Returns the transformed contents.
     def transform
-      _renderer.convert(content)
+      renderer.convert(content)
     end
 
     # Determine the extension depending on content_type.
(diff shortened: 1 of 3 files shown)


lib/jekyll/convertible.rb
```diff
@@ -244,12 +244,12 @@ def [](property)
       end
     end
 
-    private
-
-    def _renderer
-      @_renderer ||= Jekyll::Renderer.new(site, self)
+    def renderer(payload = nil)
```
mattr-: If we want to keep this from being exposed publically, we could make this protected: ¯\\_(ツ)\_/¯

mattr-:
This looks great!! Thank you!! 🥇

lib/jekyll/document.rb
```diff
@@ -273,6 +273,11 @@ def read(opts = {})
       end
     end
 
+    # The Renderer instance for this document
+    def renderer(payload = nil)
+      @renderer ||= Jekyll::Renderer.new(site, self, payload)
```
parkr: By accepting the payload but memoizing, you're breaking a contract with the user that the input will affect the return value.

```text
>> doc.renderer
#<Jekyll::Renderer@12345 @payload=nil>
>> doc.renderer(my_payload)
#<Jekyll::Renderer@12345 @payload=nil>
>> !!! I should get back a renderer that has `my_payload` set, not nil!
```

One possible solution to this is to:

```ruby
def renderer(payload = nil)
  @renderer ||= Jekyll::Renderer.new(site, self, payload)
  @renderer.tap { |r| r.payload = payload }
end
```

This will return the Renderer instance with the given payload.

ashmaroli: nice..  :+1:

ashmaroli:
closing as this could lead to unforeseen issues in the future..
(vaguely relayed by the current failing test..)

PR #6756, opened 2018-02-08
Return early unless there is a valid front matter default set

ashmaroli:
Since `FrontmatterDefaults#find` is called as part of `:default_proc` of a `Page` and `Document` object's `data` hash, it helps to return early if the configured front matter defaults are invalid or not configured at all (empty array)

### DIFF
--- a/lib/jekyll/frontmatter_defaults.rb
+++ b/lib/jekyll/frontmatter_defaults.rb
@@ -198,7 +198,7 @@ module Jekyll
     # Returns an array of hashes
     def valid_sets
       sets = @site.config["defaults"]
-      return [] unless sets.is_a?(Array)
+      return [] if sets.empty? || !sets.is_a?(Array)
 
       sets.map do |set|
         if valid?(set)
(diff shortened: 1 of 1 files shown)


pathawks:
Is this `3.8` or `4.0`?

ashmaroli:
It can be shipped in `v3.8.0` but I don't want this to be merged right now because `#valid_sets` is currently an unmemoized method (the associated method comments also document why it wasn't cached by design).
So this proposal is more harmful than beneficial (as of this writing). But can become beneficial when `#valid_sets` can involve memoization..

pathawks:
> But can become beneficial when #valid_sets can involve memoization..

Is there a PR for that? What are we waiting on?

ashmaroli:
> Is there a PR for that?

Nope..

> What are we waiting on?

We aren't. This is not included in any milestone and has a `do-not-merge` label..

pathawks:
It has an approved review, and is a Pull Request. You are not requesting that it be pulled?

Can this be closed, and you keep the branch to reopen if circumstances change?

PR #6757, opened 2018-02-08
Add 'jekyll-fontello' to plugins

ericcornelissen:
Added a small plugin I made to download a webfont from [Fontello](https://github.com/fontello/fontello) to the list of plugins, generators. Update from #6744, now a Ruby gem instead of a GitHub Gist.


oe:
@jekyllbot: :shipit: +docs

PR #6759, opened 2018-02-09
Create a single instance of PostReader per site

ashmaroli:
Currently, for every valid directory (based on the `EntryFilter`) traversed by `Reader`, a new instance of `PostReader` is created (with the same site instance) to retrieve posts and drafts from that directory.

This PR proposes to use just one `PostReader` instance to retrieve posts and drafts from all valid directories.


DirtyF:
@jekyllbot: merge +minor

PR #6760, opened 2018-02-09
Profile various stages of a site's build process

ashmaroli:
Outputs a table showing how much time has been spent on each stage in `Site#process`

```
Build Process Summary:
----------------------

  PHASE      |      TIME
-------------+-------------
  RESET      |    0.0000
  READ       |    0.3423
  GENERATE   |    0.0483
  RENDER     |   12.0781
  CLEANUP    |    0.0763
  WRITE      |    0.1856
-------------+-------------
 TOTAL TIME  |   12.7306
```


lib/jekyll/site.rb
```diff
@@ -67,13 +67,8 @@ def config=(config)
...
     def process
-      reset
-      read
-      generate
-      render
-      cleanup
-      write
-      print_stats if config["profile"]
+      return profile_process if config["profile"]
+      process_methods.each { |method| public_send(method) }
```
oe: I'm not a big fan of impacting readability of this method like this. This has historically been one of the most crucial parts to understanding Jekyll's build procedure, I don't think we should change it in this drastic of a manner.

mattr-:
Is there a way that we can get this out of the rendering classes and into their own classes so that we don't pollute the core function of those classes?

ashmaroli:
> Is there a way that we can get this out of the rendering classes and into their own classes

There is an active proposal where @parkr adds a `Jekyll::Profiler` class.. We can move this into that class if Parker is fine with the arising conflict..

ashmaroli:
@mattr-  Moved the methods for profiling various stages of the build process into (a new class) `Jekyll::Profiler`

DirtyF:
How is this helpful and actionable for end-users?

```
Build Process Summary:

| PHASE      |   TIME |
+------------+--------+
| RESET      | 0.0003 |
| READ       | 0.3453 |
| GENERATE   | 0.0335 |
| RENDER     | 5.7318 |
| CLEANUP    | 0.1145 |
| WRITE      | 0.1663 |
+------------+--------+
| TOTAL TIME | 6.3917 |
```

ashmaroli:
> How is this helpful

In the past, we had received queries from users stating that the total time calculated from the
`--profile` table does not tally with the total time output at the end of a build (e.g. `done in 1.215s`). We had to then inform the users that the `--profile` table only logs the `:render` phase. The additional time would be from IO bound stages such as reading and writing files and in some cases from multiple generators.

> actionable for end-users?

Not quite for novice users. But at least they'd have greater intel on their site.
For intermediate to advanced users, the additional info might help them optimize their `_plugins`.

DirtyF:
OK so when it's something more than 80% of users don't care, I would put it behind an option like `--profile --verbose`

ashmaroli:
> an option similar to `--profile --verbose`

I'm not so keen on introducing another CLI switch at the moment, though..

DirtyF:
@jekyll: merge +minor

PR #6763, opened 2018-02-09
Cache and retrieve escaped path components

ashmaroli:
This is to avoid calling `Addressable::URI.encode` on the same path components.

e.g. `2018-02-09-test-post.md` and `2018-02-08-first-post.md` would send `2018` and `02` twice to be encoded / escaped..

### DIFF
--- a/lib/jekyll/url.rb
+++ b/lib/jekyll/url.rb
@@ -131,7 +131,6 @@ module Jekyll
     # Returns the escaped path.
     def self.escape_path(path)
       return cache[path] if cache.key?(path)
-      raw_path = path
 
       # Because URI.escape doesn't escape "?", "[" and "]" by default,
       # specify unsafe string (except unreserved, sub-delims, ":", "@" and "/").
@@ -143,11 +142,8 @@ module Jekyll
       #   pct-encoded   = "%" HEXDIG HEXDIG
       #   sub-delims    = "!" / "$" / "&" / "'" / "(" / ")"
       #                 / "*" / "+" / "," / ";" / "="
-      path = Addressable::URI.encode(path)
-      escaped_path = path.encode("utf-8").sub("#", "%23")
-      cache[raw_path] = escaped_path
-
-      escaped_path
+      encoded_path = Addressable::URI.encode(path).encode("utf-8")
+      cache[path] = encoded_path.sub("#", "%23")
     end
 
     # Unescapes a URL path segment
--- a/lib/jekyll/url.rb
+++ b/lib/jekyll/url.rb
@@ -130,6 +130,9 @@ module Jekyll
     #
     # Returns the escaped path.
     def self.escape_path(path)
+      return cache[path] if cache.key?(path)
+      raw_path = path
+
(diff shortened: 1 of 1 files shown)


lib/jekyll/url.rb
```diff
@@ -141,7 +144,10 @@ def self.escape_path(path)
       #   sub-delims    = "!" / "$" / "&" / "'" / "(" / ")"
       #                 / "*" / "+" / "," / ";" / "="
       path = Addressable::URI.encode(path)
-      path.encode("utf-8").sub("#", "%23")
+      escaped_path = path.encode("utf-8").sub("#", "%23")
+      cache[raw_path] = escaped_path
+
+      escaped_path
```
pathawks: Are all the transient variables needed?

```ruby
  escaped_path = Addressable::URI.encode(path).encode("utf-8")
  cache[path] = escaped_path.sub("#", "%23")
end
```

This seems readable enough to me.

ashmaroli: should I not return `escaped_path` (before the closing end)?

pathawks: `cache[path] = escaped_path.sub("#", "%23")` evaluates to the right hand side of the assignment. This means:

```ruby
cache[raw_path] = escaped_path
escaped_path
```

is the same as

```ruby
cache[raw_path] = escaped_path
```

Too weird?

ashmaroli: a bit, yes. But I like how DRY it is..
just wasn't sure if that's the approach to take..

mattr-:
This looks good. Just one question, since I'm not as familiar with this code as I should be: Is there only a single filter object per site? My concern here is that our caching won't be as effective or non-helpful if this is just per post.

ashmaroli:
> Is there only a single filter object per site?

I do not understand why you're referring to a filter object here..
I ran a call-trace (`puts caller(0..3)`) on the method to confirm.. All calls to this method originate from within the `Jekyll::URL` class..

Or I missed your question by a couple of miles..

lib/jekyll/url.rb
```diff
@@ -157,5 +158,9 @@ def self.escape_path(path)
     def self.unescape_path(path)
       Addressable::URI.unencode(path.encode("utf-8"))
     end
+
+    def self.cache
```
parkr: If this is only used for `escape_path` what do you think about calling this `escape_path_cache`?

ashmaroli: make sense.. especially since the cache is defined in the Singleton class, the cache would be indifferent to multiple instances of the `URL` class..

ashmaroli:
The cache is currently `public` so that it can be reset to `nil` from elsewhere if required..

lib/jekyll/url.rb
ashmaroli: on second thoughts.. what do you think about caching *both* escaped and unescaped paths..?
i.e. stash results from both `self.escape_path` and `self.unescape_path` in the same cache...
```ruby
cache["escaped"][path]   = ...
cache["unescaped"][path] = ...
```

PR #6765, opened 2018-02-10
Update windows.md

dsolodow:
Had to add dh-autoreconf to resolve compile issues with 'sudo gem install jekyll'


DirtyF:
@jekyllbot: :shipit: +docs

PR #6766, opened 2018-02-11
Return empty content without attempting conversion

ashmaroli:


pathawks:
I wonder if there are any renderers that accept an empty string only as a trigger for something else.

ashmaroli:
> I wonder if there are any renderers that accept an empty string only as a trigger for something else.

Shipping this into the wild would be the way to be sure.. :smiling_imp: 
That aside, triggering a converter with an empty string seems sad.. no?

pathawks:
> That aside, triggering a converter with an empty string seems sad.

It wouldn't be the strangest exploitation of undefined behavior I've seen on a Jekyll site 😨

PR #6768, opened 2018-02-12
Useless privates removed

Ana06:
`private` sets the default visibility for subsequently defined methods to private, so it is not needed to repeat it for every method. :bowtie:


pathawks:
At one point, including the `private` above each method was part of our unspoken style guide.

Perhaps we should write a Rubocop check to enforce this?

ashmaroli:
@Ana06 The reason we follow this style is when there are numerous private methods, the reader (person who is reading the code) would have to either scroll his page up to check if the method is truly `private`..

Also, it helps managing PRs that involve adding / changing / removing private methods easily because there's a `private` keyword right above the method.., otherwise, the maintainer will have to spend additional time to ensure the PR isn't breaking any semver rules.. (e.g. private methods can have their signature modified without penalty..)

Ana06:
@ashmaroli, @pathawks then maybe you should use the private with arguments, that at least have some effect. And this come with strings attached, what happens if you forget the private? it is having effect anyway. 😕  And you are not following your one rule in many places of the code as well. 🤓 

A Rubocop rule would be nice if this is the style you want to follow, but then I will enforce to use the prove with arguments instead.

BTW, are also the `Private: ` comments on propuse? I would aim to remove that as well... if the method is private you don't need a comment... 🤔

ashmaroli:
Yes, we do not *enforce* this style currently.. but we won't be removing existing instances of the style because of the reasons I mentioned..

`# Private: ` comments are perhaps superfluous in their respective context.. They're not visible to `rdoc` and the like..

Can you elaborate further on the "private with arguments" statement..? Thanks

Ana06:
@ashmaroli 
> Yes, we do not enforce this style currently.. but we won't be removing existing instances of the style because of the reasons I mentioned..

in many of the cases there are two or three short private methods, so that doesn't really apply... What happens with new code? So for example, if I write a new private method, what style should then I follow?

regarding the private with arguments, this is what is currently used:
```ruby
  private
  def private_method_1
  ...
  end

  private
  def private_method_2
  ...
  end
```

and this is what I propose:

```ruby
  private :private_method_1
  def private_method_1
  ...
  end

  private :private_method_2
  def private_method_2
  ...
  end
```

ashmaroli:
Personal opinion: I do not like the "private with argument" style..

/cc @jekyll/core for inputs

---

> if I write a new private method, what style should then I follow?

You can take cues from the surrounding code..
If existing private methods have the keyword above them, then you should follow suit.. otherwise leave it.. We'll add them when have Rubocop enforcing the style.

parkr:
Typically in the past we've followed convention which says the top portion of a class is for its public functions, then you declare the `private` keyword once, then all the private methods go below that. 

Whatever decision is made should probably have a Rubocop checker written so we keep it consistent across the whole code base.

DirtyF:
I see no reason not to merge this.
Let's create a dedicated issue for automation with Rubocop

pathawks:
> Let's create a dedicated issue for automation with Rubocop
 
we have had other PRs that specifically _add_ these `private` statements.
Rather than going back and forth adding and removing, let’s tell Rubocop how it should be done, and let it enforce the style for us.

ashmaroli:
@Ana06 I have enabled Rubocop to enforce single `private` keyword per class.. go ahead and remove the remaining `private`s Rubocop will now detect.. :+1:

.rubocop.yml
```diff
@@ -57,8 +57,6 @@ Lint/UnneededRequireStatement:
   Enabled: false
 Lint/UnreachableCode:
   Severity: error
-Lint/UselessAccessModifier:
-  Enabled: false
```
pathawks: 🎉

nickskalkin:
@parkr @pathawks Hi! Please take a look at this PR #7004 
It enforces us to use one convention

nickskalkin:
Please, take a look at #7004

I think that explicit privates are preferable

nickskalkin:
Can't write long explanation now, but check out this blogpost please https://anti-pattern.com/explicit-privates

Ana06:
@nickskalkin 

> I think that explicit privates are preferable


the default in Rubocop is without them: https://rubocop.readthedocs.io/en/latest/cops_lint/#lintuselessaccessmodifier


I am not sure anymore if we want to keep them or remove them :confused:

pathawks:
> I am not sure anymore if we want to keep them or remove them

Sorry; I didn't see #7004 until after I had already approved this PR, and I can't figure out how to dismiss my review.

I would prefer to see `private` _everywhere_ rather than in _fewer_ places. This makes it much easier--when reviewing PRs--to see if any new public methods have been added or removed or changed. Prefixing every method with `private` removes any ambiguity.

At any rate, this needs a decision from @jekyll/core

DirtyF:
@Ana06 Thank you so much for opening the debate. I a big fan of minimalism but I admit I was sold _after_ reading the blog post. It's not the first time we won't stick to a Rubocop's default.

And thank to you both: now we have a way to automate this and that's the most important for us: being easy to read and being consistent.

DirtyF:
@pathawks I'm in favor on closing this in favor of #7004

ashmaroli:
Now that I see #7004 I think its the better way to proceed with.

Inline `private` doesn't *add* to the total no. of lines in a class as counted by Rubocop and at the same time states the intended accessibility/visibility of the method.

DirtyF:
@Ana06 so, after devating, Jekyll will go the Ruby Way, 🙏  please fix [remaining offenses](https://travis-ci.org/jekyll/jekyll/jobs/376946470#L614) reported by Rubocop and we'll be able to finally merge this. Sorry about the indecision.

Ana06:
ok @DirtyF! I'll fix the offenses ASAP :wink:

DirtyF:
@jekyllbot: merge +dev

Thanks @pathawks for making Travis green again 🇺🇸

DirtyF:
@Ana06 🎉

PR #6770, opened 2018-02-13
Remove unnecessary Jekyll::Page constant

ashmaroli:
Once we drop support for Ruby 2.2 in Jekyll 4.0, having the magic comment `# frozen string literal: true` is all that's needed to avoid creating multiple copies of the same string.

In that event, the constant `Jekyll::Page::FORWARD_SLASH` will be unnecessary.

### DIFF
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -11,8 +11,6 @@ module Jekyll
 
     alias_method :extname, :ext
 
-    FORWARD_SLASH = "/".freeze
-
     # Attributes for Liquid templates
     ATTRIBUTES_FOR_LIQUID = %w(
       content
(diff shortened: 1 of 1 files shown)


lib/jekyll/page.rb
```diff
@@ -64,11 +62,11 @@ def initialize(site, base, dir, name)
     #
     # Returns the String destination directory.
     def dir
-      if url.end_with?(FORWARD_SLASH)
+      if url.end_with?("/")
         url
       else
         url_dir = File.dirname(url)
-        url_dir.end_with?(FORWARD_SLASH) ? url_dir : "#{url_dir}/"
+        url_dir.end_with?("/") ? url_dir : "#{url_dir}/"
```
crispgm: Should we replace `"#{url_dir}/"` with "#{url_dir}#{FORWARD_SLASH}"?

ashmaroli: er.. um.. I'm trying to do away with the constant here..
Why would I want to interpolate a single character..?

pathawks:
This is a `+dev` change that should _not_ be backported to `3.x` because it relies on Ruby 2.3.

@jekyllbot: merge +dev

PR #6769, opened 2018-02-12
 Add ordinal_date filter

Ana06:
It is normal to write things like _posted on 12th February 2018" in the blog post. Currently to achieve this, a code similar to this one is needed:

```
{% assign d = page.date | date: "%-d" %}
{% case d %}
{% when '1' or '21' or '31' %}
{{ d }}st
{% when '2' or '22' %}
{{ d }}nd
{% when '3' or '23' %}
{{ d }}rd
{% else %}{{ d }}th
{% endcase %}
{{ page.date | date: "%B" }}
{{ page.date | date: "%Y" }}
```

I think the case is common enough to introduce a method similar to `date_to_long_string` for this. :bowtie:

PR #6773, opened 2018-02-14
Allow date filters to output ordinal days

Ana06:
It is normal to write things like _posted on 12th February 2018" in the blog post. Currently to achieve this, a code similar to this one is needed:

```
{% assign d = page.date | date: "%-d" %}
{% case d %}
{% when '1' or '21' or '31' %}
{{ d }}st
{% when '2' or '22' %}
{{ d }}nd
{% when '3' or '23' %}
{{ d }}rd
{% else %}{{ d }}th
{% endcase %}
{{ page.date | date: "%B" }}
{{ page.date | date: "%Y" }}
```

I think the case is common enough to introduce it in Jekyll, by adding some parameters to the `date_to_string` and `date_to_long_string` methods.

It supports both UK and US styles. UK is the default.

I also changed @sample_date in one of the tests to cover more cases.

### DIFF
--- a/docs/_docs/templates.md
+++ b/docs/_docs/templates.md
@@ -93,6 +93,20 @@ you come up with your own tags via plugins.
     </tr>
     <tr>
       <td>
+        <p class="name"><strong>Date to String in ordinal US style</strong></p>
+        <p>Format a date to ordinal, US, short format.</p>
+      </td>
+      <td class="align-center">
+        <p>
+         <code class="filter">{% raw %}{{ site.time | date_to_string: "ordinal", "US" }}{% endraw %}</code>
+        </p>
+        <p>
+          <code class="output">Nov 7th, 2008</code>
+        </p>
+      </td>
+    </tr>
+    <tr>
+      <td>
         <p class="name"><strong>Date to Long String</strong></p>
         <p>Format a date to long format.</p>
       </td>
@@ -102,6 +116,20 @@ you come up with your own tags via plugins.
         </p>
         <p>
           <code class="output">07 November 2008</code>
+        </p>
+      </td>
+    </tr>
+    <tr>
+      <td>
+        <p class="name"><strong>Date to Long String in ordinal UK style</strong></p>
+        <p>Format a date to ordinal, UK, long format.</p>
+      </td>
+      <td class="align-center">
+        <p>
+         <code class="filter">{% raw %}{{ site.time | date_to_long_string: "ordinal" }}{% endraw %}</code>
+        </p>
+        <p>
+          <code class="output">7th November 2008</code>
         </p>
       </td>
     </tr>
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -68,22 +68,33 @@ module Jekyll
     end
 
     # Format a date in short format e.g. "27 Jan 2011".
+    # Ordinal format is also supported, in both the UK
+    # (e.g. "27th Jan 2011") and US ("e.g. Jan 27th, 2011") formats.
+    # UK format is the default.
     #
     # date - the Time to format.
+    # type - if "ordinal" the returned String will be in ordinal format
+    # style - if "US" the returned String will be in US format.
+    #         Otherwise it will be in UK format.
     #
     # Returns the formatting String.
-    def date_to_string(date)
-      time(date).strftime("%d %b %Y")
+    def date_to_string(date, type = nil, style = nil)
+      stringify_date(date, "%b", type, style)
     end
 
     # Format a date in long format e.g. "27 January 2011".
+    # Ordinal format is also supported, in both the UK
+    # (e.g. "27th January 2011") and US ("e.g. January 27th, 2011") formats.
+    # UK format is the default.
     #
-    # date - The Time to format.
+    # date - the Time to format.
+    # type - if "ordinal" the returned String will be in ordinal format
+    # style - if "US" the returned String will be in US format.
+    #         Otherwise it will be in UK format.
     #
     # Returns the formatted String.
-    def date_to_long_string(date)
-      return date if date.to_s.empty?
-      time(date).strftime("%d %B %Y")
+    def date_to_long_string(date, type = nil, style = nil)
+      stringify_date(date, "%B", type, style)
     end
 
     # Format a date for use in XML.
@@ -355,6 +366,36 @@ module Jekyll
           end
         end
         .map!(&:last)
+    end
+
+    private
+    # month_type: Notations that evaluate to 'Month' via `Time#strftime` ("%b", "%B")
+    # type: nil (default) or "ordinal"
+    # style: nil (default) or "US"
+    #
+    # Returns a stringified date or the empty input.
+    def stringify_date(date, month_type, type = nil, style = nil)
+      return date if date.to_s.empty?
+      time = time(date)
+      if type == "ordinal"
+        day = time.day
+        ordinal_day = "#{day}#{ordinal(day)}"
+        return time.strftime("#{month_type} #{ordinal_day}, %Y") if style == "US"
+        return time.strftime("#{ordinal_day} #{month_type} %Y")
+      end
+      time.strftime("%d #{month_type} %Y")
+    end
+
+    private
+    def ordinal(number)
+      return "th" if (11..13).cover?(number)
+
+      case number % 10
+      when 1 then "st"
+      when 2 then "nd"
+      when 3 then "rd"
+      else "th"
+      end
     end
 
     private
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -11,6 +11,7 @@ module Jekyll
   module Filters
     include URLFilters
     include GroupingFilters
+    include DateFilters
 
     # Convert a Markdown string into HTML output.
     #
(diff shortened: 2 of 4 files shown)


Ana06:
I closed https://github.com/jekyll/jekyll/pull/6769 by accident

@ashmaroli I have made the changes we had discussed.

Rubocop complains about the class being to long, should I move date related method to a separate module, similarly to how it is done in `url_filters.rb`?

ashmaroli:
> should I move date related method to a separate module

I don't see the reason to *not move* this to a separate module. But I'll let the other maintainers make that decision..

DirtyF:
Are there no native methods in Ruby to handle ordinal dates? What about non-english dates format? e.g,  In french you write `Jeudi 1er Mars 2018`.

ashmaroli:
> Are there no native methods in Ruby to handle ordinal dates?

none that I'm aware of.. But interesting question..

Ana06:
@DirtyF 

> Are there no native methods in Ruby to handle ordinal dates? What about non-english dates format? e.g, In french you write Jeudi 1er Mars 2018

In the [Time Ruby class](http://ruby-doc.org/core-2.5.0/Time.html) there is nothing. We could suggest them to add it, but it won't happen soon.

I think it is a bad idea to support dates in other languages in Jekyll, but what about extending this methods in [your Date plugin](https://github.com/envygeeks/jekyll-dates)? I would be up to help. 😉 

So my suggestion is add the ordinal dates in English as it is quite general and used case, and the rest in the plugin. What do you think? 😄

DirtyF:
@Ana06 OK, I guess it's make sense to implement ordinal dates like Rails:

https://github.com/rails/rails/blob/acbcef6094d61eb8c4820295620d170743a4bd71/activesupport/lib/active_support/inflector/methods.rb#L343-L369

DirtyF:
@Ana06 II mean I looked at your implementation and it seems to match the Rail's one 👍

docs/_docs/templates.md
```diff
@@ -105,6 +105,35 @@ you come up with your own tags via plugins.
         </p>
       </td>
     </tr>
+    <tr>
+      <td>
+        <p class="name"><strong>Date to String in ordinal US style</strong></p>
+        <p>Format a date to ordinal, UK, short format.</p>
```
mattr-: Could you replace `UK` here with `US` since it looks like this is for the US format based on the context around it?

lib/jekyll/filters.rb
```diff
@@ -68,22 +68,33 @@ def slugify(input, mode = nil)
     end
 
     # Format a date in short format e.g. "27 Jan 2011".
+    # It allows to use ordinal format as well, in both UK
```
mattr-: This comment would read better if it was written thusly:
```
Ordinal format is also supported, in both the UK and US formats.
```

Please keep the example dates in this comment when you make the language change. I left them out of the above sentence just to keep that shorter.

lib/jekyll/filters.rb
```diff
@@ -68,22 +68,33 @@ def slugify(input, mode = nil)
...
+    # type - if "ordinal" the returned String will be in ordinal format
+    # style - if "US" the returned String will be in US format.
+    #         Otherwise it will be in UK format.
     #
     # Returns the formatted String.
-    def date_to_long_string(date)
-      return date if date.to_s.empty?
-      time(date).strftime("%d %B %Y")
+    def date_to_long_string(date, type = nil, style = nil)
+      generic_date_to_string(date, "B", type, style)
```
mattr-: Could you convert the magic string `B` into a constant? Same for `b` as well. Something like:
```ruby
LONG_MONTH_FORMAT = 'B'
SHORT_MONTH_FORMAT = 'b'
```

provides much more expressiveness and clarity in the code than the magic string.

lib/jekyll/filters.rb
```diff
@@ -357,6 +368,34 @@ def sort_input(input, property, order)
         .map!(&:last)
     end
 
+    private
+    # month_type: "b", "B"
+    # type: nil, "ordinal"
+    # style nil, "US"
+    def generic_date_to_string(date, month_type, type = nil, style = nil)
```
mattr-: Both `type` and `style` should default to something other than `nil` here, IMHO.

docs/_docs/templates.md
```diff
@@ -105,6 +105,35 @@ you come up with your own tags via plugins.
...
+        <p>
+         <code class="filter">{% raw %}{{ site.time | date_to_string: "ordinal", "US" }}{% endraw %}</code>
+        </p>
+        <p>
+          <code class="output">Nov 7th, 2008</code>
+        </p>
+      </td>
+    </tr>
+    <tr>
+    <tr>
```
ashmaroli: double `<tr>` here..

docs/_docs/templates.md
```diff
@@ -105,6 +105,35 @@ you come up with your own tags via plugins.
         </p>
       </td>
     </tr>
+    <tr>
+      <td>
+        <p class="name"><strong>Date to String in ordinal US style</strong></p>
+        <p>Format a date to ordinal, UK, short format.</p>
+      </td>
+      <td class="align-center">
```
ashmaroli: Please **reorder** the table rows here such that information flows gracefully. For example:

Filter Desc. | Example
--- | --- 
**Date to String** | `date_to_string`
**Date to String in UK Ordinal** | `date_to_string: "ordinal"`
**Date to String in US Ordinal** | `date_to_string: "ordinal", "US"`
**Date to Long String** | `date_to_long_string`
**Date to Long String in UK Ordinal** | `date_to_long_string: "ordinal"`
**Date to Long String in US Ordinal** | `date_to_long_string: "ordinal", "US"`

lib/jekyll/filters.rb
```diff
@@ -68,22 +68,33 @@ def slugify(input, mode = nil)
...
     # date - the Time to format.
+    # type - if "ordinal" the returned String will be in ordinal format
+    # style - if "US" the returned String will be in US format.
+    #         Otherwise it will be in UK format.
     #
     # Returns the formatting String.
-    def date_to_string(date)
-      time(date).strftime("%d %b %Y")
+    def date_to_string(date, type = nil, style = nil)
+      generic_date_to_string(date, "b", type, style)
```
ashmaroli: - instead of `generic_date_to_string()` how about `stringify_date()` ?
- instead of passing `b` and `B` as parameters, how about passing `%b` and `%B` ?

DirtyF:
@Ana06 👋 we're still interested in shipping this for the next minor version, is there anything blocking you from addressing the feedback?

Ana06:
@DirtyF 

> @Ana06 wave we're still interested in shipping this for the next minor version, is there anything blocking you from addressing the feedback?

I was waiting in case there was more feedback, I'll update this now :wink:

lib/jekyll/filters.rb
Ana06: I changed it to @ashmaroli suggestion:

> instead of passing b and B as parameters, how about passing %b and %B ?

Do you think it is already clear?

I think with that is already clear that is the use of the strftime method and IMO there is no need to add two variables for this. :smile:

lib/jekyll/filters.rb
Ana06: `nil` is equivalent to default. We can give it something more complicate, but the code will work anyway with something different. :thinking:

lib/jekyll/filters.rb
```diff
@@ -357,6 +368,34 @@ def sort_input(input, property, order)
...
+    # month_type: "b", "B"
+    # type: nil, "ordinal"
+    # style nil, "US"
+    def stringify_date(date, month_type, type = nil, style = nil)
+      return date if date.to_s.empty?
+      time = time(date)
+      if type == "ordinal"
+        day = time.day
+        ordinal_day = "#{day}#{ordinal(day)}"
+        return time.strftime("%#{month_type} #{ordinal_day}, %Y") if style == "US"
```
ashmaroli: you forgot to adjust the string interpolation here and below.. :slightly_smiling_face: 
```diff
- time.strftime("%#{month_type}..
+ time.strftime("#{month_type}..
```

Ana06: yes, I was just correcting it :wink:

Ana06:
@DirtyF I think it should be ready to merge now :wink:

lib/jekyll/filters/date_filters.rb
```diff
@@ -66,21 +64,26 @@ def date_to_rfc822(date)
...
           day = time.day
           ordinal_day = "#{day}#{ordinal(day)}"
           return time.strftime("#{month_type} #{ordinal_day}, %Y") if style == "US"
           return time.strftime("#{ordinal_day} #{month_type} %Y")
         end
+
         time.strftime("%d #{month_type} %Y")
       end
 
+      private
```
Ana06: :see_no_evil:

lib/jekyll/filters/date_filters.rb
```diff
@@ -66,21 +64,26 @@ def date_to_rfc822(date)
       end
 
       private
-      # month_type: "b", "B"
-      # type: nil, "ordinal"
-      # style nil, "US"
+      # month_type: Notations that evaluate to 'Month' via `Time#strftime` ("%b", "%B")
```
Ana06: :+1:

Ana06:
@ashmaroli should I add your correction commit to mines, so that the commit history is clean? :thinking:

lib/jekyll/filters/date_filters.rb
```diff
@@ -66,21 +64,26 @@ def date_to_rfc822(date)
...
+      #       type: nil (default) or "ordinal"
```
Ana06: to many spaces....

ashmaroli:
Feel free to update the branch as you see fit. I only pushed the commit because I didn't want to bother you unnecessarily..

Ana06:
@ashmaroli already added. I think I didn't miss anything :wink: I like having clean commit history :cupid:

ashmaroli:
> I think I didn't miss anything

no probs.. my changes seem to have been incorporated nicely.. :+1:

DirtyF:
📆 🤙 🔢 

Congrats @Ana06 for your first Jekyll contrib! 😍 

@jekyllbot: :shipit: +minor

PR #6775, opened 2018-02-15
Update docs on permalink configuration

ashmaroli:
Documents change made in https://github.com/jekyll/jekyll/commit/b6c283a4aee531a06ba6b3c9cde5e972889c7471

Other references:
  - https://github.com/jekyll/jekyll-sitemap/issues/206


DirtyF:
I feel like 99% of the time you would use `.html` maybe we should give an example use case and explain when to use `output_ext` instead?

pathawks:
@DirtyF Using `.html` doesn’t work.

pathawks:
I agree that using `.html` _should_ work, or at least display a warning. But for right now we should just make sure our docs reflect our current behavior.

DirtyF:
@jekyllbot: :shipit: +docs

pathawks:
This is a bug in the code, not just a bug in the docs.

Having a permalink template that ends with `.html` may not be ideal, but it should _absolutely_ be supported.

PR #6776, opened 2018-02-15
Codeclimate: exclude livereload.js

ashmaroli:
An asset we do not maintain shouldn't affect our Code Climate score.


oe:
@jekyllbot: :shipit: +dev

PR #6727, opened 2018-01-30
Incorporate `relative_url` filter in `link` tag

ashmaroli:
Resolves #5682 
Closes #5943 

With this, `{% link about.md %}` will return `/blog/about/` for a site configured with
 `baseurl: /blog` and `permalink: pretty`

**`BREAKING-CHANGE`**

- [x] Initial implementation
- [x] ~~Unit test/~~ Cucumber feature
~~Documentation~~

PR #6777, opened 2018-02-15
Add a cucumber feature to test link tag

ashmaroli:
Mainly to act as a control element for #6727 which passes other existing tests in spite of containing a breaking change.

### DIFF
--- a/features/step_definitions.rb
+++ b/features/step_definitions.rb
@@ -36,7 +36,7 @@ end
 Given(%r!^I have an? "(.*)" page(?: with (.*) "(.*)")? that contains "(.*)"$!) do |file, key, value, text|
   File.write(file, Jekyll::Utils.strip_heredoc(<<-DATA))
     ---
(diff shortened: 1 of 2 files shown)


features/step_definitions.rb
```diff
@@ -36,7 +36,7 @@
 Given(%r!^I have an? "(.*)" page(?: with (.*) "(.*)")? that contains "(.*)"$!) do |file, key, value, text|
   File.write(file, Jekyll::Utils.strip_heredoc(<<-DATA))
     ---
-    #{key || "layout"}: #{value || "nil"}
+    #{key || "layout"}: #{value || "none"}
```
pathawks: This is unrelated?

ashmaroli: yep totally unrelated.. thought I'd fix this silently..

DirtyF:
@jekyllbot: merge +dev

PR #6779, opened 2018-02-15
Fix documentation for `{{ page.excerpt }}`

ashmaroli:
`{{ page.excerpt }}` is *currently* meant for `Jekyll::Document` objects only, and has nothing to do with `Jekyll::Page` objects, much like `{{ page.id }}` is meant to be used for documents alone.

### DIFF
--- a/docs/_docs/posts.md
+++ b/docs/_docs/posts.md
@@ -216,10 +216,10 @@ your posts:
 {% raw %}
 ```liquid
 <ul>
-  {% for doc in site.posts %}
+  {% for post in site.posts %}
     <li>
-      <a href="{{ doc.url }}">{{ doc.title }}</a>
-      {{ doc.excerpt }}
+      <a href="{{ post.url }}">{{ post.title }}</a>
+      {{ post.excerpt }}
     </li>
   {% endfor %}
 </ul>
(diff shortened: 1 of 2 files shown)


docs/_docs/posts.md
```diff
@@ -207,18 +207,19 @@ filter by tags or any other variable created with extensions.
...
+post's data hash.
 Take the above example of an index of posts. Perhaps you want to include
 a little hint about the post's content by adding the first paragraph of each of
 your posts:
 
 {% raw %}
 ```liquid
 <ul>
-  {% for post in site.posts %}
+  {% for doc in site.posts %}
```
oe: Why change this? Makes it harder to understand for new users, IMO.

ashmaroli: .. so that they do not think that one should always just use `post` parameter for posts..

oe: But it does make more sense semantically to say `for post in posts`, doesn't it? I think this example is fine, as long as we clarify it explicitly somewhere else.

ashmaroli: okay.. I'll undo this change.

pathawks:
Could we just _add_ `excerpt` to `Page`?

Maybe default to `""` and log a warning that `excerpt` is not calculated for pages?

pathawks:
@jekyllbot: :ship: +docs

PR #6780, opened 2018-02-15
WIP: Enable Jekyll::Page objects to generate excerpts

### LINKED ISSUE #6778: Allow Jekyll::Page objects to render {{ page.excerpt }} with relevant content

---

---

## My Reproduction Steps

I'm trying to get  `{{ page.excerpt }}` from my layout file. This seem to only work for documents (including posts), but not for pages.

I created a repository that reproduce the issue: https://github.com/pixelastic/jekyll-excerpt

I basically created a new site using `jekyll new my_site`. 

Then I created a new layout, `_layouts/test_layout.html` with the following content:
```
---
layout: default
---

<dl>
  <dt>Excerpt:</dt>
  <dd>{{ page.excerpt }}</dd>
</dl>
```

I updated both `about.md` and the post example to use this layout. Only the post actually display the excerpt.

## The Output I Wanted

I expected to be able to get access to an excerpt for pages, as I can do for posts.

## Miscellaneous

_Note that this is a simplified test case. My actual use-case is from inside a plugin, where I'd like to get access to the excerpt of a `Jekyll::Page` instance. When I have a `Jekyll::Document` instance, I can call `.data['excerpt]` on it, but there is no such key for a page._

ashmaroli:
```
Proof of Concept
----------------
```
Resolves #6778

### DIFF
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -48,6 +48,7 @@ module Jekyll
   autoload :EntryFilter,         "jekyll/entry_filter"
   autoload :Errors,              "jekyll/errors"
   autoload :Excerpt,             "jekyll/excerpt"
+  autoload :Excerptible,         "jekyll/excerptible"
   autoload :External,            "jekyll/external"
   autoload :FrontmatterDefaults, "jekyll/frontmatter_defaults"
   autoload :Hooks,               "jekyll/hooks"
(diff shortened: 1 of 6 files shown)


ashmaroli:
Closing this, as re-implementing `site.pages` as a collection would be a better approach than porting just a fraction of a `Document` feature over to `Page`

PR #6781, opened 2018-02-15
Don't prompt for sudo when installing with Ubuntu WSL

dsolodow:
When doing the setup under Ubuntu WSL, had to set the bundle path to avoid issues with 'jekyll new' and 'bundle update'


ashmaroli:
Any thoughts on this, @sverrirs?

pathawks:
@jekyllbot: :ship: +doc

PR #6784, opened 2018-02-17
Fix theme gem feature

DirtyF:
theme gem feature failed with Rubygem 2.7.6 because we have TODO strings and an invalid URL in the default theme gemspec

`--force` option will skip the validation, it's OK here because we want to keep this TODO comments to show theme developers that they need to update those fields.

### DIFF
--- a/features/step_definitions.rb
+++ b/features/step_definitions.rb
@@ -216,8 +216,6 @@ end
 
 When(%r!^I decide to build the theme gem$!) do
   Dir.chdir(Paths.theme_gem_dir)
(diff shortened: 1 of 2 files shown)


ashmaroli:
um.. I like the `--force` option.. but we have a step definition for `When I decide to build the theme gem` where cucumber removes the `TODO` via a `sub` call..

You can just change it to `gsub`..
or please remove that step entirely if `--force` is all it takes to resolve the error.

DirtyF:
@jekyllbot: merge +dev

PR #6791, opened 2018-02-20
Do not hardcode locale unless certainly necessary

### LINKED ISSUE #6669: Incompatibility between Jekyll 3.7 and I18n filter plugin
- [x] I am on (or have tested on) ***macOS*** 10.13.2
- [x] I was trying to build.

---

## My Reproduction Steps

It is very common in the Jekyll community to use the [I18nFilter plugin](https://github.com/borisschapira/jekyll-bugs/blob/master/_plugins/i18n_filter.rb) to translate dates from data provided in YAML.

Until Jekyll 3.6.2, the usage of this plugin was ok. Since Jekyll 3.7, there is a problem.
I have prepared a release that reproduces the problem: https://github.com/borisschapira/jekyll-bugs/releases/tag/jekyll-3.7-i18n-filter

```
"----"
"Loading the locale…"
["/path/to/jekyll/folder/_plugins/../_data/_locales/fr_FR.yml", "/path/to/jekyll/folder/_plugins/../_data/_locales/en_US.yml"]
"Available locales:"
[:en]
"----"
```

## The Output I Wanted

```
"----"
"Loading the locale…"
["/path/to/jekyll/folder/_plugins/../_data/_locales/fr_FR.yml", "/path/to/jekyll/folder/_plugins/../_data/_locales/en_US.yml"]
"Available locales:"
["fr_FR", "en_EN"]
"----"
```

ashmaroli:
Fixes #6669 
We shouldn't be hardcoding locales unnecessarily

### DIFF
--- a/lib/jekyll/utils.rb
+++ b/lib/jekyll/utils.rb
@@ -203,7 +203,10 @@ module Jekyll
       end
 
       # Drop accent marks from latin characters. Everything else turns to ?
-      string = ::I18n.transliterate(string) if mode == "latin"
+      if mode == "latin"
+        I18n.config.available_locales = :en if I18n.config.available_locales.empty?
+        string = I18n.transliterate(string)
+      end
 
       slug = replace_character_sequence_with_hyphen(string, :mode => mode)
 
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -35,7 +35,6 @@ require "colorator"
 require "i18n"
 
 SafeYAML::OPTIONS[:suppress_warnings] = true
-I18n.config.available_locales = :en
 
 module Jekyll
   # internal requires


pathawks:
This was added in #6509. Was it necessary for something, or just boilerplate that got included?

ashmaroli:
> Was it necessary for something, or just boilerplate that got included?

It's required. 
For some reason, `I18n` raises an error if we do not pass `:en` as a locale in the event there are no locales defined..
Its probably a bug in `I18n` or in the implementation at our end, but as of this writing, its required.

pathawks:
> For some reason, I18n raises an error if we do not pass :en as a locale in the event there are no locales defined..

So removing this default could lead to problems if plugins expect it to be there?
4.0?

alextsui05:
Sorry, I wrote that line when adding latin mode and the situation is as ashmaroli says. I don't think there's any particular need other than to clear the error, and this PR looks good to limit setting it only where it's necessary.

ashmaroli:
> So removing this default could lead to problems if plugins expect it to be there?
> 4.0?

@pathawks Since this default was just added in `3.7.0`, its safe to assume that not many plugins would have been written to use the default.
While technically, it can be considered a breaking change; though, for practical purposes, this is actually just a `bug-fix` that should be shipped a.s.a.p.

Ideally, we'd release this as a backport in `3.7.3` but since the maintainers are busy with other stuff, I guess this will just ship with `3.8.0`..

pathawks:
If there were enough bug fixes to merit a `3.7.3`, I could probably do that later this week.

ashmaroli:
> If there were enough bug fixes to merit

I believe all other reported bugs have been fixed as part of `3.7.2`
Besides, there has been at least two backport releases with just one bug-fix in the past..

lib/jekyll/utils.rb
```diff
@@ -203,7 +203,10 @@ def slugify(string, mode: nil, cased: false)
       end
 
       # Drop accent marks from latin characters. Everything else turns to ?
-      string = ::I18n.transliterate(string) if mode == "latin"
+      if mode == "latin"
+        I18n.config.available_locales = :en if I18n.config.available_locales.empty?
```
Crunch09: Should we set `available_locales` to an array (`[:en]`) instead?

ashmaroli: I don't think its necessary.
If you were to add a `print I18n.config.available_locales` immediately after this line, the output would return `[:en]`

Crunch09: 👍 Thanks for checking!

oe:
@jekyllbot: merge +bug

lib/jekyll/utils.rb
milosgajdos83: This code only gets run if `mode` is set to `latin`. If it's not you break everyone else's code by removing `I18n.config.available_locales = :en` from jekyll module which is exactly what happened to me.

ashmaroli: @milosgajdos83 `I18n.config.available_locales = :en` was specifically added to the `Jekyll` module for the sole use in the `slugify` filter.
Please open a *new* separate issue for better tracking..

PR #6792, opened 2018-02-20
Always load Google Fonts over HTTPS

coliff:
safer and faster

### DIFF
--- a/docs/_includes/top.html
+++ b/docs/_includes/top.html
@@ -6,7 +6,7 @@
   <meta name="generator" content="Jekyll v{{ jekyll.version }}">
   {% feed_meta %}
   <link rel="alternate" type="application/atom+xml" title="Recent commits to Jekyll’s master branch" href="{{ site.repository }}/commits/master.atom">
-  <link rel="stylesheet" href="//fonts.googleapis.com/css?family=Lato:300,300italic,400,400italic,700,700italic,900">
+  <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Lato:300,300italic,400,400italic,700,700italic,900">
   <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" />
   <link rel="stylesheet" href="/css/screen.css">
   <link rel="icon" type="image/x-icon" href="/favicon.ico">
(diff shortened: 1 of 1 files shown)


coliff:
Agreed. I've started adding this to all the sites I work on:
`<link rel="preconnect" href="https://fonts.gstatic.com/" crossorigin>`
Shall I add to this PR or do you want to add it?

DirtyF:
@coliff good idea, please do.

I was curious to benchmark if we could do better reading https://www.zachleat.com/web/23-minutes/

ashmaroli:
@coliff Thank you for introducing me to `preconnect`!!
I was aware of `preload` and `prefetch`.., but `preconnect`, is a killer in its own right!!

TIL :beers:

coliff:
@ashmaroli -you're welcome! This blog post (https://www.cdnplanet.com/blog/faster-google-webfonts-preconnect/) has some good info about that - with evidence of the performance improvements.

ashmaroli:
I gez we could `preconnect` to `cdn.jsdelivr.net` as well for `<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" />`.. but that's story for a future PR :smiley:

DirtyF:
note for later : looking at the network waterfall in dev tools, fonts are not preloaded or cached. Might submit a PR to improve TTFB and global speed index.

As a result it's blocking the rendering:
https://webpagetest.org/result/170515_DB_196E/1/details/#waterfall_view_step1

DirtyF:
@jekyllbot: :shipit: +site

coliff:
> note for later : looking at the network waterfall in dev tools, fonts are not preloaded or cached. Might submit a PR to improve TTFB and global speed index.

Yep, there's a few more things to be done to improve the performance. I just saw that the static assets (images and CSS) have a cache life of just 10 minutes. Would be good to increase that.
REF: https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fjekyllrb.com%2F&tab=desktop

DirtyF:
These must default GitHub Pages defaults, I guess there's nothing much we can do about it.

DirtyF:
Here's an updated webpagetest with the current breaking of network loading: https://webpagetest.org/result/180220_TZ_3c17c2be14957b1923a6fa0ec6913677/1/details/#waterfall_view_step1

ashmaroli:
@coliff What's your take on using `preload` on a local copy of the Google Font file (it won't be changing thereafter) over `preconnect`?

DirtyF:
Here's a preload recipe that we could try: https://github.com/zachleat/web-font-loading-recipes/blob/master/preload.html

coliff:
@ashmaroli hmmm.. overall I think it might not be worth it. If you're hosting the fonts yourself it'll mean you're using GitHub Pages very short cache life (10 minutes rather than Google's CDN cache life of 365 days).

PR #6795, opened 2018-02-21
Improve handling non-default collection documents rendering and writing

### LINKED ISSUE #5953: Collection documents with future date and output set to false are not ignored

ashmaroli:
Resolves #5953 
Replaces #6506 

- [x] implement patch
- [x] update cucumber feature and unit test
- [x] update documentation

### DIFF
--- a/docs/_docs/collections.md
+++ b/docs/_docs/collections.md
@@ -348,6 +348,20 @@ you specified in your `_config.yml` (if present) and the following information:
   <code>{% raw %}{{ site.collections | where: "label", "myCollection" | first }}{% endraw %}</code></p>
 </div>
 
+<div class="note info">
+  <h5>Collections and Time</h5>
+  <p>Except for documents in hard-coded default collection <code>posts</code>, all documents in collections
+    you create, are accessible via Liquid irrespective of their assigned date, if any, and therefore renderable.
+  </p>
+  <p>However documents are attempted to be physically output (written to disk) only if the concerned collection
+    metadata has <code>output: true</code>. Additionally, future-dated documents are only written if
+    <code>site.future</code> <em>is also true</em>.
+  </p>
+  <p>A more fine-grained control over documents being written to disk can be further excercised by setting
+    <code>published: false</code> (<em><code>true</code> by default</em>) in the document's front matter.
+  </p>
+</div>
+
 
 ### Documents
 
--- a/docs/_docs/collections.md
+++ b/docs/_docs/collections.md
@@ -353,11 +353,11 @@ you specified in your `_config.yml` (if present) and the following information:
   <p>Except for documents in hard-coded default collection <code>posts</code>, all documents in collections
     you create, are accessible via Liquid irrespective of their assigned date, if any, and therefore renderable.
   </p>
-  <p>However documents are attempted to be physically output (written to disk) only if the concerned collection
+  <p>However documents are attempted to be written to disk only if the concerned collection
     metadata has <code>output: true</code>. Additionally, future-dated documents are only written if
     <code>site.future</code> <em>is also true</em>.
   </p>
-  <p>A more fine-grained control over documents being written to disk can be further excercised by setting
+  <p>More fine-grained control over documents being written to disk can be exercised by setting
     <code>published: false</code> (<em><code>true</code> by default</em>) in the document's front matter.
   </p>
 </div>
--- a/features/collections.feature
+++ b/features/collections.feature
@@ -116,7 +116,31 @@ Feature: Collections
     And the _site directory should exist
     And the "_site/puppies/fido.html" file should exist
 
-  Scenario: Hidden collection with document with future date
+  Scenario: Access rendered collection with future dated document via Liquid
+    Given I have a _puppies directory
+    And I have the following documents under the puppies collection:
+      | title  | date       | content             |
+      | Rover  | 2007-12-31 | content for Rover.  |
+      | Fido   | 2120-12-31 | content for Fido.   |
+    And I have a "_config.yml" file with content:
+    """
+    collections:
+      puppies:
+        output: true
+    """
+    And I have a "index.html" page that contains "Newest puppy: {% assign puppy = site.puppies.last %}{{ puppy.title }}"
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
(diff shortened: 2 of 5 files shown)


features/collections.feature
```diff
@@ -116,7 +116,31 @@ Feature: Collections
...
+    """
+    collections:
+      puppies:
+        output: true
+    """
+    And I have a "index.html" page that contains "Newest puppy: {% assign puppy = site.puppies.last %}{{ puppy.title }}"
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "Newest puppy: Fido" in "_site/index.html"
```
DirtyF: Why do we return a future document here?

features/collections.feature
```diff
@@ -139,6 +163,30 @@ Feature: Collections
...
+    """
+    collections:
+      puppies:
+        output: false
+    """
+    And I have a "index.html" page that contains "Newest puppy: {% assign puppy = site.puppies.last %}{{ puppy.title }}"
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "Newest puppy: Fido" in "_site/index.html"
```
DirtyF: Same question as above.

features/collections.feature
ashmaroli: I've documented the behavior in `_docs/collections.md`

This is based on https://github.com/jekyll/jekyll/pull/6506#issuecomment-340670610 and  https://github.com/jekyll/jekyll/pull/6506#issuecomment-340675231 by @pathawks.
You can iterate through *future docs* but won't be output / written to disk

ashmaroli: If this a breaking-change in behavior, we can ship this in `4.0`.. no biggie

DirtyF: @pathawks use case with future events makes sense. 
Hugo's `date`, `publishsate` and `expirydate` is way more flexible.

ashmaroli: If Pat Hawks use case is possible on `master` then this PR is safe to ship in `3.x` otherwise its probably a `4.0`

pathawks: On `master`, future docs are rendered and can be accessed, but are not written to disk. :+1:

lib/jekyll/collection.rb
```diff
@@ -210,12 +210,11 @@ def container
...
-      doc = Jekyll::Document.new(full_path, :site => site, :collection => self)
-      doc.read
-      if site.publisher.publish?(doc) || !write?
-        docs << doc
-      else
-        Jekyll.logger.debug "Skipped Publishing:", doc.relative_path
+      docs << Document.new(full_path, :site => site, :collection => self).tap do |doc|
+        doc.read
+        if !site.publisher.publish?(doc) && site.publisher.hidden_in_the_future?(doc)
+          Jekyll.logger.debug "Skip Publishing:", "#{doc.relative_path} has a future date"
```
pathawks: I am unable to trigger this log message. Perhaps I am misunderstanding.

https://github.com/jekyll/cases/commit/0c7256b03a1c24bf9afe2ad3c45627a1e3fabd1b

ashmaroli: can you post the *entire* verbose build output under the commit? Thanks

pathawks: **verbose** is the key word. 🤕

docs/_docs/collections.md
```diff
@@ -348,6 +348,20 @@ you specified in your `_config.yml` (if present) and the following information:
   <code>{% raw %}{{ site.collections | where: "label", "myCollection" | first }}{% endraw %}</code></p>
 </div>
 
+<div class="note info">
+  <h5>Collections and Time</h5>
+  <p>Except for documents in hard-coded default collection <code>posts</code>, all documents in collections
+    you create, are accessible via Liquid irrespective of their assigned date, if any, and therefore renderable.
+  </p>
+  <p>However documents are attempted to be physically output (written to disk) only if the concerned collection
```
oe: just say written to disk here

docs/_docs/collections.md
```diff
@@ -348,6 +348,20 @@ you specified in your `_config.yml` (if present) and the following information:
...
+    metadata has <code>output: true</code>. Additionally, future-dated documents are only written if
+    <code>site.future</code> <em>is also true</em>.
+  </p>
+  <p>A more fine-grained control over documents being written to disk can be further excercised by setting
```
oe: `More fine-grained control [...]`

oe:
The title is kind of misleading, no? There's more than that added in this PR.

ashmaroli:
> The title is kind of misleading, no?

Feel free to change the title to something that fits the proposed changes.

ashmaroli:
```
Improve handling non-default collection documents rendering and writing
```

DirtyF:
…waiting for @oe approval before merging this fix.

DirtyF:
@jekyllbot: :shipit: +fix

PR #6799, opened 2018-02-24
Add generator-jekyll-plugin instructions

codejamninja:


DirtyF:
So, given this tool has the merit to exist, and that lot of developers already have node on their machine where should we list this?

ashmaroli:
Why don't we add a `new-plugin <args>` command in the Core..?

DirtyF:
I believe it's what I suggested to @kacperduras to do in a `command` plugin. 80% of our users aren't gem developers, this belongs to a plugin.

ashmaroli:
> in the Core? Why not external gem?

So that *we (the Jekyll Team)* could update the style conventions in sync with Jekyll Core.
But @DirtyF is right, this is best served as a plugin..

codejamninja:
The reason this is built in NodeJS is that it just so happens that yeoman, one of the best generator tools is built in NodeJS. If someone builds a comparable generator to yeoman in Ruby (which would be pointless) then I might consider rebuilding this in ruby.

Using Ruby for the generator is like saying not to use Sphinx (the most comprehensive documentation tool) because it is built in Python. Use the tools that work the best for the job.

Also, most developers have NodeJS installed on their computer, and if they don't, it's not hard to install Node.

codejamninja:
@kacperduras, you didn't create a generator. You created a template. There is a very clear distinction.

codejamninja:
If you want to make this more rubyish for the developers using it, I could package the generator along with NodeJS inside an external gem. Then it would work without being forced to install NodeJS.

ashmaroli:
@codejamninja You're free to use the language of your choice to build a plugin. Irrespective of that, the Official Jekyll Docs isn't the place to recommend using a third-party plugin (which this PR proposed).

The section (which itself is being moved to a dedicated website) simply lists plugins created by the community, with a short summary.

pathawks:
@codejamninja Sorry if any of this discussion came across as bashing the tool; I know that was not the intent.

I think one of the big sticking points here is that the instructions have been added to the primary documentation, which might seem to imply that it is the Official Jekyll Way. And yet the Jekyll maintainers do not use this tool, nor has any explanation been provided detailing specifically what problem this tool is solving.

Additionally, it is kind of a big ask to tell developers that the Official Jekyll Way of creating a plugin involves installing (and then managing) a Node.js runtime. Sure, we’re all hackers here, but our community includes many different kinds of hackers. Just because somebody knows enough Ruby to start working on a Jekyll plugin does not necessarily mean they have any experience with Node.js. This is not to say that the choice of Node.js was necessarily _wrong_, just that it is one more hurdle that needs to be weighed against the advantages provided by this tool.

If the tool really does make creating Jekyll plugins substantially easier, I would recommend you promote it and spread the word. Perhaps we could even include a tutorial in our docs explaining what it offers and how to use it. However, until it has gained some traction, I hesitate to promote a tool that none of us have experience with as the Official Jekyll Way of creating plugins.

kacperduras:
@codejamninja Have you ever read what I sended?

DirtyF:
> yeoman, one of the best generator tools is built in NodeJS

Just curious, @codejamninja what features does yeoman bring to ease the release of such a generator that would make it difficult with a proper Ruby gem?

codejamninja:
I can't compare yeoman with ruby. That's like comparing WordPress with ruby. You would need to tell me a ruby generator framework to compare yeoman with.

codejamninja:
I totally understand not wanting to use this tool as part of the core jekyll system. I do not feel like anyone is bashing it. I think it's a good thing to be critical. I'm a very critical developer myself.

I find great value from this tool and so do many others. I chose Yeoman because it is the best generator / scaffolding tool I have worked with. So far nobody has given me a comparable ruby scaffolding tool to compare yeoman to. I don't even know if one exists.

codejamninja:
@kacperduras I definitely read your input. I have read this whole feed multiple times.

If you want to use a `new-plugin <args>` command in the core, either I will need to package this project into a ruby gem (theoretically this is possible), or someone will need to build a ruby scaffolding tool comparable to yeoman.

pathawks:
Ok, we have decided not to merge this PR at this time. This is not the place for a larger discussion about yeoman alternatives for Ruby.

Issue #6797, opened 2018-02-23
Install on macOS 10.13 with Ruby 2.5.0


chrisfinazzo:
- [ ] I believe this to be a bug, not a question about using Jekyll.
- [x] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [x] I ran `jekyll doctor` to check my configuration
- [ ] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [x] This is a feature request.

---

- [x] I am on (or have tested on) ***macOS*** 10+
- [ ] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux
- [ ] I am on (or have tested on) ***Fedora*** GNU/Linux
- [ ] I am on (or have tested on) ***Arch*** GNU/Linux
- [ ] I am on (or have tested on) ***Other*** GNU/Linux
- [ ] I am on (or have tested on) ***Windows*** 10+

---

- [x] I was trying to install.
- [ ] There is a broken Plugin API.
- [ ] I had an error on GitHub Pages, and I have reproduced it locally.
- [ ] I had an error on GitHub Pages, and GitHub Support said it was a Jekyll Bug.
- [ ] I had an error on GitHub Pages and I did not test it locally.
- [ ] I was trying to build.
- [ ] It was another bug.

## My Reproduction Steps

1. Add Jekyll to Gemfile
2. `bundle install`, `rbenv rehash`
3. In the current directory (`/Blog`), run `jekyll new .` to create a scaffold

## The Output I Wanted

A new site in `/Blog`

Instead, I get a conflict, which is odd because I specified the existing path:

`Conflict: /Users/Chris/Documents/Blog exists and is not empty.`

In addition, running `jekyll doctor` warns about the lack of a config file - which would have been otherwise created by the `jekyll new` command.

I recall (a few OS releases ago) running into the same issue, which required me to move `Gemfile` and its lockfile out of the directory temporarily and then create a site. Someone (@envygeeks, I think) said there would be a way to get around this in the future, but I don't remember how it was resolved.

What is the procedure for specifying the current directory? There may have been something like `--force` at one point, but I assume it was removed for safety.

DirtyF:
`jekyll new . --force` will get you goint.

pathawks:
https://github.com/jekyll/jekyll/blob/85cd6105dfc285067c906643d6201dd4584e8819/lib/jekyll/commands/new.rb#L30-L31

[...]

PR #6801, opened 2018-02-25
Remind user to resolve conflict in `jekyll new` with `--force`

ashmaroli:
Ref: https://github.com/jekyll/jekyll/issues/6797#issuecomment-368088913 by @pathawks 

/cc @chrisfinazzo

### DIFF
--- a/lib/jekyll/commands/new.rb
+++ b/lib/jekyll/commands/new.rb
@@ -27,8 +27,8 @@ module Jekyll
           new_blog_path = File.expand_path(args.join(" "), Dir.pwd)
           FileUtils.mkdir_p new_blog_path
           if preserve_source_location?(new_blog_path, options)
-            Jekyll.logger.abort_with "Conflict:",
-                      "#{new_blog_path} exists and is not empty."
+            Jekyll.logger.error "Conflict:", "#{new_blog_path} exists and is not empty."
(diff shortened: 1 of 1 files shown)


lib/jekyll/commands/new.rb
```diff
@@ -27,8 +27,8 @@ def process(args, options = {})
           new_blog_path = File.expand_path(args.join(" "), Dir.pwd)
           FileUtils.mkdir_p new_blog_path
           if preserve_source_location?(new_blog_path, options)
-            Jekyll.logger.abort_with "Conflict:",
-                      "#{new_blog_path} exists and is not empty."
+            Jekyll.logger.error "Conflict:", "#{new_blog_path} exists and is not empty."
+            Jekyll.logger.abort_with "", "Use '--force' to overwrite and create anyway."
```
oe: I'm not sure about this wording. Can we make it a bit less forceful? Using `--force` shouldn't be recommended as the default option.

ashmaroli: > bit less forceful?

haha..


> Using `--force` shouldn't be recommended as the default option.

```
"Ensure the #{new_blog_path} is empty or else try again with `--force`
to proceed and overwrite any files"
```

DirtyF:
@jekyllbot: merge +minor

PR #6802, opened 2018-02-25
add missing step for gem-based theme conversion

Strangehill:
Theme documentation steps for converting a gem-based theme into 'intrinsic' site files are missing explanation about plugins.

First brought up in jekyll/minima#157

h/t @ashmaroli 

@jekyll/documentation

### DIFF
--- a/docs/_docs/themes.md
+++ b/docs/_docs/themes.md
@@ -3,7 +3,7 @@ title: Themes
 permalink: /docs/themes/
 ---
 
-Jekyll has an extensive theme system that allows you to leverage community-maintained templates and styles to customize your site's presentation. Jekyll themes package up layouts, includes, and stylesheets in a way that can be overridden by your site's content.
+Jekyll has an extensive theme system that allows you to leverage community-maintained templates and styles to customize your site's presentation. Jekyll themes specify plugins and package up assets, layouts, includes, and stylesheets in a way that can be overridden by your site's content.
 
 ## Understanding gem-based themes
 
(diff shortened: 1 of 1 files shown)


docs/_docs/themes.md
```diff
@@ -3,7 +3,7 @@ title: Themes
 permalink: /docs/themes/
 ---
 
-Jekyll has an extensive theme system that allows you to leverage community-maintained templates and styles to customize your site's presentation. Jekyll themes package up layouts, includes, and stylesheets in a way that can be overridden by your site's content.
+Jekyll has an extensive theme system that allows you to leverage community-maintained templates and styles to customize your site's presentation. Jekyll themes package up plugins, assets, layouts, includes, and stylesheets in a way that can be overridden by your site's content.
```
ashmaroli: Theme-gems do not package plugins. The gemspec only *specifies* gem-based plugins that are then `required` by Jekyll while loading the theme-gem.
`*.rb` files in a `_plugins` directory in the theme-gem will not be loaded whatsoever (as of `v3.7.x`)

docs/_docs/themes.md
```diff
@@ -108,7 +108,33 @@ Suppose you want to get rid of the gem-based theme and convert it to a regular t
 
 To do this, copy the files from the theme gem's directory into your Jekyll site directory. (For example, copy them to `/myblog` if you created your Jekyll site at `/myblog`. See the previous section for details.)
 
-Then remove references to the theme gem in `Gemfile` and configuration. For example, to remove `minima`:
+Then you must tell Jekyll about the plugins that were packaged by the theme. You can find these plugins in the theme's gemspec file as runtime dependencies. If you were converting the Minima theme, for example, you might see:
```
ashmaroli: here as well.. plugins are not packaged by the theme, just referenced to..
```diff
- about the plugins that were packaged by the theme. You can find these plugins in the theme's gemspec...
+ about the plugins that are required by the theme. You can find these plugins listed in the theme's gemspec...
```

Strangehill:
Thank you. Correction attempted.

docs/_docs/themes.md
```diff
@@ -108,7 +108,33 @@ Suppose you want to get rid of the gem-based theme and convert it to a regular t
...
+
+plugins:
+  - jekyll-feed
+  - jekyll-seo-tag
+```
+
+```ruby
+# ./Gemfile
+
+group :jekyll_plugins do
```
pathawks: If using `group :jekyll_plugins`, the plugins do not need to be added to `_config.yml`

ashmaroli: *Note: GitHub Pages do not `require` plugins via Bundler. So adding to `_config.yml` is always better*

pathawks: Could be listed as an either/or.

Strangehill: I think I understand. I've attempted to incorporate these notes. Please check for accuracy and style.

pathawks:
I like this. This does a really good job of explaining what the issue is, and how to work around it.
It goes into sufficient detail as to actually help the user understand what is happening, rather than just providing a series of commands to execute blindly while crossing fingers.
🏆👊

docs/_docs/themes.md
```diff
@@ -108,7 +108,48 @@ Suppose you want to get rid of the gem-based theme and convert it to a regular t
...
+  gem "jekyll-feed", "~> 0.9"
+  gem "jekyll-seo-tag", "~> 2.1"
+end
+```
+
+Either way, don't forget to `bundle update`.
+
+However, if you're publishing on GitHub Pages you should update only your `_config.yml` as GitHub Pages doesn't require plugins via Bundler.
```
ashmaroli: one small correction here:
```diff
- GitHub Pages doesn't require plugins
+ GitHub Pages doesn't load plugins
```
while `require` is technically correct in Ruby-lingo, the English equivalent 'load' is better suited here

Strangehill: great. I was concerned about that. Fixed.

DirtyF:
@jekyllbot: merge +docs

PR #6807, opened 2018-02-26
always load analytics.js over HTTPS

coliff:
safer, faster and recommended way by Google
REF: https://developers.google.com/analytics/devguides/collection/analyticsjs/


DirtyF:
@jekyllbot: merge +site

PR #6809, opened 2018-02-26
Memoize `Site#site_data`

ashmaroli:
- `Site#site_data` is called [*twice*](https://github.com/jekyll/jekyll-seo-tag/blob/f19b29ea0a20c08b32e3018800da83d9da69ace5/lib/jekyll-seo-tag/author_drop.rb#L63-L64) per page / document when the popular plugin `jekyll-seo-tag` is used
- `jekyll-seo-tag` is loaded by default via recent releases of Minima
- IMO, its better addressing this here instead of patching per plugin(s)

### DIFF
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -93,6 +93,7 @@ module Jekyll
       self.pages = []
       self.static_files = []
       self.data = {}
+      @site_data = nil
       @collections = nil
       @docs_to_write = nil
       @regenerator.clear_cache
@@ -253,7 +254,7 @@ module Jekyll
     #
     # Returns the Hash to be hooked to site.data.
     def site_data
-      @site_data ||= config["data"] || data
+      @site_data ||= (config["data"] || data)
     end
 
     # The Hash payload containing site-wide data.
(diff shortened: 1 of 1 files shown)


lib/jekyll/site.rb
```diff
@@ -253,7 +253,7 @@ def categories
     #
     # Returns the Hash to be hooked to site.data.
     def site_data
-      config["data"] || data
+      @site_data ||= config["data"] || data
```
parkr: If you want to memoize the `||`, then it'd be: `@site_data ||= (config["data"] || data)`. By default, this is equivalent to `(@site_data ||= config["data"]) || data`.

parkr: This `@site_data` should also be reset in `reset`!

ashmaroli: Regarding the use of parenthesis, I did not know, this would be interpreted that way.. Thanks!

parkr: @ashmaroli You might enjoy http://tmm1.net/ruby21-profiling/ which has an example of where parentheses went awry!

ashmaroli: Thanks. I'll check it out :smiley:

oe:
@jekyllbot: merge +dev

PR #6811, opened 2018-02-27
Avoid using includes within Liquid forloops

ashmaroli:
### Logic:
  - `include` tag parses and renders the "includee" like a standalone page
  - using `include` tag inside a Liquid forloop (`{% for %}...{% endfor %}`) means having to parse and render the "includee" for each item in the looped array.
  - moving the code directly within the `for` tags can avoid the multiple parse-render step and get straight  to the point.

### Use-case in the Official Documentation site:
  - each of the "includee" are *small snippets* of markup used in a single layout.
  - based on above-stated *logic* and the fact that the includes are used in just a single layout, the *includes* have been scraped entirely and their contents added to the concerned layouts directly.


pathawks:
Does our docs site take too long to render?
Is this worth sacrificing readability/orthogonality?

ashmaroli:
> Does our docs site take too long to render?

Nope. *`10-11s` on Unices* and *`28-30s` on Windows*
Then why? &mdash; (IMO) The proposed change reduces the memory-load on Ruby.
Also, there's a *slight* gain in the overall build time: *`27-28s` on Windows*

---

> Is this worth sacrificing readability/orthogonality?

The following are just my view-points (only because of the *perceived* reduction in memory usage):
  - I feel reading / maintaining 2 layouts is better than reading / maintaining 2 layouts *and* 4 includes
  - Readability: I don't have to open 2 include files just to read through 10-15 lines of code that are not repeated elsewhere.
  - **The Official Documentation site source code can be seen by the community as a guide to best-practices with Jekyll**

ashmaroli:
@DirtyF Please state your reasons too :smiley:

DirtyF:
We should avoid doing the wrong things and being distracted by very minor concerns like this one.

ashmaroli:
Since there isn't any perceivable gain in the build time on @DirtyF 's end (macOS), I guess its not worth pushing this any further.
Closing..

DirtyF:
@ashmaroli That being said, I really appreciate your concern to track performance gain everywhere around in Jekyll. Things like parallelization discussed before should possibly make an impact on overall performance.

PR #6812, opened 2018-02-28
Build docs on CI after successful tests

ashmaroli:
..as an afterthought to a successful build


pathawks:
I like this idea. Do we want to do this for every job in the matrix, or maybe only on one?

ashmaroli:
> Do we want to do this for every job in the matrix, or maybe only on one?

I wanted it run on every tested `RUBY_VERSION` so that one can compare the build time on each version

pathawks:
> I wanted it run on every tested `RUBY_VERSION` so that one can compare the build time on each version

Travis makes no effort to give jobs any dedicated resources, so build times are _largely_ impacted by what other jobs may be running on Travis at the time.
With this in mind, and so that we can minimize the time it takes for CI to report, maybe we should limit this to one build in the matrix.

.travis.yml
```diff
@@ -1,4 +1,4 @@
-bundler_args: --without benchmark:site:development
+bundler_args: --without benchmark:development
```
pathawks: Can we change this in the matrix, so `site` continues to be excluded for all other jobs?

ashmaroli: I don't think that's possible..
Besides, why is that a problem, Travis is going to use the cached gemset anyways..?

pathawks:
I'm trying to remember the motivation behind this. Can we use the GitHub Pages health check instead?

ashmaroli:
> I'm trying to remember the motivation behind this.

The original motivation was to see if a PR had any impact on building our `docs` site. But now that we have Utterson testing against a larger corpus, this won't be needed.

PR #6813, opened 2018-02-28
Rewrite `script/rubyprof` as a Ruby script

ashmaroli:
The original script seemed to be unused as it has been left broken for a long time..
(The script was testing `Jekyll::Commands::Build.process({'source' => 'site'})`)

..and there were issues while trying to get it to run on Windows , even after the `source` was corrected to the current docs site at `./docs` (`NameError: uninitialized constant CallTreePrinter`)

So, I rewrote it as a Ruby script.

---

### Usage changes
1. Invocation:

```diff
- $ bash script/rubyprof
+ $ bundle exec ruby script/rubyprof
```
2. The generated result is now a simpler text output via `RubyProf::FlatPrinter` instead of `RubyProf::CallTreePrinter` earlier.
3. The generated filename now has a time-stamp as against a random integer earlier

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -55,11 +55,10 @@ end
 #
 
 group :benchmark do
-  gem "ruby-prof"
-
   if ENV["BENCHMARK"]
     gem "benchmark-ips"
     gem "rbtrace"
+    gem "ruby-prof"
     gem "stackprof"
   end
 end
--- a/Gemfile
+++ b/Gemfile
@@ -55,10 +55,11 @@ end
 #
 
 group :benchmark do
+  gem "ruby-prof"
+
   if ENV["BENCHMARK"]
     gem "benchmark-ips"
     gem "rbtrace"
-    gem "ruby-prof"
     gem "stackprof"
   end
 end
--- a/script/rubyprof
+++ b/script/rubyprof
@@ -1,19 +1,23 @@
-#!/usr/bin/env bash
+#!/usr/bin/env ruby
 
-export BENCHMARK=1
+require "ruby-prof"
+require File.expand_path("../lib/jekyll", __dir__)
 
-TEST_SCRIPT="Jekyll::Commands::Build.process({'source' => 'site'})"
+result = RubyProf.profile do
+  Jekyll::Commands::Build.process({
+    "source"      => File.expand_path("../docs", __dir__),
+    "destination" => File.expand_path("../docs/_site", __dir__),
+  })
+end
 
-RUBY=$(cat <<RUBY
-  require 'ruby-prof'
-  result = RubyProf.profile{ ${TEST_SCRIPT} }
-  printer = RubyProf::CallTreePrinter.new(result)
-  filename = "tmp/ruby_prof_#{rand 10000}"
-  puts "Writing profile to #{filename}"
-  file = File.open(filename, "w")
-  printer.print(file, {})
-  file.close
-RUBY
-)
+puts "\nProcessing result.."
 
-bundle exec ruby -r ./lib/jekyll -e "${RUBY}"
+dir_path  = File.expand_path("../tmp", __dir__)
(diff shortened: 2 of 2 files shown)


DirtyF:
Example output:

```
Measure Mode: wall_time
Thread ID: 70232756713420
Fiber ID: 70232762760360
Total: 27.370537
Sort by: self_time

 %self      total      self      wait     child     calls  name
  4.08      1.345     1.118     0.000     0.227   111111   Nokogiri::XML::Node#native_write_to
  3.36      0.924     0.920     0.000     0.004      441   <Class::Nokogiri::HTML::Document>#read_memory
  1.79      3.458     0.491     0.000     2.967   111111   Nokogiri::XML::Node#serialize
  1.64      3.906     0.449     0.000     3.458   249818  *Jekyll::Drops::Drop#[]
  1.63      2.342     0.445     0.000     1.897   111111   Nokogiri::XML::Node#write_to
  1.45      0.398     0.398     0.000     0.000   295944   Kernel#methods
  1.43      0.394     0.392     0.000     0.002      326   Nokogiri::XML::Node#in_context
  1.38      4.557     0.379     0.000     4.178   111111   Nokogiri::XML::Node#to_format
  1.38      0.453     0.378     0.000     0.075      168  *Kernel#load
  1.32      5.841     0.362     0.000     5.479   506985  *Class#new
  1.31      3.635     0.358     0.000     3.277    12787  *Array#select
  1.24      0.339     0.339     0.000     0.000    79279   String#=~
  1.22      0.334     0.334     0.000     0.000   305775   String#split
  1.17      0.337     0.321     0.000     0.016     8276   Nokogiri::XML::Node#add_child_node
  1.17      0.416     0.320     0.000     0.096     5136   Array#uniq
  1.06      2.364     0.291     0.000     2.073   180460   Jekyll::Filters#item_property
  1.06      0.290     0.290     0.000     0.000  1124383   Kernel#is_a?
....
```

Gemfile
```diff
@@ -55,10 +55,11 @@ end
 #
 
 group :benchmark do
+  gem "ruby-prof"
+
   if ENV["BENCHMARK"]
     gem "benchmark-ips"
     gem "rbtrace"
-    gem "ruby-prof"
```
pathawks: What's this about?

ashmaroli: so that one doesn't have to run
```
BENCHMARK=1 script/rubyprof
```
or
```
BENCHMARK=1 bundle exec ruby script/rubyprof
```

pathawks: So, should we just get rid of the `if ENV["BENCHMARK"]` guard altogether?

ashmaroli: no I prefer having that guard because, `stackprof` cannot be installed on Windows.. and this guard ensures it's not looked for by Bundler

pathawks: So change the guard to look for Windows?

ashmaroli: That would be a nice alternative, but I do not want to change all that in this PR.
This PR is dealing with just `script/rubyprof` that seemed to have been left unused.., so, changes to it wouldn't disrupt an existing workflow..

pathawks: Let’s revert this and discuss it in a separate PR then?

ashmaroli: Ping: @parkr, @oe, @DirtyF  Would like your opinion on this..
Which of the following formats do you prefer?
```
$ BENCHMARK=1 script/rubyprof
```
```
$ script/rubyprof
```
*I'll revert the change to the Gemfile only if the majority prefers the former format*

pathawks: I’m not trying to start a war. I’m just trying to understand the motivation for the change.

ashmaroli: okay.. I'll concede and revert this change for a future discussion..

DirtyF:
@jekyllbot: merge +dev

PR #6814, opened 2018-02-28
Memoize document output extension

ashmaroli:
This is to avoid creating new instances of `Jekyll::Renderer` each time `Document#output_ext` is called

### DIFF
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -90,7 +90,7 @@ module Jekyll
     #
     # Returns the output extension
     def output_ext
-      Jekyll::Renderer.new(site, self).output_ext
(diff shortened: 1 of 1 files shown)


ashmaroli:
> I really hope none of these attributes we keep memoizing change

@parkr The other alternative is to create a single private Renderer instance like how its done in `Jekyll::Convertible`
```ruby
# lib/jekyll/document.rb

def output_ext
  renderer.output_ext
end

private
def renderer
  @renderer ||= Jekyll::Renderer.new(site, self)
end

parkr:
This PR is fine by me, @ashmaroli.

oe:
@jekyllbot: merge +dev

PR #6819, opened 2018-03-01
Delegated methods after `private` keyword are meant to be private

ashmaroli:
#### Context:
Methods defined via `def_delegator` are `public` even if there is a preceding `private` keyword

#### Example:
https://github.com/jekyll/jekyll/blob/93e16428d8a27f20b16f77e466e0873692a9cb7f/lib/jekyll/drops/document_drop.rb#L65-L66

#### Proof:
```bash
$ irb
require 'jekyll'
# => true

Jekyll::Drops::DocumentDrop.private_instance_methods.include?(:fallback_data)
# => false

Jekyll::Drops::DocumentDrop.public_instance_methods.include?(:fallback_data)
# => true
```

### DIFF
--- a/lib/jekyll/drops/collection_drop.rb
+++ b/lib/jekyll/drops/collection_drop.rb
@@ -11,14 +11,11 @@ module Jekyll
       def_delegators :@obj, :label, :docs, :files, :directory,
                             :relative_directory
 
+      private def_delegator :@obj, :metadata, :fallback_data
+
       def to_s
         docs.to_s
       end
-
-      private
-
-      def_delegator :@obj, :metadata, :fallback_data
-      private :fallback_data
     end
   end
 end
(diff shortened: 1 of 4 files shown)


ashmaroli:
Not sure about `StaticFileDrop` though..:
https://github.com/jekyll/jekyll/blob/93e16428d8a27f20b16f77e466e0873692a9cb7f/lib/jekyll/drops/static_file_drop.rb#L5-L11

ashmaroli:
What's the preferred style..?
```ruby
class Foo
  extend Forwardable
  [...]

  private 

  def_delegator :@obj, :data, :fallback_data
  def_delegator :@obj, :path, :abs_path
  private :fallback_data, :abs_path

end
```
or
```ruby
class Foo
  extend Forwardable

  def_delegator :@obj, :relative_path, :path
  private def_delegator :@obj, :data, :fallback_data
  private def_delegator :@obj, :path, :abs_path

  [...]
end
```

DirtyF:
@jekyllbot: merge +fix

PR #6821, opened 2018-03-02
Refactor `highlight` tag to behave like the `raw` tag

### LINKED ISSUE #6658: Better options for fenced code-blocks, highlighting and languages the use braces.
- [ ] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [ ] I ran `jekyll doctor` to check my configuration

---

I'm using kramdown and pygmentize (we have our own plugin for the Mathematica language) for highlighting. At the moment, I have to use `{% raw %}` in many places because Mathematica uses braces heavily. Code-blocks look like this

```
{% highlight wl linenos %}
{% raw %}
calcPoints[pts : {pcurr_, pnext1_, pnext2_, rest___}, f_, result_] :=
  calcPoints[
    {pnext1 + f*(pnext2 - pnext1), pnext2, rest, pcurr},
    f,
    {result, pcurr}
  ] /; isNotTooShort[pts];

calcPoints[pts_, _, result_] := Partition[Flatten[result], 2];
{% endraw %}
{% endhighlight %}
```

This is not only ugly, it prevents that I can use code injection in IDEA which supports highlighting for languages inside Markdown files if proper fencing is used

img

Is there any way I can tell Jekyll to ignore code-blocks when processing Liquid tags? Or has the community ever considered making this an option?

ashmaroli:
Closes #6658 

With this, the `highlight` tag will no longer parse Liquid within the tag-body.
```
{% highlight ruby %}
{{ site.title }}
def print_hi(name)
  puts "Hi, #{name}"
end
print_hi('Tom')
{% endhighlight %}
```
will produce
```ruby
{{ site.title }}
def print_hi(name)
  puts "Hi, #{name}"
end
print_hi('Tom')
```
- [x] initial implementation
- [x] update tests
- [x] update documentation


lib/jekyll/tags/highlight.rb
```diff
@@ -2,7 +2,7 @@
 
 module Jekyll
   module Tags
-    class HighlightBlock < Liquid::Block
+    class HighlightBlock < Liquid::Raw
```
pathawks: 🤯 Never knew this was possible!

lib/jekyll/tags/highlight.rb
```diff
@@ -29,13 +29,11 @@ def initialize(tag_name, markup, tokens)
...
-        suffix = context["highlighter_suffix"] || ""
-        code = super.to_s.gsub(%r!\A(\n|\r)+|(\n|\r)+\z!, "")
-
+        prefix  = context["highlighter_prefix"] || ""
+        suffix  = context["highlighter_suffix"] || ""
+        code    = @body.gsub(%r!\A(\n|\r)+|(\n|\r)+\z!, "")
         is_safe = !!context.registers[:site].safe
-
-        output =
+        output  =
```
pathawks: Aligning `=` with the block above? :+1:

pathawks:
This will of course make `{% highlight ruby %}{% include example.rb %}{% endraw %}` impossible, but I agree that the trade is probably worth it.

A backwards-compatible alternative would be something like a `{% highlight-raw %}` tag.

lib/jekyll/tags/highlight.rb
ashmaroli: yep.. reads better..

ashmaroli:
> This will of course make `{% highlight ruby %}{% include example.rb %}{% endraw %}` impossible..

How is that syntax even valid??
To achieve what you meant, I'd use the backticks-fenced-block.. though I won't be able to print `linenos`

--

> A backwards-compatible alternative would be something..

I have a backwards-compatible alternative implementation locally.. but it is a new tag inheriting from `Tags::HighlightBlock` and involves some duplicate code.. I feel, library-wise this is the cleanest move..

pathawks:
> I feel, library-wise this is the cleanest move

I agree :100:

ashmaroli:
> We need some documentation for this...

@pathawks Definitely. Its already part of this PR's *todo-list*.
But weren't we adding documentation for `v4.0` to a separate branch that gets published after `v4.0` has been shipped..? Or did that change?

*Also I'm waiting for #6983 to be merged first as resolving conflicts here is a lot cleaner..*

DirtyF:
@jekyllbot: merge +major

kvz:
Sorry for the amateur question, but I've been looking at 4.0 (btw very nice work on that release! :tada: ) and how would one go about doing `{% highlight ruby %}{% include example.rb %}{% endhighlight %}` now? I've successfully changed a few occasions in my site's codebase to triple backtick fenced blocks in markdown files, but those blocks do not work in html files it seems.

ashmaroli:
@kvz That is a situation I hadn't considered..
Markdown documents allow interpolation via kramdown by using `markdown="1"` on the parent HTML container but that doesn't work for HTML source files..

Hmm.. Can you file a bug report? If we can't come up with a patch in time for Jekyll 4.0 release, this feature will be reverted.

Thank you for bringing this to my notice. :+1:

mattr-:
I'm not sure how @kvz's situation is any different that the exact situations that are called out in the PR body and the feature tests. We've called out that doing the exact example `{% highlight ruby %}{% include example.rb %}{% endhighlight %}` won't work anymore. The file type doesn't matter. 

I wouldn't call this a workaround per se, but I think the same effect could be gotten by having a file, let's call it `foo.example` with the code:
```
{% highlight ruby %}
my_fancy_ruby_code_goes_here
{% endhighlight %}
```

and then doing
```liquid
{% include foo.example %}
```

in a post.

kvz:
It’s just that the workaround won’t work in html. The workaround you propose would break the json and python files we are trying to include and highlight in our documentation site (the cool thing is we can actually run the source code and display it, too, without duplication or so)

I think with this change, there is no way to do that anymore.

Which may be worth it, not sure, you’d be best to judge. But for me would mean I’d have to stay on 3.x(?)

ashmaroli:
> I'm not sure how @kvz's situation is any different that the exact situations that are called out in the PR body and the feature tests.

@mattr- The feature tests do not test the scenario put forth by @kvz. (Using `foo.html` as a source file which won't be converted by kramdown (? or does it?) and hence use of triple backticks within `foo.html` won't be converted by kramdown.)
Moreover, the tests do not suggest a what will work. Just what won't work. (My bad!)

However, I like your suggestion. So basically, we reverse the operation.
Where one would highlight an included snippet in `v3,x`, in `4.0`, they'd be including highlighted snippets. That has to documented in the upgrading post..

ashmaroli:
@kvz Can I see a sample repository? So that I can visualize how it all comes together currently..?

kvz:
Hi, so the project I'm talking about is the site for https://transloadit.com. We'll use code snippets in our docs, [demos](https://transloadit.com/blog/), and [blog posts](https://transloadit.com/blog/) such as this one <https://transloadit.com/blog/2019/03/audio-delay/> that is executed, as well as displayed. It's nice that way as we don't have to keep 2 copies of source files, as over time, they'd certainly diverge, and what you see, has less and less to do with what you get.

For blog posts it's not so much a problem as we can use triple backticks there, but for the demos especially, or the homepage, those are more layout-heavy and as such html files, and we wouldn't be able to upgrade to Jekyll 4 without duplicating our source files as I now understand it, where 1 has the code, and the other has the code, plus `{% highlight ruby %}...{% endhighlight %}` wrappers, and we'd have to include the latter and run the former.

The repo isn't open source, so i can't share it in its entirety. But the relevant snippets I already shared higher up in the thread, I think. If you need more specifics I'm happy to provide them however.

ashmaroli:
@kvz I understand your situation but unfortunately do not see a clean solution for it.
@mattr- Should we revert this change and introduce a new `{% highlight_raw %}` tag instead?

mattr-:
@ashmaroli yup, that's the right thing to do. Thanks for opening that revert PR.

Since we'll be introducing a new tag, we can even delay the new tag addition until 4.1 if necessary. Also, I think it makes more sense to call the tag `raw_highlight` rather than `highlight_raw` but I'm not attached to any particular name for the tag.

PR #6822, opened 2018-03-02
Refactor private method `HighlightBlock#parse_options`

ashmaroli:
- return early if argument is empty
- skip checking if key `linenos` exist in the 'options' hash => if it doesn't exist, the hash will return `nil` anyways..
- guard clause makes the method more appealing.. :stuck_out_tongue:


DirtyF:
@jekyllbot: merge +dev

PR #6824, opened 2018-03-02
Disable Liquid via front matter

pathawks:
- [x] Add test for posts

This adds an option to specify `liquid: false` in front matter and disable Liquid entirely for a particular document. This would help sites that want to use fenced code blocks without worrying about the Liquid `{% raw %}` tag. This would make it easier to create Markdown documents that render the same way on GitHub and in Jekyll.

This would also make _our_ lives easier, as we could get rid of all the `{% raw %}{% endraw %}` tags on our [templates documentation page](https://github.com/jekyll/jekyll/blob/master/docs/_docs/templates.md).

This could be set using front matter defaults for even more win.


ashmaroli:
@pathawks Leaving notes here since its a WIP..
Regarding the tests:
  - `:read_yaml` doesn't call `doc.render_with_liquid?`. `Renderer#render_document` does..
  - would be better if the fixture page contained more than an empty `raw` tag..
```
---
foo: bar
liquid: false
---
### Valid Liquid
{% raw %}{{ page.foo }}{% endraw %}
{{ page.foo }}

### Invalid Liquid
{% page.foo }}
```

pathawks:
@ashmorali Generally speaking, are you on board with this idea?

ashmaroli:
> Generally speaking, are you on board with this idea?

Yep. Definitely in favor of the idea. :100:
It doesn't seem like something that is prone to bugs or something that would need long-term maintenance..

The only thing I'm not sure of is whether it should be `liquid: false` or `liquid: disabled`..

ashmaroli:
@pathawks Any updates on this?

kenman345:
shouldnt we have a spec added to test that the affirmative is also able to work as expected? 

Also, Why is convertible.rb using a ref to `self` but document.rb is not? 

As a final thought, do you need to say `return`? Cant the lines just be:
`false if self.data["liquid"] == false`

pathawks:
> Cant the lines just be:
> ```ruby
> false if self.data["liquid"] == false
> ```

Unfortunately no. In this case, we want to bail out of the method immediately if `self.data["liquid"] == false`. There is no need for the rest of the checks. That is the reason for the `return`.

It would certainly be possible to incorperate this condition into the longer condition that follows, but deciphering it would require several applications of DeMorgan's Law to figure out the intent. I think the current method reads better than:

```ruby
!(!data["liquid"] || coffeescript_file? || yaml_file? || !Utils.has_liquid_construct?(content))
```

Computationally, there isn't any meaningful difference, so we should use what is easier to read.

kenman345:
I meant that the keyword `return` is unnecessary, no?

pathawks:
> I meant that the keyword `return` is unnecessary, no?

The `return` is necessary because otherwise the expression would be evaluated and ignored. Ruby only returns the result of the last expression of the method, unless an explicit `return` is provided.

kenman345:
Ah, okay. Sorry, still learning Ruby. Just trying to streamline the code
and learn more.

pathawks:
@kenman345 No need to apologize; hacking on Jekyll is how I taught myself Ruby. Happy to have you on board :beers:

pathawks:
Ok, I have added tests to make sure that posts are correctly rendered with Liquid set to `true` and `false`.
This PR is ready for review.

pathawks:
@ashmaroli It is a boolean option with two states: on or off. For the programmer in me, it makes little sense for the type to be a string.

Obviously if this is the only thing blocking this PR I am open to other options.

ashmaroli:
*Disclaimer: The following comment is just an opinionated argument and should not influence the course of this PR.*

> For the programmer in me, it makes little sense for the type to be a string.

Perhaps. But from **my perspective**, a boolean doesn't fit here.
Why? `published: true` or say, `hidden: true` fits intuitively because the said "keys" are **adjectives**, whereas as `liquid:` is a **noun**..

```yaml
# Is the document "published" or not?
published: true

# Is the document "hidden" or not?
hidden: true

# Is the document "processed by liquid" or not?
# liquid: disabled ?
liquid: false
```

pathawks:
Does @jekyll/documentation have any input here? What would be the easiest to explain to users?

kenman345:
I feel like the key being liquid is poor. What if someone’s using Jekyll to
manage chemicals and uses liquid as a key for a property?

I think instead we should have a more direct key, such as perhaps
`process_liquid` Pr something that we would expect only the fewest of
people to have in use already.

pathawks:
The three hardest problems in programming are naming things and off by one errors.

kenman345:
I just think that if we are going to use a name like that, then either have
the tag name be defined in the `_config.yml` or make it so it’s never used
by accident.

We are trying to make people’s lives easier not harder. If we keep it as
`liquid` then I wholeheartedly agree it should be a Boolean.

Maybe a sequence `render` with the subkey
`liquid` would be best so that it can be used to enable or disable other
built in Jekyll norms on a per document/page basis overwriting the defaults
which come from the site level variable definitions.

kenman345:
Looking this over, I kinda like the idea of using the same name as the new method you added `render_with_liquid` to designate when to explicitly disable liquid rendering.

example:
```yaml
render_with_liquid: false # Do not process/render liquid elements
```

ashmaroli:
I like `render_with_liquid:` too :+1:

pathawks:
> I kinda like the idea of using the same name as the new method you added `render_with_liquid`

> I like `render_with_liquid:` too

So let it be written, so let it be done

kenman345:
Does the feature scenario have to be updated?

pathawks:
> Does the feature scenario have to be updated?

@kenman345 ✅Done

https://github.com/jekyll/jekyll/blob/15edbe0411a115611ecf9c8ca54aad8f292ccd94/features/post_data.feature#L254-L264

kenman345:
I meant those two lines starting with “and” that use the variable name we
moved away from. It’s somewhat confusing why it’s there as opposed to what
is there now or what the name of the variable is now

pathawks:
> It’s somewhat confusing why it’s there as opposed to what
> is there now or what the name of the variable is now

Would it be less confusing to just use `"Hello World {{ page.title }}"` or similar? The main part of the test is the stuff in the `{{ }}`

ashmaroli:
@kenman345  How about the following..?

```gherkin
Scenario: Use page.render_with_liquid variable
  Given I have a _posts directory
  And I have the following posts:
    | title           | render_with_liquid | date       | content                |
    | Unrendered Post | false              | 2017-07-06 | Hello {{ page.title }} |
    | Rendered Post   | true               | 2017-07-06 | Hello {{ page.title }} |
  When I run jekyll build
  And I should not see "Hello Unrendered Post" in "_site/2017/07/06/unrendered-post.html"
  But I should see "Hello {{ page.title }}" in "_site/2017/07/06/unrendered-post.html"
  And I should see "Hello Rendered Post" in "_site/2017/07/06/rendered-post.html"
```

ashmaroli:
@pathawks I think the *scenario* above :point_up_2:  would be a lot less confusing to newer contributors..

kenman345:
Agreed

pathawks:
@kenman345 @ashmaroli Thanks. I added you both as co-authors for that commit :beers:

ashmaroli:
> I added you both as co-authors

Thank you very much @pathawks :clinking_glasses: But it wouldn't make a difference because Jekyllbot would simply squash it into a single commit authored by you..

kenman345:
I work hard for my credits but appreciate the thoughtfulness. I am looking
forward to seeing this get used and all the other exciting things coming to
Jekyll

pathawks:
@ashmaroli https://github.com/parkr/auto-reply/issues/40

kenman345:
Are we waiting for the bot to have the ability to give credit? I think I'm okay if you just do a manual merge then if you really care to give credit. This being a code complete PR and self contained should not have any reliance on a bot for something we can handle for the timebeing manually.

pathawks:
> Are we waiting for the bot to have the ability to give credit?

Waiting on a review from @jekyll/core

pathawks:
@jekyllbot: :ship: +feature

kenman345:
should a similar change as was in the document.rb file be applied to the following in excerpt.rb? 

```ruby
    def render_with_liquid?
      !(coffeescript_file? || yaml_file? || !Utils.has_liquid_construct?(content))
    end
```

ashmaroli:
> be applied to the following in excerpt.rb?

I think it should. If a "document" shouldn't be rendered by Liquid, its "excerpt" shouldn't either.. 
Good catch.

kenman345:
No problem. I am working on trying to get my yml files to render and working through some of those pages. I'm having a bit of trouble but hopefully will figure it all out. Was considering opening the PR once I have a decent amount done so that maybe someone might be able to help get me over the finish line with it.

PR #6826, opened 2018-03-03
Exclude CNAME by default

willnode:
This PR is about excluding `CNAME` from site builds by default.

I know it's a GitHub things, but there are so many sites ends up including CNAME, even:

+ https://developer.github.com/CNAME
+ https://desktop.github.com/CNAME
+ https://services.github.com/CNAME
+ https://blog.github.com/CNAME
+ https://help.github.com/CNAME
+ https://hub.github.com/CNAME
+ https://webpack.js.org/CNAME
+ https://nuxtjs.org/CNAME
+ [And many many more](https://gist.github.com/willnode/40220661ca6228d7cbf2f306807ac1c9)

The effect might be a little, but personally I don't want to bother with config file just to add `CNAME` to it.

### DIFF
--- a/docs/_docs/configuration.md
+++ b/docs/_docs/configuration.md
@@ -644,7 +644,7 @@ collections:
 # Handling Reading
 safe:                 false
 include:              [".htaccess"]
-exclude:              ["Gemfile", "Gemfile.lock", "node_modules", "vendor/bundle/", "vendor/cache/", "vendor/gems/", "vendor/ruby/"]
+exclude:              ["CNAME", "Gemfile", "Gemfile.lock", "node_modules", "vendor/bundle/", "vendor/cache/", "vendor/gems/", "vendor/ruby/"]
 keep_files:           [".git", ".svn"]
 encoding:             "utf-8"
 markdown_ext:         "markdown,mkdown,mkdn,mkd,md"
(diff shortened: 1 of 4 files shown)


pathawks:
Could this be added to the `pages-gem`?

thiemo:
@willnode not including the CNAME file in the site build will break the custom domain.

I suspect the contents of the CNAME files in the site dir are parsed into the github web servers vhost config and DNS.

If you don‘t want the CNAME file to show up in your page build, don‘t include it in your source!

willnode:
@thiemo I do have a blog that already excludes CNAME in _config.yml, and it works. I just wonder if we can also exclude CNAME just like Gemfile.

> not including the CNAME file in the site build will break the custom domain.

If the site have no plugins, then no, because it [works for me](https://github.com/willnode/blog).

For those who have jekyll integrated to CI, I believe `CNAME` is already been handled automatically by that (Travis have the option).

For the rest who publish Jekyll to Github Pages manually, I believe they already have custom `exclude` list in `_config.yml`, which leaves this PR unnaffected, otherwise they'll know CNAME have been missing before the site get published by hand.

ashmaroli:
@thiemo The following is the *custom* `exclude:` setting for Jekyll's official documentation site hosted on a custom domain via GitHub Pages:
https://github.com/jekyll/jekyll/blob/93e16428d8a27f20b16f77e466e0873692a9cb7f/docs/_config.yml#L44-L48

parkr:
Excluding in github/pages-gem is the right move. Jekyll should not care about this IMO.

DirtyF:
@willnode Please rather open an issue on the [pages-gem](https://github.com/github/pages-gem/issues/) repo. Thanks.

Issue #6800, opened 2018-02-24
Add ruby environment instructions to Install guide


tedsanders:
- [ ] I believe this to be a bug, not a question about using Jekyll.
- [ ] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [ ] I ran `jekyll doctor` to check my configuration
- [ ] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [ ] This is a feature request.

---

- [ ] I am on (or have tested on) ***macOS*** 10+
- [X] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux
- [ ] I am on (or have tested on) ***Fedora*** GNU/Linux
- [ ] I am on (or have tested on) ***Arch*** GNU/Linux
- [ ] I am on (or have tested on) ***Other*** GNU/Linux
- [ ] I am on (or have tested on) ***Windows*** 10+

---

- [X] I was trying to install.
- [ ] There is a broken Plugin API.
- [ ] I had an error on GitHub Pages, and I have reproduced it locally.
- [ ] I had an error on GitHub Pages, and GitHub Support said it was a Jekyll Bug.
- [ ] I had an error on GitHub Pages and I did not test it locally.
- [ ] I was trying to build.
- [ ] It was another bug.

## My Reproduction Steps

This isn't a Jekyll bug, but a suggestion for improving the install experience for new users. Hopefully the suggestion won't take much time or thought to implement.

I have a fresh install of Linux Mint 17. Coming from using Jekyll in a Windows environment.

Went first to the install page: https://jekyllrb.com/docs/installation/#install-with-rubygems

Checked the everything was installed.

Ran the very first command to install Jekyll (gem install Jekyll).

Immediately was told by terminal that I don't have write permissions. Googled around and saw that this is a common problem among many users and distros.

I'm not saying this is a bug with Jekyll, it's more about setting up Ruby environments, but I just wanted to highlight that it's a giant user barrier when Linux users run into an error on the very first step of an installation guide. Many users thinking of trying Jekyll out may just stop here.

Therefore, I suggest adding a link or instruction or paragraph that explains the best way to get around the permissions issue. I think this will strengthen the pipeline of new Jekyll users and reduced headaches for everyone. Thanks for your time and sorry if I've said anything ignorant.

tedsanders:
[...]

PR #6828, opened 2018-03-05
Add Bundler Installation Instructions

mkasberg:
This commit addresses issue #6800. Specifically, this commit adds a
section for installing Jekyll using Bundler (as an alternative to the
gem command or RVM).

In addition, clean up the surrounding instructions a bit, and add some
helpful hints. One hint for Ubuntu users to get started quickly, and one
hint for people who run into permissions problems with `gem install
jekyll`.


docs/_docs/installation.md
```diff
@@ -37,11 +46,34 @@ Before you start, make sure your system has the following:
...
+
+Then, add Jekyll to your project's dependencies:
+
+```sh
+cd my-jekyll-project
+bundle init                          # Creates an empty Gemfile
+bundle install --path vendor/bundle  # Set install path to ./vendor/bundle
+bundle add jekyll                    # Add & install Jekyll gem
+```
+
```
ashmaroli: I'm not a fan of this method though..
A new user would get confused by having to run `*bundle exec jekyll new .` and be greeted by an error message.,. while elsewhere in the docs, we simply instruct users to initialize a jekyll project with `jekyll new [jekyll-project-path]`

docs/_docs/installation.md
```diff
@@ -37,11 +46,34 @@ Before you start, make sure your system has the following:
...
+First, install Bundler (as a system-wide gem):
+```sh
+sudo gem install bundler
+```
+
+Then, add Jekyll to your project's dependencies:
+
+```sh
+cd my-jekyll-project
```
ashmaroli: you can't `cd` into a path that doesn't exist before-hand..

docs/_docs/installation.md
```diff
@@ -37,11 +46,34 @@ Before you start, make sure your system has the following:
...
+## Install & Use Jekyll with Bundler
+
+To use Jekyll with [Bundler](https://bundler.io/), add it as a
+dependency of your project. It can be installed in a sub-directory of the
+project. This avoids permissions issues, and also allows you to use different
+versions of Jekyll with different projects.
+
+First, install Bundler (as a system-wide gem):
+```sh
+sudo gem install bundler
```
pathawks: If the goal is to avoid permissions issues, using `sudo` is a non-starter.

docs/_docs/installation.md
```diff
@@ -37,11 +46,34 @@ Before you start, make sure your system has the following:
   </p>
 </div>
 
+## Install & Use Jekyll with Bundler
```
pathawks: I don’t know if we want this `./vendor/bundle` method to be presented first, as if it is the default/recommended method :/

docs/_docs/installation.md
```diff
@@ -50,6 +82,16 @@ gem install jekyll
 All of Jekyll’s gem dependencies are automatically installed by the above
 command, so you won’t have to worry about them at all.
 
+<div class="note info">
+  <h5>Permissions Issues?</h5>
+  <p>
+    The above command performs a system-wide install of jekyll (on most
+    systems). If you get a permissions error, you can either run the command as root
+    (<code>sudo gem install jekyll</code>) or use
```
pathawks: This always leads to conflicts with multiple versions of Jekyll being installed. There have been countless bug reports from users who installed Jekyll using `sudo` without understanding what that meant. Let’s recommend installing to `~/` instead of using `sudo`

docs/_docs/installation.md
```diff
@@ -50,6 +82,16 @@ gem install jekyll
...
+    <a href="https://rvm.io/gemsets/basics">RVM</a> to manage your Gems.
```
pathawks: Isn’t RVM for managing Ruby versions? I have always used RVM in addition to Bundler. Can we provide any more detail here?

docs/_docs/installation.md
```diff
@@ -28,6 +29,14 @@ Before you start, make sure your system has the following:
   </p>
 </div>
 
+<div class="note info">
```
pathawks: Do we really need three notes in a row?
Is there a better way to present this?

DirtyF:
Thanks @mkasberg, we need to do better on install instructions for different systems. 

This feels too custom/advanced for a _default_ install compared to the [instructions for  installing jekyll on Linux on CloudCannon](https://learn.cloudcannon.com/jekyll/install-jekyll-on-linux/).

I don't mind adding a dedicated `Install on GNU/Linux` section as long as there's a compromise from linux users on what should be the _simplest_ default working method. What I see is that experienced command line users have each one a preferred method and we can't list them all. We just need a good default method.

This does not prevent to also add advanced methods and custom options like the `--path` you show here.
 
Resources:
- [how to install rbenv under Ubuntu/linux](https://gorails.com/setup/ubuntu/17.10).

docs/_docs/installation.md
DirtyF: Agreed. Let's do a section per system.

docs/_docs/installation.md
DirtyF: what is the recommended method then?

docs/_docs/installation.md
DirtyF: We  could distinct new with defaults from blank (working on a PR to have a working default directory structure. with blank)

docs/_docs/installation.md
DirtyF: is there a simpler method than [the one listed here](https://gorails.com/setup/ubuntu/17.10)?

```
sudo apt-get install libgdbm-dev libncurses5-dev automake libtool bison libffi-dev
gpg --keyserver hkp://keys.gnupg.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3
curl -sSL https://get.rvm.io | bash -s stable
source ~/.rvm/scripts/rvm
rvm install 2.5.0
rvm use 2.5.0 --default
ruby -v
```

docs/_docs/installation.md
```diff
@@ -37,11 +46,34 @@ Before you start, make sure your system has the following:
   </p>
 </div>
 
+## Install & Use Jekyll with Bundler
+
+To use Jekyll with [Bundler](https://bundler.io/), add it as a
+dependency of your project. It can be installed in a sub-directory of the
+project. This avoids permissions issues, and also allows you to use different
```
oe: `permissions` -> `permission`

docs/_docs/installation.md
oe: The only simpler method I can think of is if RVM is in your system's package manager repositories, I'm afraid.

docs/_docs/installation.md
pathawks: `gem install --user-install bundler`

mkasberg:
Thanks for all the comments everyone, it's all good feedback.

I think perhaps this pull request is changing too much too quickly. @oe brings up a good point that while Bundler may be preferable for some users, perhaps it's not a good default if you're going for the simplest method. Also, I don't do `jekyll new` too often, so I wasn't thinking about that... I'll probably close this pull request (unmerged), but I wouldn't mind following up with a simpler pull request if there's some kind of consensus.

I guess the original thing that I wanted to fix is to make sure a new user knows what to do if they run into permissions problems at the `gem install jekyll` step. Particularly if they haven't used Ruby before and just want to run Jekyll as easily as possible. I know there are lots of ways to go about it, but is there any kind of consensus on what might be best for new users?

docs/_docs/installation.md
mkasberg: Good point. No sudo.

docs/_docs/installation.md
mkasberg: Agreed. I've always installed Bundler along with Ruby via apt on Ubuntu, but if you're on LTS it's a really old version of Bundler... But you're right, we shouldn't recommend that.

pathawks:
> I guess the original thing that I wanted to fix is to make sure a new user knows what to do if they run into permissions problems at the `gem install jekyll` step. Particularly if they haven't used Ruby before and just want to run Jekyll as easily as possible.

Yes. :100:

This has been a very difficult problem for us to solve. On the one hand, we do not distribute or support Ruby or Bundler and aren't really setup to support users that are still trying to get those things installed. On the other hand, this is the biggest hurdle for new users getting started with Jekyll, and I am sure we would have happier users (and quite possibly more developers) if installing Jekyll were not by far the most difficult step toward using (or contributing to) Jekyll.

What makes it especially tricky is that, as on display in this thread, it seems there are as many different ways to install Jekyll as there are users using Jekyll. This means there isn't necessarily anything as simple as "Copy these three lines and paste them into a shell."

You have done a lot of good work here and it would be a shame to just close the PR. I wonder if this could be spun into a [tutorial](https://jekyllrb.com/tutorials/home/) for installing Jekyll on Ubuntu or something similar? Maybe for users that want to use a Docker container or install Jekyll system wide, make sure you have these dependencies and `sudo` install everything.

Thoughts @DirtyF?

DirtyF:
We need to rethink the whole install page, so that's it's easier to follow for each OS:

I was thinking of something like:

jekyll-install-macos

We would need help on the GNU/Linux part. Anyone willing to contribute?

mkasberg:
I could spin this PR off into a tutorial, "Using Jekyll as a Bundler Dependency," that's probably a better place for it. It can be done without the `sudo gem install bundler` step... I just need to figure out the best/simplest way.

I'm also willing to help with writing installation instructions for GNU/Linux, I think the idea of separate sections probably makes sense.

pathawks:
I really like this document conceptually, but I'm a bit tired at the moment so somebody from @jekyll/documentation should also review 😴:+1:

docs/_tutorials/using-jekyll-with-bundler.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Create A Jekyll Scaffold
+
+Now that Jekyll is installed, we can use it to create the scaffolding for our
+site. We need the `--force` parameter because our folder isn't empty - it
+already has some Bundler files in it. We run the `bundle install` separately
+because Jekyll gets confused if the Gemfile already exists.
+
+```sh
+bundle exec jekyll new --force --skip-bundle .
```
ashmaroli: you missed the 'dot' `.` here.. `bundle exec jekyll new .`

oe: @ashmaroli the dot is included here, at the end of the line

ashmaroli: gosh.. I did not see that..
guess my brain registered it as a 'full-stop'.. :stuck_out_tongue:

DirtyF:
@jekyllbot: merge +docs

PR #6839, opened 2018-03-10
fix minitest deprecation warning in test

Crunch09:
```
Use assert_nil if expecting nil from test/test_site.rb:65:in `block (2 levels) in <class:TestSite>'. This will fail in MT6.
```

cc: @jekyll/stability


DirtyF:
@jekyllbot: merge +dev

PR #6610, opened 2017-12-08
Change regex to sanitize and normalize filenames passed to LiquidRenderer

ashmaroli:
### Motivation:
Currently, the resulting table from building with `--profile` has **`relative_path`** for files at the source dir but an **`absolute_path`** for files within a theme-gem (which can be a long string depending on the user's Ruby Installation location.)

### Patch notes:
  - all filenames are normalized to be relative to the source-directory or theme-root.
  - if a path starts with a non-alphanumeric char, (e.g. `../../` or `~foo`), those characters are ignored.
  - no discernible effect on performance.

---

### UPDATE - 09 FEB 2018
  - prepend relative paths from theme-gem with the gem's dirname
(e.g. `minima-2.3.0/_includes/head.html`)

PR #6841, opened 2018-03-11
Optimize computing filename in LiquidRenderer

ashmaroli:
I should've made these corrections *before* #6610 got merged..
Oh well.."*Better late than never..*"

### DIFF
--- a/lib/jekyll/liquid_renderer.rb
+++ b/lib/jekyll/liquid_renderer.rb
@@ -7,8 +7,8 @@ module Jekyll
   class LiquidRenderer
     extend Forwardable
 
-    def_delegator :@site, :in_source_dir, :source_dir
-    def_delegator :@site, :in_theme_dir, :theme_dir
+    private def_delegator :@site, :in_source_dir, :source_dir
+    private def_delegator :@site, :in_theme_dir, :theme_dir
 
     def initialize(site)
       @site = site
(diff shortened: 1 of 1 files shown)


lib/jekyll/liquid_renderer.rb
```diff
@@ -53,7 +53,7 @@ def self.format_error(e, path)
     private
 
     def filename_regex
-      %r!\A(#{source_dir}/|#{theme_dir}/|\W*)(.*)!oi
+      @filename_regex ||= %r!\A(#{source_dir}/|#{theme_dir}/|\W*)(.*)!i
```
Crunch09: Just curious: Why did you get rid of the `o` modifier?

ashmaroli: The `o` modifier stands for "interpolate once" which I feel is *unnecessary" when the regex itself is going to be created once per `Site` instance since the entire method has now been memoized..

lib/jekyll/liquid_renderer.rb
```diff
@@ -7,8 +7,8 @@ module Jekyll
   class LiquidRenderer
     extend Forwardable
 
-    def_delegator :@site, :in_source_dir, :source_dir
-    def_delegator :@site, :in_theme_dir, :theme_dir
+    private def_delegator :@site, :in_source_dir, :source_dir
+    private def_delegator :@site, :in_theme_dir, :theme_dir
```
pathawks: I recognize that these were always _meant_ to be private, but at this point would it be easier to punt this until `v4.0.0`? Or is this necessary for this PR?

ashmaroli: These attributes have not been released yet.. I *added them in #6610*..
So they can be safely altered..

ashmaroli: (this is therefore just a `fix` as wel..)

pathawks:
@jekyllbot: :shipit: +minor

PR #6846, opened 2018-03-14
Assign and use Liquid variables in templates

ashmaroli:


ashmaroli:
> What's the motivation for this?

None. I just saw scope for improvement while going through the templates code

PR #6848, opened 2018-03-15
Improve regex usage in `Tags::IncludeTag`

ashmaroli:
Summary of changes:
- moved locally initialized regexps in frequently called methods to class constants
- `INVALID_SEQUENCES` now flag sequences of *multiple adjacent periods* `(..)` and *multiple adjacent forward slashes* `(//)` along with original sequences of *dot-slash* `(./)` and *slash-dot* `(/.)`
- `:validate_file_name` now first checks for `INVALID_SEQUENCES` and then proceeds to check for `VALID_FILENAME_CHARS` to `raise` sooner
- use `String#=~` over slower `String#match`

### DIFF
--- a/lib/jekyll/tags/include.rb
+++ b/lib/jekyll/tags/include.rb
@@ -21,6 +21,10 @@ module Jekyll
         (?<params>.*)
       !x
 
+      FULL_VALID_SYNTAX = %r!\A\s*(?:#{VALID_SYNTAX}(?=\s|\z)\s*)*\z!
+      VALID_FILENAME_CHARS = %r!^[\w/\.-]+$!
+      INVALID_SEQUENCES = %r![\./][\./]+!
+
       def initialize(tag_name, markup, tokens)
         super
         matched = markup.strip.match(VARIABLE_SYNTAX)
@@ -59,7 +63,7 @@ module Jekyll
       end
 
       def validate_file_name(file)
-        if file !~ %r!^[a-zA-Z0-9_/\.-]+$! || file =~ %r!\./! || file =~ %r!/\.!
+        if file =~ INVALID_SEQUENCES || file !~ VALID_FILENAME_CHARS
           raise ArgumentError, <<-MSG
 Invalid syntax for include tag. File contains invalid characters or sequences:
 
@@ -74,8 +78,7 @@ MSG
       end
 
       def validate_params
-        full_valid_syntax = %r!\A\s*(?:#{VALID_SYNTAX}(?=\s|\z)\s*)*\z!
-        unless @params =~ full_valid_syntax
+        unless @params =~ FULL_VALID_SYNTAX
           raise ArgumentError, <<-MSG
 Invalid syntax for include tag:
 
@@ -96,7 +99,7 @@ MSG
 
       # Render the variable if required
       def render_variable(context)
-        if @file.match(VARIABLE_SYNTAX)
+        if @file =~ VARIABLE_SYNTAX
           partial = context.registers[:site]
             .liquid_renderer
             .file("(variable)")
--- a/lib/jekyll/tags/include.rb
+++ b/lib/jekyll/tags/include.rb
@@ -23,7 +23,7 @@ module Jekyll
 
       FULL_VALID_SYNTAX = %r!\A\s*(?:#{VALID_SYNTAX}(?=\s|\z)\s*)*\z!
       VALID_FILENAME_CHARS = %r!^[\w/\.-]+$!
-      INVALID_SEQUENCES = %r![\./][\./]+!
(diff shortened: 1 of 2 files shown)


lib/jekyll/tags/include.rb
```diff
@@ -59,7 +63,7 @@ def parse_params(context)
       end
 
       def validate_file_name(file)
-        if file !~ %r!^[a-zA-Z0-9_/\.-]+$! || file =~ %r!\./! || file =~ %r!/\.!
+        if file =~ INVALID_SEQUENCES || file !~ VALID_FILENAME_CHARS
```
pathawks: The other two (below) are obvious wins. This one needs a closer look. I am worried about unintentionally changing behavior. It is not obvious to me that `INVALID_SEQUENCES` is the complement of `%r!^[a-zA-Z0-9_/\.-]+$!`, or what this is all about. Since the existing regexes are kind of cryptic, I might need to `git blame` to figure out what the original intent was.

pathawks:
This needs to be `3.8.0`?

ashmaroli:
> This needs to be `3.8.0`?

We can't introduce class constants in a patch release..

pathawks:
Sorry, what I meant was: does this need to be a blocker for `3.8.0` instead of maybe being slated for `3.9.0`? No matter.

lib/jekyll/tags/include.rb
pathawks: This line is nine years old.
https://github.com/jekyll/jekyll/commit/8a2a42ba7142f705bb2ef31ec204d0e3c3beef75#diff-a22f622dd552a293d1933aae37f4fb5bR10

lib/jekyll/tags/include.rb
```diff
@@ -21,6 +21,10 @@ class IncludeTag < Liquid::Tag
         (?<params>.*)
       !x
 
+      FULL_VALID_SYNTAX = %r!\A\s*(?:#{VALID_SYNTAX}(?=\s|\z)\s*)*\z!
+      VALID_FILENAME_CHARS = %r!^[\w/\.-]+$!
+      INVALID_SEQUENCES = %r![\./][\./]+!
```
pathawks: `%r![./]{2,}!` ?

lib/jekyll/tags/include.rb
ashmaroli: > It is not obvious to me that `INVALID_SEQUENCES` is the complement of `%r!^[a-zA-Z0-9_/\.-]+$!`

I switched the calls here.. (as mentioned in the PR opener..)
`INVALID_SEQUENCES = %r![\./][\./]+!`
`VALID_FILENAME_CHARS = %r!^[\w/\.-]+$!`

lib/jekyll/tags/include.rb
```diff
@@ -21,6 +21,10 @@ class IncludeTag < Liquid::Tag
         (?<params>.*)
       !x
 
+      FULL_VALID_SYNTAX = %r!\A\s*(?:#{VALID_SYNTAX}(?=\s|\z)\s*)*\z!
```
pathawks: * `(?=\s|\z)\s*)*\z` It seems like there might be multiple ways for this to match. That's bad for performance.
 * What is the purpose of the `(?=\s|\z)`?
 * This matches the empty string (but `VALID_SYNTAX` does not). Is that intentional?

I see you are just hoisting this from `validate_params` so maybe these issues should be addressed separately?

lib/jekyll/tags/include.rb
ashmaroli: is this faster or just more idiomatic?

pathawks: Easier to read. More obvious what the intent is.

lib/jekyll/tags/include.rb
ashmaroli: - `VALID_SYNTAX` is probably a poorly named constant because its actually intended to just validate `params` in the include tag instead of the entire tag markup..
- `FULL_VALID_SYNTAX` is a superset of `VALID_SYNTAX` so its allowed that the former matches more than the latter..
- `(?=\s|\z)` is to ensure that `params` is either singular `foo='bar'` or space-separated
 `foo='bar' alpha="beta"`. `foo='bar'alpha="beta"` wouldn't match.

pathawks: > `(?=\s|\z)` is to ensure that params is either singular foo='bar' or space-separated

So we can get rid of the following `\s*` and `\z`?

`%r!\A\s*(?:#{VALID_SYNTAX}(?=\s|\z)+)+!`?

"Each repetition of `VALID_SYNTAX` must be followed by either spaces or the end of the string."

pathawks:
@jekyllbot: :ship: +minor

PR #6850, opened 2018-03-15
List news excerpts on the index pages

ashmaroli:
So that `https://jekyllrb.com/news/` and `https://jekyllrb.com/news/releases/` are of smaller sizes (lesser bandwidth usage on requests..)


DirtyF:
We don't want to display the first sentence without any more context (or edit all existing posts for a marginal gain).

screenshot 2018-03-15 15 46 15

PR #6854, opened 2018-03-17
Add support for `*.xhtml` files

ExE-Boss:
This adds proper support for `*.xhtml` files when using `jekyll serve`.


pathawks:
I haven't seen `.xhtml` actually used as an extension; only XHTML served with `.html` extension.

Can you describe why this change is needed?

DirtyF:
This does not look like a 90% user case, I'm tempted to label this as a `wontfix`

ExE-Boss:
`.xhtml` is the extension for [`application/xhtml+xml`](https://www.iana.org/assignments/media-types/application/xhtml+xml), whereas `.html` automatically denotes [`text/html`](https://www.iana.org/assignments/media-types/text/html). Because of this XHTML files won’t be served correctly as `application/xhtml+xml` without the `.xhtml` extension.

Also, GitHub Pages supports `.xhtml` correctly, while this doesn’t.

pathawks:
> `.xhtml` is the extension for `application/xhtml+xml`

Where are you getting this information? I do not believe that is correct.

pathawks:
I tend to agree with @DirtyF that this change wouldn't add value for the majority of our users, but if it brings the behavior of `jekyll serve` closer to that of GHP... 🤷‍♂️

/cc: @jekyll/stability

oe:
One thing - this could have some tests, just to future-proof it.

ExE-Boss:
@oe I’ve now implemented the tests (I actually kind‑of forgot about this until I checked my open PRs because [of this article I came across on Medium prompting readers to do so](https://medium.com/pnpm/why-package-managers-need-hook-systems-b8125d8b3dc7)).

oe:
@jekyllbot: merge +minor

oe:
@ExE-Boss Thank you for coming back to this!

PR #6856, opened 2018-03-18
Store and retrieve converter instances for Jekyll::Filters via a hash

ashmaroli:
Store converter instances into, and retrieve from a hash.
Since the associated filters *may* be called frequently within single Liquid template for various inputs, and the template itself *may* be called for a ton of documents / pages in a large site, its better to opt for hash-lookups over multiple `Array#find`

### Summary:
  - Move to using multiple hash lookups against multiple `Array#find`
  - Existing local variables have been dropped since they serve no bigger purpose than to be created, used once and immediately *garbage-collected* away
  - `@converter_instances` remains defined for the duration of current `Site` instance since initialized converters don't change when the site regenerates

### DIFF
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -281,9 +281,10 @@ module Jekyll
     # klass - The Class of the Converter to fetch.
     def find_converter_instance(klass)
       @find_converter_instance ||= {}
-      @find_converter_instance[klass] = \
-        converters.find { |klass_| klass_.instance_of?(klass) } || \
-        raise("No Converters found for #{klass}")
+      @find_converter_instance[klass] ||= begin
+        converters.find { |converter| converter.instance_of?(klass) } || \
+          raise("No Converters found for #{klass}")
+      end
     end
 
     # klass - class or module containing the subclasses.
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -276,6 +276,17 @@ module Jekyll
     end
     alias_method :to_liquid, :site_payload
 
+    # Store converter instances for faster multiple retrievals
+    #
+    # klass - The Class of the Converter instance to be retrieved
+    #
+    # Returns the requested Converter instance or raises an exception via
+    # `Jekyll::Site#find_converter_instance`
+    def converter_instance_of(klass)
+      @converter_instances ||= {}
+      @converter_instances[klass] = find_converter_instance(klass)
+    end
+
     # Get the implementation class for the given Converter.
     # Returns the Converter instance implementing the given Converter.
     # klass - The Class of the Converter to fetch.
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -276,23 +276,13 @@ module Jekyll
     end
     alias_method :to_liquid, :site_payload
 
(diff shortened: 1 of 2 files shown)


pathawks:
Why not just change `find_converter_instance` to memoize? Does this really result in a big performance boost?

ashmaroli:
> Why not just change `find_converter_instance` to memoize? Does this really result in a big performance boost?

We can't memoize by the traditional approach (using `||=`) because the method takes a parameter and the Site class is already packed with `instance_variables`..
So, its better we take the alternative approach using a hash to cache and retrieve objects..

The performance boost is not much to write home about.. therefore it won't be perceivable to the end user..

pathawks:
> because the method takes a parameter

I must be missing something. Why not use the same hash table approach from `converter_instance_of` here, but in the `find_converter_instance` method instead? That way, any code that already uses `find_converter_instance` benefits.

> therefore it won't be perceivable to the end user

So why is it worth the additional complexity?

ashmaroli:
> Why not use the same hash table approach from `converter_instance_of` here, but in the `find_converter_instance` method instead? That way, any code that already uses `find_converter_instance` benefits

Ah.. I see it now..
Shot myself in the foot back there when i said `Site` is already loaded with `instance_variables`..

ashmaroli:
> So why is it worth the additional complexity?

If a layout were to call `markdownify` filter 20 times and this layout was used by 50 documents,
`converters.find { |klass_| klass_.instance_of?(klass) }` would be executed `20*50` times..
Now if another layout in the same site had similar calls for `smartify` or `scssify` filters............

lib/jekyll/site.rb
```diff
@@ -279,9 +279,10 @@ def site_payload
     # Get the implementation class for the given Converter.
     # Returns the Converter instance implementing the given Converter.
     # klass - The Class of the Converter to fetch.
-
     def find_converter_instance(klass)
-      converters.find { |klass_| klass_.instance_of?(klass) } || \
+      @find_converter_instance ||= {}
+      @find_converter_instance[klass] = \
+        converters.find { |klass_| klass_.instance_of?(klass) } || \
```
Crunch09: I realise this didn't change in your PR but why is the block variable called `klass_` instead of e.g. `converter` ? Seems a bit odd to me.

lib/jekyll/site.rb
```diff
@@ -279,9 +279,10 @@ def site_payload
     # Get the implementation class for the given Converter.
     # Returns the Converter instance implementing the given Converter.
     # klass - The Class of the Converter to fetch.
-
     def find_converter_instance(klass)
-      converters.find { |klass_| klass_.instance_of?(klass) } || \
+      @find_converter_instance ||= {}
+      @find_converter_instance[klass] = \
```
Crunch09: I'm probably missing something but shouldn't this be `@find_converter_instance[klass] ||=` ?

ashmaroli: You're correct @Crunch09. Nice catch :+1:

lib/jekyll/site.rb
ashmaroli: Agreed. I can change it since it's part of the diff..

pathawks:
@jekyllbot: :ship: +minor

PR #6859, opened 2018-03-18
Fixed Minor Grammar Issue

EarthSchlange:
This adds the missing 'and' between major updates ... detailed instructions.


pathawks:
I agree that this is awkwardly worded. I _think_ it is referring to instructions for upgrading from one major version to the next (ie: 2.x to 3.x).

PR #6860, opened 2018-03-19
Minimize array allocations in the `where` filter

ashmaroli:
Since `Kernel#Array` creates an array out of the returned `item_property`, calling `map!` here only modifies elements of the said array which is already being done currently (albeit by creating more copies of the array..)

This reduces a lot a Array allocations especially when a large *`Collection`* is piped to the `where` filter multiple times in Liquid..


oe:
@jekyllbot: merge +dev

PR #6861, opened 2018-03-19
WIP: add utility methods to delete leading and trailing substring w/o regexps

ashmaroli:
- [x] initial implementation
- [x] basic benchmark script
- [ ] unit tests


ashmaroli:
Basically an experiment to replace `gsub(%r!\A#{substr}|#{substr}\z!, "")` , `sub(%r!\A#{substr}!, "")` and `sub(%r!#{substr}\z!, "")` with something that doesn't involve regex..

PR #6871, opened 2018-03-25
Introduce a 'verbatim' configuration option

feuerbach:
I wanted to publish an R markdown file ("Rmd") with a YAML front matter on my jekyll website, so that I could link to it and others could download it. However, jekyll recognized the file as markdown and compiled it to html, which is not what I wanted.

With this PR, you can include

```
  verbatim:
    - files
```

in your `_config.yml`, and everything under the `files` directory will be copied verbatim, even if it contains a YAML front matter.

This PR is just to solicit feedback. If you're open to merging it, I'll write the tests and documentation.

### DIFF
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -22,6 +22,7 @@ module Jekyll
         Gemfile Gemfile.lock node_modules vendor/bundle/ vendor/cache/ vendor/gems/
         vendor/ruby/
       ),
+      "verbatim"            => [],
       "keep_files"          => [".git", ".svn"],
       "encoding"            => "utf-8",
       "markdown_ext"        => "markdown,mkdown,mkdn,mkd,md",
--- a/lib/jekyll/entry_filter.rb
+++ b/lib/jekyll/entry_filter.rb
@@ -62,6 +62,17 @@ module Jekyll
       end
     end
 
+    def verbatim?(entry)
+      glob_include?(site.verbatim, relative_to_source(entry)).tap do |verbatim|
+        if verbatim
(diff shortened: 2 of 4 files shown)


chrisfinazzo:
Hi @feuerbach, good to have you here.

I see how this could be an issue as `.rmd` files are not simply Markdown, but rather Markdown+R-specific data. When you have time, feel free to send a PR which describes this config option.

cc/ @jekyll/stability

DirtyF:
I don't feel like we should add another config option for this, I'd rather see Jekyll's conventions adapt to this edge case, e.g. with a special folder that gets copied as is.

parkr:
If you're able to use `+++` instead of `---` for your YAML front matter, then I'd recommend something like that for now!

ashmaroli:
@feuerbach Apart from "posts", every other (non-*excluded*) file is destined to be "converted" *if the file contains YAML Front Matter **at the very beginning of the file***..

If you can edit the contents of the `rmd` files, the easiest thing to do is to simply add an empty line at the very beginning of the file..

feuerbach:
@ashmaroli yes, I am aware of that. YAML front matter is not something jekyll-specific, and Rmd files contain it (at the very beginning of the file) by design.

feuerbach:
@parkr I haven't heard of a `+++` front matter. Anyway, the point is that you should be able to drop any files in the `files` folder without even thinking how they start, and be sure that jekyll won't get in your way.

ashmaroli:
Closing this as it seems `.Rmd` files are not processed (as a Markdown file) by Jekyll on `master`.

Issue #6252, opened 2017-07-27
Remove usage of default_proc everywhere


parkr:
For any Hash using default_proc, remove it and replace its functionality through some other means (often by manually setting values or transforming keys).

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>3.3-stable</code> or <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, please consider building it first as a plugin. Jekyll 3 introduced [hooks](http://jekyllrb.com/docs/plugins/#hooks) which provide convenient access points throughout the Jekyll build pipeline whereby most needs can be fulfilled. If this is something that cannot be built as a plugin, then please provide more information about why in order to keep this issue open.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>3.3-stable</code> or <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, please consider building it first as a plugin. Jekyll 3 introduced [hooks](http://jekyllrb.com/docs/plugins/#hooks) which provide convenient access points throughout the Jekyll build pipeline whereby most needs can be fulfilled. If this is something that cannot be built as a plugin, then please provide more information about why in order to keep this issue open.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

Issue #5355, opened 2016-09-14
Jekyll Hooks :site, :pre_render (or probably the payload itself) leaks into configurations.


envygeeks:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [ ] This is a feature request.

---
- [x] I am on (or have tested on) **_macOS**_ 10+
- [x] I am on (or have tested on) **_Debian/Ubuntu**_ GNU/Linux
- [x] I am on (or have tested on) **_Fedora**_ GNU/Linux
- [x] I am on (or have tested on) **_Arch**_ GNU/Linux
- [ ] I am on (or have tested on) **_Other**_ GNU/Linux
- [x] I am on (or have tested on) **_Windows**_ 10+

---
- [ ] I was trying to install.
- [x] There is a broken Plugin API.
- [ ] I had an error on GitHub Pages, and I have reproduced it locally.
- [ ] I had an error on GitHub Pages, and GitHub Support said it was a Jekyll Bug.
- [ ] I had an error on GitHub Pages and I did not test it locally.
- [x] I was trying to build.
- [x] It was another bug.
## My Reproduction Steps

``` ruby
Jekyll::Hooks.register :site, :pre_render do |site, payload|
  payload.site["seriously"] = "Yeah... seriously."
  $stdout.puts(site.config[
    "seriously"
  ])
end
```
### Output

```
jordon@tardis ~/development/<secret>/spec/fixture → jekyll b -w --trace
Configuration file: /home/jordon/development/<secret>/spec/fixture/_config.yml
            Source: /home/jordon/development/<secret>/spec/fixture
       Destination: /home/jordon/development/<secret>/spec/fixture/_site
 Incremental build: disabled. Enable with --incremental
      Generating... 
Yeah... seriously.
                    done in 0.022 seconds.
 Auto-regeneration: enabled for '/home/jordon/development/<secret>/spec/fixture'
```
## The Output I Wanted

I want it not to leak.

Crunch09:
@envygeeks With leaking you mean that it should not be possible to update the configuration by setting the payload in a hook, right? 😅

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

[...]

PR #6874, opened 2018-03-25
WIP: allow `pages` and `documents` to be marshalled

Crunch09:
I'd really like to help bring parallel builds to jekyll. One of the blockers which has been mentioned in previous PRs / issues like #5355 , #5344, #5823 was that `Jekyll::Page` and `Jekyll::Document` instances can't be marshalled as one of their instance variables is a Hash with a [`default_proc`](https://docs.ruby-lang.org/en/2.5.0/Hash.html#method-i-default_proc).

@parkr opened #6252 to track the removal of `default_proc` and mentioned in https://github.com/jekyll/jekyll/pull/6401#issuecomment-341541036 that removing the `default_proc` before marshalling and re-applying it after the marshalled object has been loaded would be best, so this is what this PR tries to achieve.

It is still WIP because it turned out that a [`Liquid::Context`](https://github.com/Shopify/liquid/blob/master/lib/liquid/context.rb) also can't be marshalled so it is removed before marshalling a `Jekyll::Drops::Drop` object. I don't know if that would be acceptable so would love some feedback on the current approach from @jekyll/core before writing better and more tests.

### DIFF
--- a/lib/jekyll/drops/drop.rb
+++ b/lib/jekyll/drops/drop.rb
@@ -209,6 +209,18 @@ module Jekyll
         return yield(key) unless block.nil?
         return default unless default.nil?
       end
+
+      private
+      def marshal_dump
+        (instance_variables - [:@context]).map do |attr|
+          { attr => instance_variable_get(attr) }
+        end.inject(:merge)
+      end
+
+      private
+      def marshal_load(marshalled_data)
+        marshalled_data.each { |key, value| instance_variable_set(key, value) }
+      end
     end
   end
 end
--- a/lib/jekyll/drops/drop.rb
+++ b/lib/jekyll/drops/drop.rb
@@ -212,7 +212,7 @@ module Jekyll
 
       private
       def marshal_dump
-        (instance_variables - [:@context]).map do |attr|
+        (instance_variables - [:@context]).map! do |attr|
           { attr => instance_variable_get(attr) }
         end.inject(:merge!)
       end
--- a/lib/jekyll/drops/drop.rb
+++ b/lib/jekyll/drops/drop.rb
@@ -214,7 +214,7 @@ module Jekyll
       def marshal_dump
         (instance_variables - [:@context]).map do |attr|
           { attr => instance_variable_get(attr) }
-        end.inject(:merge)
+        end.inject(:merge!)
       end
 
       private
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -36,7 +36,7 @@ module Jekyll
         categories_from_path(collection.relative_directory)
       end
 
-      set_default_proc
+      data.default_proc = data_default_proc
 
       trigger_hooks(:post_init)
     end
@@ -505,8 +505,8 @@ module Jekyll
     end
 
     private
-    def set_default_proc
-      data.default_proc = proc do |_, key|
+    def data_default_proc
+      proc do |_, key|
         site.frontmatter_defaults.find(relative_path, collection.label, key)
       end
     end
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -3,6 +3,7 @@
 module Jekyll
   class Document
     include Comparable
+    include Utils::MarshalWithoutDefaultProc
     extend Forwardable
 
     attr_reader :path, :site, :extname, :collection
@@ -35,9 +36,7 @@ module Jekyll
         categories_from_path(collection.relative_directory)
       end
 
-      data.default_proc = proc do |_, key|
-        site.frontmatter_defaults.find(relative_path, collection.label, key)
-      end
+      set_default_proc
 
       trigger_hooks(:post_init)
     end
@@ -502,6 +501,13 @@ module Jekyll
     def generate_excerpt
       if generate_excerpt?
         data["excerpt"] ||= Jekyll::Excerpt.new(self)
+      end
+    end
+
+    private
+    def set_default_proc
+      data.default_proc = proc do |_, key|
+        site.frontmatter_defaults.find(relative_path, collection.label, key)
       end
     end
   end
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -414,6 +414,12 @@ module Jekyll
       })
     end
 
+    def data_default_proc
+      proc do |_, key|
+        site.frontmatter_defaults.find(relative_path, collection.label, key)
+      end
+    end
+
     private
     def merge_categories!(other)
       if other.key?("categories") && !other["categories"].nil?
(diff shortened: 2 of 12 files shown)


lib/jekyll/drops/drop.rb
```diff
@@ -209,6 +209,18 @@ def fetch(key, default = nil, &block)
         return yield(key) unless block.nil?
         return default unless default.nil?
       end
+
+      private
+      def marshal_dump
+        (instance_variables - [:@context]).map do |attr|
+          { attr => instance_variable_get(attr) }
+        end.inject(:merge)
```
ashmaroli: `array.inject(:merge)` was known to be *very slow..!* (Not sure if it still is..) in comparison to `array.inject(:merge!)` and `array.each_with_object` (..though not sure that `#each_with_object` defines the `default_proc` for the *internal memo*)

parkr:
@Crunch09 I think we can do this a bit more simply. What do you think about something like a `Marshaller` class which can handle this? Copying data seems brittle. I'd rather go in, remove the `default_proc`, marshal, then add the `default_proc` back in. Run this in Ruby or IRB:

```ruby
class Page
  attr_reader :data
  
  def initialize
    @name = 'foo'
    @dir = 'baz'
    @data = {'hi' => 'there', dear: 'friends'}
    @data.default_proc = data_default_proc
  end
  
  def data_default_proc
    proc { "in the proc" }
  end
end

class Marshaller
  def self.dump(obj)
    obj.data.default_proc = nil
    Marshal.dump(obj)
  ensure
    reset_default_proc(obj)
  end
  
  def self.load(bytes)
    obj = Marshal.load(bytes)
    reset_default_proc(obj)
    obj
  end
  
  def self.reset_default_proc(obj)
    return unless obj.respond_to?(:data) && obj.data
    obj.data.default_proc = obj.data_default_proc
  end
end

page = Page.new
p page
dumped = Marshaller.dump(page)
p dumped
loaded_page = Marshaller.load(dumped)
p loaded_page
p loaded_page.data['unknown_key']
```

lib/jekyll/drops/drop.rb
Crunch09: Good catch, thanks! 👍  I just ran a quick benchmark and `inject(:merge)` is indeed much, much slower than `injcect(:merge!)`. Will update

Crunch09:
@parkr Thanks for your feedback! I really like the abstraction of a dedicated `Marshaller` class. 💯 
The problem we still have with that though is that `Marshal.dump(page_object)` will also cause a marshalling of all instance variables of `page_object`. This will subsequently cause a marshalling of `documents` (as `pages` and `documents` are connected via a common `site` object).
Those `documents` still have the `default_proc` on their `data` attribute and the marshalling will fail for that reason.

*Simplified* example:
```ruby
class Page
  attr_reader :data, :document
  
  def initialize
    @data = {'hi' => 'there', dear: 'friends'}
    @data.default_proc = data_default_proc
  end
  
  def data_default_proc
    proc { "in the proc" }
  end
end

class Document
  attr_reader :data, :page
  
  def initialize
    @data = {'hello' => 'world', hi: 'there'}
    @data.default_proc = data_default_proc
  end
  
  def data_default_proc
    proc { "in the proc" }
  end
end


page = Page.new
page.document = Document.new
Marshaller.dump(page)
# will fail, `page` object itself would be fine
# but it has an associated `document` with a default proc
```

I think we can still improve the code especially to no longer copy `data` as you mentioned but i think we need to use `marshal_dump` and `marshal_load` to do it. I will try to update and push a new version tomorrow!

lib/jekyll/drops/drop.rb
ashmaroli: it would be nice to have that `inject-merge-vs-inject-merge!` benchmark added to *our* Benchmark Suite.. (perhaps better in a separate PR)

lib/jekyll/drops/drop.rb
```diff
@@ -209,6 +209,18 @@ def fetch(key, default = nil, &block)
         return yield(key) unless block.nil?
         return default unless default.nil?
       end
+
+      private
+      def marshal_dump
+        (instance_variables - [:@context]).map do |attr|
```
ashmaroli: How may times will this method get called in an average-to-large site?
We can reduce Array allocations by using `map!` here, since `Array#-` already creates a copy of given array, `instance_variables`)

Crunch09: It's going to be called at most once for every drop. I just checked it with the jekyll docs site:
```ruby
ObjectSpace.each_object(Jekyll::Drops::Drop).count # => 257
page = ObjectSpace.each_object(Jekyll::Page).first
Marshal.dump(page) # `marshal_dump` in `Jekyll::Drops::Drop` is called 252 times
```

Thanks for the suggestion with `#map!` 👍  i updated the code

ashmaroli:
@Crunch09 I was thinking about this lately and I always tend to get stuck on the "concurrency" factor..
Say, we have the technical possibility to marshal all of our objects, but then what exactly does the parallel build involve?

By going "parallel", does it mean we may "read" all files and directories at the same time? or run our "generators" in parallel, or "render" in parallel..?
Is all of that possible since we essentially run all those "processes" by *iterating* (one-by-one) through arrays viz. `site.pages`, `site.documents`, etc, etc...

Crunch09:
@ashmaroli Sorry for the late response!

> Say, we have the technical possibility to marshal all of our objects, but then what exactly does the parallel build involve?

I'm also not quite sure about how enabling marshalling for pages and documents would help with concurrency but it was mentioned in previous discussions as a requirement / blocker.

> By going "parallel", does it mean we may "read" all files and directories at the same time? or run our "generators" in parallel, or "render" in parallel..?

Yeah, i think ultimately it would be great to parallelize as many parts of the build process as we can. As a first step i wanted to look at `render` and see if we could run that in parallel. I've started to look at bundler's [parallel installer](https://github.com/bundler/bundler/blob/master/lib/bundler/installer/parallel_installer.rb) as an example approach for concurrenct tasks.

ashmaroli:
@Crunch09 Thank you for responding.
From past experience, my opinion is that both "concurrent" and "parallel" rendering is "very difficult" because of Jekyll's internal architecture.

For example, if `index.md" gets rendered *slightly before* `site.posts` get rendered, then the `for loop` in "index.html" is *may contain unrendered* post titles and excerpts.. This could be a may be or may be not situation but possible and difficult to debug..

Yeah, its related to the current PR, just wanted to mention it so that you'll have it in the back of your head as part of the bigger picture.

Crunch09:
Thanks @ashmaroli :) 👍  We'll definetely need to create a dependency tree to handle these kind of issues!

pathawks:
@Crunch09 

> We'll definetely need to create a dependency tree to handle these kind of issues!

Coming soon 😉

PR #6881, opened 2018-03-28
Docs: Install Jekyll on macOS

DirtyF:
This PR aims at providing detailed explanation to install Jekyll on macOS.

screenshot update after the review:

jekyll-install-macos


docs/_docs/installation.md
```diff
@@ -4,42 +4,51 @@ description: Official guide to install Jekyll on macOS, GNU/Linux or Windows.
...
+- [Ruby](https://www.ruby-lang.org/en/downloads/) version 2.2.5 or above, including all development headers (ruby installation can be checked by running `ruby -v`)
+- [RubyGems](https://rubygems.org/pages/download) (which you can check by running `gem -v`)
+- [GCC](https://gcc.gnu.org/install/) and [Make](https://www.gnu.org/software/make/) (in case your system doesn't have them installed, which you can check by running `gcc -v`,`g++ -v`  and `make -v` in your system's command line interface)
 
 ## Install on macOS {#macOS}
 
-We only cover latest macOS High Sierra 10.13 here, who ships a new Ruby version, older systems will need to install Homebrew, see below.
+We only cover latest macOS High Sierra 10.13 here, who comes with Ruby 2.3.3, older systems will need to [install a more recent Ruby version via Homebrew](#homebrew).
 
 First, you need to [install Xcode from the AppStore](https://itunes.apple.com/fr/app/xcode/id497799835?mt=12). Then to install the command-line tools, open a terminal and run:
```
coliff: Suggest to use URL:
`https://itunes.apple.com/app/xcode/id497799835?mt=12`
instead of country-specific

coliff: AppStore should be two words (App Store)

docs/_docs/installation.md
```diff
@@ -4,42 +4,51 @@ description: Official guide to install Jekyll on macOS, GNU/Linux or Windows.
...
 ruby -v
 2.3.3
-gem --version
-2.7.6
+
+gem update --system
 ```
 
-Great, let's install Jekyll, we'll also need bundler to help us handle plugins and themes:
+Great, let's install Jekyll, we'll also need [bundler](http://bundler.io/) to help us handle [plugins](../plugins) and [themes](../themes):
```
coliff: http://bundler.io/ can be HTTPS - so https://bundler.io/

docs/_docs/installation.md
```diff
@@ -4,42 +4,51 @@ description: Official guide to install Jekyll on macOS, GNU/Linux or Windows.
...
+Before you start, make sure your system has the following:
+
+- [Ruby](https://www.ruby-lang.org/en/downloads/) version 2.2.5 or above, including all development headers (ruby installation can be checked by running `ruby -v`)
+- [RubyGems](https://rubygems.org/pages/download) (which you can check by running `gem -v`)
+- [GCC](https://gcc.gnu.org/install/) and [Make](https://www.gnu.org/software/make/) (in case your system doesn't have them installed, which you can check by running `gcc -v`,`g++ -v`  and `make -v` in your system's command line interface)
 
 ## Install on macOS {#macOS}
 
-We only cover latest macOS High Sierra 10.13 here, who ships a new Ruby version, older systems will need to install Homebrew, see below.
+We only cover latest macOS High Sierra 10.13 here, who comes with Ruby 2.3.3, older systems will need to [install a more recent Ruby version via Homebrew](#homebrew).
```
coliff: Shoudl this be:
> We only cover latest macOS High Sierra 10.13 here, which comes with Ruby 2.3.3...
(*which* not *who*)

docs/_docs/installation.md
chrisfinazzo: Better idea - the Mac App Store is terrible, [avoid][] it if you possibly can. I know this requires auth to get past the login, but if we're being honest, most people installing on macOS probably know about this already and can deal with this requirement.

[avoid]: developer.apple.com

docs/_docs/installation.md
```diff
@@ -4,42 +4,51 @@ description: Official guide to install Jekyll on macOS, GNU/Linux or Windows.
...
 
-Installing Jekyll should be straight-forward if all requirements are met.
-
+- [Requirements](#requirements)
 - [Install on macOS](#macOS)
-- [Install on GNU/Linux](#Linux)
 - [Install on Windows](../windows/)
+- [Upgrade Jekyll](#upgrade-jekyll)
+
+Installing Jekyll should be straight-forward if all requirements are met.
```
chrisfinazzo: "...if your system meets the requirements"

Consistent with details under the Requirements heading.

docs/_docs/installation.md
```diff
@@ -4,42 +4,51 @@ description: Official guide to install Jekyll on macOS, GNU/Linux or Windows.
...
 ```sh
 ruby -v
 2.3.3
-gem --version
-2.7.6
+
+gem update --system
 ```
 
-Great, let's install Jekyll, we'll also need bundler to help us handle plugins and themes:
```
chrisfinazzo: Feels like this should be two sentences.

"Great, let's install Jekyll. We'll also need Bundler to help us handle plugins and themes:"

From that point on, continue.

docs/_docs/installation.md
chrisfinazzo: Better idea - the Mac App Store is terrible, [avoid][] it if you possibly can. I know this requires auth to get past the login, but if we're being honest, most people installing on macOS probably know about this already and can deal with this requirement.

To that end, I have found [this][] document to be more useful regarding how to obtain Xcode.

[avoid]: developer.apple.com

[this]:  http://ericasadun.com/2016/03/22/xcode-upgrades-lessons-learned/

DirtyF:
Thanks @coliff @chrisfinazzo adressed your feedback.

docs/_docs/installation.md
```diff
@@ -1,138 +1,164 @@
...
-<div class="note info">
-  <h5>Installing Xcode Command-Line Tools</h5>
-  <p>
-    If you run into issues installing Jekyll's dependencies which make use of
-    native extensions and are using macOS, you will need to install Xcode
-    and the Command-Line Tools it ships with. Download them in
-    <code>Preferences &#8594; Downloads &#8594; Components</code>.
-  </p>
-</div>
+Great, let's install Jekyll. We also need [bundler](https://bundler.io/) to help us handle [plugins](../plugins) and [themes](../themes):
```
oe: bundler is capitalized iirc

docs/_docs/installation.md
```diff
@@ -1,138 +1,164 @@
...
+Developers often use [rbenv](https://github.com/rbenv/rbenv) to manage multiple Ruby versions.
 
 ```sh
-git clone git://github.com/jekyll/jekyll.git
-cd jekyll
-script/bootstrap
-bundle exec rake build
-ls pkg/*.gem | head -n 1 | xargs gem install -l
+# Install rbenv and ruby-build
+brew install rbenv
```
oe: does the homebrew installation of rbenv include ruby-build?

docs/_docs/installation.md
```diff
@@ -1,138 +1,164 @@
...
-Before you start developing with Jekyll, you may want to check that you're up to date with the latest version. To find your version of Jekyll, run one of these commands:
+Before you start developing with Jekyll, you may want to check that you're up to date with the latest version. To find the currently installed version of Jekyll, run one of these commands:
 
 ```sh
 jekyll --version
 gem list jekyll
 ```
 
-You can also use [RubyGems](https://rubygems.org/gems/jekyll) to find the current versioning of any gem. But you can also use the `gem` command line tool:
+You can use RubyGems to find [the current versioning of Jekyll](https://rubygems.org/gems/jekyll). Another way to check if you have the latest version is to run the command `gem outdated`. This will provide a list of all the gems on your system that need to be updated. If you aren't running the latest version, run this command:
```
oe: s/versioning/version

docs/_docs/installation.md
chrisfinazzo: Just tried it...

`brew info` lists it as "recommended" for `rbenv` and will download the latest tarball during `rbenv` installation process.

On macOS 10.13.

oe: ok cool!

docs/_docs/installation.md
```diff
@@ -1,138 +1,164 @@
...
 
-<div class="note info">
-  <h5>Running Jekyll on Windows</h5>
-  <p>
-    While Windows is not officially supported, it is possible to get Jekyll running
-    on Windows. Special instructions can be found on our
-    <a href="../windows/#installation">Windows-specific docs page</a>.
-  </p>
-</div>
+We only cover latest macOS High Sierra 10.13 here, which comes with Ruby 2.3.3, older systems will need to [install a more recent Ruby version via Homebrew](#homebrew).
```
mattr-: Let's remove `latest` from this sentence since we're explicitly referencing the version number and code name and once a new release comes out, it won't be the latest anymore.

docs/_docs/installation.md
```diff
@@ -1,138 +1,164 @@
...
 
-## Install with RubyGems
+First, you need to [install Xcode from the App Store](https://itunes.apple.com/app/xcode/id497799835?mt=12) or [from Apple Developer website](https://developer.apple.com/xcode/). Then to install the command-line tools, open a terminal and run:
```
mattr-: Technically, XCode isn't required. On macOS 10.13, you can just run `xcode-select --install` in a terminal window to get the command line tools installed.

docs/_docs/installation.md
```diff
@@ -1,138 +1,164 @@
...
+
+# Install gems
+bundle install
 ```
 
-This will install the latest pre-release. If you want a particular pre-release,
-use the `-v` switch to indicate the version you'd like to install:
+From there you can either use a [theme](../themes/) or create your own layouts.
+
+### Install via Homebrew {#homebrew}
```
mattr-: If macOS 10.13 has the appropriate version requirements for Ruby out of the box, I'm not sure I see a reason to add a section on installing via Homebrew. 

If we want to keep this section. I think it would be better if we rearrange the documentation to be something along the lines of:

```
                 |-> Set up Ruby included with the OS  +
Install on macOS +                                     | -> Install Jekyll -> Generate your first website
                 |-> Install a newer Ruby via Homebrew +

```

I was quite confused when I got to this section for two reasons:
1. We don't install Jekyll via Homebrew, so just what the heck is going on here :smiley:
2. We already set up Ruby, we don't need to do it again.

What do y'all think?

docs/_docs/installation.md
```diff
@@ -1,138 +1,164 @@
...
-[RubyGems](https://rubygems.org/pages/download). At the terminal prompt,
-simply run the following command to install Jekyll:
+Check your Ruby version, and upgrade [RubyGems](https://rubygems.org/pages/download):
 
 ```sh
-gem install jekyll
+ruby -v
+2.3.3
+
+gem update --system
```
mattr-: Do we require a newer version of Rubygems outside of what's installed by default? If not, we shouldn't recommend that people upgrade Rubygems on their system.

DirtyF:
@mattr- @oe thanks for the feedack, updated the screenshot accordingly.

Most users can stop reading after the first part (working with macOS system ruby version), other sections are optional, so I kept the the `jekyll new` part at the beginning.

PR #6882, opened 2018-03-28
Docs: describe difference between tags and categories

steelman:


DirtyF:
@jekyllbot: merge +docs

PR #6885, opened 2018-03-30
Allow front matter defaults to be applied properly to documents gathered under custom `collections_dir`

### LINKED ISSUE #6879: Config globs not working as expected

---

---

- [x] I was trying to build.
- [ ] It was another bug.

## My Reproduction Steps

I'm currently assuming this is user error, so I am not providing repro steps at this point. If this sounds like a possible bug, I'll figure out a way to create a test repo.

I'd like to upgrade from 3.7.0 to 3.7.3, but the way path globbing changed has broken things, and my attempts to update based on the docs haven't yielded success.

I have a collection set up with various types of subfolders that require different defaults, sort of like this:

```
_collections/ (collections_dir)
  _people/
    employees/
      personA.md
      personB.md
    alumni/
      personC.md
      personD.md
```

**old value**
`path: _collections/_people/employees/`

**new value**
`path: _collections/_people/employees/*`

I assumed this was the way to continue to glob directories, but all of my defaults are no longer set, causing breakage.

When I do a verbose log, I see a bunch of this:

`Globbed Scope Path: _collections/_people/employees/cameron-mcefee.md`

I assume this means the glob is matching the files correctly (never mind that it seems repeat this for every single file), so I'm confused about why the defaults don't get set.

ashmaroli:
Resolves #6879 

- [x] test if reported issue is reproducible
- [x] initial implementation


lib/jekyll/frontmatter_defaults.rb
```diff
@@ -129,6 +130,14 @@ def path_is_subpath?(path, parent_path)
       false
     end
 
+    def strip_collections_dir(path)
+      collections_dir  = @site.config["collections_dir"]
+      slashed_coll_dir = "#{collections_dir}/"
+      return path if collections_dir.empty?
```
oe: I feel like this and the following line could be consolidated into one block to clear up confusion.

oe:
@jekyllbot: merge +bug

PR #6886, opened 2018-03-30
Cache computing valid front matter defaults sets

ashmaroli:
Potential `breaking-change`

- memoize `FrontmatterDefaults#valid_sets` to reduce Array allocations
- could be a potential `breaking-change` as plugins using existing behavior to alter the `defaults:` configuration and have it take effect will no longer work as expected.
**A better alternative for these plugins would be to use Jekyll's `:after_init` hook.**


pathawks:
Any benchmarks?

PR #6887, opened 2018-03-30
Store regexp in a constant

ashmaroli:


pathawks:
@jekyllbot: :shipit: +minor

PR #6888, opened 2018-03-30
Cache matched defaults sets for given parameters

ashmaroli:
..to avoid unnecessary Array allocations due to `Array#select` for the same given parameters.


pathawks:
Any benchmarks?

pathawks:
Tried doing some benchmarking, but got the error `undefined method `superclass' for nil:NilClass`

pathawks:
```text
  v3.8.2    287.54
  master    304.47
   #6888    284.61
```

Looks like a nice improvement :+1:

ashmaroli:
Why is our `master` taking so much extra time..?
@pathawks can you run the benchmark once more, please?

DirtyF:
@jekyllbot: merge +minor

ashmaroli:
Awesome!!

PR #6889, opened 2018-03-31
Refactor server start up command

kenchan0130:
I would like to use [jekyll-admin](https://github.com/jekyll/jekyll-admin).
However, we can not use it in the latest stable jekyll.

Therefore, I thought that it is necessary to review the current configuration.

I thought that server instance is appropriate from the point of view that it is meaningful to do it when setup rather than a function (`start_up_webrick `) to start WEBrick.

I have plan to patch jekyll-admin side when this PR is merged.

- [x] fix test

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -101,7 +101,11 @@ module Jekyll
           register_reload_hooks(opts) if opts["livereload"]
           setup(destination)
 
+          start_up_livereload(opts)
           start_up_webrick(opts, destination)
+
+          launch_browser @server, opts if opts["open_url"]
+          boot_or_detach @server, opts
         end
 
         def shutdown
@@ -220,16 +224,17 @@ module Jekyll
 
         private
         def start_up_webrick(opts, destination)
-          if opts["livereload"]
-            @reload_reactor.start(opts)
-          end
-
           @server = WEBrick::HTTPServer.new(webrick_opts(opts)).tap { |o| o.unmount("") }
           @server.mount(opts["baseurl"].to_s, Servlet, destination, file_handler_opts)
 
           Jekyll.logger.info "Server address:", server_address(@server, opts)
-          launch_browser @server, opts if opts["open_url"]
-          boot_or_detach @server, opts
+        end
+
+        private
+        def start_up_livereload(opts)
+          if opts["livereload"]
+            @reload_reactor.start(opts)
+          end
         end
 
         # Recreate NondisclosureName under utf-8 circumstance
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -99,10 +99,10 @@ module Jekyll
           opts = configuration_from_options(opts)
           destination = opts["destination"]
           register_reload_hooks(opts) if opts["livereload"]
-          setup(destination)
+          pre_setup(destination)
+          setup_webrick(opts, destination)
 
-          start_up_livereload(opts)
-          start_up_webrick(opts, destination)
+          @reload_reactor.start(opts) if opts["livereload"]
 
           launch_browser @server, opts if opts["open_url"]
           boot_or_detach @server, opts
@@ -183,7 +183,7 @@ module Jekyll
         # and making sure our destination exists.
 
         private
-        def setup(destination)
+        def pre_setup(destination)
           require_relative "serve/servlet"
 
           FileUtils.mkdir_p(destination)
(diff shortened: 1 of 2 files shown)


ashmaroli:
@kenchan0130 What exactly is the error from `jekyll-admin`?

DirtyF:
👎 if this breaks our current tests

kenchan0130:
@ashmaroli @mattr- 

Thank you for your review.

I'm sorry I should have explained to you more clearly.

The problem I want to solve is https://github.com/jekyll/jekyll-admin/issues/458.

I think WEBrick server start up should concentrate on server startup and is sending this patch.


@DirtyF 

I fixed the test.

lib/jekyll/commands/serve.rb
```diff
@@ -220,16 +224,17 @@ def webrick_opts(opts)
...
           @server = WEBrick::HTTPServer.new(webrick_opts(opts)).tap { |o| o.unmount("") }
           @server.mount(opts["baseurl"].to_s, Servlet, destination, file_handler_opts)
 
           Jekyll.logger.info "Server address:", server_address(@server, opts)
-          launch_browser @server, opts if opts["open_url"]
-          boot_or_detach @server, opts
+        end
+
+        private
+        def start_up_livereload(opts)
```
ashmaroli: I don't think this need to be a new method.. it's going to be called just once anyways..
So it better to simply *start livereload* directly in the `:process` method.

```ruby
def process(opts)
  [...]
  @reload_reactor.start(opts) if opts["livereload"]
  start_up_webrick(opts, destination)

  launch_browser(@server, opts) if opts["open_url"]
  boot_or_detach(@server, opts)
end
```

lib/jekyll/commands/serve.rb
```diff
@@ -101,7 +101,11 @@ def process(opts)
           register_reload_hooks(opts) if opts["livereload"]
           setup(destination)
 
+          start_up_livereload(opts)
           start_up_webrick(opts, destination)
+
+          launch_browser @server, opts if opts["open_url"]
+          boot_or_detach @server, opts
```
ashmaroli: Since you're refactoring these lines, might as well clean them up..
Its a common style in Jekyll to improve readability by using parantheses to scope parameters.

```ruby
launch_browser(@server, opts) if opts["open_url"]
boot_or_detach(@server, opts)
```

kenchan0130: OK, I see.

I'll try it.

lib/jekyll/commands/serve.rb
kenchan0130: I do not have any particular opinion, so I follow your suggestions.

https://github.com/jekyll/jekyll/pull/6889/commits/7596a20348fc2b3b3ff964a12b0d549e64f6ec20

lib/jekyll/commands/serve.rb
kenchan0130: Fixed by https://github.com/jekyll/jekyll/pull/6889/commits/b5b48c15efa895353e3798a4a664b5c4e4735500

lib/jekyll/commands/serve.rb
```diff
@@ -219,17 +223,11 @@ def webrick_opts(opts)
         #
 
         private
-        def start_up_webrick(opts, destination)
-          if opts["livereload"]
-            @reload_reactor.start(opts)
-          end
-
+        def setup_webrick(opts, destination)
           @server = WEBrick::HTTPServer.new(webrick_opts(opts)).tap { |o| o.unmount("") }
```
kenchan0130: I think that it may be possible to erase the instance variable `@server` by using `attr_accessor` .

lib/jekyll/commands/serve.rb
```diff
@@ -219,17 +223,11 @@ def webrick_opts(opts)
         #
 
         private
-        def start_up_webrick(opts, destination)
-          if opts["livereload"]
-            @reload_reactor.start(opts)
-          end
-
+        def setup_webrick(opts, destination)
```
ashmaroli: you can't change the method name (even if it is a `private` method) because it will break Jekyll Admin's monkey-patch which is against your primary goal with this PR

kenchan0130: Since I moved the location of processing such as boot and open browser,  I think that it is necessary to patch the jekyll-admin side in either case.

I am thinking to apply the following patch.

```ruby
class Serve < Command
  class << self
    def setup_webrick(opts, destination)
      super(opts, destination)
      jekyll_admin_monkey_patch
    end
    
    def jekyll_admin_monkey_patch
      @server.mount "/admin", Rack::Handler::WEBrick, JekyllAdmin::StaticServer
      @server.mount "/_api",  Rack::Handler::WEBrick, JekyllAdmin::Server
      Jekyll.logger.info "JekyllAdmin mode:", ENV["RACK_ENV"] || "production"
    end
  end
end
```

kenchan0130:
I am very thankful if anyone can respond.

Thank you.

ashmaroli:
@kenchan0130 This PR attempts to *fix* an issue that occurs from Jekyll 3.7.0.

So irrespective of whether the Jekyll Core Maintainers decides to ship a fix for this via `3.8.1` or `3.7.4` or not all (as the root cause of the issue here is regarding a method that was denoted as `private`), the *proposed* patch should not induce `jekyll-admin` to make changes to its Ruby code as well..

Either Jekyll makes a clean resolution to allow `jekyll-admin` **use its existing monkey-patch** or
Jekyll Admin makes the change to adapt to latest Jekyll code.. not both..

Just my opinion..

/cc @oe

oe:
If it's a problem with an external monkey-patch, I'd rather see that get updated than making some obscure changes here. It's also way more likely that this would land in a Jekyll Admin release than landing here before 4.0.

ashmaroli:
Closing as this should be ideally handled by `jekyll-admin` itself..

PR #6896, opened 2018-04-03
Fixed typo in description of categories and tags

stdll:
I came across this when I read your documentation. I think mentioning the same page twice doesn't make sense, so I assumed you meant the respective pages for the two tags.


DirtyF:
@jekyllbot: :shipit: +docs

PR #6897, opened 2018-04-03
Add missing ul-tag

stdll:
In scenario 5, the result list is surrounded by `<ul>` tags. In scenario 6 they are missing, causing the list to move to the left and causing a rendering issue at https://jekyllrb.com/tutorials/navigation/#scenario-6-applying-the-active-class-for-the-current-page under "Result".


DirtyF:
@jekyllbot: :shipit: +docs

That's two in a row @stdll 👍

PR #6898, opened 2018-04-03
doc: add liquid tag plugin jekyll-onebox for html previews

rriemann:


DirtyF:
@jekyllbot: merge +docs

PR #6905, opened 2018-04-04
WIP: add a faster filter to render smart quotes

ashmaroli:
`Proof-of-concept`

Introduces a filter that uses [RubyPants](https://github.com/jmcnevin/rubypants) to render smart-quotes.
The included benchmark script showed this version to be `~5x` faster on my system

```
            smartify     7.000  i/100ms
         smartypants    38.000  i/100ms
            smartify     75.026  (A± 1.3%) i/s -    378.000  in   5.039730s
         smartypants    378.489  (A± 6.3%) i/s -      1.900k in   5.042268s

Comparison:
         smartypants:      378.5 i/s
            smartify:       75.0 i/s - 5.04x  slower
```

/cc @agriffis @pathawks

### DIFF
--- a/benchmark/smartify-vs-smartypants
+++ b/benchmark/smartify-vs-smartypants
@@ -1,0 +1,37 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require "benchmark/ips"
+require File.join(__dir__, "../lib/jekyll")
+
+class DummyFilter
+  include Jekyll::Filters
+  attr_accessor :site, :context
+
+  def initialize
+    config = Jekyll.configuration("skip_config_files" => true)
+    @site = Jekyll::Site.new(config)
+    @context = Liquid::Context.new(@site.site_payload, {}, { :site => @site })
+  end
+end
+
+SAMPLE_TEXT = %q{'Twas the night before `yesterday', when the 'quick' brown fox jumped over the "lazy" sleeping dog.}
+
+def smartify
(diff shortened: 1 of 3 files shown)


agriffis:
This looks great @ashmaroli 

I'd be interested in two things:

1. Ability to change the options in `_config.yml`

2. Consider a different set of defaults. (This could get into bike shedding, but hopefully useful to consider before merging.) My preferred defaults are: `[1, :named_entities, :prevent_breaks]` (the defaults in rubypants, along with the integer options, are historical at this point)
    * Typographically en dashes aren't as useful as em dashes, they're barely different from hyphens. So it makes sense to render both `--` and `---` as em dashes by default.
    * I don't care much about `:named_entities` versus `:character_entities`, so either way.
    * It's a good thing typographically to `:prevent_breaks`.

ashmaroli:
@agriffis Thank you very much for your suggestions.
I intentionally kept *configurability* out of the loop so that the filter remains being a *slim & faster alternative* for those users who're using the `smartify` filter to convert just "dumb quotes" and "dashes" to typographic alternatives.

Once the Jekyll Core team expresses interest in merging this (esp. since RubyPants does not convert `>`, `<` , `>>`, `<<`), I'will consider fleshing this out to a utilize a *new* `Converter` class under the hood.

The current set of `*options` were chosen since they were *more in sync* with results obtained from the `smartify` filter.

PR #6626, opened 2017-12-12
Upgrade Rubocop to 0.52.1

pathawks:
Fixes #6625

PR #6915, opened 2018-04-08
Bump Rubocop to 0.54.x

### LINKED ISSUE #6655: Update dependency constraint to allow for rubocop v0.52.1
Hey there! :wave:

The constraint I found was `~> 0.51.0`, and the latest version available is `0.52.1`.

crispgm:
- [x] Update Gemfile
- [x] Fix cops

Fix #6655 #6830 #6865 
Closes #6626

### DIFF
--- a/lib/jekyll/drops/site_drop.rb
+++ b/lib/jekyll/drops/site_drop.rb
@@ -26,17 +26,17 @@ module Jekyll
       end
 
       def posts
-        @site_posts ||= @obj.posts.docs.sort { |a, b| b <=> a }
+        @posts ||= @obj.posts.docs.sort { |a, b| b <=> a }
       end
 
       def html_pages
-        @site_html_pages ||= @obj.pages.select do |page|
+        @html_pages ||= @obj.pages.select do |page|
           page.html? || page.url.end_with?("/")
         end
       end
 
       def collections
-        @site_collections ||= @obj.collections.values.sort_by(&:label).map(&:to_liquid)
+        @collections ||= @obj.collections.values.sort_by(&:label).map(&:to_liquid)
       end
 
       # `{{ site.related_posts }}` is how posts can get posts related to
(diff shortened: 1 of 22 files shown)


DirtyF:
@crispgm Do you intend to fix remaining offenses?

crispgm:
@DirtyF I may have a try. And I wonder why there are even more issues using 0.51.0.

crispgm:
@DirtyF there are a lot of empty comments in the codebase. Should remove them or add some simple comment?

ashmaroli:
> there are a lot of empty comments in the codebase. Should I disable it?

disable the cop only for `/features/`

crispgm:
@DirtyF Finally, it's happy again! And it's time to make maintainers happy. :)

lib/jekyll/drops/site_drop.rb
```diff
@@ -25,6 +25,7 @@ def key?(key)
         (@obj.collections.key?(key) && key != "posts") || super
       end
 
+      # rubocop:disable Naming/MemoizedInstanceVariableName
```
ashmaroli: Should we just disable this cop globally?

test/test_site.rb
```diff
@@ -213,6 +213,7 @@ def generate(site)
       @site.process
       # exclude files in symlinked directories here and insert them in the
       # following step when not on Windows.
+      # rubocop:disable Lint/NestedPercentLiteral
```
ashmaroli: IMO this comment ought to be placed at the top level (just before `should "sort pages alphabetically" do`) and the enabler-comment at the corresponding end..

crispgm: Are there any best practice on this? Here I disable the minimum scope for the cops.

DirtyF: Ideally we don't add any rubocop comments and lists files to exclude for now in the rubocop config file and anyone can pick a file and fix it like we did before.

crispgm: @Di r

DirtyF:
@jekyllbot: :shipit: +dev

PR #6916, opened 2018-04-09
Improve comment included in the starter index.md

ashmaroli:
A bit verbose but I guess it gets the message across with clarity.

### DIFF
--- a/lib/site_template/index.md
+++ b/lib/site_template/index.md
@@ -1,6 +1,15 @@
 ---
-# You don't need to edit this file, it's empty on purpose.
-# Edit theme's home layout instead if you wanna make some changes
-# See: https://jekyllrb.com/docs/themes/#overriding-theme-defaults
+# This file is empty because by default, the "content" is adequately realised
+# by Minima's 'home' layout.
+#
+# Nevertheless, feel free to customize this page as you like. Anything you
+# add after the front matter dashes will be rendered and written into the
+# generated "index.html". You may also alter this front matter block itself
+# if you'd like to, Jekyll won't mind.
+#
+# Checkout the following URL for information on customizing layouts and other
+# files bundled inside a theme-gem:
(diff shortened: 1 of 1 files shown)


lib/site_template/index.md
```diff
@@ -1,6 +1,15 @@
 ---
-# You don't need to edit this file, it's empty on purpose.
-# Edit theme's home layout instead if you wanna make some changes
-# See: https://jekyllrb.com/docs/themes/#overriding-theme-defaults
+# This file is empty because by default, the "content" is adequately realised
```
oe: I feel like this line is just going to confuse newcomers even more. What about replacing this with the original `You don't need to edit this file, it's empty on purpose` or some variation of it? If a user wants more information, they can look at the URL provided later on.

lib/site_template/index.md
```diff
@@ -1,6 +1,15 @@
...
+# by Minima's 'home' layout.
+#
+# Nevertheless, feel free to customize this page as you like. Anything you
+# add after the front matter dashes will be rendered and written into the
+# generated "index.html". You may also alter this front matter block itself
+# if you'd like to, Jekyll won't mind.
+#
+# Checkout the following URL for information on customizing layouts and other
```
oe: should be "Check out"

lib/site_template/index.md
```diff
@@ -1,6 +1,15 @@
...
+# files bundled inside a theme-gem:
```
oe: "theme-gem" shouldn't be separated by dashes

lib/site_template/index.md
ashmaroli: I specifically want to do away with `You don't need to edit this file, it's empty on purpose`. It sounds like we're imposing something on the user..

ashmaroli: I like zis.. :+1:

PR #6917, opened 2018-04-09
Regenerate unconditionally unless its an incremental build

ashmaroli:
When its not an incremental build, exit early instead of walking the `case` tree and arriving at the same conclusion..


oe:
@jekyllbot: merge +dev

PR #6921, opened 2018-04-13
Repair the inexact content

sctop:
Actually, when you commit to GitHub, the "jekyll" command will Recompiling on your site. But there not here. So I add.

### DIFF
--- a/docs/_docs/variables.md
+++ b/docs/_docs/variables.md
@@ -93,11 +93,18 @@ following is a reference of the available data.
     </tr>
     <tr>
       <td><p><code>site.pages</code></p></td>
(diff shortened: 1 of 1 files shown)


DirtyF:
I don't know, what is the problem here, but this has nothing to do in this section.

Also GitHub Pages has a proper documentation
https://help.github.com/articles/about-github-pages-and-jekyll/#jekylls-build-process

sctop:
What I want to say is that when you submit it up, GitHub will still compile once (even if you have compiled it once).
Some users may have seen this phenomenon, but novices don't know.
So I want to add it.

PR #6926, opened 2018-04-14
Update Windows install docs

larskanis:
This updates install instructions for jekyll on Windows. I split it into 3 commits to make the changes more comparable. Also see the commit messages for detailed change description.

We're using jekyll and github pages for hosting https://rubyinstaller.org , so that we continuously verify that both run well together. Now that RubyInstaller [comes with builtin Devkit](https://rubyinstaller.org/2018/03/29/rubyinstaller-2.5.1-1-and-2.4.4-1-released.html), I beleave it's significant easier to setup than WSL. This is why I moved the RubyInstaller way to the front in the last commit.


larskanis:
@DirtyF The Chocolatey install description is currently outdated. It's for the old RubyInstaller until 2.3.

Chocolatey was helpful in the past to install nokogiri, because it provides a Chocolatey libxml package, which was necessary for older nokogiri versions. However things have changed, so that nokogiri ships with up-to date binary extensions and it builds fine with it's bundled libxml sources on Windows (since years). Additionally with the introduction of MSYS2 into RubyInstaller-2.4 and newer, `pacman` can be used to install libxml2 and libxslt as described [here](https://github.com/oneclick/rubyinstaller2/tree/rubyinstaller-2.5.1-1#install-gems-with-c-extensions-and-additional-library-dependencies).

Chocolatey can still be used with Ruby 2.4 and newer. It is described [in the RubyInstaller Wiki](https://github.com/oneclick/rubyinstaller2/wiki/FAQ#user-content-choco-install) and in the description of the [Chocolatey packages for Ruby](https://chocolatey.org/packages/ruby) and [for MSYS2](https://chocolatey.org/packages/msys2). However I don't think it has much value in the context of jekyll nowadays. I also found it more fragile than just using the RubyInstaller.

DirtyF:
@larskanis Thanks for the explanation.

Would be nice to have two approvals from @jekyll/windows people here.

docs/_docs/windows.md
```diff
@@ -7,142 +7,26 @@ While Windows is not an officially-supported platform, it can be used to run Jek
...
-   --with-xml2-lib=C:\Chocolatey\lib\libxml2.redist.2.7.8.7\build\native\bin\v110\x64\Release\dynamic\cdecl^
-   --with-iconv-include=C:\Chocolatey\lib\libiconv.1.14.0.11\build\native\include^
-   --with-iconv-lib=C:\Chocolatey\lib\libiconv.redist.1.14.0.11\build\native\bin\v110\x64\Release\dynamic\cdecl^
-   --with-xslt-include=C:\Chocolatey\lib\libxslt.1.1.28.0\build\native\include^
-   --with-xslt-lib=C:\Chocolatey\lib\libxslt.redist.1.1.28.0\build\native\bin\v110\x64\Release\dynamic
-```
-
-#### Install github-pages
+You need RubyInstaller-2.4 or newer with installed Devkit.
+RubyInstaller versions before 2.4 are not compatible to jekyll!
```
ashmaroli: This statement is wrong actually..
Jekyll is fully compatible with Ruby 2.3.x installed via RubyInstaller-2.3.3 and *mingw* DevKit..
([AppVeyor log](https://ci.appveyor.com/project/jekyll/jekyll/build/job/5revhriu0irh38lx#L120) for corroboration.)

For a library (as of `v3.8.0`) that [tests fine even on Ruby 2.2.x](https://ci.appveyor.com/project/jekyll/jekyll/build/job/7i6tgblqvign4igv#L122), it would be incoherent for a Quickstart section to state otherwise..

I'm not against *recommending* Ruby Versions actively supported by `RI-2`, but disagree with stating that Jekyll is incompatible with *any version older than Ruby-2.4*..

docs/_docs/windows.md
```diff
@@ -154,30 +38,18 @@ source 'https://rubygems.org'
 gem 'github-pages', group: :jekyll_plugins
 ```
 
- * **Note:** We use an unsecure connection because SSL throws exceptions in the version of Ruby
  * Open a command prompt, target your local blog repository root, and install github-pages: `bundle install`
 
+After this process you should have github-pages installed on your system and you can host your blog with `jekyll s`.
```
ashmaroli: Can we recommend the "long-form" of the command here as well, please?
```
...and you can host your blog with `jekyll serve` or simply `jekyll s`.
```

docs/_docs/windows.md
larskanis: @ashmaroli My mistake - this statement is wrong. I tried `gem install github-pages` with RubyInstaller-2.3. It installs and seems to work. Still there are the SSL issues with those older releases, so I would still recommend RubyInstaller-2.4 and up.

docs/_docs/windows.md
larskanis: Sounds good - I'll change this.

docs/_docs/windows.md
ashmaroli: `github-pages` and `jekyll` are two different programs.
`github-pages` is a `jekyll` wrapper that has a highly constrained [environment](https://pages.github.com/versions/).

`https://jekyllrb.com/` concerns only about `jekyll`..

larskanis:
Due to the comment of @ashmaroli [above](https://github.com/jekyll/jekyll/pull/6926#discussion_r181662096), I removed the section about `github-pages` in the [last commit](https://github.com/jekyll/jekyll/pull/6926/commits/1e00b1a04658756f2800cb0aa899df509a4dc606). The installation procedure is already described in [github-pages.md](http://localhost:4000/docs/github-pages/), and there isn't anything special on Windows any longer.

DirtyF:
@jekyllbot: merge +docs

PR #6927, opened 2018-04-16
WIP: Do not advise users to install Jekyll outside of Bundler

chrisfinazzo:
This probably should have been in #6881. It seems reasonable to me that Jekyll should be under Bundler's purview. This is consistent w/ instructions [here][] and is useful if you don't run `jekyll new` (which calls `bundle install` internally). The wording is a bit repetitive, so that might need to be tweaked.

[here]: https://jekyllrb.com/docs/installation/

### DIFF
--- a/docs/_docs/quickstart.md
+++ b/docs/_docs/quickstart.md
@@ -7,8 +7,11 @@ permalink: /docs/quickstart/
 If you already have a full [Ruby](https://www.ruby-lang.org/en/downloads/) development environment with all headers and [RubyGems](https://rubygems.org/pages/download) installed (see Jekyll's [requirements](/docs/installation/#requirements)), you can create a new Jekyll site by doing the following:
 
 ```sh
-# Install Jekyll and Bundler gems through RubyGems
-gem install jekyll bundler
+# Install Bundler through RubyGems
+gem install bundler
+
+# Add Jekyll to your Gemfile and install its dependencies
+bundle add jekyll && bundle install
 
 # Create a new Jekyll site at ./myblog
 jekyll new myblog
@@ -26,7 +29,7 @@ If you encounter any unexpected errors during the above, please refer to the [tr
 
 ## About Bundler
 
-`gem install jekyll bundler` installs the [jekyll](https://rubygems.org/gems/jekyll/) and [bundler](https://rubygems.org/gems/bundler) gems through [RubyGems](https://rubygems.org/). You need only to install the gems one time &mdash; not every time you create a new Jekyll project. Here are some additional details:
+`gem install bundler` installs the [bundler](https://rubygems.org/gems/bundler) gem through [RubyGems](https://rubygems.org/). You need only to install it once &mdash; not every time you create a new Jekyll project. Here are some additional details:
 
 * `bundler` is a gem that manages other Ruby gems. It makes sure your gems and gem versions are compatible, and that you have all necessary dependencies each gem requires.
 * The `Gemfile` and `Gemfile.lock` files inform Bundler about the gem requirements in your site. If your site doesn't have these Gemfiles, you can omit `bundle exec` and just run `jekyll serve`.
--- a/docs/_docs/quickstart.md
+++ b/docs/_docs/quickstart.md
@@ -29,13 +29,15 @@ If you encounter any unexpected errors during the above, please refer to the [tr
 
 ## About Bundler
 
-`gem install bundler` installs the [bundler](https://rubygems.org/gems/bundler) gem through [RubyGems](https://rubygems.org/). You need only to install it once &mdash; not every time you create a new Jekyll project. Here are some additional details:
+`gem install bundler` installs the [bundler](https://rubygems.org/gems/bundler) gem through [RubyGems](https://rubygems.org/). You only need to install it once &mdash; not every time you create a new Jekyll project. Here are some additional details:
 
 * `bundler` is a gem that manages other Ruby gems. It makes sure your gems and gem versions are compatible, and that you have all necessary dependencies each gem requires.
 * The `Gemfile` and `Gemfile.lock` files inform Bundler about the gem requirements in your site. If your site doesn't have these Gemfiles, you can omit `bundle exec` and just run `jekyll serve`.
 
 * When you run `bundle exec jekyll serve`, Bundler uses the gems and versions as specified in `Gemfile.lock` to ensure your Jekyll site builds with no compatibility or dependency conflicts.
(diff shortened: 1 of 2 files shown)


pathawks:
The purpose of `gem install jekyll` is so users can run `jekyll new MySite` anywhere and get a new site scaffolding.

This _does_ mean that there may be a difference between the system Jekyll and the Bundler Jekyll. If this issue needs fixing, it might require a more wholistic view of how we expect users to install/run Jekyll then this PR currently provides.

chrisfinazzo:
> This does mean that there may be a difference between the system Jekyll and the Bundler Jekyll. If this issue needs fixing, it might require a more wholistic view of how we expect users to install/run Jekyll then this PR currently provides.

I think this is the key here. No question, this is incomplete at the moment. Over time, the trend in the documentation has become more "Bundler-oriented" and it is expected to be part of your gem management (on some level) with regard to Jekyll projects. #4368 was an early attempt at this, but I think things have matured to the point where it warrants consideration again.

Obviously, we shouldn't dictate environments if at all possible, but _not_ doing this at some point just sounds like a compatibility headache that users shouldn't have to worry about. It doesn't solve the entire problem (which itself is nigh-impossible because software is hard) but it does make the management of these things a bit easier by removing some complexity from the equation.

pathawks:
> but I think things have matured to the point where it warrants consideration again.

Agreed.

/cc: @jekyll/documentation

DirtyF:
We should link to https://jekyllrb.com/tutorials/using-jekyll-with-bundler/ in the [About Bundler](https://jekyllrb.com/docs/quickstart/#about-bundler) section, and even merge the whole so that we keep it a quickstart guide.

docs/index.html
```diff
@@ -45,7 +45,12 @@ <h3>Get up and running <em>in&nbsp;seconds</em>.</h3>
         <p class="line">
           <span class="path">~</span>
           <span class="prompt">$</span>
-          <span class="command">gem install jekyll bundler</span>
+          <span class="command">gem install bundler</span>
+        </p>
+        <p class="line">
+          <span class="path">~</span>
+          <span class="prompt">$</span>
+          <span class="command">bundle add jekyll && bundle install</span>
```
crispgm: Should we use `&amp;` instead here?

docs/_docs/quickstart.md
```diff
@@ -7,8 +7,11 @@ permalink: /docs/quickstart/
 If you already have a full [Ruby](https://www.ruby-lang.org/en/downloads/) development environment with all headers and [RubyGems](https://rubygems.org/pages/download) installed (see Jekyll's [requirements](/docs/installation/#requirements)), you can create a new Jekyll site by doing the following:
 
 ```sh
-# Install Jekyll and Bundler gems through RubyGems
-gem install jekyll bundler
+# Install Bundler through RubyGems
+gem install bundler
+
+# Add Jekyll to your Gemfile and install its dependencies
+bundle add jekyll && bundle install
```
ashmaroli: One would need to run `bundle init` or be certain they have a valid `Gemfile` before they can run `bundle add`

docs/_docs/quickstart.md
```diff
@@ -26,7 +29,7 @@ If you encounter any unexpected errors during the above, please refer to the [tr
 
 ## About Bundler
 
-`gem install jekyll bundler` installs the [jekyll](https://rubygems.org/gems/jekyll/) and [bundler](https://rubygems.org/gems/bundler) gems through [RubyGems](https://rubygems.org/). You need only to install the gems one time &mdash; not every time you create a new Jekyll project. Here are some additional details:
+`gem install bundler` installs the [bundler](https://rubygems.org/gems/bundler) gem through [RubyGems](https://rubygems.org/). You need only to install it once &mdash; not every time you create a new Jekyll project. Here are some additional details:
```
ashmaroli: You may consider rewording this sentence to:
```diff
- You need only to install it once &mdash; not every time you create a new Jekyll project.
+ You only need to install it once &mdash; not every time you create a new Jekyll project.
```

docs/_docs/quickstart.md
pathawks: I’m not sure where this Gemfile would exist. Some directory that is a parent of _all_ of a users Jekyll sites?

ashmaroli: This is why *the docs currently recommend* the `jekyll new` approach after `gem install jekyll`. Its a lot simpler..

docs/index.html
chrisfinazzo: @crispgm, I used it because it is the convention in Bash to indicate “Run this command, now run that command” after the first one completes. I don’t know enough with regard to character encodings to say if this would cause Unicode issues and not display correctly in some browsers.

docs/_docs/quickstart.md
chrisfinazzo: Ok, will update. Might be a little while as I just finished restoring from a Time Machine backup.

Tl;dr, 10.13.4 "delta" Installer is 💩 .

smash-computer

docs/index.html
chrisfinazzo: Tried it out just to be sure, looks the same, so I'm 👍🏼 on this.

docs/_docs/quickstart.md
chrisfinazzo: I considered whether or not to expicitly mention `bundle init` here, as it is already described in some detail on the tutorial that we will be linking. Is that just repeating ourselves for no reason?

screen shot 2018-04-16 at 3 40 17 pm

ashmaroli: `bundle add jekyll` is detailed in that tutorial as well.. So, your concern is actually moot since `bundle init` is a necessary step for `bundle add jekyll` to run successfully..

DirtyF:
@jekyllbot: merge +docs

docs/index.html
```diff
@@ -45,7 +45,12 @@ <h3>Get up and running <em>in&nbsp;seconds</em>.</h3>
         <p class="line">
           <span class="path">~</span>
           <span class="prompt">$</span>
-          <span class="command">gem install jekyll bundler</span>
+          <span class="command">gem install bundler</span>
+        </p>
+        <p class="line">
+          <span class="path">~</span>
+          <span class="prompt">$</span>
+          <span class="command">bundle add jekyll &amp; &amp; bundle install</span>
```
chrisfinazzo: I'm...not that clever. Sorry about this - as written, it's a syntax error. Fix is in #6969.

PR #6936, opened 2018-04-18
Improve regex for matching relative_path w/ date

ashmaroli:
PRO:
  -  `~1.60x faster` when matching *valid relative paths*
  - upto `1.90x` faster when matching *invalid relative paths with two categories (`foo/bar/_posts/*.md`)*

~~CON: `~1.40x slower` when matching *invalid relative paths*~~

(Benchmark script attached with PR)

### DIFF
--- a/benchmark/posts-filepath-regex
+++ b/benchmark/posts-filepath-regex
@@ -15,9 +15,9 @@ DATED_FILES = %w(
 POST_FILENAMES = DATED_FILES.concat(
   DATED_FILES.map do |filename|
     [
-      "_posts/" + filename,
-      "_posts/category_a/" + filename,
-      "_posts/category_a/category_b/" + filename
+      '_posts/' + filename,
+      'category_a/_posts/' + filename,
+      'category_a/category_b/_posts/' + filename
     ]
   end
 ).flatten.freeze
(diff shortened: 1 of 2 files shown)


ashmaroli:
> Could you please describe what this change is doing?

I added a couple of *lazy-tokens* to reduce the backtracks with a successful match.
`^(?:.+/)*` on `master` would eagerly match the full `relative_path` and then backtrack as required to match successfully consequent position. This can be a drain when there are numerous posts with really long filenames.
`^(?:.+?/)*?` in the proposal only consumes part of the `relative_path` (approx. before the `yyyy-mm-dd-`) as required, to match successfully at the consequent position..

The included benchmark validates the above hypothesis.

pathawks:
That's cool :+1:

It would be nice to test this on a number of real world sites, just to make sure that the _valid_ case really is more common than the _invalid_ case (I am assuming it would be)

ashmaroli:
Probably not worth pursuing this..

pathawks:
If this is faster in all cases it seems like an easy win. I just want to make sure a developer going through the code can mentally parse the RegExp.

ashmaroli:
The gain from this change is probably insignificant for the users..

PR #6938, opened 2018-04-19
Generate posts with `.md` extension

ashmaroli:
..and promote the shorter extension because:
- our `Document::DATE_FILENAME_MATCHER` parses the shorter extension [faster](https://github.com/jekyll/jekyll/pull/6936#issuecomment-382605969) (~1.05x)
- shorter extension is easier to type (when manually creating posts via File Explorer..) :stuck_out_tongue:


ashmaroli:
> I think we're still missing the big picture here...

Agreed. I'm so lost at the low-level, I barely see the *big picture* anymore.. Would definitely be helpful if *someone* can leave pointers to the *big picture..*

mattr-:
I'm going to close this. `.markdown` is more explicit and while `.md` is shorter and everything, the performance improvement is such a micro-optimization that I don't think it's important to focus on.

We're at a point with our performance optimization work that we're going to need to figure out how we fundamentally restructure jekyll to be more performant, rather than trying to squeeze extra performance out of an architecture that was not designed with performance in mind.

PR #6951, opened 2018-04-23
Memoize computing excerpt's relative_path

ashmaroli:
because `relative_path` is called multiple times (at least *3 times per post)*..

(This *definition* was introduced in `3.8.0`)


DirtyF:
@jekyllbot: merge +minor

PR #6956, opened 2018-04-25
Add version number for group_by_exp doc

karel-3d:


docs/_docs/templates.md
```diff
@@ -178,7 +178,7 @@ you come up with your own tags via plugins.
     <tr>
       <td>
         <p class="name"><strong>Group By Expression</strong></p>
-        <p>Group an array's items using a Liquid expression.</p>
+        <p>Group an array's items using a Liquid expression. Jekyll v3.4.0 & later.</p>
```
DirtyF: @karel-3d we have an include for that:
`{% include docs_version_badge.html version="3.4.0" %}`

karel-3d: I copy-pasted this from few lines earlier, where_exp. Should I change both?

DirtyF: yes, please.

docs/_docs/templates.md
```diff
@@ -178,7 +179,8 @@ you come up with your own tags via plugins.
     <tr>
       <td>
         <p class="name"><strong>Group By Expression</strong></p>
-        <p>Group an array's items using a Liquid expression.</p>
+        <p>Group an array's items using a Liquid expression.
+        {% include docs_version_badge.html version="3.4.0" %}</p>
```
pathawks: Note: We don't want to use `{%- -%}` syntax here, because we want a space after the period before the badge.

DirtyF:
@jekyllbot: merge +docs

karel-3d:
Thanks!

PR #6959, opened 2018-04-26
Liquify documents unless published == false

### LINKED ISSUE #6958: Published: false no longer working
Since updating to 3.8 we have noticed pages/items in collections are being outputted even if they have `published: false` in the front matter.

E.g. we have a collection "partners" and there are 9 items in the collection, 3 of which were set to published false due to us still finalising things.

This was fine until we updated but now all 9 items are shown in the collection instead of the 6 that were set to published: true.

Is this a bug or has published been deprecated? If so, what to?

Thanks

ashmaroli:
Resolves #6958 

- `data["published"] == false` decides if the concerned `doc` should be accessible via Liquid or if it should be processed and ultimately written to disk..
- future-dated documents in both *visible* (`output: true`) and *hidden* (`output: false`) collections **continue** to be accessible via Liquid, but will be written only if `site.future == true` and document is part of a *visible* collection.

### DIFF
--- a/lib/jekyll/collection.rb
+++ b/lib/jekyll/collection.rb
@@ -210,12 +210,9 @@ module Jekyll
     private
 
     def read_document(full_path)
-      docs << Document.new(full_path, :site => site, :collection => self).tap do |doc|
-        doc.read
-        if !site.publisher.publish?(doc) && site.publisher.hidden_in_the_future?(doc)
-          Jekyll.logger.debug "Skip Publishing:", "#{doc.relative_path} has a future date"
-        end
-      end
+      doc = Document.new(full_path, :site => site, :collection => self)
+      doc.read
+      docs << doc unless doc.data["published"] == false
     end
 
     private
--- a/features/collections.feature
+++ b/features/collections.feature
@@ -140,6 +140,44 @@ Feature: Collections
     And I should see "Newest puppy: Fido" in "_site/index.html"
     And the "_site/puppies/fido.html" file should exist
 
+  Scenario: Access rendered and published collection documents via Liquid
+    Given I have a _puppies directory
+    And I have the following documents under the puppies collection:
+      | title  | date       | content             | published |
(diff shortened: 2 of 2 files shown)


lib/jekyll/collection.rb
```diff
@@ -210,12 +210,9 @@ def container
...
     def read_document(full_path)
-      docs << Document.new(full_path, :site => site, :collection => self).tap do |doc|
-        doc.read
-        if !site.publisher.publish?(doc) && site.publisher.hidden_in_the_future?(doc)
-          Jekyll.logger.debug "Skip Publishing:", "#{doc.relative_path} has a future date"
-        end
-      end
+      doc = Document.new(full_path, :site => site, :collection => self)
+      doc.read
+      docs << doc unless doc.data["published"] == false
```
parkr: These sorts of concerns should be in the Publisher, don’t you think? It’s intention was to be this sort of filter.

ashmaroli:
> previously, you could reference this document even though it wasn’t output, right? But now you can’t?

You still can reference *non-output documents* as long as there is no explicit `published: false` in the front matter..

lib/jekyll/collection.rb
ashmaroli: I didn't want to introduce a new `public_method` in a patch release..

pathawks:
Did this change with `v3.8.0`? Which commit changed the previous behavior?

Or is this an enhancement?

ashmaroli:
@pathawks Yes, this is a `fix`.. for a bug introduced via #6795

lib/jekyll/collection.rb
mattr-: I don't see a problem with introducing a new public method in a patch release.

ashmaroli:
> ..would like to see that happen at some point later.

Definitely.. it'll be introduced in `4.0`..

DirtyF:
@jekyllbot: merge +fix

pathawks:
@jekyllbot: merge +fix

PR #6967, opened 2018-04-30
Loggers should accept both numbers and symbols

pathawks:
From #6305:

> Jekyll's logger should start trying to conform to standard logging practices in Ruby. That said, the logger level should accept both a :symbol (if you really want,) and a number, this way users aren't taken aback when they know what log level they want. This commit makes it so that the logger level can be set normally and within Jekyll's own constraints.

```ruby
Jekyll.logger.log_level = 0
Jekyll.logger.log_level = Logger::DEBUG
Jekyll.logger.log_level = :debug
```


ashmaroli:
@pathawks How does this benefit our regular users or plugin developers..?

pathawks:
@ashmaroli I think the benefit here is that it makes our logger behave more like standard Ruby loggers.

ashmaroli:
okay.. will review after the tests pass..

pathawks:
@ashmaroli Travis is now ✅

ashmaroli:
*(explicitly requesting a review from @parkr (instead of a team) because something about this doesn't feel right to me..)*

DirtyF:
@jekyllbot: merge +dev

Issue #6248, opened 2017-07-26
Feature proposal: report syntax errors in the web page


osteele:
Problem: Frequently while running `jekyll serve --watch` and editing a file, I'll introduce a syntax error and not notice, because I'm just looking back and forth between the editor and the browser, and don't have the terminal in my field of view.

One solution, that requires modifying user behavior, is for the user to have all three panes up and scan them all.

Another is to implement a “syntax error overlay”. The [create-react-app](https://github.com/facebookincubator/create-react-app) project reports syntax errors in the web page, as well as on the console. [This PR](https://github.com/facebookincubator/create-react-app/pull/744) begins with a screenshot that illustrates this.

This would be a nice feature for Jekyll too. It could look something like this:

<img width="843" alt="screenshot of safari 7-26-17 6-55-08 pm" src="https://user-images.githubusercontent.com/674/28647464-5041acea-7235-11e7-9a63-b8e715432494.png">

(This particular screenshot is from a Jekyll clone. That's why the error message is different from what Jekyll would report. I've found it helpful there. If anyone wants to take this on, you are of course welcome to use [the clone's CSS and Liquid template](https://github.com/osteele/gojekyll/blob/master/server/server.go).)

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>3.3-stable</code> or <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, please consider building it first as a plugin. Jekyll 3 introduced [hooks](http://jekyllrb.com/docs/plugins/#hooks) which provide convenient access points throughout the Jekyll build pipeline whereby most needs can be fulfilled. If this is something that cannot be built as a plugin, then please provide more information about why in order to keep this issue open.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

pathawks:
This would be pretty cool, so long as it only happened in the dev environment. :+1:

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>3.3-stable</code> or <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

[...]

PR #6974, opened 2018-05-01
WIP: report syntax errors in the web page

nickskalkin:
Hi! Here is my implementation of this feature #6248 

It works for me, but I think that implementation may not be very good. Can you give me some advices?
Also I haven't written any tests and docs, as I think I will be asked to rewrite my solution...

### DIFF
--- a/test/test_live_reload_template.rb
+++ b/test/test_live_reload_template.rb
@@ -5,7 +5,7 @@ require "helper"
 class TestLiveReloadTemplate < JekyllUnitTest
   context "rendering livereload script with different options" do
     context "with port only" do
-      subject { LiveReloadTemplate.new("livereload_port" => "35729") }
+      subject { LiveReloadTemplate.new("livereload_port" => 35_729) }
 
       should "return correct script" do
         assert_equal subject.template, <<-TEMPLATE
--- a/test/test_live_reload_template.rb
+++ b/test/test_live_reload_template.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+require "helper"
+
+class TestLiveReloadTemplate < JekyllUnitTest
+  context "rendering livereload script with different options" do
+    context "with port only" do
+      subject { LiveReloadTemplate.new("livereload_port" => "35729") }
+
+      should "return correct script" do
+        assert_equal subject.template, <<-TEMPLATE
+<script>
+  document.write(
+    '<script src=\"http://' +
+    (location.host || 'localhost').split(':')[0] +
+    ':35729/livereload.js?snipver=1&amp;port=35729\"' +
+    '></' +
+    'script>');
+</script>
+        TEMPLATE
+      end
+    end
+
+    context "with all possible options" do
+      subject do
+        LiveReloadTemplate.new(
+          "livereload_port"      => "35729",
+          "livereload_min_delay" => "500",
+          "livereload_max_delay" => "1000"
+        )
+      end
+
+      should "return correct script" do
+        assert_equal subject.template, <<-TEMPLATE
+<script>
+  document.write(
+    '<script src=\"http://' +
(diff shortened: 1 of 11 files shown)


DirtyF:
Thanks for giving a shot at this ❤️ 

My 2 cents on tests and documentation: Writing tests is the best way to make sure your code does what he's supposed to do. You won't need to change them even if there's need for refactoring during the review, it will even be easier for you to refactor with the help of the tests. The same goes for documentation. 😄

DirtyF:
I can't test that locally with jekyll's website: 

```

**Comment by nickskalkin** [comment]
@ashmaroli I've tested only Liquid errors. For example I've changed `{% include ... %}` to `{% failed_include .... %}`

Here is an example how it looks for Jekyll site:

screen shot 2018-05-01 at 22 45 34

Some notes:
I haven't run jekyll site with `rake` task, only by `bundle exec jekyll --livereload`. Mention `livereload`, as my code work ONLY when livereload is `true`.

**Comment by ashmaroli** [comment]
@nickskalkin I tested this out locally using Jekyll's `docs` folder by adding `{% foobar %}` to `_docs/code_of_conduct.md` and `_tutorials/orderofinterpretation.md`, and using `master` versions of `jekyll-mentions` and `jemoji` plugins.
 Here are my findings: *(all commands are from within the `docs` directory and prepended with `bundle exec`)*
  - When I run `jekyll build`, the build aborts with an exception from our `jekyll-mentions` plugin on `_docs/code_of_conduct.md`
  - When I run `jekyll serve`, the build aborts as above.
  - When I run `jekyll serve --livereload`, the following can be observed:
    - When I visit `/tutorials/orderofinterpretation/`, the error template is rendered &mdash; **your implementation works!**
    - When I visit `/docs/conduct/`, the error template is **not rendered** &mdash; *your implementation got overridden by Jekyll's `:post_render` hook called by `jekyll-mentions` plugin*

So, for an initial draft :+1:  :clap:
Good start!!

**Comment by ashmaroli** [comment]
And regarding the code, I recommend *memoizing* the methods that are called multiple times but do not change for given build session, especially `LiveReloadTemplate#template` and `ErrorPrinter#error_styles` so that they do not get executed 1000 times in a large site when a *layout for 1000 posts* contains a Liquid Syntax error

**Comment by nickskalkin** [comment]
@ashmaroli thanks for the feedback! I'll improve my code (about memoizing, docs and tests). But maybe I didn't get you right, is current behaviour OK? As far as I understand I need to fix this:

> When I visit /docs/conduct/, the error template is not rendered — your implementation got overridden by Jekyll's :post_render hook called by jekyll-mentions plugin

**Comment by ashmaroli** [comment]
> As far as I understand I need to fix this:

Yes, that's correct.

**Comment by ashmaroli** [comment]
One other thing you **have to consider:** 
*If the Liquid Syntax Error occurs in a layout or an include, the rendered error is misleading..*

I don't know if you can overcome that hurdle...

**Comment by pathawks** [comment]
@nickskalkin Please do _not_ spend any time memorizing variables or performing premature optimizations.

Write readable code, and if we find a bottleneck we will deal with it on a case-by-case basis 👍🏼

**Comment by nickskalkin** [comment]
There are some tests and fix of post render hook

File: test/test_live_reload_template.rb [hunk 1]
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+require "helper"
+
+class TestLiveReloadTemplate < JekyllUnitTest
+  context "rendering livereload script with different options" do
+    context "with port only" do
+      subject { LiveReloadTemplate.new("livereload_port" => "35729") }
```
ashmaroli: should this be `35_729` like its in other tests..?

nickskalkin: OK, I'll change!

PR #6976, opened 2018-05-02
yajl-ruby update to v1.4.0

### LINKED ISSUE #6965: Update dependency constraint to allow for yajl-ruby v1.4.0
Hey there! :wave:

I noticed that the constraint you have for yajl-ruby doesn't allow for the latest version to be used.

The constraint I found was `~> 1.3.1`, and the latest version available is `1.4.0`.

nickskalkin:
Resolves #6965

Tested on High Sierra, Ruby 2.5.1, all works fine

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -80,7 +80,7 @@ group :jekyll_optional_dependencies do
     gem "pygments.rb", "~> 1.0"
     gem "rdiscount", "~> 2.0"
     gem "redcarpet", "~> 3.2", ">= 3.2.3"
-    gem "yajl-ruby", "~> 1.4.0"
+    gem "yajl-ruby", "~> 1.3"
   end
 
   # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
--- a/Gemfile
+++ b/Gemfile
@@ -80,7 +80,7 @@ group :jekyll_optional_dependencies do
     gem "pygments.rb", "~> 1.0"
     gem "rdiscount", "~> 2.0"
     gem "redcarpet", "~> 3.2", ">= 3.2.3"
-    gem "yajl-ruby", "~> 1.3.1"
+    gem "yajl-ruby", "~> 1.4.0"
   end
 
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -80,7 +80,7 @@ group :jekyll_optional_dependencies do
     gem "pygments.rb", "~> 1.0"
     gem "rdiscount", "~> 2.0"
     gem "redcarpet", "~> 3.2", ">= 3.2.3"
-    gem "yajl-ruby", "~> 1.3.1"
+    gem "yajl-ruby", "~> 1.4.0"
```
pathawks: Any reason not to use `"~> 1.3"` ?

ashmaroli:
Please edit `Issue #6965` in the PR body to `Resolves #6965`

Gemfile
nickskalkin: Just updated as mentioned in #6965

nickskalkin:
@ashmaroli done

Gemfile
ashmaroli: Using `"~> 1.3"` is better..

pathawks:
@nickskalkin Thanks for opening this PR :beers:

I think, so long as `yajl-ruby` follows semantic versioning, we should change the constraint to `"~> 1.3"`
This will allow any version of `yajl-ruby` from `1.3.0` up to (but not including) `2.0.0`. This means that our users will be able to take advantage of `1.5.0` as soon as it is released, instead of waiting for another PR like this one.

Gemfile
nickskalkin: Ah, got it, sorry :) Read '1.3.1' instead of '1.3'

pathawks:
If you're interested, you can read all about RubyGems version specifiers here: http://guides.rubygems.org/patterns/#pessimistic-version-constraint

pathawks:
Thanks for the PR :tada:

@jekyllbot: merge +dev

PR #6977, opened 2018-05-02
We are not using Ruby 2.2 anymore

pathawks:


DirtyF:
@jekyllbot: merge +major

PR #6978, opened 2018-05-02
Drop support for older versions of Rouge

ashmaroli:
Its just Rouge 3.x and above from here onwards


DirtyF:
@jekyllbot: merge +major

PR #6980, opened 2018-05-03
Ignore posts and drafts without front matter

### LINKED ISSUE #6688: Inconsistent behavior with Front Matter in Posts and other collections
- [x] This is a bug in Jekyll
Jekyll Version: 3.7.0 on Windows

### To Reproduce:
  1. Create a fresh Jekyll workspace
  2. Open boilerplate post and insert a newline at the very beginning:
```
\n
---
title: Welcome To Jekyll
---
```
  3. Configure a new collection and add the new line to a document in this as well
```
\n
---
title: Collections and documents
---
```
  4. Build site
  5. Inspect contents in `_site`

### Expectations:
I expected to see the generated files for both the boilerplate post document and the newly added collection document.

### What I got instead:
  - The post got processed into the HTML file as expected.
  - The collection document got returned as if it were a static file.

### Inference:
  - Clearly, an inconsistent behavior across collections..

### Other observations:
  - **Posts without a Front Matter block get processed into destination.**
  - Regular pages do not get processed if the file does not begin with the Front Matter dashes.

ashmaroli:
Fix #6688 

If a "post" or "draft" contains YAML front matter, create the `Jekyll::Document` instance otherwise its a `StaticFile`


DirtyF:
What about compatibility with https://github.com/benbalter/jekyll-optional-front-matter?

ashmaroli:
I don't see why it should be a problem since the plugin's _raison d'être_ is to _nullify_ this core behavior. Though granted that the plugin may have to be _enhanced_ to cover posts and drafts once this is merged

mattr-:
Are we going to move away from the notion that things without front matter default to pages or static files?

ashmaroli:
> Are we going to move away from the notion that things without front matter default to pages or static files?

Nope. We're going to *cement* the notion that Jekyll only processes files with YAML Front Matter.

mattr-:
Could you define `processes`? Because in my head, moving static files over to the output directory is Jekyll processing them. 😄

mattr-:
My point being, we need to be ultra clear about the language here so we don't indicate the wrong thing to our users.

DirtyF:
What about markdown files without front matter? I would expect those would be converted into HTML, maybe we have defaults frontmatter in our config and that's enough.

Can we please state what we're trying to achieve and why from a user's perspective?

DirtyF:
```Given I have 1000 markdown files
When I run Jekyll build
Then All my files are copied as static files and not rendered

PR #6984, opened 2018-05-04
Handle static files in collection with a subclass

ashmaroli:
### Background:
`Jekyll::StaticFile` currently contains numerous methods that vary based on whether the class was initialized with a `Collection` object passed as a parameter leading to complicated code (code smell)

This PR attempts to resolve the code smell by extracting the logic to handle static files in collections with a dedicated class inherited from `Jekyll::StaticFile`

### Outcome:
This implementation changes the public API of the `StaticFile` class in addition to cleaning it up. The changes are:
  - `StaticFile.new` now takes *only* 4 parameters
  - `StaticFile#placeholders` is no longer necessary and has been removed
  - `StaticFile#type` always returns `nil`

The new class `Jekyll::CollectionStatic` overrides the relevant methods from its superclass and exhibits behavior *similar* to `Jekyll::Document`

### DIFF
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -73,6 +73,7 @@ module Jekyll
   autoload :LiquidRenderer,      "jekyll/liquid_renderer"
   autoload :Site,                "jekyll/site"
   autoload :StaticFile,          "jekyll/static_file"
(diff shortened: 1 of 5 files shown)


ashmaroli:
No longer interested in pursuing this..

PR #6987, opened 2018-05-04
Removed Redcarpet

### LINKED ISSUE #6981: Remove support for Redcarpet
Redcarpet isn't maintained very well at this point and it feels like we should do one of the following:
- Move support for Redcarpet to a plugin
- Remove it entirely

I vote for the second option. It doesn't make sense for us to create one more plugin for ourselves to maintain.

nickskalkin:
Resolves #6981


docs/_docs/configuration.md
```diff
@@ -728,50 +725,6 @@ mutually exclusive.
 The various Markdown renderers supported by Jekyll sometimes have extra options
 available.
 
-### Redcarpet
```
crispgm: Should we leave a message to inform the deprecation here?

nickskalkin: I don't know if it makes sense to leave a deprecation message when support of Redcarpet will be removed (in 4.0). Maybe I should do another PR with deprecation warnings? And merge this PR to the master before 4.0?

ashmaroli: @nickskalkin Please don't bother adding deprecation messages to our *documentation*. Chances are that those messages will never be read..

DirtyF: Changes to the documentation shouldn't be live before v4.0. I'd advise to commit those changes in a separate commit in the `docs-40` branch

DirtyF:
@nickskalkin Still need to revert changes to the docs in this PR before merging 👍

nickskalkin:
@DirtyF done!

PR #6988, opened 2018-05-04
Remove support for rdiscount

jekyllbot:
Fix #6988 

- [x] update Documentation (in https://github.com/jekyll/jekyll/commit/1d8d6be243a4a6697c70c47f2ede47458d88bcc7)


ashmaroli:
@DirtyF Conflicts ahoy! :wink:

lib/jekyll/converters/markdown.rb
```diff
@@ -43,7 +42,7 @@ def get_processor
       # are not in safe mode.)
 
       def valid_processors
-        %w(rdiscount kramdown) + third_party_processors
+        %w(kramdown) + third_party_processors
```
ashmaroli: lets use `#concat` or `#unshift` here to avoid creating another Array for no reason..
```ruby
%w(kramdown).concat(third_party_processors)
```

lib/jekyll/converters/markdown.rb
```diff
@@ -52,7 +51,7 @@ def valid_processors
 
       def third_party_processors
         self.class.constants - \
-          %w(KramdownParser RDiscountParser PRIORITIES).map(
+          %w(KramdownParser PRIORITIES).map(
             &:to_sym
           )
```
ashmaroli: This can now be condensed into a single line now IMO.. and use `#map!`
```ruby
%w(KramdownParser PRIORITIES).map!(&:to_sym)
```

DirtyF:
That leaves us with a single Markdown engine by default, we might want to support Commonmark out of the box later.

ashmaroli:
Removed my *requests for refactoring the methods* (and updated the branch accordingly).
The refactoring will be for a separate PR

ashmaroli:
> That leaves us with a single Markdown engine by default, we might want to support Commonmark out of the box later.

We can either bundle [`jekyll-commonmark-ghpages`](https://github.com/github/jekyll-commonmark-ghpages) or *its dependency* `jekyll-commonmark` like how we bundle `jekyll-sass-converter`

DirtyF:
@jekyllbot: merge +major

PR #6990, opened 2018-05-05
Remove documentation for using Redcarpet

nickskalkin:
There are only docs changes.

### DIFF
--- a/docs/_docs/configuration.md
+++ b/docs/_docs/configuration.md
@@ -690,9 +690,6 @@ liquid:
 
 # Markdown Processors
 
-redcarpet:
-  extensions: []
-
 kramdown:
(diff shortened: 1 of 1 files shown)


nickskalkin:
Or I should leave a deprecation warning instead of deleting Redcarpet section entirely?

ashmaroli:
> leave a deprecation warning instead

Deprecation means to discourage the use of a feature that is being phased out..
This contents of this PR will be public only at the time `v4.0` is being shipped (by when the feature has already been removed).. There's no point in sending out a warning at that time, is there? :slightly_smiling_face:

DirtyF:
@jekyllbot: merge +docs

Issue #6947, opened 2018-04-22
Extra forward slash in URLs when config["url"] has a trailing forward slash


inconvergent:
- [?] I believe this to be a bug, not a question about using Jekyll.
- [x] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [ ] I ran `jekyll doctor` to check my configuration
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [ ] This is a feature request.

---

- [ ] I am on (or have tested on) ***macOS*** 10+
- [x] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux
- [ ] I am on (or have tested on) ***Fedora*** GNU/Linux
- [ ] I am on (or have tested on) ***Arch*** GNU/Linux
- [ ] I am on (or have tested on) ***Other*** GNU/Linux
- [ ] I am on (or have tested on) ***Windows*** 10+

---

- [ ] I was trying to install.
- [ ] There is a broken Plugin API.
- [ ] I had an error on GitHub Pages, and I have reproduced it locally.
- [ ] I had an error on GitHub Pages, and GitHub Support said it was a Jekyll Bug.
- [ ] I had an error on GitHub Pages and I did not test it locally.
- [ ] I was trying to build.
- [x] It was another bug.

## My Reproduction Steps

I am opening this at the request of @pathawks (https://github.com/jekyll/jekyll-redirect-from/issues/178#issuecomment-383382550). It was initially posted here 
https://github.com/jekyll/jekyll-redirect-from/issues/178#issuecomment-383383152

The issue relates to redirects (using the redirect-from plugin) on github pages vs. my local setup.
 
I've tested this on the same version of jekyll (3.7.3) and jekyll-redirect-from (0.13.0) as what is listed as current on github: https://pages.github.com/versions/

When using the redirect plugin i get `//` between the domain name and the path. 

When I'm using redirect_from on my jekyll page it works as expected locally. but on on the github page there is an extra slash. 

### As an example.

i redirect like this: `redirect_from: /summary/2017/`

I don't know it is relevant, but my baseurl is "" (empty string). 

When i curl the two different pages i get the following:

```
  curl -v  'http://localhost:4000/summary/2017/' 
*   Trying ::1...
* connect to ::1 port 4000 failed: Connection refused
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 4000 (#0)
[...]

PR #6991, opened 2018-05-06
WIP: Sanitize url and baseurl values on loading config

ashmaroli:
Benefits Jekyll's URLFilters &mdash; Jekyll doesn't have to sanitize the values on each call to the filter(s).

### Summary:
  - `config["url"]` and `config["baseurl"]` will be *reset* to `""` if the set value is `nil`, or an instance other than that of `String`, or if the value contains `".."`
  - `config["url"]` will have all *trailing* slash(es) removed
  - `config["baseurl"]` will have all *trailing* slash(es) and any extra slashes

### DIFF
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -99,7 +99,7 @@ module Jekyll
       # problems and backwards-compatibility.
       def from(user_config)
         Utils.deep_merge_hashes(DEFAULTS, Configuration[user_config].stringify_keys)
-          .add_default_collections
+          .add_default_collections.sanitize_url!.sanitize_baseurl!
       end
     end
 
@@ -283,6 +283,26 @@ module Jekyll
           " file accordingly."
         config[new] = config.delete(old)
       end
+    end
+
+    def sanitize_url!
+      value = ensure_string_without_trailing_slash(self["url"])
+      tap { |config| config["url"] = value }
+    end
+
+    def sanitize_baseurl!
+      value = ensure_string_without_trailing_slash(self["baseurl"])
+      value = "/#{value}".squeeze("/") unless value.empty?
+      tap { |config| config["baseurl"] = value }
+    end
+
+    private
+    def ensure_string_without_trailing_slash(value)
+      value = "" unless value.is_a?(String)
+      value = "" if value.include?("..")
+      value = value.dup
+      value.chomp!("/") while value.end_with?("/")
+      value
     end
 
     private
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -99,7 +99,9 @@ module Jekyll
       # problems and backwards-compatibility.
       def from(user_config)
         Utils.deep_merge_hashes(DEFAULTS, Configuration[user_config].stringify_keys)
-          .add_default_collections.sanitize_url!.sanitize_baseurl!
+          .add_default_collections
+          .sanitize_url!
+          .sanitize_baseurl!
       end
     end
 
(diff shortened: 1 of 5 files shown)


pathawks:
>`config["url"]` and `config["baseurl"]` will be reset to "" if the set value is nil, or an instance other than that of `String`, or if the value contains `".."`

I am wondering if it would be better for us to just error out of the build if `config["url"]` is an `Array` or something goofy like that. _Or_ maybe we should just `to_s` everything, in case the `baseurl` is `1234` instead of `"1234"`.

I'm not sure.

lib/jekyll/configuration.rb
```diff
@@ -99,7 +99,7 @@ class << self
       # problems and backwards-compatibility.
       def from(user_config)
         Utils.deep_merge_hashes(DEFAULTS, Configuration[user_config].stringify_keys)
-          .add_default_collections
+          .add_default_collections.sanitize_url!.sanitize_baseurl!
```
pathawks: Is it more readable if each of these is on a separate line?

lib/jekyll/configuration.rb
```diff
@@ -285,6 +285,26 @@ def renamed_key(old, new, config, _ = nil)
       end
     end
 
+    def sanitize_url!
```
pathawks: Should these be `private`? Why might a plugin want access to them?

lib/jekyll/configuration.rb
ashmaroli: Yes definitely. But, Rubocop complains that the class has exceeded the number of lines allowed..

pathawks:
https://github.com/jekyll/jekyll/blob/925fb6da45cbfeefa2accb7fa3faf99803c48a2d/test/test_filters.rb#L570-L577

This test thinks we should not attempt to use `baseurl` if it is improperly set. I think, rather than silently ignoring a wrong `baseurl`, we should either warn the user that something is broken, or just fail to build.

https://github.com/jekyll/jekyll/blob/925fb6da45cbfeefa2accb7fa3faf99803c48a2d/test/test_filters.rb#L597-L601

This test thinks we should always run `to_s` on `baseurl`.

https://github.com/jekyll/jekyll/blob/925fb6da45cbfeefa2accb7fa3faf99803c48a2d/test/test_commands_serve.rb#L261-L270

This test expects `site.url` to always end in a trailing slash.

ashmaroli:
> This test expects site.url to always end in a trailing slash.

You're wrong there.. :smiley:
The test in **`test_command_serve.rb`** tests to see if Jekyll overrides the given value (with `localhost:4000`) in `env("production")`. Yeah, the test doesn't make it obvious though..

pathawks:
> The test in **`test_command_serve.rb`** tests to see if Jekyll overrides the given value (with `localhost:4000`) in `env("production")`.

That is the true purpose of the test, yes. A (perhaps unintentional) second function of the test is that it expects `site.url` to always have a trailing slash, and complains when it does not (as is currently the case). I think we can just remove the trailing slash from ` assert_equal "https://jekyllrb.com/"` and call it good; the test would still be making sure it isn't set to `localhost` or anything else.

ashmaroli:
> it expects `site.url` to always have a trailing slash

#6947 reports that its not to good behavior to expect.. removing the trailing `/` in the test..

ashmaroli:
@pathawks leaving out changes to URLFilters in this PR.. I think we'll need to decide how to handle situations where the `config` has been altered erroneously by a plugin..

test/test_commands_serve.rb
```diff
@@ -262,10 +262,10 @@ def serve(opts)
         should "not update the site url" do
           expect(Jekyll).to receive(:env).and_return("production")
           expect(Jekyll::Commands::Serve).to receive(:start_up_webrick)
-          @merc.execute(:serve, { "watch" => false, "url" => "https://jekyllrb.com/" })
+          @merc.execute(:serve, { "watch" => false, "url" => "https://jekyllrb.com" })
```
pathawks: This will probably break many sites in weird ways. We need to document this really well.

PR #6993, opened 2018-05-07
Revert update to docs meant for v4.0

ashmaroli:
introduced via #6269
This feature will be released only in `v4.0`

TODO: Push this diff into `docs-40` once this has been merged.

/cc @DirtyF (I recommend a manual merge for this)


ashmaroli:
@DirtyF You're welcome. But do note that there may be conflicts on the `docs-40` because this content already exists in the `docs-40` branch..

PR #6995, opened 2018-05-07
Replace individual core team emails with link to team page

oe:
`at jekyllrb dot com` emails have proven to be like honey for spam and phishing emails, and as such this shouldn't use them because _actual_ requests regarding the Code of Conduct get lost. This links to the team page, which we should also update with the core team's email addresses (the non-jekyllrb.com variant).


pathawks:
Are they catching so much spam _because_ they are listed publicly on the website? If we replace them with different email addresses, won't we have similar problems?

Would it be better to replace them with proper contact forms?

oe:
@pathawks Nope. They're not on the website, that's the thing. They're currently only in the README.

PR #6998, opened 2018-05-08
Support for binary operators in where_exp filter

### LINKED ISSUE #6754: `and` and `or` operator in where_exp
- [x] I am on (or have tested on) ***Fedora*** GNU/Linux
- [x] I have reproduced it locally.
```
{% assign guides = site.pages
    | where_exp: "item" , "item.title  contains 'Data' or item.title  contains 'Groups'"
 %}
```
I got : `Error: Liquid syntax error (.../guide.html line 12): Expected end_of_string but found id included`

Are the operators `or` and `and` not supported in where_exp ?

Note: actually `and` can be done by chaining many where_exp but not `or`.

ashmaroli:
Resolves #6754 

Adapts [code from `Liquid::If`](https://github.com/Shopify/liquid/blob/4-0-stable/lib/liquid/tags/if.rb#L85-L107) to parse the Liquid expression.

/cc @Morikko


pathawks:
I really want this feature, and it seems like the code is working, but I cannot figure out how it is working. If it makes sense to another maintainer, cool. If not, I wonder if we could make the code a bit clearer so it will be easier to maintain?

ashmaroli:
> I wonder if we could make the code a bit clearer so it will be easier to maintain?

@pathawks I'll try and include comments.

mattr-:
@jekyllbot: merge +minor

DirtyF:
@ashmaroli This isn't documented?

PR #6999, opened 2018-05-08
WIP: Basic `incremental` support for Pages from plugins

ashmaroli:
This sets up the foundation that would allow plugins like `jekyll-feed` and `jekyll-sitemap` to generate a `feed.xml` or `sitemap.xml` only if their respective dependencies have been modified.

Of course, the above-mentioned plugins will have to implement the necessary code at their end to *track their output-file's dependencies
For example, the default `feed.xml` revolves around just the posts in a site. So, the following patch to the plugin's generator should suffice in fully realizing the incremental hook:

```ruby
    def generate(site)
      @site = site
      return if file_exists?(feed_path)
      absolute_path = site.in_source_dir(feed_path)
      site.regenerator.add_external(absolute_path)
      site.posts.docs.each do |doc|
        site.regenerator.add_dependency(absolute_path, doc.path)
      end
      site.pages << content_for_file(feed_path, feed_source_path)
    end
```

/cc @mattr- @benbalter @pathawks

### DIFF
--- a/lib/jekyll/regenerator.rb
+++ b/lib/jekyll/regenerator.rb
@@ -42,8 +42,24 @@ module Jekyll
       return true unless File.exist?(path)
 
       metadata[path] = {
-        "mtime" => File.mtime(path),
-        "deps"  => [],
+        "remote" => false,
+        "mtime"  => File.mtime(path),
+        "deps"   => [],
+      }
+      cache[path] = true
+    end
+
+    # Add a path outside the site's source directory, to the metadata.
+    # Since the file at that path doesn't physically exist, we'll only
+    # concern ourselves with its dependencies.
+    #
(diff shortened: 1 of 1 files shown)


ashmaroli:
@pathawks  Thanks for your feedback.
As of this writing, the proposed approach doesn't expect to do anything with Liquid at all.. Its just a rudimentary "hook" that would allow plugins to inject data regarding *"externally-located `Page`s"* into the metadata file..

In theory, this approach should automatically pick up new additions to a given array of files, but yes, I'll have to rethink this because "jekyll-feed" will soon be generating feeds for "other collections" and "categories"..

So, automatically grabbing the context by parsing the current template is a very nice idea.. :+1:

PR #7003, opened 2018-05-09
Textile is only supported through a converter plugin

### LINKED ISSUE #6989: Migration docs should mention removal of Textile from core
My original issue turned out not to be a bug; the problem is that I upgraded from an early build of Jekyll recently, and was unaware that Textile support now requires a plugin.  This is a matter of documentation only - the migration pages should mention when this occurred, and the 'directory structure' page should be more clear that using Textile requires the plug-in.

DirtyF:
Fix #6989


DirtyF:
@jekyllbot: 🚢  +docs

PR #7004, opened 2018-05-09
Rubocop cop for checking inline access modifier

nickskalkin:
Hi! This PR is continuation of this issue: #6768 

I've made a Cop for checking inline access modifiers, so it forces use to use one style. Take a look please

### DIFF
--- a/lib/jekyll/cleaner.rb
+++ b/lib/jekyll/cleaner.rb
@@ -16,12 +16,10 @@ module Jekyll
       FileUtils.rm_rf(metadata_file) unless @site.incremental?
     end
 
-    private
-
     # Private: The list of files and directories to be deleted during cleanup process
     #
     # Returns an Array of the file and directory paths
-    def obsolete_files
+    private def obsolete_files
       out = (existing_files - new_files - new_dirs + replaced_files).to_a
       Jekyll::Hooks.trigger :clean, :on_obsolete, out
       out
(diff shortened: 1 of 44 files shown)


mattr-:
I vote no. Idiomatic ruby is to have one private block per class unless required for other reasons. I know of no other project that does it like this.

mattr-:
And to be clear, if the team consensus is to do private for every method, I'll go along with it. I don't favor this style though and I don't understand it's appeal.

ashmaroli:
I like this.. a bit verbose but adds a lot to clarity.

@nickskalkin Does the cop support `--autocorrect`?
(I'm away from my system, so can't test it myself..)

ashmaroli:
@mattr- the one thing that I like about this style is that I can grok right away if the method is intended to be private, protected or public and that too without affecting the no. of lines in the class..
(contrast this against having to scroll up to make sure that the method is actually private, protected or public.. makes reading code that much easier..)

oe:
-1, adding `private` to every block is superfluous and non-idiomatic, especially if doc comments already state the exact same.

DirtyF:
closing in favor of #6768 then, thanks @nickskalkin for helping us decide once and for all on this 😅

nickskalkin:
@oe What is non-idiomatic? In some languages access modifiers prepend every method :) 

`especially if doc comments already state the exact same.` Then we should enforce contributors to write comments before every private/protected method? Because when the file is large it's hard to determine if you are in private/protected section.

oe:
@nickskalkin I'd be wary of changing something that's a universal standard in the Ruby world, especially considering that we're one of the largest and most well-known Ruby projects out there. I also have concerns that this would be confusing to beginners and newcomers to our project and Ruby in general.

nickskalkin:
@oe Ok, thanks for your explanation!

PR #7006, opened 2018-05-10
Load Rouge into memory as part of the library

ashmaroli:
..instead of relying on Kramdown to load `rouge` as `Rouge` is Jekyll's default highlighter..


pathawks:
Not all sites use syntax highlighting.
Is there a benefit to doing it one way over another?

PR #7010, opened 2018-05-13
Use OpenSSL instead of GnuTLS

isaacdgoodman:
By default, the test runs with GnuTLS. I experienced problems in testing various external sites as GnuTLS doesn't recognize some of the certificates. By using OpenSSL, these certificates are trusted.


kenman345:
Note: HTML-Proofer documentation actually makes this recommendation for optimal configuration/compatibility.

kenman345:
Actually, its also recommended to add 
`    - ca-certificates`

to the packages section defined for the openssl thing.

PR #7015, opened 2018-05-15
Add whitespace control to LIQUID_TAG_REGEX

### LINKED ISSUE #7009: Erroneous "excerpt modified" warnings and build errors in non-HTML pages
- [X] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [X] I ran `jekyll doctor` to check my configuration
- [X] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/

---

- [X] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux

---

- [X] I was trying to build.
- [ ] It was another bug.

## My Reproduction Steps

A minimal working example is as follows:

1. Create a non-HTML page (e.g. `test.xml`) with this basic structure:

```
---
---

{%- for post in site.posts %} 
xyzzy
{% endfor %}
```

2. Then attempt to generate the page with `jekyll serve`.
3. Result:

```
Configuration file: <path>/_config.yml
            Source: <path>
       Destination: <path>
      Generating... 
           Warning: Excerpt modified in <file>!
                    Found a Liquid block containing separator ' ' and has been modified with the appropriate closing tag.
                    Feel free to define a custom excerpt or excerpt_separator in the document's Front Matter if the generated excerpt is unsatisfactory.
             Error: could not read file <path>: undefined method `superclass' for nil:NilClass
         AutoPages: Disabled/Not configured in site.config.
        Pagination: Complete, processed 1 pagination page(s)
                    done in <x.yyyyyyyyy> seconds.
 Auto-regeneration: enabled for '<path>'
    Server address: http://127.0.0.1:4000
````

Adding something separated by a blank line above the for loop, or removing the whitespace-removing dash (`{% for ...` instead of `{%- for ...`) leaves the warnings but removes the build errors, e.g.:

````
---
---

X

{%- for post .... }
...
````

These errors also occur in similar minimal working examples with if loops:

````
---
---

{% if true %} 
xyzzy
{% endif %}
````

## The Output I Wanted

I'd expect the output of the `for` loop or `if` statement, or if there's something wrong with that syntax, an error or warning unrelated to excerpts, which aren't used anywhere on the page.

### PR DESCRIPTION
**Pull request by kylebarbour**
Fixes #7009. Simply adds `-?` to the beginning and end of the regexes used in `/lib/jekyll/excerpt.rb` to allow for Liquid whitespace control characters.

### DIFF
--- a/lib/jekyll/excerpt.rb
+++ b/lib/jekyll/excerpt.rb
@@ -128,7 +128,7 @@ module Jekyll
     #
     # Returns excerpt String
 
-    LIQUID_TAG_REGEX = %r!{%\s*(\w+).+\s*%}!m
+    LIQUID_TAG_REGEX = %r!{%-?\s*(\w+).+\s*-?%}!m
     MKDWN_LINK_REF_REGEX = %r!^ {0,3}\[[^\]]+\]:.+$!
 
     def extract_excerpt(doc_content)
@@ -141,7 +141,7 @@ module Jekyll
         head =~ LIQUID_TAG_REGEX
         tag_name = Regexp.last_match(1)
 
-        if liquid_block?(tag_name) && head.match(%r!{%\s*end#{tag_name}\s*%}!).nil?
+        if liquid_block?(tag_name) && head.match(%r!{%-?\s*end#{tag_name}\s*-?%}!).nil?
           print_build_warning
           head << "\n{% end#{tag_name} %}"
         end
--- a/lib/jekyll/excerpt.rb
+++ b/lib/jekyll/excerpt.rb
@@ -158,6 +158,11 @@ module Jekyll
 
     def liquid_block?(tag_name)
       Liquid::Template.tags[tag_name].superclass == Liquid::Block
+    rescue NoMethodError => e
+      Jekyll.logger.error "Error:",
+        "This page's excerpt appears to contain a Liquid tag which couldn't be " \
+        "parsed.\n" + e.message
+      raise
     end
 
     def print_build_warning
--- a/test/source/_posts/2018-05-15-closed-liquid-block-excerpt-whitespace-control.md
+++ b/test/source/_posts/2018-05-15-closed-liquid-block-excerpt-whitespace-control.md
@@ -1,0 +1,9 @@
+---
+title:  LIQUID_TAG_REGEX excerpt whitespace control test
+layout: post
+---
+
+{%- for post in site.posts -%}
+    You are in a maze of twisty little passages, all alike.
+    There's lots more to say about this, but that's enough for now.
+{%- endfor -%}
--- a/test/source/_posts/2018-05-15-excerpt-whitespace-control-variable.md
+++ b/test/source/_posts/2018-05-15-excerpt-whitespace-control-variable.md
@@ -1,0 +1,7 @@
+---
+title:  LIQUID_TAG_REGEX excerpt whitespace control test
+layout: post
+---
+
+{%- assign xyzzy = 'You are in a maze of twisty little passages, all alike.' %}
+{{- xyzzy -}}
--- a/test/source/_posts/2018-05-15-open-liquid-block-excerpt-whitespace-control.md
+++ b/test/source/_posts/2018-05-15-open-liquid-block-excerpt-whitespace-control.md
@@ -1,0 +1,10 @@
+---
+title:  LIQUID_TAG_REGEX excerpt whitespace control test
+layout: post
+---
+
+{%- for post in site.posts -%}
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

**Comment by pathawks** [comment]
Oh sweet! Thanks for catching this @kylebarbour :+1:

@ashmaroli Does `LIQUID_TAG_REGEX` have any tests? Does this PR need to add any, or maybe we should make a note to tackle that later?

**Comment by ashmaroli** [comment]
@kylebarbour Thank you for opening the PR.
The regex changes look good. This does need a test too though.

You may refer to #6724  to  see what tests are currently in place to test this functionality. If it looks confusing to you, the following note should guide you:

All you need to do is (at minimum):
  - add two new posts in the `test/source/_posts` directory, each containing  Liquid Block-type tag with the `whitespace_control` character.
One post (test subject) will contain only the "opening" construct as *part of the excerpt.* (The "closing" construct is to be written after a blank line). The other post (control subject) will contain the same content but without the blank line before the "closing" construct.
  - append assertions to `test/test_excerpt.rb` using the above two posts.
  - adjust `test/test_generated_site.rb`.

Run `script/test test/test_excerpt.rb` to see if your patch passes the base test.

*(Optional): Once you're comfortable with the tests, you can add and test more posts with various valid placement of the whitespace-control character.*

*If you still find it hard to add tests for this PR, feel free to inform so here.*

**Comment by kylebarbour** [comment]
Thanks for guiding me through the process! I'll take a stab at it.

**Comment by kylebarbour** [comment]
@ashmaroli Take a look and see what you think &mdash; I think this all works!

**Comment by ashmaroli** [comment]
@kylebarbour The tests look good.  :+1:

The message to Jekyll Logger converts all "whitespace" to a "single space" character. Therefore, `Jekyll.logger.error "Error:", "foo\nbar"` will simply output `Error: foo bar`..

So, IMO this should be in 2 separate calls:

```ruby
Jekyll.logger.error "Error:",
  "This page's excerpt appears to contain a Liquid tag which couldn't be parsed"
Jekyll.logger.error "", e.message
```

*Disclaimer: The suggestion above has not been tested locally..*

Since, I can't test locally, I'm not sure if Jekyll has to log `e.message` at all, since we're going to be calling `raise` with the same message..

kylebarbour:
Great, I've separated them. You're right, logging `e.message` is redundant, so I've removed it.

kylebarbour:
Actually, for some reason, that error message doesn't seem to be generating now that I'm checking this morning (not sure what happened), so I'm trying to fix it.

lib/jekyll/excerpt.rb
```diff
@@ -158,6 +158,11 @@ def extract_excerpt(doc_content)
 
     def liquid_block?(tag_name)
       Liquid::Template.tags[tag_name].superclass == Liquid::Block
+    rescue NoMethodError
+      Jekyll.logger.error "Error:",
+        "Excerpt in #{doc.relative_path} appears to contain a Liquid tag which " \
+        "couldn't be parsed."
+      raise
```
pathawks: This seems like a nice compromise to me for something like a `3.8.2`: a much more helpful error message without any material change in behavior :+1:

ashmaroli: Can we shorten the message a bit..? :
```ruby
Jekyll.logger.error "Error:",
  "Liquid tag in #{doc.relative_path} couldn't be parsed."
```

pathawks: I would favor a more descriptive error message over a less descriptive one, but if you think this is sufficient, :+1:

lib/jekyll/excerpt.rb
```diff
@@ -158,6 +158,11 @@ def extract_excerpt(doc_content)
 
     def liquid_block?(tag_name)
       Liquid::Template.tags[tag_name].superclass == Liquid::Block
+    rescue NoMethodError
+      Jekyll.logger.error "Error:",
+        "A Liquid tag in the excerpt of #{doc.relative_path} couldn't be " \
+        "parsed."
```
pathawks: This makes a lot of sense to me :+1:

lib/jekyll/excerpt.rb
ashmaroli: This is the reason why we decided to add a helpful error message.. :wink:

kylebarbour:
Thanks all! You were really helpful and fun to work with &mdash; as a truly rookie programmer I appreciate it.

pathawks:
@kylebarbour Thanks so much for all your work on this. Hoping this won't be your last contribution :wink::beers:

ashmaroli:
@pathawks Would you mind handling the backport of this patch for a `v3.8.2`..?

PR #7019, opened 2018-05-17
added new variable "file_size" for static_files

### LINKED ISSUE #7017: Adding `file_size` to static_files in Jekyll

---

Hi,

I'm new to Jekyll/Ruby, but I would like to add a new 'file_size' variable to static files to let users know the size of the file that they are going to download. 

I figured that I needed to start by adding the following method into */lib/jekyll/static_file.rb*:

```
    # Returns the file size in bytes
    def file_size
      File.size(path)
    end
```

However, I don't know how to go about making sure that this new feature works in the updated Jekyll before I add tests/submit a pull request. Are there any good resources for me to refer to?

Thanks!

prestonlimlianjie:
Create new liquid variable `file_size` for static files to allow Jekyll to display the download size of static files.

Fix #7017

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,7 @@ source "https://rubygems.org"
 gemspec :name => "jekyll"
 
 gem "rake", "~> 12.0"
+gem 'filesize', '~> 0.1.1'
 
 group :development do
   gem "launchy", "~> 2.3"
--- a/lib/jekyll/drops/static_file_drop.rb
+++ b/lib/jekyll/drops/static_file_drop.rb
@@ -4,7 +4,7 @@ module Jekyll
   module Drops
     class StaticFileDrop < Drop
       extend Forwardable
-      def_delegators :@obj, :name, :extname, :modified_time, :basename
+      def_delegators :@obj, :name, :extname, :modified_time, :basename, :file_size
       def_delegator :@obj, :relative_path, :path
       def_delegator :@obj, :type, :collection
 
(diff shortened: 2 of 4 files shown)


pathawks:
I don't have a real strong opinion on this, but it might help if we had some solid use cases for why this feature is needed.

Also, does this read the file size in constant time from the filesystem, or is this something that could take a long time if used with large files?

oe:
It feels like this should be an optional feature, maybe implemented through a small plugin.

prestonlimlianjie:
I'm building a company corporate website using Jekyll hosted on GitHub Pages; I'm allowing the download of PDF reports/forms on the site.

For better UX, I want to provide the users with the file size before they download any file. As such, users can decide whether to download it on their mobile data or waiting until they have WiFi connection before downloading said file.

It'd be great if it were in the Jekyll core feature set so in the (hopefully not-too-distant) future, it can be put into GitHub Pages too. And I think it's quite a simple/small feature too :)

DirtyF:
> I'm building a company corporate website using Jekyll hosted on GitHub Pages.

There are other hosting solutions without plugins limitation — while keeping your code on GitHub.
This is not a sufficient reason to add core features to Jekyll.

ashmaroli:
FWIW, my vote is on marking this as `plugin-feature`

pathawks:
@prestonlimlianjie Thanks for that information.

I think this is a good candidate for a plugin feature. If it gains widespread popularity, we can always add it to Jekyll core later.

PR #7027, opened 2018-05-19
Fix --unpublished not affecting collection documents

### LINKED ISSUE #7026: Extend jekyll serve's --unpublished flag to also filter collections

---

---

## My Reproduction Steps

- Have a Jekyll collection
- Set `published:  false` in the front-matter of a collection item
- Run `jekyll serve --unpublished`

## The Output I Wanted

At this point unpublished items from the collection will still show on the Jekyll site, whereas unpublished posts will not. This behaviour occurs across multiple of my Jekyll sites. I've also created a fresh install using the quick start and confirmed that it occurs there also.

I'm not sure whether this is a bug or the feature working as designed. If it's the latter I'd like to make the case that intuitively the `--unpublished` flag should apply to both posts and collections given their `published` front-manner is identical and users might expect that commands which discriminate based on `published` status would apply equally to both. I generally use `--unpublished` to work on content that is still in a draft stage but as I use collections more and more not having the ability to easily switch between the 'draft' and 'live' version of the site's collections means having to find work arounds to show/hide content.

I'm not super familiar with Ruby but have dug around the source to try and get a handle for where the parameter is passed and how it is fed through to filter the posts. Before attempting a PR I'd like to check whether or not this is an intended behaviour or if there is some sort of technical hurdle that obstructs changing the behaviour here. In the case that this is a trivial change in the code I would very much anyone taking this on as while I'd like to improve my ruby skills I'm not confident I have the time or ability to do so anytime soon.

Thanks also to everyone who helps develop and maintain Jekyll; I've been using it for about five years now and it's always been exactly what I needed!

philipbelesky:
Fixes #7026 

Note that the changes to `collection.rb` different slightly to what was suggested — the `fmt` linter gave me a warning against using parens with `unless`. Please let me know if any further modifications are needed.

As a sidenote thanks for the very quick and friendly prompts for how to add this PR. Getting a chance to learn about Cucumber was great!

### DIFF
--- a/features/collections.feature
+++ b/features/collections.feature
@@ -263,14 +263,14 @@ Feature: Collections
     And the "_site/puppies/snowy.html" file should not exist
     And the "_site/puppies/hardy.html" file should not exist
 
-  Scenario: Access rendered but unpublished collection via Liquid
+  Scenario: Access rendered collection with future date and unpublished flag via Liquid
     Given I have a _puppies directory
     And I have the following documents under the puppies collection:
       | title  | date       | content             | published |
       | Rover  | 2007-12-31 | content for Rover.  | true      |
       | Figor  | 2007-12-31 | content for Figor.  | false     |
-      | Snowy  | 2007-12-31 | content for Snowy.  | true      |
-      | Hardy  | 2007-12-31 | content for Hardy.  | false     |
+      | Snowy  | 2199-12-31 | content for Snowy.  | true      |
+      | Hardy  | 2199-12-31 | content for Hardy.  | false     |
     And I have a "_config.yml" file with content:
     """
     collections:
@@ -286,7 +286,7 @@ Feature: Collections
     And I should not see "<div>Figor</div>" in "_site/index.html"
     And I should not see "<div>Hardy</div>" in "_site/index.html"
     And the "_site/puppies/rover.html" file should exist
-    And the "_site/puppies/snowy.html" file should exist
+    And the "_site/puppies/snowy.html" file should not exist
     And the "_site/puppies/figor.html" file should not exist
     And the "_site/puppies/hardy.html" file should not exist
     When I run jekyll build --unpublished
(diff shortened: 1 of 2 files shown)


ashmaroli:
Thanks for opening the PR.
Removing the superfluous parentheses are fine.

I'd like to leave a few comments regarding the included cucumber test:
  - Unpublished documents are never rendered.
When one passes the `--unpublished` switch, Jekyll overrides the front matter setting of `published: false` which leads to *all* non-future dated documents to be *rendered*.
  - "Snowy" and "Hardy" are meant to be **future-dated** like in the preceding scenario. We need this to test how the proposed patch affects such documents.
(Documents other than posts differ in behavior here. **Future-dated documents should be accessible via Liquid, but not written to destination unless `--future` switch has been passed as well.**)

philipbelesky:
Thanks for the feedback! I have created a second test to account for the scenario when `output: false` and incorporated a consideration of future-dated documents into both tests. Please let me know if anything else is required or you'd prefer the PR closed and re-opened as a single commit.

ashmaroli:
*In case anyone's wondering why the proposed fix doesn't involve the `Publisher` class here.., this is because it requires "enhancing" `Publisher`'s API to get the desired outcome, which is outside the scope of a "patch" release, as per Principles of SemVer..

This will be handled by updating #6997 for `v4.0`*

features/collections.feature
```diff
@@ -263,6 +263,104 @@ Feature: Collections
...
+    collections:
+      puppies:
+        output: true
+    """
+    And I have a "index.md" page that contains "{% for puppy in site.puppies %}<div>{{ puppy.title }}</div>{% endfor %}"
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "<div>Rover</div>" in "_site/index.html"
+    But I should see "<div>Snowy</div>" in "_site/index.html"
```
DirtyF: A document with a future date gets published even without `--future` option?

DirtyF:
> Future-dated documents should be accessible via Liquid

Why is that?

ashmaroli:
> Future-dated documents should be accessible via Liquid
> > Why is that?

@DirtyF For example, one might have a collection labelled "events" with documents for past, current and upcoming events which are then rendered using Liquid `for` loop..

lib/jekyll/collection.rb
```diff
@@ -212,7 +212,9 @@ def container
     def read_document(full_path)
       doc = Document.new(full_path, :site => site, :collection => self)
       doc.read
-      docs << doc unless doc.data["published"] == false
+      unless doc.data["published"] == false && !site.unpublished
```
pathawks: This `unless` makes my head hurt.

"**Unless** **not** the site is **un**published"

features/collections.feature
pathawks: I don't think it gets _output_, but it does get _rendered_ and is accessible via Liquid. It's kind of weird behavior. 🤷‍♂️

lib/jekyll/collection.rb
pathawks:
```ruby
if site.unpublished || doc.data["published"] != false
  docs << doc
end
```

❓

pathawks:
Wow, great work here! I appreciate the tests that have been added :tada:

lib/jekyll/collection.rb
ashmaroli: Reversing the conditional leads us to another cleaner alternative:
```ruby
if site.unpublished || doc.published?
  docs << doc
end
```

philipbelesky: Agreed, just pushed up a commit that amends this.

pathawks:
@jekyllbot: :shipit: +fix

PR #7037, opened 2018-05-23
Clarify definition of 'draft'

AliLynne:
Made it more clear that there shouldn't be a date in the filename. This was confusing to me as I thought it meant that there shouldn't be a date in the front matter. It took me awhile to figure out that I just needed to not put a date on the filename.

### DIFF
--- a/docs/_docs/drafts.md
+++ b/docs/_docs/drafts.md
@@ -3,7 +3,7 @@ title: Working with drafts
 permalink: /docs/drafts/
 ---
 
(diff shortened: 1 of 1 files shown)


chrisfinazzo:
Welcome @AliLynne, good to have you here.

DirtyF:
Thanks @AliLynne, for the record there is an official plugin to help working with drafts and posts: https://github.com/jekyll/jekyll-compose

DirtyF:
@jekyllbot: :shipit: +docs

🙌 @AliLynne

PR #7041, opened 2018-05-28
Install Jekyll on Windows: Add Ruby to PATH 

anonymoustafa:
I experienced difficulties using fantastic jekyll cause, my windows os only could understand jekyll commands in bin folder of ruby installation folder. After searching internet I found that I have to add the executive files route on my hard drive to the system path.

### DIFF
--- a/docs/_docs/windows.md
+++ b/docs/_docs/windows.md
@@ -19,7 +19,7 @@ We only cover RubyInstaller-2.4 and newer here, older versions need to [install 
 2. Open a new command prompt window from the start menu, so that changes to the `PATH` environment variable becomes effective.
    Install Jekyll and Bundler via: `gem install jekyll bundler`
 3. Check if Jekyll installed properly: `jekyll -v`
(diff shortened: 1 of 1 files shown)


XhmikosR:
1. I don't use Ruby 2.5 but
2. If you have chosen to add Ruby to PATH it works just fine and this is not needed for sure.

ashmaroli:
I agree with @XhmikosR here. Adding Ruby to the `PATH` is a part of installing Ruby, which is beyond the scope of Jekyll Documentation.

kacperduras:
I'm with @XhmikosR and @ashmaroli.

DirtyF:
@anonymoustafa Thanks for the note, even if it does not make it to our docs, this could be useful for people landing on this PR 👍

oe:
@DirtyF I think we could at least add a small note? This is legitimately a thing that people stumble into while following the installation instructions from _our docs_.

PR #7042, opened 2018-05-28
override `url` in development environment in `build` and `serve`

### LINKED ISSUE #6397: Cannot override JEKYLL_ENV in `jekyll build`

Crunch09:
Previously this was only the case when running `serve` which caused inconsistencies between running `build` and `serve`.

This fixes #6397.

@jekyll/build


features/site_configuration.feature
```diff
@@ -247,3 +247,16 @@ Feature: Site configuration
...
+
+  Scenario: override url in development environment
+    Given I have a _posts directory
+    And I have a configuration file with:
+      | key         | value               |
+      | url         | tobeoverwritten.dev |
+    And I have the following posts:
+      | title   | date       | content      |
+      | foobar  | 2018-01-01 | {{site.url}} |
+    When I run jekyll build ENV=development
```
ashmaroli: Since this is the default value, shouldn't we simply test for `jekyll build`..?

ashmaroli:
This needs to be well-documented that simply running `jekyll build` will henceforth add `http://localhost:4000` to all links **unless `JEKYLL_ENV` has been overridden beforehand.**

Crunch09:
@oe Thanks, missed that scenario and updated it, as using the URL might not be the simplest example anymore to read from the `_config.yml`.
@ashmaroli Thanks for you review! I updated the test now and also happy to add documentation once we decided that we really want this to be the default behaviour for `jekyll build` as well.

pathawks:
This change bothers me. When I started using Jekyll, I had only a passing knowledge of environment variables, and would have been very frustrated if there seemed to be know good way for me to convince Jekyll to respect my set `site.url`. I think we need to be very careful about when we decide to intervene and override an option that has been explicitly set by the user.

I understand that currently `jekyll build` and `jekyll serve` are not consistent, but I am not sure that the solution is more "magic" in `build`.

DirtyF:
@pathawks Would a message in the console output hel ? Should the user opt in for this behavior?

`By default Jekyll runs in developement environment, to build your site with your defined production URL prefix your command with JEKYLL_ENV=production`

oe:
@DirtyF This wouldn't be consistent syntax with all shells, I'd say we either make it opt-in or just add it to the docs (with examples for different shells)

ashmaroli:
I'm personally :-1:  with this change because I feel having to set an environment variable each time I generate a production build is not user-friendly.
The current behavior is a lot better: Develop and preview my site locally with `bundle exec jekyll serve` but build a production-ready site with a simple `bundle exec jekyll build`

chrisfinazzo:
@ashmaroli,  I eventually got to a point where this issue pushed me into trying Gulp because the rewrites broke links constantly. This PR makes that not a problem anymore. @oe and @DirtyF's suggestion of an opt-in seems like a reasonable middle ground.

I have to type fewer characters and that initial setup with environment variables pays off down the road because I don't have to think about what happens when each task runs. In my case, deploying pushes to `master` and SSH's to my server in one step.

DirtyF:
> having to set an environment variable each time I generate a production build is not user-friendly.

A user-friendly production build should be scripted and automated. And luckily we have CI or services like GitHub Pages that will do that for you. Just set an environment variable to your CI and forget it.

pathawks:
> A user-friendly production build should be scripted and automated.

A user-friendly _development_ build should be scripted and automated. 😕

Crunch09:
Sorry for my long absence on the jekyll project and thanks all for your comments.
Should we add an opt-in to the config? To me this also seemes to be the most sensible solution given the valid concern.

ashmaroli:
> Sorry for my long absence on the jekyll project

@Crunch09 You don't have to apologize. Everyone contributes on their own time.
Glad to have you back!! :smiley:

mattr-:
I'm not sure that this solves #6397 as expected in the report. By default, nothing should change unless I override it from the command line. IOW, unless I'm running `JEKYLL_ENV=development be jekyll build`, I should expect to get the value from the production environment.

Crunch09:
@mattr- You're right, but the current state is kind of a dilemma as the environment already falls back to `development` whether `JEKYLL_ENV=development` is provided or not: https://github.com/jekyll/jekyll/blob/4d6edb3a269e37860aabd2d809d18edb947514f0/lib/jekyll.rb#L96-L98

This PR is just trying to produce the same output for the `development` environment regardless if it's used with `build` or `serve`.

PR #7049, opened 2018-05-31
Add Hints for some Improved Travis Config in Doc

patrickfav:
Adds hint about caching, notifications and uses ruby which is present in trusty dockers on Travis.

### DIFF
--- a/docs/_docs/continuous-integration/travis-ci.md
+++ b/docs/_docs/continuous-integration/travis-ci.md
@@ -90,7 +90,7 @@ Your `.travis.yml` file should look like this:
 ```yaml
 language: ruby
 rvm:
-- 2.3.3
+  - 2.4.1
 
 before_script:
  - chmod +x ./script/cibuild # or do this locally and commit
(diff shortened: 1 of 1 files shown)


ashmaroli:
What is the need to replace `Ruby 2.3` with `Ruby 2.4`? Jekyll 3.x works fine with Ruby 2.2 and above..

patrickfav:
The idea was to use a Ruby version that does not need to be downloaded (ie. is available on the trusty docker) to speed up build. The choice is between 2.3.4 or 2.4.1, so I just used the latest one. AFAIK jenkyll will remove support for 2.2 in v4, so I just want to make the guide "future proof" (and newer versions have security fixes?). Is there any reason to use 2.3.x instead of 2.4.x?

ashmaroli:
There's nothing wrong in using Ruby 2.3 over the newer versions.. nor the other way around..
I was wondering what the rationale behind your edits would be..

DirtyF:
> Is there any reason to use 2.3.x instead of 2.4.x?

As it's still officially supported until 2019-03-31, we also support branch 2.3
https://www.ruby-lang.org/en/downloads/branches/

Note that GitHub Pages is using branch 2.4
https://pages.github.com/versions/

pathawks:
@jekyllbot: :shipit: +docs

PR #7064, opened 2018-06-07
Livereload sleep

jamesluberda:
This change adds an option to `jekyll serve` to delay auto-regeneration when using livereload to address the scenarios in which livereload is likely to return a 404 error due to rapid successive system-level file modifications to a page being watched by a client. In these scenarios, a sequence of rapidly executed file modifications result in a page being repeatedly auto-regenerated and unavailable during the window in which the client requests the modified page following the first modification.

The scenarios which this option is primarily intended to address concern file updates made through remote file transfers, and is based specifically upon experience with Dropbox (and to a lesser degree, pCloud). A single change to the contents of a file hosted by or linked to Dropbox and served by Jekyll typically results in three, and sometimes four visible system-level file modifications. These, in turn, result in a number of rapid subsequent callbacks from [Listen](https://github.com/guard/listen) with consequent auto-regenerations. As a result of the file being made temporarily unavailable by the auto-regenerations immediately following the first, the client often receives a 404 error rather than the requested content. This behavior has also been observed, though rarely, with sftp-based file transfers, which generally only result in two system-level file modifications within an interval short enough to trigger only one auto-regeneration. 

With the option `--livereload-sleep [SECONDS]` enabled, [jekyll-watcher](https://github.com/jekyll/jekyll-watch) will pass along the specified livereload-sleep value, or a default of 1, to Listen as the `wait_for_delay` parameter (overriding Listen's default of 0.10 seconds), delaying the callbacks, thus causing auto-regeneration to occur only once across rapid successive changes to a file within the specified "sleep" time. As a result, the client's request is unlikely to occur with an auto-regeneration underway, and will instead receive the updated content as expected.

This pull request provides the code to make the option available to the user. An additional pull request has been made for jekyll-watcher to implement the functionality.

I also noted that the base livereload option has limited documentation at present, with notice appearing the release notes for 3.7.0 and in usage.md. No mention in configuration.md. I could propose some content based on the initial PR, [#5142](https://github.com/jekyll/jekyll/pull/5142), under this or another PR, if desired/appropriate.


jamesluberda:
Just checking in--is there anything more needed to have this reviewed and merged along with its supporting PR jekyll/jekyll-watch#68 ?

PR #7067, opened 2018-06-08
Make file name matcher check on base name of entry

ainopara:
If  a folder in `_posts` named `2018-06-09-Test.textbundle`, which can pass the matcher check, all files in the folder will pass the check no matter what name they have.


ashmaroli:
@ainopara Will you be able to write some tests for this..?

kenman345:
@ainopara This looks pretty solid to me. Can you show me the issue you're having in a repository so I can mock that up for tests to help you along?

PR #7084, opened 2018-06-19
Add `highlight` `start_line` option to specify starting line number

gasi:
When including partial code snippets from a larger source file, it can be useful to specify at which line the code starts. This adds option `start_line=<n>` to `highlight` Liquid tag.

Was on the fence between choosing Pygment `linenostart` and Rouge `start_line` option and went with the latter for readability, but I am open to changing it if project lead(s) feel strongly.


gasi:
@DirtyF @pathawks Would you mind taking a look? Happy to make any changes needed to land this. Thanks!

lib/jekyll/tags/highlight.rb
```diff
@@ -83,6 +85,7 @@ def parse_options(input)
         end
 
         options[:linenos] = "inline" if options[:linenos] == true
+        options[:start_line] = Integer(options[:start_line]) unless options[:start_line].nil?
```
oe: nit: I'd personally try to avoid `unless` statements because they're a bit harder to wrap your head around than (in my opinion) an `if !(condition)`.

test/test_tags.rb
```diff
@@ -198,6 +214,14 @@ def highlight_block_with_opts(options_string)
           @result
         )
       end
+
+      should "render markdown with pygments with line numbers and custom start line" do
```
DirtyF: Note: Support for Pygment will be dropped in 4.0

lib/jekyll/tags/highlight.rb
pathawks: @oe I _believe_ that Rubocop complains about `if ! condition` and suggests using `unless condition` instead.

gasi: Indeed, @pathawks got it right:

3__bash_and_1__bash

I don’t feel strongly if there is another preferred way to express this.

test/test_tags.rb
gasi: Thanks, that’s good to know. That might factor into API naming as `linenos` comes from Pygment (see below).

gasi:
@oe and @DirtyF: First of all, thanks for the quick review!

> @DirtyF: Quick test without an include: it modifies the line numbers but always start at the beginning of the code snippet.

Good point! That is by design as `{% highlight %}` is only concerned with presentation and not with how or where the data comes from. Maybe that’s why Pygment’s naming convention of `linenostart` is clearer — it expresses that it’s just where the line number starts, not at which line the code listing begins to be included (although that’d be nice to solve with a more code-centric `include` tag).

Given your feedback, I was going to propose adopting `linenostart` from Pygment. But then you mentioned Pygment will be dropped in 4.0. Also, `linenostart` is a poor API name for googling. I was trying to find this functionality by searching ’jekyll highlight start line‘, etc. but all that came up was Rouge’s `start_line` API, hence the choice above.

Ultimately, I’d propose `linenostart` to align with existing `linenos` and long-term, when Pygments is dropped in 4.0, add clearer APIs such as:

- `line_numbers` (instead of `linenos`)
- `line_numbers_start` (instead of `linenostart`)

What do you think?

lib/jekyll/tags/highlight.rb
oe: @pathawks dangit!

oe: disregard my nit then

pathawks:
> and long-term, when Pygments is dropped in 4.0, add clearer APIs

I don't believe this feature will be released until 4.0, by which time Pygments will have already been dropped. Therefore, I don't think we need to worry much about Pygments for this PR.

gasi:
> @pathawks I don't believe this feature will be released until 4.0, by which time Pygments will have already been dropped. Therefore, I don't think we need to worry much about Pygments for this PR.

Understood. To clarify, I agree that we can drop the implementation support for Pygments above but one thing that will presumably survive the dropping of Pygments is the `{% highlight ruby linenos %}` API as `linenos` comes straight from Pygments: http://pygments.org/docs/formatters/

If we wanted to have a consistent API, we’d also use Pygments `linenostart` naming, but I find that overall to be poor API naming. Since Pygments is dropped, maybe we can alias `linenos` with `line_numbers` and adopt `start_line` straight from Rouge (or `line_numbers_start`): https://github.com/jneen/rouge/tree/ea7675f741ee28f3f177ff32a9bde192742ffc59#formatters

ashmaroli:
@gasi, not that its a concern but what if `start_line` is used without `linenos` option?

gasi:
> @ashmaroli: not that its a concern but what if `start_line` is used without `linenos` option?

It’s currently a no-op. What’s Jekyll’s philosophy? Should we fail fast?

ashmaroli:
> It’s currently a no-op

@gasi I was wondering if you could wire it up to a single option..

`{% highlight ruby start_line=42 %}` automatically *enables* the `linenos` option internally..

pathawks:
@ashmaroli That makes sense to me :+1:

dAmnation69:
is this enhancement available in the latest version ?

ashmaroli:
@gasi Will you be able to update this branch against the current `master`? Thanks.

gasi:
> @ashmaroli  Will you be able to update this branch against the current `master`? Thanks.

Unfortunately, no. This PR is over a year old and I haven’t worked with Jekyll since. I won’t be able to rebase and test this in the medium-term. Feel free to take over this PR to land the feature.

ashmaroli:
@gasi No issues. Thank you for getting the ball rolling.

PR #7086, opened 2018-06-19
Added plugin json-get.

iBobik:


DirtyF:
@jekyllbot: :shipit: +docs

PR #7095, opened 2018-06-25
Instructions to view theme’s files under Linux

sanmai:
Added instructions to view theme’s files under Linux


pathawks:
>  A Finder or Explorer window opens showing the theme's files and directories.

This is not true of the `find` command; not sure if that’s a problem.

I wonder if it would be better to omit the `cd` and just `find $(bundle show minima)` directly?

sanmai:
On Linux it's largely dependent on desktop environment and/or distribution used. Although it appears that `xdg-open` is the closest thing to macOS's `open` there is. I've updated this PR to use `xdg-open`.

DirtyF:
@jekyllbot: merge +docs

PR #7104, opened 2018-07-02
Fix tests to have "expected value" as the first argument passed to `assert_equal`

pathawks:
The expected value needs to come _before_ the actual. Otherwise, the test results are confusing.

From [the docs](https://apidock.com/ruby/Test/Unit/Assertions/assert_equal):

```ruby
assert_equal(exp, act, msg = nil)
```

### DIFF
--- a/test/test_page_without_a_file.rb
+++ b/test/test_page_without_a_file.rb
@@ -62,7 +62,7 @@ class TestPageWithoutAFile < JekyllUnitTest
 
           # assert differences with Jekyll::PageWithoutAFile instance
           if basic_attrs.include?(prop)
-            assert_equal @page[prop], value, "For <page[\"#{prop}\"]>:"
+            assert_equal value, @page[prop], "For <page[\"#{prop}\"]>:"
           else
             assert_nil @page[prop]
           end
--- a/test/test_page_without_a_file.rb
+++ b/test/test_page_without_a_file.rb
@@ -41,6 +41,10 @@ class TestPageWithoutAFile < JekyllUnitTest
 
       should "have basic attributes defined in it" do
         regular_page = setup_page("properties.html", :klass => Page)
+        # assert a couple of attributes accessible in a regular Jekyll::Page instance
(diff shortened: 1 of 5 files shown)


pathawks:
I honestly can't figure out what the purpose of this test is anyway. 🤷‍♂️

ashmaroli:
@pathawks I've updated the test comments to better express intention. Hope its clear now.

ashmaroli:
Failing test is unrelated to this PR. Seems like an issue with JRuby archive at `https://s3.amazonaws.com/jruby.org/downloads`

test/test_page_without_a_file.rb
```diff
@@ -56,13 +56,17 @@ def render_and_write
           "url"       => "/properties.html",
         }
         attrs.each do |prop, value|
-          # assert the props being accessible in a Jekyll::Page instance
+          # assert a couple of attributes accessible in a regular Jekyll::Page instance
           assert_equal "All the properties.\n", regular_page["content"]
           assert_equal "properties.html", regular_page["name"]
```
pathawks: @ashmaroli Any idea what the purpose of these `regular_page` tests is?

ashmaroli: Ah.. these should be "*variants*" too.. I'll override your commit.. stay tuned..

ashmaroli: Never mind.. Its best left as is.. :grin:

pathawks:
I think this is all ready for review, and would specifically help with some work that I am doing.

Can we get a review from @jekyll/core?

ashmaroli:
There probably is more tests in various other files, that needs fixing. But manually going through each is not practical, I guess..

pathawks:
> There probably is more tests in various other files, that needs fixing.

I have only been fixing them as I encounter them. If we find more in the future, we can fix them as they arise. :+1:

ashmaroli:
manual-merge ?

PR #7108, opened 2018-07-04
WIP: Cache rendered includes

pathawks:
Many sites use includes for things like headers or footers that are identical on all pages.

Instead of rendering these includes for each page that includes them, it would be nice to detect when an include does not rely on anything page specific and cache the rendered include.

For starters, we can cache any includes that do not have any Liquid (which is not a very common case). Next, we could cache includes that only access `{{ site.* }}` (which is a more common case) by creating a special Liquid context which is capable of reporting if anything that hasn't been whitelisted has been accessed.

This PR is in a very early stage, and feedback is appreciated.

### DIFF
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -71,6 +71,7 @@ module Jekyll
   autoload :Regenerator,         "jekyll/regenerator"
   autoload :RelatedPosts,        "jekyll/related_posts"
   autoload :Renderer,            "jekyll/renderer"
+  autoload :ReportingContext,    "jekyll/reporting_context"
   autoload :LiquidRenderer,      "jekyll/liquid_renderer"
   autoload :Site,                "jekyll/site"
   autoload :StaticFile,          "jekyll/static_file"
--- a/lib/jekyll/reporting_context.rb
+++ b/lib/jekyll/reporting_context.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module Jekyll
+  class ReportingContext < Liquid::Context
+    SAFE_KEYS = %w(jekyll site).freeze
+
+    def find_variable(key)
+      @cachable = false unless SAFE_KEYS.include? key
+      super
+    end
+
+    def cachable?
+      @cachable != false
+    end
+  end
+end
--- a/lib/jekyll/tags/include.rb
+++ b/lib/jekyll/tags/include.rb
@@ -112,6 +112,9 @@ module Jekyll
         raise IOError, could_not_locate_message(file, includes_dirs, safe)
       end
 
+      # rubocop:disable Metrics/AbcSize
+      # rubocop:disable Metrics/CyclomaticComplexity
+      # rubocop:disable Metrics/MethodLength
       def render(context)
         site = context.registers[:site]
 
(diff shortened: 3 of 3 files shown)


DirtyF:
Related: https://github.com/benbalter/jekyll-include-cache

kenman345:
I see this is not passing right now. I am wondering if perhaps this is a place we might also wish to use `||=`[Memoization](https://en.wikipedia.org/wiki/Memoization)? Say perhaps that a header or footer actually is identical on all pages but is still using `_config.yml` or `site` level variables in liquid tags. 

Like, a header might be a part of a theme, the theme meant to be flexible as its a gem, so it would have the `site.url` in it and `site.title` in it perhaps, but otherwise nothing unique per page. Memoization when that is the case could prove to make it even more commonly returning a cached version than this original proposal and still be pretty safe as long as they are only using the values from site level and not any logical expressions to determine output.

pathawks:
@DirtyF Thanks for bringing that up. **jekyll-include-cache** was a big part of the inspiration for this. :+1:

kenman345:
I wonder if perhaps we could render all includes with the `site` data and then if they still have additional tags, then we cache to an includes object the `site` rendered partial, and use that when we dont have an include cached yet.

Essentially, this is like saying, well `serve` command doesnt reload the `_config.yml` data anyways so lets just process all those things right at the beginning and store in a secondary cache the data instead of re-opening it each time because its still going to be different each time its asked to be included in a page. 

So during development its slightly faster and during a single build, it would still contain the same benefits as you were hoping but is simply unable to gain any additional performance out of my suggestion as its not re-rendering. 

I guess what I'm saying is my idea of caching the includes is a bit more aggressive but nets us benefits when developing, and maintains the same performance as you were originally aiming for (or maybe slightly better) with your current PR.

pathawks:
Does not seem like this is any faster.

| #7108                                    |                333.52 |
| `master`                                 |                334.94 |

ashmaroli:
> Does not seem like this is any faster.

One way to compare this against Ben's plugin is by having a branch/PR that adds the plugin to Jekyll out-of-the-box, and then run Utterson against that branch..

pathawks:
I'm sure Ben's plugin does make things faster, but he is using a different approach that does not try to perform Magic™. It is not, however, a drop-in replacement for `{% include %}`, so it could not be tested via Utterson (because each site would need to be tweaked to use `{% include_cached %}`).

ashmaroli:
> so it could not be tested via Utterson

Ah yes, I forgot about that.. Of course you could monkey-patch to override our `include` tag... but yeah, not probably worth it, since its a different approach..

kenman345:
Monkey-patch...haven’t heard that one but it’s gonna get used at work a lot.

pathawks:
This has been updated to use the new Cache API.

It also depends on #7190 to avoid trying to dump a `Hash` with default proc.

DirtyF:
@jekyllbot: merge +major

ashmaroli:
@DirtyF  This was a PR to merge into `dont-dump-undumpable` branch.. not `master`..
Should nt be registered in the `master` branch yet or in the `HISTORY.markdown`..

ashmaroli:
The code here seems to broken #7190 
I'm restoring this branch so that whatever the issue is, can be revisited once #7190 is merged into `master`..

PR #7112, opened 2018-07-06
GitHub enables you to use themes from other repos

bvhme:
Although GitHub pages do not support pulling in themes from Gems it does support pulling in themes from other repositories and thus can be used in a similar way.


chrisfinazzo:
Welcome @bvhme, good to have you here.

DirtyF:
@jekyllbot: merge +docs

PR #7114, opened 2018-07-07
add jekyll-xml-source

mcred:
Add Jekyll XML Source to the Plugins list. 

Signed-off-by: Derek Smart <derek@grindaga.com>

### DIFF
--- a/docs/_docs/plugins.md
+++ b/docs/_docs/plugins.md
@@ -940,6 +940,7 @@ You can find a few useful plugins at the following locations:
 - [jekyll-pwa-plugin](https://github.com/lavas-project/jekyll-pwa): A plugin provides PWA support for Jekyll. It generates a service worker in Jekyll build process and makes precache and runtime cache available in the runtime with Google Workbox.
 - [jekyll-algolia](https://community.algolia.com/jekyll-algolia/): Add fast and relevant search to your Jekyll site through the Algolia API.
 - [jekyll-get](https://github.com/18F/jekyll-get): Download data from external JSON API sources to use in generating a site.
+- [jekyll-xml-source](https://github.com/mcred/jekyll-xml-source): Download XML and RSS from external sites for use in site data.
 
 <div class="note info">
   <h5>Submit your gem plugins</h5>


pathawks:
This isn't a Gem. I don't remember if we are now only allowing new plugins if they are packaged as Gems, or if that is only something that we had discussed but not yet implemented.

DirtyF:
@pathawks Ideally we'd query jekyll-plugins label on GitHub and automatically list the plugins :)

mcred:
I read through https://jekyllrb.com/docs/contributing/ and looked at most of the existing plugins and they all use the "copy this to your _plugins folder' instruction. If it needs to be a gem, I can do that. Just let me know.

ashmaroli:
@mcred Whether it should be a gem or just a plain `.rb` file is upto the developer.
If the code *is stable and won't be changing* in the near future, then it doesn't have to be a Rubygem. But if the code is going to be refined with bug-fixes, or enhanced with additional features, then its best released as a Rubygem so that the end-user will be able to switch versions.

jamesluberda:
@mcred One other thing to consider in general when deciding gem vs. plain `.rb` file copied to the `_plugins` folder is that if the user is running the `github-pages` gem, the `plugins_dir` configuration value is set at build time to a random string, meaning that plugins in that dir (or any other user-specified dir) will not run. A gem can (even if it won't run on GitHub).

ashmaroli:
@jamesluberda Correction. Third-party plugins won't run with `github-pages` irrespective of whether it's a gem or a plain `.rb` file.
The phenomenon you're describing is an undocumented loophole though. Locally, `github-pages` does not block Bundler. A gem-based plugin *may run* with `github-pages` if the plugin-gem is listed under the `:jekyll_plugins` group in the `Gemfile`.

mcred:
@jamesluberda In this case, the file is very simple and won't be undergoing more development. Re: `github-pages`: I'm aware of that they use the `--safe` flag during build which disallows non white-listed plugins. I was not aware of the Bundler loophole. I've solved this problem by using Travis CI and a controlled build step.

ashmaroli:
@mcred Another point in favor of creating a Rubygem is usage statistics. Publishing to `rubygems.org` will give you insight into *how many times a gem has been downloaded..* :slightly_smiling_face:

pathawks:
Sorry I derailed this conversation. I thought this question had already been settled; clearly it has not. Let's discuss our policy somewhere else.

pathawks:
@jekyllbot: merge +docs

PR #7117, opened 2018-07-08
post_url: Loosen up post name regex

perlun:
## Background

When using the [jekyll-multiple-languages-plugin](https://github.com/Anthony-Gaudino/jekyll-multiple-languages-plugin) to be able to publish a blog in multiple languages, the `relative_path` changes semantics. Instead of being `_posts/2017-09-26-more-memory-leaks-jvm-and-jruby`, it will now be `_i18n/en/_posts/2017-09-26-more-memory-leaks-jvm-and-jruby` - each generated "site" will have its subdirectory below the `_i18n` directory.

## The problem

Unfortunately, this breaks the regex being modified in this PR. It makes the assumption that `_posts` is indeed the top-level folder name. The `post_url` tag will still _work_, but it will print out the dreaded deprecation warning for each blog post being linked to.

## The suggested solution

The suggested approach taken in this PR is the "simplest way possible". By loosening up the regex, taking away the "start-of-string" ^ character, everything works fine - no deprecation warning any more. And, all tests still pass.

Are you fine with this or would you want something fancier? What I like about the approach taken here is that it really feels like an application of the KISS principle to me. As simple as possible, but no simpler than that.

### DIFF
--- a/lib/jekyll/tags/post_url.rb
+++ b/lib/jekyll/tags/post_url.rb
@@ -17,7 +17,7 @@ module Jekyll
         end
 
         escaped_slug = Regexp.escape(slug)
-        @name_regex = %r!^_posts/#{path}#{date}-#{escaped_slug}\.[^.]+|
+        @name_regex = %r!_posts/#{path}#{date}-#{escaped_slug}\.[^.]+|
           ^#{path}_posts/?#{date}-#{escaped_slug}\.[^.]+!x
       end
 


ashmaroli:
@perlun I've always been against changing Core logic just to suit the shenanigans of a third-party plugin.
If you're able to demonstrate that the `post_url` tag doesn't work well in a site (without third-party plugins) where the posts are not at the top-level, e.g. at `en/random/_posts/2018-07-09-whatever.md` then its a valid case for consideration. Otherwise, its upto the plugin to *override the concerned code* (regex in this case) as necessary.

pathawks:
> Otherwise, its upto the plugin to override the concerned code (regex in this case) as necessary.

I think this is the correct path forward here. That plugin does some very unconventional things, building the site multiple times without each build knowing about the others. I think that plugin should patch the tag with the known correct path so that checking can remain as strict as possible.

Loosening the regex could break things, and provides no gain to most users.

perlun:
@ashmaroli 

> If you're able to demonstrate that the post_url tag doesn't work well in a site (without third-party plugins) where the posts are not at the top-level, e.g. at en/random/_posts/2018-07-09-whatever.md then its a valid case for consideration.

I'm fairly convinced that it does; will see if I can find some time to dig deeper into this (vacation and not so much hacking time at the moment).

@pathawks 

> > Otherwise, its upto the plugin to override the concerned code (regex in this case) as necessary.
> 
> I think this is the correct path forward here. That plugin does some very unconventional things, building the site multiple times without each build knowing about the others. I think that plugin should patch the tag with the known correct path so that checking can remain as strict as possible.

I can buy into this thinkingl; I understand what you mean. I looked briefly and the `jekyll-multiple-languages-plugin` project unfortunately seems rather dead, so I'm not sure it's worth trying to send a PR there unfortunately. If so, I guess someone would have to fork it and maintain it.

----

(Somewhat off-topic: if someone reading this knows of a good multi-language plugin for Jekyll that plays well with Jekyll core, please speak up - I'm open for suggestions. If it can be solved without the need for a plugin, even better - I'm all ears.)

oe:
@perlun i18n support is kind of sort of on the 4.0 roadmap-ish!

pathawks:
Closing this for now. Hopefully we can get to the root of the problem with better i18n support soon :+1:

perlun:
Thanks @pathawks and @oe. I filed a PR towards the plugin for now (https://github.com/Anthony-Gaudino/jekyll-multiple-languages-plugin/pull/126), will use that hwile joyfully anticipating 4.0 to hit the scene, some day. :tada:

PR #7119, opened 2018-07-09
Use .md for welcome post template

sonicdoe:
Both [`index.md`](https://github.com/jekyll/jekyll/blob/v3.8.3/lib/site_template/index.md) and [`about.md`](https://github.com/jekyll/jekyll/blob/v3.8.3/lib/site_template/about.md) use the `.md` extension, so this makes it more consistent.

For example: `2018-07-09-welcome-to-jekyll.markdown` → `2018-07-09-welcome-to-jekyll.md`


ashmaroli:
Related #6938

sonicdoe:
Thank you for linking to https://github.com/jekyll/jekyll/pull/6938, I wasn’t aware of it.

I’d argue that `.md` is, in fact, more well-known than `.markdown` since a lot of open-source projects have a `README.md` (and even GitHub’s default readme filename is `README.md`).

Nevertheless, I’d be fine with `.markdown`. I’d suggest renaming `index.md` and `about.md` then, though.

ashmaroli:
> I’d suggest renaming `index.md` and `about.md` then, though

Personally, I'm not in favor of using the `.markdown` extension for `index.md` and `about.md` as I prefer the shorter extensions over the ugly verbose ones.
But since you're playing the *consistency* card, I have no comments.. :wink:

pathawks:
Because this is a starter site, I think we should be as verbose as possible. `.markdown` makes it very easy for new users to understand what the file contains, and it is easy for users to rename the files `.md` if they prefer.

oe:
I prefer `.markdown` in this specific case. The filename already has a lot of numbers and dashes, so making at least the file extension verbose is a good thing.

sonicdoe:
Thanks for giving it a thought and making a decision. Would you like to use `.markdown` for the pages then too (the consistency card 🙂)? This was actually the reason why I opened the pull request in the first place: I wondered if there’s a reason pages use a different extension than posts.

pathawks:
@sonicdoe I like the idea of anything created by `jekyll new` having the `.markdown` extension.

For anything else, I don't think it matters so much.

Issue #7120, opened 2018-07-09
Reboot.css


chrisfinazzo:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I ran `jekyll doctor` to check my configuration
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [x] This is a feature request.

---

- [x] I am on (or have tested on) ***macOS*** 10+

---

- [x] I was trying to build.

As Normalize doesn't bring many changes with it from v7 to v8, it seems appropriate to revisit our reset stylesheet. In #6491, @envygeeks [mentioned][] that Bootstrap's Reboot might be worth a look. Furthermore, Bootstrap itself has migrated away from Normalize to use Reboot in their 4.0 release. If desired, I can open a PR to swap it in - thinking this could go in 4.0, but it's up to @oe and @DirtyF to make that call.

[mentioned]: https://github.com/jekyll/jekyll/pull/6491#issuecomment-340347240

That said, there is one issue I've come across that needs to be resolved: TWBS has numerous mixins and other SASS directives which currently trigger a build failure. Importing all of them feels like overkill and hand-tweaking values is no one's idea of a good time.

Thoughts?

cc\ @jekyll/documentation

---

## My Reproduction Steps

After editing the appropriate files, rebuild the site as normal `bundle exec jekyll build`.

<img width="585" alt="screen shot 2018-07-09 at 11 05 47 am" src="https://user-images.githubusercontent.com/1534882/42458763-62979220-8368-11e8-95c9-e8a7f23cb5e4.png">

## The Output I Wanted

A successful build to appear in `_site`.

mattr-:
Does reboot depend on bootstrap or is it a standalone thing? I'm thinking we may need to extract a pure CSS version of Reboot rather than pulling the SASS version in to avoid a bootstrap dependency or upgrade. If any of this is just flat out wrong, I apologize. I haven't taken the time to get up to speed with bootstrap 4.

chrisfinazzo:
Ah, yes, that's correct. It was developed in parallel with Bootstrap, so we'd need to have a compiled version. I've updated the link and the build now succeeds. I'll post a link to the PR so you can take a look.

@DirtyF This was @envygeeks' suggestion, so he's probably better equipped to answer that than I am. I just wanted to see if it was possible so that there would be an alternative if Normalize ceased active development.

DirtyF:
[...]

PR #7121, opened 2018-07-09
Testing the Reboot reset

chrisfinazzo:
Based on discussion in #7120.


docs/_sass/_reboot.scss
```diff
@@ -0,0 +1,8 @@
...
+/*# sourceMappingURL=bootstrap-reboot.min.css.map */
```
pathawks: Are we actually generating a source map somewhere?

DirtyF:
Could you provide a diff image of the change to see how this affects the current website?

Looks like Reboot does more things then Normalize https://codepen.io/ncerminara/embed/RLMwmy?height=500&slug-hash=RLMwmy&default-tabs=html,result&host=https://codepen.io&embed-version=2

docs/_sass/_reboot.scss
chrisfinazzo: Not that I'm aware of.

Does cURL permit something like `sourcemap=none` which would suppress this? I've used it when converting by hand, but I'm not sure it applies to a process like Rake tasks 🤔

chrisfinazzo:
@DirtyF OK, will do.

docs/_sass/_reboot.scss
pathawks: Oh, I see: This is from the pre-generated CSS file that we are downloading.
No, not much we can do about this, I think.

chrisfinazzo: @pathawks Yeah, given how this is structured, it's redundant.

chrisfinazzo:
Sidebar margins and headers seem to be affected - looks like `margin-top` is 0 in a bunch of places and `h2`'s are sized as titles. We would need to override these, so it's not a drop-in replacement at this point.

**Reboot.css**
screen shot 2018-07-09 at 3 26 31 pm

**Normalize.css**
screen shot 2018-07-09 at 3 30 18 pm

DirtyF:
@chrisfinazzo thanks for the feedback. Good to know, another thing to consider is the file size:

- Normalize 5.91 KB 
- Reboot  11.2 KB

Given that Reboot is bigger and needs additional rules, without bringing any gain to the website users the moment, I'll close this for now, we might reconsider our reset when we rewrite CSS, but it's far from our priorities at the moment 😄

PR #7123, opened 2018-07-10
Update quickstart.md

Awjin:
Hi team, had some questions about our audience:

- Who is our main reader?
- Are they non-technical, or do they program?
- Should we explain Bundler, or trust them to handle the abstraction?

And based on answers to the above, should we structure this article any differently?


chrisfinazzo:
Hi @Awjin,

While these are valid questions, and certainly things we've discussed in the past, might I suggest you open an Issue instead to address this? As written, this PR is just a comment in `quickstart.md`.

@jekyll/documentation, please close.

PR #7130, opened 2018-07-12
Scan `assert_equal` methods and rectify any offenses with a custom Rubocop cop

### LINKED ISSUE #7107: assert_equal called with constant as `actual`
In several places in our tests, we call `assert_equal` with `expected` and `actual` swapped.

```ruby
assert_equal @affected.data["key"], "val"
```

I wonder if it would be possible to create a Rubocop cop that complains when the second parameter passed to `assert_equal` is clearly a constant.

ashmaroli:
Closes #7107 

- [x] Implement Cop
- [x] Initial Run with detected offenses &mdash; indirectly test if the cop works
- [x] Correct offenses with `--autocorrect, -a`
- [x] CI Passes 100% &mdash; indirectly test if the autocorrect aspect works properly

### DIFF
--- a/History.markdown
+++ b/History.markdown
@@ -34,6 +34,7 @@
   * add jekyll-xml-source (#7114)
   * Add the jekyll-firstimage filter plugin (#7127)
   * Use a real theme in the example (#7125)
+  * Update docs about post creation (#7138)
 
 ### Minor Enhancements
 
@@ -43,6 +44,7 @@
   * Suggest re-running command with --trace on fail (#6551)
   * Update item_property to return numbers as numbers instead of strings (#6608)
   * Use .markdown for page templates (#7126)
+  * Fix custom 404 page for GitHub pages (#7132)
 
 ### Major Enhancements
 
@@ -70,10 +72,16 @@
   * Update rubocop version to 0.57.x ### -docs (#7078)
   * Example of CircleCI deployment through CircleCI v2 (#7024)
   * Fix Rubocop offences in test files (#7128)
+  * fix up refute_equal call (#7133)
+  * Fix incorrectly passed arguments to assert_equal (#7134)
 
 ### Bug Fixes
 
   * Add call to unused method `validate_options` in `commands/serve.rb` (#7122)
+
+### feature
+
+  * Disable Liquid via front matter (#6824)
 
 ## 3.8.3 / 2018-06-05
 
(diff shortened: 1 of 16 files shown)


kenman345:
Should we extend this for other assert statements and refute statements? 

I.e. `refute_match`, `assert_match`

ashmaroli:
> Should we extend this for other assert statements and refute statements?
> 
> I.e. `refute_match`, `assert_match`

Do they have similar API? i.e., are they defined as `assert_match(expected, actual, msg=nil)
(I'll have to check)

kenman345:
I will look but we also have `refute_equal` that most certainly follows the same pattern

ashmaroli:
Apparently, they do have similar or same API..
https://apidock.com/ruby/MiniTest/Assertions/assert_match

ashmaroli:
I dunno if we should have a Cop for every definition.. Might as well have a dedicated `gem` for it then.. :stuck_out_tongue:

kenman345:
I think the idea of the cop is to ensure that we are consistent with a standard so at least refute_equal should follow for this cop and I’m indifferent about a secondary cop for matches assert or refutes if that’s deemed worthy of also being enforced

ashmaroli:
Correct. And I agree completely.
`refute_equal` can be checked by extending this PR's cop a bit.

But all of that is beyond my authority. I'll defer to the Core maintainers
/cc @jekyll/core

DirtyF:
> I dunno if we should have a Cop for every definition.. Might as well have a dedicated gem for it then.. 

A `rubocop-minitest` gem would be nice 🤖 💎

ashmaroli:
> A `rubocop-minitest` gem would be nice

my thoughts exactly..

kenman345:
Okay but does that mean do not merge this in and/or make the adjustments to make this work for all assert and refute methods that follow that pattern?

kenman345:
Also note: [assert_same](https://apidock.com/ruby/v1_9_3_392/MiniTest/Assertions/assert_same) and [refute_same](https://apidock.com/ruby/v1_9_3_392/MiniTest/Assertions/refute_same) also have the same pattern as is in this cop.

If you want to make a gem for this stuff, I'm all for it. I would love to try and help with that but I think an autocorrect function from this cop handling these other methods as well could help clean up this repository in the meantime.

DirtyF:
@jekyllbot: +dev

PR #7132, opened 2018-07-13
Fix custom 404 page for GitHub pages

ScottKillen:
Github Pages was not using the custom 404 page on my site.

I implemented the attached fix per [Github](https://help.github.com/articles/creating-a-custom-404-page-for-your-github-pages-site/) and now all is as it should be.

### DIFF
--- a/lib/site_template/404.html
+++ b/lib/site_template/404.html
@@ -1,4 +1,5 @@
 ---
+permalink: /404.html
 layout: default
 ---
 


ashmaroli:
At first glance, I think this change is unnecessary because a `./404.html` page will always get output into `./_site/404.html`. @ScottKillen Can you tell me a few cases where the output page won't be at `./_site/404.html`?

ScottKillen:
I created a new site with `jekyll new` and 404 worked fine locally on jekyll server. When I pushed to GitHub, 404 did not work from hosted site. I then googled and found this directive at GitHub help. I do not think that this has anything to do with a problem in jekyll...as I said the local server works fine; however,  GitHub only sees my 404 page if I follow their instructions to add this directive.

I should also add for completeness that I am hosting on GitHub with a custom domain should you wish to try and duplicate.

ashmaroli:
Thanks for getting back. I'm fine with this change if it reduces a lot of confusion when hosting on `github-pages`

DirtyF:
@jekyllbot: merge +minor

PR #7133, opened 2018-07-13
fix up refute_equal call

kenman345:
Some of the `assert_equal` calls have been fixed in othe PRs but this is a `refute_equal` that was sending the actual first and then the expected, which is against the expected pattern and might as well be cleaned up.

### DIFF
--- a/test/test_collections.rb
+++ b/test/test_collections.rb
@@ -87,7 +87,8 @@ class TestCollections < JekyllUnitTest
     end
 
     should "contain only the default collections" do
-      refute_equal @site.collections, {}
+      expected = {}
+      refute_equal expected, @site.collections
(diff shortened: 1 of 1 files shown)


kenman345:
I basically pulled down @ashmaroli 's PR branch and tried my hand at extending the new cop to handle the other minitests that match that pattern and the autocorrect as well. 

While that PR is not yet merged and its still not determined if it will or if it should include those types of changes, I figured we could at least take advantage of the correction it would make.

DirtyF:
Good catch @kenman345 👍 

@jekyllbot: merge +dev

pathawks:
Thanks, @kenman345 🎉🍻

PR #7136, opened 2018-07-15
Optimize rendering Liquid templates

### LINKED ISSUE #7135: Includes are parsed multiple times in a build session
## Steps to reproduce this bug

  - Edit the following code by inserting calls to `jekyll.logger`:
https://github.com/jekyll/jekyll/blob/16e9820dac7c29b5ad835ec7ca1023301f7963d0/lib/jekyll/tags/include.rb#L149-L155

    ```diff
               if cached_partial.key?(path)
    +            Jekyll.logger.info "Cached Include:", path.green
                 cached_partial[path]
               else
    +            Jekyll.logger.warn "Parsed Include:", path
    ```

- Create a new jekyll source with `jekyll new` and edit the `Gemfile` to point to the Jekyll version that contains the above change.
- Run `bundle exec jekyll build`

## The Output I Wanted

I expected to see a mix of `Cached Include: <path>` (green) and `Parsed Include: <path>` (yellow) output in my terminal.

## The Output I Got

Just a stream of `Parsed Include: <path>` (yellow)

ashmaroli:
Fixes #7135 
by caching parsed templates and then rendering those templates non-destructively

### DIFF
--- a/docs/_docs/upgrading/3-to-4.md
+++ b/docs/_docs/upgrading/3-to-4.md
@@ -19,9 +19,32 @@ If you're using Ruby >= 2.3.0, go ahead and fetch the latest version of Jekyll:
 gem update jekyll
 ```
 
----
 
-  *Insert sections here*
+### Template rendering
+
+We've slightly altered the way Jekyll parses and renders your various templates to improve
+the overall build times. Jekyll now parses a template once, caches it internally and then
+renders the parsed template multiple times as required by your pages and documents.
+
+The downside to this is that some of the community-authored plugins may not work as they
+previously used to.
+
+#### For Plugin-authors
+
+* If your plugin depends on the following code: `site.liquid_renderer.file(path).parse(content)`,
+note that the return value (`template`, an instance of *`Liquid::Template`*), from that line will
+always be the **same object** for a given `path`. <br/>
+The *`template`* instance is then rendered as previously, with respect to the `payload` passed to it.
+You'll therefore have to ensure that *`payload`* is not memoized or cached in your plugin instance.
+
+* If its a requirement that `template` you get from the above step *be different* at all times,
+you can invoke *`Liquid::Template`* directly:
+
+
+  ```diff
+  - template = site.liquid_renderer.file(path).parse(content)
+  + template = Liquid::Template.parse(content)
+  ```
 
 ---
 
--- a/lib/jekyll/liquid_renderer.rb
+++ b/lib/jekyll/liquid_renderer.rb
@@ -18,6 +18,7 @@ module Jekyll
 
     def reset
       @stats = {}
+      @cache = {}
     end
 
     def file(filename)
@@ -50,10 +51,18 @@ module Jekyll
       "#{error.message} in #{path}"
     end
 
+    # A persistent cache to store and retrieve parsed templates based on the filename
+    # via `LiquidRenderer::File#parse`
+    #
+    # It is emptied when `self.reset` is called.
+    def cache
+      @cache ||= {}
+    end
+
     private
 
     def filename_regex
-      @filename_regex ||= %r!\A(#{source_dir}/|#{theme_dir}/|\W*)(.*)!i
+      @filename_regex ||= %r!\A(#{source_dir}/|#{theme_dir}/|/*)(.*)!i
     end
 
     def new_profile_hash
--- a/lib/jekyll/liquid_renderer/file.rb
+++ b/lib/jekyll/liquid_renderer/file.rb
@@ -10,8 +10,9 @@ module Jekyll
 
       def parse(content)
         measure_time do
-          @template = Liquid::Template.parse(content, :line_numbers => true)
+          @renderer.cache[@filename] ||= Liquid::Template.parse(content, :line_numbers => true)
         end
+        @template = @renderer.cache[@filename]
 
         self
       end
@@ -24,6 +25,7 @@ module Jekyll
         end
       end
 
+      # This method simply 'rethrows any error' before attempting to render the template.
       def render!(*args)
         measure_time do
           measure_bytes do
--- a/lib/jekyll/tags/include.rb
+++ b/lib/jekyll/tags/include.rb
@@ -90,13 +90,7 @@ module Jekyll
 
       # Render the variable if required
       def render_variable(context)
-        if @file =~ VARIABLE_SYNTAX
-          partial = context.registers[:site]
-            .liquid_renderer
-            .file("(variable)")
-            .parse(@file)
-          partial.render!(context)
-        end
+        Liquid::Template.parse(@file).render(context) if @file =~ VARIABLE_SYNTAX
       end
 
       def tag_includes_dirs(context)
--- a/lib/jekyll/tags/link.rb
+++ b/lib/jekyll/tags/link.rb
@@ -17,9 +17,7 @@ module Jekyll
 
       def render(context)
         site = context.registers[:site]
-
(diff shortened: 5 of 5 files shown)


pathawks:
Can you explain a bit about what was wrong with the previous approach, and how this fixes it?

ashmaroli:
> Can you explain a bit about what was wrong with the previous approach, and how this fixes it?

Earlier, each call to `template = site.liquid_renderer.file(path).parse(content)` would parse the `content` and return a *new* template instance which is then rendered in place by a subsequent `render!`.

With the proposed change, `template` is populated with a previously parsed `content` and is subsequently rendered non-destructively with `render` thereby allowing the same template to be rendered multiple times.

This is a `breaking-change` for plugins like `jekyll-seo-tag` that relies on receiving new instances of say, `_includes/head.html` every time..

pathawks:
Thanks for explaining; that all makes good sense.

Can you point to where this breaks `jekyll-seo-tag` in that codebase? That plugin doesn't rely on the `{% include %}` tag at all, so I think I'm missing something here.

pathawks:
> |ref | build time in seconds|
> |:-- | --: |
> | #&zwnj;7136 | 267.73 |
> | `master` | 334.20 |

Woah 😻

DirtyF:
This currently breaks our tests and as expected all websites pages inherit the very same metadata when using `{% seo %}` 😕

ashmaroli:
> Are you saying that, if an include contains `{% seo %}`, it will get rendered once and cached for that include? That's not great.

@pathawks Nope, I ain't sayin that *include* will get rendered and cached. What I'm saying is, **there will be just one instance of `Jekyll::SeoTag`**
So, to work with this proposal, the plugin **will have to make the following change:**
```diff

  def render(context)
    @context = context
-   SeoTag.template.render!(payload, info)
+   SeoTag.template.render(payload, info)
  end

[...]

+ # The `drop` should not be cached since there is going to be just
+ # one instance of this class per `{% seo %}`
+ # i.e., if you're going to use `{% seo %}` in two templates that are
+ # collectively used by 50 documents, there's just going to be
+ # **2 instances of this class** instead of a **100**.
  def drop
-   @drop ||= Jekyll::SeoTag::Drop.new(@text, @context)
+   Jekyll::SeoTag::Drop.new(@text, @context)
  end
```

ashmaroli:
@DirtyF The failing tests have been taken care of and rogue commits squashed. All that's left now is to document the breaking-change.

pathawks:
I wish Utterson had the ability to diff two builds of a site to make sure there were no unintentional chances.

https://github.com/jekyll/Utterson/issues/19

pathawks:
I committed @ashmaroli's changes to [the `no-cache-drop` branch](https://github.com/jekyll/jekyll-seo-tag/commit/82204a6e31136466510eb99083988ec3c6f3784e) of **jekyll-seo-tag** for some testing. It looks like it makes things just a bit slower when used with the current version of Jekyll, but might be worth it for this performance boost.

| `jekyll 3.8.3 + jekyll-seo-tag 2.5.0`    |                335.19 |
| `jekyll 3.8.3 + jekyll-seo-tag no-cache-drop` |                337.11 |
| `jekyll optimize-liquid-rendering + jekyll-seo-tag no-cache-drop` |                267.04 |

ashmaroli:
> It looks like it makes things just a bit slower

yeah., I had seen that locally as well. I came up with a workaround for that:
```ruby
    def drop
      if context.registers[:site].liquid_renderer.respond_to?(:cache)
        Jekyll::SeoTag::Drop.new(@text, @context)
      else
        cached_drop
      end
    end

    def cached_drop
      @cached_drop ||= Jekyll::SeoTag::Drop.new(@text, @context)
    end
```

kenman345:
It seems like a few sites might have an issue with this and the plugins they use. Perhaps we can implement the caching to happen so its caching seo-tag once per site and not once per machine? I havent looked over the code but while this would go into a major version release it does feel like something that a ton of plugins will have to get updates to work properly and that can greatly reduce the rollout of the latest and greatest if it breaks a lot of plugins that users then need to wait on a fix that they might not fully understand. If nothing else, we should be able to make a new `jekyll` command line tool, or note one that already exists, that will help clear out the cache so they can force cleanout of the cache locally.

pathawks:
> it does feel like something that a ton of plugins will have to get updates to work properly

If we had a tool that could do regression tests, it would allow us to identify which plugins might need updating, and we could just open PRs on those plugins before 4.0 is released. This way, popular plugins are ready to go from day one, and we would have detailed instructions on our website for anybody who uses a custom plugin.

kenman345:
I like that, I would be happy to try and help with that. 

I looked it up and `jekyll clean` should work to clean this up, no? If not, maybe we extend that to help those out while they wait for new gems to be released for the less common gems?

ashmaroli:
@kenman345 
> once per site and not once per machine

The `cache` persists only for a given `site` instance and gets emptied when the site *regenerates*.


> note one that already exists, that will help clear out the cache

To clear the cache out, all the user would need to do, is to get the site to regenerate.

kenman345:
That makes sense. Thanks for making that clear for me

pathawks:
The [`{% github_edit_link %}` tag](http://jekyll.github.io/github-metadata/edit-on-github-link/) will need [updating](https://github.com/jekyll/github-metadata/blob/47855c340dc4ac4e7c7b4db7ab6349c0286ca50e/lib/jekyll-github-metadata/edit-link-tag.rb#L84) for the same reason as `{% seo %}`.

ashmaroli:
> The `{% github_edit_link %}` tag will need updating for the same reason as `{% seo %}`.

Yes, it will need updating too..

DirtyF:
Could this change affect popular plugins like jekyll-assets?

pathawks:
There is one site that builds correctly with this PR, but fails in very strange ways when build with current Jekyll `master`. Investigating.

ashmaroli:
> not for helping plugin and theme developers with keeping in line with recommendations

@kenman345 Does the README at `rubocop-jekyll` really say that..?

> Rubocop-Jekyll could be used to have a cop that identifies and warns against the now breaking pattern and an autocorrect method to help rewrite it properly.

`rubocop-jekyll` is mainly a **Linter**. It is not a plugin that extends Jekyll's functionality.
Detecting breaking changes or notifying about them should ideally be handled by the classes `Jekyll::Commands::Doctor` or `Jekyll::Deprecator`.

That said, I like the idea of detecting breaking changes programmatically.

kenman345:
I’ll take a look next time I have a moment free and see what it might take
to extend those classes for the changes. I agree it makes sense that it be
inside Jekyll itself if anything since it guarantees people have it and
it’s tightly coupled to the version installed too

kenman345:
@ashmaroli I looked it over and `Jekyll::Deprecator` is definitely not the right area, and `jekyll doctor` seems to be more mass-user centric, not at all focusing on any installed plugins/gems that might affect experience/the code that goes into them. 

If anything were to be added to Jekyll commands that would help plugin creators, it would seem that it has no home right now. :-(

That said, does Jekyll have any utils/plugins widely used by plugin creators that we could add to?

ashmaroli:
@pathawks Any updates on this..? :hourglass_flowing_sand:
(Migration to the Cache API will be in a future PR only)

DirtyF:
>  Migration to the Cache API will be in a future PR only

So should we close this PR and wait for the proper implementation?

DirtyF:
@jekyllbot: merge +minor

Issue #7113, opened 2018-07-06
Feature Request: Collections of Yaml files to be output as pages


kenman345:
I was trying to use a collection of yaml files to output a bunch of data. Each file is representative of a set of data I want on a given page. Right now I have a rake task to make a `.md` file in a secondary collection with the `output: true` setting just to then set the data by first looking up from my original set of data. I was wondering if we can enable functionality wherein when `output: true` is defined for a collection and the collection has a `.yml` file, that it considers it like frontmatter is, and makes it into a page with the defaults defined for the type of page if not present.

kenman345:
Can I get a little love here? I understand my issue may not warrant an update but instead be a plugin but I was hoping to get some feedback and guidance from a @Jekyll/Core Team member to figure out the approach I should take

ashmaroli:
> Can I get a little love here?

I'm having a hard time understanding your `feature-request`.  Perhaps you could try simplifying your request application by stating the following in as simple sentences as possible:
  - What feature are you requesting for?
  - Why do you think this feature would benefit more than 60% of our users
  - If I start with a boilerplate site, what steps at minimum, should I take to use this feature?
  - What is your expected result?

pathawks:
I don't understand the issue. If you rename the `.yml` files to `.html` does it output the pages as you would like?

kenman345:
@pathawks I tried changing the .yml extension of the files to .html and it does not output properly either.

Basically, I take .yml files, I set the `_config.yml` to have a collection for the folder they are on with the settings:
```yml
collections:
  artists:
    output: true
    layout: artist
```

The output if I leave it as a .yml file is that I get a blank .html file (0 kb). 
If I change it to .html, then the output is literally just of the content of the `yaml` file, not generated using any of my layouts that the `artist` layout inherits from, or any elements that `artist` contains.

kenman345:
My project is a personal one. I want to be able to use my NAS and a Ruby script to generate an up to date set of data to this projects proper folder so that I can use YAML to represent my current set of music collection metadata. 

I have most things set up but I currently have been having rake copy my `_data` files to my collection folder and append `---` to them when I generate the site. I would much rather have Jekyll assume the frontmatter since my YML files all are output by scripts, which means they start with 
```yaml
---
```
at the top. So its annoying that its not assumed to be frontmatter.

[...]

PR #7147, opened 2018-07-17
Treat YAML files in collections the same as .md files, and render them when output => true

kenman345:
Just getting some visibility while I continue to work on this. I have not gotten this to the point where it works with what I am trying to do but the basis is the request I made in #7113 and the suggested starting point I was directed to take to accomplish the change.

### DIFF
--- a/lib/jekyll/collection.rb
+++ b/lib/jekyll/collection.rb
@@ -58,7 +58,7 @@ module Jekyll
       filtered_entries.each do |file_path|
         full_path = collection_dir(file_path)
         next if File.directory?(full_path)
-        if Utils.has_yaml_header?(full_path) || read_yaml_file?(full_path)
+        if Utils.has_yaml_header? full_path
           read_document(full_path)
         else
           read_static_file(file_path, full_path)
@@ -202,10 +202,6 @@ module Jekyll
     end
 
     private
-
-    def read_yaml_file?(full_path)
-      write? && %w(.yaml .yml).include?(File.extname(full_path))
-    end
 
     def container
       @container ||= site.config["collections_dir"]
--- a/lib/jekyll/collection.rb
+++ b/lib/jekyll/collection.rb
@@ -58,7 +58,7 @@ module Jekyll
       filtered_entries.each do |file_path|
         full_path = collection_dir(file_path)
         next if File.directory?(full_path)
-        if Utils.has_yaml_header? full_path
+        if Utils.has_yaml_header?(full_path) || read_yaml_file?(full_path)
           read_document(full_path)
         else
           read_static_file(file_path, full_path)
@@ -202,6 +202,10 @@ module Jekyll
     end
 
     private
+
+    def read_yaml_file?(full_path)
+      write? && %w(.yaml .yml).include?(File.extname(full_path))
+    end
 
     def container
       @container ||= site.config["collections_dir"]
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -132,7 +132,7 @@ module Jekyll
 
     # Determine whether the document is a YAML file. And if it is meant to be written.
     #
-    # Returns true if the file should be output and 
+    # Returns true if the file should be output and
     #   extname is either .yml or .yaml, false otherwise.
     def render_yaml_file?
       yaml_file? && write?
@@ -168,7 +168,8 @@ module Jekyll
     #   true otherwise.
     def render_with_liquid?
       return false if data["render_with_liquid"] == false
-      !(coffeescript_file? || yaml_file? || !Utils.has_liquid_construct?(content)) || render_yaml_file?
+      !(coffeescript_file? || yaml_file? || !Utils.has_liquid_construct?(content))\
+      || render_yaml_file?
     end
 
     # Determine whether the file should be rendered with a layout.
@@ -274,9 +275,9 @@ module Jekyll
         @data = SafeYAML.load_file(path)
       else
         begin
-          merge_defaults 
+          merge_defaults
           if yaml_file?
-            read_yml_content()
+            read_yml_content
           else
             read_content(opts)
           end
@@ -459,7 +460,7 @@ module Jekyll
       end
     end
 
-    def read_yml_content()
+    def read_yml_content
       data_file = SafeYAML.load_file(path)
       self.content = data_file["content"] || ""
       merge_data!(data_file, :source => "YAML front matter")
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -169,7 +169,7 @@ module Jekyll
     def render_with_liquid?
       return false if data["render_with_liquid"] == false
       !(coffeescript_file? || yaml_file? || !Utils.has_liquid_construct?(content))\
-      || render_yaml_file?
+      || (render_yaml_file? && Utils.has_liquid_construct?(content))
     end
 
     # Determine whether the file should be rendered with a layout.
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -130,6 +130,14 @@ module Jekyll
       %w(.yaml .yml).include?(extname)
     end
 
+    # Determine whether the document is a YAML file. And if it is meant to be written.
+    #
+    # Returns true if the file should be output and 
+    #   extname is either .yml or .yaml, false otherwise.
+    def render_yaml_file?
+      yaml_file? && write?
+    end
+
     # Determine whether the document is an asset file.
     # Asset files include CoffeeScript files and Sass/SCSS files.
     #
@@ -160,7 +168,7 @@ module Jekyll
     #   true otherwise.
     def render_with_liquid?
       return false if data["render_with_liquid"] == false
-      !(coffeescript_file? || yaml_file? || !Utils.has_liquid_construct?(content))
+      !(coffeescript_file? || yaml_file? || !Utils.has_liquid_construct?(content)) || render_yaml_file?
     end
 
     # Determine whether the file should be rendered with a layout.
@@ -175,7 +183,7 @@ module Jekyll
     # Returns false if the document is set to `layouts: none`, or is either an
     #   asset file or a yaml file. Returns true otherwise.
     def place_in_layout?
-      !(asset_file? || yaml_file? || no_layout?)
+      !(asset_file? || yaml_file? || no_layout?) || render_yaml_file?
     end
 
     # The URL template where the document would be accessible.
@@ -262,12 +270,16 @@ module Jekyll
     def read(opts = {})
       Jekyll.logger.debug "Reading:", relative_path
 
-      if yaml_file?
+      if yaml_file? && !render_yaml_file?
         @data = SafeYAML.load_file(path)
       else
         begin
-          merge_defaults
-          read_content(opts)
+          merge_defaults 
+          if yaml_file?
+            read_yml_content()
+          else
+            read_content(opts)
+          end
           read_post_data
         rescue StandardError => e
           handle_read_error(e)
@@ -445,6 +457,12 @@ module Jekyll
         data_file = SafeYAML.load(Regexp.last_match(1))
         merge_data!(data_file, :source => "YAML front matter") if data_file
       end
+    end
+
+    def read_yml_content()
+      data_file = SafeYAML.load_file(path)
+      self.content = data_file["content"] || ""
+      merge_data!(data_file, :source => "YAML front matter")
     end
 
     def read_post_data
--- a/History.markdown
+++ b/History.markdown
@@ -36,6 +36,7 @@
   * Use a real theme in the example (#7125)
   * Update docs about post creation (#7138)
   * Add DEV Community&#39;s Jekyll tag to community page (#7139)
+  * Initialize upgrading doc for v4.0 (#7140)
 
 ### Minor Enhancements
 
--- a/History.markdown
+++ b/History.markdown
@@ -35,6 +35,7 @@
   * Add the jekyll-firstimage filter plugin (#7127)
   * Use a real theme in the example (#7125)
   * Update docs about post creation (#7138)
+  * Add DEV Community&#39;s Jekyll tag to community page (#7139)
 
 ### Minor Enhancements
 
--- a/docs/_docs/upgrading/3-to-4.md
+++ b/docs/_docs/upgrading/3-to-4.md
@@ -1,0 +1,28 @@
+---
+title: Upgrading from 3.x to 4.x
+permalink: /docs/upgrading/3-to-4/
+---
+
+Upgrading from an older version of Jekyll? A few things have changed in Jekyll 4
+that you'll want to know about.
+
+Before we dive in, you need to have at least Ruby 2.3.0 installed. Run the following
+in your terminal to check
+
+```sh
+ruby -v
+```
+
+If you're using Ruby >= 2.3.0, go ahead and fetch the latest version of Jekyll:
+
+```sh
+gem update jekyll
+```
+
+---
+
+  *Insert sections here*
+
+---
+
+*Did we miss something? Please click "Improve this page" above and add a section. Thanks!*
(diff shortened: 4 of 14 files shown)


ashmaroli:
For starters, `Jekyll::Document` does not `include Convertible`. So no point trying to meddle there.

ashmaroli:
> basis is the request I made in `#7113` and the suggested starting point I was directed to take to accomplish the change.

If you can achieve what you're looking for via an alternate route, by all means, please do..

kenman345:
Are you saying convertible is the thing not to change?

Also, as mentioned in that feature request it appears that part of this functionality is there. 

I’ll try and see about making this a plugin if it’s deemed too aggressive to be a core feature. It seems a bit silly to need to make a plugin for something that’s so close to being fully supported and a new version of Jekyll is in the works, allowing for such an enhancement to make sense depending on how much effort the actual change takes

ashmaroli:
> Are you saying convertible is the thing not to change?

Yes. `Convertible` is a `module` that is used by `Jekyll::Page` and `Jekyll::Layout` . It is not used by `Jekyll::Document` at all.

asteroidsbg:
Cccccccccccc ccccccccOh"&I*&(&c(&*"u3uoli)o2*&u2cc;*(h(&²I"u2

On Jul 17, 2018 20:37, "Ken Salomon" <notifications@github.com> wrote:

Are you saying convertible is the thing not to change?

Also, as mentioned in that feature request it appears that part of this
functionality is there.

I’ll try and see about making this a plugin if it’s deemed too aggressive
to be a core feature. It seems a bit silly to need to make a plugin for
something that’s so close to being fully supported and a new version of
Jekyll is in the works, allowing for such an enhancement to make sense
depending on how much effort the actual change takes

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/jekyll/jekyll/pull/7147#issuecomment-405664896>, or mute
the thread

ashmaroli:
> if it’s deemed too aggressive to be a core feature

Its not too aggressive for the Core. Its just that there hasn't been any demand for it.. So the maintainers might as well dedicate their time and effort on important things

kenman345:
Working to add in the unit tests. These look to be the last changes I need to make to the `lib` files. 

@ashmaroli Take a look at the changes I made. I do not think its asking much to keep this maintained, nor would I ask of anyone that they have to maintain it. I would happily help maintain the code.

Also of note, the change I made adds a `render_yaml_file?` method which if decided to extend this functionality in the future, could be renamed to something more generic and the return be combined with additional `||` statements to make adding some other file to this somewhat simple. 

I may need some help getting all the tests but I think the documentation on making enhancements will get me a good amount further than I am now (considering I havent added anything yet...)

lib/jekyll/document.rb
```diff
@@ -262,12 +270,16 @@ def published?
...
-      if yaml_file?
+      if yaml_file? && !render_yaml_file?
         @data = SafeYAML.load_file(path)
       else
         begin
-          merge_defaults
-          read_content(opts)
+          merge_defaults 
+          if yaml_file?
+            read_yml_content()
```
ashmaroli: **Tip:** This line isn't going to be evaluated because `ln 275` above has already established that the current document is not a YAML File

ashmaroli:
> I do not think its asking much to keep this maintained

Neither do I.. The reason I let you tackle this issue is because the "diff" I had in mind and suggested towards was very small and easy for a Jekyll Codebase Novice.

**Tip: Start small, expand once established.**
First implement just the minimum change required to get a rendered collection full of YAML files with
`---` at the top of the files. Write tests for them.
Then proceed to include YAML files without the `---` at the top

lib/jekyll/document.rb
kenman345: This does hit as the if statement is evaluating if it is a yaml file but not to be written. We then only get into that else if the yaml file should be written or its not a yaml file at all.

lib/jekyll/collection.rb
```diff
@@ -58,7 +58,7 @@ def read
       filtered_entries.each do |file_path|
         full_path = collection_dir(file_path)
         next if File.directory?(full_path)
-        if Utils.has_yaml_header? full_path
+        if Utils.has_yaml_header?(full_path) || read_yaml_file?(full_path)
```
kenman345: This is something I had a concern about. I wanted to ensure taht yml files being processed was not just for yaml files with `---` at the top, but what I've seen a lot of yaml files be, which is just the data. 

I am thinking though that perhaps since we want the data to be read to the `site.data` that we might want to eliminate where I check below if its meant to be written, and instead just check that its a yml file.

lib/jekyll/document.rb
```diff
@@ -262,12 +271,16 @@ def published?
     def read(opts = {})
       Jekyll.logger.debug "Reading:", relative_path
 
-      if yaml_file?
+      if yaml_file? && !render_yaml_file?
```
kenman345: could change `!render_yaml_file?` to `!write?` so it makes more sense that we get to the else statement.

In the case that we make it to the `else` statement, we merely only need to check that we have a yaml file to know if we should process it properly since this line already determines that we cannot get to the `else` without expecting to read and merge the data with defaults.

ashmaroli:
@kenman345 Please slow down and take baby steps. I'll not be able to help you if you're going to have your fingers into many pies at once.
I want you to focus on just one specific use-case:

```gherkin
Given that I have a directory "_movies"
And I have "_movies/avengers.yaml" generated by an external script
And I have "_movies/iron-man.yml" generated by an external script
And I have "collections: ['movies']" in my config file
When I run jekyll build
Then "_site/movies/avengers.html" should exist
Then "_site/movies/iron-man.html" should exist
```

*Hint: YAML files generated by an external script has `---` at the beginning*

kenman345:
I believe the hyphens can actually be removed in a script by specifying the
Yaml.dump with options.

I know a lot of people and projects that do not use yaml via script. They
add to yaml by hand and do not know what the hyphens are and even remove
them sometimes.

ashmaroli:
> I believe the hyphens can actually be removed in a script by specifying the Yaml.dump with options

I`m not saying that we shouldn't support this feature for YAML files without the `---`. What i'm saying is first get the feature ready for YAML files with `---`. Then *extend* the feature to support YAML files without `---`

kenman345:
Ah, got it thanks. I will focus my efforts on the feature/step fdefinitions for this first where we assume the yaml has the frontmatter type beginning, and secondly where its a yaml file but not with the hyphens. 

@ashmaroli can you weigh in on my comment about the collection.rb file so I can make the appropriate change if any and update/add in the proper tests and feature definition for it?

ashmaroli:
> can you weigh in on my comment about the collection.rb file..?

Leave `collection.rb` untouched (no diff with respect to `master`) for the first phase.

kenman345:
@ashmaroli so as you can see from my current changes to `collections.feature` and `test_collections.rb`, I have the new files `just_yaml.yaml` and `just_yml.yml` in the `_methods/` folder and updated the feature steps to include them, but with removing the changes to the collection.rb that I should also remove these updates? 

If so, then where would I add in tests and feature definitions for this new feature as no `documents.feature` exists.

ashmaroli:
> where would I add in tests and feature definitions for this new feature as no documents.feature exists

You can create a `yaml_collections.feature`

kenman345:
I am building that now but trying to figure out how deep to dive into it. I am looking at the collections feature definition and trying to only put in definitions of things that prove out that the yaml collection is working in unique ways and some of the core collection features are being made use of still work.

kenman345:
@ashmaroli I seem to have run into an issue where my test scenarios are failing due to .yml being the ending and not .html files. I am thinking my local environment was running with `permalinks: pretty` so I wasnt seeing it. 

Looking over the code, we generally use converters to identify an output extension swap. I was wondering if I can get some help on that since yaml doesnt really need a converter of its own does it?

ashmaroli:
> Looking over the code, we generally use converters to identify an output extension swap

If you look carefully, you'll see that "output extension" is determined by either the document's `converter` or by its `permalink`. So, if your data file has something like `permalink: /foo.html`, then what do you think the outcome will be..?

If you wish to simply have a "default output extension", well, you can add an accessor method for `:output_ext` and then:
```ruby
self.output_ext = ".html" unless permalink
```

--

> if I can get some help on that since yaml doesnt really need a converter of its own does it?

Well, I can't say with certainty that YAML files do not need converters. Depends on your implementation of this feature.
Converters "act upon" `content`.  Does `Jekyll::Document` objects created from your YAML files have "content" or do they just have "data"..?
If they do have "content", what do you plan to do with it? Output it verbatim into the generated HTML file or convert it into HTML..?

These are "design" questions that I cannot answer "for you". Its entirely your call but will decide whether the core maintainers here will accept this PR or not.

kenman345:
I think I figured it out. I can make my own converter but I feel like with the recommendation before that was to set self.content = @data["content"] || "" when pulling in the yml file, I think it makes sense to actually just add the .yml and .yaml extensions to the list of extensions the markdown converter handles. 

I believe this would be in `configuration.rb`:
```ruby
"markdown_ext"        => "markdown,mkdown,mkdn,mkd,md,yaml,yml",
```

Does that sound reasonable? The expectation is if you wanted more than the layouts to take effect, then the `content` in the `data['content']` key would define the values rendered into the {{content}} field of the lowest order layout page. This would be considered markdown. 

If we did it this way, then we would simply need to add to the documentation if you have previously set that value set in your config, that you will need to add .yml and .yaml if you wish to have them output as .html. Otherwise, this functionality would not affect your site at all, since it will still output a blank .yml file today if it existed in a collection meant to be output.

kenman345:
Note: I use the term `we` in the below not assuming a role in the Jekyll team but more as a "I would like to help drive these efforts if deemed relevant/warranting changes" type of thing. 

I have renamed the PR and pushed up my commit will the unit tests and changes discussed. The yaml_collections.feature has been added and covers the enhancement made in this PR. 

I noticed that if we have a non-script generated file for `.yml` file (i.e. it has no `---` on the first line), then it actually is not getting pulled into the collection data,and is treated as a static file. Since the intention of YAML files in your source folder is to use as data or as a generated page (after this is merged), I think we might want to consider that a potential bug and make a different PR to check if the extension is of a yaml file in our collection.rb class to include it, the same as it would be added to the collection data if it were in the `_data` folder.

ashmaroli:
I'm personally :-1: on the changes here.

kenman345:
Would making collections have an Output_data flag be better instead? What
exactly is the thing you have problems with, maybe it can be done better or
perhaps moved to a plugin if this isn’t worth merging

kenman345:
@ashmaroli something I thought about as I was making this PR was that perhaps I might be taking a narrow minded approach to this enhancement. I am wanting it to make it into a .html file but YAML is just a data set thats more readable, but translates directly to its origin syntax of JSON. I was thinking the default approach might be that we actually have .yml files output as .json files when included in a collection meant to be output. If you wish not to include the .yml at all could use a `output_data: false` flag on the collection to indicate not to do that. And if they want to have it generate as HTML then they would be able to set `output_data_ext: .html` with the assumed default being `.json` when empty.

ashmaroli:
@kenman345 I'm not in favor of merging this implementation because:
  - First and foremost, you opted to include `.yaml` and `.yml` into `site.config["markdown_ext"]` which is very counter intuitive and definitely hacky.
  - there's an added complexity to check if the file is a YAML file or if the collection is meant to be written.. I think its a case of over-engineering (and that leads to unforeseen bugs)
  - More of a personal nit than a concern, I do not like how your git commits are not simple modular changes with a short message (less than 50 chars) that imply the underlying intent. It might as well be one single commit so that future-visitors to this page do not wonder what the individual commits mean..

kenman345:
All very fair and valid.

I can remove the site config change in favor of a converter of its own.

The added check could actually be refined, let me take a second look to
making it more readable.

What are your thoughts on my Scenarios I pointed out above? Happy to drop
this apt in favor of a different change but it feels broken in its current
form as per my scenarios

ashmaroli:
Additionally, in my opinion, the current behavior that collection entries that have front matter dashes `---` at the very start are read-in as `Document` objects whereas those without `---` are simply `StaticFile` objects, **is not a bug that needs to be "fixed"**

kenman345:
@ashmaroli I understand the frontmatter thing and the static file aspect but I am saying that we have an inconsistency happening currently. 

The most pertinent is if `output: true` then we get script generated yaml files being made on the output side but empty files. 
The second being unlike in the `_data` folder, non-script generated yaml files arent being processed into the collection as part of its data set.
The third being that the id's can match when you have a `.md` and a  `.yml` file of the same name ion the same folder. Whereas if we had some of these changes, then the data file actually gets merged into the .md files data set. (Of note, this merges except the `[content]` part as its an odd scenario. If this PR is to get merged then I can add in a fix to marge the data together. Its something I only thought about now but similar things happen if you have .md and a .markdown file in the same folder already.)

pathawks:
I'm not opposed to adding a `Converter` for YAML files, but I think we should stay away from all the changes to `lib/jekyll/document.rb`.

If there is no way to add a YAML Converter without adding a lot of complexity in places like `Document` then this might be better as a plugin.

kenman345:
@pathawks thanks for weighing in. I believe a lot of the changes can indeed be accomplished without modifying the document.rb file but! we would definitely need to make some change as the data file currently gets read in without being merged to our site defaults. If we simply keep the logic in place that if its set to output the data file in the collection that we merge in the data, versus just setting @data to the data file with no merge if not outputting the file. Otherwise this change wont work as core functionality.

I am not sure exactly how this works as a plugin. Would a plugin basically handle adding in the converter, that seems easy, but also able to handle things as they are in my PR? Or would it have to implement its own logic thats basically mimicing the documents.rb file?

ashmaroli:
As it stands, we currently assume that the object returned from `SafeYAML.load_file(path)` is a Hash. But YAML files can be of an `Array` or simply `String`s. There is no validation or tests tpo prove otherwise.

kenman345:
I would like to close out this PR in favor of a new proposed means of attacking the problems based on #5208 being properly addressed. If the necessary changes are made to that PR and include making collections handling data files consistently as per `_data` folder has been, then the proposed new PR I would make would be to make data files be merged if the ids are the same. So that an `avengers.json` and `avengers.yml` file both get merged, but also `avengers.md` would be able to be rendered without any `site.movies` traversing, since `avengers.yml` and/or `avengers.json` would be merged into its pages data set. 

This change would allow use of scripts to grab external source data and as directly as possible add it into the mix of data sets for a generated page such that it can easily be swapped out in favor of newer versions of that data set and `.csv` and `.tsv` files would not have to have frontmatter in them in order to be used in a collection and/or generated page when not in `_data` folder. 

Since Jekyll is meant to not force everyone to be a programmer and instead make static sites generate easily, the idea of merging the data sets when ids would match makes sense to me at least to allow users to forego the difficulty in properly traversing the `site.___` model to find the set of data, but instead immediately start using it. 

This proposed solution would also mean if you just have a `.yml` file in a collection to be output, then it would still make the empty `.yml` file output it does now, but if it had a similarly named `.md` file that it would not output the `.yml` file anymore, but instead output the .html of the `.md` file using the data from the frontmatter and from the `.yml` file as if it were in the frontmatter of the `.md` to begin with.

@ashmaroli I think this would make a lot of sense for the Jekyll 4.0.0 release, and it personally makes more sense to me at least to achieve what I was hoping for in this PR without imposing a large change that is convoluted in execution. If I can get some buy in, I would happily put together a PR for fixing up the #5208 PR branch so it can be merged and I can subsequently add in a PR for my own suggestion after that.

kenman345:
Seeing as I managed to create, from what I can tell, is everything thats needed to get the data directory as a collection in my PR #7166 so that it can be merged into #5208, I am closing this PR. 

I am closing it in hopes my other efforts can be reviewed and focused on instead of this one. If #5208 were to get merged, then a lot of effort would be reduced and my PR would be much simpler with less code changed. In effect, the overlap between whats needed to achieve both makes this one quite messy. I would want to revisit such an enhancement only after the more clear and approachable change has been made. Thanks for the advice along the way guys and I shall pick this idea back up at a later time.

PR #7152, opened 2018-07-21
Add support for hyphenation in gem name as structure for plugin gems listed in `_config.yml`

jamesluberda:
This PR is to add Jekyll support for gem-based plugins that follow [Bundler](https://bundler.io/v1.16/man/bundle-gem.1.html)/[RubyGems](https://guides.rubygems.org/name-your-gem/) recommended gem structures as expressed by hyphenated gem names, i.e. a gem named `jekyll-persist` having as its base script `lib/jekyll/persist.rb` (rather than `lib/jekyll-persist.rb`).

Currently, Jekyll's `require` of `_config.yml` listed plugins via  `require_with_graceful_fail` in `lib/jekyll/external.rb` will raise a `LoadError` and (erroneously) declare a gem not installed if its hyphenated name is indicative of the gem's internal structure.

This change adopts Bundler's handling of hyphenated gem names in [`lib/bundler/runtime.rb`](https://github.com/bundler/bundler/blob/master/lib/bundler/runtime.rb), making Jekyll's internal gem support consistent with Bundler's, with both Jekyll and Bundler having the same success or failure in loading hyphenated gems regardless as to whether the hyphens indicate structure or not.

For reference, while the code is modeled on that found in `lib/bundler/runtime.rb`, if another approach is preferred, I would be happy to adapt the change accordingly. Any suggestions as to how to reduce the method length would also be welcome.


ashmaroli:
@jamesluberda This **definitely** needs tests. At a glance, I doubt if the second `LoadError => e` will be called on the second-pass. Having tests clear doubts.

pathawks:
[Failing on jruby](https://travis-ci.org/jekyll/jekyll/jobs/406939064#L1010)

jamesluberda:
Interesting. I see it breaks under JRuby specifically. Runs as expected/intended under MRI. Will give it another go to make it work under both. Any ideas what JRuby is doing differently here, i.e. gotchas I may not be familiar with regarding exception handling?

pathawks:
> Any ideas what JRuby is doing differently here, i.e. gotchas I may not be familiar with regarding exception handling?

I wish I could help you, but I know nothing about JRuby and have done my best to stay away from it.

Here is the line that is failing:
https://github.com/jamesluberda/jekyll/blob/4fbf87d8cca4567caa212c076b0a6caf5acf114c/lib/jekyll/external.rb#L73-L75

I am guessing that JRuby uses a different format for its warning message that does not match `%r!^cannot load such file -- (.+)$!i`

ashmaroli:
@jamesluberda Thanks for adding the test. Now I have another issue. I find it odd that Jekyll would complain the following when the user simply listed `gems: ['foo-bar']`:
```
Dependency Error: Yikes! It looks like you don't have foo/bar or one of its dependencies
installed. In order to use Jekyll as currently configured, you'll need to install this gem.
```

jamesluberda:
@pathawks thanks for the pointer, @ashmaroli thanks for the fix! Regarding the `gems` line, what should be output be? (or shouldn't there have been any?)

ashmaroli:
> Regarding the `gems` line, what should the output be?

@jamesluberda Given that a user has listed `gems: ['foo-bar']` or `plugins: ['foo-bar']` in their config file, they would expect to see the following error message:
```
Dependency Error: Yikes! It looks like you don't have 'foo-bar' or one of its dependencies
installed. [...]
```
instead of
```
Dependency Error: Yikes! It looks like you don't have 'foo/bar' or one of its dependencies
installed. [...]
```

jamesluberda:
@ashmaroli Ah yes, the problem is that when the second error is raised because a hyphen was found and tr'd but the gem is still not loadable, it is of course raised against the new require in which the hyphens have been tr'd into slashes. I did a re-write (uncommitted) before I saw your fix which preserves the name used in the config file for Jekyll's error message (the name referenced by raise remains modified if so), but it doesn't use the drop-through error handling I had originally implemented.

In the version that I originally implemented and you fixed, unfortunately the original `name` var is not available for use in the messaging, at least not that I could see, or manage to pass along in some fashion.

lib/jekyll/external.rb
```diff
@@ -52,24 +52,34 @@ def version_constraint(gem_name)
...
+            else
+              raise
+            end
           end
         end
+      rescue LoadError => e
+        e.message =~ %r! -- (.+)$!i
+        specimen = Regexp.last_match[1]
+        Jekyll.logger.error "Dependency Error:", <<-MSG
+          Yikes! It looks like you don't have #{specimen} or one of its dependencies installed.
```
pathawks: Let's use `name` instead of `specimen` in this line.

`name` should still have the original, potentially hyphenated name of the Gem.

ashmaroli: @pathawks *I changed it to `specimen` because `foo/bar` is ideally not the given `name`*
But hopefully that should change before this PR gets approved.

ashmaroli:
> the original name var is not available for use in the messaging, at least not that I could see, or manage to pass along in some fashion.

@jamesluberda I manged to come up with a solution (locally) by extracting the `Jekyll.logger` call to a separate `private` method. I'll let you do the honors or refine that idea at your discretion.

lib/jekyll/external.rb
pathawks: What I mean is that we should use `name`, with its original value.

lib/jekyll/external.rb
```diff
@@ -52,34 +52,45 @@ def version_constraint(gem_name)
...
         Array(names).each do |name|
           begin
             Jekyll.logger.debug "Requiring:", name.to_s
             require name
-          rescue LoadError
+          rescue LoadError => e
             if name.include?("-")
-              require name.tr("-", "/")
+              begin
+                require name.tr("-", "/")
```
pathawks: Why not `require_with_graceful_fail name.tr("-", "/")` and get rid of the duplication?
I guess we would still have the problem of maintaining the original `name`…

ashmaroli: @pathawks IMO, this route is necessary to first check if `foo-bar` can be required otherwise rescue and try loading `foo/bar` or re-raise otherwise.

Can you post your suggested Ruby code..?

ashmaroli:
@jamesluberda Your commit is exactly what I had come up with.. the only difference was the private method name and the local variables for the exceptions:
```ruby
      def require_with_graceful_fail(names)
        Array(names).each do |name|
          begin
            Jekyll.logger.debug "Requiring:", name.to_s
            require name
          rescue LoadError => exception
            if name.include?("-")
              begin
                Jekyll.logger.debug "Requiring:", name.to_s
                require name.tr("-", "/")
              rescue LoadError => error
                load_error_message(name, error)
                raise Jekyll::Errors::MissingDependencyException, name
              end
            else
              load_error_message(name, exception)
              raise Jekyll::Errors::MissingDependencyException, name
            end
          end
        end
      end

      private

      def load_error_message(plugin_name, exception)
        Jekyll.logger.error "Dependency Error:", <<~MSG
          Yikes! It looks like you don't have #{plugin_name} or one of its dependencies installed.
          In order to use Jekyll as currently configured, you'll need to install this gem.

          The full error message from Ruby is: '#{exception.message}'

          If you run into trouble, you can find helpful resources at https://jekyllrb.com/help/!
        MSG
      end
```

lib/jekyll/external.rb
ashmaroli: @pathawks I like the DRY version. Perhaps we can re-substitute the `/` in the name in the error-message..? Or would that be confusing for people who have configured `plugins: ['foo/bar']`

parkr:
I think we should display both things we try:

```ruby
def require_with_graceful_exit(gems)
  Array(gems).each do |gem|
    begin
      require(gem) rescue LoadError require(gem.tr(“-“, “/“))
    rescue LoadError
      Jekyll.logger.error “Fatal:”, “could not load gem #{gem.inspect} (or #{gem.tr(“-“, “/“)})”
      # ... add helpful output which guides user to check config & Gemfile & run with bundle install, then link out to the website
      # alternatively we just provide a detailed error page on the website which goes through troubleshooting steps!
    end
  end 
end
```

ashmaroli:
@parkr Nice to see your input, but I'm afraid `require(gem) rescue LoadError require(gem.tr('-', '/'))` is not how inline rescue works (at least in Ruby 2.3)..

Can you please point me to documentation regarding 
```ruby
try_something rescue ExceptionClass try_something_else
```

All I see is about using
```ruby
try_something rescue try_something_else
```

jamesluberda:
I was unable to get the inline code proposed by @parkr working, but I have put together a DRYer version than is currently committed, reflecting the initial implementation with improvements suggested here by @ashmaroli, @parkr, and @pathawks. That said, the less-DRY version currently committed still appears to be the only way to preserve an error message that reflects precisely what the user has specified in config, including the potential cases where the user does identify the gem by pathname, i.e. "foobar/foobar".

Here is part of the DRYer alternative I have to what is currently committed. It includes an effort to make the custom error message reflect the two possibilities for a failed require of a hyphenated gem, but report normally in non-hyphenated cases. It does, however, somewhat incorrectly pose both possibilities in the case of a required but missing pathed gem listed in config such as "foobar/foobar." The use of recursion also results in logging both requires even in the case of a hyphenated gem being successfully loaded with the second require, which could be confusing, though perhaps not worth worrying over.

Thoughts? @pathawks, since yours is the initial change request?

```ruby
      def require_with_graceful_fail(names)
        Array(names).each do |name|
          begin
            Jekyll.logger.debug "Requiring:", name.to_s
            require name
          rescue LoadError => e
            if name.include?("-")
              require_with_graceful_fail name.tr("-", "/")
            else
              requires = name.include?("/") ? "#{name.tr("/", "-")} or #{name}" : name
              log_require_with_graceful_fail_error(requires, e)
              raise Jekyll::Errors::MissingDependencyException, requires
            end
          end
        end
      end
```

jamesluberda:
Does the codeclimate cognitive complexity issue need to be addressed for this to go forward? It seems to me that CC is objecting largely to the DRYer code I recently committed, with its use of recursion and rescue catching (based on my reading of https://docs.codeclimate.com/docs/cognitive-complexity). We could revert to the less DRY code, which I suspect would not raise this issue with CC.

ashmaroli:
I think CC is mainly complaining about the use of recursion with multiple outcomes.
Every `if..else` or `.. ? .. : ..` gets added and compounded with recursion..

If you can reduce the "Cognitive Complexity" rating, well and good. If not, it doesn't matter..

PR #7154, opened 2018-07-23
Name some of our Travis builds

pathawks:
This doesn't gain us much, since most of our builds are not explicitly defined, but hopefully it adds a bit of clarity for somebody who is not used to looking at our Travis builds.


DirtyF:
And if it's not clear enough, we can add emojis 😄 

 - Rubocop 🤖 
- Default site 🏠

PR #7157, opened 2018-07-25
Reduce array traversal in Jekyll::Reader

ashmaroli:
This proposal is mainly to replace the following steps with optimized alternatives and see what Utterson feels about it.
```ruby
dot_dirs  = dot.select { |file| File.directory?(@site.in_source_dir(base, file)) }
dot_pages = dot_files.select { |file| Utils.has_yaml_header?(@site.in_source_dir(base, file)) }
```
Also `dot_files = (dot - dot_dirs)` simply exists to create another array `dot_static_files` ...

### DIFF
--- a/lib/jekyll/reader.rb
+++ b/lib/jekyll/reader.rb
@@ -39,13 +39,21 @@ module Jekyll
 
       return unless File.directory?(base)
 
+      dot_dirs = []
+      dot_pages = []
+      dot_static_files = []
+
       dot = Dir.chdir(base) { filter_entries(Dir.entries("."), base) }
-      dot_dirs = dot.select { |file| File.directory?(@site.in_source_dir(base, file)) }
-      dot_files = (dot - dot_dirs)
-      dot_pages = dot_files.select do |file|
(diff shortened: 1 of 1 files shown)


pathawks:
Not a performance optimization, but maybe more readable :+1:

ashmaroli:
> Not a performance optimization

Strange indeed. @pathawks Do you have any input on how / why my logic is flawed here..?

pathawks:
@ashmaroli Unfortunately, no. Logic seems sound; I also would have guessed this would be faster. 🤷‍♂️

ashmaroli:
> I can run one locally, if you like

No, that's not necessary. But, thanks anyways..

pathawks:
| #7157                                    |                287.63 |
| `master`                                 |                289.64 |

On my laptop, results show a couple seconds in the other direction. It seems like we're doing the same basic things either way, but I'd say this change is a bit more readable :+1:

ashmaroli:
@jekyllbot: merge +dev

PR #7158, opened 2018-07-25
Remove 'cache_dir' during `jekyll clean`

### LINKED ISSUE #4227: Consolidate caches and create a _cache folder for *all* to use.

pathawks:
Fixes #4227

### DIFF
--- a/lib/jekyll/commands/clean.rb
+++ b/lib/jekyll/commands/clean.rb
@@ -22,10 +22,12 @@ module Jekyll
           options = configuration_from_options(options)
           destination = options["destination"]
           metadata_file = File.join(options["source"], ".jekyll-metadata")
+          cache_dir = File.join(options["source"], options["cache_dir"])
           sass_cache = ".sass-cache"
 
           remove(destination, :checker_func => :directory?)
           remove(metadata_file, :checker_func => :file?)
+          remove(cache_dir, :checker_func => :directory?)
           remove(sass_cache, :checker_func => :directory?)
         end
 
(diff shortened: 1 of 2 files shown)


pathawks:
We will probably want to move `.sass-cache/` and `.jekyll-metadata` into this directory.

parkr:
We have site.in_source_dir() which is a helpful method for not worrying about File.join’s. What do you think about site.in_cache_dir() as well? Doesn’t have to be this PR, but I think it would add value for plug-in authors.

pathawks:
> We had talked about making this `_cache` instead of a dot directory. What do you think?

My original instinct was to use `_cache/`, but I changed it to the dot directory because of 
https://github.com/jekyll/jekyll/issues/4227#issuecomment-310653092 and because it's what we use for `.jekyll-metadata`. 

I have no preference regarding the default name. If somebody has a preference, we can change it :+1:

parkr:
Ah, ok!

pathawks:
@jekyllbot: :shipit: +major

PR #7162, opened 2018-07-27
Add version badge for date filters with ordinal

kplattret:
I was getting errors when trying to use the new `date_to_long_string: "ordinal"` parameter on my Jekyll website. I found out that this feature was introduced in Jekyll 3.8.0 (jekyll/jekyll#6773) and realised that the reason why I couldn't use it was because I had installed Jekyll through the `github_pages` gem, which only supports version 3.7.3 still today.

This adds a version badge to the two filters documented here, so that it may help others if they get the same error.

(cc @jekyll/documentation)


DirtyF:
@kplattret Thanks for the edit 👍 

@jekyllbot: merge +docs

PR #7163, opened 2018-07-28
Add closing tags for <a>

enh6:


pathawks:
I’m shocked that these are missing! Maybe lost to a bad merge?

chrisfinazzo:
@pathawks That's the only thing that I can think of. Would HTML Proofer's link testing catch this? cc/ @gjtorikian

ashmaroli:
@chrisfinazzo The markup here is inside code-blocks. Would html-proofer flag contents of `<pre>` and `<code>`?

chrisfinazzo:
@ashmaroli I don't see anything in proofer's documentation about preformatted text, so my inclination is to think that it wouldn't.

Most of the tests seem to be oriented as a "best practices/most reputable sites do this" kind of idea. Other than making sure that links actually return a 200 response, it doesn't do much validation.

PR #7168, opened 2018-07-31
Add TSV to list of supported _data files.

kenman345:
Noticed that this was missing from the website/documentation. It is technically supported for a very long time, seems odd it was missed.

Moved the rest of the wording down to logical spacing to keep in line with formatting.


DirtyF:
I hate when our docs don't reflect all the features, thanks for fixing Ken  🥗 :s

DirtyF:
I hate when our docs don't reflect all the features, thanks for fixing Ken  🥗 🌞  

@jekyllbot: merge +docs

PR #7160, opened 2018-07-26
Add `Site#in_cache_dir` helper method

ashmaroli:
Based on https://github.com/jekyll/jekyll/pull/7158#issuecomment-407924585

PR #7169, opened 2018-07-31
Add Cache class

pathawks:
This adds a basic caching mechanism to Jekyll.

Any internal Jekyll class, or any Jekyll plugin, can create its own cache. That cache will be shared among all instances of that class or plugin. The cache is essentially a key-value store.

The following operations can be performed on a cache:
 * Get the `value` associated with a `key`
 * Set the `value` associated with a `key`
 * Delete the `value` associated with a `key`
 * Check if a `key` exists in the cache
 * Clear the cache

Additionally, all caches can be cleared at once.

The cache as implemented here is not persistent at all. The idea is that most users would use a custom cache plugin to provide a persistent backend (if that is desired). This is not being done out of the box because of security concerns, and because there are many times when a persistent cache would be neither necessary or desired (like GitHub Pages).

An example would be a plugin that uses Ruby's `PStore` to write caches to disk so that they would persist across `jekyll build`s.

Cached values need not be Strings. This means we could cache (ie) parsed Liquid templates.

Currently, any cache plugin would be expected to monkey patch `Jekyll::Cache` to use a custom backend.

PRs that could be refactored to take advantage of `Jekyll::Cache`: #7159 #7136 #7108

### DIFF
--- a/docs/_tutorials/cache_api.md
+++ b/docs/_tutorials/cache_api.md
@@ -1,0 +1,87 @@
+---
+layout: tutorials
+permalink: /tutorials/cache-api/
+title: Cache API
+---
+
+Jekyll includes a caching API, which is used both internally as well as exposed
+for plugins, which can be used to cache the output of deterministic functions to
+speed up site generation. This cache will be persistent across builds, but
+cleared when Jekyll detects any changes to `_config.yml`.
+
+## Jekyll::Cache.new(name) → new_cache
+
+If there has already been a cache created with `name`, this will return a
+reference to that existing Cache. Otherwise, create a new Cache called `name`.
+
+If this Cache will be used by a Gem-packaged plugin, `name` should either be the
+name of the Gem, or prefixed with the name of the Gem followed by `::` (if a
+plugin expects to use multiple Caches). If this Cache will be used internally by
+Jekyll, `name` should be the name of the class that is using the Cache (ie:
+`"Jekyll::Converters::Markdown"`).
+
+Cached objects are shared between all Caches created with the same `name`, but
+are _not_ shared between Caches with different names. There can be an object
+stored with key `1` in `Jekyll::Cache.new("a")` and an object stored with key
+`1` in `Jekyll::Cache.new("b")` and these will not point to the same cached
+object. This way, you do not need to ensure that keys are globally unique.
+
+## getset(key) {block}
+
+This is the most common way to utilize the Cache.
+
+`block` is a bit of code that takes a lot of time to compute, but always
+generates the same output given a particular input (like converting Markdown to
+HTML). `key` is a `String` (or an object with `to_s`) that uniquely identifies
+the input to the function.
+
+If `key` already exists in the Cache, it will be returned and `block` will never
+be executed. If `key` does not exist in the Cache, `block` will be executed and
+the result will be added to the Cache and returned.
+
+```ruby
+def cache
+  @@cache ||= Jekyll::Cache.new("ConvertMarkdown")
+end
+
+def convert_markdown_to_html(markdown)
+  cache.getset(markdown) do
+    expensive_conversion_method(markdown)
+  end
+end
+```
+
+In the above example, `expensive_conversion_method` will only be called once for
+any given `markdown` input. If `convert_markdown_to_html` is called a second
+time with the same input, the cached output will be returned.
+
+Because posts will frequently remain unchanged from one build to the next, this
+is an effective way to avoid performing the same computations each time the site
+is built.
+
+## clear
+
+This will clear all cached objects from a particular Cache. The Cache will be
+empty, both in memory and on disk.
+
+
+### The following methods will probably only be used in special circumstances
+
+## cache[key] → value
+
+Fetches `key` from Cache and returns its `value`. Raises if `key` does not exist
+in Cache.
+
+## cache[key] = value
+
+Adds `value` to Cache under `key`.
+Returns nothing.
+
+## key?(key) → true or false
+
+Returns `true` if `key` already exists in Cache. False otherwise.
+
+## delete(key)
+
+Removes `key` from Cache.
+Returns nothing.
--- a/lib/jekyll/cache.rb
+++ b/lib/jekyll/cache.rb
@@ -17,11 +17,20 @@ module Jekyll
       @@caches[name] ||= {}
       @cache = @@caches[name]
     end
-    # rubocop:enable Style/ClassVars
 
     def self.clear
+      @@caches ||= {}
       @@caches.clear
     end
+    # rubocop:enable Style/ClassVars
+
+    # rubocop:disable Lint/UselessSetterCall
+    def self.clear_if_config_changed(config)
+      cache = Jekyll::Cache.new "Jekyll::Cache"
+      clear unless cache["config"] == config
+      cache["config"] = config
+    end
+    # rubocop:enable Lint/UselessSetterCall
 
     def getset(key)
       if key?(key)
--- a/lib/jekyll/cache.rb
+++ b/lib/jekyll/cache.rb
@@ -22,5 +22,13 @@ module Jekyll
     def self.clear
       @@caches.clear
     end
+
+    def getset(key)
+      if has_key?(key)
+        @cache[key]
+      else
+        @cache[key] = yield
+      end
+    end
   end
 end
--- a/lib/jekyll/cache.rb
+++ b/lib/jekyll/cache.rb
@@ -4,7 +4,7 @@ module Jekyll
   class Cache
     extend Forwardable
 
-    def_delegators :@cache, :[], :[]=, :clear, :delete, :has_key?
+    def_delegators :@cache, :[], :[]=, :clear, :delete, :key?
 
     # Get an existing named cache, or create a new one if none exists
     #
@@ -24,7 +24,7 @@ module Jekyll
     end
 
     def getset(key)
-      if has_key?(key)
+      if key?(key)
         @cache[key]
       else
         @cache[key] = yield
--- a/lib/jekyll/cache.rb
+++ b/lib/jekyll/cache.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+module Jekyll
+  class Cache
+    extend Forwardable
+
+    def_delegators :@cache, :[], :[]=, :clear, :delete, :has_key?
+
+    # Get an existing named cache, or create a new one if none exists
+    #
+    # name - name of the cache
+    #
+    # Returns nothing.
+    # rubocop:disable Style/ClassVars
+    def initialize(name)
+      @@caches ||= {}
+      @@caches[name] ||= {}
+      @cache = @@caches[name]
+    end
+    # rubocop:enable Style/ClassVars
+
+    def self.clear
+      @@caches.clear
+    end
+  end
+end
--- a/lib/jekyll/cache.rb
+++ b/lib/jekyll/cache.rb
@@ -56,10 +56,9 @@ module Jekyll
     # Returns nothing.
     def []=(key, value)
       @cache[key] = value
-      if @@disk_cache_enabled
-        path = path_to(hash(key))
-        dump(path, value)
-      end
+      return unless @@disk_cache_enabled
+      path = path_to(hash(key))
+      dump(path, value)
     end
 
     # If an item already exists in the cache, retrieve it
@@ -78,10 +77,9 @@ module Jekyll
     # Returns nothing.
     def delete(key)
       @cache.delete(key)
-      if @@disk_cache_enabled
-        path = path_to(hash(key))
-        File.delete(path)
-      end
+      return unless @@disk_cache_enabled
+      path = path_to(hash(key))
+      File.delete(path)
     end
 
     # Check if `key` already exists in this cache
@@ -104,11 +102,11 @@ module Jekyll
     def self.clear_if_config_changed(config)
       config = config.inspect
       cache = Jekyll::Cache.new "Jekyll::Cache"
-      unless cache.key?("config") && cache["config"] == config
-        clear
-        cache = Jekyll::Cache.new "Jekyll::Cache"
-        cache["config"] = config
-      end
+      return if cache.key?("config") && cache["config"] == config
+      clear
+      cache = Jekyll::Cache.new "Jekyll::Cache"
+      cache["config"] = config
+      nil
     end
 
     private
--- a/lib/jekyll/cache.rb
+++ b/lib/jekyll/cache.rb
@@ -54,13 +54,14 @@ module Jekyll
         path = path_to(hash(key))
         dump(path, value)
       end
-      value
     end
 
     def getset(key)
       self[key]
-    rescue
-      self[key] = yield
+    rescue StandardError
+      value = yield
+      self[key] = value
+      value
     end
 
     def delete(key)
--- a/lib/jekyll/cache.rb
+++ b/lib/jekyll/cache.rb
@@ -24,17 +24,17 @@ module Jekyll
       FileUtils.mkdir_p(path_to) if @@disk_cache_enabled
     end
 
-    # Clear all caches
-    def self.clear
-      delete_cache_files
-      @@caches.each_value(&:clear)
-    end
-
     # Disable Marshaling cached items to disk
     def self.disable_disk_cache!
       @@disk_cache_enabled = false
     end
     # rubocop:enable Style/ClassVars
+
+    # Clear all caches
+    def self.clear
+      delete_cache_files
+      @@caches.each_value(&:clear)
+    end
 
     # Clear this particular cache
     def clear
--- a/lib/jekyll/cache.rb
+++ b/lib/jekyll/cache.rb
@@ -21,7 +21,6 @@ module Jekyll
       @@base_dir ||= File.expand_path(".jekyll-cache/Jekyll/Cache")
       @cache = @@caches[name] ||= {}
       @name = name.gsub(%r![^\w\s-]!, "-")
-      FileUtils.mkdir_p(path_to) if @@disk_cache_enabled
     end
 
     # Disable Marshaling cached items to disk
--- a/lib/jekyll/cache.rb
+++ b/lib/jekyll/cache.rb
@@ -19,7 +19,7 @@ module Jekyll
       @@base_dir ||= File.expand_path(".jekyll-cache/Jekyll/Cache")
       @@caches ||= {}
       @cache = @@caches[name] ||= {}
-      @name = name
+      @name = name.gsub(%r![^\w\s-]!, "-")
       FileUtils.mkdir_p(path_to) if @@safe
     end
 
--- a/lib/jekyll/cache.rb
+++ b/lib/jekyll/cache.rb
@@ -138,6 +138,14 @@ module Jekyll
       FileUtils.rm_rf(path_to) if @@disk_cache_enabled
     end
 
+    # Delete all cached items from all caches
+    #
+    # Returns nothing.
+    def self.delete_cache_files
+      FileUtils.rm_rf(@@base_dir) if @@disk_cache_enabled
+    end
+    private_class_method :delete_cache_files
+
     # Load `path` from disk and return the result
     # This MUST NEVER be called in Safe Mode
     # rubocop:disable Security/MarshalLoad
@@ -162,13 +170,5 @@ module Jekyll
         Marshal.dump(value, cached_file)
       end
     end
-
-    # Delete all cached items from all caches
-    #
-    # Returns nothing.
-    def self.delete_cache_files
-      FileUtils.rm_rf(@@base_dir) if @@disk_cache_enabled
-    end
-    private_class_method :delete_cache_files
   end
 end
--- a/lib/jekyll/cache.rb
+++ b/lib/jekyll/cache.rb
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
 
+require "digest"
+require "fileutils"
+require "pstore"
+
 module Jekyll
   class Cache
     extend Forwardable
-
-    def_delegators :@cache, :[], :[]=, :clear, :delete, :key?
 
     # Get an existing named cache, or create a new one if none exists
     #
@@ -13,31 +15,113 @@ module Jekyll
     # Returns nothing.
     # rubocop:disable Style/ClassVars
     def initialize(name)
+      @@base_dir ||= File.expand_path(".jekyll-cache/Jekyll/Cache")
       @@caches ||= {}
-      @@caches[name] ||= {}
-      @cache = @@caches[name]
+      @cache = @@caches[name] ||= {}
+      @name = name
+      FileUtils.mkdir_p(path_to)
     end
 
     def self.clear
-      @@caches ||= {}
-      @@caches.clear
+      delete_cache_files
+      @@caches.each_value(&:clear)
     end
     # rubocop:enable Style/ClassVars
 
-    # rubocop:disable Lint/UselessSetterCall
-    def self.clear_if_config_changed(config)
-      cache = Jekyll::Cache.new "Jekyll::Cache"
-      clear unless cache["config"] == config
-      cache["config"] = config
+    def clear
+      delete_cache_files
+      @cache.clear
     end
-    # rubocop:enable Lint/UselessSetterCall
 
-    def getset(key)
-      if key?(key)
-        @cache[key]
+    def [](key)
+      return @cache[key] if @cache.key?(key)
+      path = path_to(hash(key))
+      if File.file?(path) && File.readable?(path)
+        @cache[key] = load(path)
       else
-        @cache[key] = yield
+        raise
       end
     end
+
+    def getset(key)
+      return @cache[key] if @cache.key?(key)
+      path = path_to(hash(key))
+      if File.file?(path) && File.readable?(path)
+        value = load(path)
+      else
+        value = yield
+        dump(path, value)
+      end
+      @cache[key] = value
+    end
+
+    def []=(key, value)
+      @cache[key] = value
+      path = path_to(hash(key))
+      dump(path, value)
+    end
+
+    def delete(key)
+      @cache.delete(key)
+      path = path_to(hash(key))
+      File.delete(path)
+    end
+
+    def key?(key)
+      return true if @cache.key?(key)
+      path = path_to(hash(key))
+      File.file?(path) && File.readable?(path)
+    end
+
+    # rubocop:disable Style/ClassVars
+    def self.clear_if_config_changed(config)
+      config = config.inspect
+      cache = Jekyll::Cache.new "Jekyll::Cache"
+      unless cache.key?("config") && cache["config"] == config
+        delete_cache_files
+        @@caches = {}
+        cache = Jekyll::Cache.new "Jekyll::Cache"
+        cache["config"] = config
+      end
+    end
+    # rubocop:enable Style/ClassVars
+
+    private
+
+    def path_to(hash = nil)
+      @base_dir ||= File.join(@@base_dir, @name)
+      return @base_dir if hash.nil?
+      File.join(@base_dir, hash[0..1], hash[2..-1]).freeze
+    end
+
+    def hash(key)
+      Digest::SHA2.hexdigest(key).freeze
+    end
+
+    def delete_cache_files
+      FileUtils.rm_rf(path_to)
+    end
+
+    # rubocop:disable Security/MarshalLoad
+    def load(path)
+      cached_file = File.open(path, "rb")
+      value = Marshal.load(cached_file)
+      cached_file.close
+      value
+    end
+    # rubocop:enable Security/MarshalLoad
+
+    def dump(path, value)
+      dir, _file = File.split(path)
+      FileUtils.mkdir_p(dir)
+      cached_file = File.open(path, "wb")
+      Marshal.dump(value, cached_file)
+      cached_file.close
+    end
+
+    def self.delete_cache_files
+      FileUtils.rm_rf(@@base_dir)
+    end
+    private_class_method :delete_cache_files
   end
 end
--- a/lib/jekyll/cache.rb
+++ b/lib/jekyll/cache.rb
@@ -15,12 +15,14 @@ module Jekyll
     # Returns nothing.
     # rubocop:disable Style/ClassVars
     def initialize(name)
-      throw unless defined? @@safe
+      unless defined? @@disk_cache_enabled
+        @@disk_cache_enabled = true
+      end
       @@base_dir ||= File.expand_path(".jekyll-cache/Jekyll/Cache")
       @@caches ||= {}
       @cache = @@caches[name] ||= {}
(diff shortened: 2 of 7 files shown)


parkr:
This is an example of an in-memory Cache. You also have an example in another PR of a disk cache. Do you think we should have this class allow consumers to specify which type they want? So Cache.new(“compiled_markdown”, :disk) would write to disk but Cache.new(“related_posts”, :memory) would just write to memory?

pathawks:
> Do you think we should have this class allow consumers to specify which type they want?

I was thinking that this would be entirely up to whatever caching plugin the user chooses to use. I know that there are Jekyll installations (such as GitHub Pages) where it would never make sense to use a persistent cache. What would happen in these settings if a consumer requested a disk cache? Would we fallback to using a memory cache? Would we assume that nothing would be worth caching?

There are some things that we would for sure like to cache, even if only in memory. Parsed Liquid templates would fall into this category. If we could only parse them once per build, there are performance improvements to be had. If we could get away with parsing them less than once per build, there might be even more improvements.
On the other hand, there are things that might be dangerous to persist. I would be afraid to cache related posts, since they depend on so many things, and if one post was added or deleted or modified it might affect everything in the related posts cache. This should never persist across builds, and I was assuming was beyond the scope of this PR.

I guess I see a couple variables:
 * Is something worth caching if we do not have a cache that persists across builds? (Converted Markdown is probably not)
 * Is something safe to persist across builds? (Related posts are probably not, with only a key-value store)

I will look into LRU caches. I agree that we would probably need some size limit.

pathawks:
>  What do you think about an LRU Cache which has a maximum size? There are some really great single-file open source LRU Caches out there you could base this approach off of.

Thinking a bit more about this.

I really think Jekyll Core should only expose some hooks for more powerful cache plugins. I don't think that we should reinvent the wheel by including our own LRU implementation, nor do I think that we should `require` a specific LRU gem if most power users will be using a different caching plugin anyway.

What is required for some caching API to be accepted in Jekyll Core? Should I create a plugin that would use a Git style cache to back this caching?

parkr:
Jekyll Core is in the business of providing primitives. I think a cache is a great type of primitive. Users should be unaware that it’s an LRU Cache, and yes I think we should choose the implementation and implement it here as a primitive. I think we should do some more work to decide what the end user should use (and who the end user is). If the end user is all plugin authors, then what is the purpose of the cache? When should it be used? This will help us decide what the API will look like. The current implementation seems aimed toward plugins which do very expensive work the output of which can be reused. Whether it’s a disk cache or an in-memory cache is a matter of whether the cached values are useful in subsequent builds. LRU is just a convenient algorithm for capping the size in a smart way. 

So how would you like to use this capability as a plugin writer? In core? I’d start there and work backwards.

pathawks:
> So how would you like to use this capability as a plugin writer? In core? 

Currently, we use memoization in many places as a sort of poor man's cache. #7136 is a great example of this. These templates will often not change between builds, so there is nothing preventing us from caching these templates across builds.

When we memoize all these things, we are not placing any upper limit on the amount of memory we are using. Because of this, I assumed that it would not be a requirement for this PR.

How would we decide how much memory to use? Should this be user configurable?
Are we keeping track of how much memory Jekyll uses in general for memoization?

> The current implementation seems aimed toward plugins which do very expensive work the output of which can be reused.

I'm sorry if it appears that way. My primary focus is on places in _core_ that do very expensive work. I was going to avoid making those changes until we had an API that we could all agree on, but perhaps I should include those changes in this PR (which would make it a monster of a PR) so that we call all see the examples I have in mind?

My intent was to provide some hooks so that, instead of _just_ memoizing things like that, a site could add a plugin which would provide persistent storage for all of these things. Since writing objects to a disk cache involves Marshaling objects around, this is probably not something that Jekyll would allow in Safe Mode. How should we handle this? Should we fall back to using a memory backed cache? Should we just avoid caching anything?

> what is the purpose of the cache? When should it be used?

The cache should be used when we have a deterministic function that is fairly time expensive. We can then store the input-output pair, and we never have to worry about keeping track of things like dependencies.

There are many instances where a function is deterministic assuming the configuration does not change. (If a user changes Markdown options, then _all_ Markdown output could possibly change.) For this reason, it makes sense to me to clear the whole cache if the site's config changes. That eliminates another class of problems our current incremental approach suffers.

> I think we should do some more work to decide what the end user should use (and who the end user is).

I know that some large users would rather use a database as a persistent store than files on disk. This functionality should be available, either in Core or from a plugin. It does not make sense to me for Jekyll Core to include functionality for communicating with a DB backed cache.

ashmaroli:
Noting a couple of things below for posterity.

> The cache should be used when we have a deterministic function that is fairly time expensive.

Agreed. Parsing and Rendering Liquid Templates is one such function. Given that parsing a Liquid template is a time-expensive function (documented as [inline-comment](https://github.com/Shopify/liquid/blob/70e75719dee67a51c6e94d8706baa9b19f2958f1/lib/liquid/template.rb#L3-L8)) and a parsed template outputs the same markup for a given Hash data (ergo, *deterministic*).
The unpleasant situation is when the Hash data (usually optimized by subclassing `Liquid::Drop`) passed to a template itself is a time-expensive process. Since the templates are a cached entity, the Hash data **has to be a variable determined by the given `context`**. Context cannot be efficiently cached because it is a composite of all Liquid data at play for a given "page / document".

In light of above,
  - Parsing Liquid template files can be optimized by caching because they *generally never change* from one pass to another. Even with the use of Jekyll's `:pre_render` hooks, this can be assumed to be safe.
  - Rendering `Liquid::Tag` can be optimized by caching **provided they do not depend on data from the "page / document" itself**

The Core can use the cache to optimize `{% include %}` and other bundled Tags.
The Plugin authors can use the cache in similar fashion to optimize rendering their plugins where applicable.

DirtyF:
I'm having a hard time understanding what event will trigger a new version of the cache (except for config changes), I hope one can document this with a sketch explaining how it works 😅

pathawks:
@DirtyF This cache is only implemented in memory, so it will start empty for each build.
If the cache were saved to disk, it would be cleared if the config has changed between builds, or on `jekyll clean`. It seems like you have a specific concern in mind?

ashmaroli:
@pathawks Given that a Cache instance currently holds a parsed Liquid Template, what happens when I "edit" the previously parsed template..?

pathawks:
If we use the _content_ as the key to the cache (instead of something like the filename), if the content changes then it will parse the new content and add that to the hash. Similarly, if you had 100 identical Liquid Templates, we would parse the first one and cache it, and then the other 99 would all just be retrieved from the cache.

When it comes to rendering Markdown or Liquid Templates, the cache does not even know the file name; just that a particular input produces a particular output. That's why I mentioned deterministic functions earlier; that's the only time when this kind of cache is applicable.

This is how Git works. At the blob level, Git doesn't care about file names or anything. If you have a million identical files, Git can store them internally as just one blob (which then would be mapped to a million filenames).

ashmaroli:
> we use the content as the key to the cache (instead of something like the filename)

I see. How is the template then retrieved from the cache..?

ashmaroli:
@pathawks Please do not pull the changes in my PR onto this branch. That would essentially render my PR obsolete when this PR gets merged before that..

pathawks:
@ashmaroli Specifics were asked for, so I am trying to provide specifics. It is difficult to have a conversation about how to build a caching API without any examples of how that API would be implemented.

mattr-:
> How would we decide how much memory to use? Should this be user configurable?

We shouldn't decide how much memory to use (and I'm not aware of anything in Ruby that lets us do so). I also don't think this should be user configurable.

> Are we keeping track of how much memory Jekyll uses in general for memoization?

Not within the application itself. It's possible to profile memory usage, but I'm unconvinced that folks that have problems with Jekyll's performance are related to its memory usage. :smile:

>  I was going to avoid making those changes until we had an API that we could all agree on, but perhaps I should include those changes in this PR

That's definitely the way to go in my mind. If you have concrete examples that would be part of another PR that you can include in this one, I think that would be great to have to provide more context. It's much easier to investigate impact with smaller changes.

----
@ashmaroli Why do you need changes to this PR? There should be no reason for any PR to wait on another one before being merged, especially if one PR has approval and another one doesn't. Unless there are specific code changes that need to be made, would you please dismiss your review?

ashmaroli:
> There should be no reason for any PR to wait on another one before being merged

@mattr- Actually, there is a valid reason here.
  - This PR has pulled in the entire significant set of changes from an Approved PR #7136 .
     **If by any reason this PR gets merged to `master` before `#7136` does**, then:
       * the other PR ends up with just the "insignificant portion" (documentation update)
       * my contribution to the "significant portion" gets lost without a trace. Even the courtesy mention in the commit message in this PR is going to be lost when JekyllBot does the squash-merge.

That should not happen to any contributor irrespective of whether they are a first-time-contributor or a long-term-contributor, and is a violation of Principles of Collaboration, irrespective of whether the "mistake" was made by the Owner, or someone with Administrator privileges or a novice contributor.

DirtyF:
If this PR make #7136 insignificant, it's because this is a better implementation, with a broader vision. I don't see any violantion to any principle of collaboration here, just a smarter solution. We just should be grateful to @pathawks for his work here, and play collective.

ashmaroli:
> I don't see any violation to any principle of collaboration here

@DirtyF So there's no value in my observation / discovery that there's a lot of performance improvement to be achieved by caching parsed Liquid templates just because I used a class-wide cache and Pat simply replaces that with a module-wide cache..?

Why is everyone so hostile????

lib/jekyll/cache.rb
```diff
@@ -0,0 +1,118 @@
...
+    def path_to
+      File.expand_path(".jekyll-cache/Jekyll/Cache/#{@name}")
+    end
+
+    def hash(key)
+      Digest::SHA2.hexdigest(key)
+    end
+
+    def cached_hash_path(hash)
+      ".jekyll-cache/Jekyll/Cache/#{@name}/#{hash}"
```
pathawks: It would be easy to split this into `...#{@name}/#{hash[0..1]}/#hash{2..-1}` if we would prefer. We would just need to always make sure the subdirectory exists before writing the file.

mattr-:
How hard would it be to convince you to use https://github.com/SamSaffron/lru_redux for this? I'm having a hard time understanding why we're rolling our own cache when we can use something like `lru_redux`.

pathawks:
> How hard would it be to convince you to use https://github.com/SamSaffron/lru_redux for this?

**lru_redux** does not provide a persistent cache. We need persistence across builds to benefit from things like cached Markdown conversion. Would we Marshal the whole cache between builds?

The only thing that using **lru_redux** would gain us over the straightforward memoization I've used here is placing a hard limit on the size of the cache.

There are a couple other hurdles to using an LRU cache replacement policy:
 * What is the size of the cache?
   - Do _we_ set that size, or does the user? Will there be a larger cache on machines with more memory?
 * Is the size of the cache fixed per build, or per plugin?
   - Will something that has been cached for Jekyll Core be ejected to store something for **jekyll-feed*?
 * Why Least Recently Used?
   - For our purposes, we care less about how recently it was used and more about how long it takes to `yield`. If something is used on every page of a site but takes an insignificant amount of time to generate from scratch, and something else is used only twice during the build but takes a long time to generate, it seems like we should be focused on caching the thing that takes a long time to generate.

With infinite knowledge, we would assign each object a score of `time to yield * number of occurrences`, and then use something like a Huffman coding algorithm to pick which objects to cache such that we maximize the amount of time that we can save.

mattr-:
I was mostly interested in the thread safety that is provided by `lru_redux` but if it doesn't meet any of our other needs, then it's not worth it.

pathawks:
`getset` should be thread safe, so long as the block will `yield` the same output each time it is called.

I think the only time this might not be thread safe would be for a plugin to use `[]` and `[]=` for something where the value being set depends on the value being retrieved, right? If necessary, we could add an atomic setter; I just haven't yet seen any scenario where it would be useful.

Is there a use case I am not thinking of?

pathawks:
> What do you think about an LRU Cache which has a maximum size?

in practice, it seems this API can actually _reduce_ Jekyll's memory footprint.

In 177c1ca09e1be5f8a2924aed6c74bfa7d53079c5, I tested this API with cached Markdown, Liquid templates, and syntax highlighting, along with modified versions of `jekyll-feed` and `jekyll-seo-tag` that took advantage of caching. Jekyll's memory footprint was reduced by half when building https://github.com/tomjoht/tomjoht.github.io (the site in Utterson's list with the largest memory footprint). For sites that did not benefit from this caching, Jekyll's memory footprint only grew by <9% (80MB to 87MB).

How should the disk cache work with safe mode? Should we just disable reading from the disk cache during safe mode?

What more needs to be done here before this can be merged?

mattr-:
From a quick glance at the code, yes, we should disable the disk cache in safe mode.

pathawks:
The safe mode semantics are a bit odd. I was trying to mimic the [plugin `safe` semantics](https://github.com/jekyll/jekyll/blob/035ea729ff5668dfc96e7f56a86d214e5a633291/lib/jekyll/plugin.rb#L53-L63).

lib/jekyll/cache.rb
```diff
@@ -0,0 +1,139 @@
...
+      cached_file = File.open(path, "rb")
+      value = Marshal.load(cached_file)
+      cached_file.close
+      value
+    end
+    # rubocop:enable Security/MarshalLoad
+
+    def dump(path, value)
+      return unless @@safe
+      dir, _file = File.split(path)
```
ashmaroli: Why not simply `dir = File.dirname(path)` ??

lib/jekyll/cache.rb
```diff
@@ -0,0 +1,139 @@
...
+      FileUtils.mkdir_p(dir)
+      cached_file = File.open(path, "wb")
+      Marshal.dump(value, cached_file)
+      cached_file.close
+    end
```
ashmaroli: This can be simplified by using a block..
```ruby
File.open(path, "wb") do |cached_file|
  Marshal.dump(value, cached_file)
end
```

lib/jekyll/cache.rb
```diff
@@ -0,0 +1,139 @@
...
+    #
+    # Returns nothing.
+    # rubocop:disable Style/ClassVars
+    def initialize(name)
+      @@base_dir ||= File.expand_path(".jekyll-cache/Jekyll/Cache")
+      @@caches ||= {}
+      @@safe ||= true
+      @cache = @@caches[name] ||= {}
+      @name = name
+      FileUtils.mkdir_p(path_to) if @@safe
```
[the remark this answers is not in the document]
ashmaroli: > Would @@safe = true unless defined? @@safe work?

No. `FileUtils.mkdir_p(path_to) if @@safe` would still be *unchecked*. 
Instead you should call `FileUtils.mkdir_p(path_to) if @@safe` **before** handling `@@safe`:
```ruby
FileUtils.mkdir_p(path_to) if @@safe
@@safe = true if @@safe.nil?
```

lib/jekyll/cache.rb
```diff
@@ -0,0 +1,139 @@
...
+      else
+        raise
+      end
+    end
+
+    def getset(key)
+      return @cache[key] if @cache.key?(key)
+      path = path_to(hash(key))
+      if @@safe && File.file?(path) && File.readable?(path)
+        value = load(path)
```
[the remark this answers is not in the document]
pathawks: I really wanted to just call `#[]` from `#getset`, but it won't work. They both need to do very different things in the else block.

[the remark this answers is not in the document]
ashmaroli: Agreed.

lib/jekyll/cache.rb
```diff
@@ -0,0 +1,139 @@
...
+        dump(path, value) if @@safe
+      end
+      @cache[key] = value
+    end
+
+    def []=(key, value)
+      @cache[key] = value
+      if @@safe
+        path = path_to(hash(key))
+        dump(path, value)
```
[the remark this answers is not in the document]
ashmaroli: > If there were tremendous performance benefits

Unfortunately, no. There are no discernible perf benefits from **not using** lvars *unnecessarily*. Just that its odd that GC would be triggered for a variable that is going to be used just once..

lib/jekyll/cache.rb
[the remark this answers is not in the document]
pathawks: _Or_, since `#[]` just throws if it the key doesn't exist, `#getset` could `rescue` and do it's thing? I don't know if this would make the code easier or more difficult to follow.

```ruby
def getset(key)
  self[key]
rescue
  value = yield
  dump(path, value) if @@safe
  @cache[key] = value
end
```

ashmaroli:
@pathawks in case you are wondering why AppVeyor is failing consecutively, it's because colons are not allowed in a dirname. i.e., **You cannot create a directory labelled `Jekyll::Cache` on Windows**

lib/jekyll/cache.rb
```diff
@@ -0,0 +1,134 @@
...
+      @cache.clear
+    end
+
+    def [](key)
+      return @cache[key] if @cache.key?(key)
+      path = path_to(hash(key))
+      if @@safe && File.file?(path) && File.readable?(path)
+        @cache[key] = load(path)
+      else
+        raise
```
pathawks: We should probably create a new exception for this

lib/jekyll/cache.rb
```diff
@@ -0,0 +1,135 @@
...
+    extend Forwardable
+
+    # Get an existing named cache, or create a new one if none exists
+    #
+    # name - name of the cache
+    #
+    # Returns nothing.
+    # rubocop:disable Style/ClassVars
+    def initialize(name)
+      throw unless defined? @@safe
```
ashmaroli: What "catches" this throw here?

ashmaroli:
All this use of class variables and instance variables is lowering the "Aesthetics" of the Ruby Class and raising the "Prone to bugs" quotient, IMO..
`@@safe, @@base_dir, @base_dir, @@caches, @cache`...

@pathawks I suggest you try defining the class attributes within the Singleton class:
```ruby
class Cache
  class << self
    attr_reader :base .., # and so forth..
  end
end
```
The above can also allow writing tests for the various attributes

DirtyF:
It may be outside the scope of this PR but we should add `.jekyll-cache` to default `.gitignore` in site and theme builder templates.

ashmaroli:
> we should add `.jekyll-cache` to default `.gitignore`

@DirtyF `master` already initializes `.jekyll-cache` as the default cache_dir. So a separate PR is better..

ashmaroli:
@pathawks Wouldn't `Jekyll::Cache.new('Jekyll::Foo')` and `Jekyll::Cache.new('Jekyll--Foo')` be separate objects in memory but write to the same path..?
That is a race condition waiting to happen.... no?

pathawks:
> Wouldn't Jekyll::Cache.new('Jekyll::Foo') and Jekyll::Cache.new('Jekyll--Foo') be separate objects in memory but write to the same path..?

I am open to suggestion, but in practice this will never happen accidentally. When would we have one cache named `Jekyll::Foo` and another named `Jekyll--Foo`? For the name of the particular cache, we use either the name of the class (in Core) or the name of the Gem (in plugins). Even if two caches were being stored in the same directory on disk, there would need to be a hash collision for any problem to arise. If you find a SHA2 collision, you could publish those findings. The only reason we are using separate subdirectories on disk is to make it possible for a particular plugin to clear all of its own cached objects without needing to keep track of those objects.

If a malicious plugin tries to override a directory and mess with cached objects things could break, but _anytime_ a user loads a malicious plugin things could break.

lib/jekyll/cache.rb
pathawks: Nothing. The idea is that `@@safe` must be defined before the cache can be used. `@@safe` is set when Jekyll is initialized. I'm going to change the semantics of this :+1:

pathawks:
> until after you finalize the API

Is the API not finalized? I cannot really move further until we can come to a consensus about how this should work.

ashmaroli:
> Is the API not finalized?

Oh, I thought it wasn't since you were changing the semantics of a couple of methods.
If you say, its ready to be finalized, then it is ready for another review.. Will post one shortly..

lib/jekyll/cache.rb
```diff
@@ -0,0 +1,175 @@
...
+      File.file?(path) && File.readable?(path)
+    end
+
+    # Compare the current config to the cached config
+    # If they are different, clear all caches
+    #
+    # Returns nothing.
+    def self.clear_if_config_changed(config)
+      config = config.inspect
+      cache = Jekyll::Cache.new "Jekyll::Cache"
```
ashmaroli: Would delegating `Jekyll::Cache.new "Jekyll::Cache"` to a dedicated method called `:initialize_self` be over-engineering?
```ruby
def initialize_self
  Jekyll::Cache.new("Jekyll::Cache") # or simply `new("Jekyll::Cache")`
end
```

lib/jekyll/cache.rb
```diff
@@ -0,0 +1,175 @@
...
+    # rubocop:enable Security/MarshalLoad
+
+    # Given a path and a value, save value to disk at path
+    # This should NEVER be called in Safe Mode
+    #
+    # Returns nothing.
+    def dump(path, value)
+      return unless @@disk_cache_enabled
+      dir = File.dirname(path)
+      FileUtils.mkdir_p(dir)
```
ashmaroli: Personal nit:
```diff
- dir = File.dirname(path)
- FileUtils.mkdir_p(dir)
+ FileUtils.mkdir_p(File.dirname(path))
```

lib/jekyll/cache.rb
```diff
@@ -0,0 +1,175 @@
+# frozen_string_literal: true
+
+require "digest"
+require "fileutils"
+require "pstore"
+
+module Jekyll
+  class Cache
+    extend Forwardable
```
ashmaroli: What's the role of `Forwardable` in this class..?

lib/jekyll/cache.rb
```diff
@@ -0,0 +1,175 @@
...
+
+    # rubocop:disable Style/ClassVars
+    @@caches = {}
```
ashmaroli: Personal nit:
Can we name this as `@@stash` instead of resembling `@cache`?

lib/jekyll/cache.rb
pathawks: 🤷‍♂️ I'm not sure. This is the only place we are using it.

To be consistent with the way I am using the cache in other places, we could call the method `cache`, but it might be odd to have a `cache` method in the `Cache` class.

Yo Dawg, I hear you like caches...

ashmaroli:
> We don't really document Jekyll's Ruby API on the docs site, do we?

The remaining classes' Ruby API is pretty simple, straight-forward and not intended to be consumed on a varying scale.

> Can we assume that developers will read the comments to learn the API?

That's not a good idea IMO, its better that the author translate their vision into text rather than leaving things open for interpretation.
Not every plugin developer is a Ruby pro.
Besides, the other maintainers can also refer that documentation when users post "creative queries" on using the Cache API..

pathawks:
@ashmaroli Ok, I am on board with adding docs to the site :+1:

Any idea where they should go in the site's hierarchy?

lib/jekyll/cache.rb
```diff
@@ -0,0 +1,175 @@
...
+    end
+
+    # Disable Marshaling cached items to disk
+    def self.disable_disk_cache!
+      @@disk_cache_enabled = false
+    end
+    # rubocop:enable Style/ClassVars
+
+    # Clear all caches
+    def self.clear
```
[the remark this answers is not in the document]
pathawks: I don't have strong feelings, but I like the symmetry between `individual_cache.clear` and `Jekyll::Cache.clear`. One is clearly clearing a particular cache, while the other is acting on the whole cache system.

If you think `Jekyll::Cache.clear_all` is more clear, that doesn't bother me. It should maybe be a `private_class_method` anyway 🤷‍♂️

ashmaroli:
> Any idea where they should go in the site's hierarchy?

The best place would be the Tutorials.. We can then link to that document in our release post and upgrading document.

lib/jekyll/cache.rb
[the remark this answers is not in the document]
ashmaroli: `¯\_(ツ)_/¯` your call..

docs/_tutorials/cache_api.md
```diff
@@ -0,0 +1,87 @@
+---
+layout: tutorials
+permalink: /tutorials/cache-api/
+title: Cache API
+---
+
+Jekyll includes a caching API, which is used both internally as well as exposed
```
DirtyF: When did Jekyll introduced this API?

Ex: Jekyll v4 introduced a caching API…

docs/_tutorials/cache_api.md
```diff
@@ -0,0 +1,87 @@
...
+for plugins, which can be used to cache the output of deterministic functions to
```
DirtyF: What functions?

docs/_tutorials/cache_api.md
```diff
@@ -0,0 +1,87 @@
...
+speed up site generation. This cache will be persistent across builds, but
+cleared when Jekyll detects any changes to `_config.yml`.
+
+## Jekyll::Cache.new(name) → new_cache
```
DirtyF: Can we introduce sections with intent and keep the code snippet for examples?

docs/_tutorials/cache_api.md
```diff
@@ -0,0 +1,87 @@
...
+
+If there has already been a cache created with `name`, this will return a
+reference to that existing Cache. Otherwise, create a new Cache called `name`.
+
+If this Cache will be used by a Gem-packaged plugin, `name` should either be the
+name of the Gem, or prefixed with the name of the Gem followed by `::` (if a
+plugin expects to use multiple Caches). If this Cache will be used internally by
+Jekyll, `name` should be the name of the class that is using the Cache (ie:
+`"Jekyll::Converters::Markdown"`).
```
DirtyF: Could we give an example for a jekyll plugin here?

docs/_tutorials/cache_api.md
pathawks: How could I word this better? I am thinking like a programmer, and can’t think with words.

“You have a plugin. Parts of that plugin meet the requirements of a ‘function.’ Caching can speed up those parts of your plugin.”

docs/_tutorials/cache_api.md
pathawks: Was trying to mimic RubyDoc.
I am open to suggestions.

docs/_tutorials/cache_api.md
DirtyF: Examples are always a safe bet.

ashmaroli:
@pathawks I have a couple of suggestions and would like your inputs on them..:

  - Could the `:initialize` be such that it expects a simple string or an Object instance..?

    ```ruby
    class Jekyll::Cache
      def initialize(obj)
        name_str = \
          if obj.is_a? String
            obj
          elsif obj.respond_to? :class  # Class Instance
            obj.class.name
          elsif obj.respond_to? :name   # Module or Singleton Class
            obj.name
          else
            "Jekyll::Cache::GenericObject"
          end
        @name = # process(name_str)
        [...]
      end
    end
    ```
    The initialization can then simply be **`Jekyll::Cache.new(self)`** for Jekyll's core classes and for plugins under their own unique namespace.
  - Could the cache paths be simplified to `"./.jekyll-cache/<top_level_namespce>/<descendant/...>"`?
    e.g. For **`Jekyll::Renderer`** it'd be  **`".jekyll-cache/Jekyll/Renderer/<path_to(hash)>"`** but for **`JekyllFeed::Generator`**, it'd be **`".jekyll-cache/JekyllFeed/Generator/<path_to(hash)>"`**

pathawks:
> Could the :initialize be such that it expects a simple string or an Object instance..?

Thought about that. I prefer the “no magic” approach.

> Could the cache paths be simplified…

We are not the only ones using the `.jekyll-cache` directory, nor are we the first ones. We need to be very cautious to avoid namespace collisions, which is why I have chosen to keep all of our stuff in a subdirectory. This also gives us more flexibility if we radically change the way we are storing files on disk: we could just change the name of the subdirectory and never worry about collisions.

lib/jekyll/cache.rb
oe: I think this usage is fine. There is such a thing as over-engineering.

lib/jekyll/cache.rb
```diff
@@ -0,0 +1,175 @@
...
+    @@caches = {}
+    @@disk_cache_enabled = true
+
+    # Get an existing named cache, or create a new one if none exists
+    #
+    # name - name of the cache
+    #
+    # Returns nothing.
+    def initialize(name)
+      @@base_dir ||= File.expand_path(".jekyll-cache/Jekyll/Cache")
```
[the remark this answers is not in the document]
ashmaroli: This is the only query that remains unanswered..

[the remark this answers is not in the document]
pathawks: What we need to do is make this relative to `site.source` instead of `pwd`.
I’m not sure of the syntactically correct way to do that.

I agree this does not belong here.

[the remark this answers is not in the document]
ashmaroli: #7160 once merged, will give you the helper method to call a path within `.jekyll-cache` relative to `site.source`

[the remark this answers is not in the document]
pathawks: The problem is that we need to get an instance of `site`. The helper method you added is not static.

I hate singletons.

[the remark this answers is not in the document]
ashmaroli: I didn't realize that this class doesn't link to a particular Site instance.
But we do have access to the `config` hash...
I guess, the issue will have to be handled via a separate PR after **this PR** is merged to `master`.

lib/jekyll/cache.rb
```diff
@@ -0,0 +1,168 @@
...
+    # If an item already exists in the cache, retrieve it
+    # Else execute code block, and add the result to the cache, and return that
+    #   result
+    def getset(key)
+      self[key]
+    rescue StandardError
+      value = yield
+      self[key] = value
+      value
+    end
```
ashmaroli: can't the `rescue` block be simplified to the following?
```ruby
rescue
  self[key] = yield
end
```

pathawks: No, because we cannot assume that `[]=` returns `value`.

ashmaroli: I see.. Fair enough..

pathawks: Currently, our Rubocop settings forbid setting an explicit return value in `[]=`.
I did consider this though.

DirtyF:
> Make it work, make it right, make it fast.™ — Kent Beck

Caching is hard, but it looks like we have _something_ that works —  would love to see some benchmarking on the `max-cache` branch to see how it performs in the real world if this gets approved.

@jekyll/core : Can we go with this first implementation or are there any memory/edge case concerns that prevents us from shipping a first iteration for this caching API?

oe:
@jekyllbot: merge +minor

ashmaroli:
@pathawks May I delete this branch..?

PR #7171, opened 2018-08-02
Fix links in Code Climate badges

ashmaroli:
The badges currently point to `https://github.com/jekyll/jekyll/blob/master/codeclimate` and `https://github.com/jekyll/jekyll/blob/master/coverage` erroneously


DirtyF:
Thanks for the edit.

PR #7173, opened 2018-08-03
Output Jekyll Version while debugging

ashmaroli:
To allow one to deduce what version of Jekyll was used to build the site by simply reading through the complete `--verbose` output


pathawks:
@jekyllbot: merge +major

PR #7175, opened 2018-08-07
Cache highlights

pathawks:
This uses #7169 to cache highlighted code blocks.

### DIFF
--- a/lib/jekyll/tags/highlight.rb
+++ b/lib/jekyll/tags/highlight.rb
@@ -31,22 +31,25 @@ module Jekyll
       def render(context)
         prefix = context["highlighter_prefix"] || ""
         suffix = context["highlighter_suffix"] || ""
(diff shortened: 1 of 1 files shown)


lib/jekyll/tags/highlight.rb
```diff
@@ -67,6 +70,10 @@ def sanitized_opts(opts, is_safe)
 
       OPTIONS_REGEX = %r!(?:\w="[^"]*"|\w=\w|\w)+!
 
+      def cache
+        Jekyll::Cache.new("Jekyll::Tags::HighlightBlock")
```
ashmaroli: Can't this be inlined into the `:render` method both itself..? Or is it going to be used in another method call in a future PR?

pathawks: Readability and consistency.

PR #7179, opened 2018-08-09
Add info how to deploy using pre-push git hook

vfreefly:
Add to  docs/_docs/deployment-methods.md info how to deploy using git pre-push hook.


DirtyF:
Thanks @vfreefly ❤️ 

@jekyllbot: :shipit:  +docs

PR #7181, opened 2018-08-13
Corrected sample usage of postfiles

dmalan:


DirtyF:
@jekyllbot: merge +docs

DirtyF:
Thanks @dmalan 👍

PR #7182, opened 2018-08-15
Strip extra slashes via `Jekyll.sanitized_path`

ashmaroli:
Depending on whether the asserted scenario is outside the scope of `Jekyll.sanitized_path`, this PR can be seen as a bug-report-cum-resolution combo or just an edge-case noise.

/cc @benbalter for his opinion on the relevance of the asserted scenario


ashmaroli:
Thank you @benbalter :smiley:

DirtyF:
@jekyllbot: merge +minor

PR #7183, opened 2018-08-15
Cache converter in renderer

pathawks:


DirtyF:
Could we avoid empty description issues and e.g. briefly explain why caching the converted content should be done here?

DirtyF:
@jekyllbot: merge +major

PR #7184, opened 2018-08-16
Ignore cache directory

DirtyF:
Following #7169 

When creating a new site or a new theme, make sure we don't version the default cache directory.


DirtyF:
@jekyllbot: merge +major

PR #7187, opened 2018-08-18
Respect SOURCE_DATE_EPOCH environment variable (if set) over Time.now

lamby:
Whilst working on the Reproducible Builds effort [0], we noticed
that jekyll builds sites that are not reproducible.

This is because site.time uses the current time of day and thus
anything then shipping a Jekyll site (eg. as documentation) will
therefore not be reproducible.

This commit prefers the value in the SOURCE_DATE_EPOCH environment
variable [1] if it is set and config.time is also not set.

 [0] https://reproducible-builds.org/
 [1] https://reproducible-builds.org/specs/source-date-epoch/

### DIFF
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -78,26 +78,11 @@ module Jekyll
       Jekyll.logger.info @liquid_renderer.stats_table
     end
 
-    # Get Time
-
-    def now
-      now = if ENV["SOURCE_DATE_EPOCH"].nil?
-              Time.now
-            else
-              Time.at(ENV["SOURCE_DATE_EPOCH"].to_i).gmtime
-            end
-      self.time = if config["time"]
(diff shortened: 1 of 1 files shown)


oe:
@lamby Can you fix the RuboCop offense as well as add a simple test? Thanks so much!

ashmaroli:
@DirtyF For a given `Jekyll::Site` instance `site`, calling `site.now` should not *set* `site.time`. Only `site.reset` should do that..
So can I request you to amend your commit to simply do the following:

```ruby
def reset
  self.time = now
  self.layouts = {}
  self.pages = []
  [...]
end

# preferably a private method
def now
  # your implementation
end
```

straight-shoota:
It would be nice to get support for `SOURCE_DATE_EPOCH`. It's a shame that Jekyll doesn't support this yet 😢

Issue #6948, opened 2018-04-22
Jekyll 4.0 Ideas


oe:
:wave: Hello everyone! Summer is coming, and so is the implementation period for Jekyll 4.0. That's right, it's time to get some breaking changes in. To accommodate for this, we're opening this issue to collect interesting ideas that people would like to see implemented in 4.0. __Keep in mind that even if an idea receives a lot of support, there's no guarantee that it'll get implemented__ — that depends on if someone is free to actually implement it. We're all volunteers here, keep that in mind.

Feel free to revive old feature requests, too, just not something that we've explicitly rejected.

For an organized view of how we're consolidating ideas and features, check out our Project board: https://github.com/jekyll/jekyll/projects/2

laukstein:
I vote for #6293 Markdown links bidi support, in `<>`, `[]()` by default.

laukstein:
Custom HTTP headers compatibility for Github Pages (Webrick isn't planned to support https://github.com/github/pages-gem/issues/415)

pathawks:
@laukstein Let’s keep this restricted to Jekyll features. We don’t have any control over GitHub Pages.

letrastudio:
I'd love to see automatic generation of tag and category archive pages in core (plugins like [jekyll-archives](https://github.com/jekyll/jekyll-archives/) do this, but none are whitelisted in GitHub Pages). I wrote up #6952 with my detailed rationale and a proposal of how it could work.

Here's the TL;DR:

> Having vanilla Jekyll auto-generate archive pages for every tag and category is probably a good idea now, and I think I've come up with a good, clean way to do it while staying true to Jekyll's philosophy.

Code-wise, this snippet gets most of it across:

```yaml
collections:
  tags:
    output: true
    permalink: /tags/:name/
  categories:
    output: true
    permalink: /categories/:name/
```

mbrav:
I know that this is markdown and not necessarily jekyll related, but my life would have been a thousand times easier if jekyll supported inline footnotes like so  ‘[^Footnote, p. 123]’ as apposed to the more tedious:

‘’’
This is first reference[^one], this is the second[^two]

[^one]: Footnote, p.1
[^two]: Footnote, p.2

‘’’

DirtyF:
@mbrav This depedens on the Markdown engine. Please ask  on [Kramdown](https://github.com/gettalong/kramdown) or [CommonMark](http://commonmark.org/).

asipple1:
[...]

PR #7188, opened 2018-08-19
Always exclude certain paths from being processed

### LINKED ISSUE #4791: Inclusion should take priority over exclusion in the configuration
I thought I had this figured out, but it doesn't seem to be working. If there's any more information I can provide to help facilitate this, don't hesitate.

Thanks again @envygeeks for the Twitter help earlier!

3.1.2

Windows 10
###### What did you do?

My `_config` contains the following:

```
exclude: [assets/node_modules, assets/bower_components]
include: [assets/bower_components/jQuery/dist]
```

I expected the subfolders specified in the include option to take priority over the excluded parent directory.

The exclude option takes total priority. `_site` doesn't contain a `bower_components` folder whatsoever.

ashmaroli:
Fixes #4791 

```ruby
DEFAULT_EXCLUDES = [
  ".sass-cache", ".jekyll-cache",
  "gemfiles", "Gemfile", "Gemfile.lock",
  "node_modules",
  "vendor/bundle/", "vendor/cache/", "vendor/gems/", "vendor/ruby/"
]
```

## Pros
  - This will simplify users' overhead when defining the `exclude` array in their config file(s). e.g. Accidentally defining a `exclude: ["Gemfile"]` setting will never halt the build by processing contents of  the `vendor` directory.

## Cons
  - This will no longer allow *advanced users* to **override** the default array &mdash; any entry in the users' `exclude` array is going to be **appended** to the default array. (**`BREAKING-CHANGE`**)

## Tasks
  - [x] Initial implementation
  - [x] Unit tests
  - [x] Edit config file generated via `jekyll new`
  - [x] Update Documentation


kenman345:
This looks nice, would reduce the developer special knowledge that goes with the configuration of the excludes and wondering why you need all those, or not understanding why setting one file in the exclude automatically means you had more than before. Themes get a little cleaner too. I like!

What happens if someone wants to include these items though? Can you be sure to add tests/document what happens if these same items are marked in the `include` key? Who wins?

ashmaroli:
Thank you @kenman345 though the `site.include` is still a gray area. Even by **explicitly including 
 `node_modules/bazinga.js`**, one would not be able to copy that file to destination..

ashmaroli:
Update: With https://github.com/jekyll/jekyll/pull/7188/commits/aac927d36e9438e656afc89e3f79da84c0ee25d3, `site.config["include"]` can override `site.config["exclude"]` with fine-grain control:
For example, for a site configured to exclude `node_modules/` directory
  - to include all entries of `node_modules/`, simply add the directory to the `include` array.
  - to include only a couple of entries from `node_modules/` directory, simply list the individual relative paths in the array

:tada:

kenman345:
Is this actually a breaking change? Since it just makes for less explicit configuration its more technical debt of prior configuration files to include all these items in the exclude key, but is it really a breaking thing since it would still work?

kvz:
From https://github.com/jekyll/jekyll/issues/6948#issuecomment-419928579:

> @kvz I've submitted a PR that should handle what you're looking for, as a side-effect.. Would you be able to give that branch a test-run..?
>
># Gemfile
>gem 'jekyll', git: 'https://github.com/jekyll/jekyll.git', ref: 'refs/pull/7188/head'
>
>Feedback invited at the PR's url

Thanks @ashmaroli this looks really good. Unfortunately we're not a Ruby shop (and i have no Ruby skills). My Ruby setup is a big smiling pile of horsedoo with rbenv/rvm/gem/permission conflicts due to this :) 

We've been happily using the Docker container of Jekyll to work around this and build static websites for our open source products.

If there was a docker container i'd love to try it out! But short of that, looking at this code, I _think_ this will indeed solve the problem we voiced in #4791 and it's going to be massively beneficial in speeding up Jekyll for our use case.

ashmaroli:
@kvz Updated the cucumber feature to **confirm** that `exclude: "**"` `include: index.md` will indeed generate only `_site/index.html`..

kvz:
💪💪💪🚀❤️

tim-kos:
Awesome beans, dear Sirs! Xmas + Easter + Birthday feelings here. <3

docs/_docs/upgrading/3-to-4.md
```diff
@@ -21,7 +21,32 @@ gem update jekyll
...
+vendor/gems/
+vendor/ruby/
+```
+
+What's new is that this array __cannot be overriden anymore__. You can forcibly include
+directories or files that have been excluded by default, though. This would look like this:
+
+```yml
+include:
+  - node_modules/.bin/**
```
ashmaroli: umm.. actually, I haven't tested if `include` works as expected with wildcards..

oe: oh! should probably do that then :P

ashmaroli: @oe Local tests show that wildcards do not work with includes yet.. (Can you confirm..?)
But I feel fixing that is outside the scope of the current PR...

oe: yeah, i'm just going to change the example in this scenario

ashmaroli:
I'll resolve the merge conflicts once this gets an approval from the "build" team as well..

DirtyF:
@jekyllbot: merge +major

PR #7189, opened 2018-08-20
Replace regex arg to :gsub with a string arg

ashmaroli:
Because, `:gsub` with string arguments are faster..

### DIFF
--- a/lib/jekyll/tags/include.rb
+++ b/lib/jekyll/tags/include.rb
@@ -41,9 +41,9 @@ module Jekyll
           markup = markup[match.end(0)..-1]
 
           value = if match[2]
-                    match[2].gsub(%r!\\"!, '"')
+                    match[2].gsub('\\"', '"')
                   elsif match[3]
-                    match[3].gsub(%r!\\'!, "'")
+                    match[3].gsub("\\'", "'")
                   elsif match[4]
                     context[match[4]]
                   end


oe:
Why does the build time increase?

ashmaroli:
> Why does the build time increase?

Have no idea about that..
This is actually a micro-optimization that shouldn't have **any effect** on the build times..

pathawks:
@jekyllbot: merge +dev

This has a slight but measurable improvement to performance. In the future though, let’s be more measured with performance PRs and avoid any/all “micro-optimizations.”

pathawks:
> Why does the build time increase?

Oops. I was reading the report wrong. Yeah, this makes performance worse rather than better. 🤷‍♂️

ashmaroli:
> this makes performance worse rather than better.

I don't see why it should ever worsen the performance..

oe:
@pathawks Could you rerun Utterson with this PR and a rolled-back version of `master` to see whether these build times are consistent?

ashmaroli:
@oe Hitting the `revert` button above will create a rollback PR. Should I..?

oe:
@ashmaroli Let's hold off on it until we verify that this actually decreased performance.

ashmaroli:
That's the idea:
  1. Create a rollback PR by hitting the `revert` button above.
  2. Wait till Utterson profiles the diff
  3. If results show decreased performance (current `master` is faster), then close PR. If results show increased performance (current `master` is slower), then merge PR..

oe:
Oh, I guess that works too! Go ahead then.

PR #7193, opened 2018-08-23
Add note to "env at build" about multi-config

tordans:
There are two ways that allow to work with Jekyll and multiple environments. The section "Specifying a Jekyll environment at build time" is the best place to learn about those. But until now, this section only described one of the two options. This change add a note-area to the page that references the other option. I also removed the paragraph about the "avoid editing the config file", since the second option that is now described, is the better solution for this use case.

This edition is based on the feedback in comment https://github.com/jekyll/jekyll/issues/6948#issuecomment-415078430.

This PR tries to extend the docs with the least changes possible.

The other solution would be to re-write the section "Specifying a Jekyll environment at build time" into something like "Working with multiple environments" which then explains both solutions with their pro-contra. However, in such a case I wonder why anyone would even use the JEKYLL_ENV option, since the config-solution will always solve both use cases (content and configuration). – In any case, I hope this minimal change can be merged regardless of a possible broader rework of the section.

### DIFF
--- a/docs/_docs/configuration.md
+++ b/docs/_docs/configuration.md
@@ -461,6 +461,8 @@ The default value for `JEKYLL_ENV` is `development`. Therefore if you omit `JEKY
 
 Your environment values can be anything you want (not just `development` or `production`). Some elements you might want to hide in development environments include Disqus comment forms or Google Analytics. Conversely, you might want to expose an "Edit me in GitHub" button in a development environment but not include it in production environments.
 
+By specifying the option in the build command, you avoid having to change values in your configuration files when moving from one environment to another.
+
 <div class="note info">
   <p>
     <code>JEKYLL_ENV</code> allows for environment specific content. In case you need environment specific configuration options, use the <a href="#build-command-options">build command option</a>, for example <code>--config _config.yml,_config.development.yml</code>. Settings in later files override settings in earlier files.
--- a/docs/_docs/configuration.md
+++ b/docs/_docs/configuration.md
@@ -461,7 +461,11 @@ The default value for `JEKYLL_ENV` is `development`. Therefore if you omit `JEKY
 
 Your environment values can be anything you want (not just `development` or `production`). Some elements you might want to hide in development environments include Disqus comment forms or Google Analytics. Conversely, you might want to expose an "Edit me in GitHub" button in a development environment but not include it in production environments.
 
-By specifying the option in the build command, you avoid having to change values in your configuration files when moving from one environment to another.
+<div class="note info">
+  <p>
+    <code>JEKYLL_ENV</code> allows for environment specific content. In case you need environment specific configuration options, use the <a href="#build-command-options">build command option</a>, for example <code>--config _config.yml,_config.development.yml</code>. Settings in later files override settings in earlier files.
+  </p>
+</div>
 
 ## Front Matter defaults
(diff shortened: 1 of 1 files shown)


docs/_docs/configuration.md
```diff
@@ -461,7 +461,11 @@ The default value for `JEKYLL_ENV` is `development`. Therefore if you omit `JEKY
 
 Your environment values can be anything you want (not just `development` or `production`). Some elements you might want to hide in development environments include Disqus comment forms or Google Analytics. Conversely, you might want to expose an "Edit me in GitHub" button in a development environment but not include it in production environments.
 
-By specifying the option in the build command, you avoid having to change values in your configuration files when moving from one environment to another.
```
DirtyF: Let's not remove this sentence. `JEKYLL_ENV=production bundle exec jekyll build` is [commonly used](https://github.com/search?l=Shell&o=desc&q=JEKYLL_ENV&s=indexed&type=Code), and does not prevent to use multiple configuration files at the same time.

tordans: I feel like this is redundant, makes it harder to understand and also kind of wrong.

What is says is "you can move stuff from the config inside the view to then use liquid in your view with inline-if-else to change it, based on the jekyll.environment."

Why should the docs direct people to move config values in the views, if there is a clean way to have multiple config values per env.

Since I now explain this solution one sentence below, I feel like its harder to understand with both options there.

docs/_docs/configuration.md
```diff
@@ -461,7 +461,11 @@ The default value for `JEKYLL_ENV` is `development`. Therefore if you omit `JEKY
...
+<div class="note info">
+  <p>
+    <code>JEKYLL_ENV</code> allows for environment specific content. In case you need environment specific configuration options, use the <a href="#build-command-options">build command option</a>, for example <code>--config _config.yml,_config.development.yml</code>. Settings in later files override settings in earlier files.
```
ashmaroli: *Novice users* are probably going to be confused as to what qualifies as `environment specific content` and `environment specific configuration options`. You may want to add a paragraph explaining those with a simple example.

Existing templates *may have* a Liquid block like the following:
```
{% if jekyll.environment == "production" %}
  Render this!
{% else %}
  Render something else
{% endif %}
```
The above is not a wrong usage, and `jekyll.environment` can only be altered via `JEKYLL_ENV` env.

tordans: @ashmaroli My goal is to stitch the missing link between the "build-command-options" and this area. The section above your comment explains in detail how jekyll.environment works; I did not change this. My change is at the end of the section, linking to a different way to solve very similar things.

You comment feels like a rework of the whole paragraph https://github.com/jekyll/jekyll/blob/67c597c5b42018843e70fe514cfe6c8104526ac0/docs/_docs/configuration.md#specifying-a-jekyll-environment-at-build-time which IMO should be a different PR.

docs/_docs/configuration.md
tordans: @ashmaroli thats very interesting and a strong +1 to re-work this whole paragraph. However, my change only tries to fix the bug which is a missing link between two solutions.

tordans:
Close since the docs changed recently. See https://github.com/jekyll/jekyll/pull/7266.

Issue #4901, opened 2016-05-15
Incremental generation not working


matthewmcvickar:
- I believe this to be a bug, not a question about using Jekyll.
- I Updated to the latest Jekyll (or) if on Github Pages to the latest `github-pages`
- I Read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- I am on (or have tested on) ***Mac OS*** 10+
- It was another bug.

## My Reproduction Steps

- Run initial build:

    ```shell
    jekyll build --incremental --watch
    Incremental build: enabled
        Generating...
                            done in 32.004 seconds.
    Auto-regeneration: enabled for '/Users/matthewmcvickar/Sites/foo'
    ```

- Change the most recent (or any post) post (insert a new line or character, whatever).

    ```shell
    Regenerating: 1 file(s) changed at 2016-05-15 11:23:13 ...done in 32.837523 seconds.
    ````

## The Output I Wanted

This output should have taken less than a second, since only file was changed. It took as long as the initial generation, however, so it seems as though the entire site has been regenerated.

## Configuration Details

Here is my configuration (personal information like `title` and `author` configurations excluded from the top):

```yaml
baseurl:     ''
url:         localhost
permalink:   /:year/:month/:title/

# Jekyll Build Settings.
# Don't include these files and directories when building the site.
exclude: [
  'docs',
  'Gemfile',
  'Gemfile.lock',
  'Gruntfile.js',
  'node_modules',
  'package.json',
  'README.md'
]

# YAML Front Matter Defaults.
# These settings will set the YAML front matter for all pages and posts. This
# allows us to not have to set `layout: [page\post]` on each page and post.
defaults:
-
  scope:
    type: 'pages'
  values:
    layout: 'page'
    is_page: true
-
  scope:
    type: 'posts'
  values:
    layout: 'post'
    is_post: true

# Enable Git-Flavored Markdown.
# Kramdown is the default Markdown renderer.
kramdown:
  input: GFM


# --- #


# OctoPress.
# These settings determine how OctoPress will generate posts when using the
# `octopress new [page/post]` commands.
post_ext:    'md'
page_layout: ''
post_layout: ''


# --- #


# Archives.
jekyll-archives:
  enabled:
    - year
    - month
    - categories
  layouts:
    year:     archive-time
    month:    archive-time
    category: archive-taxonomy
  permalinks:
    year:     '/:year/'
    month:    '/:year/:month/'
    category: '/:name/'


# --- #


[...]

PR #7201, opened 2018-08-28
Hopeful patch for transient --incremental issue

steinbachr:
This PR is in response to an issue I've seen popup transiently (most often in CI containers) whereby regenerating a site using the `--incremental` flag and with no changes isn't seeing a speedup in build time (basically the issue described in https://github.com/jekyll/jekyll/issues/4901). 

After injecting some logging statements, I sourced the issue to the case when path dependencies had been modified. In this case, the metadata was never being updated to reflect the new `mtime` of the dependency. Therefore, `--incremental` was never able to see build time improvements, even on subsequent runs.

Sorry no test case is included, my Ruby-fu is pretty weak as is and I didn't think I'd be able to cobble anything together in a reasonable timeframe.

### DIFF
--- a/lib/jekyll/regenerator.rb
+++ b/lib/jekyll/regenerator.rb
@@ -178,15 +178,19 @@ module Jekyll
     def existing_file_modified?(path)
       # If one of this file dependencies have been modified,
       # set the regeneration bit for both the dependency and the file to true
+      deps_modified = false
       metadata[path]["deps"].each do |dependency|
-        return cache[dependency] = cache[path] = true if modified?(dependency)
+        if modified?(dependency)
+          add(dependency)
+          deps_modified = true
+        end
       end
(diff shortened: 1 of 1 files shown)


lib/jekyll/regenerator.rb
```diff
@@ -178,15 +178,19 @@ def regenerate_document?(document)
     def existing_file_modified?(path)
       # If one of this file dependencies have been modified,
       # set the regeneration bit for both the dependency and the file to true
+      deps_modified = false
       metadata[path]["deps"].each do |dependency|
-        return cache[dependency] = cache[path] = true if modified?(dependency)
+        if modified?(dependency)
+          add(dependency)
```
mattr-: does `add` here have the same behavior as the above line?

PR #7204, opened 2018-08-29
Feature/hash based generation

steinbachr:
This is related to PR https://github.com/jekyll/jekyll/pull/7201.

Even after applying the patch from the referenced PR, we were still seeing issues in CI (CircleCI to be specific) containers with `incremental` builds not seeing speedups. Specifically, when using CircleCI `restore_cache`/`save_cache` functionality to persist the state of `.jekyll-metadata`/`.jekyll-cache`. The cause, it would appear, is that whenever the cache was restored, the `mtime` of the restored files were set to the current time, while the `.jekyll-metadata` stored the true mtime's.

This PR introduces a generic solution by adding a new configuration option - `incremental_key` - which (for now) can be one of `mtime` (default) or `hash`. When set to hash, the SHA256 hash of the file contents is used as the basis for comparison when determining whether a file has been modified.

### DIFF
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -44,6 +44,7 @@ module Jekyll
       "lsi"                 => false,
       "excerpt_separator"   => "\n\n",
       "incremental"         => false,
+      "incremental_key"     => "mtime",
 
       # Serving
       "detach"              => false, # default to not detaching the server
--- a/lib/jekyll/regenerator.rb
+++ b/lib/jekyll/regenerator.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+require 'digest'
 
 module Jekyll
   class Regenerator
@@ -33,6 +34,16 @@ module Jekyll
       end
     end
 
+    # Get the value for 'path' indicating whether or not it has been modified
+    #
+    # Returns any serializable
+    def mkey(path)
+      return File.mtime(path) unless site.incremental_key == 'hash'
+
+      sha256 = Digest::SHA256.file path
+      return sha256.hexdigest
+    end
+
     # Add a path to the metadata
     #
     # Returns true, also on failure.
@@ -40,7 +51,7 @@ module Jekyll
       return true unless File.exist?(path)
 
       metadata[path] = {
-        "mtime" => File.mtime(path),
+        "mkey" => mkey(path),
         "deps"  => [],
       }
       cache[path] = true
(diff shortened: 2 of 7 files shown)


ashmaroli:
@steinbachr So, were you able to achieve a significant amount of performance boost by applying this patch..?

steinbachr:
@ashmaroli yeah as mentioned in the PR, the `incremental` feature wasn't yielding any speedup at all. In our particular project, builds went from 300s -> 14s when using the `hash` incremental key introduced by this PR.

parkr:
Hashing a file will always be more reliable. I’m curious how this compares to mtime checks in performance in the best case. What system calls are required?

steinbachr:
Canonically, it seemed the build times were comparable.

That said, via `incremental_key` you can switch easily between compiling with a hash based vs. mtime key, so benchmarking shouldn't be too tricky.

parkr:
Ok, I’ll do some testing! If it’s just as oerformant, i would recommend moving to only use hashes in Jekyll 4

DirtyF:
> Canonically, it seemed the build times were comparable.

Utterson shows it's a tiny bit slower, would need more benchmark though.

ashmaroli:
@DirtyF I wouldn't rely on Utterson's report for changes to `--incremental` feature because Utterson doesn't build in incremental mode at all..

ashmaroli:
@steinbachr Apologies for leaving this idle for so long.
I would like to test this out locally on a regular system (outsider container / VM / Docker).
Do you remember the steps your team took to come up with the conclusion that build times dropped by 200+secs for the incremental build with Digest hashes..?

steinbachr:
@ashmaroli sorry this was a while ago, so the details are fuzzy. However, I imagine a large part of the reason why we saw such a drastic improvement is simply that our project is very large for a Jekyll project (> 4,000 files), so uncached builds are very slow (on the order of 4m)

mattr-:
I'll pick this up and run with it. It won't land in Jekyll 4.0, but probably 4.1

mattr-:
TODO for this PR:
 1. Fix the conflicts
 1. Switch to hashing by default
 1. Adapt metadata that contains `mtime` to use `mkey` instead, on the fly at build time, so that we're forward compatible.

If the original author doesn't want to pick up this work, another person would need to pull this PR down to their local copy, make any changes, and push up a new branch and open a new PR.

PR #7211, opened 2018-09-02
Small benchmark refactoring

vbrazo:
Refactoring of some duplicated lines in the `benchmark/schwartzian_transform.rb`

### DIFF
--- a/benchmark/schwartzian_transform.rb
+++ b/benchmark/schwartzian_transform.rb
@@ -90,26 +90,21 @@ end
 Correctness.new(site_docs, "redirect_from".freeze).assert!
 Correctness.new(site_docs, "title".freeze).assert!
 
-# First, test with a property only a handful of documents have.
-Benchmark.ips do |x|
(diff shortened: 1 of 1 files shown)


ashmaroli:
@vbrazo Please explain the exact code smell that is being fixed here..

ashmaroli:
@vbrazo Oh, is that all.. I would've refactored it in the following manner, but your implementation is better.. :+1: 

```ruby
SAMPLE = {
  "sparse"     => "redirect_from",
  "non-sparse" => "title"
}

SAMPLE.each do |property, meta_key|
  Benchmark.ips do |x|
    x.config(time: 10, warmup: 5)
    x.report("sort_by_property_directly with #{property} property") do
      sort_by_property_directly(site_docs, meta_key)
    end
    x.report("schwartzian_transform with #{property} property") do
      schwartzian_transform(site_docs, meta_key)
    end
    x.compare!
  end
end
```

DirtyF:
@jekyllbot: merge +dev

PR #7214, opened 2018-09-02
Add some minor improvements to image loading in Showcase page

codegaze:
Hey, great work on the showcase page! :)

I did some minor changes to the loading states of this page's (`/showcase/`) images. There isn't any functional change but rather visual. 

- Removes the loading animated image from the "showcases" view and sets a default height percentage for each image to avoid page "jumping" on image loading.
- Added a small "spacer" gif in order to have a src to load on render and a background color. When the image loads has a small fade in effect.

Didn't know if there is a specific way/logic to add new styles so I improvised a bit and tried to minimize changes that might affect something else.

Cheers 🍻


DirtyF:
Neat and smooth, thanks @codegaze 

@jekyllbot: :shipit: +docs

PR #7216, opened 2018-09-03
fixes a typo?

jlvallelonga:
"Shows a help message or these commands" should be "Shows a help message for these commands" I think

### DIFF
--- a/docs/_docs/usage.md
+++ b/docs/_docs/usage.md
@@ -13,7 +13,7 @@ You can use this command in a number of ways:
 * `jekyll doctor` - Outputs any deprecation or configuration issues
 * `jekyll new-theme` - Creates a new Jekyll theme scaffold
 * `jekyll clean` - Removes the generated site and metadata file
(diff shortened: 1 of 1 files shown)


DirtyF:
This isn't a typo, jekyll help also displays commands listed on the page. Thanks for watchin'.

jlvallelonga:
Oh did you mean to say "Shows a help message and these commands" then? The word "or" in this circumstance would indicate alternatives meaning that either it shows a help message or it shows these commands, not both. I don't think that usage really makes sense for a command that returns the same result every time you use it. It's confusing. Maybe the description from the actual `jekyll help` result would be better here: "Show the help message, optionally for a given subcommand."

Not a huge deal, but figured it was confusing for me so it might be confusing for someone else.

DirtyF:
@jlvallelonga fixed in https://github.com/jekyll/jekyll/commit/56dd010c8293b77209e87e4294bffc204a8e7ca3 thanks for your feedback.

PR #7223, opened 2018-09-06
removes quotes from markdown for assets

jlvallelonga:
Having quotes in the markdown was breaking the references to images and other assets


DirtyF:
Thanks @jlvallelonga 😄 

@jekyllbot merge +docs

PR #7225, opened 2018-09-07
Add support.crowdin.com to showcases page

anton-yurkiv:
Crowdin - localization management platform
Docs site built with jekyll
localization powered by jekyll-multiple-languages-plugin and Crowdin


DirtyF:
Thanks for using Jekyll.

As stated the showcase already list almost of *our* favorite websites, and isn't meant to list all Jekyll websites. We might revisit how we display the showcase and re-open this PR, but right now I feel like it deserves its purpose: show that Jekyll is used in many great ways.

PR #7224, opened 2018-09-06
3.7.x: EntryFilter#filter symlink fix

parkr:
In `EntryFilter#filter`, we check for `include?` before we check for `symlink?`, which allows symlinks to be read in a build when they shouldn't be by just including them.

I am committing a failing test in one commit, and a fix in another commit so we can demonstrate what fails.

PR #7226, opened 2018-09-07
Security: fix `include` bypass of `EntryFilter#filter` symlink check

parkr:
In `EntryFilter#filter`, we check for `include?` before we check for `symlink?`, which allows symlinks to be read in a build when they shouldn't be by just including them.

/cc https://github.com/jekyll/jekyll/pull/7224


DirtyF:
👍 once we appease Rubocop

parkr:
@jekyllbot: merge +bug

PR #7228, opened 2018-09-07
3.8.x: security: fix `include` bypass of `EntryFilter#filter` symlink check

parkr:
Backports #7226 for 3.8.x series of releases.


ashmaroli:
Why is the base branch for this `master` instead of `3.8-stable`..?

PR #7229, opened 2018-09-07
3.6.x: security: fix `include` bypass of `EntryFilter#filter` symlink check

parkr:
Backports #7226 for the 3.6.x series of releases.


DirtyF:
@jekyllbot: merge +bug

Issue #6578, opened 2017-11-26
Make site.source the cynosure of a Jekyll Site


ashmaroli:
Related: #6060 

Basically, replace dependency on `Dir.pwd` with `site.source` as the primary focus.
For example, the **ultimate paths** just before processing ought to be..:
```ruby
# options denote derivation from CLI switches

config["source"]          = options["source"] || Dir.pwd
config["destination"]     = options["destination"] || File.join(config["source"], "_site")
config["layouts_dir"]     = options["layouts"] || File.join(config["source"], "_layouts")
config["includes_dir"]    = File.join(config["source"], "_includes")
config["data_dir"]        = File.join(config["source"], "_data")
config["collections_dir"] = File.join(config["source"], "")
config["plugins_dir"]     = options["plugins_dir"] || File.join(config["source"], "_plugins")
```

mattr-:
Agreed. Everything should revolve around `site.source`

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>3.3-stable</code> or <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, please consider building it first as a plugin. Jekyll 3 introduced [hooks](http://jekyllrb.com/docs/plugins/#hooks) which provide convenient access points throughout the Jekyll build pipeline whereby most needs can be fulfilled. If this is something that cannot be built as a plugin, then please provide more information about why in order to keep this issue open.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

mattr-:
I really want to get 4.0 out the door, so i'm going to push this back. It'll probably end up being a breaking change, so 5.0 for this one.

PR #7232, opened 2018-09-09
Configure cache_dir

### LINKED ISSUE #7220: Custom `cache_dir` configuration not honored
1. `jekyll new cache-test --skip-bundle && cd cache-test`
2. Edit `./Gemfile` to point to latest `jekyll/jekyll:master`
3. `bundle install`
4. Edit `./_config.yml` by inserting `cache_dir: ".foo-cache"`
5. `bundle exec jekyll build`
6. **Observe that `./.jekyll-cache` directory is created instead of `./.foo-cache` directory**

ashmaroli:
Resolves #7220 

Additionally, this **ensures** that the cache is always "within the source_dir" and also sanitizes the `cache_dir` configuration..

### DIFF
--- a/features/cache.feature
+++ b/features/cache.feature
@@ -1,0 +1,28 @@
+Feature: Cache
+  As a developer who likes to create plugins
+  I want to be able to cache certain aspects across multiple builds
+  And retrieve the cached aspects when needed
+
+  Scenario: Default Cache directory
+    Given I have an "index.md" page that contains "{{ site.title }}"
+    And I have a configuration file with "title" set to "Hello World"
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the .jekyll-cache directory should exist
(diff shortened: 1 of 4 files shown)


ashmaroli:
@jekyllbot: merge +minor

Convincible:
Why was it important that the cache directory be in the source directory? I want to store my cache somewhere else entirely!

ashmaroli:
> Why was it important that the cache directory be in the source directory?

- Partially influenced by https://github.com/jekyll/jekyll/issues/6578
- Since, this is a very new feature, I felt limiting its flexibility will allow loosening as required.
- If not at `source_dir`, the other location would be at `$PWD` which (in my head) appeared to be brittle implementation of the *"disk-cache"*. My `$PWD` could be anywhere and Jekyll would build the site properly as long as I provided proper values to the `--source` CLI switch. So, there was a chance that (all those) location(s) would be littered with `.jekyll-cache` directories in turn reducing its effectiveness.

PR #7240, opened 2018-09-11
WIP Add support for basic localization

ashmaroli:
*Disclaimer: Not a ground-breaking feature :wink:*

---

**This feature would simplify creating and maintaining templates and themes with locale support via Jekyll's data directory.**

### Example Case

For example, in the popular theme [Minimal Mistakes](https://github.com/mmistakes/minimal-mistakes) by @mmistakes, one would find numerous instances of providing locale support by using [the following syntax](https://github.com/mmistakes/minimal-mistakes/blob/fc3e7a61769253f10d325525b0a7a334166f49ac/_includes/comments.html#L2);
```
{{ site.data.ui-text[site.locale].comments_label | default: "Comments" }}
```

With the proposed native support in Jekyll, @mmistakes (or a user customizing the template) can achieve the same result from the following syntax:
```
{{ locale.comments_label | default: "Comments" }}
```
along with the following configuration for compatibility:
```yaml
# _config.yml

locales_dir: ui-text
locale: en-US
```

### DIFF
--- a/features/locales.feature
+++ b/features/locales.feature
@@ -1,0 +1,78 @@
+Feature: Localization Support
+  As a polyglot blogger who likes to write
+  I want to be able to easily switch my site localization
+  And render non-content strings accordingly
+
+  Scenario: Default configuration
+    Given I have an "index.md" page that contains "{{ locale.greeting }} Visitor!"
+    And I have a _data/locales directory
+    And I have a "_data/locales/en.yml" file that contains "greeting: Hello"
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "Hello Visitor!" in "_site/index.html"
+
+  Scenario: Default configuration with custom locale
+    Given I have an "index.md" page that contains "{{ locale.greeting }} Visitor!"
+    And I have a configuration file with "locale" set to "pirate"
+    And I have a _data/locales directory
+    And I have a "_data/locales/pirate.yml" file that contains "greeting: Ahoy there"
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "Ahoy there Visitor!" in "_site/index.html"
+
+  Scenario: Custom configuration and custom locale
+    Given I have an "index.md" page that contains "{{ locale.greeting }} Visitor!"
+    And I have a configuration file with:
+      | key         | value     |
+      | locales_dir | languages |
+      | locale      | pirate    |
+    And I have a _data/locales directory
+    And I have a _data/languages directory
+    And I have a "_data/locales/pirate.yml" file that contains "greeting: Valar Morghulis"
+    And I have a "_data/languages/pirate.yml" file that contains "greeting: Ahoy there"
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "Ahoy there Visitor!" in "_site/index.html"
+
+  Scenario: Custom configuration and custom locales in a single data file
+    Given I have an "index.md" page that contains "{{ locale.greeting }} Visitor!"
+    And I have a configuration file with:
+      | key         | value     |
+      | locales_dir | languages |
+      | locale      | pirate    |
+    And I have a _data directory
+    And I have a "_data/languages.yml" file with content:
+    """
+    en:
+      greeting: Welcome
+
+    pirate:
+      greeting: Ahoy there
+    """
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "Ahoy there Visitor!" in "_site/index.html"
+
+  Scenario: Locale data file with atypical keys
+    Given I have an "index.md" page that contains "{{ locale.good_morning }} Visitor!"
+    And I have a configuration file with:
+      | key         | value     |
+      | locales_dir | languages |
+      | locale      | polyglot  |
+    And I have a _data directory
+    And I have a "_data/languages.yml" file with content:
+    """
+    polyglot:
+      good_morning : Good Morning
+      Good Morning : Goede morgen
+      Good-Morning : Bonjour
+      Good.Morning : Buenos días
+    """
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "Buenos días Visitor!" in "_site/index.html"
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -71,6 +71,7 @@ module Jekyll
   autoload :Regenerator,         "jekyll/regenerator"
   autoload :RelatedPosts,        "jekyll/related_posts"
   autoload :Renderer,            "jekyll/renderer"
+  autoload :LocaleHandler,       "jekyll/locale_handler"
   autoload :LiquidRenderer,      "jekyll/liquid_renderer"
   autoload :Site,                "jekyll/site"
   autoload :StaticFile,          "jekyll/static_file"
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -14,6 +14,7 @@ module Jekyll
       "layouts_dir"         => "_layouts",
       "data_dir"            => "_data",
       "includes_dir"        => "_includes",
+      "locales_dir"         => "locales", # always relative to data_dir
       "collections"         => {},
 
       # Handling Reading
@@ -53,6 +54,7 @@ module Jekyll
       "show_dir_listing"    => false,
 
       # Output Configuration
+      "locale"              => "en",
       "permalink"           => "date",
       "paginate_path"       => "/page:num",
       "timezone"            => nil, # use the local timezone
--- a/lib/jekyll/drops/locale_drop.rb
+++ b/lib/jekyll/drops/locale_drop.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+module Jekyll
+  module Drops
+    class LocaleDrop < Drop
+      extend Forwardable
+
+      mutable false
+      private def_delegator :@obj, :data, :fallback_data
+    end
+  end
+end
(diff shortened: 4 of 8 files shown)


oe:
This looks great! I'm having a little bit of trouble understanding what the use-case for this is. Don't most people who want to add localization to their site want to generate two or more locales at once? If so, I don't see much of a point in a glorified `site.data` abstraction. If there's a possibility that we could build on this PR to enhance i18n features, though, then I'm all for it.

ashmaroli:
> bit of trouble understanding what the use-case for this is.

The specific use-case for this has been illustrated in the PR's main-post.

> Don't most people who want to add localization to their site want to generate two or more locales at once?

Yes, I agree with that.The reasons why I chose this as **the starting point** are:
  - I18n is a complex being. The best way to tame it is via baby-steps..
  - For Jekyll as a handy tool, this proposal lays the foundation for basic i18n support in the Core. Users would simply have to define and maintain their own "translations" as data files.
They're already able to do this in Jekyll 3.x, But this proposal will add an API (the `{{ locale }}` namespace) that will streamline usage and maintenance based on current configuration.

> If there's a possibility that we could build on this PR to enhance i18n features

**Theoretically**, in the future, `Jekyll::LocaleHandler` can have an *attribute set* by iterating through an array of "preferred_locales" which can then be utilized to render locale-specific copies of the current site..
But all that will require a lot more brainstorming...

DirtyF:
We should favor convention over configuration here and define where to store translations by default (e.g `_i18n` or `_locales`).

ashmaroli:
> We should favor convention over configuration here

I chose `<source>/_data/locales` for simplicity here: **Piggy-back on existing infrastructure.**
That avoids having to add more code and logic to:
  - consider `<source>/_locales` as a special directory.
  - consider contents of `<source>/_locales` as bona fide primitives (neither a `Page`, nor a `Document`, nor a `StaticFile`) instead of just being a data-file.

Or if your concern was to have `<source>/_data/_locales` as the default locales_dir, I do not see the point of having a directory with a leading underscore inside a designated special directory...

But that said, considering the recent developments with converting `_data` into a default collection (which I'm not a fan of), I have no further comments..

ashmaroli:
@letrastudio I just chanced upon [your comment](https://github.com/jekyll/jekyll/issues/6948#issuecomment-415056402) in the now frozen wishlist for Jekyll 4.0 and thought you'd like to see this..
Would this PR benefit / conflict with the work you're doing on i18n..?

letrastudio:
Thanks for letting me know @ashmaroli, I hadn’t come across this PR, and I certainly can contribute to the discussion!

I agree a `{{ locale }}` global variable is the right way to go on the Liquid side, but I think this PR is aiming too small with what it should encompass. String localization is very important, but it’s just one part of i18n (some i18n sites might not even use multiple languages). If string localization were to be Jekyll's first step into i18n, I'd recommend something like `{{ locale.strings }}`, allowing for other future attributes like `{{ locale.lang }}` or `{{ locale.baseurl }}`.

Maybe string localization isn't really what the first step into i18n should be — it's already pretty easy to solve in Jekyll, using `_data` as is. What I find to be most difficult is localizing entire documents, and setting up relationships between them. So if I have this `about.md`:

```
---
title: About
image: profile.jpg
email: me@example.com
---
Hello!
```

I should be able to have a localized copy of that document that only includes translations of localizable data: `title` and `content`, keeping it DRY. So right off the bat, assuming that each locale can simply correspond to a parallel build of the site wouldn’t work; to build the localized page, the original page would need to be accessed as well, to get `image` and `email`.

I think solving i18n for entire documents should really be the first step, although I do recognize it’s not quite a _baby step_. I have some ideas on how to do that — more comments to come soon, as well as some code examples!

DirtyF:
Ideally Jekyll could map translations either via filenames or directories : 

```about.md
about.es.md
about.fr.md

about.md
/es/about.md
/fr/about.md

**Comment by letrastudio** [comment]
@ashmaroli I haven't yet had the chance to test out the code, so my opinion is based mostly on the feature file — I might not be fully grasping all the implications of what you're trying to do here.

It's exciting to see progress on this front! Hopefully I'll have time over the weekend to look at this more closely. I also plan to publish a demo repo showing how I'm dealing with i18n in Jekyll 3, which I'm hoping will help scope out the problem.

A quick note, though:

> I feel those should be handled under the hood without being visible in the templates..

My experience with i18n tells me we should be careful when abstracting away certain concepts for the sake of ease of use. Because of how diverse locales can be, the approach should be as hands-off as possible in order to avoid locking out certain languages, writing systems, etc.

I'm leaning towards locales being defined in _config.yml like collections, allowing for arbitrary metadata:

```yaml
locales:
  default:
    lang: en
    name: English
  pt:
    lang: pt-PT
    name: Português
  zh:
    baseurl: /mo
    lang: zh-Hant-MO
    name: 繁體中文
  ar:
    baseurl: /ae
    lang: ar-AE
    name: عربى
    direction: rtl
```

We could then loop through `site.locales` to build things like language switcher menus. And write this sort of thing in templates:

```liquid
<html lang="{{ locale.lang }}" dir="{{ locale.direction | default: 'ltr' }}">
```

But I absolutely agree that translated text strings should go in `_data`, as you suggest.

ashmaroli:
@letrastudio I switched into `[WIP]` because this shouldn't be merged in its current state.
I'm trying to upgrade the functionality from being *just syntactic sugar* to *something more..*

Your feedback has been very enlightening so far.. :+1:

ashmaroli:
I've extracted this proposal into a separate repository: https://github.com/ashmaroli/jekyll-locale
This will allow collecting ideas / contributions / discussions and implementing functionality without tampering Jekyll's git log..

Whether it gets *upgraded* into an official plugin is not a concern at the moment..

letrastudio:
Okay, I finally took the time to finish up my ideas and put everything together in a repo: [letrastudio/jekyll-i18n](https://github.com/letrastudio/jekyll-i18n). It’s a live [GitHub Pages site](
https://letrastudio.github.io/jekyll-i18n/) and aims to be a thorough example using four locales.

I’ve been working on this on-and-off since April, it feels good to finally open source it! 🎉 (thanks to @ashmaroli for the final push I needed to get it done)

The repo includes both my i18n method for Jekyll 3 and a full proposal for 4.0 in [a separate branch](
https://github.com/letrastudio/jekyll-i18n/tree/jekyll-4-proposal). Each branch includes a comprehensive README that should be enough to explain how it all works.

@ashmaroli please take a look if you can and let me know what you think. Some of the core ideas are at odds with what you’re proposing… My solution evolved from trying to do it in Jekyll 3, so some parts might not be optimal, but I think it would work quite well in terms of flexibility and ease of use for Jekyll users. I haven’t yet had a chance to fully delve into your repo, but I plan to do so this weekend — I’ll add some more feedback then.

ashmaroli:
@letrastudio I took a look at [`jekyll-i18n`](https://letrastudio.github.io/jekyll-i18n/).
It's *good* as a community-developed theme that can be forked and customized by a user.
Congratulations on getting it across the finish-line!!

*But unfortunately, its not what Jekyll Core is looking for..*

Jekyll Core is vying to add a *feature* that can be used by any site / theme *with minimal performance drain, with minimal configuration, with minimal maintenance-overhead, and with Ruby bearing the computational brunt instead of Liquid..* (the latter case being very slow for moderate-to-large sites)
That said, the one thing Ruby implementation *cannot do* is reverse the "page layout" for RTL locales. That will have to be handled by a Liquid Template only

I couldn't pinpoint the reason why your proposal for Jekyll-4.0 would not work in Jekyll-3.0.. What new Core Feature is the 4.0-proposal using..?

letrastudio:
@ashmaroli I get the feeling you might have missed the [`jekyll-4-proposal` branch](
https://github.com/letrastudio/jekyll-i18n/tree/jekyll-4-proposal) — sorry if the distinction wasn’t clear.

The `master` branch is meant to be exactly what you describe: throwing lots of liquid at the problem to make it work in Jekyll 3 with no plugins. It’s slow and cumbersome to configure, but can be helpful to this effort by working as a sort of hacked-together prototype.

The `jekyll-4-proposal` branch, while not proposing any *Ruby* code, defines how native i18n APIs could be designed, from a user perspective. None of the code in that branch works with Jekyll 3 as is, and the README is also different. You can even see how much configuration and liquid code would be reduced for the demo site in this [diff](
https://github.com/letrastudio/jekyll-i18n/commit/e90895716a61c3929911f7f0f54f6ced96abcbad).

Let me know if this makes more sense.

ashmaroli:
> The `jekyll-4-proposal` branch, while not proposing any _Ruby_ code, defines how native i18n APIs could be designed, from a user perspective.

Ah! That inference (`how native i18n APIs could be designed`) is what I missed from the initial read-through..
The proposal looks interesting.. I'll *try to incorporate some of the ideas outlined there into the `jekyll-locale`* plugin.

My recommendation though, is to:
  - Go through what the plugin is capable of doing *now*..
  - Split your "proposal" into multiple *bite-sized* "feature-request" issue tickets at the plugin's repository. That way,
    - I can decide whether or not to implement a feature
    - Users reading through a feature-request ticket will be able to easily grok how the feature got conceptualized, and eventually why it got implemented / modified / discarded..

untra:
Hi @ashmaroli @letrastudio 👋 

Just checking in: whats the status of Jekyll 4.0 and the priority of this PR? Will there be a dedicated blessed i18n support for Jekyll going into a release soon? Is this PR going to receive further development?

I manage the [polyglot](https://github.com/untra/polyglot) Jekyll i18n gem, and I've put off updating it in anticipation of this PR and an official jekyll-4.0 blessed i18n solution, and that my effort wouldn't be worth it considering the official ongoing development there.

Otherwise I may reheat these leftovers and take a stab at this myself.

ashmaroli:
@untra I extracted the basis for this PR into a separate gem named [`jekyll-locale`](https://github.com/ashmaroli/jekyll-locale)
Feel free to take its **`master`** branch for a test-ride and provide feedback especially since you have prior experience with I18n.

mattr-:
@untra There won't be anything that lands in Jekyll 4.0 for i18n. I'd recommend updating `polyglot` when you can find the spare cycles to do so. 🙂

PR #7241, opened 2018-09-11
explicit location of where to create blog.html

stephencweiss:
I initially saved my blog.html file to the wrong directory. Making it explicit on row 65 may help other beginners avoid confusion.


DirtyF:
👍 default path is givne for all other files

@jekyllbot: merge +docs

stephencweiss:
Actually it wasn’t. I get that index.html went to root - but it wasn’t immediately clear to me that blog.html would be the same (about.md for example we treated differently ). It’s a small thing, but as a beginner, I made this mistake.

PR #7250, opened 2018-09-16
Re-implement handling Liquid blocks in excerpts

### LINKED ISSUE #7248: Excerpt throws error when `highlight` and `raw` tags are combined at the begining of a document.
After investigating [this SO question](https://stackoverflow.com/questions/52324134/getting-an-liquid-exception-liquid-syntax-error-while-using-jekyll), I can reproduce an error by combining highlight and raw tags.

## Reproduction Steps
bundle exec jekyll new mysite

Replace "Welcome to Jekyll" post content by :

```
---
front matter
---
{% highlight python %}
{% raw %}
  # line 1
  # line 2
{% endraw %}
{% endhighlight %}
```
This works well.

Add a line between two lines of code :

```
---
front matter
---
{% highlight python %}
{% raw %}
  # line 1

  # line 2
{% endraw %}
{% endhighlight %}
```

Jekyll throws :
>  Liquid Exception: Liquid syntax error (line 2): 'raw' tag was never closed in _posts/2018-09-14-welcome-to-jekyll.markdown/#excerpt
  Liquid Exception: Liquid syntax error (minima-2.5.0/_includes/head.html line 2): 'raw' tag was never closed included in /_layouts/default.html
             Error: Liquid syntax error (minima-2.5.0/_includes/head.html line 2): 'raw' tag was never closed included 
             Error: Run jekyll build --trace for more information.

While :

```
---
front matter
---
{% highlight python %}

{% raw %}
  # line 1
  # line 2
{% endraw %}
{% endhighlight %}
```
Is warning :

>  Warning: Excerpt modified in _posts/2018-09-14-welcome-to-jekyll.markdown!
    Found a Liquid block containing separator ' ' and has been modified with the appropriate closing tag.
    Feel free to define a custom excerpt or excerpt_separator in the document's Front Matter if the generated excerpt is unsatisfactory.

ashmaroli:
Resolves #7248 

This re-implementation should be able to handle multiple Liquid blocks in excerpts.

@djacquel I invite you to test this branch against various valid combinations of Liquid blocks so that we can get rid of this issue for good. Thanks in advance.

@jekyll/core I request this PR to be considered a.s.a.p and be backported to the `3.8-stable` series


lib/jekyll/excerpt.rb
```diff
@@ -131,36 +131,44 @@ def render_with_liquid?
     #
     # Returns excerpt String
 
-    LIQUID_TAG_REGEX = %r!{%-?\s*(\w+).+\s*-?%}!m
+    LIQUID_TAG_REGEX = %r!{%-?\s*(\w+).+?\s*-?%}!m
```
mattr-: The capture for this regular expression isn't 100% correct. A liquid tag of `{%highlight%}` will lose the last letter, meaning that the capture group will contain `highligh` instead of `highlight`

I wrote a quick test case for this on Rubular as well, as an extra example: http://rubular.com/r/umPU5g5PiH

lib/jekyll/excerpt.rb
```diff
@@ -169,12 +177,11 @@ def liquid_block?(tag_name)
...
       Jekyll.logger.warn "Warning:", "Excerpt modified in #{doc.relative_path}!"
-      Jekyll.logger.warn "",
-                         "Found a Liquid block containing separator '#{doc.excerpt_separator}'" \
-                         " and has been modified with the appropriate closing tag."
-      Jekyll.logger.warn "",
-                         "Feel free to define a custom excerpt or excerpt_separator in the" \
-                         " document's Front Matter if the generated excerpt is unsatisfactory."
+      Jekyll.logger.warn "", "Found a Liquid block containing separator" \
+                         " #{doc.excerpt_separator.inspect} "
+      Jekyll.logger.warn "", "and has been modified with the appropriate closing tag."
```
mattr-: This line here reads a bit funky. Could you change the warning to say something like the code block below?
```
Found a Liquid block containing the excerpt separator '#{doc.excerpt_separator.inspect}'. The block has been modified with the appropriate closing tag.
```

lib/jekyll/excerpt.rb
ashmaroli: Thank you for catching this @mattr- :slightly_smiling_face:

lib/jekyll/excerpt.rb
ashmaroli: Sure. Not a problem..

ashmaroli:
> do we need to be concerned about liquid object tag pairs (`{{` and `}}`)?

@mattr- No, we shouldn't handle deviation from the formats used by the *majority* of the users: **`{{variable}}`** or **`{{ variable }}`**.

mattr-:
@ashmaroli Sorry, not sure I understood you. Are you saying that since `{%` and `%}` tag pairs are the most common, we don't need to handle `{{` and `}}` tag pairs?

ashmaroli:
> Are you saying that since `{%` and `%}` tag pairs are the most common, we don't need to handle `{{` and `}}` tag pairs?

@mattr- Not at all..
We are handling only `{%` and `%}` for **`Liquid::Block`** type tags. i.e. tags that have an opener
(e.g. **`{% raw %}`**) and a closer (e.g. **`{% endraw %}`**). Regular tags like **`{% assign %}`** are not handled.

ashmaroli:
Oh.. The Utterson Report doesn't look good.. :slightly_frowning_face: 

ref | build time in seconds
-- | --
#7250 | 344.02
master | 329.06

pathawks:
@ashmaroli Correctness wins over performance 👍🏼

lib/jekyll/excerpt.rb
```diff
@@ -169,12 +177,11 @@ def liquid_block?(tag_name)
...
-                         "Found a Liquid block containing separator '#{doc.excerpt_separator}'" \
-                         " and has been modified with the appropriate closing tag."
-      Jekyll.logger.warn "",
-                         "Feel free to define a custom excerpt or excerpt_separator in the" \
-                         " document's Front Matter if the generated excerpt is unsatisfactory."
+      Jekyll.logger.warn "", "Found a Liquid block containing the excerpt separator" \
+                         " #{doc.excerpt_separator.inspect}. "
+      Jekyll.logger.warn "", "The block has been modified with the appropriate closing tag."
+      Jekyll.logger.warn "", "Feel free to define a custom excerpt or excerpt_separator in the"
+      Jekyll.logger.warn "", "document's Front Matter if the generated excerpt is unsatisfactory."
```
pathawks: These are all the same warning, not separate messages, correct?

ashmaroli: @pathawks Yes. These *calls* are for the same warning.
I split them out into separate calls so that the messages appear to be formatted on the CLI.

I can't use a heredoc to emulate the formatting because Jekyll's `logger` squashes all whitespace into a single `space` char, which causes heredoc messages to simply print as a single line message.

oe:
@jekyllbot: merge +bug

PR #7251, opened 2018-09-17
Cache include contents

ashmaroli:
..to limit disk usage (with numerous includes)


ashmaroli:
The results are out!
Utterson doesn't look very impressed with this route:

ref | build time in seconds
-- | --
#7251 | 328.61
master | 327.63

Sad..

PR #7252, opened 2018-09-17
Cache URL Filter results based on given input

ashmaroli:
**In theory**,
- URL Filters are generally more commonly used in various templates than other liquid filters
- Avoids processing the same input for every template
  - Reduced String allocations
  - Reduced calls to `Addressable::URI`

*Disclaimer: Performance benefits depend on use of the filters in place of `prepend: site.baseurl` and
`prepend: site.url`*


ashmaroli:
*Mr. Utterson,
&nbsp; &nbsp; Please assess the changes proposed here send in your report a.s.a.p
&nbsp;
Regards,
A*

PR #7260, opened 2018-09-19
Fix a small grammar error/typo in the docs

jtappa:
I was just re-reading through the docs and found this small blip, thought I'd correct it.


DirtyF:
@jekyllbot: merge +docs

PR #7261, opened 2018-09-20
Search Front matter defaults for Page objects with relative_path

ashmaroli:
So that it's consistent with `Document` objects, `StaticFile` objects and also with `Convertible#to_liquid`
https://github.com/jekyll/jekyll/blob/4271495fcaad4073b2375b76089041acd16ffb49/lib/jekyll/document.rb#L39
https://github.com/jekyll/jekyll/blob/4271495fcaad4073b2375b76089041acd16ffb49/lib/jekyll/static_file.rb#L37
https://github.com/jekyll/jekyll/blob/4271495fcaad4073b2375b76089041acd16ffb49/lib/jekyll/convertible.rb#L119

(This will have a *minor* performance gain in sites with numerous `pages` and uses `front matter defaults` since the paths are cached internally now..)

### DIFF
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -50,7 +50,7 @@ module Jekyll
       read_yaml(File.join(base, dir), name)
 
       data.default_proc = proc do |_, key|
-        site.frontmatter_defaults.find(File.join(dir, name), type, key)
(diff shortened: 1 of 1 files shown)


DirtyF:
@jekyllbot: merge +minor

PR #7263, opened 2018-09-20
Check if scope applies to type before given path

ashmaroli:
Since `#applies?` requires both conditions to be `true`, let us run the more expensive method only if required


oe:
@jekyllbot: merge +bug

PR #7266, opened 2018-09-22
environments.md: reference the build command options that allows multiple config files

tordans:
This is a new take on https://github.com/jekyll/jekyll/pull/7193 after the docs where restructured recently. 

I reference the build command options as a valid solutions to handle environments.

Why is this change important: It took me more than a year an several github issues/comments to understand this feature / learn of its existence. I think the main reason is, I always searched for a solution for multiple environments and only found this page which does not reference the build command that allows multiple config files.


DirtyF:
@jekyllbot: merge +docs

PR #7268, opened 2018-09-23
Update 10-deployment.md

GuillermoYanez:
Fixed HTML <link> tag for styles.css path in "_layouts/default.html" listing.


oe:
@jekyllbot: merge +docs

PR #7270, opened 2018-09-24
Document two methods, simplify one of the methods

GregPK:
Very simple PR. Documented two methods and slightly changed one of them.

Reasons:
 - Undocumented public methods.
 - Memoization of `source_file_mtime` dropped since it was used only once.


DirtyF:
@GregPK Thanks for your first PR on Jekyll 🙏 

@jekyllbot: merge +fix

PR #7272, opened 2018-09-25
Avoid unnecessary duplication of pages array

ashmaroli:
- This is a 🐛 bug fix.

## Summary

`site.pages` is already a flat array of `pages`. Calling `flatten` on it and creating a new Array instance is unnecessary.


DirtyF:
@jekyllbot: merge +fix

PR #7273, opened 2018-09-25
Memoize the return value of Site#documents

ashmaroli:
- This is a :bug: fix 

## Summary

Memoize the return value of Site#documents as it is a potentially expensive method that involves `Enumerable#reduce` and multiple array allocations.

This issue was discovered while building the *"Official Docs Site"*

## Context

Source of *"performance drain"*:
https://github.com/jekyll/jekyll/blob/1e9cd27dc40c26624abb34ad947b314c9571ccc7/lib/jekyll/drops/site_drop.rb#L11

Use of drain in the *"Official Docs Site"*
https://github.com/jekyll/jekyll/blob/1e9cd27dc40c26624abb34ad947b314c9571ccc7/docs/_includes/docs_option.html#L2
https://github.com/jekyll/jekyll/blob/1e9cd27dc40c26624abb34ad947b314c9571ccc7/docs/_includes/docs_ul.html#L3

### DIFF
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -94,6 +94,7 @@ module Jekyll
       self.data = {}
       @site_data = nil
       @collections = nil
+      @documents = nil
       @docs_to_write = nil
       @regenerator.clear_cache
       @liquid_renderer.reset
@@ -321,7 +322,7 @@ module Jekyll
     #
     # Returns an Array of all Documents
     def documents
-      collections.reduce(Set.new) do |docs, (_, collection)|
+      @documents ||= collections.reduce(Set.new) do |docs, (_, collection)|
         docs + collection.docs + collection.files
       end.to_a
     end


ashmaroli:
@DirtyF Since this optimization doesn't involve a critical path (used definitely by every site), Utterson may not report any difference in the build times..
Will you be able to post the "build time" for our `docs` on `master` and this branch, on macOS..?

The times on my **Windows** system are:

branch | build time
--- | ---
`master` | 32.225 s
`PR-7273` | 27.536 s

ashmaroli:
Banging my head on the wall here..
I simply do not understand why...
...why this would consistently show to be *slower than current `master`* :sweat:

ashmaroli:
Thank you for testing, @DirtyF :+1:

DirtyF:
@ashmaroli I took a median value from three runs, because it may vary from 1 or 2 seconds even on a 11s build. So it doesn't surprise me that it could vary from a few seconds on many websites, maybe our test suite don't use `site.documents`. 

P.S: If we have to do a lot of memoization, I wonder why we wouldn't consider a helper like https://github.com/matthewrudy/memoist

ashmaroli:
> I took a median value from three runs, because it may vary from 1 or 2 seconds even on a 11s build.

I understand that. Even the values I posted were a median of 3 runs..
But while I got a *gain of 3-4s* at my end, you got a *loss of 2-3s* at your end for the same docs site.. 
Now that's perplexing..
What's even more perplexing is the report from Utterson running sites that do not use `site.documents` shows a loss of 3-5s at average..

---

> I wonder why we wouldn't consider a helper like https://github.com/matthewrudy/memoist

I'm aware of that gem. But opted against it because using an instance_variable and a Hash directly gives us more fine-grained control as to what cache can be flushed and when they can be flushed.
Depending on a third-party gem for something so trivial can lead to unexpected bugs..

DirtyF:
@ashmaroli woops, sorry, bad copy-paste, it is the other way around for jekyll docs website, it showed an average ~2s gain. 😅

ashmaroli:
> woops, sorry, bad copy-paste

LOL! That explains it, then!! :laughing: 
Should've assumed it to be the case since you approved the PR shortly after.. :sweat_smile:

ashmaroli:
> This negatively impacts incremental regeneration when adding a new document, does it not?

@mattr- **In theory**, it should **not have** an impact on incremental regen because `Site#reset` is called, irrespective of `--incremental`..
The one ripple-effect this may have is with *generator*-plugins that *inject* a new object into a collection's `docs` or `files` and expect `Site#documents` to reflect that..

DirtyF:
@jekyllbot: merge +minor

PR #7276, opened 2018-09-26
Memoize Site#post_attr_hash

ashmaroli:
- This is a 🐛 bug fix.
- The test suite passes (run `script/cibuild` to verify this)

## Summary

Two common *callers*, `Site#tags` and `Site#categories`, are frequently called in Liquid Templates via the `SiteDrop` class

## Context

https://github.com/jekyll/jekyll/blob/44a8035d135b9f2a5b0102fcec84d3c9a87a79af/lib/jekyll/drops/site_drop.rb#L11
https://github.com/jekyll/jekyll/blob/44a8035d135b9f2a5b0102fcec84d3c9a87a79af/lib/jekyll/site.rb#L243-L249

### DIFF
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -92,6 +92,7 @@ module Jekyll
       self.pages = []
       self.static_files = []
       self.data = {}
+      @post_attr_hash = {}
       @site_data = nil
       @collections = nil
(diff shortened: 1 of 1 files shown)


DirtyF:
@jekyllbot: merge +minor

PR #7277, opened 2018-09-26
Check key in collections only if it isn't "posts"

ashmaroli:
- **This is a :roll_eyes: micro-optimization.**

## Summary

Check if keys of hash `site.collections` includes given drop-key only if the drop-key isn't `"posts"`


DirtyF:
@jekyllbot: merge +fix

Issue #7185, opened 2018-08-16
Add filter for generating URL paths which are relative to the current document


BigBlueHat:
Currently there is no way to generate a path (intended for an `href` or `src`) which is relative to the output documents location within the site tree.

Right now, there are two filters available:
 - `absolute_url` = `http://example.com/path/to/thing`
 - `relative_url` = `/path/to/thing`

What's missing is the ability to create a path which is relative to the current page (when generated) such as `to/thing` when referenced from a document in `/path/`. This would be especially useful when dealing with sites published to multiple locations where the use of `baseurl` would cause inconsistencies across domains (i.e. `example.com/projects/subsite/` vs. `example.github.io/subsite/`).

I'd like to propose the addition of a `relative_path` filter which could be used from pages and *templates* which would create such relative paths.

This name comes from terminology used in the URI spec [RFC 3986](https://tools.ietf.org/html/rfc3986#section-4.2)'s section on Relative Reference:
> A relative reference that does not begin with a slash character is termed a relative-path reference.

The current `relative_url` could be *optionally* aliased (and eventually removed?) as `absolute_path` to make the outputs clearer. Again, quoting from the URI spec:
> A relative reference that begins with a single slash character is termed an absolute-path reference.

The end result would be these three filters for creating URLs throughout templates and Liquid-enhanced content:
 - `absolute_url` = `http://example.com/path/to/thing`
 - `absolute_path` = `/path/to/thing` (currently `relative_url`)
 - `relative_path` = `to/thing` (when used from `/path/` for instance)

Thanks for considering this!
🎩 

---
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [x] This is a feature request.
- [x] I am on (or have tested on) ***Windows*** 10+

BigBlueHat:
Also, I'm aware this is a perennial request (re: #6360 and #6362), but as noted in both those threads there are many Jekyll deployment scenarios (such as GitHub Pages) where adding Plugins is not an option. Additionally, the expectation is that relative paths would be core to any content system dealing in tree structures, so having this "on hand" all the time (without any additional installation) would be invaluable (and appreciated).

DirtyF:
Rather than another filter, this could be a global (config) option that when set would rewrite all relative URLs (not the absolute ones), e.g. a relative link to `/about/` would become a  relative path to `../../about/`.

pathawks:
[...]

PR #7281, opened 2018-09-27
Add relative_path filter

switzersc:
- [x] I read the contributing document at https://jekyllrb.com/docs/contributing/

- [x] I've added tests (if it's a bug, feature or enhancement)
- [x] I've adjusted the documentation (if it's a feature or enhancement): Comments in the code, although I haven't submitted a PR to the docs repo.
- [x] The test suite passes (run `script/cibuild` to verify this)

## Summary

This PR adds the ability to create a path which is relative to the current page via a new `relative_path` filter.

## Context

This provides what I hope will be an uncontroversial solution to the much discussed problem of not being able to generate a path which is relative to the output document's location within the site tree. I have utilized the work already done in #6362 (which is locked) and updated it to utilize the term `relative_path` which is more inline with the URI spec [RFC 3986](https://tools.ietf.org/html/rfc3986#section-4.2)'s section on Relative Reference:

    A relative reference that does not begin with a slash character is termed a relative-path reference.

Related issues: 
#6360 
#7185

### DIFF
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -51,9 +51,10 @@ module Jekyll
       #
       # Returns the supplied path relativized to the current page.
       def relativize_url(input)
-        pageUrl = @context.registers[:page]["url"]
-        pageDir = Pathname(pageUrl).parent
-        Pathname(input).relative_path_from(pageDir).to_s
+        return if input.nil?
+        page_url = @context.registers[:page]["url"]
+        page_dir = Pathname(page_url).parent
+        Pathname(input).relative_path_from(page_dir).to_s
       end
 
       # Strips trailing `/index.html` from URLs to create pretty permalinks
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -48,8 +48,9 @@ module Jekyll
       # input - a path relative to the project root
       #
       # Returns the supplied path relativized to the current page.
-      def relativize_url(input)
+      def relative_path(input)
         return if input.nil?
+
         input = ensure_leading_slash(input)
         page_url = @context.registers[:page]["url"]
         page_dir = Pathname(page_url).parent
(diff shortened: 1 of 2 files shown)


ashmaroli:
@switzersc I couldn't find a test for a scenario where a page (or the entire site itself) could have a permalink set to generate "pretty" url.
For example, say there's a source file `./about.md` with `permalink: /about/` in its front matter. How would the **output file** of another page `./author/bio.md` (`_site/author/bio.html`) link to the output from the `./about.md` source file..?

switzersc:
@ashmaroli good point! I'll add a feature test.

dmalan:
Just wanted to chime in with a vote of support for this filter (and, in turn, the use cases so thoughtfully described by @aioobe in https://github.com/jekyll/jekyll/pull/6362). We've bumped up against the limitations of the existing `relative_url` in the context of reverse proxies whereby we have a site like `https://foo.com/bar/` proxying to `https://bar.com/` (as is possible with services like @netlify), the result of which is that the absolute paths generated by `relative_url` break the former. (And we'd like both URLs to work, particularly for testing.) For instance, if `https://foo.com/bar/index.html` proxies to `https://bar.com/index.html` and `index.html` has a tag like

```html
<script src="/js/scripts.js">
```

that path will incorrectly resolve to `https://foo.com/js/scripts.js` rather than to `https://foo.com/bar/js/scripts.js`. Particularly since GitHub Pages (on which use our same custom theme for other sites, thanks to @benbalter's [jekyll-remote-theme](https://github.com/benbalter/jekyll-remote-theme)) doesn't allow custom plugins, having `relative_path` would be a wonderful tool in Jekyll's core toolkit.

Thank you!

dmalan:
@DirtyF, just wanted to check in to see if @switzersc's PR resonates with the team? Thanks!

mattr-:
This PR needs a feature test (aka a cucumber test) and a docs update describing the filter before it's  ready to merge.

switzersc:
Hey y'all! Sorry I haven't wrapped this up yet with a test and docs. I'll aim to add those this week; in the meantime, help is welcome :)

PR #7282, opened 2018-09-28
Docs: Add title tag to item in RSS template

andy5995:

It adds a `<title>` tag to the `items` section of the rss example feed. The existing title tag is for the site title, but each item entry should also have a title.


DirtyF:
@jekyllbot: merge +docs

andy5995:
Thank you, @chrisfinazzo

PR #7291, opened 2018-10-02
Docs: Add link tag to item in RSS template

andy5995:
If this is missing, the feed validator at https://validator.w3.org/feed/
will output the message ```This feed does not validate.

    line 28, column 4: Missing channel element: link [help]

            </channel>```

/cc @jekyll/documentation


DirtyF:
@jekyllbot: merge +docs

PR #7299, opened 2018-10-04
Automatically load _config.toml

argv-minus-one:

## Summary

When looking for a `_config.*` file to load, also look for `_config.toml`.

## Description

Currently, Jekyll can load a TOML configuration file, but will only do so if explicitly told to with a command line option. With this PR, Jekyll will automatically look for a `_config.toml` file as well.

Default behavior remains the same: `_config.yml` and `_config.yaml` are tried first. If those don't exist, and `_config.toml` does exist, then that will be used instead.

### DIFF
--- a/docs/_docs/configuration.md
+++ b/docs/_docs/configuration.md
@@ -4,9 +4,9 @@ permalink: /docs/configuration/
 ---
 
 Jekyll gives you a lot of flexibility to customize how it builds your site. These
-options can either be specified in a `_config.yml` file placed in your site’s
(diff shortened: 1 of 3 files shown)


DirtyF:
❗️☕️ ➡️ The coffeescript failing test is not specific to this PR, it's already been happening on other PRs. Relaunching the test suite could make it pass, it looks random. 🤷‍♂️

DirtyF:
@jekyllbot: :shipit: +minor

Issue #7255, opened 2018-09-18
Validating the encoding of a post's content may cause 'undefined method `valid_encoding?' for nil:NilClass'


ironjan:
- [x] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
  - My version is 3.7.4 which is marked as latest on https://github.com/jekyll/jekyll/releases
  - https://jekyllrb.com/ shows 3.8.4 as current version
  - I checked the latest code and if my guess is correctly, the cause is still there
- [x] I am on (or have tested on) ***Other*** GNU/Linux
  - Manjaro. Arch based.
---
- [x] I had an error on GitHub Pages, and I have reproduced it locally.
- [x] I was trying to build.

## My Reproduction Steps

First of all, after some digging I found that the cause was human error: I created my posts as yaml files, not as markdown. This seems to have caused unintended behavior inside jekyll.

1. `git clone git@github.com:codeforpb/Kneipenplan.git`
2. `git checkout 53c417e8ae8138ad81a655c5d2ee07e38d46be5b`
3. `jekyll serve --verbose`

```
[...]
           Reading: food/_posts/1970-01-01-das_elfte_gebot.md
jekyll 3.7.4 | Error:  undefined method `valid_encoding?' for nil:NilClass
```

Please note that `food/_posts/1970-01-01-das_elfte_gebot.md` was innocent.

After renaming all yaml-files from `.yml` to `.md` the issue was fixed. 

## The Output I Wanted

Jekyll did not build the site and did not show any of the offending files. The change proposed in the linked PR does two things:

1. Adds a nil-check so that `jekyll serve` does not crash
2. Increases the log-level to warn so that offending files are listed

IssuehuntBot:
@issuehuntfest has funded $60.00 to this issue. [See it on IssueHunt](https://issuehunt.io/repos/65252/issues/7255)

ashmaroli:
Resolved via #7302

IssuehuntBot:
@issuehuntfest has cancelled funding for this issue.(Cancelled amount: $60.00) [See it on IssueHunt](https://issuehunt.io/repos/65252/issues/7255)

PR #7302, opened 2018-10-07
Skip processing posts that can not be read

lxxxvi:
Adresses #7255 

Issue #7255 suggests to skip processing posts that can not be read due to wrong (?) file extension and log/print the exception. 
The fix will also catch 

```
'undefined method `valid_encoding?' for nil:NilClass'
```

during startup of `jekyll serve`.

If anyone can help me pointing to the right place/approach to add a corresponding test, I'd love to do that.

### DIFF
--- a/lib/jekyll/readers/post_reader.rb
+++ b/lib/jekyll/readers/post_reader.rb
@@ -36,28 +36,7 @@ module Jekyll
     def read_publishable(dir, magic_dir, matcher)
       read_content(dir, magic_dir, matcher)
         .tap { |docs| docs.each(&:read) }
-        .select(&method(:readable?))
-        .select(&method(:publishable?))
-    end
-
-    def readable?(doc)
-      if doc.content.nil?
-        Jekyll.logger.debug "Skipping:", "Content in #{doc.relative_path} is nil"
-        false
-      elsif !doc.content.valid_encoding?
-        Jekyll.logger.debug "Skipping:", "#{doc.relative_path} is not valid UTF-8"
-        false
-      else
-        true
-      end
-    end
-
-    def publishable?(doc)
-      site.publisher.publish?(doc).tap do |will_publish|
-        if !will_publish && site.publisher.hidden_in_the_future?(doc)
-          Jekyll.logger.debug "Skipping:", "#{doc.relative_path} has a future date"
-        end
-      end
+        .select(&method(:processable?))
     end
 
     # Read all the content files from <source>/<dir>/magic_dir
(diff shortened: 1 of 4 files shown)


DirtyF:
Looks we don't have any test for `post_reader` yet, feel free to add some in https://github.com/jekyll/jekyll/tree/master/test

lib/jekyll/readers/post_reader.rb
```diff
@@ -34,19 +34,30 @@ def read_posts(dir)
...
-              end
-            end
-          else
-            Jekyll.logger.debug "Skipping:", "#{doc.relative_path} is not valid UTF-8"
-            false
-          end
+      read_content(dir, magic_dir, matcher)
+        .tap { |docs| docs.each(&:read) }
+        .select(&method(:readable?))
+        .select(&method(:publishable?))
```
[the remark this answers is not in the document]
lxxxvi: @ashmaroli 
Yes, thanks for pointing to that flaw.

I moved `:readable?` and `:publishable?` into a single new (private) method and named it `:processable?`.
Then rubocop complained, had to move `:publishable?` into a separate method.

lxxxvi:
@DirtyF 
Thanks for your help. I added `test_post_reader.rb`, I hope I've done it right.
Let me know if anything is not quite right or can be improved.

lib/jekyll/readers/post_reader.rb
```diff
@@ -34,19 +34,9 @@ def read_posts(dir)
...
-              end
-            end
-          else
-            Jekyll.logger.debug "Skipping:", "#{doc.relative_path} is not valid UTF-8"
-            false
-          end
-        end
+      read_content(dir, magic_dir, matcher)
+        .tap { |docs| docs.each(&:read) }
+        .select(&method(:processable?))
```
ashmaroli: @lxxxvi I know that the use of `&method` is a clever shortcut. But can we have the "conventional" block form instead..?
(unless you can produce some benchmarks that show using `&method` is a faster alternative..)
```ruby
      read_content(dir, magic_dir, matcher)
        .tap { |docs| docs.each(&:read) }
        .select { |doc| processable?(doc) }
```

test/source/_posts/2008-02-03-wrong-extension.yml
```diff
@@ -0,0 +1,7 @@
+---
+layout: default
+title: This file extension is skipped
+category: test_post_reader
+---
+
+Title says it all
```
ironjan: I would suggest to add some explanation here and [submitted a corresponding PR](https://github.com/lxxxvi/jekyll/pull/1) to you.

lxxxvi:
@ashmaroli 
Is there anything I need to do here? The "WIP" check says it's still pending...

ashmaroli:
@lxxxvi The WIP check seems to be facing a minor glitch.
Rest assured, this PR will be merged when one more maintainer approves the proposal.

oe:
@jekyllbot: merge +minor

PR #7306, opened 2018-10-09
Added missing semicolon

DreamingInBinary:
It wouldn't work for me without one, I am new to Sass so I could be missing something :)


DirtyF:
It is working without the final semicolon with Ruby Sass 3.6.0, the last one isn't mandatory in SCSS - like in CSS.
That does not mean this is a good practice, let's add the final semicolon.

@jekyllbot: merge +docs

PR #7310, opened 2018-10-10
feat: enhance --blank scaffolding

### LINKED ISSUE #5260: New option: create a blank site, without a template by default

## Current behaviour

1. Create new site with `jekyll new .`
2. Preview site with `jekyll serve`
3. Result: A site with a custom template, need to remove a lot of SASS and html files to start from zero

## Expected behaviour

1. Create new site with `jekyll new . --blank`
2. Preview site with `jekyll serve`
3. Result: Get a site without a template. Start developing.

It would be nice to have an option in `jekyll new`  to create a site from scratch. At least without a template. Everytime I'm starting a new website I have to delete a lot of the SASS and HTML that comes by default.

It can be an option, like `jekyll new . --blank` or something similar.

I searched for this feature on the docs and the issues and I wasn't able to find anything similar. Maybe it's already supported!

jekyllbot:
Fix #5260 

## Context 

Current blank template could be enhanced so that users can start developing right away.

```
As a developer
I want to be able to start developing a Jekyll site from scratch with default scaffolding
```

```gherkin
Given I start a new Jekyll project
When I run `bundle exec jekyll new SITENAME --blank`
Then I can start developing from a complete default scaffolding:
```

```
├── _config.yml
├── _data
├── _drafts
├── _includes
├── _layouts
│   └── default.html
├── _posts
├── _sass
│   └── main.scss
├── assets
│   └── css
│       └── main.scss
└── index.html
```

### Demo

[asciicast](https://asciinema.org/a/205751)


DirtyF:
We could go a bit further and create at least `_layouts/default.html`

ashmaroli:
> We could go a bit further and create at least `_layouts/default.html`

Yeah that'd be nice since all the themes at GHP has just a `default` layout..

lib/jekyll/commands/new.rb
```diff
@@ -43,7 +43,10 @@ def process(args, options = {})
 
         def create_blank_site(path)
           Dir.chdir(path) do
-            FileUtils.mkdir(%w(_layouts _posts _drafts))
+            FileUtils.mkdir(%w(_data _drafts _includes _layouts _posts _sass assets assets/css))
+            FileUtils.touch("_config.yml")
+            FileUtils.touch("_sass/main.scss")
+            FileUtils.touch("assets/css/main.scss")
```
ashmaroli: You shouldn't just `touch` these files.. They're useless that way..
* `_config.yml` should never be empty. It will not validate otherwise.
* `assets/css/main.scss` should `import "main"` otherwise there's no point having a `_sass/main.scss`.
* Might as well *add something* to `_sass/main.scss` since its just sitting there..

DirtyF: empty `config.yml` is fine, I tested the generated *blank* scaffold, and no error on `jekyll build`.

ashmaroli: Strange..
https://github.com/jekyll/jekyll/blob/153f24c4e4e00574fce074d8cc261a5f754a0e58/lib/jekyll/configuration.rb#L304-L309
https://github.com/jekyll/jekyll/blob/153f24c4e4e00574fce074d8cc261a5f754a0e58/lib/jekyll/configuration.rb#L171-L173

oe: I believe YAML works in mysterious ways here where an empty file counts as a `Hash`.

ashmaroli: okay...
So empty config files default to an empty Hash..
https://github.com/jekyll/jekyll/blob/153f24c4e4e00574fce074d8cc261a5f754a0e58/lib/jekyll/configuration.rb#L134-L135

mattr-:
Some context I'm missing from this PR:

We already had `--blank` and it did something. What did it do before and how does this change it?

mattr-:
oh, never mind. it's in the diff. 🤦‍♂️

lib/jekyll/commands/new.rb
DirtyF: @ashmaroli addressed your concerns, `--blank` now provides a default layout linking to a generated CSS, and  *blank* config file. I don't think we should go further.

lukepighetti:
Glad to see this PR. Thanks for the hard work!

ashmaroli:
But you should fix the failing test before this gets merged..

oe:
@jekyllbot: merge +minor

PR #7312, opened 2018-10-10
CI(Appveyor): shallow clone with 5 last commits

XhmikosR:


parkr:
@XhmikosR Can you explain what you're doing here? The only substantive change I see is `clone` moved from `10` to `5`.

XhmikosR:
Yup, nothing else. The rest is a cosmetic change, since environments makes more sense to be first to me :)

parkr:
Ok, works for me!

DirtyF:
@jekyllbot: merge +dev

tedmelvin25:
Hi

PR #7315, opened 2018-10-12
Handle files with trailing dots in their basename

### LINKED ISSUE #7246: Double dots in filenames

## My Reproduction Steps

This is a corner case first signaled by @AxelTerizaki for a plugin I wrote. However, it does not depend on the plugin and might be an unintended behavior of Jekyll.

1. create a new Jekyll site (`jekyll new test`)
2. create a source file with a final dot before the extension, e.g., `my-file..md`
2. build

## The Output I Wanted

`_site/my-file..html`  or `_site/my-file./index.html`

## The Output I get

`_site/my-file/html.html`

## Additional Notes

Similarly:  `test..test..test.md` gets mapped to `_site/test/test/test.html` (which might promote it to a feature, rather than a bug)

ashmaroli:
- This is a 🐛 bug fix.
- I've added tests.

## Summary

Render files with trailing dots in their "basename" (filename w/o extension) into correct url instead of generating incorrect URLs or aborting the build entirely

## Context

Resolves #7246

### DIFF
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -119,8 +119,10 @@ module Jekyll
     #   and with the collection's directory removed as well.
     # This method is useful when building the URL of the document.
     #
+    # NOTE: `String#gsub` removes all trailing periods (in comparison to `String#chomp`)
+    #
     # Examples:
-    #   When relative_path is "_methods/site/generate.md":
+    #   When relative_path is "_methods/site/generate...md":
     #     cleaned_relative_path
     #     # => "/site/generate"
     #
@@ -486,6 +488,7 @@ module Jekyll
       end
 
       # slugs shouldn't end with a period
+      # `String#gsub!` removes all trailing periods (in comparison to `String#chomp!`)
       slug.gsub!(%r!\.*\z!, "")
 
       # Try to ensure the user gets a title.
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -127,7 +127,9 @@ module Jekyll
     # Returns the cleaned relative path of the document.
     def cleaned_relative_path
       @cleaned_relative_path ||=
-        relative_path[0..-extname.length - 1].sub(collection.relative_directory, "")
+        relative_path[0..-extname.length - 1]
+          .sub(collection.relative_directory, "")
+          .gsub(%r!\.*\z!, "")
     end
 
     # Determine whether the document is a YAML file.
@@ -482,6 +484,9 @@ module Jekyll
       elsif relative_path =~ DATELESS_FILENAME_MATCHER
         slug, ext = Regexp.last_match.captures
       end
+
+      # slugs shouldn't end with a period
+      slug.gsub!(%r!\.*\z!, "")
 
       # Try to ensure the user gets a title.
       data["title"] ||= Utils.titleize_slug(slug)
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -119,7 +119,7 @@ module Jekyll
     # Returns nothing.
     def process(name)
       self.ext = File.extname(name)
-      self.basename = name[0..-ext.length - 1]
+      self.basename = name[0..-ext.length - 1].gsub(%r!\.+\z!, "")
     end
 
     # Add any necessary layouts to this post
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -116,10 +116,11 @@ module Jekyll
     #
     # name - The String filename of the page file.
     #
+    # NOTE: `String#gsub` removes all trailing periods (in comparison to `String#chomp`)
     # Returns nothing.
     def process(name)
       self.ext = File.extname(name)
-      self.basename = name[0..-ext.length - 1].gsub(%r!\.+\z!, "")
+      self.basename = name[0..-ext.length - 1].gsub(%r!\.*\z!, "")
     end
 
     # Add any necessary layouts to this post
--- a/lib/jekyll/static_file.rb
+++ b/lib/jekyll/static_file.rb
@@ -95,7 +95,7 @@ module Jekyll
     #
     # Returns false if the file was not modified since last time (no-op).
     def write(dest)
-      dest_path = destination(dest)
+      dest_path = destination(dest).squeeze(".")
 
       return false if File.exist?(dest_path) && !modified?
 
@@ -120,7 +120,7 @@ module Jekyll
       {
         :collection => @collection.label,
         :path       => relative_path[
-          @collection.relative_directory.size..relative_path.size],
+          @collection.relative_directory.size..relative_path.size].squeeze("."),
         :output_ext => "",
         :name       => "",
         :title      => "",
@@ -134,7 +134,7 @@ module Jekyll
       @url ||= if @collection.nil?
                  relative_path
                else
-                 ::Jekyll::URL.new(
+                 URL.new(
                    :template     => @collection.url_template,
                    :placeholders => placeholders
                  )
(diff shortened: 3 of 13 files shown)


ashmaroli:
> Why don't we warn the user that he should rename files with consecutive dots correctly?

We could warn the user as well.
Especially since the destination url is being changed as a direct result of this proposal. i.e. `./foobar...md` and `./foobar.md` both will render into the same "destination url"
That said, this was already a possible outcome if the user had a `./foobar.md` and `./foobar.markdown`..


> What about `.file.html`, `fi.le.html`, `file.ht.ml`?

`.file.html` will be seen as a dotfile and hence ignored by default.
`file.ht.ml` will have `.ml` as an extension for which we do not have a `converter` by default. So that'll simply get rendered into `_site/file.ht.ml`
`fi.le.html` will be converted into `_site/fi.le.html` as well

`fi.le.md` will be converted into `_site/fi.le.html`


I chose to fix this corner-case rather than issue a warning because I felt Jekyll should be smart about such things rather than issue a warning and go about its business..

ashmaroli:
(I amended the commit to fix cucumber tests as well..)

lib/jekyll/document.rb
```diff
@@ -127,7 +127,9 @@ def basename
     # Returns the cleaned relative path of the document.
     def cleaned_relative_path
       @cleaned_relative_path ||=
-        relative_path[0..-extname.length - 1].sub(collection.relative_directory, "")
+        relative_path[0..-extname.length - 1]
+          .sub(collection.relative_directory, "")
+          .gsub(%r!\.*\z!, "")
```
parkr: You might be able to use `.chomp(".")` here instead. [String#chomp](https://ruby-doc.org/core-2.2.0/String.html#method-i-chomp)

lib/jekyll/document.rb
```diff
@@ -483,6 +485,9 @@ def populate_title
         slug, ext = Regexp.last_match.captures
       end
 
+      # slugs shouldn't end with a period
+      slug.gsub!(%r!\.*\z!, "")
```
parkr: You might be able to use `.chomp(".")` here instead. [String#chomp](https://ruby-doc.org/core-2.2.0/String.html#method-i-chomp)

lib/jekyll/page.rb
```diff
@@ -119,7 +119,7 @@ def url_placeholders
     # Returns nothing.
     def process(name)
       self.ext = File.extname(name)
-      self.basename = name[0..-ext.length - 1]
+      self.basename = name[0..-ext.length - 1].gsub(%r!\.+\z!, "")
```
parkr: You might be able to use `.chomp(".")` here instead. [String#chomp](https://ruby-doc.org/core-2.2.0/String.html#method-i-chomp)

lib/jekyll/static_file.rb
```diff
@@ -95,7 +95,7 @@ def write?
     #
     # Returns false if the file was not modified since last time (no-op).
     def write(dest)
-      dest_path = destination(dest)
+      dest_path = destination(dest).squeeze(".")
```
parkr: Can you explain what this is doing and why it's needed in addition to the above?

lib/jekyll/static_file.rb
```diff
@@ -120,7 +120,7 @@ def placeholders
       {
         :collection => @collection.label,
         :path       => relative_path[
-          @collection.relative_directory.size..relative_path.size],
+          @collection.relative_directory.size..relative_path.size].squeeze("."),
```
parkr: Can you explain what this is doing and why it's needed beyond the modification we're doing above?

lib/jekyll/static_file.rb
```diff
@@ -134,7 +134,7 @@ def url
       @url ||= if @collection.nil?
                  relative_path
                else
-                 ::Jekyll::URL.new(
+                 URL.new(
```
parkr: This was just to make sure we get the Jekyll::URL. Generally speaking I like to use the fully-qualified class name to reduce ambiguity.

lib/jekyll/document.rb
ashmaroli: I intentionally did not use `String#chomp` here because `#chomp` removes only one trailing character.
```ruby
"hello-world...".chomp(".")           # => "hello-world.."
"hello-world...".gsub(%r!\.*\z!, "")  # => "hello-world"
```

lib/jekyll/document.rb
ashmaroli: I intentionally did not use `String#chomp` here because `#chomp` removes only one trailing character.
```ruby
"hello-world...".chomp(".")           # => "hello-world.."
"hello-world...".gsub(%r!\.*\z!, "")  # => "hello-world"
```

lib/jekyll/page.rb
ashmaroli: I intentionally did not use `String#chomp` here because `#chomp` removes only one trailing character.
```ruby
"hello-world...".chomp(".")           # => "hello-world.."
"hello-world...".gsub(%r!\.*\z!, "")  # => "hello-world"
```

lib/jekyll/static_file.rb
ashmaroli: `String#squeeze` reduces multiple occurrences of a substring into a single occurrence.
```ruby
"my-cool-avatar...png".squeeze(".")  # => "my-cool-avatar.png"
```

> why it's needed in addition to the above?

The "above" sanitization is in a different class of code..
The "above" regex sanitization was in `Jekyll::Page` and `Jekyll::Document` where `basename` would determine the generated `url`

"This" is in `Jekyll::StaticFile`, where the `destination url` is *generally only* determined by the static-file's `path` and `relative_path`..
https://github.com/jekyll/jekyll/blob/fd6e01fd6a33365deaf19832f25f1b77e1d93697/lib/jekyll/static_file.rb#L133-L142

https://github.com/jekyll/jekyll/blob/fd6e01fd6a33365deaf19832f25f1b77e1d93697/lib/jekyll/static_file.rb#L119-L128

lib/jekyll/static_file.rb
ashmaroli: `String#squeeze` reduces multiple occurrences of a substring into a single occurrence.
```ruby
"my-cool-avatar...png".squeeze(".")  # => "my-cool-avatar.png"
```

> why it's needed in addition to the above?

The "above" sanitization is in a different class of code..
The "above" regex sanitization was in `Jekyll::Page` and `Jekyll::Document` where `basename` would determine the generated `url`

"This" is in `Jekyll::StaticFile`, where the `destination url` is *generally only* determined by the static-file's `path` and `relative_path`..
https://github.com/jekyll/jekyll/blob/fd6e01fd6a33365deaf19832f25f1b77e1d93697/lib/jekyll/static_file.rb#L133-L142

https://github.com/jekyll/jekyll/blob/fd6e01fd6a33365deaf19832f25f1b77e1d93697/lib/jekyll/static_file.rb#L119-L128

ashmaroli:
@parkr I've added some comments and a new method to handle sanitization: `StaticFile#cleaned_relative_path` along the same lines of `Document#cleaned_relative_path`

Please re-review when possible.
Thanks.

mattr-:
@jekyllbot: merge +bug

fauno:
sorry to return to this but i just got hit by it.  i'm migrating a customer's site to jekyll and they have lots of files that should maintain their previous url. i found a few files with double dots before the extension were "smartly" corrected to a single dot, breaking symlinks i made to tidy things a bit. so i was trying to understand the rationale behind this, why would you correct static file names anyway?

ashmaroli:
> why would you correct static file names anyway?

That was for consistency. But why people would have their file's basename with trailing dots is beyond me..

Nevertheless, I'm not sure if we should revert the sanitization of *static file* basenames in a future Jekyll release..
/cc @jekyll/build

PR #7316, opened 2018-10-14
Measure the no. of times a template gets rendered

ashmaroli:
- This is a 🐛 bug fix.

## Summary

As a direct consequence of #7136, the `count` field in the `--profile` output table returns `1` for includes rendered for each "page" / "document", which doesn't provide the user with valuable information.

This PR restores the "profile data"


DirtyF:
@jekyllbot: merge +fix

PR #7319, opened 2018-10-17
Add Tidelift to site sponsors

oe:
This PR adds Tidelift, who are "sponsoring" us (e.g. giving us a share of their income from their subscription model). It should be said that ideally, we would give them a larger spot, potentially with a bit of text, since the more people visit their site and subscribe to them, the more money we make. I've held off on doing that in this PR since I'm not sure how the rest of the team feels about it.

We could also add a reference in our README file, or do a shoutout on Twitter. Thoughts?


oe:
@DirtyF Good idea, this is what it looks like now:

PR #7335, opened 2018-10-24
Document Jekyll Filters with YAML data

ashmaroli:
- This is a 🔦 documentation change.

## Summary
This change is mainly to leverage Jekyll's Core functionality and avoid having to write and maintain HTML in the docs-site's source-files.
By having a Liquid Template take over, contributors need not worry about the markup &mdash; simply provide the necessary data.

PR #7336, opened 2018-10-24
docs: Page variables as YAML data

DirtyF:
Move page API to YAML files. (following #7335)

What if Jekyll API was available as JSON output, could IDE plugins use it to implement autocompletion for instance?

### DIFF
--- a/docs/_data/jekyll_variables.yml
+++ b/docs/_data/jekyll_variables.yml
@@ -1,27 +1,24 @@
 global:
   - name: site
     description: >-
-      Site wide information + configuration settings from
-      <code>_config.yml</code>. See below for details.
+      Site wide information + configuration settings from <code>_config.yml</code>.
+      See below for details.
   - name: page
     description: >-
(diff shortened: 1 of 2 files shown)


ashmaroli:
The following seems like repeated code to me:
```
<div class="mobile-side-scroller">
<table>
  <thead>
    <tr>
      <th>Variable</th>
      <th>Description</th>
    </tr>
  </thead>
  <tbody>
  {% for var in site.data.jekyll_variables.global %}
    <tr>
      <td><p><code>{{ var.name }}</code></p></td>
      <td><p>{{- var.description -}}</p></td>
    </tr>
  {% endfor %}
  </tbody>
</table>
</div>
```
@DirtyF What do you think of the idea of abstracting it to an "include"..?

chrisfinazzo:
Yes, please DRY where you can.

DirtyF:
@ashmaroli will try that in another PR.

PR #7342, opened 2018-10-27
Remove redundant instruction comment

aidan-fitz:
This is a change to an :octocat: issue template.

## Summary

- [x] Removed redundant instruction comment. Template originally said "What should we do instead?" twice.
- [x] Added exclamation marks to top and bottom comments for extra "friendliness".

## Context

N/A

## Semver Changes

None, because this doesn't affect the codebase.


DirtyF:
@jekyllbot: merge +docs

PR #7345, opened 2018-10-31
Cache globbed paths in front matter defaults

ashmaroli:
- This is an :zap: optimization fix.

## Summary

`Dir.glob` is a known to be a slow function (especially on Windows).
Caching the result of this function helps to avoid performance issues from globbing the same pattern
multiple times.


ashmaroli:
(Utterson test cases do not use *globbing patterns* in front matter defaults..)

DirtyF:
Here's some quick benchmark with a single glob pattern:

```
defaults:
  - scope:
      path: "assets/**/*.{js,css,png,svg,ico,ttf,woff,woff2}"
    values:
      sitemap: false
```

| without glob (master) | glob (master) | glob (#7345) |
|-----------------------|--------------|---------------|
| ~8.5s                  | ~17s        | ~13.5s           |

DirtyF:
@jekyllbot: merge +minor

PR #7351, opened 2018-11-03
Addresses bundle not found.

EarthSchlange:
## Summary

Adds code so the ```bundle: command not found``` error that results from not having the gems location define on the server does not happen and allows gems to be installed when running a ```$ git push origin```.

## Type of Change


DirtyF:
@jekyllbot: merge +docs

PR #7357, opened 2018-11-06
Log a warning when the slug is empty

mokhan:
This is a 🐛 bug fix.

## Summary

When slugifying an emoji character an empty string is returned unless you use a slug mode of `:pretty`.

```irb
irb(main):002:0> x = "💎"
=> "💎"
irb(main):003:0> Jekyll::Utils.slugify(x)
=> ""
irb(main):004:0> Jekyll::Utils.slugify(x, mode: nil)
=> ""
irb(main):005:0> Jekyll::Utils.slugify(x, mode: :pretty)
=> "💎"
```

The proposed change in this pull request will raise an error instead of returning an empty string. 

I have marked this PR as WIP as I am not sure if this change warrants a documentation update or not. Please advise.

## Context

This PR is based on a proposal described in https://github.com/jekyll/jekyll-archives/pull/129. 

## Semver Changes

I recommend a patch version increment. This `shouldn't'` change any existing behaviour except in cases where slugs are getting returned as a blank string. In which case I think raising an error so that the issue can be identified is best.

### DIFF
--- a/lib/jekyll/utils.rb
+++ b/lib/jekyll/utils.rb
@@ -222,6 +222,8 @@ module Jekyll
       slug.gsub!(%r!^\-|\-$!i, "")
 
       slug.downcase! unless cased
+      raise Errors::InvalidPermalinkError if slug.empty?
+
       slug
     end
 
--- a/lib/jekyll/utils.rb
+++ b/lib/jekyll/utils.rb
@@ -222,7 +222,7 @@ module Jekyll
       slug.gsub!(%r!^\-|\-$!i, "")
 
       slug.downcase! unless cased
-      Jekyll.logger.warn "Warning:", "Empty `slug` generated for '[#{string}]'." if slug.empty?
+      Jekyll.logger.warn("Warning:", "Empty `slug` generated for '#{string}'.") if slug.empty?
       slug
     end
 
--- a/lib/jekyll/utils.rb
+++ b/lib/jekyll/utils.rb
@@ -222,8 +222,7 @@ module Jekyll
       slug.gsub!(%r!^\-|\-$!i, "")
 
       slug.downcase! unless cased
-      raise Errors::InvalidPermalinkError if slug.empty?
-
+      Jekyll.logger.warn "Warning:", "Empty `slug` generated for '[#{string}]'." if slug.empty?
       slug
     end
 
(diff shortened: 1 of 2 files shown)


ashmaroli:
The proposal looks good..
But thinking about this as a whole, I find it odd that the build would:
  - `raise` for no fault of the user.. All they did was probably have a `{{ page.category | slugify }}`..
  - `raise` without any message to the user..

mokhan:
Instead of raising an error, do you think it would be a better user experience to record an error in the log? This would allow the user to see that there is a potential issue that they may want to look at but it wouldn't interrupt the entire build process.

ashmaroli:
> it would be a better user experience to record an error in the log?

**IMHO**, we could simply log a warning.
```
   Warning: Empty `slug` generated for '[input]'.
```

ping: @jekyll/build for their collective opinion

mokhan:
The value of this change in the current state is low, but the risk is high. Logging a warning makes sense to me.

I will wait for the collective wisdom before making the next move.

mattr-:
I'm in favor of logging an error and not throwing an exception. I think it makes sense for users to see that a post doesn't have a permalink. I don't think the build should be aborted.

lib/jekyll/utils.rb
```diff
@@ -222,6 +222,7 @@ def slugify(string, mode: nil, cased: false)
       slug.gsub!(%r!^\-|\-$!i, "")
 
       slug.downcase! unless cased
+      Jekyll.logger.warn "Warning:", "Empty `slug` generated for '[#{string}]'." if slug.empty?
```
ashmaroli:
```suggestion
      Jekyll.logger.warn("Warning:", "Empty `slug` generated for '#{string}'.") if slug.empty?
```

mokhan:
The build failure is:

```bash
Failure:
TestRegenerator#test_: The site metadata should read legacy YAML metadata.  [/home/travis/build/jekyll/jekyll/test/test_regenerator.rb:182]
Minitest::Assertion: Expected: 2018-11-08 22:27:33 UTC
  Actual: 2018-11-08 22:27:33 +0000
```

I'm not quite sure how I triggered this failure. locale?

DirtyF:
@jekyllbot: merge +fix

PR #7361, opened 2018-11-08
Use keyword arguments instead of a Hash argument

ashmaroli:
- This is an :zap: optimization change

## Summary

Testing to see if this has a visible gain in performance..
The included benchmark script shows that using keyword args are about ~2.5x faster than a Hash argument

### DIFF
--- a/benchmark/hash-vs-keywords.rb
+++ b/benchmark/hash-vs-keywords.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+# ---------------------------------------------------------
+# Adapted from: https://gist.github.com/palexander/8592640
+# ---------------------------------------------------------
+
+require "benchmark/ips"
+
+NAME  = "Test Name"
+EMAIL = "test@example.org"
+
+class Person
+  attr_accessor :name, :email
+
+  def set_with_hash(options = {})
+    @name  = options[:name]
+    @email = options[:email]
+  end
+
+  def set_with_keywords(name:, email:)
+    @name  = name
+    @email = email
+  end
+end
+
+Benchmark.ips do |x|
+  x.report("hash arg") do
+    p = Person.new
+    p.set_with_hash(name: NAME, email: EMAIL)
+  end
+
+  x.report("keyword args") do
+    p = Person.new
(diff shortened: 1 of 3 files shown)


ashmaroli:
Utterson reports a ~3s gain:

ref | build time in seconds
-- | --
#7361 | **232.63**
master | 235.23

mattr-:
Is it possible to estimate the impact this will have on plugins, sites, etc? Switching to keyword arguments is a breaking change. I don't have any problems with that, but it would be good to know and  document the cases where we expect breakage as part of an upgrade guide for 4.0

ashmaroli:
@mattr- To me, this change does not classify as "breaking" since the constructor signature has not changed. It just took a more explicit form.
Moreover, our test-suite that mocks new instances (esp. of `Document` and `URL`) pass without any modification. So I'm pretty confident that this would have zero impact on plugins.

mattr-:
@ashmaroli The constructor signature has changed though. Options that could be present in the hash, should this PR gets merged, are no longer permitted. In general, a move like this from less explicit to more explicit is a breaking change.

Consider the following method:
```ruby
def foo(test:, test2:)
  puts "#{test} #{test2}"
end
```

If you call it with an extra parameter, you'll get an argument error:
```ruby
irb(main):009:0> foo(test: "foo", test2: "bar", test3: "baz")
ArgumentError: unknown keyword: test3
        from (irb):5:in `foo'
        from (irb):9
        from /usr/bin/irb:11:in `<main>'
```

The hash variant does not raise an error, and if the method never uses the extra pairs in the hash, there's not a problem:
```ruby
def foo(hash)
  puts "#{hash}"
end
```
```ruby
irb(main):013:0> foo(test: "foo", test2: "bar", test3: "baz")
{:test=>"foo", :test2=>"bar", :test3=>"baz"}
```


We need to document that we've changed the signatures of `Jekyll::Document` and `Jekyll:Url` so that users can make adjustments since both of these classes are public API.

ashmaroli:
@mattr- I proposed this change under the impression that it wouldn't be a breaking change.
But in the light of your explanation, I no longer wish to proceed with this and add complexity for minor perf gains.

Thank you very much for the detailed explanation.

PR #7365, opened 2018-11-09
Update excludes for CodeClimate Analyses

ashmaroli:
Exclude superfluous files that do not affect our CodeClimate Score
(Any file that isn't `lib/**/*.rb`)


DirtyF:
@jekyllbot: merge +dev

PR #7368, opened 2018-11-11
Update 07-assets.md

henryhu712:
## Summary

Use absolute path to make it clearer.

### DIFF
--- a/docs/_docs/step-by-step/07-assets.md
+++ b/docs/_docs/step-by-step/07-assets.md
@@ -49,12 +49,12 @@ content:
 
 The empty front matter at the top tells Jekyll it needs to process the file. The
 `@import "main"` tells Sass to look for a file called `main.scss` in the sass
-directory (`_sass/` by default).
+directory (`/_sass/` by default).
 
 At this stage you'll just have a main css file. For larger projects, this is a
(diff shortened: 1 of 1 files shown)


ashmaroli:
@henryhu712 There is no need for the leading slash in either of the places you've changed.
The `_sass` directory is at the root of a Jekyll source directory, not at the root of the system directory..

henryhu712:
@ashmaroli I agree with you. I promote the pull request because there is an expression of /assets/css/style.sass just previous it. The absolute and relative paths confused me. It needs to change another.

ashmaroli:
> there is an expression of /assets/css/style.sass just previous it. The absolute and relative paths confused me

Let me make it clearer for you.
The two expressions are of different types. `/assets/css/style.sass` is an **URL** relative to the webserver root. `_sass/main.scss` is a **file path** relative to the Jekyll source directory.

The URL is usually "generated" points to the "generated file". The file path points to the "source file".

henryhu712:
Thank you. How can I know they are url path or file path without your explanation?

DirtyF:
All folders beginning with an underscore will not be copied in the destination folder, and thus will not become an URL.

PR #7372, opened 2018-11-12
Switch to `:install_if` for wdm gem

XhmikosR:
The reason for this is that the Gemfile.lock is consistent now between all platforms, thus `bundle install --deployment` does not break anymore.

Before: https://travis-ci.org/twbs/bootstrap/builds/441675281
After: https://travis-ci.org/twbs/bootstrap/builds/454076791

I'm no Bundler/Ruby guru for sure, so it'd be nice if we could verify the minimum bundler version this needs, and if there's any downside. So far I haven't seen any issues, and I will land this in Bootstrap soon.

### DIFF
--- a/lib/jekyll/commands/new.rb
+++ b/lib/jekyll/commands/new.rb
@@ -91,7 +91,7 @@ module Jekyll
             # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
             gem "tzinfo-data", platforms: [:mingw, :mswin, :x64_mingw, :jruby]
             # Performance-booster for watching directories on Windows
-            gem "wdm", "~> 0.1.0" if Gem.win_platform?
+            gem "wdm", "~> 0.1.1", :install_if => Gem.win_platform?
 
           RUBY
         end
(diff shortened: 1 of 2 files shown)


lib/jekyll/commands/new.rb
```diff
@@ -91,7 +91,7 @@ def gemfile_contents
             # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
             gem "tzinfo-data", platforms: [:mingw, :mswin, :x64_mingw, :jruby]
             # Performance-booster for watching directories on Windows
-            gem "wdm", "~> 0.1.0" if Gem.win_platform?
+            gem "wdm", "~> 0.1.1", :install_if => Gem.win_platform?
```
ashmaroli:
```suggestion
            gem "wdm", "~> 0.1.0", :install_if => Gem.win_platform?
```

ashmaroli:
> it'd be nice if we could verify the minimum bundler version this needs, and if there's any downside.

Based on [this commit](https://github.com/bundler/bundler/commit/6342a6bf2bb4a51902fdea14ef6d80c1fb6847e0), the minimum Bundler version required is **`v1.10`**..
Plus, I've tested this locally on my Windows system and **there're no issues** with building on Windows.

ashmaroli:
Note: Technically, this is **not a Windows issue**.
But it has been labelled `windows` because the change is *closely related* to builds on Windows platform

PR #7375, opened 2018-11-12
Upgrade liquid-c to v4.0

### LINKED ISSUE #7374: Update dependency constraint to allow for liquid-c v4.0.0
Hey there! :wave:

I noticed that the constraint you have for liquid-c doesn't allow for the latest version to be used.

The constraint I found was `~> 3.0`, and the latest version available is `4.0.0`.

DirtyF:
Fix #7374 

🆙 💎


DirtyF:
@jekyllbot: :shipit: +minor

According to basic benchmarking, that could boost Liquid rendering up to 10%

PR #7381, opened 2018-11-15
Remove installation instructions with Homebrew

leafac:
The `ruby` formula was changed to [keg-only], which means the `ruby(1)` executable is no longer linked and the installation instructions no longer work—the system Ruby continues to be used.

[keg-only]: https://github.com/Homebrew/homebrew-core/pull/33503


DirtyF:
@jekyllbot: merge +docs

PR #7382, opened 2018-11-16
Excerpt handling of custom and intermediate tags

### LINKED ISSUE #7380: Excerpt handling in 3.8.5 generates Liquid errors despite valid Liquid syntax
## My Environment

| Operating System | Ubuntu 18.04.1 LTS |
| `jekyll`         | 3.8.5     |
| `github-pages`   | no     |

---

## Expected Behaviour

Valid Liquid syntax should not throw errors when excerpted; excerpts should gracefully handle Liquid syntax that is correct in the full page (or at least explain what the issue is). This seems related to the [changes in excerpt handling made in 3.8.5](7250) as I do not get errors using jekyll 3.8.4.

## Current Behavior

In some cases, valid Liquid syntax results in syntax errors related to the excerpt

    Error: A Liquid tag in the excerpt of _test/test.xml couldn't be parsed.
    Error: could not read file /tmp/test/_test/test.xml: undefined method `ancestors' for nil:NilClass

Calling `jekyll _3.8.4_ build` generates no errors. Putting `excerpt: none` in the front matter also eliminates the error.

## Code Sample

This isn't my use case (my use case is generating an RSS file), but here's a minimal working example:

Directory tree:

```
.
├── _config.yml
└── _test
    └── test.xml
```

_config.yml:

```
collections:
    test:
        output:     true
```

_test/test.xml:

```
---
---

{% if true %}
xyzzy
{% else %}
plugh
{% endif %}
```

Output:

```
$ jekyll build
Configuration file: /tmp/test/_config.yml
            Source: /tmp/test
       Destination: /tmp/test/_site
      Generating... 
             Error: A Liquid tag in the excerpt of _test/test.xml couldn't be parsed.
             Error: could not read file /tmp/test/_test/test.xml: undefined method `ancestors' for nil:NilClass
                    done in 0.012 seconds.
```

kylebarbour:
This is a 🐛 bug fix.

## Summary

Resolves #7380. This updates `lib/jekyll/excerpts.rb` to improve handling of Liquid blocks which are not properly closed in the excerpt. User-defined custom blocks can now be auto-closed and intermediate logic in blocks (e.g. `else`, `elsif`, `break`, and `continue`) no longer generate issues.

### DIFF
--- a/test/source/_plugins/custom_block.rb
+++ b/test/source/_plugins/custom_block.rb
@@ -10,5 +10,5 @@ module Jekyll
   end
 end
 
-Liquid::Template.register_tag("do_nothing", DoNothingBlock)
-Liquid::Template.register_tag("do_nothing_other", DoNothingOther)
+Liquid::Template.register_tag("do_nothing", Jekyll::DoNothingBlock)
+Liquid::Template.register_tag("do_nothing_other", Jekyll::DoNothingOther)
--- a/test/source/_plugins/custom_block.rb
+++ b/test/source/_plugins/custom_block.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+# For testing excerpt handling of custom tags
+
+module Jekyll
+  class DoNothingBlock < Liquid::Block
+  end
+
+  class DoNothingOther < Liquid::Tag
+  end
+end
+
+Liquid::Template.register_tag("do_nothing", DoNothingBlock)
+Liquid::Template.register_tag("do_nothing_other", DoNothingOther)
--- a/test/test_excerpt.rb
+++ b/test/test_excerpt.rb
@@ -283,10 +283,10 @@ class TestExcerpt < JekyllUnitTest
       @excerpt = @post.data["excerpt"]
 
       assert_includes @post.content.split("\n\n")[0].strip, "{% continue %}"
-      assert_equal true, DoNothingBlock.ancestors.include?(Liquid::Block)
-      assert_equal false, DoNothingOther.ancestors.include?(Liquid::Block)
-      assert_match "DoNothingBlock", Liquid::Template.tags["do_nothing"].name
-      assert_match "DoNothingOther", Liquid::Template.tags["do_nothing_other"].name
+      assert_equal true, Jekyll::DoNothingBlock.ancestors.include?(Liquid::Block)
+      assert_equal false, Jekyll::DoNothingOther.ancestors.include?(Liquid::Block)
+      assert_match "Jekyll::DoNothingBlock", Liquid::Template.tags["do_nothing"].name
+      assert_match "Jekyll::DoNothingOther", Liquid::Template.tags["do_nothing_other"].name
     end
 
     should "close open block tags, including custom tags, and ignore others" do
--- a/test/test_excerpt.rb
+++ b/test/test_excerpt.rb
@@ -274,4 +274,29 @@ class TestExcerpt < JekyllUnitTest
       assert_equal true, @excerpt.is_a?(Jekyll::Excerpt)
     end
   end
+
+  context "An excerpt with Liquid tags" do
+    setup do
+      clear_dest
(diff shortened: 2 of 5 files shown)


test/source/_plugins/custom_block.rb
```diff
@@ -0,0 +1,14 @@
...
+module Jekyll
+  class DoNothingBlock < Liquid::Block
+  end
+
+  class DoNothingOther < Liquid::Tag
+  end
+end
+
+Liquid::Template.register_tag("do_nothing", DoNothingBlock)
+Liquid::Template.register_tag("do_nothing_other", DoNothingOther)
```
ashmaroli: These should be under the `Jekyll` namespace..
```ruby
Liquid::Template.register_tag("do_nothing", Jekyll::DoNothingBlock)
Liquid::Template.register_tag("do_nothing_other", Jekyll::DoNothingOther)
```

test/test_excerpt.rb
```diff
@@ -274,4 +274,29 @@ def do_render(document)
...
+  context "An excerpt with Liquid tags" do
+    setup do
+      clear_dest
+      @site = fixture_site
+      @post = setup_post("2018-11-15-excerpt-liquid-block.md")
+      @excerpt = @post.data["excerpt"]
+
+      assert_includes @post.content.split("\n\n")[0].strip, "{% continue %}"
+      assert_equal true, DoNothingBlock.ancestors.include?(Liquid::Block)
+      assert_equal false, DoNothingOther.ancestors.include?(Liquid::Block)
```
ashmaroli: Here as well..
```ruby
      assert_equal true, Jekyll::DoNothingBlock.ancestors.include?(Liquid::Block)
      assert_equal false, Jekyll::DoNothingOther.ancestors.include?(Liquid::Block)
```

test/source/_plugins/custom_block.rb
kylebarbour: Doesn't putting them in `module Jekyll` automatically do that? When I tested it looked like they were.

test/test_excerpt.rb
kylebarbour: I'll update and add that :)

test/source/_plugins/custom_block.rb
kylebarbour: I've added the test below which seems to confirm that they're in the `Jekyll` namespace.

ashmaroli: Only if `Liquid` is registering the tag within the `Jekyll` module:
```ruby
module Jekyll
  class DoNothingBlock < Liquid::Block
    # [...]
  end

  Liquid::Template.register_tag("do_nothing", DoNothingBlock)
end

Comment (kylebarbour) [reply test/source/_plugins/custom_block.rb]: I'm so sorry to be dumb, but isn't that what's written above?

Comment (ashmaroli) [reply test/source/_plugins/custom_block.rb]: Do not be sorry. What I see in your plugin is:
```ruby
module Jekyll
  class DoNothingBlock < Liquid::Block
    # [...]
  end
end

Liquid::Template.register_tag("do_nothing", DoNothingBlock)
```

kylebarbour: Can `assert_match "Jekyll::DoNothingBlock"` pass if it's not in the Jekyll namespace though?

ashmaroli: It should raise an `uninitialized constant` exception, ideally. But honestly, I do not know why the tests are passing.

I extracted your "plugin" into a freshly generated Jekyll Project and it failed as expected..
:man_shrugging:

kylebarbour: OK! I'll just set it to be as you suggest :)

DirtyF:
@jekyllbot: merge +fix

ashmaroli:
@DirtyF This needs to be back-ported to `3.8-stable` too.. (Just a reminder)

PR #7383, opened 2018-11-16
Fix dead link and misleading prose

leafac:

https://github.com/jekyll/jekyll/pull/7381 left a few problems behind: a dead link to the section that was removed, and prose that can be confusing to the newcomer.


DirtyF:
@jekyllbot: merge +docs

PR #7385, opened 2018-11-18
Fix content management section

akshatkedia:

## Summary

Updated Siteleaf to match it's official name.

Remove Contentful because the link is broken and they no longer list Jekyll under their ecosystem page.


oe:
@jekyllbot: merge +docs

PR #7394, opened 2018-11-26
docs: Fix group_by_exp filter example

joeshannon:

## Summary

* Add missing colon to truncate filter

* In example output remove '...' since the truncate
filter is used with the optional argument (string
to append) which in this case is an empty string


joeshannon:
Although removing the '...' from the name is correct, perhaps it loses context?

pathawks:
Does @jekyllbot not see review comments?

@jekyllbot: merge +docs

PR #7395, opened 2018-11-28
adjust team page listings

oe:
- Removed ordered lists, and replaces them with unordered lists
- Remove myself as project lead
- Sort all person lists by name


mattr-:
@jekyllbot: merge +docs

PR #7404, opened 2018-12-06
Documents should be able to render their date

### LINKED ISSUE #7244: In a collection, `page.date` only generates on incremental builds

---

---

- [x] I had an error on GitHub Pages, and I have reproduced it locally.
- [ ] I was trying to build.
- [ ] It was another bug.

## My Reproduction Steps

In brief:

1. Create a vanilla Jekyll site with a collection. On a page in the collection, use `{{ page.date }}`. Do not add a `date` to the YAML frontmatter.
2. Serve without `--incremental`: the page date *does not* generate.
3. Serve with `--incremental`: the page date *does* generate.

I've created a repo to demonstrate the issue [here](https://github.com/arthurattwell/jekyll-page-date-bug), served on GH Pages [here](https://arthurattwell.github.io/jekyll-page-date-bug/).

Notes in more detail:

- when running `jekyll serve` without `--incremental`, pages in collections do not get a `page.date`;
- but when running `jekyll serve --incremental`, pages in collections do get the build timestamp as their `page.date`. That timestamp is the time of the last time the page built.

For instance:

1. Build the site with `jekyll serve` at 12:00. The `page.date` does not generate on the collection page.
2. Stop the server.
3. Build the site with `jekyll serve --incremental` at 12:01. The `page.date` *does* appear on the collection page as 12:00.

    So the `--incremental` build does rebuild the collection page, but using the previous build's timestamp.

4. With the incremental build still running, save the collection-page's markdown file to trigger a rebuild of that page only. Say this is at 12:02.
5. The collection page now *does* include the generated `page.date` with a new timestamp of 12:02.

Notes:

- Are collection pages meant to get an automatic date? [The docs](https://jekyllrb.com/docs/variables/#page-variables) suggest they should. So I assume that the page date that is generated with `--incremental` is not unexpected.
- Obviously this issue applies only when the collection page does *not* include a `date:` in its YAML page frontmatter. When page frontmatter includes a `date:`, that `page.date` always generates as expected.
- This behaviour is the same whether we `jekyll serve` or `jekyll build` and `jekyll build --incremental`.
- Pages outside collections (like this home page) never get an automatic `page.date`. This may or may not be expected behaviour but is not the subject of this test.

ashmaroli:
- This is a 🐛 bug fix.

## Summary

Documents should be able to relay their `date` attribute when queried for via `{{ page.date }}`

## Context

Fixes #7244

### DIFF
--- a/lib/jekyll/drops/document_drop.rb
+++ b/lib/jekyll/drops/document_drop.rb
@@ -12,7 +12,7 @@ module Jekyll
       mutable false
 
       def_delegator :@obj, :relative_path, :path
-      def_delegators :@obj, :id, :output, :content, :to_s, :relative_path, :url
+      def_delegators :@obj, :id, :output, :content, :to_s, :relative_path, :url, :date
 
       private def_delegator :@obj, :data, :fallback_data
 
--- a/lib/jekyll/drops/excerpt_drop.rb
+++ b/lib/jekyll/drops/excerpt_drop.rb
@@ -7,6 +7,10 @@ module Jekyll
         @obj.doc.data["layout"]
       end
 
+      def date
+        @obj.doc.date
+      end
+
       def excerpt
         nil
       end
--- a/test/test_document.rb
+++ b/test/test_document.rb
@@ -67,6 +67,12 @@ class TestDocument < JekyllUnitTest
       assert_equal "foo.bar", @document.data["whatever"]
     end
 
+    should "know its date" do
+      assert_nil @document.data["date"]
+      assert_equal Time.now.strftime("%Y/%m/%d"), @document.date.strftime("%Y/%m/%d")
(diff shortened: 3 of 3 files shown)


mattr-:
So, the code is fine. 🙂 My concern is that it doesn't fix the root of the problem.

Why does
https://github.com/jekyll/jekyll/blob/0f9467c61bab4f2fd53cc627185ab92cad11a838/lib/jekyll/drops/drop.rb#L86-L94 fail to make the `date` method on `Jekyll::Document` accessible to liquid as part of the drop? I think if we can answer that question, then we can fix this without adding more delegators to `Document::Drop`

mattr-:
Never mind, I got this confused. We have to add the `date` method to the drop classes because they won't be delegated to otherwise.

ashmaroli:
@jekyllbot: merge +bug

PR #7407, opened 2018-12-07
Remove alt attribute from a tags

mattmassicotte:

## Summary

I was following a long with this tutorial, when I noticed the use of the `alt` attribute in `a` tags. I didn't realized this was allowed, so I looked it up. It appears that it actually is not. Here are my sources:

https://www.w3.org/TR/html5/text-level-semantics.html#the-a-element
https://www.w3schools.com/tags/att_alt.asp

## Context

no related issues

## Semver Changes

doc-only change, so I don't think there is a version change needed

### DIFF
--- a/docs/_tutorials/navigation.md
+++ b/docs/_tutorials/navigation.md
@@ -53,7 +53,7 @@ docs:
 <h2>{{ site.data.samplelist.docs_list_title }}</h2>
 <ul>
    {% for item in site.data.samplelist.docs %}
-      <li><a href="{{ item.url }}" alt="{{ item.title }}">{{ item.title }}</a></li>
+      <li><a href="{{ item.url }}">{{ item.title }}</a></li>
    {% endfor %}
 </ul>
 ```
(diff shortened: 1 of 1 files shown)


ashmaroli:
@mattmassicotte Thank you catching this.
Can you replace the `alt` attribute with `title` instead?
```html
<a href="foo.html" title="Tool-tip">Link</a>
```

mattmassicotte:
@ashmaroli @DirtyF updated with title tags. Also added them in a few places where they were missing, for consistency in the document.

DirtyF:
> The title attribute is used to provide additional information to help clarify or further describe the purpose of a link. — https://www.w3.org/TR/WCAG20-TECHS/H33.html

Here we don't provide additional information, so I'll get rid of those redundant attributes.

ashmaroli:
@DirtyF Why do you think `title` attributes in the `<a>` tags are redundant?
Consider the following links by hovering over them:
<ul>
<li><a href="https://jekyllrb.com">Official Docs</a></li>
<li><a href="http://talk.jekyllrb.com" title="Jekyll Talk">Official Forum</a></li>
</ul>

ashmaroli:
@mattmassicotte I've removed the redundant attributes for you. Sorry for the hassle.

mattmassicotte:
Thanks everyone - glad to have helped (somehow :) )

PR #7408, opened 2018-12-08
Update collections.md

edouard-gv:
Having lists of collection type instead of attribute does not seem to work.

### DIFF
--- a/docs/_docs/collections.md
+++ b/docs/_docs/collections.md
@@ -13,7 +13,7 @@ example here's a collection of staff members:
 
 ```yaml
 collections:
-  - staff_members
+  staff_members
 ```
 
 You can optionally specify metadata for your collection in the configuration:


ashmaroli:
@edouard-gv Collections can configured via both YAML Lists and YAML dictionary of key-value pairs:
```yaml
collections:
  - members
  - tutorials
```
```yaml
collections:
  members:
    output: false
  tutorials:
    output: true
```

ashmaroli:
> does not seem to work

This statement is too vague. If you provide details as to why you think it does not seem to work, I can perhaps help you resolve the situation.

edouard-gv:
Undestood @ashmaroli . Moreover, I am sorry for my proposition is not working when no metadata is specified for a collection. I was just fooled by the fact that you have to change from a yaml list to a yaml map when adding metadata attribute to the collection. It thus more a yaml misunderstanding from myself. I let you judge if you like to add a simple parenthesis indicating that removing the "-" is he not a typo, without altering the simplicity of your documentation.

ashmaroli:
> my proposition is not working when no metadata is specified for a collection.

@edouard-gv Documents in a collection (other than posts) are not rendered into destination by default.
You have to explicitly set `output: true` in the collection's metadata.
This is alluded to, in our [documentation](https://jekyllrb.com/docs/collections/#output):

jekyllrb com_docs_collections_


> the fact that you have to change from a yaml list to a yaml map when adding metadata attribute to the collection.

Correct. I agree that our documentation is not clear in this aspect.

PR #7409, opened 2018-12-09
Fix grammatical error in permalinks.md

manumathew23:


DirtyF:
@jekyllbot: merge +docs

PR #7410, opened 2018-12-09
Dont write static files from unrendered collection

### LINKED ISSUE #7389: Collection assets still output even when output:false
## My Environment

| Operating System |    OSX 10.14.1        |
| `jekyll`         | 3.8.5     |
| `github-pages`   | No     |

A collection is defined in _config.yml and the output suppressed:
```
collections:
  my_collection:
    output: false
```

I have a mixture of images and markdown pages inside the _my_collection folder.
---

## Expected Behaviour

I expect all content in that collection to not be output, including assets like images. I understand there is a complex behaviour around _posts for historical reasons, however I'm only talking about custom collections here.

## Current Behavior

All markdown pages are suppressed, but other files go to the output. If the flag was called "page_output: false" then this would make sense, but the flag is just "output: false" which would lead me to believe we should be able to disable the whole collection at will.

ashmaroli:
- This is a 🐛 bug fix/
- I've added tests

## Summary

Collections other than `posts` can contain static assets along with Markdown files.
If a collection is initialized with metadata `output: false`, then neither its "documents" nor its "files" should be *output* to the destination directory.

## Context

Resolves #7389

### DIFF
--- a/features/collections.feature
+++ b/features/collections.feature
@@ -6,16 +6,20 @@ Feature: Collections
   Scenario: Unrendered collection
     Given I have an "index.html" page that contains "Collections: {{ site.methods }}"
     And I have fixture collections
+    And I have a "_methods/static-file.txt" file that contains "Static Content {{ site.title }}"
     And I have a configuration file with "collections" set to "['methods']"
     When I run jekyll build
     Then I should get a zero exit status
     And the _site directory should exist
+    But the _site/methods directory should not exist
     And the "_site/methods/configuration.html" file should not exist
+    And the "_site/methods/static-file.txt" file should not exist
 
   Scenario: Rendered collection
     Given I have an "index.html" page that contains "Collections: output => {{ site.collections[0].output }} label => {{ site.collections[0].label }}"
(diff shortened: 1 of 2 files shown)


DirtyF:
@jekyllbot: merge +fix

dcabo:
After upgrading our site to Jekyll 4 we noticed our `/_uploads` asset collection wasn't now being copied into the rendered site. It was set to `output: false`.

I've just found this issue that explains that this is not a regression, but an intended change. Which is fine with me. But I think then [the official documentation](https://jekyllrb.com/docs/collections/#add-content) is out of date, as it still describes the old behaviour: "If no front matter is provided, Jekyll will consider it to be a static file and copy it to the destination (e.g. _site) without processing." I would suggest adding a short note about this in the [3.x to 4.x migration guide](https://jekyllrb.com/docs/upgrading/3-to-4/): it took us a while to find out the reason for our issue, it could help others.

Thanks!

ashmaroli:
@dcabo Sorry for the inconvenience caused. It never occurred to me that this was a **`breaking-change`** and therefore never considered highlighting this change in the docs.

Thanks for reporting.

dcabo:
No worries, happy to help.

PR #7411, opened 2018-12-10
Assess impact of disabling Cache API on Utterson samples

ashmaroli:
**This is an `experiment`! Do not review**


ashmaroli:
**Loss of performance: 10 seconds**

ref | build time in seconds
-- | --
`#7411` | 150.66
`master` | 140.32

PR #7412, opened 2018-12-10
dev: test oldest and latest Ruby only

DirtyF:
This is a 🔧  CI config change. 

## Summary

Given we don't use any new features from Ruby > 2.3, we can speed up our CI builds a bit and save some resources 🌱  by not testing the intermediary Ruby version.


DirtyF:
@jekyllbot: merge +dev

PR #7413, opened 2018-12-12
Update 07-assets.md

dkalev:

## Summary

Updated some changes to the directory structure. With Jekyll 3.7.4 and minima theme the links to css can be added to `_includes/head.html` and the navbar is in `_includes/header.html`. Also clarified that we are a creating `styles.scss` and then referencing `styles.css`.


DirtyF:
@jekyllbot: merge +docs

PR #7419, opened 2018-12-13
Theme gems: ensure directories aren't symlinks

parkr:
This is a 🐛 bug fix.

## Summary

This PR fixes an issue where symlinks could be used by themes to bypass security protections. In this specific case, theme _directories_ could be symlinks (e.g. `theme_dir/assets -> /etc`) that break out of our very carefully-constructed cage. In this case, we want to re-apply the theme root (e.g. `theme_dir/assets -> /etc -> theme_dir/etc`).

This fix & test was written by @dbussink.

## Semver Changes

I'd like to backport this to 3.8 and 3.7.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,6 +28,7 @@ group :test do
   gem "rubocop", "~> 0.57.2"
   gem "test-dependency-theme", :path => File.expand_path("test/fixtures/test-dependency-theme", __dir__)
   gem "test-theme", :path => File.expand_path("test/fixtures/test-theme", __dir__)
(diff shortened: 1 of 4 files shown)


parkr:
> @parkr Can you add some inline comments above `Theme#realpath_for` as to why there are two calls to `Jekyll.sanitized_path` there..?

Sure!

parkr:
@ashmaroli updated!

parkr:
@jekyllbot: merge +bug

PR #7422, opened 2018-12-14
Add a custom inspect string for StaticFile objects

ashmaroli:
- I've added tests.

## Summary

Since a `StaticFile` object is initialized with the `site` instance, the object's default debug string contains the entire config data of the site.
Providing a custom string ensures that we get a simple result allowing faster debugging.
Additionally, this maintains consistency with core classes `Jekyll::Page` and `Jekyll::Document`

## Context

A direct response to the output in [test log](https://travis-ci.org/jekyll/jekyll/jobs/467560220#L738)

### DIFF
--- a/lib/jekyll/static_file.rb
+++ b/lib/jekyll/static_file.rb
@@ -152,6 +152,12 @@ module Jekyll
       @defaults ||= @site.frontmatter_defaults.all url, type
     end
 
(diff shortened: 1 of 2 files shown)


ashmaroli:
@jekyllbot: merge +minor

PR #7427, opened 2018-12-16
Allow custom sorting of collection documents

### LINKED ISSUE #5754: Specify order of collection items and set previous/next links accordingly

ashmaroli:

## Summary

Allow custom sorting of Document objects in collections by via the collection's metadata.
The two routes available are by either setting a `sort_by` metadata string or by setting an `order` array, with the latter route having a greater precedence.

## Context

Resolves #5754 
This is also a fresh take on #5904 

/cc @gvwilson

### DIFF
--- a/docs/_docs/collections.md
+++ b/docs/_docs/collections.md
@@ -111,6 +111,58 @@ You can link to the generated page using the `url` attribute:
 There are special [permalink variables for collections](/docs/permalinks/) to
 help you control the output url for the entire collection.
 
+{% if site.version == '4.0.0' %}{% comment %} Remove this encapsulation when v4.0 ships {% endcomment %}
+
+## Custom Sorting of Documents
+
+By default, documents in a collection are sorted by their paths. But you can control this sorting via the collection's metadata.
+
+### Sort By Front Matter Key
+
+Documents can be sorted based on a front matter key by setting a `sort_by` metadata to the front matter key string. For example,
+to sort a collection of tutorials based on key `lesson`, the configuration would be:
+
+```yaml
+collections:
+  tutorials:
+    sort_by: lesson
+```
+
+The documents are arranged in the increasing order of the key's value. If a document does not have the front matter key defined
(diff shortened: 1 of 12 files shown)


hosnas:
is this also a solution for [this pr](https://github.com/jekyll/jekyll/pull/5857)?

ashmaroli:
@hosnas No. This PR does not replace #5857

mattr-:
@jekyllbot: merge +major

PR #7429, opened 2018-12-18
Create symlink only if target is accessible

ashmaroli:
- Introduce a test-helper to generate symlinks only if the test environment has access to the target file. If access is denied, skip creating symlink (and therefore the test itself) with a terminal output
- Delete with `{ force: true }` so that the test doesn't abort if symlink wasn't created

Would like to backport this to stable branches which include the changed tests: `3.6-stable`, `3.7-stable` and `3.8-stable`

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -190,6 +190,13 @@ class JekyllUnitTest < Minitest::Test
       skip msg.to_s.magenta
     end
   end
+
+  def symlink_if_allowed(target, sym_file)
+    FileUtils.ln_sf(target, sym_file)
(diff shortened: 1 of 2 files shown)


parkr:
Without having looked too closely, could we do this with the theme dir symlink too?

ashmaroli:
> could we do this with the theme dir symlink too?

Yes, we could.
I left it out so that this commit could be backported without conflicting the existing backport PRs..

ashmaroli:
@jekyllbot: merge +dev

PR #7434, opened 2018-12-24
Replace `name` in Page#inspect with relative_path

ashmaroli:
- This is a 🐛 bug fix

## Summary

Pagination pages (from `jekyll-paginate`) have the same `name` attribute value : `index.html`.
Therefore, every pagination page will output the same `inspect` string.

Apart from pagination pages, the output would be the same for pages at paths `./index.md` and `./about/index.md`


DirtyF:
@jekyllbot: merge +fix

PR #7436, opened 2018-12-24
Bump JRuby version used for Travis CI builds

ashmaroli:
https://www.jruby.org/2018/12/06/jruby-9-2-5-0.html


ashmaroli:
Will be bumped as part of #7454.

PR #7438, opened 2018-12-25
Test with the new Ruby v2.6

ashmaroli:
- https://www.ruby-lang.org/en/news/2018/12/25/ruby-2-6-0-released/
- https://github.com/rvm/rvm/pull/4542


ashmaroli:
@jekyllbot: merge +dev

PR #7440, opened 2018-12-26
Remove warnings and fixes for deprecated config

ashmaroli:
## Summary

- Stop outputting Deprecation Warnings due to outdated configuration
- Stop "fixing" outdated configuration into the expected config and:
  - `Configuration#backwards_compatibilize` has been replaced with `#validate`
  - simply ignore outdated keys instead of assigning their values to their contemporary counterparts
  - simply raise `Jekyll::Errors::InvalidConfigurationError` if the value is invalid
- Deprecated method `Configuration#fix_common_issues` has been removed


lib/jekyll/configuration.rb
```diff
@@ -377,15 +292,12 @@ def check_maruku(config)
...
-        raise Jekyll::Errors::InvalidConfigurationError,
-              "'plugins' should not be a string, but was: " \
-              "#{config["plugins"].inspect}. Use 'plugins_dir' instead."
-      end
+      return unless config.key?("plugins")
+      return unless config["plugins"].is_a?(String)
+
+      raise Jekyll::Errors::InvalidConfigurationError,
+            "'plugins' should be set as an array, but was: #{config["plugins"].inspect}. " \
+            "Use 'plugins_dir' instead to set the directory of your plugins."
```
DirtyF: Maybe explain that one key if for gems, and the other is for Ruby files here?

```
Error: Gems plugins should be listed in the `plugins` array (#{config["plugins"].inspect} detected).
Use `plugins_dir` instead to set the directory for Ruby plugins. (`_plugins` by default)
```

ashmaroli: Yeah.. that's a good idea :+1:

jekyll.gemspec
```diff
@@ -59,6 +59,10 @@ Gem::Specification.new do |s|
       * Our `post_url` tag now comes with the `relative_url` filter incorporated into it.
         You shouldn't prepend `{{ site.baseurl }}` to `{% post_url 2019-03-27-hello %}`
         For further details: https://github.com/jekyll/jekyll/pull/7589
-    ----------------------------------------------------------------------------------
+
+      * Support for deprecated configuration options have been removed. We will no longer
```
DirtyF:
```suggestion
      * Support for deprecated configuration options has been removed. We will no longer
```

DirtyF:
@jekyllbot: merge +minor

PR #7448, opened 2019-01-02
Fix Interpreter warning from Jekyll::Renderer

mojavelinux:
Change OR operator to a lazy assignment in Jekyll::Renderer#layouts

This is a 🐛 bug fix.

## Summary

When warnings are enabled, Jekyll emits a rather insidious warning message (inidious because of how many times it occurs).

> warning: instance variable @layouts not initialized

This patch fixes the problem.

## Context

Based on other methods in the Jekyll::Renderer class, I believe the intention is to use a lazy assignment operator here instead of an OR operation.

If that's not the case, then I can update the patch to allow the OR operator to be used without emitting a warning.

## Semver Changes

Patch.

### DIFF
--- a/lib/jekyll/renderer.rb
+++ b/lib/jekyll/renderer.rb
@@ -9,6 +9,7 @@ module Jekyll
       @site     = site
       @document = document
       @payload  = site_payload
+      @layouts  = nil
     end
 
     # Fetches the payload used in Liquid rendering.


pathawks:
I don’t think the goal here is memorization.

Looking back at Git history, we used to set `@layouts` in the initalizer. I wonder if it would be safer to just explicitly initialize it properly; that way we could still check `layouts.nil?`
(I don’t know how important that is; just thinking aloud.

mojavelinux:
@pathawks That's the insight I was hoping for. I couldn't tell from the test suite which way to go here. All tests pass in either case.

If you'd like me to switch to just removing the warning (and not adding memoization), I can proceed accordingly (by setting the value in the initializer like you suggested).

pathawks:
> I can proceed accordingly (by setting the value in the initializer like you suggested).

Yeah, let’s go that route instead. 👍🏼

ashmaroli:
IMHO, there wouldn't be much of an issue because with this memoization, calling `renderer.layouts` is going to return the value of either `@layouts` or `site.layouts` anyway..

mojavelinux:
Code updated.

DirtyF:
Thanks @mojavelinux 👍 

@jekyllbot: merge +fix

mojavelinux:
Thank you all for reviewing!

PR #7471, opened 2019-01-11
Relax version constraint on classifier-reborn gem

ashmaroli:
To allow testing and building with latest version available


DirtyF:
@jekyllbot: merge +dev

PR #7472, opened 2019-01-11
Update resources.md

tschuhmacher:
deleted formester.com / not available, link redirects somewhere else

## Summary

Deleted the link to Formester from the list 'Forms', because it's not working as it should. The url formester.com redirects to strange places and Formester seems out of business.


DirtyF:
@jekyllbot: merge +docs

Issue #4056, opened 2015-10-26
Support for TextBundle


diethardsteiner:
I very much like the concept behind textbundle.
I am currently using Jekyll for generating my blog posts and one thing that is quite an inconvenience is that I always have to change the image file references when publishing the article. I think it would be great if not only Markdown apps supported textbundle but also static website generators like Jekyll, so that I don't have to change file references any more when publishing an article. 
You can find info about TextBundle [here](http://textbundle.org/). I also talked to the developers and they would be happy to work together with you.

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least
three months.

The resources of the Jekyll Core team are limited, and so we are asking for your help.

If you can still reproduce this error on the `3.0-stable` branch or on `master`,
please reply with all of the information you have about it in order to keep the issue open.

Thank you for all your contributions.

ijaaz:
I also would definitely like that!!

parkr:
Closing as stale. This would be a great plugin idea for anyone searching around 😉

PR #7473, opened 2019-01-12
Docs: store SSL key and cert in site source

arthurattwell:

This adds a small note that the SSL key and certificate must be stored in or symlinked from the site source when using `--ssl-cert` and `--ssl-key`.

Users (like me and in #4056) may assume that keys and certs can be stored anywhere. I have not found documentation that notes that they should be in the site source.

However, I am concerned that we should also note that keys and certs should never be added to source control. I have not added this because the Jekyll docs may not be the appropriate place to be giving security advice.


DirtyF:
@jekyllbot: merge +docs

Issue #7475, opened 2019-01-14
Jekyll-mention load error on "jekyll serve" or "bundle jekyll server"


Kimwangqing:
Hello, 

I met this issue when running "jekyll serve" or "bundle jekyll server" on bash,  but got blocked in an error:

$ bundle exec jekyll serve  
Configuration file: D:/Kim.github.io/_config.yml                                   Dependency Error: Yikes! It looks like you don't have jekyll-mentions or one o f its dependencies installed. In order to use Jekyll as currently configured, yo u'll need to install this gem. The full error message from Ruby is: 'cannot load  such file -- jekyll-mentions' If you run into trouble, you can find helpful res ources at https://jekyllrb.com/help/!                                            jekyll 3.8.5 \| Error:  jekyll-mentions
--

Actually, Jekyll-mentions gem existed in gem list jekyll, that means I have installed the gem. You can find the gem is included in gem list:

$ gem list jekyll                                                                
jekyll (3.8.5)                                                                   jekyll-feed (0.11.0)                                                             jekyll-gist (1.5.0)                                                              jekyll-mentions (1.4.1)                                                          jekyll-sass-converter (1.5.2)                                                    jekyll-seo-tag (2.5.0)                                                           jekyll-sitemap (1.2.0)                                                           jekyll-watch (2.1.2)
--

The running background is on Windows 8.1 and:
$ ruby -v                                                                        
ruby 2.4.5p335 (2018-10-18 revision 65137) [x64-mingw32]
$ gem -v                                                                         
2.6.14.3
$ bundle exec jekyll -v                                                          
jekyll 3.8.5
--

I installed jekyll-mentions version is the latest version, that is 1.4.1, and the runtime dependency is compatible with jekyll ~>3.0, and I installed jekyll is 3,8.5 version that is ranged the scope. 

Could you please guide me to go ahead? 

Cheers！

ashmaroli:
You should include `jekyll-mentions` in your Gemfile as well when you're running jekyll with `bundle exec`
```ruby
# Gemfile

gem "jekyll-mentions"
```

Kimwangqing:
I added "jekyll-mentions" in gemfile, and then ran "bundle exec jekyll server", but here is another error (I installed jekyll version 3.8.5):

$ bundle exec jekyll serve                                                       
[...]

PR #7476, opened 2019-01-14
Remind user to include gem in the Gemfile on error

ashmaroli:
Novice users don't know that running with `bundle exec` requires their plugin-gems to be listed in their Gemfile as well.

Related: #7475


ashmaroli:
> What about the `jekyll_plugins` group?

That's not a requirement in this context.

ashmaroli:
@jekyllbot: merge +minor

PR #7489, opened 2019-01-22
Update rubocop to ~> 0.63.1

### LINKED ISSUE #7480: Update dependency constraint to allow for rubocop v0.63.0
Hey there! :wave:

The constraint I found was `~> 0.62.0`, and the latest version available is `0.63.0`.

### LINKED ISSUE #7490: Update dependency constraint to allow for rubocop v0.63.1
Hey there! :wave:

The constraint I found was `~> 0.62.0`, and the latest version available is `0.63.1`.

luisenrike:
## Summary

Update rubocop to ~> 0.63.0

## Context

Fixes #7480
Fixes #7490

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,7 +26,7 @@ group :test do
   gem "nokogiri", "~> 1.7"
   gem "rspec"
   gem "rspec-mocks"
-  gem "rubocop", "~> 0.63.0"
+  gem "rubocop", "~> 0.63.1"
   gem "test-dependency-theme", :path => File.expand_path("test/fixtures/test-dependency-theme", __dir__)
   gem "test-theme", :path => File.expand_path("test/fixtures/test-theme", __dir__)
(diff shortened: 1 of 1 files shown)


ashmaroli:
`rubocop-0.63.1` is already out though..

ashmaroli:
Yes, that is correct. Since the bug in `0.63.0` has been resolved, we should not support it longer.
Say a user has only `rubocop-0.63.0` installed. If he were to clone the repo after this PR is merged, he would get an error when he runs our `script/fmt`.
But if we were to not support `v0.63.0`, he would be forced to update to the resolved version.

luisenrike:
👍 that makes sense, so it should be `gem "rubocop", "~> 0.63.0", ">= 0.63.1"`, right?

ashmaroli:
```
gem "rubocop", "~> 0.63.1"
```
should suffice IMO..

ashmaroli:
Thank you very much @luisenrike 

@jekyllbot: merge +dev

PR #7492, opened 2019-01-23
Upgrade kramdown dependency to v2.x

### LINKED ISSUE #7487: Update dependency constraint to allow for kramdown v2.0.0
Hey there! :wave:

I noticed that the constraint you have for kramdown doesn't allow for the latest version to be used.

The constraint I found was `~> 1.14`, and the latest version available is `2.0.0`.

### LINKED ISSUE #7509: Update dependency constraint to allow for kramdown v2.1.0
Hey there! :wave:

I noticed that the constraint you have for kramdown doesn't allow for the latest version to be used.

The constraint I found was `~> 1.14`, and the latest version available is `2.1.0`.

ashmaroli:
- This is an 🙋 enhancement.

## Summary

- Drop support for `kramdown-1.x`: It will no longer receive any updates.
- With v2.0, kramdown has split many of its accessory features into separate extensions shipped in individual gems:
    - kramdown-parser-gfm
    - kramdown-converter-pdf
    - kramdown-syntax-coderay
    - kramdown-math-mathjaxnode
    - kramdown-math-sskatex
    - kramdown-math-katex
    - kramdown-math-ritex
    - kramdown-math-itex2mml
- Since Jekyll has set `kramdown.input => "GFM"` by default, the extension `kramdown-parser-gfm` is listed as a `runtime_dependency` and will be loaded automatically if `kramdown.input` config hasn't changed.
- Every other extension **has to be listed in the `Gemfile`** (as applicable to the use-case) for utilizing the concerned feature.
- `kramdown-converter-pdf` extension will not be used by Jekyll Core. It is best suited to be handled by a plugin.

## Context

Resolves #7487 
Resolves #7509


ashmaroli:
What we mean to say is that we only deal with converting Markdown to HTML, with additional rendering of Math equations. To render Markdown to PDF, simply loading the `kramdown-converter-pdf` is not going to suffice. A dedicated plugin has to define the `Converter` subclass for it.

mattr-:
> To render Markdown to PDF, simply loading the `kramdown-converter-pdf` is not going to suffice. A dedicated plugin has to define the `Converter` subclass for it.

Can you call this out in the documentation? We shouldn't let people expect that using that particular kramdown plugin is going to just work like a lot of the others will.

ashmaroli:
> Can you call this out in the documentation?

I already have:
```
`kramdown-converter-pdf` will be ignored by Jekyll Core. Use a third-party plugin to convert Markdown to PDF.
```

mattr-:
Sorry. I wasn't clear earlier. 

I think it would be a good idea to provide guidance in the docs that a custom `Converter` subclass/plugin would be required in order to use that plugin.

ashmaroli:
@mattr-  Updated guide with an example plugin..

DirtyF:
@jekyllbot: merge +major

PR #7498, opened 2019-01-25
Don't check if site URL is absolute if it is nil

ashmaroli:
- This is a 🐛 bug fix.

## Summary

Resolves the following scenario:

```ruby
url = nil
Addressable::URI.parse(url).absolute?

# => undefined method `absolute?' for nil:NilClass
```

### DIFF
--- a/lib/jekyll/commands/doctor.rb
+++ b/lib/jekyll/commands/doctor.rb
@@ -161,7 +161,7 @@ module Jekyll
         end
 
         def url_absolute(url)
-          return true if url && Addressable::URI.parse(url).absolute?
+          return true if url && Addressable::URI.parse(url.to_s).absolute?
 
           Jekyll.logger.warn "Warning:", "Your site URL does not seem to be absolute, "\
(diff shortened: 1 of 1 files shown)


ashmaroli:
> How did this _ever_ work?

The boilerplate config file generated by `jekyll new` has `url: ''`. So, the chances of Doctor failing with an error were less.

lib/jekyll/commands/doctor.rb
```diff
@@ -161,7 +161,7 @@ def url_valid?(url)
         end
 
         def url_absolute(url)
-          return true if Addressable::URI.parse(url).absolute?
+          return true if url && Addressable::URI.parse(url.to_s).absolute?
```
parkr: Should we just `return false unless url` instead?

ashmaroli: I think we should allow just `String` instances to be parsed..

DirtyF:
@jekyllbot: merge +fix

PR #7503, opened 2019-01-29
Changed order of steps

byebyers:
When setting up CircleCI for my Jekyll site I needed to have the site built first before running HTMLProofer. So I switched their order to reflect that. Other than that this guide works like a charm!


DirtyF:
Thanks for the edit @byebyers 👍 

@jekyllbot: merge +docs

PR #7511, opened 2019-02-01
Line 31  replace bundle exec jekyll serve with jekyll serve

ajayianwansedo:
When i used bundle exec jekyll serve, Jekyll did not load in localhost:4000.
however, when i used jekyll serve, it worked!

### DIFF
--- a/docs/_docs/index.md
+++ b/docs/_docs/index.md
@@ -28,7 +28,7 @@ cd myblog
 ```
 5. Build the site and make it available on a local server
 ```
-bundle exec jekyll serve
+jekyll serve
 ```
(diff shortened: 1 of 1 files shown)


ashmaroli:
Sorry @ajayianwansedo This change will not be accepted, for the same reasons described by @chrisfinazzo

DirtyF:
On my system Jekyll is now aliased to bundler, it that a bundler 2 thing?

```
$ alias jekyll
jekyll=bundled_jekyll

$ which bundled_jekyll
bundled_jekyll () {
	_run-with-bundler jekyll $@
}
```

Issue #7516, opened 2019-02-10
Blog post timezone bug/misunderstanding


Ramobo:
# Example

Post date:
`date: 2019-02-10 15:00`

Configuration file:
`timezone: Etc/GMT`

For clarity below: my timezone is UTC-3.

# Expected behaviour

Built post date when viewing locally from UTC-3: 12:00

# Actual behaviour

Built post time when viewing locally from UTC-3: 15:00

They are 3 hours apart, but seemingly inverted.

From my understanding, the time on posts should be in the timezone specified in the configuration file, and when a user views that page, the displayed time is the post time converted from the configuration timezone to the user's system's timezone, but that doesn't seem to be the case.

ashmaroli:
@Ramobo What OS are you using?
What are your "expected" and "actual" behaviors for TZ offset at `UTC-5`

Ramobo:
I'm using Windows 10.

For UTC-5 (using 15:00 post time with the timezone Etc/GMT):
Expected: 10:00
Actual: 15:00
They don't seem to change with the system timezone.

Now I'm doing some more testing, and it looks like the post time is absolute, and the offset offsets the time by that offset, no matter your timezone. What's the point of that, then?

How are these offsets even supposed to work?

ashmaroli:
@Ramobo The offsets are supposed to work as following (w.r.t Jekyll on Windows and `tzinfo-v1.2.5`):
- A post's date is *absolute* if it doesn't specify the timezone
  ```yaml
  # Timezone-agnostic date
  date: 2019-02-10 15:00
  ```
  ```yaml
  # Timezone-dependent date
  date: 2019-02-10 15:00:00 -0300
  ```
- The offset is calculated w.r.t to the configured `timezone`

  date | timezone setting | result
  ---- | ---- | ----
  2019-02-10 15:00:00 -0300 | N/A | 2019-02-10 15:00:00 -0300
  2019-02-10 15:00:00 -0300 | **`Etc/GMT`** | 2019-02-10 18:00:00 +0000
  2019-02-10 15:00:00 -0300 | **`America/New_York`** | 2019-02-10 13:00:00 -0500

*Note: The above observations are for `gem 'tzinfo', '~> 1.0'`. `tzinfo-2.0` introduced changes to offset computation logic.*

Ramobo:
[...]

PR #7521, opened 2019-02-13
Upgrade WinTZ utility to use TZInfo-2.0

ashmaroli:
- This is an 🙋 enhancement.

## Summary

- TZInfo-2.0 returns local time with the correct UTC Offset.
- WinTZ utility will now generate inaccurate results with TZInfo-1.x

## Context

Discovered while investigating #7516

### DIFF
--- a/docs/_docs/upgrading/3-to-4.md
+++ b/docs/_docs/upgrading/3-to-4.md
@@ -46,6 +46,21 @@ you can invoke *`Liquid::Template`* directly:
   + template = Liquid::Template.parse(content)
   ```
 
+
+### Timezone in Windows
+
+Timezone utility for Jekyll on Windows now requires `tzinfo-2.0` and above.
+Simply add / update the gem listing in your `Gemfile`:
+
+```ruby
+# Gemfile
+
+# Windows does not include zoneinfo files, so bundle the tzinfo-data gem
+# and associated library.
+gem "tzinfo-data", platforms: [:mingw, :mswin, :x64_mingw, :jruby]
+gem "tzinfo", "~> 2.0"
+```
+
 ---
 
 *Did we miss something? Please click "Improve this page" above and add a section. Thanks!*
--- a/docs/_docs/upgrading/3-to-4.md
+++ b/docs/_docs/upgrading/3-to-4.md
@@ -49,7 +49,7 @@ you can invoke *`Liquid::Template`* directly:
 
 ### Timezone in Windows
 
-Timezone utility for Jekyll on Windows now requires `tzinfo-2.0` and above.
+Timezone handling for Jekyll on Windows now requires `tzinfo-2.0` and above.
 Simply add / update the gem listing in your `Gemfile`:
 
 ```ruby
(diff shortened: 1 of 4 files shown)


lib/jekyll/commands/new.rb
```diff
@@ -88,8 +88,14 @@ def gemfile_contents
             group :jekyll_plugins do
               gem "jekyll-feed", "~> 0.6"
             end
-            # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
-            gem "tzinfo-data", platforms: [:mingw, :mswin, :x64_mingw, :jruby]
+
+            # Windows and JRuby does not include zoneinfo files, so bundle the tzinfo-data gem
+            # and associated library.
+            install_if -> { RUBY_PLATFORM =~ %r!mingw|mswin|java! } do
```
XhmikosR: Any reason you don't use the same check as in the wdm gem? `Gem.win_platform?`

XhmikosR:
Sorry, I'm not sure what to test here being I never bothered with timezones and Jekyll. I only had one comment, not sure if it helps.

lib/jekyll/commands/new.rb
ashmaroli: I wanted to test for both Windows and JRuby with one operation.

ashmaroli:
> I'm not sure what to test here being I never bothered with timezones and Jekyll.

Okay, not an issue.
I saw that you're still a member of the `@jekyll/windows` team with review privileges. So I thought that you might have some insights on this PR overall.

Thank you for your time. :smiley:

docs/_docs/upgrading/3-to-4.md
```diff
@@ -46,6 +46,21 @@ you can invoke *`Liquid::Template`* directly:
   + template = Liquid::Template.parse(content)
   ```
 
+
+### Timezone in Windows
+
+Timezone utility for Jekyll on Windows now requires `tzinfo-2.0` and above.
```
mattr-: This sentence doesn't make a lot of sense to me. Perhaps `Timezone utilities` works here instead?

ashmaroli: It's in the singular because there's only one utility method for this in all of Jekyll: `Jekyll::Utils::WinTZ.calculate(timezone)`

Perhaps I should start the sentence with a definitive stance, then?
```diff
- Timezone utility for Jekyll on Windows now requires `tzinfo-2.0` and above.
+ The timezone utility for Jekyll on Windows now requires `tzinfo-2.0` and above.
```

mattr-: I'd rather be more generic and say `Timezone handling for Jekyll on Windows`...

It doesn't matter to the end user that there's only one. 🙂

XhmikosR:
@ashmaroli I meant I never bothered with Jekyll and timezones, not with Jekyll, and timezones :)

NP, feel free to CC me for Windows issues since it's my main OS.

DirtyF:
@jekyllbot: merge +minor

PR #7522, opened 2019-02-13
Correct example update to default.html

chapb:

## Summary
I believe the link to the stylesheet should be defined as "/assets/css/styles.scss" as per the above line "First create a Sass file at /assets/css/styles.scss with the following content:".

## Context
Correction in documentation. No relation to GitHub issue(s).

## Semver Changes
Unknown


ashmaroli:
@chapb We link to the `.css` file because the Sass file (with `.scss`) gets processed into CSS file (with `.css`) just like how `.html` files are generated from `.md` files..

chapb:
Sorry - just realized that as I was playing around with it further.

PR #7529, opened 2019-02-15
Compute relative_path for Page and StaticFile consistently

ashmaroli:
- This is a 🐛 bug fix.

## Summary

- Stripped leading slash from the  `relative_path` for a static file `'/assets/foo.js' => 'assets/foo.js'`
- Use the same algorithm to compute `relative_path` for `Page` and `StaticFile` instances
- Replace lazy-initialization of `relative_path` in `Page` instances with an `attr_reader`

### DIFF
--- a/docs/_docs/upgrading/3-to-4.md
+++ b/docs/_docs/upgrading/3-to-4.md
@@ -46,6 +46,20 @@ you can invoke *`Liquid::Template`* directly:
   + template = Liquid::Template.parse(content)
   ```
 
+### Relative path of Static files
+
+Static files no longer have a leading forward slash in their relative paths. So, if your plugin
+includes an additional step to remove the leading forward slash to bring it at par to those of
+a Page / Document object, then it is no longer necessary either.
+
+More importantly, if you're linking to static assets via our `{% link ... %}` tag, then please
+ensure that you correct them as otherwise the built-in validation will break your build:
+
+```diff
+- {% link /css/screen.css %}
++ {% link css/screen.css %}
+```
+
 ---
 
 *Did we miss something? Please click "Improve this page" above and add a section. Thanks!*
--- a/jekyll.gemspec
+++ b/jekyll.gemspec
@@ -52,6 +52,10 @@ Gem::Specification.new do |s|
       * Our `link` tag now comes with the `relative_url` filter incorporated into it.
         You should no longer prepend `{{ site.baseurl }}` to `{% link foo.md %}`
         For further details: https://github.com/jekyll/jekyll/pull/6727
+
+      * StaticFile objects no longer have a leading slash in their relative path
+        So please correct `{% link /css/screen.css %}` to `{% link css/screen.css %}`
+        For further details: https://github.com/jekyll/jekyll/pull/7529
     ----------------------------------------------------------------------------------
   MSG
 end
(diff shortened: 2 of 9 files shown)


DirtyF:
What problem are you trying to solve here?

ashmaroli:
> What problem are you trying to solve here?

@DirtyF Sorry that it is not readily graspable.
Consider the following code:
https://github.com/jekyll/jekyll/blob/9cb27144b03d8fc1bdd7d961501744038fa505be/lib/jekyll/tags/link.rb#L27-L28
See how we had to add additional logic to consider relative_path discrepancy in static files?
I intend to resolve that technical debt (which has effects elsewhere as well) with this PR.

DirtyF:
Sorry, but I still don't get what problem it solves for the user, except it adds a breaking change.

mattr-:
This breaks the convention that Jekyll operates from the top-level site directory and is confusing to the user about where the file actually comes from. If I have `css/screen.css` in a file in `_posts`, it implies that we'll look for `_posts/css/screen.css` rather than `$SITE_DIR/css/screen.css` which is what we'll actually be doing.

I appreciate the efforts you're putting in to remove technical debt. The technical debt here is ok to continue keeping for a better and more straightforward user experience. If you still want to normalize path computations and can do so without making a breaking change, I'll happily merge that PR. 🙂

ashmaroli:
@mattr- Thank you for looking into this.
*I'm simply stripping the leading slash in the relative_path of static files,* so that it is consistent with relative paths of a `Page` and a `Document`.

> If I have `css/screen.css` in a file in `_posts`, it implies that we'll look for `_posts/css/screen.css` rather than `$SITE_DIR/css/screen.css` which is what we'll actually be doing.

I agree with this in general. But that's where our `{% link ... %}` tag comes in.
**`{% link 'css/screen.css' %}` renders the generated `url` attribute instead of the physical relative path of the file.** So the result would simply be `{{ site.baseurl }}/css/screen.css` which points to the physical location `$DEST_DIR/css/screen.css`.

The breaking-change is only because our `link` tag will not be able to find an object with a `relative_path` attribute that *equals* given path. Apart from that there is no other negative effect for the end-user.

mattr-:
i'm coming at this from the principle of least surprise here. If I write in HTML, I would do it like:
```html
<link rel="/css/screen.css" type="stylesheet">
```
and it comes from the baseurl of my site, and everything is fine and I'm a happy user. Let's say that I run into a scenario where I actually need to care about `site.baseurl` and I discover that there's a fancy helper in Jekyll for that. I then translate my HTML above into liquid, like so:

```
{% link "/css/screen.css" %}
```

If it doesn't work like that, we break the principle of least surprise, make the user do more work due to looking up documentation, and with this change would break every single site that's using the `link` helper currently, all because we think that normalizing our code is more important than the user experience (which it's not).

I feel like people know that working with static files is going to be different than working with posts or collections, and we have tags for those sorts of things already, so it's a different use case, IMO.

ashmaroli:
Oh well.. :man_shrugging: 
With due respect, Jekyll 4.0 is already going to break sites (result in 404 links) using `{% link ... %}` tag because we have now incorporated the `relative_url` filter into the tag.

DirtyF:
If we break `link` tag, we might as well make it behave as expected: `{% link "/css/screen.css" %}`

PR #7530, opened 2019-02-15
Docs: check if var exists before include tag

jorgson:
Otherwise it will result in 

Liquid Exception: Invalid syntax for include tag. File contains invalid characters or sequences: Valid syntax: {% include file.ext param='value' param2='value' %} in FILE.html


DirtyF:
@jorgson Thanks for the edit 👍 

@jekyllbot: merge +docs

PR #7538, opened 2019-02-18
Clarify docs on collections regarding the need for front matter

kimbaudi:

## Summary

The document on Collections states that collection documents without front matter will not be generated, but this is not true. Those documents without front matter will be regarded as [static-files](https://jekyllrb.com/docs/static-files/) and copied over to the destination (e.g., `_site`).

You need to always add front matter to collections documents (especially .md files). If `output: true` is set in collections metadata, then collections documents with front matter will get generated (i.e., converted to .html and copied to `_site`). If `output: false` (the default value) is set in collections metadata, then collections documents with front matter will not get generated (i.e., not copied to `_site`).

## Context

Consider the following 2 collections configurations:

```shell
collections:
  foo:
    output: false
```

```shell
collections:
  foo:
    output: true
```

In both cases, if I have the following document `_foo/test.md` without front matter:

```md
# Test
```

When I build the site (e.g., `bundle exec jekyll build`), the content of `_site` folder is:

```
_site
└── foo
    └── test.md
```

But the doc states: 
> If no front matter is provided, Jekyll will not generate the file in your collection.

Now if I add front matter to the document `_foo/test.md`:

```md
---
---
# Test
```

Running `bundle exec jekyll build` with `output: false` will not generate the file in my collection. (e.g., `_site` will not have `foo/test.md`).

And running `bundle exec jekyll build` with `output: true` will generate the file in my collection. (e.g., `_sites` will have `foo/test.html`).

### DIFF
--- a/docs/_docs/collections.md
+++ b/docs/_docs/collections.md
@@ -45,7 +45,7 @@ Create a corresponding folder (e.g. `<source>/_staff_members`) and add
 documents. Front matter is processed if the front matter exists, and everything
 after the front matter is pushed into the document's `content` attribute. If no front
 matter is provided, Jekyll will consider it to be a [static file](/docs/static-files/)
-and copy it to the destination without processing (e.g. `_site`). If front matter
+and copy it to the destination (e.g. `_site`) without processing. If front matter
 is provided, Jekyll will not process and copy the file in your collection by default
 unless `output: true` is set in the collections metadata.
 
--- a/docs/_docs/collections.md
+++ b/docs/_docs/collections.md
@@ -44,7 +44,10 @@ collections:
 Create a corresponding folder (e.g. `<source>/_staff_members`) and add
 documents. Front matter is processed if the front matter exists, and everything
 after the front matter is pushed into the document's `content` attribute. If no front
-matter is provided, Jekyll will not generate the file in your collection.
+matter is provided, Jekyll will consider it to be a [static file](https://jekyllrb.com/docs/static-files/)
+and copy it to the destination without processing (e.g. `_site`). If front matter
+is provided, Jekyll will not process and copy the file in your collection by default
+unless `output: true` is set in the collections metadata.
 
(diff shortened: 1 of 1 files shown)


docs/_docs/collections.md
```diff
@@ -44,7 +44,10 @@ collections:
 Create a corresponding folder (e.g. `<source>/_staff_members`) and add
 documents. Front matter is processed if the front matter exists, and everything
 after the front matter is pushed into the document's `content` attribute. If no front
-matter is provided, Jekyll will not generate the file in your collection.
+matter is provided, Jekyll will consider it to be a [static file](https://jekyllrb.com/docs/static-files/)
```
ashmaroli: Please consider using host-agnostic links so that they can be traversed locally:
 ```suggestion
matter is provided, Jekyll will consider it to be a [static file](/docs/static-files/)
```

File: docs/_docs/collections.md [hunk 2]
```diff
@@ -44,7 +44,10 @@ collections:
...
+and copy it to the destination without processing (e.g. `_site`). If front matter
```
ashmaroli:
```suggestion
and copy it to the destination (e.g. `_site`) without processing. If front matter
```

docs/_docs/collections.md
```diff
@@ -44,7 +44,10 @@ collections:
...
+is provided, Jekyll will not process and copy the file in your collection by default
```
ashmaroli: Further clarification:
```suggestion
is provided, Jekyll will process the file in your collection but will not write to disk
```

docs/_docs/collections.md
```diff
@@ -44,7 +44,10 @@ collections:
...
+unless `output: true` is set in the collections metadata.
```
ashmaroli: typo:
```suggestion
unless `output: true` is set in the collection's metadata.
```

DirtyF:
@jekyllbot: merge +docs

PR #7542, opened 2019-02-20
Locate include file once per context

ashmaroli:
- This is an :alembic: experiment.

## Summary

Experiment to test optimization by reducing disk-access while rendering *includes*.


ashmaroli:
Hmm.. Disk-access cannot be gauged using VM environment of CI builds.. :man_shrugging: 
Needs results from the wild..

PR #7551, opened 2019-02-26
Test parsing date with zone in post's front matter

### LINKED ISSUE #7550: The wrong date can be used in the url
## My Environment

This is building on the latest version of github-pages.

## Expected Behaviour

For a post with a front matter date of `date: '2015-05-28 23:23:39.00000000 -05:00'`, the URL generated should have the date portion as `2015/05/28`. For additional (older) information, see #1069.

## Current Behavior

This issue reported in #1069 seems to still exist.
Look at ~~https://scottdorman.github.io/2015/05/29/introducing-jquery-gotop/~~ https://scottdorman.github.io/2015/05/28/introducing-jquery-gotop/ as an example.

The date portion of the link gets set to `2015/05/29` even though the front matter has `date: '2015-05-28 23:23:39.00000000 -05:00'` and my _config.yml has `timezone: America/New_York`. I have tried the front matter date in several variations and none of them seem to make any difference.

This issue affects a bunch of pages for me and also has an impact on some automatic redirection logic I have set up since the date portions of the links don't match up.

If there's something I need to change, I'm all for changing it but it seems like this is an issue with how Jekyll is parsing the date front matter variable to generate the URL. If the date has timezone information provided (which mine does) it should always use that timezone.

Interestingly enough, as I was writing this I decided to try setting the date to the corresponding UTC time (2015-05-29 03:23:39) which still had no effect. However, when I set the date back to the original and **removed** the time zone portion, it was generated with the correct date in the URL.  This behavior seems widly inconsistent, as other pages seem to work just fine with the time zone in the date (as a string) (see https://github.com/scottdorman/scottdorman.github.io/blob/master/_posts/2018/2018-08-18-saving-bootstrap-component-state.md as an example). It seems like where things start to go wrong is when the time approaches that roll-over period in UTC (so, for my time zone any dates earlier than 19:00:00 generate with the correct date and anything after would appear to generate with the wrong date).

## Code Sample

The repository where this is happening is https://github.com/scottdorman/scottdorman.github.io. See files/URLs referenced above for specific examples.

ashmaroli:
- I've added tests

## Context

Tests if #7550 is reproducible on `master` across platforms
Closes #7550 if and when CI builds successfully on Travis and AppVeyor.

### DIFF
--- a/features/post_data.feature
+++ b/features/post_data.feature
@@ -39,6 +39,17 @@ Feature: Post data
     And the _site directory should exist
     And I should see "Post date: 27 Mar 2009" in "_site/2009/03/27/star-wars.html"
 
+  Scenario: Use post.date variable with valid date and timezone
+    Given I have a _posts directory
+    And I have a configuration file with "timezone" set to "America/New_York"
+    And I have the following post:
+      | title     | date                       | content                 |
+      | Star Wars | 2009-03-27 23:23:39 -05:00 | Luke, I am your father. |
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "Luke, I am your father." in "_site/2009/03/27/star-wars.html"
+
   Scenario: Use post.date variable with invalid
     Given I have a _posts directory
     And I have a "_posts/2016-01-01-test.md" page with date "tuesday" that contains "I have a bad date."


mattr-:
Does this do anything to resolve #5963?

ashmaroli:
> Does this do anything to resolve #5963?

@mattr- Honestly, I did not even realize that a similar issue was previously reported. The current pull request is a direct reaction to the failing cucumber test in https://github.com/jekyll/jekyll/pull/7551/commits/4e68111595411b14c33014bfa3201b4bff497003 in this pull request.
(Notably, the tests **fail on Travis on all tested Ruby versions** but **not on AppVeyor**)

I'll try and test further to check if this issue is a regression due to https://github.com/jekyll/jekyll/commit/cf5c689d9414c7ba5b6300e44178b2735faf8c06 shipped in Jekyll-3.7.1

ashmaroli:
@mattr- From my investigation, the inference is that #5963 was resolved by #6697.
This PR attempts to clear out an aspect overlooked in #6697: post data with date, time and time zone.
I feel this issue was overlooked in the tests because the scenarios were set to UTC time zone.

*Disclaimer: I'm not currently sure if the tests here will fail when changes due to DST kick in..*

scottdorman:
What will it take to move this PR forward and get it merged (besides the obvious of fixing the merge conflicts)? It's been over a year since the PR was submitted and about 6 months since the changes in the PR were approved.

ashmaroli:
@scottdorman Sorry for the delay, but technically, we're sort of at a stalemate here.
The tests show that the behavior is flaky and changes with DST changes.

And then there's the issue with dependencies.. The current timezone management in Jekyll requires the gem `tzinfo-1.x` but then that would cause issues with an indirect dependency gem named `activesupport` whose newer versions do not support
`tzinfo-1.x`...

So in other words, its a messy situation..
Was hoping if someone from the community could propose a better solution..

scottdorman:
@ashmaroli Thanks for the update. On the surface, it seems like such a simple thing to fix. :) It seems like doing something around using UTC dates here would be the solution, or at least part of the solution, since that should remove the issues around DST changes.

PR #7554, opened 2019-03-05
[Feature] Theme inheritance

### LINKED ISSUE #7344: feat: gem-based themes inheritence
## Summary

Support for parent/child themes allowing users to build upon existing themes. 
See https://github.com/benbalter/jekyll-remote-theme/issues/35#issuecomment-429888249

## Motivation

 - develop starter themes and allow to build child themes upon
 - distribute child themes as gems

## Guide-level explanation

  - A child theme is a theme than expand an existing gem-based theme 

  - Only available for gem-based themes
  - A child can list a theme dependency in its gemspec (and `theme/_config.yml` ?)
  - Jekyll can apply a parent theme, and a child theme
  - Users have only to declare the child-theme in the main` _config.yml`

seshrs:

## Summary

This PR adds the concept of "theme inheritance" to Jekyll. This means that a developer who wants to make a small number of changes to an existing theme does not have to make a copy of all of the original theme's files. (This would also simplify the process of keeping the child theme up-to-date with the parent theme — they would simply have to bump up the parent theme's version number in the `gemspec`.)

Currently while building and rendering, Jekyll will first search for files (includes, layouts, assets and stylesheets) in the local directory, before searching the site's theme's directory. After this PR is merged, Jekyll would continue searching for these files in the theme's "parent" theme, and will continue recursively searching its parents as well. See #7374 for further context and discussion.

In addition to the code changes made to implement this feature, I have also added tests (and Cucumber feature tests), along with updating the documentation. To help make reviewing these changes easier, I have split my changes thematically across different commits.

## Context

Closes #7344.

***Note:** I am new to Ruby and to open-source contributions in general. I'd appreciate any feedback and suggestions to make this PR and the code changes conform better to the Jekyll standards. Thanks!*

### DIFF
--- a/lib/jekyll/layout.rb
+++ b/lib/jekyll/layout.rb
@@ -27,15 +27,19 @@ module Jekyll
 
     # Initialize a new Layout.
     #
-    # site - The Site.
-    # base - The String path to the source.
-    # name - The String filename of the post file.
-    def initialize(site, base, name)
+    # site  -  The Site.
+    # base  -  The String path to the source.
+    # name  -  The String filename of the post file.
+    # theme -  The Theme that contains this layout. Defaults to nil.
+    def initialize(site, base, name, theme = nil)
       @site = site
       @base = base
       @name = name
 
-      if site.theme && site.theme.layouts_path.eql?(base)
+      if theme && theme.layouts_path.eql?(base)
+        @base_dir = theme.root
+        @path = site.in_theme_dir_with_theme(theme, base, name)
+      elsif site.theme && site.theme.layouts_path.eql?(base)
         @base_dir = site.theme.root
         @path = site.in_theme_dir(base, name)
       else
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -31,17 +31,19 @@ module Jekyll
 
     # Initialize a new Page.
     #
-    # site - The Site object.
-    # base - The String path to the source.
-    # dir  - The String path between the source and the file.
-    # name - The String filename of the file.
+    # site  - The Site object.
+    # base  - The String path to the source.
+    # dir   - The String path between the source and the file.
+    # name  - The String filename of the file.
     def initialize(site, base, dir, name)
       @site = site
       @base = base
       @dir  = dir
       @name = name
-      @path = if site.in_theme_dir(base) == base # we're in a theme
-                site.in_theme_dir(base, dir, name)
+
+      theme = site.theme_with_root(base)
+      @path = if theme # we're in a theme
+                site.in_theme_dir_with_theme(theme, base, dir, name)
               else
                 site.in_source_dir(base, dir, name)
               end
--- a/lib/jekyll/readers/layout_reader.rb
+++ b/lib/jekyll/readers/layout_reader.rb
@@ -14,9 +14,11 @@ module Jekyll
           Layout.new(site, layout_directory, layout_file)
       end
 
-      theme_layout_entries.each do |layout_file|
-        @layouts[layout_name(layout_file)] ||= \
-          Layout.new(site, theme_layout_directory, layout_file)
+      site.theme_list.each do |theme|
+        theme_layout_entries(theme).each do |layout_file|
+          @layouts[layout_name(layout_file)] ||= \
+            Layout.new(site, theme_layout_directory(theme), layout_file, theme)
+        end
       end
 
       @layouts
@@ -26,8 +28,10 @@ module Jekyll
       @layout_directory ||= site.in_source_dir(site.config["layouts_dir"])
     end
 
-    def theme_layout_directory
-      @theme_layout_directory ||= site.theme.layouts_path if site.theme
+    def theme_layout_directory(theme)
+      return nil unless theme
+
+      theme.layouts_path
     end
 
     private
@@ -36,8 +40,12 @@ module Jekyll
       entries_in layout_directory
     end
 
-    def theme_layout_entries
-      theme_layout_directory ? entries_in(theme_layout_directory) : []
+    def theme_layout_entries(theme)
+      if theme_layout_directory(theme)
+        entries_in(theme_layout_directory(theme))
+      else
+        []
+      end
     end
 
     def entries_in(dir)
--- a/Gemfile
+++ b/Gemfile
@@ -31,6 +31,7 @@ group :test do
   gem "rubocop", Gem.win_platform? ? "~> 0.64.0" : "~> 0.65.0"
   gem "test-dependency-theme", :path => File.expand_path("test/fixtures/test-dependency-theme", __dir__)
   gem "test-theme", :path => File.expand_path("test/fixtures/test-theme", __dir__)
+  gem "test-theme-inheritance", :path => File.expand_path("test/fixtures/test-theme-inheritance", __dir__)
   gem "test-theme-symlink", :path => File.expand_path("test/fixtures/test-theme-symlink", __dir__)
 
   gem "jruby-openssl" if RUBY_ENGINE == "jruby"
--- a/docs/_docs/themes.md
+++ b/docs/_docs/themes.md
@@ -300,6 +300,30 @@ While this feature is to enable easier adoption of a theme, the restrictions ens
 This feature will let the theme-gem to work with *theme-specific config variables* out-of-the-box.
 {% endif %}
 
+{% if site.version == '4.0.0' %}
+{% comment %} Remove this encapsulation when `v4.0` ships {% endcomment %}
+
+### Theme inheritance {%- include docs_version_badge.html version="4.0.0" -%}
+
+If you are only adding features (or making small changes) to an existing theme, you don't have to make a copy of all the theme's files. Theme inheritance allows you to reuse a "parent" theme's assets, layouts, includes and stylesheets. To use theme inheritance, add to your theme's `gemspec` metadata:
+
+{% raw %}
+```ruby
+Gem::Specification.new do |s|
+  s.name        = "test-theme-inheritance"
+  
+  # Add the theme dependency
+  s.add_runtime_dependency("test-theme", "~> 1.0")
+  # And mark it as a parent
+  s.metadata    = {
+    "parent_theme" => "test-theme"
+  }
+end
+```
+{% endraw %}
+
+{% endif %}
+
 ### Documenting your theme
 
 Your theme should include a `/README.md` file, which explains how site authors can install and use your theme. What layouts are included? What includes? Do they need to add anything special to their site's configuration file?
--- a/features/theme_inheritance.feature
+++ b/features/theme_inheritance.feature
@@ -1,0 +1,52 @@
+Feature: Theme Inheritance
+  As a developer who wants to use a pre-existing theme
+  I want to be able to extend an existing theme
+  In order to distribute my changes without having to watch for changes in the parent theme
+
+  Scenario: A child theme using a parent theme's assets
+    Given I have a configuration file with "theme" set to "test-theme-inheritance"
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "From your theme" in "_site/assets/base.js"
+    And I should see "Another from theme inheritance" in "_site/assets/another.js"
+
+  Scenario: Layouts in the child theme overriding the parent theme's layouts
+    Given I have a configuration file with "theme" set to "test-theme-inheritance"
+    And I have an "test.md" file with content:
+      """
+      ---
+      layout: override
+      ---
+      """
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "override.html from test-theme-inheritance" in "_site/test.html"
+    And I should see ""
+
+  Scenario: Using layouts from the parent theme are used
+    Given I have a configuration file with "theme" set to "test-theme-inheritance"
+    And I have an "test.md" file with content:
+      """
+      ---
+      layout: test-layout
+      ---
+      """
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "test-layout.html from test-theme" in "_site/test.html"
+
+  Scenario: Layouts from the child theme should be available
+    Given I have a configuration file with "theme" set to "test-theme-inheritance"
+    And I have an "test.md" file with content:
+      """
+      ---
+      layout: unique
+      ---
+      """
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "unique.html from test-theme-inheritance" in "_site/test.html"
--- a/lib/jekyll/readers/theme_assets_reader.rb
+++ b/lib/jekyll/readers/theme_assets_reader.rb
@@ -10,22 +10,24 @@ module Jekyll
     def read
       return unless site.theme&.assets_path
 
-      Find.find(site.theme.assets_path) do |path|
-        next if File.directory?(path)
+      site.theme_list.each do |theme|
+        Find.find(theme.assets_path) do |path|
+          next if File.directory?(path)
 
-        if File.symlink?(path)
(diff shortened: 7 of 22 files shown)


ashmaroli:
Awesome First Contribution, @seshrs!!

I do have a couple of *minor* concerns though:
  - The sample spec in the documentation uses `{% raw %}..{% endraw %}`. This is unnecessary since there are no Liquid constructs (`{{...}}` or `{%...%}`) within. Did you try building the docs locally without them?
  - The sample spec in the documentation mentions listing the "parent-theme" as a `runtime_dependency`. But the "fixture theme" **doesn't have** such a declaration.
Moreover, recent versions of Jekyll attempts to automatically "load" all of the theme's runtime_dependencies. **I'm guessing** that Ruby will probably throw a `LoadError` exception if it tries to "load" a Rubygem that does not contain Ruby files.
  - The Cucumber feature does not **clarify** overriding at the users' end.
If a user has `<source_dir>/_layouts/home.html` with some customization, how does that play with a theme that in turn inherits from another theme-gem both of which contains a `_layouts/home.html` ?
  - Lastly, though not a Jekyll core-feature, I think the documentation could have a cursory mention regarding *possible incompatibility* with **GitHub's `remote-theme` feature**.

seshrs:
Thanks for your comments @ashmaroli! I've pushed a commit that addresses your suggestions for the docs and the feature specs. (Regarding `remote-theme`: If theme inheritance is added to Jekyll, I hope an analogous feature will become available on `remote-theme` as described in benbalter/jekyll-remote-theme#35.)

I'm not sure I understand Gemfiles and gemspecs well enough to address your second concern. The reason the fixture theme works without having a runtime dependency is because `test-theme` is already included in the [`Gemfile`](https://github.com/jekyll/jekyll/pull/7554/files#diff-8b7db4d5cc4b8f6dc8feb7030baa2478R33).
When I remove `test-theme` from the `Gemfile`, the tests fail as expected. However, when I then add `test-theme` as a runtime dependency to the `gemspec` instead, the test still fails with the error `Dependency Error: Yikes! It looks like you don't have test-theme or one of its dependencies installed...`, even if `test-theme` is included in the `Gemfile`.

In short, I am now unsure if adding a parent theme as a runtime dependency is sufficient/necessary, and I am unsure how to test this locally. Do you have thoughts or suggestions?

ashmaroli:
> the test still fails with the error `Dependency Error: Yikes! It looks like...`, even if `test-theme` is included in the `Gemfile`.

Thank you for testing this @seshrs.
This is exactly what I was concerned of. The `Dependency Error: Yikes!..` is simply a user-friendly wrapper in Jekyll for the `LoadError` exception from Ruby.

> I am now unsure if adding a parent theme as a runtime dependency is sufficient/necessary

My opinion is that a dependency declaration is beneficial for two reasons:
  - `runtime_depencies` are installed along with the concerned gem. This ensures that parent-theme(s) are installed along with the inherited-theme gem.
  - letting end-users know what theme(s) and version(s) the concerned theme-gem is inheriting from.

The following code handles the "automatic loading" of a theme's dependencies:
https://github.com/jekyll/jekyll/blob/68e633a56ec712acde4b2527b6f00125b419997a/lib/jekyll/plugin_manager.rb#L35-L46
It'd be dope if you can come up with a way to **not attempt loading gems** that are jekyll-themes.
Good luck!

seshrs:
Thanks for the pointer @ashmaroli! I've pushed a commit that prevents loading theme runtime dependencies that are in the chain of parent themes. Preliminary testing with external themes also seems to work as expected!

ashmaroli:
Thank you for addressing all of my concerns @seshrs. This is really an exciting feature to look forward to in v4.0. :crossed_fingers:

DirtyF:
I'd like if we could have feedback from theme developers here: @mmistakes @daviddarnes do any one of you plan to use such a feature? What should we check for here?

daviddarnes:
Thanks for looping me in @DirtyF, and big thanks for working on this @seshrs! It's sounding great and I'm excited to try this out. The immediate case would be theme developers using more vanilla themes as a basis for more bespoke themes. This is certainly what I'm planning on doing. Couple of queries:
- What happens with assets like scss files? Will it work in the similar manner to themes now?
- I believe somewhere there is an option to import the theme code you're using into the site so you can remove it as a dependancy (sorry but couldn't find the name of it). Will that command respect the inheritance order?

Thanks!

ashmaroli:
@seshrs I've added some commits to update the tests and fixtures in this branch. Looks like configuration from parent-theme(s) are not inheritable with the current implementation.

(The ability to read-in a config file from within a theme-gem is an unreleased `v4.0` feature.)

seshrs:
Thanks @daviddarnes and @DirtyF! In response to your questions:

- Stylesheets and assets would work just like layouts. A child theme can reference and use a parent theme's stylesheets and assets (just like local files can reference a theme's files). Similarly, a child theme can override stylesheets or assets from the parent theme by creating a file with the same name and same relative path (just like local files can override a theme's files).
I've added to `feature/theme_inheritance.feature` to document this behavior.
- I am not aware of the command, and couldn't find it in the few minutes I tried searching. (Given https://jekyllrb.com/docs/themes/#converting-gem-based-themes-to-regular-themes, it seems like removing a theme dependency needs to be done manually.) @DirtyF, @ashmaroli, do you know whether Jekyll provides an option to automate the process?

@ashmaroli: Ah, I missed that! I think I've fixed it in my most recent commit by modifying `Site::load_theme_configuration`. If the theme does not define its own `_config.yml`, Jekyll will continue to search its parent theme(s).

ashmaroli:
> I believe somewhere there is an option to import the theme code you're using into the site so you can remove it as a dependancy
> > it seems like removing a theme dependency needs to be done manually. @ashmaroli, do you know whether Jekyll provides an option to automate the process?

@daviddarnes If by "theme code" you meant layouts, includes, sass and assets, then Jekyll Core doesn't provide such an option / command. The closest is a [plugin I wrote](https://github.com/ashmaroli/jekyll-extract) (not under active maintenance)
@seshrs if by "theme dependency" you meant `runtime_dependency` of a theme, then no there's no option to disable their "automatic loading" by Jekyll. Perhaps, this PR should also test if `runtime_dependencies` are inherited from a parent-theme.. :wink:

seshrs:
@ashmaroli Good insight, I hadn't accounted for that! I've pushed a commit that fixes it and modified the tests to detect such an issue in future.

daviddarnes:
Ah that makes sense @ashmaroli, the jekyll-extract plugin is what I was thinking of

mattr-:
@seshrs This is looking good to me at first glance. I appreciate you taking the time to contribute.

There doesn't appear to be any backwards compatible changes here, so I'd like to save this feature for the 4.1 release. IMO, we've let the 4.0 release drag on long enough and we need to close it off and :ship: it.

Would you mind updating the references to `4.0` to `4.1` please?

seshrs:
Thanks @mattr-! I've updated the version references in the docs in bec42fe4dcdac29125c7a9df3acf4466de996805.

seshrs:
I merged from `master` to resolve merge conflicts, and updates some of the tests because the expected strings differed slightly.

However, the `theme_inheritance.feature` tests don't work any more. Specifically, they fail while importing a partial SASS file from the parent theme's `_sass` directory. I wasn't able to trace the root-cause of this error, but I noticed that there was a recent major release for [`jekyll-sass-converter`](https://github.com/jekyll/jekyll-sass-converter). Does something need to be changed for the SASS converter to detect parent-theme SASS files? /cc @ashmaroli @DirtyF 

(I'll continue to look into fixing this over the next few days.)

lib/jekyll/page.rb
```diff
@@ -31,17 +31,19 @@ class Page
...
     # Initialize a new Page.
     #
-    # site - The Site object.
-    # base - The String path to the source.
-    # dir  - The String path between the source and the file.
-    # name - The String filename of the file.
+    # site  - The Site object.
+    # base  - The String path to the source.
+    # dir   - The String path between the source and the file.
+    # name  - The String filename of the file.
```
ashmaroli: Why these changes..?

lib/jekyll/readers/layout_reader.rb
```diff
@@ -26,8 +28,10 @@ def layout_directory
       @layout_directory ||= site.in_source_dir(site.config["layouts_dir"])
     end
 
-    def theme_layout_directory
-      @theme_layout_directory ||= site.theme.layouts_path if site.theme
+    def theme_layout_directory(theme)
```
ashmaroli: You shouldn't change the signature of a `public` method. Please define a new method to wrap around the existing one.

ashmaroli (dismissed):

ashmaroli:
> Does something need to be changed for the SASS converter to detect parent-theme SASS files?

@seshrs Looks like there is a need to do that to support this feature.
Specifically, the v2.0 of the sass-converter sets up the sass partials' `load paths` instead of Jekyll.

lib/jekyll/readers/layout_reader.rb
DirtyF: @ashmaroli is that the only thing preventing us from shipping this feature? Can we help here to move this forward?

DirtyF:
Until we make sure the tests pass here, this won't make in 4.1.

lib/jekyll/readers/layout_reader.rb
seshrs: Sounds good. I've updated the PR to add a new method `theme_layout_directory_for_theme(theme)` instead.

seshrs:
My apologies for dropping this for so long 😓 But I want to try to push this to completion!

I figured out what changes need to be made to `jekyll-sass-converter`, and proposed jekyll/jekyll-sass-converter#114. I can confirm that the changes work together (I could get the spec in `theme_inheritance.feature` to pass locally). However, because of the dependencies, it looks like we will need three PRs to implement theme-inheritance:

1. Merge #8581 that adds the `Site::theme_list` method. It just returns an array containing the site's theme (if specified).
2. Merge jekyll/jekyll-sass-converter#114.
3. Merge this PR.

@ashmaroli, @DirtyF, does this sound reasonable to you? Let me know if I've missed something. Thanks for your patience and support! 😃

ashmaroli:
@seshrs Thank you for getting back on this.
Why was #8581 opened as a separate PR. I don't see how it has significance outside the context of the pull request here.

seshrs:
> Why was #8581 opened as a separate PR? I don't see how it has significance outside the context of the pull request here.

@ashmaroli I could be mistaken — I assumed that this PR and jekyll/jekyll-sass-converter#114 could only be merged if CI tests pass. But if it’s possible to merge these two PRs in conjunction without requiring CI tests to pass, then I will close #8581. :)

ashmaroli:
Yes CI needs to pass for us to be confident of the change.
Since there is no other use of #8581, please close that and include those changes here.. so that we need to track just two pull requests.

seshrs:
Okay sounds good! The changes from #8581 are already in this PR, so it’s ready for review. Let me know what I can do to help. Thanks!

ashmaroli:
Hmm.. the only workaround I can think of is utilizing a temporary Gemfile entry in both pull requests.
For example, edit the Gemfile in this repository to point to the other PR:
```diff
+ gem "jekyll-sass-converter", :github => "jekyll/jekyll-sass-converter", :ref => "ref/pull/114/head"
```
And a likewise counterpart in the sister PR..

seshrs:
😲 I made a similar change to the Gemfile to test locally, but I didn't realize I could also specify a GitHub ref! I've temporarily updated the Gemfiles for both PRs.

seshrs:
Hi! I just thought I’d check in on this PR. @ashmaroli do you think there are any other blockers yo address before we can merge this? 

I don’t think there are any backwards-incompatible changes here, so maybe this feature could make Jekyll 4.3? 🤞

ashmaroli:
@seshrs First of all, apologies for ignoring this proposal for so long.
I need sometime to re-wrap my head around this..
You could bring this branch up-to-date by running:

```sh
# where upstream points to this repository
git pull upstream master
```

lib/jekyll/layout.rb
```diff
@@ -15,15 +15,19 @@ class Layout
...
+    # base  -  The String path to the source.
+    # name  -  The String filename of the post file.
+    # theme -  The Theme that contains this layout. Defaults to nil.
+    def initialize(site, base, name, theme = nil)
       @site = site
       @base = base
       @name = name
 
-      if site.theme && site.theme.layouts_path.eql?(base)
+      if theme && theme.layouts_path.eql?(base)
```
ashmaroli, line 27:
```suggestion
      if theme&.layouts_path.eql?(base)
```

lib/jekyll/layout.rb
```diff
@@ -15,15 +15,19 @@ class Layout
...
+        @base_dir = theme.root
+        @path = site.in_theme_dir_with_theme(theme, base, name)
+      elsif site.theme && site.theme.layouts_path.eql?(base)
```
ashmaroli, line 30:
```suggestion
      elsif site.theme&.layouts_path.eql?(base)
```

lib/jekyll/page.rb
```diff
@@ -39,8 +39,10 @@ def initialize(site, base, dir, name)
       @base = base
       @dir  = dir
       @name = name
-      @path = if site.in_theme_dir(base) == base # we're in a theme
-                site.in_theme_dir(base, dir, name)
+
+      theme = site.theme_with_root(base)
+      @path = if theme # we're in a theme
```
ashmaroli, line 44: Unnecessary local variable initilization. Inlining is fine:
```suggestion
      @path = if site.theme_with_root(base) # we're in a theme
```

lib/jekyll/plugin_manager.rb
```diff
@@ -36,10 +36,14 @@ def require_gems
     #
     # Returns false only if no dependencies have been specified, otherwise nothing.
     def require_theme_deps
-      return false unless site.theme.runtime_dependencies
+      runtime_dependencies = site.theme_list.map(&:runtime_dependencies)
+      runtime_dependencies.flatten!
```
ashmaroli, line 40:
```suggestion
      runtime_dependencies = site.theme_list.map(&:runtime_dependencies).tap(&:flatten!)
```

ashmaroli (commented):
Some minor code-style updates.

ashmaroli:
This sure is a monster feature. Kudos for implementing it. Some additional comments:

- I don't see tests for multiple levels of inheritance. (Does the overriding work as expected without hiccups?  Does `site.theme_list` contain all the ancestors? etc etc..)

seshrs:
@ashmaroli I decided to add a new Cucumber feature for multi-level theme inheritance in f8000ccc5c2e521563e50636ce2294806b174ad1. The tests are very similar to the single-level inheritance feature descriptions. Let me know if you think those tests should live in a single feature file instead!

ashmaroli:
@parkr Could you take a look at this, please? Thanks.

parkr:
Adding theme inheritance (especially of N>2 themes) makes me nervous. I had a horrible time trying to write a child theme in WordPress once. It's just impossible to keep things in sync with the parent themes. I'm worried about the upgrade path (the child theme would need to be really strict about the parent theme's version) and I'm worried about N themes (child of child of child of child of child of...) being a headache for theme creators and users ("which file in which theme is this content coming from?"). 

PROs:
- less forking of themes for small modifications that can be applied to more sites (since the child theme is a theme)
- more playful sampling/customization of existing themes for suites of websites

CONs:
- Jekyll sites and themes tend to drift quickly and transparently to the user so upgrading could be really onerous.
- upgrading a new theme requires that all themes in the stack up to the desired theme are updated first. With one absentee maintainer (update, release to RG.org), we end up with forks and re-parenting the whole stack of themes
- customization requires knowing all themes pretty well since it's file-based overrides (we almost need mappings like for CSS/JS after minification or compilation)
- Jekyll remote themes (used heavily on GHP) makes gemspec reading and stuff potentially more confusing (since the theme is not in the Gemfile)

I'm open to it, but I'm not sure we've addressed the potential consequences here. What do you think about them?

PR #7557, opened 2019-03-08
Notify about 'github-pages' within :jekyll_plugins

ashmaroli:

## Summary

We shouldn't recommend adding `github-pages` to the `:jekyll_plugins` group because it results in automatic loading of all whitelisted dependencies irrespective of whether they're listed in the config file.

### DIFF
--- a/docs/_docs/github-pages.md
+++ b/docs/_docs/github-pages.md
@@ -32,16 +32,28 @@ gems.
 Note that GitHub Pages runs in `safe` mode and only allows [a set of whitelisted
 plugins](https://help.github.com/articles/configuring-jekyll-plugins/#default-plugins).
 
-To use the currently-deployed version of the gem in your project, add the
-following to your `Gemfile`:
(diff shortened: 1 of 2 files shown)


docs/_docs/github-pages.md
```diff
@@ -32,17 +32,29 @@ gems.
...
 ```
-
 Be sure to run `bundle update` often.
 
+<div class="note warning">
+  <h5><code>github-pages</code> gem and the <code>:jekyll_plugins</code> group</h5>
+  <p>
+    When <code>github-pages</code> gem is included in the <code>:jekyll_plugins</code>
+    group in your Gemfile, its whitelisted plugins are automatically enabled and loaded
+    into memory. This will result in longer build times and unneeded processing.
```
parkr: This is for the GitHub Pages documentation page, so I think we want the local experience to be as close to the true GitHub Pages experience. I vote we leave the group there since that's what you'll see on the website.

PR #7561, opened 2019-03-10
Install Docs that Work on MacOS 10.14

### LINKED ISSUE #3984: gem install jekyll failed on Mac OS X 10.11

### LINKED ISSUE #7608: docs: Small change in Jekyll on macOS - avoid root installs
## Motivation

Avoid the event where the user install `jekyll` and `bundler` as `root` on macOS, before telling them it's best not to do this. The way the documentation is written on [Jekyll on macOS](https://jekyllrb.com/docs/installation/macos/) this might happen. If the user is following along and using the commands they'll end up installing this as root.

## Suggestion

Change the recommendation of installing `jekyll` and `bundler` to default with `--user-install`, since this i recommended. Then (maybe) give the alternative method of installing them globally. Wouldn't someone who needs to install them globally know how to do this anyway?

liamrosenfeld:

## Summary

When following the current documentation on MacOS 10.14 you recieve the following error because of SIP Protections:

```
ERROR: While executing gem ... (Gem::FilePermissionError)
You don't have write permissions for the /Library/Ruby/Gems/2.0.0 directory.
```

This PR updates the instructions so that on Mojave you do not run into that error. It also improves the overall clarity of the page by organizing it clearer subsections.

## Context

Resolves #3984
Resolves #7608

### DIFF
--- a/docs/_docs/installation/macos.md
+++ b/docs/_docs/installation/macos.md
@@ -3,19 +3,19 @@ title: Jekyll on macOS
 permalink: /docs/installation/macos/
 ---
 
+## Install Command Line Tools
 First, you need to install the command-line tools to be able to compile native extensions, open a terminal and run:
 
 ```sh
 xcode-select --install
 ```
 
-## Set up Ruby
+## Install Ruby
 
 Jekyll requires Ruby > {{ site.min_ruby_version }}.
-As macOS Mojave 10.14 comes only with ruby 2.3.x, you'll have to install Ruby through Homebrew.
+As macOS Mojave 10.14 comes only with ruby 2.3.x, you'll have to install a newer version of Ruby.
 
-### Install latest stable Ruby with Homebrew {#brew}
-
+### With Homebrew {#brew}
 To run the latest Ruby version you need to install it through [Homebrew](https://brew.sh).
 
 ```sh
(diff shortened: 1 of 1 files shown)


liamrosenfeld:
@pathawks the only way to access `/usr/local/bin/`, which is where command line tools must be stored on MacOS 10.14, is using sudo. If there is another way just let me know and I'll update this immediately.

liamrosenfeld:
@pathawks and @chrisfinazzo, I updated it with a working local install method for Mojave. Please let me know what you think of it.

liamrosenfeld:
@pathawks, I updated it with a working local install method for Mojave and fixed the merge conflicts.

docs/_docs/installation/macos.md
```diff
@@ -95,11 +69,67 @@ Now you can install the Ruby version of our choice, let's go with current latest
...
+sudo gem install bundler
+sudo gem install -n /usr/local/bin/ jekyll
+```
+
+#### Before Mojave (<10.14)
+
+You only have to run:
+
+```sh
+sudo gem install bundler jekyll
```
DirtyF: 😨

liamrosenfeld: They are first presented with how to local install and the global install is immediately followed by a warning advising against it. But in the case they do want to global install, it seems like it would be better to save them the time to figure it out on 10.14.

liamrosenfeld:
@pathawks, I made local install front and center while the global install has a warning and is after the local install.

liamrosenfeld:
@pathawks, I implemented your requested changes by putting local install front and center while the global install has a warning and is after the local install.

If you want, I could remove the global section altogether, but in the rare cases that a global install would be preferred, I think it would be best if the docs save the user from being confused by random errors caused by SIP. 

Also, this fixes the gem path which caused an error on any local install because the docs currently contain an older version's directory.

mattr-:
@jekyllbot: merge +docs

liamrosenfeld:
Thank you!

PR #7566, opened 2019-03-11
Use communicative method parameters

ashmaroli:
## Summary

Communicative parameter names improve readability.
Changing method signature of `Configuration#renamed_key` because the *"unused parameter"* was never used even [at the time of its inception](https://github.com/jekyll/jekyll/commit/399606c54447b5c4853981ae52cf40c30195c3be).


DirtyF:
@jekyllbot: merge +fix

PR #7572, opened 2019-03-14
Fix unnecessary allocations via StaticFileReader

ashmaroli:
- This is a 🐛 bug fix.

## Summary

`Array#map` returns a new array derived from the given array after the iteration.
`Array#each` returns self after the iteration.

In `StaticFileReader#read`, `@unfiltered_content` is being returned instead of the array created by `Array#map`


ashmaroli:
@jekyllbot: merge +bug

PR #7577, opened 2019-03-17
Push Markdown link refs to excerpt only as required

### LINKED ISSUE #7576: Excerpt: Footnote definition <...> is unreferenced - ignoring
## For Reproducing
1. Clone https://github.com/TWiStErRob/repros
2. `cd repros/jekyll/excerpt-footnotes`
2. `bundle install`
3. `bundle exec jekyll build`

## Expected:
```
Configuration file: P:/projects/bugs/jekyll-footnotes/_config.yml
            Source: P:/projects/bugs/jekyll-footnotes
       Destination: P:/projects/bugs/jekyll-footnotes/_site
      Generating...
       Jekyll Feed: Generating feed for posts
                    done in 0.04 seconds.
```
## Actual:
```
Configuration file: P:/projects/bugs/jekyll-footnotes/_config.yml
            Source: P:/projects/bugs/jekyll-footnotes
       Destination: P:/projects/bugs/jekyll-footnotes/_site
      Generating...
       Jekyll Feed: Generating feed for posts
  Kramdown warning: Footnote definition for '1' on line 4 is unreferenced - ignoring
                    done in 0.04 seconds.
```

## Details
The problem seems to be that the abstract doesn't contain a reference to the footnote, which is a reasonable occurrence. The `excerpt_separator` makes no difference, I'm just using an explicit one on my site and it shows where it is better for this repro.

What makes the warning go away:
 * removing `jekyll-feed` from `Gemfile`
 * changing to `abstract [^1]`
 * downgrading to Jekyll 3.6.3, here's the diff between 3.6.3 and 3.7.4:
    ```
    Using jekyll-watch 2.1.2 (was 1.5.1)
    Using rouge 3.3.0 (was 2.2.1)
    Using jekyll 3.7.4 (was 3.6.3)
    ```

This only started showing up when I updated the version, very likely https://github.com/jekyll/jekyll/pull/6522 brought the problem to the surface. That PR is a really good change since it helps fixing issues in my md files, but having an issue I can't fix makes the build noisy and opens up the possibility of not seeing other warnings.

ashmaroli:
- This is a 🐛 bug fix.
- I've added tests

## Summary

Currently if Markdown reference-style links are present anywhere in a document's body, the reference definition(s) are included into the document's excerpt which results in kramdown throwing a warning (when configured to log warnings) for those reference definitions that are *not used* in the excerpt.

This PR proposes to push only those link definitions that are necessary during the excerpt's rendering.

*The additional changes are simply to reduce complexity scores as flagged by RuboCop*.

## Context

Resolves #7576

### DIFF
--- a/lib/jekyll/excerpt.rb
+++ b/lib/jekyll/excerpt.rb
@@ -139,7 +139,7 @@ module Jekyll
       return head if tail.empty?
 
       head = sanctify_liquid_tags(head) if head.include?("{%")
-      definitions = extract_mkdown_ref_link_defs(head, tail)
+      definitions = extract_markdown_link_reference_defintions(head, tail)
       return head if definitions.empty?
 
       head << "\n\n" << definitions.join("\n")
@@ -165,7 +165,7 @@ module Jekyll
       head
     end
 
-    def extract_mkdown_ref_link_defs(head, tail)
+    def extract_markdown_link_reference_defintions(head, tail)
       [].tap do |definitions|
         tail.scan(MKDWN_LINK_REF_REGEX).each do |segments|
           definitions << segments.join if head.include?(segments[0])
--- a/lib/jekyll/excerpt.rb
+++ b/lib/jekyll/excerpt.rb
@@ -132,33 +132,46 @@ module Jekyll
     # Returns excerpt String
 
     LIQUID_TAG_REGEX = %r!{%-?\s*(\w+)\s*.*?-?%}!m.freeze
-    MKDWN_LINK_REF_REGEX = %r!^ {0,3}\[[^\]]+\]:.+$!.freeze
+    MKDWN_LINK_REF_REGEX = %r!^ {0,3}(?:(\[[^\]]+\])(:.+))$!.freeze
 
     def extract_excerpt(doc_content)
       head, _, tail = doc_content.to_s.partition(doc.excerpt_separator)
-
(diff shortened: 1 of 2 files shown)


lib/jekyll/excerpt.rb
```diff
@@ -132,34 +132,47 @@ def render_with_liquid?
...
+
+        modified = true
+        head << "\n{% end#{tag_name} %}"
+      end
+
+      print_build_warning if modified
+      head
+    end
+
+    def extract_mkdown_ref_link_defs(head, tail)
```
mattr-:
```suggestion
    def extract_markdown_link_reference_defintions(head, tail)
```

If rubocop is going to complain here about the long function name, we should suppress that warning. I want longer, more readable method names more than I want to make rubocop happy.

mattr-:
Dope. Thanks!

@jekyllbot:  merge +bug

ashmaroli:
That's strange. Wonder why Jekyllbot did not respond..

mattr-:
@jekyllbot:  merge +bug

mattr-:
RIP jekyllbot. :wink:

DirtyF:
@jekyllbot: merge +bug

lib/jekyll/excerpt.rb
TWiStErRob: So this was not an issue in the end?

PR #7578, opened 2019-03-17
DRY: Delegate datetime methods in Drops::UrlDrop

ashmaroli:
## Summary

There is a lot of repetition among the date-time methods in `Jekyll::Drops::UrlDrop`.
This PR proposes to use some metaprogramming to simplify the class structure.


pathawks:
This seems to be a lot more difficult to understand. Isn’t DRY supposed to make code easier to read rather than more difficult?

mattr-:
This looks like a case where DRYing up the code actually leads to less readability than the duplication that's there. Let's leave this the way it is.

Issue #6939, opened 2018-04-19
Liquid `where` filter should fail gracefully when property does not exist 


calebhicks:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [x] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [x] I ran `jekyll doctor` to check my configuration
- [x] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [x] This is a feature request.

---

- [x] I am on (or have tested on) ***macOS*** 10+
- [ ] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux
- [ ] I am on (or have tested on) ***Fedora*** GNU/Linux
- [ ] I am on (or have tested on) ***Arch*** GNU/Linux
- [ ] I am on (or have tested on) ***Other*** GNU/Linux
- [ ] I am on (or have tested on) ***Windows*** 10+

---

- [ ] I was trying to install.
- [ ] There is a broken Plugin API.
- [ ] I had an error on GitHub Pages, and I have reproduced it locally.
- [ ] I had an error on GitHub Pages, and GitHub Support said it was a Jekyll Bug.
- [ ] I had an error on GitHub Pages and I did not test it locally.
- [x] I was trying to build.
- [ ] It was another bug.

## My Reproduction Steps

I have a number of pages that have a `context` value that may look like the following:

```yaml
{"Name"=>"Full Stack Web Development", "Curriculum Sets"=>["recaXooVhMqOWC7vJ", "recxZPsJ4zlmCbY6x", "recvV85PJcBALdsBg", "recneLUBwwQw4xPAs"], "Short ID"=>"cs-fsw", "type"=>"course", "id"=>"recUjPmBMgGVAwSA2", "base"=>"lambda-school-curriculum"}
```

I want to filter `site.pages` by the `context.type`, using the following code:

```
{% assign course_pages = site.pages | where: "context.type", "course" %}
```

Jekyll fails with the following trace:

```
  Liquid Exception: undefined method `[]' for nil:NilClass in /_layouts/default.html
bundler: failed to load command: jekyll (/Users/calebhicks/.rbenv/versions/2.4.1/bin/jekyll)
NoMethodError: undefined method `[]' for nil:NilClass
[...]

PR #7580, opened 2019-03-20
Detect `nil` and empty values in objects with `where` filter

### LINKED ISSUE #6038: Since 3.2, where-filter incorrectly handles missing/null properties

ashmaroli:
- This is an 🙋 enhancement.
- I've added tests 
- I've adjusted the documentation

## Summary

With this, one can pass `nil`, Liquid literals `empty` or `blank` to detect `nil` and empty values in the given array of objects.

*Additionally, this also slightly optimizes current operation by avoiding unnecessary array iteration due to chaining `#map` with `#include` (especially with an array property containing numerous elements)*

## Context

Resolves #6038
Closes #6039

### DIFF
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -161,13 +161,15 @@ module Jekyll
 
     # Filter an array of objects
     #
-    # input - the object array
-    # property - property within each object to filter by
-    # value - desired value
+    # input    - the object array.
+    # property - the property within each object to filter by.
+    # value    - the desired value.
+    #            Cannot be an instance of Array nor Hash since calling #to_s on them returns
+    #            their `#inspect` string object.
     #
     # Returns the filtered array of objects
     def where(input, property, value)
-      return input if property.nil? || value.nil?
+      return input if !property || value.is_a?(Array) || value.is_a?(Hash)
       return input unless input.respond_to?(:select)
 
       input    = input.values if input.is_a?(Hash)
@@ -182,8 +184,8 @@ module Jekyll
       # stash or retrive results to return
       @where_filter_cache[input_id][property][value] ||= begin
         input.select do |object|
-          Array(item_property(object, property)).map!(&:to_s).include?(value.to_s)
-        end || []
+          compare_property_vs_target(item_property(object, property), value)
+        end.to_a
       end
     end
 
@@ -321,6 +323,23 @@ module Jekyll
           end
         end
         .map!(&:last)
+    end
+
+    # `where` filter helper
+    def compare_property_vs_target(property, target)
+      case target
+      when NilClass
+        # won't catch `""`
+        return true if Array(property).empty?
+      when Liquid::Expression::MethodLiteral # `empty` or `blank`
+        return true if Array(property).join == target.to_s
+      else
+        Array(property).each do |prop|
+          return true if prop.to_s == target.to_s
+        end
+      end
+
+      false
     end
 
     def item_property(item, property)
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -329,8 +329,7 @@ module Jekyll
     def compare_property_vs_target(property, target)
       case target
       when NilClass
-        # won't catch `""`
-        return true if Array(property).empty?
+        return true if property.nil?
       when Liquid::Expression::MethodLiteral # `empty` or `blank`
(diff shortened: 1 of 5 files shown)


lib/jekyll/filters.rb
```diff
@@ -323,6 +325,23 @@ def sort_input(input, property, order)
         .map!(&:last)
     end
 
+    # `where` filter helper
+    def compare_property_vs_target(property, target)
+      case target
+      when NilClass
+        # won't catch `""`
+        return true if Array(property).empty?
```
mattr-: Why are we converting to an array here vs. just calling `.nil?` or something similar?

lib/jekyll/filters.rb
```diff
@@ -323,6 +325,23 @@ def sort_input(input, property, order)
...
+      when Liquid::Expression::MethodLiteral # `empty` or `blank`
```
mattr-: Are `empty` and `blank` the only values that would cause this `case` statement to match?

test/test_filters.rb
```diff
@@ -844,6 +844,11 @@ def to_liquid
         assert_equal 2, @filter.where(@array_of_objects, "color", "red").length
       end
 
+      should "filter objects with null properties appropriately" do
+        array = [{}, { "color" => nil }, { "color" => "" }, { "color" => "text" }]
+        assert_equal 2, @filter.where(array, "color", nil).length
```
mattr-: Why is `2` the value we're expecting here? `nil` and empty strings are not the same thing. Is a liquid thing and not a ruby thing?

lib/jekyll/filters.rb
ashmaroli: Ah! You're right. This can be simplified to remove the overlap with Liquid's `empty` literal. :+1:

lib/jekyll/filters.rb
ashmaroli: I'm not entirely certain but seems so from [this code](https://github.com/Shopify/liquid/blob/f2f467bdbc248e7bf26b4f5e552cfbce6052f811/lib/liquid/expression.rb#L20-L21):
I don't see any other special keywords documented on their site or wiki.

test/test_filters.rb
ashmaroli: This is due to code for compatibility with [existing **tests**](https://github.com/jekyll/jekyll/blob/6905c80470d403cb53a40d3e62798b72491903d3/test/test_filters.rb#L940-L955) that check for stringified values.

lib/jekyll/filters.rb
mattr-: Yup. That looks fine after looking at the liquid code now. :thumbsup:

ashmaroli:
@mattr-  In ticket #6939 the reporter mentions using the `where` filter as following:
```
{% assign course_pages = site.pages | where: "context.type", "course" %}
```
Is that a valid syntax for `where` filter? or should it be used **only as part of `where_exp` filter?**

mattr-:
> Is that a valid syntax for where filter? or should it be used only as part of where_exp filter?

No, I don't think that example is valid syntax. `type` in the example from #6939 is a nested attribute in another collection. This should probably be written as a combination of an iteration over `site.pages` and then using the `where` filter on `page.context` for the proper `type` value.

mattr-:
@jekyllbot: merge +minor

PR #7583, opened 2019-03-22
Target Ruby 2.4 syntax in RuboCop scans

ashmaroli:
## Summary

Now that `master` requires at least Ruby 2.4.0, let RuboCop suggest using Ruby 2.4 syntax over legacy syntax.

### DIFF
--- a/lib/jekyll/commands/serve/servlet.rb
+++ b/lib/jekyll/commands/serve/servlet.rb
@@ -186,7 +186,7 @@ module Jekyll
           key = res.header.keys.grep(%r!content-type!i).first
           typ = res.header[key]
 
-          unless typ =~ %r!;\s*charset=!
+          unless %r!;\s*charset=!.match?(typ)
             res.header[key] = "#{typ}; charset=#{@jekyll_opts["encoding"]}"
           end
         end
--- a/lib/jekyll/commands/serve/servlet.rb
+++ b/lib/jekyll/commands/serve/servlet.rb
@@ -29,7 +29,7 @@ module Jekyll
         end
 
         def inline?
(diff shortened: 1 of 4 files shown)


ashmaroli:
@DirtyF I made some changes to code associated with `livereload` as well.. Can you do some manual testing on macOS to be sure nothing's broken as such?
Thanks.

lib/jekyll/commands/serve/servlet.rb
```diff
@@ -29,15 +29,15 @@ def chunked?
         end
 
         def inline?
-          @response["Content-Disposition"].start_with?("inline")
+          @response["Content-Disposition"].to_s.start_with?("inline")
```
mattr-: why would these ever not be strings?

ashmaroli: In these tests: https://travis-ci.org/jekyll/jekyll/builds/509863125

mattr-: Can we use `@response.fetch("Content-Disposition", "")` instead? `fetch` conveys the intent better by saying "we expect this to be here, but in case it's not, use this value instead"

ashmaroli: We can't use `#fetch` here because `@response` is an instance of [`WEBrick::HTTPResponse`](https://ruby-doc.org/stdlib-2.4.0/libdoc/webrick/rdoc/WEBrick/HTTPResponse.html) which doesn't respond to `#fetch`

mattr-: :sob: :sob: :sob:

mattr-:
@jekyllbot: merge +dev

PR #7584, opened 2019-03-22
Remove Jekyll::Utils#strip_heredoc in favor of a Ruby > 2.3 built in

ashmaroli:
## Summary

Since Jekyll 4.0 requires at least Ruby 2.4, `Jekyll::Utils.strip_heredoc` is redundant as we may as well use the native operator `<<~`

This is a `breaking change` because we'll be removing `Jekyll::Utils.strip_heredoc` with this.


mattr-:
@jekyllbot: merge +major

PR #7589, opened 2019-03-27
Incorporate `relative_url` within `post_url` tag

### LINKED ISSUE #7353: feat: implement `relative_url` fil in `post_url` tag
## Summary

Implement URL filters for `post_url` tag to reflect `link`tag behavior and/or mark `post_url` as obsolete in v4.

## Motivation

Make links easier in Jekyll.

## Guide-level explanation

If we deprecate `post_url` Jekyll would raise an error at build:

```
Error: {% post_url 2017-10-03-an-old-post %} found in _posts/2018-02-04-a-more-recent-post.md 
post_url is now deprecated, please use link tag instead : {% link _posts/2017-10-03-an-old-post.md %}
```

See #6727

ashmaroli:
- I've added tests
- I've adjusted the documentation (via `spec.post_intall_message`)

## Summary

With this, `{% post_url 2019-03-27-hello %}` will return `/blog/2019/03/27/hello.html` for a site configured with `baseurl: blog`

**`BREAKING-CHANGE`**

## Context

Resolves #7353

### DIFF
--- a/docs/_docs/upgrading/3-to-4.md
+++ b/docs/_docs/upgrading/3-to-4.md
@@ -22,6 +22,25 @@ and fetch the latest version of Jekyll:
 gem update jekyll
 ```
 
+<div class="note warning">
+  <h5><code>post_url</code> Tag and Baseurl</h5>
+  <p>&nbsp;</p>
+  <p>
+    The <code>post_url</code> tag now incorporates the <code>relative_url</code> filter within itself
+    and therefore automatically prepends your site's <code>baseurl</code> to the post's <code>url</code>
+    value.
+  </p>
+  <p>
+    Please ensure that you change all instances of the <code>post_url</code> usage as following:
+  </p>
+
+{% highlight diff %}
+- {{ site.baseurl }}/{% post_url 2018-03-20-hello-world.markdown %}
++ {% post_url 2018-03-20-hello-world.markdown %}
+{% endhighlight %}
+</div>
+
+
 ## Template rendering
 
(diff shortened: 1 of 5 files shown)


ashmaroli:
@DirtyF I've documented this change within the gemspec's `post_install_message`. Isn't that enough?

chrisfinazzo:
As this is a breaking change for nearly everything, would it be appropriate to put this info in `warning` block to remind people they will need to fix their links? 🤔

ashmaroli:
> be appropriate to put this info in `warning` block

@chrisfinazzo Sounds fair. I'll add a warning block to the upgrading docs

chrisfinazzo:
@DirtyF, I guess... 👌🏼

My only point there was because this change affects a fundamental behavior that has worked in a certain way for _so_ _so_ _so_ long, it feels half-assed to _not_ highlight this since almost everyone will encounter it at some point.

DirtyF:
@chrisfinazzo we have a post-install message for everyone to see.

People might get confused though because the link path _will_ change and this can remained unnoticed because we won't output a warning in the console, and not everyone has automatic link validation with tools link `htmlproofer`.

DirtyF:
@jekyllbot: merge +major

mattr-:
I appreciate that there's a warning for this feature now. Not everybody is going to read the post install message. With a change like this, it's important to highlight it in as many ways as possible and I'm glad that's what we're doing here. Thanks everybody! :heart:

PR #7599, opened 2019-04-03
Update 07-assets.md

c0deflow:

## Summary
After I started the step by step tutorial, I got a bit confused at the point where I need to add the main.scss file. First, we get introduced to create a Sass file at /assets/css/styles.scss, so at this moment we are focused on the folder 'assets'. After that we should create a styles.css under _sass, for me it was not 100% clear if I should create this folder insede assets/_sass or /assets/css/_sass


ashmaroli:
Since one is introduced to the `assets` folder via path `/assets/css/styles.scss`, let's introduce `_sass` in the same fashion.
Please change:
```diff
- Create a `_sass/main.scss` on the root level with the following content:
+ create a Sass file at `/_sass/main.scss` with the following content:
```

DirtyF:
@jekyllbot: merge +docs

PR #7600, opened 2019-04-03
Fix link space

cogniti:
## Summary

Current code produces wrong link space.
This patch fixes the issue.


DirtyF:
@jekyllbot: merge +docs

PR #7607, opened 2019-04-07
Simplify couple of includes in the docs site

ashmaroli:
## Summary

Standard Liquid Filter `default` can be used to handle conditional rendering of two variables


DirtyF:
@jekyllbot: merge +docs

PR #7610, opened 2019-04-07
Avoid generating empty classnames

ashmaroli:
This change is to avoid generating empty classnames (e.g. `<li class="">...</li>` when the associated Liquid condition evaluates to `false`.

## Summary
`<li class="current">...</li>` => `<li class="current">...</li>` (no change)
`<li class="">...</li>` => `<li>...</li>`


ashmaroli:
@jekyllbot: merge +doc

PR #7613, opened 2019-04-11
DRY private methods in LiquidRenderer::Table

ashmaroli:
## Summary

Minor internal refactoring in `LiquidRenderer::Table` to DRY some private methods.

The diff may look confusing but the crux of this proposal is to introduce a private method `#inject_data` that takes an initial string, unfreezes it, then yields it to a given block and finally appends a `\n` to the resulting string before returning it.

### DIFF
--- a/lib/jekyll/liquid_renderer/table.rb
+++ b/lib/jekyll/liquid_renderer/table.rb
@@ -16,44 +16,43 @@ module Jekyll
       private
 
       def generate_table(data, widths)
-        str = +"\n"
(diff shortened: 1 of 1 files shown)


mattr-:
This refactoring isn't worth it, IMO. DRY is not a rule that must always be adhered to. This introduces an extra level of indirection that actually makes the code harder to understand and reason about. The duplication here is incidental and has no impact on the design or readability of the code in question.

PR #7617, opened 2019-04-15
Ideally, drops should be read-only

ashmaroli:
## Summary

Currently `Drops::UnifiedPayloadDrop` can be altered via `accessors` and via `:[]=`. This PR limits the writeable methods to just `:[]=`
i.e. Prefer `payload["content"] = document.output` over `payload.content = document.output`


mattr-:
There is not enough context here to make an informed decision regarding why this change is necessary, why it's bad that Drops aren't read-only, and why this is the appropriate fix. Could you please add some additional information here?

ashmaroli:
I don't have enough evidence to back this idea. Closing..

PR #7167, opened 2018-07-30
Document configuring gem environment in Ubuntu on Windows

argium:
Moved WSL/Ubuntu on Windows above native install as it's a lot simpler.
Updated WSL instructions to not run gem as sudo
Added troubleshooting steps to undo installing gem with sudo.

PR #7622, opened 2019-04-18
Update installation documentation for Windows, missing sudo

JarekKala:

## Summary

Added missing sudo to installation instructions for Windows via Bash.

### DIFF
--- a/docs/_docs/installation/windows.md
+++ b/docs/_docs/installation/windows.md
@@ -60,7 +60,7 @@ sudo apt-get install ruby2.5 ruby2.5-dev build-essential dh-autoreconf
 Next let's update our Ruby gems:
 
 ```sh
-gem update
+sudo gem update
 ```
 
 Now all that is left to do is install Jekyll.


ashmaroli:
@JarekKala  The Core Team has always advocated the "no sudo" policy. Why do you think its necessary..?

JarekKala:
@ashmaroli I needed it to run it when I did these steps. I got error:

ERROR:  While executing gem ... (Gem::FilePermissionError)
    You don't have write permissions for the /var/lib/gems/2.5.0 directory

I also needed sudo for next step so maybe I did something wrong.

ashmaroli:
@JarekKala You may have to follow the steps outlined in the PR #7167:
> [configure `gem`](https://jekyllrb.com/docs/troubleshooting/#no-sudo) so that packages are installed in your $HOME directory so `sudo` is not required.  Add the following to the bottom of your .bashrc file:
>
> ```sh
> # Ruby exports
>
> export GEM_HOME=$HOME/gems
> export PATH=$HOME/gems/bin:$PATH
> ```
>
>Then reload it:
>
> ```sh
> source .bashrc
> ```

Please try the above and let us know.. Thanks.

docs/_docs/installation/windows.md
```diff
@@ -60,7 +60,7 @@ sudo apt-get install ruby2.5 ruby2.5-dev build-essential dh-autoreconf
 Next let's update our Ruby gems:
 
 ```sh
-gem update
+sudo gem update
```
MichaelCurrin, line 63: The install step below is done _without_ sudo, therefore this should also exclude it.

In fact, if you need permissions errors, there are at least two solutions without using `sudo`:

- Change the permissions of the shared gems folder (`/usr/.../...`) to be editable by all users.
- Use `gem update --user-install` and `gem install GEM_NAME --user-install` to install to the user's `~/.gem` directory.

This is a complex topic - in fact `--user-install` may become an implied optional in newer `gem` command.

So it is best to leave this as it was.

MichaelCurrin (changes requested):

docs/_docs/installation/windows.md
MichaelCurrin:
```
gem update --help

...
        --[no-]user-install          Install in user's home directory instead
                                     of GEM_HOME.
```

PR #7625, opened 2019-04-19
Reduce allocations from Jekyll::Document instances

ashmaroli:
## Summary

The intention with this pull request is to swap non-destructive Array methods with their destructive counterparts to reduce allocations from the intermediate Array objects in `Jekyll::Document` objects (which are generally the most numerous in a Jekyll site).

In theory, there shouldn't be any noticeable change in performance because of this change because the amount of method calls remain more or less the same. However, there may be a slight improvement in large sites due to lesser GC cycles.


mattr-:
Allocations will affect performance. Is there any way we can get benchmark numbers from this change?

ashmaroli:
> Is there any way we can get benchmark numbers from this change?

@mattr- Yep there is a way.. However, I feel that any gains in performance here may get swallowed up in the `site.render` phase..

### Benchmark script
```ruby
# frozen_string_literal: true
# __FILE__ : benchmark.rb

require 'benchmark/ips'
require 'bundler/setup'
require 'jekyll'

def read_site
  site = Jekyll::Site.new(
    Jekyll.configuration(
      "source" => File.expand_path("docs", __dir__),
      "destination" => File.expand_path("docs/_site", __dir__),
      "skip_config_files" => true,
      "safe" => true
    )
  )
  site.read
  site.generate
end

Benchmark.ips do |x|
  x.report(" ORIGINAL Documents".yellow) { read_site }
  x.report("OPTIMIZED Documents".green) { read_site }
  x.hold! ".benchmark_temp"
  x.compare!
end
```
### BASH script (to execute the benchmark script..)
```bash
#!/usr/bin/env bash

# Fetch pull_request ref from remote ('upstream')
git fetch upstream pull/7625/head:reduce-allocations-from-documents

# Run benchmark - phase I with `master` branch
git checkout master
ruby benchmark.rb

# Run benchmark - phase II with pull request branch
git checkout reduce-allocations-from-documents
ruby benchmark.rb
```

ashmaroli:
### Profiler summary
```diff
--- master branch https://travis-ci.org/jekyll/jekyll/jobs/532959217
+++ PR branch     https://travis-ci.org/jekyll/jekyll/jobs/533119881

- Total allocated: 532.65 MB (5301707 objects)
+ Total allocated: 532.63 MB (5301387 objects)
  Total retained:  18.34 MB (96237 objects)
```

ashmaroli:
### Profiler summary
```diff
--- master branch https://travis-ci.org/jekyll/jekyll/jobs/532959217
+++ PR branch     https://travis-ci.org/jekyll/jekyll/jobs/533216008

- Total allocated: 532.65 MB (5301707 objects)
- Total retained:  18.34 MB (96237 objects)
+ Total allocated: 532.46 MB (5300602 objects)
+ Total retained:  18.34 MB (96245 objects)
```

lib/jekyll/document.rb
```diff
@@ -404,28 +418,29 @@ def respond_to_missing?(method, *)
...
     # Returns nothing.
     def categories_from_path(special_dir)
-      superdirs = relative_path.sub(%r!#{special_dir}(.*)!, "")
-        .split(File::SEPARATOR)
-        .reject do |c|
-        c.empty? || c == special_dir || c == basename
-      end
+      superdirs = relative_path.sub(Document.superdirs_regex(special_dir), "")
+      superdirs = superdirs.split(File::SEPARATOR)
+      superdirs.reject! { |c| c.empty? || c == special_dir || c == basename }
```
mattr-: It doesn't look like you need all three assignments for this.
```ruby
 superdirs = relative_path.sub(%r!#{special_dir}(.*)!, "").
        split(File::SEPARATOR).
        reject! { |c| c.empty? || c == special_dir || c == basename }
```

Whether or not you want to change it, I'll leave up to you. Just pointing it out. 😄

ashmaroli: There's a small catch though. I can't chain `:reject!` to the end because it will return `nil` in some cases.. :smiley:

mattr-:
@jekyllbot: merge +minor

PR #7627, opened 2019-04-21
Delegate --profile tabulation to `terminal-table`

ashmaroli:
## Summary

Delegate to a third-party gem ([`terminal-table`](https://github.com/tj/terminal-table)) instead of maintaining in-house code.

### Misc.
 - There is a very slight *change in output* in case anyone is parsing the output via a plugin.
 - *If it is of any relevance, [`github-pages`](https://rubygems.org/gems/github-pages/versions/197) gem also uses `terminal-table` to output tables..*


DirtyF:
@jekyllbot: merge +minor

PR #7638, opened 2019-04-29
Store list of expected extnames in a constant

ashmaroli:
This is a :zap: micro-optimization fix.

## Summary

Storing an invariable set of extensions in a class constant to avoid allocating memory (exactly 40bytes per array) for every *document* or *page* rendered.


mattr-:
Marking as a dev fix since this is better code design IMHO and doesn't have a massive impact on allocations (40Kb with 1000 documents) 

@jekyllbot: merge +dev

PR #7640, opened 2019-05-02
Doc: Data file section adds TSV

crispgm:

## Summary

Add TSV in data file doc, which is actually already supported.


DirtyF:
@jekyllbot: merge +docs

PR #7642, opened 2019-05-02
Allow excerpts to be generated for Page objects

### LINKED ISSUE #6778: Allow Jekyll::Page objects to render {{ page.excerpt }} with relevant content

ashmaroli:

## Summary

Allow excerpts to be generated for `Jekyll::Page` objects.

*Currently just a bare-bones feature, `{{ page.excerpt }}` for a `Jekyll::Page` will yield just the rendered output string of the Page's excerpt-source.*

## Context

Resolves #6778


ashmaroli:
Unfortunately since this breaks existing builds, it is not ready to move out of the *draft* status.

ashmaroli:
@jekyllbot: merge +minor

seth-reeser:
Hello, this had an unintended consequence, I would recommend stripping html, stripping newlines, and even maybe truncating. I was using page.excerpt as a meta description entry:

`<meta name="description" content="{% if page.excerpt %}{{ page.excerpt | strip_html | strip_newlines | truncate: 160 }}{% else %}{{ site.description }}{% endif %}">`

In an example of an `excerpt` not being defined for a page, this would have evaluated to the `site.description`, but now I'm getting markup generated for me that breaks because of HTML being included.

ashmaroli:
@seth-reeser Please open a new issue ticket and include the expected and actual results in the ticket for a discussion.

PR #7406, opened 2018-12-06
Add `type` attribute to Document instances

ashmaroli:
- This is a 🙋 feature or enhancement.
-  I've added tests
- The test suite passes locally

## Summary

- **Maintains consistency** between `Jekyll::Page` and `Jekyll::StaticFile` where instances the latter two classes respond to the `:type` method.
- **Optimizes Front Matter Defaults for Documents** by not needing to check for both `@doc.collection.label` and `@doc.collection.label.to_sym` as is done currently

PR #7643, opened 2019-05-03
Reduce string allocations with better alternatives

ashmaroli:
## Rationales

- interpolation results in a new object allocation for each call. So, interpolation within an iteration block is wasteful unless the interpolated variable is the block parameter.
- Regarding the changes to `frontmatter_defaults.rb`:
  - Private method `#path_is_subpath?` checks a derived string against a string value derived from the second argument to the method. Instead of generating a new string from a `Pathname` object, its better to just pass String objects directly.
https://github.com/jekyll/jekyll/blob/8673cdc4cf161fe21a693e26f608242372832a3c/lib/jekyll/frontmatter_defaults.rb#L136-L139
  - `slashed_coll_dir = collections_dir.empty? ? "/" : "#{collections_dir}/"` is better since it avoids allocating interpolated strings for default configuration. Surprisingly, its a lot *faster* too..!
    ```ruby
    # frozen_string_literal: true

    require 'benchmark/ips'

    def interpolate(collections_dir)
      "#{collections_dir}/"
    end

    def conditional(collections_dir)
      collections_dir.empty? ? "/" : "#{collections_dir}/"
    end

    ["", "foo"].each do |dir|
      return unless interpolate(dir) == conditional(dir)

      Benchmark.ips do |x|
        x.report("interpolate '#{dir}'") { interpolate(dir) }
        x.report("conditional '#{dir}'") { conditional(dir) }
        x.compare!
      end
    end
    ```
    ```
          interpolate ''    73.729k i/100ms
          conditional ''    88.071k i/100ms
          interpolate ''      1.991M (A± 0.7%) i/s -      9.953M in   5.000580s
          conditional ''      3.412M (A± 2.9%) i/s -     17.086M in   5.013017s

    Comparison:
          conditional '':  3411816.5 i/s
          interpolate '':  1990549.5 i/s - 1.71x  slower

       interpolate 'foo'    73.693k i/100ms
       conditional 'foo'    72.454k i/100ms
       interpolate 'foo'      1.992M (A± 0.6%) i/s -     10.022M in   5.031876s
       conditional 'foo'      1.911M (A± 0.8%) i/s -      9.564M in   5.006282s

    Comparison:
       interpolate 'foo':  1991828.9 i/s
       conditional 'foo':  1910513.6 i/s - 1.04x  slower
    ``` 

## Context

String allocations will be further reduced by shipping #7406


ashmaroli:
### Profiler summary
```diff
--- master branch https://travis-ci.org/jekyll/jekyll/jobs/532959217
+++ PR branch     https://travis-ci.org/jekyll/jekyll/jobs/532974783

- Total allocated: 532.65 MB (5301707 objects)
- Total retained:  18.34 MB (96237 objects)
+ Total allocated: 532.07 MB (5287525 objects)
+ Total retained:  18.34 MB (96241 objects)
```

lib/jekyll/entry_filter.rb
```diff
@@ -123,5 +123,11 @@ def glob_include?(enum, entry)
         end
       end
     end
+
+    private
+
+    def source_path
```
mattr-: :thumbsup:

reduces allocations and assigns a name to a concept in this class that was implicit before. thanks!

mattr-:
@jekyllbot: merge +minor

PR #7644, opened 2019-05-03
Indicate where the _sass folder is by default

bonjour-anuj:
To keep it consistent with rest of the documents where a file or directory is created at root level.


chrisfinazzo:
👋🏼 @bonjour-anuj, good to have you here.

mattr-:
@jekyllbot: merge +docs

PR #7649, opened 2019-05-06
add support for journaling

zimbatm:

## Summary

When maintaining a journal it doesn't make sense to name the file slug.
Just create one entry per day.

Actually coming up with a slug name before the post if even written
doesn't make sense. Half of the time I end-up renaming the slug, which
is cumbersome to do.

This change makes it possible to use vimwiki to create new post entries.


fullstopslash:
I actively want this feature as well.

MichaelCurrin:
Useful for me too.

What does one use as the title then? Blank, "Journal"?

zimbatm:
Ok, glad to hear that this could be merged if it gets fixed.

Something has changed in the code that ignores the slug and renders the page at `test/dest/2019/05/05/2019-05-05.html`. Debugging...

ashmaroli:
> Something has changed in the code

the reason could be the following:
```ruby
slug ||= basename_without_ext
```
That line ensures that `slug` is assigned `basename_without_ext` when `slug` was either `false` / `nil`.
Once that line gets evaluated, `slug` will be whatever value is stored in the variable `basename_without_ext`.

zimbatm:
I found two issues:
* `slug ||= basename_without_ext` as you mentioned. This is needed when neither regexp are matching so I cleaned-up the branching logic to make that distinction a bit clearer.
* `gsub!` was somehow messing with the "index" string. This was fixed by using `"index".dup`. I assume that `gsub!` is being used for performance reasons and to be the common case vs "index".

ashmaroli:
> * `gsub!` was somehow messing with the "index" string.

I don't see how a string `"index"` can choke `gsub!`. So your conclusion here is not justified. Needs more investigating.

ashmaroli:
> * This was fixed by using `"index".dup`

I see.
`"index"` is a frozen string. So you can't `gsub!` it.

zimbatm:
I fixed a suggestion from rubocop to replace `.dup` with a unary `+`. Everything is good except the windows CI, do you know if master is also broken on windows as well?

PR #7654, opened 2019-05-09
Encode and unencode urls only as required

ashmaroli:
This is an :zap: optimization change.

## Summary

Avoid wasting resources on encoding simple alphanumeric urls that additionally contain just a slash or a period or a dash.


pathawks:
If this is actually a significant performance boost, I will have lost all faith in Addressable.

mattr-:
@jekyllbot: merge +minor

PR #7655, opened 2019-05-13
Use the same Hash object as the default argument

ashmaroli:
- This is a **micro-optimization** change.

## Summary

Use a Module constant to avoid multiple allocations of default parameter hash.

## Rationale

Calling a method defined with a default parameter as an empty Hash object can lead to allocating a new empty Hash object for every call if the argument is omitted.

For example, in the script below:
```ruby
class Foo
  def read(opts = {})
    opts
  end
end

opts = { :mode => "read" }

100.times { Foo.new.read(opts) }
100.times { Foo.new.read }
```

The first `100.times` block only allocates 100 `Foo` objects.
But the second `100.times` block allocates 100 `Foo` objects and 100 `Hash` objects.

## Context

MemoryProfiler reports the follows stats for this change:

### Before
```
Total allocated: 9.96 MB (54896 objects)
Total retained:  743.62 kB (1998 objects)
```

### After
```
Total allocated: 9.95 MB (54706 objects)
Total retained:  743.62 kB (1998 objects)
```

### DIFF
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -40,6 +40,9 @@ require "i18n"
 SafeYAML::OPTIONS[:suppress_warnings] = true
 
 module Jekyll
+  # To allocate a single non-mutable Hash object as the default argument to methods applicable.
+  EMPTY_HASH = {}.freeze
+
   # internal requires
   autoload :Cleaner,             "jekyll/cleaner"
   autoload :Collection,          "jekyll/collection"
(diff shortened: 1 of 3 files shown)


mattr-:
I like the intent behind this but this is not a change I'm willing to accept because doesn't make a large enough difference to the memory profile for the changes being made and requires additional thought on the part of the developer to remember to use this new constant when writing new code.

PR #7657, opened 2019-05-13
Initialize mutations for Drops only if necessary

ashmaroli:
- This is a **micro-optimization** change.

## Summary

Since `@mutations` are *technically* not required for *immutable* Drops (has `mutable: false`), its better to move its initialization outside the Drop's constructor.

## Context

MemoryProfiler reports the following for this change: (run against the `docs/` site)

### Before
```
Total allocated: 9.96 MB (54896 objects)
Total retained:  743.62 kB (1998 objects)
```
### After
```
Total allocated: 9.95 MB (54555 objects)
Total retained:  730.02 kB (1658 objects)
```

### DIFF
--- a/lib/jekyll/drops/drop.rb
+++ b/lib/jekyll/drops/drop.rb
@@ -30,7 +30,6 @@ module Jekyll
       # Returns nothing
       def initialize(obj)
         @obj = obj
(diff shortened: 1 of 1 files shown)


mattr-:
@jekyllbot: merge +minor

PR #7658, opened 2019-05-13
Initialize and reset glob_cache only as necessary

ashmaroli:
- This is a 🐛 bug fix.

## Summary

`@glob_cache` shouldn't be initialized nor reset if `scope["path"]` doesn't contain glob patterns.

## Context

Improves change implemented by #7345


mattr-:
@jekyllbot: merge +bug

PR #7663, opened 2019-05-14
bump rubocop version to 0.69.0

### LINKED ISSUE #7660: Update dependency constraint to allow for rubocop v0.69.0
Hey there! :wave:

The constraint I found was `~> 0.68.0`, and the latest version available is `0.69.0`.

GabrielRMuller:
## Summary

Bumped Rubocop's version on the gemfile so it can update to its newest version

## Context
Resolves #7660

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,7 +26,7 @@ group :test do
   gem "nokogiri", "~> 1.7"
   gem "rspec"
   gem "rspec-mocks"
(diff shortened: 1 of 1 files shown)


ashmaroli:
Thank you for submitting this @GabrielRMuller. But there has already been a prior submission..
Duplicate of #7656

PR #7668, opened 2019-05-16
Coerce all urls to lowercase

### LINKED ISSUE #7667: Jekyll uses inconsistent URL for pages and posts under the same folder
## My Environment

| Operating System |      Windows 10      |
| `jekyll`         | No   |
| `github-pages`   | Latest     |

---

## Expected Behaviour
For below structure
```
root\
    LeetCode\
        _posts\
        index.md
```
I expect Jekyll to use same string for the folder name "LeetCode" in the URL for index.md and the posts like this:
- "/LeetCode/" for the page
- "/LeetCode/Post-Title/" for the posts

## Current Behavior
Jekyll use lowercase category name for the post URL so the URLs of page and posts are inconsistent:
- "/LeetCode/" for the index page
- "/leetcode/Post-Title/" for the posts (please be aware that "LeetCode" becomes lowercase "leetcode")

## Sample
You can head to my Github pages to verify what I said
- a post https://www.raymanblog.ml/leetcode/886-Possible-Bipartition/
- 404 with "leetcode" in URL https://www.raymanblog.ml/leetcode/
- accessible in this URL https://www.raymanblog.ml/LeetCode/

ashmaroli:
- This is a 🐛 bug fix.
- This is a :warning: **`breaking-change`**

## Context

Resolves #7667 
/cc @RaymanNg

### DIFF
--- a/features/post_data.feature
+++ b/features/post_data.feature
@@ -129,19 +129,6 @@ Feature: Post data
     And the _site directory should exist
     And I should see "Post category: movies" in "_site/movies/2009/03/27/star-wars.html"
 
-  Scenario: Use post.categories with uppercase characters
-    Given I have a BroCode directory
-    And I have a BroCode/_posts directory
-    And I have a _layouts directory
-    And I have the following post in "BroCode":
-      | title     | date       | layout | category | content                 |
-      | Star Wars | 2009-03-27 | simple | epic     | Luke, I am your father. |
-    And I have a simple layout that contains "Post category: {{ page.categories }}"
-    When I run jekyll build
-    Then I should get a zero exit status
-    And the _site directory should exist
-    And I should see "Post category: BroCode" in "_site/BroCode/epic/2009/03/27/star-wars.html"
-
   Scenario: Use post.tags variable
     Given I have a _posts directory
     And I have a _layouts directory
@@ -180,7 +167,7 @@ Feature: Post data
     When I run jekyll build
     Then I should get a zero exit status
     And the _site directory should exist
-    And I should see "Post categories: scifi and Movies" in "_site/scifi/movies/2009/03/27/star-wars.html"
+    And I should see "Post categories: scifi and Movies" in "_site/scifi/Movies/2009/03/27/star-wars.html"
 
   Scenario: Use post.categories variable when category is in YAML
     Given I have a _posts directory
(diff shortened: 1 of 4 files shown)


mattr-:
I'm not sure we can merge this. Not every system we support in Jekyll supports case sensitive file and directory names (thanks macOS). Coercing to lowercase is the only way we can maintain consistent behavior.

mattr-:
We should coerce page names to lowercase instead, so that everything is lowercase.

ashmaroli:
> We should coerce page names to lowercase instead, so that everything is lowercase.

Okay. That'd still make this a **`breaking-change`**.. Is that fine?

mattr-:
> Okay. That'd still make this a breaking-change.. Is that fine?

Yup. It's a good time to fix this bug, since we haven't frozen 4.0 for breaking changes yet.

mattr-:
@ashmaroli I was sweeping through open Jekyll issues and noticed you added the `do-not-merge` label to this. What's needed to resolve this PR?

pathawks:
If I manually specify a URL in the front matter of a page, there will be no way for me to include uppercase characters?
Can redirects use uppercase characters (a la the redirect plugin)?

ashmaroli:
@pathawks If this merges, one will not be able to include uppercase characters in their URL ever. The redirect plugin won't help either because this is like the *one function to rule them all..*

@mattr- I was apprehensive about the effects from merging this. So, took this [to the community](https://talk.jekyllrb.com/t/feedback-request-lowercase-urls-in-jekyll-4-0/2917/6). The following comment by @geraldb stands out:
>  *the good old Wikipedia, for example is using mixed (or dare I say “natural”) case e.g. https://en.wikipedia.org/wiki/Strange_Case_of_Dr_Jekyll_and_Mr_Hyde and I’m sure there are many others too.*

Ryuno-Ki:
Hm, I wonder, how this affects links to YouTube. Haven't checked, whether the casing there makes an actual difference …

alzeih:
I think jekyll should aim to support all the characters allowed in a URI, which includes uppercase:
https://tools.ietf.org/html/rfc3986#section-2.3

DirtyF:
I'd rather us stick to the RFC of the IETF here.

PR #7674, opened 2019-05-19
Add a utility method to test for absolute URI

ashmaroli:
## Summary

Proposing a utility method intended to be a drop-in replacement for `Addressable::URI.parse(string).absolute?` as the latter generates *a lot of intermediate objects* before testing the given string.

*Note: While using Addressable can raise an exception for an *invalid* string, the utility would just return boolean `false` for the same string.*

### DIFF
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -12,7 +12,8 @@ module Jekyll
         return if input.nil?
 
         input = input.url if input.respond_to?(:url)
-        return input if Addressable::URI.parse(input.to_s).absolute?
+        input = input.to_s
+        return input if Jekyll::Utils.absolute_uri?(input)
 
         site = @context.registers[:site]
         return relative_url(input) if site.config["url"].nil?
@@ -32,7 +33,8 @@ module Jekyll
         return if input.nil?
 
         input = input.url if input.respond_to?(:url)
-        return input if Addressable::URI.parse(input.to_s).absolute?
+        input = input.to_s
+        return input if Jekyll::Utils.absolute_uri?(input)
 
         parts = [sanitized_baseurl, input]
         Addressable::URI.parse(
--- a/lib/jekyll/utils.rb
+++ b/lib/jekyll/utils.rb
@@ -17,6 +17,13 @@ module Jekyll
     SLUGIFY_PRETTY_REGEXP = Regexp.new("[^[:alnum:]._~!$&'()+,;=@]+").freeze
     SLUGIFY_ASCII_REGEXP = Regexp.new("[^[A-Za-z0-9]]+").freeze
 
+    # Regular expression used to parse an URI.
+    #
+    # Obtained from: https://www.ietf.org/rfc/rfc3986.txt
+    #   (ref: 'Appendix B. Parsing a URI Reference with a Regular Expression')
+    # Also used by the Addressable gem to parse URIs.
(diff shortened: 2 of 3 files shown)


ashmaroli:
## Benchmark script
```ruby
# frozen_string_literal: true

require 'benchmark/ips'
require 'addressable/uri'

URI_REGEX = %r!^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?$!

def addressable(str)
  Addressable::URI.parse(str).absolute?
rescue => e
  false
end

def utility(str)
  str =~ URI_REGEX
  !!Regexp.last_match(1)
end

[
  "c:/assets/css/main.css",
  "://assets/css/main.css",
  "//assets/css/main.css",
  "/assets/css/main.css",
  "assets/css/main.css",
  "#://assets/css/main.css",
  "/#assets/css/main.css",
  "http://assets/css/main.css",
  "random://assets/css/main.css",
].each do |str|
  next unless addressable(str) == utility(str)

  Benchmark.ips do |x|
    x.report('addressable') { addressable(str) }
    x.report('utility') { utility(str) }
    x.compare!
  end
end
```

ashmaroli:
### Profiler summary
```diff
--- master branch https://travis-ci.org/jekyll/jekyll/jobs/534012098
+++ PR branch     https://travis-ci.org/jekyll/jekyll/jobs/534364958

- Total allocated: 515.58 MB (4903210 objects)
- Total retained:  18.27 MB (95954 objects)
+ Total allocated: 513.03 MB (4872802 objects)
+ Total retained:  18.27 MB (95954 objects)
```

mattr-:
Two questions:
- Do you have time based benchmarks? If we're going to replace `Addressable`, the data for doing so needs to be completely convincing.
- Why not go straight for the source? Addressable is a gem and is on GitHub. Landing changes there that would improve allocations would benefit the Ruby community as a whole, in addition to ourselves.

ashmaroli:
> Do you have time based benchmarks? ...

@mattr- I have [posted the code to benchmark](https://github.com/jekyll/jekyll/pull/7674#issuecomment-493732364) the two scenarios with various inputs. I did not directly post the results here because of the difference between Windows and Linux/macOS everyone else on Jekyll Core Team is using..

> Landing changes there that would improve allocations would benefit the Ruby community as a whole..

Yes, this is the ideal step for the entire FOSS community in general. I agree with you :100: I've opened PRs to "patch" Addressable (just not exactly this implementation).
But, there is a catch though..
Addressable has a certain way of going about its business: *First `parse` a string into an URI instance and then call methods onto those instances* (Not unlike what Jekyll does with *its* primitives, or what Liquid does with *its* Tags and Expressions,,)
Proposing such a fundamental change is *never a welcome contribution*. Even Jekyll wouldn't accept a proposal those is orthogonal to its current design..

ashmaroli:
@mattr- There has been no response to a proposal at https://github.com/sporkmonger/addressable/pull/344

ashmaroli:
@benesch Out of curiosity, does this have a positive effect on the build times for CockroachDB docs..?

benesch:
> @benesch Out of curiosity, does this have a positive effect on the build times for CockroachDB docs..?

Perhaps a bit! It's hard to say for sure, as I don't have the most scientific setup. Here are three timings without the patch, and three timings with the patch, as reported by `jekyll build`:

```
WITHOUT
14.489s
15.075
14.59

WITH
14.485s
13.8
14.093
```

ashmaroli:
> Here are three timings without the patch, and three timings with the patch

Thanks. The two sets more or less look the same. The minor gain is probably insignificant.

ashmaroli:
Closing for now.
May revisit in future..

PR #7679, opened 2019-05-24
Update log output for an invalid theme directory

### LINKED ISSUE #7630: `Invalid theme folder: _sass` when using Github Pages with remote_theme
## My Environment

| Operating System | tested on macOS 10.14 and Ubuntu Linux thru `circleci/ruby:2.4.1`           |
| `jekyll`         | Latest (v3.8.5)|
| `github-pages`   | Latest (v198, just released yesterday) |
| `Jekyll-remote-theme`   | Latest (v0.3.1) |

---

## Expected Behaviour

A normal build without the warning: `Invalid theme folder: _sass`.

## Current Behavior

Hi, we have a Jekyll theme build for our own Jekyll sites on a separate repo accessed thru [jekyll-remote-theme](https://github.blog/2017-11-29-use-any-theme-with-github-pages/) and all have been working just fine until yesterday that we upgraded from `v3.7.4` to `v3.8.5` (based on the Github [pages-gem v198](https://github.com/github/pages-gem/releases/tag/v198) release).

When using Jekyll `v3.8.x`, there is a warning showing us: `Invalid theme folder: _sass`, this warning goes away if Jekyll is downgraded to `v3.7.4`.

The warning isn't affecting us because the `/_sass` folder **does exists** on the Jekyll theme repo and all SCSS files are precompiled just fine. We think that this warning is telling us that the `/_sass` folder doesn't exist on the main repo but that it's expected and it shouldn't produce any warning.

Screen Shot 2019-04-24 at 12 15 56 PM

What do you suggest?

cetinajero:
## Summary
This fixes an issue since v3.8 where-in a non-actionable message is output to the terminal regarding a directory within current theme (gem / remote) that could either be missing, or be inaccessible, or includes a symlink loop.

- [x] Update to more verbose terminal output (327bf5b)
- [x] Change the log-level for this message to `debug` since it is not actionable for the end-user. (f5a3f40)

The proposed message is:
```
   Theme exception: The '_sass' directory does not exist at theme's root,
                    is either not accessible or includes a symbolic link loop
```

## Context

Fixes #7630

cc: @ashmaroli

### DIFF
--- a/lib/jekyll/theme.rb
+++ b/lib/jekyll/theme.rb
@@ -63,7 +63,8 @@ module Jekyll
       # However, symlinks are allowed to point to other directories within the theme.
       Jekyll.sanitized_path(root, File.realpath(Jekyll.sanitized_path(root, folder.to_s)))
     rescue Errno::ENOENT, Errno::EACCES, Errno::ELOOP
-      Jekyll.logger.warn "Invalid theme folder:", folder
+      Jekyll.logger.warn "Warning:", "The '#{folder}' directory does not exist at theme's root,"
+      Jekyll.logger.warn "", "is either not accessible or includes a symbolic link loop"
       nil
     end
 
--- a/lib/jekyll/theme.rb
+++ b/lib/jekyll/theme.rb
@@ -63,8 +63,8 @@ module Jekyll
       # However, symlinks are allowed to point to other directories within the theme.
       Jekyll.sanitized_path(root, File.realpath(Jekyll.sanitized_path(root, folder.to_s)))
     rescue Errno::ENOENT, Errno::EACCES, Errno::ELOOP
-      Jekyll.logger.warn "Warning:", "The '#{folder}' directory does not exist at theme's root,"
-      Jekyll.logger.warn "", "is either not accessible or includes a symbolic link loop"
+      Jekyll.logger.debug "Theme exception:", "The '#{folder}' directory does not exist at theme's root,"
+      Jekyll.logger.debug "", "is either not accessible or includes a symbolic link loop"
       nil
     end
 
--- a/lib/jekyll/theme.rb
+++ b/lib/jekyll/theme.rb
@@ -63,11 +63,11 @@ module Jekyll
       # However, symlinks are allowed to point to other directories within the theme.
       Jekyll.sanitized_path(root, File.realpath(Jekyll.sanitized_path(root, folder.to_s)))
     rescue Errno::ENOENT, Errno::EACCES, Errno::ELOOP => e
-      handle_exception(e, folder)
+      log_realpath_exception(e, folder)
       nil
     end
 
-    def handle_exception(err, folder)
+    def log_realpath_exception(err, folder)
       return if err.is_a?(Errno::ENOENT)
 
       case err
--- a/lib/jekyll/theme.rb
+++ b/lib/jekyll/theme.rb
@@ -62,10 +62,20 @@ module Jekyll
       # escape the theme root.
       # However, symlinks are allowed to point to other directories within the theme.
       Jekyll.sanitized_path(root, File.realpath(Jekyll.sanitized_path(root, folder.to_s)))
-    rescue Errno::ENOENT, Errno::EACCES, Errno::ELOOP
-      Jekyll.logger.debug "Theme exception:", "The '#{folder}' directory does not exist at theme's root,"
-      Jekyll.logger.debug "", "is either not accessible or includes a symbolic link loop"
+    rescue Errno::ENOENT, Errno::EACCES, Errno::ELOOP => e
+      handle_exception(e)
       nil
+    end
+
+    def handle_exception(err)
+      return if err.is_a?(Errno::ENOENT)
+
+      case err
+      when Errno::EACCES
+        Jekyll.logger.debug "Theme error:", "Directory '#{folder}' is not accessible."
+      when Errno::ELOOP
+        Jekyll.logger.debug "Theme error:", "Directory '#{folder}' includes a symbolic link loop."
+      end
     end
 
     def gemspec
--- a/lib/jekyll/theme.rb
+++ b/lib/jekyll/theme.rb
@@ -63,18 +63,18 @@ module Jekyll
       # However, symlinks are allowed to point to other directories within the theme.
       Jekyll.sanitized_path(root, File.realpath(Jekyll.sanitized_path(root, folder.to_s)))
     rescue Errno::ENOENT, Errno::EACCES, Errno::ELOOP => e
-      handle_exception(e)
+      handle_exception(e, folder)
       nil
     end
 
-    def handle_exception(err)
+    def handle_exception(err, folder)
       return if err.is_a?(Errno::ENOENT)
 
       case err
       when Errno::EACCES
-        Jekyll.logger.debug "Theme error:", "Directory '#{folder}' is not accessible."
+        Jekyll.logger.error "Theme error:", "Directory '#{folder}' is not accessible."
       when Errno::ELOOP
-        Jekyll.logger.debug "Theme error:", "Directory '#{folder}' includes a symbolic link loop."
+        Jekyll.logger.error "Theme error:", "Directory '#{folder}' includes a symbolic link loop."
       end
     end
 
--- a/Gemfile
+++ b/Gemfile
@@ -31,6 +31,7 @@ group :test do
   gem "rubocop-performance"
   gem "test-dependency-theme", :path => File.expand_path("test/fixtures/test-dependency-theme", __dir__)
   gem "test-theme", :path => File.expand_path("test/fixtures/test-theme", __dir__)
+  gem "test-theme-skinny", :path => File.expand_path("test/fixtures/test-theme-skinny", __dir__)
   gem "test-theme-symlink", :path => File.expand_path("test/fixtures/test-theme-symlink", __dir__)
 
   gem "jruby-openssl" if RUBY_ENGINE == "jruby"
--- a/features/theme.feature
+++ b/features/theme.feature
@@ -57,6 +57,17 @@ Feature: Writing themes
     And I should see "From your site." in "_site/assets/application.coffee"
     And I should see "From your site." in "_site/assets/base.js"
 
+  Scenario: A theme with *just* layouts
+    Given I have a configuration file with "theme" set to "test-theme-skinny"
+    And I have an "index.html" page with layout "home" that contains "The quick brown fox."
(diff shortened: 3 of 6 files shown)


lib/jekyll/theme.rb
```diff
@@ -63,7 +63,8 @@ def realpath_for(folder)
       # However, symlinks are allowed to point to other directories within the theme.
       Jekyll.sanitized_path(root, File.realpath(Jekyll.sanitized_path(root, folder.to_s)))
     rescue Errno::ENOENT, Errno::EACCES, Errno::ELOOP
-      Jekyll.logger.warn "Invalid theme folder:", folder
+      Jekyll.logger.warn "Path '#{folder}' does not exist, is not accessible "\
```
ashmaroli: Since we're aiming for clarity here, lets be verbose.. Split into two logger calls for shorter readable msgs:
```
Warning: The '#{folder}' directory does not exist at the theme's root,
         is either not accessible or includes a symbolic link loop
```

ashmaroli:
```suggestion
      Jekyll.logger.warn "Warning:", "The '#{folder}' directory does not exist at theme's root,"
      Jekyll.logger.warn "", "is either not accessible or includes a symbolic link loop"
```

cetinajero: Yes, I see what you mean.

With the two logger calls the message properly aligns with the rest of the log:

```
      Generating... 
           Warning: The '_sass' directory does not exist at theme's root,
                    is either not accessible or includes a symbolic link loop
      Remote Theme: Using theme ...
```

👍

ashmaroli:
I'm not so sure if `Jekyll.logger.debug "Warning:", "..."` gels well.. However, I'm not able to think of a better alternative either :man_shrugging:

cetinajero:
> I'm not so sure if Jekyll.logger.debug "Warning:", "..." gels well.. However, I'm not able to think of a better alternative either 🤷‍♂

haha yes!, I also thought that.

cetinajero:
What about `Theme exception:` @ashmaroli?, I have gave it some thoughts and that seems the best approach I can think of...

In my `bundle exec` with `debug` enabled it can be see it like this:

```
      Generating... 
             Theme: cetinajero/jekyll-theme-marketing
      Theme source: /private/var/folders/h1/95pfc94s5plcr_sjg6ypr64c0000gn/T/jekyll-remote-theme-20190524-14114-14lbpvi
   Theme exception: The '_sass' directory does not exist at theme's root,
                    is either not accessible or includes a symbolic link loop
      Remote Theme: Using theme cetinajero/jekyll-theme-marketing
      Remote Theme: Downloading https://codeload.github.com/cetinajero/jekyll-theme-marketing/zip/master to /var/folders/h1/95pfc94s5plcr_sjg6ypr64c0000gn/T/jekyll-remote-theme-20190524-14114-1vqogbd.zip
      Remote Theme: Unzipping /var/folders/h1/95pfc94s5plcr_sjg6ypr64c0000gn/T/jekyll-remote-theme-20190524-14114-1vqogbd.zip to /private/var/folders/h1/95pfc94s5plcr_sjg6ypr64c0000gn/T/jekyll-remote-theme-20190524-14114-14lbpvi

```

ashmaroli:
> What about `Theme exception:`

@cetinajero I had thought about `Theme Error:` but **dropped it** because it isn't an error every time. If a theme doesn't use Sass partials, then it is not an error / exception. Same applies if a theme doesn't use includes..

cetinajero:
So should we name it just a `Theme notice:` @ashmaroli?

ashmaroli:
We may have to settle with handling the Exception classes individually:
```ruby
def realpath_for(folder)
  Jekyll.sanitized_path(root, File.realpath(Jekyll.sanitized_path(root, folder.to_s)))
rescue Errno::ENOENT, Errno::EACCES, Errno::ELOOP => e
  handle_exception(e)
  nil
end

private

def handle_exception(err)
  case err
  when Errno::ENOENT
    nil # no-op
  when Errno::EACCES
    Jekyll.logger.debug "Theme Error:", "Directory '#{folder}' is not accessible."
  when Errno::ELOOP
    Jekyll.logger.debug "Theme Error:", "Directory '#{folder}' includes a symbolic link loop."
  end
end
```
whatsay?

cetinajero:
> whatsay?

Great @ashmaroli!

Can I only suggest 3 things?

 - the `realpath_for` is already a private function, it is defined here:

https://github.com/jekyll/jekyll/blob/f5a3f40ec1e1412e6566cd84d1e05990e175020e/lib/jekyll/theme.rb#L52-L59

  - What about omitting the following code block? It seems rather unnecessary besides explicity demonstrates at source code level that non action is intended. If you prefer to stick to it just let me know.

```ruby
  when Errno::ENOENT
    nil # no-op
```

Can I suggest going to `Theme error:` besides `Theme Error:`, this follows the same pattern as `Theme source:`

 https://github.com/jekyll/jekyll/blob/f5a3f40ec1e1412e6566cd84d1e05990e175020e/lib/jekyll/theme.rb#L12

## TL;DR:

This is my suggestion:

```ruby
def realpath_for(folder)
  Jekyll.sanitized_path(root, File.realpath(Jekyll.sanitized_path(root, folder.to_s)))
rescue Errno::ENOENT, Errno::EACCES, Errno::ELOOP => e
  handle_exception(e)
  nil
end

def handle_exception(err)
  case err
  when Errno::EACCES
    Jekyll.logger.debug "Theme error:", "Directory '#{folder}' is not accessible."
  when Errno::ELOOP
    Jekyll.logger.debug "Theme error:", "Directory '#{folder}' includes a symbolic link loop."
  end
end
```

ashmaroli:
I kept that "explicit case branch" since that is probably the most commonly occurring scenario out of the three. In such cases, Ruby need not check the other two branches before returning `nil`.

The alternative is a guard-cause:
```ruby
return if err.is_a?(Errno::ENOENT)

case err
[...]
```

cetinajero:
> ```ruby
> return if err.is_a?(Errno::ENOENT)
> ```

Neat! 👌 

Let me do this changes, do you prefer a whole rebase of this branch?

ashmaroli:
You need not force-push every commit. Our merge-bot `@jekyllbot` squashes all commits into one, at the time of merge..

cetinajero:
Got it

cetinajero:
Do you prefer to stick to the comments?

```
      # This resolves all symlinks for the theme subfolder and then ensures that the directory
      # remains inside the theme root. This prevents the use of symlinks for theme subfolders to
      # escape the theme root.
      # However, symlinks are allowed to point to other directories within the theme.
```

ashmaroli:
Let that comment stay as is. It is outside the scope of this PR..

ashmaroli:
I think we can now change the `log_level` to `:error` since a non-accessible directory or symlink-loop should be brought to the theme-author's notice and rectified..

cetinajero:
> I think we can now change the `log_level` to `:error` since a non-accessible directory or symlink-loop should be brought to the theme-author's notice and rectified..

Yes! haha... you are probably right, and now I'm also realizing that we need to pass the `folder` variable to `handle_exception`, check this:

```
jekyll 3.8.5 | Error:  undefined local variable or method `folder' for #<Jekyll::RemoteTheme::Theme:0x00007fcdbc979a78>
```

But Travis didn't see that 🤔, we should add a new test case don't you think?

ashmaroli:
> we need to pass the `folder` variable to `handle_exception`

Ah yes. My bad..
The overall diff looks fine now.

> we should add a new test case

Ideally, we should. But mocking a scenario involving a theme that contains only a certain directory, requires setting up fixtures.
You're welcome to give it a shot by using cues from existing tests. If you're not able to, I shall push a commit after a couple of days..

cetinajero:
> The overall diff looks fine now.

🎉 

> Ideally, we should. 

Actually I tried throughout the day but didn't come to anything useful at all, the test case is very unique and isolated but maybe on the next one I'll accomplish it.

lib/jekyll/theme.rb
```diff
@@ -62,11 +62,22 @@ def realpath_for(folder)
       # escape the theme root.
       # However, symlinks are allowed to point to other directories within the theme.
       Jekyll.sanitized_path(root, File.realpath(Jekyll.sanitized_path(root, folder.to_s)))
-    rescue Errno::ENOENT, Errno::EACCES, Errno::ELOOP
-      Jekyll.logger.warn "Invalid theme folder:", folder
+    rescue Errno::ENOENT, Errno::EACCES, Errno::ELOOP => e
```
mattr-: I think it makes more sense here to not rescue all three exceptions, but to rescue them individually, so that it's more explicit what's happening here without the extra indirection of the method call
```suggestion
    rescue Errno::ENOENT
      # it's ok for things to not exist inside of the theme
    end
    rescue Errno::EACCES => e
      Jekyll.logger.error "Theme error:", "Directory '#{folder}' is not accessible."
    end
    rescue Errno::ELOOP => e
      Jekyll.logger.error "Theme error:", "Directory '#{folder}' includes a symbolic link loop."
    end
```

We can do that now, save it for a later PR, or just let it drop. I'm not going to be that picky about it.

cetinajero: 👍 👍 

Then every rescue should include a `nil` so it doesn't return a truthy, right? (see https://github.com/jekyll/jekyll/pull/5195/commits/0278abe3f95a74e07ddd5334b6de8f167aaa5056)

ashmaroli: There's a chance that RuboCop might complain about having too many lines in a method definition or perhaps some other `Metric` cop flags in.. So let us defer this change for a future PR.

lib/jekyll/theme.rb
```diff
@@ -62,11 +62,22 @@ def realpath_for(folder)
...
+      handle_exception(e, folder)
       nil
     end
 
+    def handle_exception(err, folder)
```
parkr: `log_realpath_exception` ?

ashmaroli: Yeah, that'd communicate the intent clearly.. :+1:

ashmaroli: @cetinajero Will you be able to push the changes right away..?

cetinajero: For sure!

ashmaroli:
@jekyllbot: merge +bug

ashmaroli:
Thank you for submitting this patch @cetinajero

cetinajero:
Thanks you too for your time and review @ashmaroli, I'll be around.

✌️

PR #7680, opened 2019-05-24
Added documentation for Live Reload feature

bomberstudios:

## Summary

Live Reloading was was added in 3.7.0 (see https://jekyllrb.com/news/2018/01/02/jekyll-3-7-0-released/) but it's not mentioned anywhere in the docs. This PR does something about it.

### DIFF
--- a/docs/_docs/configuration/options.md
+++ b/docs/_docs/configuration/options.md
@@ -365,6 +365,7 @@ before your site is served.
         <p class="description">Reload the site automatically on the browser when content is edited.</p>
       </td>
       <td class="align-center">
+        <p><code class="option">livereload: true</code></p>
         <p><code class="flag">--livereload</code></p>
       </td>
     </tr>
--- a/docs/_docs/configuration/options.md
+++ b/docs/_docs/configuration/options.md
@@ -362,7 +362,7 @@ before your site is served.
     <tr class="setting">
       <td>
         <p class="name"><strong>Live Reload</strong></p>
-        <p class="description">Reload the site automatically on the browser when content is edited.</p>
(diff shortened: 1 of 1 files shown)


docs/_docs/configuration/options.md
```diff
@@ -352,17 +352,26 @@ before your site is served.
...
       </td>
       <td class="align-center">
         <p><code class="option">baseurl: URL</code></p>
         <p><code class="flag">--baseurl URL</code></p>
       </td>
     </tr>
+    <tr class="setting">
+      <td>
+        <p class="name"><strong>Live Reload</strong></p>
+        <p class="description">Reload the site automatically on the browser when content is edited.</p>
```
ashmaroli: A small correction here:
*The entire site is not automatically reloaded on the browser. Just the page that has been edited.*
For example, if you are currently at `http://localhost:4000/` and you edit `about.md` that resolves into `http://localhost:4000/about/`, then the *homepage* need not reload automatically..

(Note: *Do confirm above before making a change to this PR..*)

docs/_docs/configuration/options.md
```diff
@@ -352,17 +352,26 @@ before your site is served.
...
+      </td>
+      <td class="align-center">
+        <p><code class="flag">--livereload</code></p>
```
DirtyF:
```suggestion
        <p><code class="option">livereload: true</code></p>
        <p><code class="flag">--livereload</code></p>
```

DirtyF: You can also specify that option in a config file (.e.g. `_config_local.yml`)

bomberstudios: Oh, I wasn't aware of that, and it's super useful for our use case. Thanks!

PR #7684, opened 2019-05-30
Revert memoizing Site#docs_to_write and #documents

ashmaroli:
## Summary

Memoizing `Site#docs_to_write` in Jekyll 3.8 caused plugin `jekyll-redirect-from` to break.
Therefore, reverting that and associated memoization of `Site.documents` made for Jekyll 4.0

## Context

https://github.com/jekyll/jekyll-redirect-from/issues/199

### DIFF
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -326,15 +326,15 @@ module Jekyll
     #
     # Returns an Array of Documents which should be written
     def docs_to_write
-      @docs_to_write ||= documents.select(&:write?)
+      documents.select(&:write?)
     end
 
     # Get all the documents
     #
     # Returns an Array of all Documents
     def documents
-      @documents ||= collections.reduce(Set.new) do |docs, (_, collection)|
-        docs + collection.docs + collection.files
+      collections.each_with_object(Set.new) do |(_, collection), set|
+        set.merge(collection.docs).merge(collection.files)
       end.to_a
     end
 


ashmaroli:
### Profiler summary
```diff
--- master branch https://travis-ci.org/jekyll/jekyll/jobs/538411746
+++ PR branch     https://travis-ci.org/jekyll/jekyll/jobs/539098738

- Total allocated: 508.65 MB (4727576 objects)
- Total retained:  18.28 MB (95956 objects)
+ Total allocated: 677.28 MB (4835694 objects)
+ Total retained:  18.28 MB (95954 objects)
```

Based on above stats, reverting memoization has a *significant impact* on the memory usage (of building Jekyll's Documentation site)

/cc @parkr, @mattr-

lib/jekyll/site.rb
```diff
@@ -326,14 +326,14 @@ def relative_permalinks_are_deprecated
...
-      @docs_to_write ||= documents.select(&:write?)
+      documents.select(&:write?)
     end
 
     # Get all the documents
     #
     # Returns an Array of all Documents
     def documents
-      @documents ||= collections.reduce(Set.new) do |docs, (_, collection)|
+      collections.reduce(Set.new) do |docs, (_, collection)|
```
parkr: Can you retest the memory allocations for each_with_object instead of reduce? The order of the block arguments are slightly different and you’d do `docs.merge(collection.docs)`.

The idea here is that the reduce method allocations a number of object in each iteration, but each_with_object simply iterates over your existing objects and lets you build the set with each iteration.

lib/jekyll/site.rb
```diff
@@ -326,15 +326,16 @@ def relative_permalinks_are_deprecated
...
     end
 
     # Get all the documents
     #
     # Returns an Array of all Documents
     def documents
-      @documents ||= collections.reduce(Set.new) do |docs, (_, collection)|
-        docs + collection.docs + collection.files
+      collections.each_with_object(Set.new) do |(_, collection), set|
+        collection.docs.each  { |item| set << item }
```
parkr: Let's use `set.merge(collection.docs).merge(collection.files)`. [According to the documentation](https://ruby-doc.org/stdlib-2.6.3/libdoc/set/rdoc/Set.html#method-i-merge), it mutates the set. I don't see a reason to do this second iteration manually when the Ruby core team can add optimizations for us in the future 😄

ashmaroli:
### Profiler summary
```diff
--- master branch https://travis-ci.org/jekyll/jekyll/jobs/539200911
+++ PR branch     https://travis-ci.org/jekyll/jekyll/jobs/539283137

- Total allocated: 508.63 MB (4727212 objects)
- Total retained:  18.28 MB (95956 objects)
+ Total allocated: 549.56 MB (4782845 objects)
+ Total retained:  18.28 MB (95954 objects)
```

ashmaroli:
Thank you @parkr. The allocations from `#reduce` *went down considerably* with `#each_with_object`.

ashmaroli:
Squashed the two commits into one so that Parker gets attributed properly..

ashmaroli:
@jekyllbot: merge +bug

ashmaroli:
Ow man! Jekyllbot doesn't take co-authors into consideration at all....... :disappointed:

PR #7697, opened 2019-06-05
Memoize SiteDrop#documents to reduce allocations

ashmaroli:
## Summary

`Site#documents` cannot be memoized so that `Site#docs_to_write` can access the latest state of the attribute.

Since this `SiteDrop#documents` will be called after the `Site#pre_render` hook, the `Site#documents` shouldn't thereafter change and can therefore be safely memoized.

### DIFF
--- a/lib/jekyll/drops/site_drop.rb
+++ b/lib/jekyll/drops/site_drop.rb
@@ -42,7 +42,8 @@ module Jekyll
       # latest state of the attribute.
       #
       # Since this method will be called after `Site#pre_render` hook, the `Site#documents`
-      # shouldn't change and can therefore be safely memoized.
+      # array shouldn't thereafter change and can therefore be safely memoized to prevent
+      # additional computation of `Site#documents`.
       def documents
         @documents ||= @obj.documents
       end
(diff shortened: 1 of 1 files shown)


ashmaroli:
### Profiler summary
```diff
--- master branch https://travis-ci.org/jekyll/jekyll/jobs/541575706
+++ PR branch     https://travis-ci.org/jekyll/jekyll/jobs/541775998

- Total allocated: 549.57 MB (4783239 objects)
+ Total allocated: 508.77 MB (4726486 objects)
  Total retained:  18.28 MB (95954 objects)
```

mattr-:
Are there other hooks that run that would affect this if we memoize it?

ashmaroli:
> Are there other hooks that run that would affect this

The available hooks that are related to `Site#documents` are:

```ruby
:post_read, :pre_render, :post_render, :post_write
```

To my knowledge, methods defined for Liquid drops are first invoked when a Liquid template is rendered which occurs only as part of `Site#render`.
Properly implemented `generators` simply *append* objects prior to the `render` phase to be automatically rendered as part of the `Site#render` call.

ashmaroli:
IMO, this should also be backported to the `3.8-stable` branch because we recently reverted the memoization of `Site#documents`. (The next release from `3.8-stable` shouldn't be *slower* than `3.8.5`)

mattr-:
@jekyllbot: merge +minor

PR #7698, opened 2019-06-06
Improve API surrounding Jekyll::Cache

ashmaroli:
## Summary

- `Jekyll::Cache` no longer responds to **`:cache_dir=`**
  The singleton class attributes are now read-only. A `Jekyll::Site` instance can only *bootstrap* the Cache class with itself.
- Initializing a `Jekyll::Site` will automatically *bootstrap* the `Cache` singleton class with configured `cache_dir` and toggle caching to disk.
  - Caching to disk is disabled if `site.safe`
  - Caching to disk is disabled if `config["cache_dir"]` equals `""`, `"/"` or any other data-type other than a string.
- private singleton_method `delete_cache_files` have been removed to avoid confusion with public instance_method `#delete_cache_files`
- **`Jekyll::Site#in_cache_dir` will now return `nil` if `site.safe` or `cache_dir.nil?`**

### DIFF
--- a/lib/jekyll/cache.rb
+++ b/lib/jekyll/cache.rb
@@ -12,7 +12,7 @@ module Jekyll
 
     class << self
       # class-wide cache location
-      attr_accessor :cache_dir
+      attr_reader :cache_dir
 
       # class-wide directive to write cache to disk
       attr_reader :disk_cache_enabled
@@ -20,15 +20,26 @@ module Jekyll
       # class-wide base cache reader
       attr_reader :base_cache
 
+      # Setup cache attributes for given +site+
+      #
+      # site - A Jekyll::Site instance
+      #
+      # Returns nothing
+      def bootstrap(site)
+        @cache_dir = determine_cache_dir(site)
+        disable_disk_cache! if @cache_dir.nil? || site.safe
+        nil
+      end
+
       # Disable Marshaling cached items to disk
       def disable_disk_cache!
         @disk_cache_enabled = false
       end
 
-      # Clear all caches
+      # Delete all cached items from the disk cache and clear all in-memory caches
       def clear
-        delete_cache_files
-        base_cache.each_value(&:clear)
+        FileUtils.rm_rf(cache_dir) if disk_cache_enabled && cache_dir
+        base_cache.clear
       end
 
       # Compare the current config to the cached config
@@ -48,11 +59,14 @@ module Jekyll
 
       private
 
-      # Delete all cached items from all caches
-      #
-      # Returns nothing.
-      def delete_cache_files
-        FileUtils.rm_rf(@cache_dir) if disk_cache_enabled
+      def determine_cache_dir(site)
+        configured_path = site.config["cache_dir"]
+        return unless configured_path.is_a?(String)
+
+        cache_path = site.in_source_dir(configured_path)
+        return cache_path unless cache_path == site.in_source_dir("")
+
+        nil
       end
     end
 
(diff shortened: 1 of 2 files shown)


lib/jekyll/cache.rb
```diff
@@ -136,14 +150,16 @@ def key?(key)
...
     end
 
     private
 
     # Given a hashed key, return the path to where this item would be saved on disk.
     def path_to(hash = nil)
-      @base_dir ||= File.join(Jekyll::Cache.cache_dir, @name)
+      return unless disk_cache_enabled?
+
+      @base_dir ||= File.join(Jekyll::Cache.cache_dir, "Jekyll/Cache", @name)
```
mattr-: Can we use lower case here?
```suggestion
      @base_dir ||= File.join(Jekyll::Cache.cache_dir, "jekyll/cache", @name)
```

mattr-: Also, does this need to be a dot directory or anything like that?

lib/jekyll/cache.rb
```diff
@@ -12,23 +12,34 @@ class Cache
...
 
       # class-wide base cache reader
       attr_reader :base_cache
 
+      # Setup cache attributes for given +site+
+      #
+      # site - A Jekyll::Site instance
+      #
+      # Returns nothing
+      def bootstrap(site)
```
mattr-: I'd rather be more explicit about the interface here. Let's change the bootstrap signature a bit:
```suggestion
      def bootstrap(configured_cache_dir, safe_mode)
```

It also removes the dependency on the `site` object, which is really only used here and in `determine_cache_dir`

lib/jekyll/cache.rb
```diff
@@ -12,23 +12,34 @@ class Cache
...
       def disable_disk_cache!
         @disk_cache_enabled = false
       end
 
-      # Clear all caches
+      # Delete all cached items from the disk cache and clear all in-memory caches
       def clear
-        delete_cache_files
-        base_cache.each_value(&:clear)
+        FileUtils.rm_rf(cache_dir) if disk_cache_enabled && cache_dir
```
mattr-:
```suggestion
        FileUtils.rm_rf(cache_dir) if disk_cache_enabled?
```

lib/jekyll/cache.rb
ashmaroli: 1. `Jekyll/Cache` was the path that Pat Hawks originally chose:
https://github.com/jekyll/jekyll/blob/7096885e98a72a3ed8cbe4c2c666d0f490211ea9/lib/jekyll/site.rb#L472

However, I don't see why we shouldn't opt for the downcased names.. perhaps, it was to avoid collision with existing directories created by third-party plugins..? Not sure.

2. There's no need for it to be a dot directory because the default value for `Jekyll::Cache.cache_dir` is `.jekyll-cache`. If a user overrides it, the onus is on them.

lib/jekyll/cache.rb
ashmaroli: I intentionally added the dependency to `site` because we've no other way to refer to the `site` object. Though, I don't remember why I needed to reference the `site` object anyway.. :thinking:

ashmaroli: The `determine_cache_dir` is wired to check if the configured cache-dir isn't the same as the current source-dir and adapt accordingly. So, we need the `site` object here as this logic can't be moved into the `Jekyll::Site` class.

lib/jekyll/cache.rb
ashmaroli: This won't do because `:clear` is a class method and `:disk_cache_enabled?` an instance_method..

PR #7701, opened 2019-06-07
Use String#end_with? to check if entry is a backup

ashmaroli:
## Summary

`String#[]` allocates memory for the substring but `String#end_with?` doesn't (and therefore faster as well).


DirtyF:
@jekyllbot: merge +dev

PR #7702, opened 2019-06-07
Use regexp to filter special entries

ashmaroli:
## Summary

Using an anchored regular expression with `Regexp#match?` is more optimal than checking if the *first character* of given `entry` exists in an array (because):
- No need for array traversal.
- No need to allocate memory for the `entry[0..0]` substring.
- `Regexp#match?` doesn't allocate `MatchData` objects.

### DIFF
--- a/lib/jekyll/entry_filter.rb
+++ b/lib/jekyll/entry_filter.rb
@@ -7,6 +7,8 @@ module Jekyll
       ".", "_", "#", "~",
     ].freeze
 
+    SPECIAL_LEADING_CHAR_REGEX = %r!\A#{Regexp.union(SPECIAL_LEADING_CHARACTERS)}!o.freeze
+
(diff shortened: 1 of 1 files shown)


lib/jekyll/entry_filter.rb
```diff
@@ -7,6 +7,8 @@ class EntryFilter
       ".", "_", "#", "~",
     ].freeze
 
+    SPECIAL_LEADING_CHAR_REGEX = %r!\A#{Regexp.union(SPECIAL_LEADING_CHARACTERS)}!o.freeze
```
mattr-: This is the only place we use `SPECIAL_LEADING_CHARACTERS` in this class, so we can just remove it.

mattr-:
@jekyllbot: merge +dev

PR #7705, opened 2019-06-09
Reject entries that are just dots in a directory

ashmaroli:
- This is a 🐛 bug fix.

## Summary
Globbing a directory can result in the inclusion of `.` and `..` in the array returned if dots are allowed to be matched, e.g. in https://github.com/jekyll/jekyll/blob/13d31c4c8be13ab1536a0139a48952b142e81682/lib/jekyll/collection.rb#L80

Collection directory or not, such entries serve no purpose in a Jekyll Build and should be filtered away at all times. Moreover, doing so allows us to avoid wasting memory and time on checking if such entries qualify to be passed through for rendering.

### DIFF
--- a/lib/jekyll/entry_filter.rb
+++ b/lib/jekyll/entry_filter.rb
@@ -31,6 +31,9 @@ module Jekyll
 
     def filter(entries)
       entries.reject do |e|
+        # Reject this entry if it is just a "dot" representation.
+        #   e.g.: '.', '..', '_movies/.', 'music/..', etc
+        next true if e.end_with?(".")
         # Reject this entry if it is a symlink.
         next true if symlink?(e)
         # Do not reject this entry if it is included.


ashmaroli:
We save a *teeny bit* of memory allocations from this change:

### Profiler summary
```diff
--- master branch https://travis-ci.org/jekyll/jekyll/jobs/541828395
+++ PR branch     https://travis-ci.org/jekyll/jekyll/jobs/543313989

- Total allocated: 548.75 MB (4762592 objects)
+ Total allocated: 548.71 MB (4761952 objects)
  Total retained:  18.28 MB (95954 objects)
```

mattr-:
@jekyllbot: merge +fix

ashmaroli:
@mattr- I think Jekyllbot doesn't respond to `+fix`..

@jekyllbot: merge +bug

PR #7708, opened 2019-06-12
fix link to Site Source config

jamesrhea:
Link HTML was missing an end quote on the `href` attribute, and so was showing as plain text.


DirtyF:
@jekyllbot: merge +docs

PR #7709, opened 2019-06-13
Update TestTags in sync with Rouge v3.4

ashmaroli:
## Summary

Rouge 3.4 released [recently](https://rubygems.org/gems/rouge/versions/3.4.1) and it is causing our CI builds to fail.
It includes a *minor change* in [the output from HTMLTable formatter](https://github.com/rouge-ruby/rouge/commit/a050f9e6155326f11b6ca73fa3a66bc16fafde65). According to the author of the change:
> The rationale is that it's important for the end of the lines stay clean to allow postprocessing to operate on all lines in the same way. When the closing `</pre>` bumps up against the last line, it requires special processing.


DirtyF:
@jekyllbot: merge +dev

PR #7710, opened 2019-06-14
Reduce EntryFilter#relative_to_source allocations

ashmaroli:
## Summary

By avoiding generation of unnecessary strings for files at the root to be excluded.
When the `base_dir` equals the `source_dir`, the result from `#relative_to_source` will just produce the given `entry` with a leading slash:
```ruby
# When `base_dir` == `source_dir` or "#{source_dir}/"
# base_dir[site.source] = "" produces => "" or "/" respectively
#   both of which ultimately yield the same string:
File.join("", "index.md")   # => "/index.md"
File.join("/", "about.md")  # => "/about.md"
```

### DIFF
--- a/lib/jekyll/entry_filter.rb
+++ b/lib/jekyll/entry_filter.rb
@@ -20,13 +20,12 @@ module Jekyll
 
     def derive_base_directory(site, base_dir)
       base_dir[site.source] = "" if base_dir.start_with?(site.source)
+      base_dir.gsub!(%r!\A#{File::SEPARATOR}!o, "")
       base_dir
     end
 
     def relative_to_source(entry)
-      File.join(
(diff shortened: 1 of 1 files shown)


ashmaroli:
Hmm.. Not much of a gain though.. :frowning_face:

### Profiler summary
```diff
--- master branch https://travis-ci.org/jekyll/jekyll/jobs/545674924
+++ PR branch     https://travis-ci.org/jekyll/jekyll/jobs/545736397

- Total allocated: 532.7 MB (4608936 objects)
- Total retained:  18.36 MB (95384 objects)
+ Total allocated: 532.7 MB (4608745 objects)
+ Total retained:  18.36 MB (95387 objects)
```

PR #7723, opened 2019-06-26
Replace `String#=~` with `String#match?`

ashmaroli:
## Summary

Prefer the faster `String#match?` over `String#=~` when a `MatchData` is not relevant.

The `:has_yaml_header?` utility method just reads the first line of the given `file` and checks against a regex without capture groups. So even if someone was using the Matchdata generated by this method, it is not a proper usage.

Therefore, this change can be considered to be a non-breaking change.

### DIFF
--- a/lib/jekyll/utils.rb
+++ b/lib/jekyll/utils.rb
@@ -135,7 +135,7 @@ module Jekyll
     # Returns true if the YAML front matter is present.
     # rubocop: disable PredicateName
     def has_yaml_header?(file)
-      !!(File.open(file, "rb", &:readline) =~ %r!\A---\s*\r?\n!)
+      File.open(file, "rb", &:readline).match? %r!\A---\s*\r?\n!
     rescue EOFError
       false
     end


parkr:
Nice simplification! Do you have a profile of before/after to help drive the performance claim? I agree that not creating MatchData should be better but just curious _how_ much better. 😄

ashmaroli:
> to help drive the performance claim

@parkr To be honest, this particular change can't be seen as a major optimization because the gain from not allocating a `MatchData` object is overwhelmed by the allocation of the string result of
`File.open(file, "rb", &:readline)`.
However, the `profile:memory` rake task we have yields the following:
```diff
--- master branch https://travis-ci.org/jekyll/jekyll/jobs/550642211
+++ PR branch     https://travis-ci.org/jekyll/jekyll/jobs/550687954

- Total allocated: 491.46 MB (4553703 objects)
+ Total allocated: 491.42 MB (4553354 objects)
  Total retained:  18.43 MB (95814 objects)
```

DirtyF:
@ashmaroli that memory footprint keeps getting down 👍  📉

ashmaroli:
@jekyllbot: merge +bug

PR #7725, opened 2019-06-27
Cache interim paths in EntryFilter#glob_include?

ashmaroli:
## Summary

For every `EntryFilter` object in a given `site`, `#glob_include?` generates intermediate path strings for each entry in a traversed subdirectory: `entry_with_source = File.join(site.source, entry)` and for every entry within arrays `site.include` and `site.exclude`: `EntryFilter.prefix_source`, both of which remains unchanged for a given build session.

### DIFF
--- a/lib/jekyll/entry_filter.rb
+++ b/lib/jekyll/entry_filter.rb
@@ -5,10 +5,10 @@ module Jekyll
     attr_reader :site
     SPECIAL_LEADING_CHAR_REGEX = %r!\A#{Regexp.union([".", "_", "#", "~"])}!o.freeze
 
-    def self.prefix_source(src, item)
-      @prefix_source ||= {}
-      @prefix_source[src] ||= {}
-      @prefix_source[src][item] ||= File.join(src, item)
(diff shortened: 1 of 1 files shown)


DirtyF:
I'm curious about the gain here. Any memory benchmark to share?

ashmaroli:
Surely.

### Profiler Summary

```diff
--- master branch https://travis-ci.org/jekyll/jekyll/jobs/550642211
+++ PR branch     https://travis-ci.org/jekyll/jekyll/jobs/551346938

- Total allocated: 491.46 MB (4553703 objects)
- Total retained:  18.43 MB (95814 objects)
+ Total allocated: 490.63 MB (4545529 objects)
+ Total retained:  18.62 MB (97457 objects)
```

ashmaroli:
The gain isn't *much* for our `docs` site. But for an extensively nested site like `https://github.com/tomjoht/tomjoht.github.io` will see a significant relief from memory usage.
(But, I don't have public results for that site though)

PR #7728, opened 2019-06-28
Memoize fallback_data for Drop

ashmaroli:
To avoid generating a new Hash object on every call

### DIFF
--- a/lib/jekyll/drops/url_drop.rb
+++ b/lib/jekyll/drops/url_drop.rb
@@ -125,7 +125,7 @@ module Jekyll
       private
 
       def fallback_data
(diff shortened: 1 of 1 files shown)


DirtyF:
Any possible side effect to memoization here?

ashmaroli:
No. It is safe. Another drop with a memoized fallback:
https://github.com/jekyll/jekyll/blob/2265e82181a1d96da9910cad233b479fda2d5967/lib/jekyll/drops/unified_payload_drop.rb#L19-L23

DirtyF:
@jekyllbot: merge +dev

PR #7729, opened 2019-06-28
Cache sanitized paths to reduce String allocations

ashmaroli:
- This is an :zap: optimization change.

## Summary

`Jekyll.sanitized_path` is one of the most frequently called methods in the library. Additionally, it is also wrapped to provide `Site#in_source_dir` and similar helper methods.

However, the algorithm involves generation of multiple intermediate strings.
Memoizing this method should reduce String allocations from a build session considerably.

PR #7732, opened 2019-06-30
Add PathManager class to cache interim paths

ashmaroli:

## Summary

`File.join` allocates an Array and returns a new String on every call using **the same arguments**. Caching the result means reduced memory usage. However, the caches are never flushed so that they can be used even when a site is regenerating.
The results are frozen to deter mutation of the cached string.

Additionally caches result from logic formerly exposed as part of `Jekyll.sanitized_path`

## Context

Clubs the changes proposed in #7725 and #7729 and extracts the logic into a new singleton class.
Closes #7725 
Closes #7729

### DIFF
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -162,14 +162,14 @@ module Jekyll
     end
 
     # Public: Ensures the questionable path is prefixed with the base directory
-    #         and prepends the questionable path with the base directory if false.
+    #         and that the resulting path is a descendant of the base directory.
     #
     # base_directory - the directory with which to prefix the questionable path
     # questionable_path - the path we're unsure about, and want prefixed
     #
-    # Returns the sanitized path.
+    # Returns a frozen sanitized path or the frozen instance of `base_directory`.
     def sanitized_path(base_directory, questionable_path)
-      return base_directory if base_directory.eql?(questionable_path)
+      return base_directory.freeze if base_directory.eql?(questionable_path)
 
       PathManager.sanitized_join(base_directory, questionable_path)
     end
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -65,6 +65,7 @@ module Jekyll
   autoload :LogAdapter,          "jekyll/log_adapter"
   autoload :Page,                "jekyll/page"
   autoload :PageWithoutAFile,    "jekyll/page_without_a_file"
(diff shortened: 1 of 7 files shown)


ashmaroli:
### Profiler summary

```diff
--- master branch https://travis-ci.org/jekyll/jekyll/jobs/551856715
+++ PR branch     https://travis-ci.org/jekyll/jekyll/jobs/552392586

- Total allocated: 482.41 MB (4533274 objects)
- Total retained:  18.43 MB (95810 objects)
+ Total allocated: 478.66 MB (4497543 objects)
+ Total retained:  18.75 MB (98289 objects)
```

DirtyF:
```

**Comment by ashmaroli** [comment]
> This triggers an error on jekyll serve

Shoot! That's unfortunate. That means we can't cache results of `Jekyll.sanitized_path`
Thanks for testing this @DirtyF

**Comment by ashmaroli** [comment]
### Profiler summary

```diff
--- master branch https://travis-ci.org/jekyll/jekyll/jobs/551856715
+++ PR branch     https://travis-ci.org/jekyll/jekyll/jobs/552438024

- Total allocated: 482.41 MB (4533274 objects)
- Total retained:  18.43 MB (95810 objects)
+ Total allocated: 481.03 MB (4518452 objects)
+ Total retained:  18.58 MB (96930 objects)
```

DirtyF:
It works now, on a side note, jekyll website build in 3.7s with a warm cache on my macbook air ⚡️

ashmaroli:
> jekyll website build in 3.7s

Is that with the updated branch..?

DirtyF:
with this branch yes.

ashmaroli:
Nice. Thanks for the info :+1:

ashmaroli:
I'm expecting to see more reduction in memory reports here once kramdown ships a new release.. So good prospects overall.

mattr-:
> It works now, on a side note, jekyll website builds in 3.7s with a warm cache on my macbook air ⚡️

What were the numbers before?

ashmaroli:
> liked it even better if we could have overridden `File.join`.. but I also understand that isn't feasible.

Its better we don't override Core Ruby classes so that we can always fall back to those classes where our cached behavior isn't desirable. :slightly_smiling_face:

mattr-:
@jekyllbot: merge +minor

PR #7746, opened 2019-07-10
Update showcase.yml

nuket:
This is a small static site I put together using Jekyll, jekyll-asciidoc, Tailwind CSS, PostCSS, PurgeCSS, and cssnano.

## Summary

Add an new example to the showcase that isn't a corporate landing page.

## Context

Not related to any issue, only related to having fun.

### DIFF
--- a/docs/_data/showcase.yml
+++ b/docs/_data/showcase.yml
@@ -239,3 +239,8 @@
     - marketing-site
     - documentation
     - software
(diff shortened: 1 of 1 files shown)


DirtyF:
Thanks but we're not adding new showcases entries at the moment. Feel free to share your site on talk.jekyllrb.com

PR #7758, opened 2019-07-22
Reduce allocations by using #each_with_object

ashmaroli:
## Summary

The loop within `config.reduce({}) { |hsh, (k, v)| hsh.merge(k.to_s => v) }` is going to run for each
key-value pair in `config` and on each iteration, `hsh.merge` is going to allocate a new Hash object.

Therefore, if a config file has 20 **top-level key-value pairs**, this method would've allocated at least 20 Hash objects (of increasing size)..
This can be bad for sites that have numerous top-level config keys (e.g. forks of [`minimal-mistakes`](https://github.com/mmistakes/minimal-mistakes/blob/master/_config.yml) theme repo)

### DIFF
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -98,7 +98,7 @@ module Jekyll
     #
     # Return a copy of the hash where all its keys are strings
     def stringify_keys
-      reduce({}) { |hsh, (k, v)| hsh.merge(k.to_s => v) }
+      each_with_object({}) { |(k, v), hsh| hsh[k.to_s] = v }
     end
 
     def get_config_value_with_override(config_key, override)
@@ -254,7 +254,9 @@ module Jekyll
 
       # Ensure we have a hash.
       if config["collections"].is_a?(Array)
-        config["collections"] = Hash[config["collections"].map { |c| [c, {}] }]
+        config["collections"] = config["collections"].each_with_object({}) do |collection, hash|
+          hash[collection] = {}
+        end
       end
 
       config["collections"] = Utils.deep_merge_hashes(
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -4,7 +4,7 @@ module Jekyll
   class Configuration < Hash
     # Default options. Overridden by values in _config.yml.
     # Strings rather than symbols are used for compatibility with YAML.
-    DEFAULTS = Configuration[{
+    DEFAULTS = {
       # Where things are
       "source"              => Dir.pwd,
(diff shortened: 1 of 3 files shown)


lib/jekyll/configuration.rb
```diff
@@ -98,7 +98,7 @@ def from(user_config)
     #
     # Return a copy of the hash where all its keys are strings
     def stringify_keys
-      reduce({}) { |hsh, (k, v)| hsh.merge(k.to_s => v) }
+      reduce({}) { |hsh, (k, v)| hsh.merge!(k.to_s => v) }
```
parkr: I'm going to suggest my good ol' friend `each_with_object` here instead:

```ruby
>> conf = {foo: "bar", "baz" => "boz"}
=> {:foo=>"bar", "baz"=>"boz"}
>> conf.each_with_object({}) { |(k,v), hash| hash[k.to_s] = v }
=> {"foo"=>"bar", "baz"=>"boz"}
```

ashmaroli: Is there a difference in the performance?

mattr-: Even if there is, `each_with_object` is easier to understand.

ashmaroli: I checked.. and I can confirm that there is a performance issue with `reduce.merge!` as well..
For future readers, the reason is because `k.to_s => v` in `merge!(k.to_s => v)` is a new `Hash` in each iteration..!!

lib/jekyll/configuration.rb
```diff
@@ -4,7 +4,7 @@ module Jekyll
   class Configuration < Hash
     # Default options. Overridden by values in _config.yml.
     # Strings rather than symbols are used for compatibility with YAML.
-    DEFAULTS = Configuration[{
```
parkr: We might want to keep this class declaration

ashmaroli:
@parkr Sorry, I found one more instance of `Hash[array.map { |k,v| [k,v] }]` after you approved the PR.

ashmaroli:
The gains aren't much, but probably *non-trivial* for much larger sites:

### Profiler Summary
```diff
--- master branch https://travis-ci.org/jekyll/jekyll/jobs/561428064
+++ PR branch     https://travis-ci.org/jekyll/jekyll/jobs/562630268

- Total allocated: 479.11 MB (4483645 objects)
- Total retained:  18.54 MB (96252 objects)
+ Total allocated: 479.06 MB (4482911 objects)
+ Total retained:  18.54 MB (96251 objects)
```

DirtyF:
@jekyllbot: merge +dev

PR #7761, opened 2019-07-24
Replace redundant Array#map with Array#each

ashmaroli:
## Summary

- The result of `Array#map` isn't being saved here. Therefore the generated array is just wasteful.
- Corrects the code-comment to better reflect the implementation.

### DIFF
--- a/lib/jekyll/readers/page_reader.rb
+++ b/lib/jekyll/readers/page_reader.rb
@@ -9,14 +9,13 @@ module Jekyll
       @unfiltered_content = []
     end
 
-    # Read all the files in <source>/<dir>/ for Yaml header and create a new Page
-    # object for each file.
+    # Create a new `Jekyll::Page` object for each entry in a given array.
     #
-    # dir - The String relative path of the directory to read.
+    # files - An array of file names inside `@dir`
     #
-    # Returns an array of static pages.
+    # Returns an array of publishable `Jekyll::Page` objects.
     def read(files)
-      files.map do |page|
+      files.each do |page|
         @unfiltered_content << Page.new(@site, @site.source, @dir, page)
       end
       @unfiltered_content.select { |page| site.publisher.publish?(page) }


DirtyF:
What bug does that fix?

ashmaroli:
> What bug does that fix?

Generating a copy of an array unnecessarily is not a bug?
Then this is just an *`internal`* change..

DirtyF:
What's the impact? When is this array used?

ashmaroli:
> When is this array used?

Remember `site.pages` usually found in Liquid templates? Well, `site.pages` is the result of concatenation of arrays returned by *the method* patched in this PR.
Every time Jekyll encounters a *non-special* directory, this method is invoked and an array is returned.


> What's the impact?

Depends on a given site's structure, if it has numerous standalone pages at root, or numerous standalone pages nested in directories or just posts or just documents from other collections, or if its a mix of all these.
Of all the scenarios, this change would only (positively) affect sites with numerous standalone pages nested in numerous directories.

I didn't post the usual `### Profile summary` stats because Rouge shipped a new version (`v3.7.0`) yesterday with new lexers (which results in increased memory and object allocations). Therefore, the difference between `master` (logged with `rouge-3.6.0`) and this PR (logged with `rouge-3.7.0`) is going to be incorrect.

*(If I were to restart the build for `master` to log the build with `rouge-3.7.0`, the differences I posted in other unmerged PRs are going end up being incorrect).*

DirtyF:
I still don't get when `site.pages` wasn't returning the complete list of pages. 🤷‍♂

ashmaroli:
The gains with this are still dubious :man_shrugging: 

### Profiler Summary

```diff
--- master branch https://travis-ci.org/jekyll/jekyll/jobs/563073466
+++ PR branch     https://travis-ci.org/jekyll/jekyll/jobs/562993536

- Total allocated: 481.24 MB (4508234 objects)
+ Total allocated: 481.24 MB (4508266 objects)
  Total retained:  18.80 MB (98498 objects)
```

DirtyF:
@jekyllbot: merge +dev

PR #7767, opened 2019-08-03
Implement item versions of certain Array filters

### LINKED ISSUE #7730: pop and shift invalid return
Jekyll pop and shift return remaining items instead of popped or shifted items. This is the exact opposite of Ruby proper:

> Removes the last element from self and returns it

https://ruby-doc.org/core/Array.html#method-i-pop

> Removes the first element of self and returns it

https://ruby-doc.org/core/Array.html#method-i-shift

and, frankly, every other programming language I have ever worked with.

mattr-:
This is a 🐛 bug fix. 

# Summary

This provides `pop_item` and `shift_item` filters, [as described](https://github.com/jekyll/jekyll/issues/7730#issuecomment-517927806)

There's several items left to do on this. I'm posting the draft PR in an attempt to get some collaboration going since I ran out of time to go any further at the moment.

 - [x] Implement `pop_item`
 - [ ] Implement `shift_item`
 - [ ] Deprecate the _behavior_ of `pop`
 - [ ] Deprecate the _behavior_ of `shift`
 - [ ] Fix the docs for `pop` and `shift`
 - [ ] Add new docs for `pop_item` and `shift_item`

Fixes #7730

### DIFF
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -257,6 +257,17 @@ module Jekyll
       new_ary
     end
 
+    def pop_item(array, num = nil)
+      return array unless array.is_a?(Array)
+
+      unless num.nil?
+        num = Liquid::Utils.to_integer(num)
+        array.pop(num)
+      else
+        array.pop
+      end
+    end
+
     def push(array, input)
       return array unless array.is_a?(Array)
 
--- a/test/test_filters.rb
+++ b/test/test_filters.rb
@@ -1256,6 +1256,26 @@ class TestFilters < JekyllUnitTest
       end
     end
 
+    context "pop_item filter" do
+      should "return the last item in the array by default" do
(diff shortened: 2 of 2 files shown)


ashmaroli:
@mattr- Since there is no certainty on when `v5.0` would even start development, let's not deprecate any of the filters right away. Instead, let us simply:
- Add a new set of filters that *mutate* given object which act as proxies to Ruby Array methods. (Would `pop!`, `shift!`, `unshift!` and `push!` be technically allowed as valid `Liquid::Filters`?)
- Document the expected behavior *clearly*.

mattr-:
>Since there is no certainty on when v5.0 would even start development, let's not deprecate any of the filters right away

Yes, let's deprecate them. It's a signal to the user. We're communicating to them that `pop_item` and `shift_item` will go away in 5.0. If we already know that, why would we hide it?

mattr-:
Not planning on finishing this.

PR #7769, opened 2019-08-03
Fix: rubocop offenses

DirtyF:
[Fix current CI errors](https://travis-ci.org/jekyll/jekyll/jobs/567204636#L465)

### DIFF
--- a/lib/jekyll/excerpt.rb
+++ b/lib/jekyll/excerpt.rb
@@ -155,7 +155,7 @@ module Jekyll
       tag_names.flatten!
       tag_names.reverse_each do |tag_name|
         next unless liquid_block?(tag_name)
-        next if head =~ endtag_regex_stash(tag_name)
+        next if endtag_regex_stash(tag_name).match?(head)
 
         modified = true
         head << "\n{% end#{tag_name} %}"
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -368,6 +368,7 @@ module Jekyll
       end
     end
 
+    # rubocop:disable Performance/RegexpMatch
     # return numeric values as numbers for proper sorting
     def parse_sort_input(property)
       number_like = %r!\A\s*-?(?:\d+\.?\d*|\.\d+)\s*\Z!
@@ -375,6 +376,7 @@ module Jekyll
 
       property
     end
+    # rubocop:enable Performance/RegexpMatch
 
     def as_liquid(item)
       case item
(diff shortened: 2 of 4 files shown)


mattr-:
Thank you!

DirtyF:
@mattr- I didn't expect this to break the current tests :/

lib/jekyll/excerpt.rb
```diff
@@ -155,7 +155,7 @@ def sanctify_liquid_tags(head)
       tag_names.flatten!
       tag_names.reverse_each do |tag_name|
         next unless liquid_block?(tag_name)
-        next if head =~ endtag_regex_stash(tag_name)
+        next if head&.match?(endtag_regex_stash(tag_name))
```
ashmaroli: Since `head` will only ever be `nil` or a String **and** that `endtag_regex_stash(tag_name)` will always be a regex, we can ditch the defensive coding and use `Regex#match?` instead:
```suggestion
        next if endtag_regex_stash(tag_name).match?(head)
```

lib/jekyll/filters.rb
```diff
@@ -371,7 +371,7 @@ def item_property(item, property)
     # return numeric values as numbers for proper sorting
     def parse_sort_input(property)
       number_like = %r!\A\s*-?(?:\d+\.?\d*|\.\d+)\s*\Z!
-      return property.to_f if property =~ number_like
+      return property.to_f if property&.match?(number_like)
```
ashmaroli: `property` here can be any Ruby primitive: Array / Hash / String / Integer.., etc
Therefore, we really can't replace `=~` (which is actually defined for [every object](https://ruby-doc.org/core-2.4.0/Object.html#method-i-3D~)) with a method (`match?`) defined only for String and Regexp.

The best solution here would be to disable RuboCop for just this method.

lib/jekyll/readers/post_reader.rb
```diff
@@ -50,7 +50,7 @@ def read_publishable(dir, magic_dir, matcher)
     # Returns klass type of content files
     def read_content(dir, magic_dir, matcher)
       @site.reader.get_entries(dir, magic_dir).map do |entry|
-        next unless entry =~ matcher
+        next unless entry&.match?(matcher)
```
ashmaroli: Prefer `Regexp#match?` as it is `nil`-compliant. `Regexp#match?(nil)` will not return an error.
```suggestion
        next unless matcher.match?(entry)
```

ashmaroli:
Thanks @DirtyF 

@jekyllbot: merge +dev

PR #7290, opened 2018-10-01
Remove configuration of theme sass files from Core

ashmaroli:
- I read the contributing document at https://jekyllrb.com/docs/contributing/
- This is a 🐛 bug fix.

## Summary

Configuring Sass from theme-gem should be handled by the bundled `jekyll-sass-converter` plugin

## Context

So that when the converter plugin chooses to drop support for `sass-ruby`, Jekyll wouldn't care..

## Requirements

- [x] Move the "configuring function" to `jekyll-sass-converter` plugin (https://github.com/jekyll/jekyll-sass-converter/pull/80)
- [x] `jekyll-sass-converter` ships a new version containing the "function"
- [x] Travis CI and AppVeyor Passes

PR #7772, opened 2019-08-04
Bump sass-converter to 2.0.0.pre.beta

DirtyF:
Needed for #7290 

Tests will need updating given now we output sourcemap files by default.


ashmaroli:
@DirtyF  Based on https://github.com/jekyll/jekyll-sass-converter/pull/86 CI logs, Bundler is able to install the pre-release automatically. So, ***if the CI passes completely here***, `sass-converter-2.0` can be shipped..

ashmaroli:
Moved all additional changes made here to the branch for #7290 itself since that would allow further experimentation..

PR #7777, opened 2019-08-14
Swapped "env ruby" out with "gemfile.lock"

RAXNATION23:
`And` hopes it runns smoothly

### DIFF
--- a/script/console
+++ b/script/console
@@ -1,4 +1,4 @@
-#!/usr/bin/env ruby
+#!/usr/bin/gemfile.lock
 
 require 'pry'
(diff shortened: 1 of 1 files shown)


ashmaroli:
You'll need to provide concrete documentation / proof to justify this change.
What exactly is the rationale behind this..?

mattr-:
Hello! 👋

Unfortunately, I don't think this change is going to have the effect you're intending and it's not something that we can accept into the main Jekyll repo. If you're having issues getting Jekyll to work with your site, please head over to https://talk.jekyllrb.com to get help from the wider community there.

Thanks!

PR #7779, opened 2019-08-14
Test with JRuby 9.2.9.0

ashmaroli:
https://www.jruby.org/2019/08/12/jruby-9-2-8-0.html
https://www.jruby.org/2019/10/30/jruby-9-2-9-0.html


DirtyF:
@jekyllbot: merge +dev

PR #7780, opened 2019-08-15
`jekyll-commonmark` as default Markdown converter

ashmaroli:

## Summary

There were talks to have `jekyll-commonmark` (using `commonmarker` written in C) as the default Markdown converter in Jekyll 4.0 to improve build times and memory usage.

It never materialized in order to avoid creating issues with *installing native extensions* on various platforms.

However, Jekyll 4.0 has already moved to include C++-based `libsass` (via `sassc-ruby` and `jekyll-sass-converter 2.0`) in the bundle.

## Context

Requires https://github.com/jekyll/jekyll-commonmark/pull/34

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,9 @@ gemspec :name => "jekyll"
 # refinements introduced in i18n-1.3.0
 gem "i18n", "~> 1.2.0" if RUBY_ENGINE == "jruby"
 
+# Temporarily point to a WIP ref in the "jekyll/jekyll-commonmark" repository.
+gem "jekyll-commonmark", :github => "jekyll/jekyll-commonmark", :ref => "refs/pull/34/head"
+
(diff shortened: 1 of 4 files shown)


ashmaroli:
Issues:
- `commonmarker` can't be installed on JRuby (platform `java`)
- Differences in rendered output.

DirtyF:
@parkr also mentioned merging commonmark and commonmark-ghpages at some point to enable GFM extensions (table, footnotes).

mattr-:
This isn't something I want to include this late in the 4.0 release cycle and I'm not super confident about basically dropping JRuby support either. I like the idea though and would like to save it for later.

ashmaroli:
@mattr- I agree with you totally. The side-effects are too significant to ignore.
I didn't upgrade this from a "draft PR" status because I myself wasn't ready to have this shipped.

PR #7783, opened 2019-08-17
Update Ruby version used in Travis-CI example

Muffo:

## Summary

Travis environment comes with Ruby 2.4.5 pre-installed.
Using this version saves about 20s of build time since 2.4.1 no longer needs to be downloaded and installed every time.

I guess many people (including me) generally do a copy paste of the scripts, so it makes sense to optimize them.


DirtyF:
@jekyllbot: merge +doc

PR #7789, opened 2019-08-18
Test Ruby 2.7.0-preview1 on Travis

ashmaroli:
## Summary

Test with unreleased Ruby versions to detect any deprecation and incompatibilities.


ashmaroli:
No longer necessary..

PR #7792, opened 2019-08-19
Actually conditionally include liquid-c

benesch:
The gem is named `liquid-c`, but the correct require is `liquid/c`. I'm
fairly certain the conditional require has been wrong since it was
introduced five years ago, and it went unnoticed because there is of
course no error message when the gem is unavailable.

### DIFF
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -190,7 +190,7 @@ module Jekyll
     end
 
     # Conditional optimizations
-    Jekyll::External.require_if_present("liquid-c")
+    Jekyll::External.require_if_present("liquid/c")
   end
 end
 


ashmaroli:
Awesome catch @benesch!!

All this while, I was wondering why the build time for the official docs site didn't improve as expected with `liquid-c`.. Now I know..!!
Thank you very much!

ashmaroli:
@mattr- Can we squeeze this into our 4.0 milestone?

DirtyF:
@jekyllbot: merge +fix 😅

ashmaroli:
@DirtyF Will you be able to update the release post for Jekyll 4.0 with this contributor's name or should I..?

benesch:
Woo, thanks for the quick merge! 🎉 

> All this while, I was wondering why the build time for the official docs site didn't improve as expected with liquid-c.. Now I know..!!

The speedup is still pretty minor, sadly, as the C extension doesn't help much with rendering, just parsing. I'm poking at that too, though!

PR #7793, opened 2019-08-19
Memoize absolute_url and relative_url filters

benesch:
These two filters are quite slow, as Addressable::URI.parse is quite
slow. Fortunately, they are pure functions and so can be easily
memoized. This one change shaves 8s off the CockroachDB docs build
(github.com/cockroachdb/docs).

### DIFF
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -9,17 +9,8 @@ module Jekyll
       #
       # Returns the absolute URL as a String.
       def absolute_url(input)
-        return if input.nil?
-
-        input = input.url if input.respond_to?(:url)
-        return input if Addressable::URI.parse(input.to_s).absolute?
-
-        site = @context.registers[:site]
-        return relative_url(input) if site.config["url"].nil?
-
-        Addressable::URI.parse(
-          site.config["url"].to_s + relative_url(input)
-        ).normalize.to_s
+        cache = (@context.registers[:cached_absolute_urls] ||= {})
+        cache[input] ||= compute_absolute_url(input)
       end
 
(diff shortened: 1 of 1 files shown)


lib/jekyll/filters/url_filters.rb
```diff
@@ -10,16 +10,20 @@ module URLFilters
...
-        input = input.url if input.respond_to?(:url)
-        return input if Addressable::URI.parse(input.to_s).absolute?
-
-        site = @context.registers[:site]
-        return relative_url(input) if site.config["url"].nil?
-
-        Addressable::URI.parse(
-          site.config["url"].to_s + relative_url(input)
-        ).normalize.to_s
+        cache = (@context.registers[:cached_absolute_urls] = {})
```
ashmaroli: Wouldn't this line cause `@context.registers[:cached_absolute_urls]` to be reset to `{}` on every call to the filter..?
Or am I reading it wrong?

dstiers-1109:
Thanks

lib/jekyll/filters/url_filters.rb
benesch: Gah, good eyes! It's the `relative_url` caching that makes all the difference for the Cockroach docs, since `relative_url` is used a fair bit in the base layout. I added caching to `absolute_url` for good measure (presumably there's some other Jekyll site out there that's using absolute URLs instead of relative URLs everywhere), and just botched the job. Thanks! Should be fixed now.

benesch:
Whoops, done. Had to refactor a bit to get the AbcSize lint to quiet down.

DirtyF:
Thanks @benesch 👍 

@jekyllbot: merge +fix

PR #7805, opened 2019-08-29
Add dropped 'title: Staff' to the code

iraszl:
'title: Staff' was part of the page previously, but then dropped. Without it the meta title is incorrect.

## Summary

A small fix to the code shown. The title part was dropped, so I re-added it. Otherwise it may confuse users learning Jekyll.


ashmaroli:
@jekyllbot: merge +docs

PR #7813, opened 2019-09-05
Make number_of_words respect CJK characters

iBug:

## Summary

CJK languages generally consider every single character a word.

This PR modifies Liquid filter "number_of_words" so it counts CJK characters correctly.

## Context

My personal blog uses Minimal Mistakes theme, which relies on this filter to generate a "reading time". However, as I write both English and Chinese, the latter of which doesn't use spaces to separate words, my Chinese articles get a "1 minute read" as a result regardless of how much I write. This is wrong.

I am currently using [a hand-crafted plugin](https://github.com/iBug/iBug-source/commit/3fcb40a6c12f77720ea4edd5fc82911c788ed405) to resolve this issue, but I hope that it can be solved on the Jekyll side. The changes is mostly a port of my own plugin.


mattr-:
@iBug Thanks for the contribution!

It looks like there's one more change to make for rubocop, and then this should be good to merge

lib/jekyll/filters.rb
```diff
@@ -122,7 +122,8 @@ def normalize_whitespace(input)
     #
     # Returns the Integer word count.
     def number_of_words(input)
-      input.split.length
+      cjk_regex = %r(\p{Han}|\p{Katakana}|\p{Hiragana}|\p{Hangul})
```
mattr-:
```suggestion
      cjk_regex = %r!\p{Han}|\p{Katakana}|\p{Hiragana}|\p{Hangul}!
```

This should make rubocop happy.

iBug:
@mattr- Thank you for your guidance. I've reviewed RuboCop output and applied your suggested changes. I'm rather new to Ruby actually.

By the way, I've updated the documentation text as well. Hope it helps.

iBug:
Sorry for the hibernation of this PR. I spent a lot of time learning Ruby development and Jekyll. With the last commit 360ecdabf82e61e3dbaa77316598ade412316ffb, `bundle exec rake test` gives no errors. Is that enough?

ashmaroli:
@iBug Thank you for returning to this. The test looks sufficient to me.

doersino:
I'm not sure whether including Katakana, Hiragana and Hangul (which all encode *syllables*, not words) in the regex is an *ideal* solution, but I do agree that it's better than the status quo.

Also note that while Japanese (Katakana, Hiragana) doesn't make use of ASCII spaces in writing, Korean (Hangul) does.

iBug:
@doersino I'm aware of that, but in general CJK languages don't count "words", but "characters" instead (think how Microsoft Word does it). For ASCII spaces, since CJK chars are normalized into spaces before western (Latin, Cyrillic etc.) "words" are split, they don't matter here.

iBug:
@ashmaroli Any chance someone could take a look at this? I think this is whatsoever better than before and thus worth having.

ashmaroli:
Yes, I believe this is better than current `master`.
Pinging @mattr- for input.

ashmaroli:
Looking at this again...
Say, I were to have `{{ page.content | number_of_words }}` in my layout, with the `page` containing only latin characters.
Based on the current proposed implementation:
```
input.scan(cjk_regex).length + input.gsub(cjk_regex, " ").split.length
```
`input.scan(cjk_regex).length` would allocate an empty array.
`input.gsub(cjk_regex, " ")` would duplicate `input` or `page.content` in our sample case (`String#gsub` duplicates given string irrespective of substitution) and then
`input.gsub(cjk_regex, " ").split` will generate an array of all the words in `page`.

In other words, a situation that was already bloated with `master` version (`page.content` is usually a large string) is going to bloat further with no apparent gain for the user.

Since, the percentage of users that would have CJK chars in their site's contents would be much lower than the percentage of users without CJK chars in their site's contents, I'm withholding approval on the change.

@iBug, I recommend that you come up with a solution acceptable for both sides of the table.

doersino:
Suggestion:

1. Check whether the `input` contains any CJK characters. This could be implemented relatively cheaply.
2. If so, do what this pull request proposes.
3. Otherwise, do what's currently implemented.

ashmaroli:
> Check whether the `input` contains any CJK characters

That is one solution.
Another would be to have *an optional parameter* (set to ignore CJK chars by default). That way, the `input` string need not be checked on each run (`input` string can be long; checking the parameter would be a constant effort; the number of times this filter is called can be large in large sites.)

Benchmarking (and optionally profiling) the two alternatives can help decide which would be the better choice.

iBug:
@ashmaroli @doersino I've pushed a patch for this. Thank you for your precious analyses and suggestions!

A summary of the new commit:

- According to @doersino 's suggestions, I first check for CJK characters, and go the old way if none is found
- According to @ashmaroli 's analyses, `String#gsub` is heavily bloated for creating copies of (usually) large strings. Using `String#split` with a regex as the argument (presumably) solves this by avoiding the big copy.

ashmaroli:
@iBug in case you missed my suggestion earlier: https://github.com/jekyll/jekyll/pull/7813#issuecomment-632248204

Additionally, you need not push commits in a hurry. It is only going to trigger unnecessary CI builds. Take your time, investigate why the test-suite is failing, run tests locally and then push the commit.

iBug:
@ashmaroli That's a good point. I'll probably come back soon, like a few hours or days or so. Thank you for your tips.

iBug:
@ashmaroli I don't know how to profile so I only did some benchmarking.

Conclusions first:

I ran the script multiple times but couldn't obtain consistent results as values from each run varies a lot, some times 2x as much. Despite that, the comparison between values from the same run would still give an insight into how things work out. It isn't surprising to see that the "auto detect" mode runs as slow as the "let user decide it" mode when enabled.

The three documents are the longest ones I could find. Their sizes are

```text
1.txt   21293
2.txt   16637
3.txt   13778
```

Keep in mind, though, that these timings are 2000x amplified. Paired with the fact that this only affects the performance where `number_of_words` is heavily employed, I would say it's unlikely this change leads to extra load or build time on the majority of websites.

### Results

```text
                   user     system      total        real
Original 1     0.452091   0.000057   0.452148 (  0.452193)
Original 2     0.946308   0.000014   0.946322 (  0.946364)
Original 3     0.248744   0.000000   0.248744 (  0.248842)
Auto 1         1.734279   0.000000   1.734279 (  1.734335)
Auto 2         3.335226   0.000000   3.335226 (  3.335416)
Auto 3         2.967918   0.000000   2.967918 (  2.968079)
Overhead 1     0.449488   0.000000   0.449488 (  0.449499)
Overhead 2     1.028173   0.000000   1.028173 (  1.028210)
Overhead 3     0.179704   0.000000   0.179704 (  0.179709)
User 1         1.724147   0.000000   1.724147 (  1.728664)
User 2         2.282451   0.000000   2.282451 (  2.285302)
User 3         2.764753   0.000000   2.764753 (  2.764977)
```

### Test environment

- Intel Core i7-8850H (sufficient cooling), 64 GiB DDR4 3200 dual-channel
- Ubuntu 20.04 LTS
- Ruby 2.7.0p0 (from Ubuntu Focal, package `1:2.7+1`)

### Test code

```ruby
#!/usr/bin/ruby

require "benchmark"

# Sources:
#
# 1.txt - https://github.com/iBug/iBug-source/raw/20406be391bcfe93190861d2a20da1c14faed116/_posts/2019-12-28-mass-crawl-douban-with-aws.md
#     Top 20 lines removed (Front Matter)
#     Predominantly Latin text
#
# 2.txt - https://github.com/iBug/iBug-source/raw/bbb421f433ba8880e867e6917c7f215e02e720dc/_cn/2018-12-19-chisel-101.md
#     Top 6 lines removed (Front Matter)
#     Mixed Latin and CJK text
#
# 3.txt - https://github.com/iBug/iBug-source/raw/bbb421f433ba8880e867e6917c7f215e02e720dc/_cn/2019-08-14-speech-at-msc-2019.md
#     Top 12 lines removed (Front Matter)
#     Predominantly CJK text

text_1 = File.read "1.txt"
text_2 = File.read "2.txt"
text_3 = File.read "3.txt"

# Original state
def foo_0(input)
  input.split.length
end

# Auto-detect CJK
def foo_1(input)
  cjk_chars = '\p{Han}\p{Katakana}\p{Hiragana}\p{Hangul}'
  cjk_count = input.scan(Regexp.new("[#{cjk_chars}]")).length
  return input.split.length if cjk_count.zero?

  word_regex = "[^#{cjk_chars}\\s]+"
  cjk_count + input.scan(Regexp.new(word_regex)).length
end

# User-select, disabled
def foo_2(input)
  mode = "default"

  if mode == "cjk"
    cjk_chars = '\p{Han}\p{Katakana}\p{Hiragana}\p{Hangul}'
    cjk_count = input.scan(Regexp.new("[#{cjk_chars}]")).length
    word_regex = "[^#{cjk_chars}\\s]+"
    cjk_count + input.scan(Regexp.new(word_regex)).length
  else
    input.split.length
  end
end

# User-select, enabled
def foo_3(input)
  mode = "cjk"

  if mode == "cjk"
    cjk_chars = '\p{Han}\p{Katakana}\p{Hiragana}\p{Hangul}'
    cjk_count = input.scan(Regexp.new("[#{cjk_chars}]")).length
    word_regex = "[^#{cjk_chars}\\s]+"
    cjk_count + input.scan(Regexp.new(word_regex)).length
  else
    input.split.length
  end
end

########## Start Benchmarking ##########

matrix = [
  ["Original 1", :foo_0, text_1],
  ["Original 2", :foo_0, text_2],
  ["Original 3", :foo_0, text_3],
  ["Auto 1", :foo_1, text_1],
  ["Auto 2", :foo_1, text_2],
  ["Auto 3", :foo_1, text_3],
  ["Overhead 1", :foo_2, text_1],
  ["Overhead 2", :foo_2, text_2],
  ["Overhead 3", :foo_2, text_3],
  ["User 1", :foo_1, text_1],
  ["User 2", :foo_1, text_2],
  ["User 3", :foo_1, text_3],
]

Benchmark.bm 12 do |bm|
  matrix.each do |title, func, arg|
    bm.report title do
      2000.times do
        method(func).call arg
      end
    end
  end
end
```

ashmaroli:
Thank you for the extensive report and hats off for all the work done here.
From the benchmark results, it is clear that `auto mode` and `cjk mode` are slow in comparison to the `default mode`.

Since I'm assuming that `non_cjk_userbase.size > cjk_userbase.size`, I'm going to favor `Overhead *` results here.

iBug:
@ashmaroli With the last commit 21bf5a5dbd64fba88de9dce6c8b720f81fd97872 pushed, I've changed the code to the "Overhead Mode" (optional but disabled).

I've also updated the tests so both the new behavior and the old behavior are tested (essentially testing the optional parameter). On my local environment both `rake test` and `rubocop -D` passes (of course preceded by `bundle exec`). Please let me know if any extra work is required.

Again, I'm very grateful to you two for being super helpful and patient on guiding me through this seemingly simple PR. It's certainly been a wonderful experience participating in this great open-source project and there's no denying that I've learned a lot from this.

ashmaroli:
You're welcome @iBug 
If you still have the benchmark script on hand, try the following version:
```ruby
def number_of_words(input, mode = nil)
  cjk_regex = %r![\p{Han}\p{Katakana}\p{Hiragana}\p{Hangul}]!
  word_regex = %r![^\p{Han}\p{Katakana}\p{Hiragana}\p{Hangul}\s]+!
  case mode
  when "cjk"
    input.scan(cjk_regex).length + input.scan(word_regex).length
  when "auto"
    cjk_count = input.scan(cjk_regex).length
    cjk_count.zero? ? input.split.length : cjk_count + input.scan(word_regex).length
  else
    input.split.length
  end
end
```

iBug:
@ashmaroli Sure. I've put the new test code on top of the original one in order to reduce clutter (avoid making this thread too long). The new results deserve a separate post so here they are.

(Keep in mind that each function is run `10000.times`)

```text
                 user     system      total        real
Original 1   4.099273   0.000369   4.099642 (  4.099745)
Original 2   4.565014   0.000003   4.565017 (  4.565135)
Original 3   1.514981   0.000000   1.514981 (  1.515023)
Auto 1      11.171122   0.000000  11.171122 ( 11.171594)
Auto 2      17.005940   0.000001  17.005941 ( 17.006658)
Auto 3      16.100305   0.000000  16.100305 ( 16.100708)
User1- 1     4.443806   0.000000   4.443806 (  4.444016)
User1- 2     4.111661   0.000000   4.111661 (  4.111847)
User1- 3     1.607525   0.000000   1.607525 (  1.607591)
User1+ 1    20.131011   0.000000  20.131011 ( 20.131919)
User1+ 2    16.699124   0.000000  16.699124 ( 16.699834)
User1+ 3    16.022139   0.000000  16.022139 ( 16.022594)
User2- 1     4.704363   0.000000   4.704363 (  4.704513)
User2- 2     3.944270   0.000000   3.944270 (  3.944377)
User2- 3     2.134202   0.000000   2.134202 (  2.134379)
User2* 1    11.197304   0.000000  11.197304 ( 11.197632)
User2* 2    16.641564   0.000000  16.641564 ( 16.642144)
User2* 3    16.273725   0.000000  16.273725 ( 16.274143)
User2+ 1    19.845742   0.000000  19.845742 ( 19.846762)
User2+ 2    16.796103   0.000000  16.796103 ( 16.796858)
User2+ 3    15.658962   0.003737  15.662699 ( 15.663274)
```

Though, I still prefer the "auto mode" and would like to have it as the default behavior. Benchmark results show an increase of less than 1 ms increase per 20 KB source processed. This level of performance penalty falls *far behind* the deviation from computers' intrinsics. To be exact, this is not going to be observable in real runs of Jekyll.

crispgm:
@iBug i guess you should add some tests on CJK languages other than Chinese. :)

ashmaroli:
@iBug The takeaway from my suggestion above is:
- The *default* behavior will be the existing behavior.
- I'm willing to ship the `auto` mode option as well.
- Using `%r!...!` is better than `Regexp.new('...')` (*the latter generates a regex on each call*).

ashmaroli:
@crispgm Will you be able to provide sample texts in Japanese or Korean?

iBug:
Is it possible to make the default option a site-wide setting? Let's say, something like `site.number_of_words_mode`.

ashmaroli:
> Is it possible to make the default option a site-wide setting?

That would delay this PR a lot.
Better alternative would be to use include files.

```
// sample include file

{{ include.text | number_of_words: 'auto' }}
```
```
// template or page

{% assign foobar = 'some content' %}
{% include number_of_words.html text = foobar %}
```

crispgm:
> @crispgm Will you be able to provide sample texts in Japanese or Korean?

@ashmaroli okay, here they are
```
こんにちは世界！
안녕하세요 세상!
```
and powered by Google Translate.

ashmaroli:
@crispgm Ah! good ol' Google Translate. :) But you forgot to mention how many words they are.
We can't know if the output is correct until there's something to check against.. :wink:

crispgm:
> @crispgm Ah! good ol' Google Translate. :) But you forgot to mention how many words they are.
> We can't know if the output is correct until there's something to check against.. 😉

This is true. I updated on last posts. :)

ashmaroli:
@iBug Please update the documentation as well so that users can differentiate when to use the two modes. Once again, thank you for your patience and effort.

iBug:
@ashmaroli I was already planning to do so, but I failed to identify which pieces of files I should look into.

ashmaroli:
> which pieces of files

Adding to what you have in `docs/_data/jekyll_filters.yml` should be sufficient.

iBug:
@ashmaroli With the latest commit, I think it's almost there.

ashmaroli:
Thank you @iBug and everyone that participated here.
@jekyllbot: merge +minor

PR #7821, opened 2019-09-19
Test where filter handling numeric property values

### LINKED ISSUE #7820: "where" filter doesn't appear to accept integer values in 4.0
- I updated to the latest Jekyll
  - I ran `jekyll doctor` to check my configuration

## My Environment

| Operating System |    MacOS 10.14.5        |
| `jekyll`         | Latest  (4.0)   |
| `github-pages`   | No     |

---

## Expected Behaviour

- when making a template assignment
- and using a where filter,
- and value is an integer,

we should return the corresponding assigned object

## Current Behavior

Since upgrade from 3.x to 4.0, I no longer return an object if the value is a number. ("where_exp" does work, so does a "where" filter with a string value).

## Code Sample

Assume an "orgs" collection, containing hr.md with `deptid: 2765` and in some site template I only want to pull data about the HR department:
`
{%- assign dept = site.orgs | where: "deptid", 2765 | first -%}
`

`{{ dept.other_metadata }}` returns relevant data in 3.x, nothing in 4.0.

these would work however:

`
{%- assign dept = site.orgs | where: "deptid", "hr" | first -%}
`
if hr.md's id == "hr" instead; success.

`
{%- assign dept = site.orgs | where_exp: "dept", "dept.deptid == 2765" | first -%}
`
where_exp with INT; also success.

Maybe this is just a change in the latest liquid, either way might be worth documentation for migrations.

ashmaroli:
## Context

Closes #7820


DirtyF:
@mattr- Do you mind shipping a 4.0.1?

ashmaroli:
Before we ship this, I would like to know if merging #7878 would fix #7820 as well. If it does, then this PR need not be merged..

ashmaroli:
Since #7820 has been resolved on `master` (via #7878), this has been refactored to augment the test-suite.

DirtyF:
@jekyllbot: merge +dev

PR #7785, opened 2019-08-17
CI: GitHub Actions

DirtyF:
Simple test to see if we could use GitHub Actions to test on Linux and Windows.

PR #7822, opened 2019-09-22
Ci/GitHub actions

ethomson:
This is a 🙋 feature or enhancement. Picking up  #7785 and seeing if I can sort out the failures.


ethomson:
Alas, I'm going to have to 👀 more deeply at the line ending issues.

ashmaroli:
@ethomson Thank you for working on this. A few suggestions in case you're planning to push more commits:
- The primary reason for failures at the moment is RuboCop offenses. Please run `script/fmt -a` (or `bundle exec rubocop -a` to correct them.)
- It'd be better if the `#temp_dir` was moved into the `test/helper.rb` file and refactored to automatically join given arguments (see `module DirectoryHelpers` in the test helper for clarification / reference).

That said, in case you're not able to make the above changes, let me know. I shall push them myself. :smiley: 

P.S. *Is there a way to configure GitHub Actions suite to re-run individual check runs?*

ethomson:
Happy to make these changes - let me know if that's not what you have in mind - it looks like there's still a few test failures that I'm not sure I understand yet, could you help me understand them?

> P.S. Is there a way to configure GitHub Actions suite to re-run individual check runs?

Unfortunately, no, not yet.  You can re-run the entire build workflow at the moment, not individual check runs.

ashmaroli:
> it looks like there's still a few test failures that I'm not sure I understand yet

Actually, those failures are false negatives. They usually disappear with reruns.
Overall, the diff looks good to me.

ashmaroli:
@ethomson Looks like there is bug when a *rerun check-suite* event is triggered. The re-run event is not being intercepted properly. The Actions for the latest commit are still *in the yellow..*

ethomson:
>  Looks like there is bug when a rerun check-suite event is triggered. The re-run event is not being intercepted properly. The Actions for the latest commit are still in the yellow..

Ouch, thanks for letting me know.  I think that this was part of the [availability issue](https://www.githubstatus.com/incidents/b54zv9nz8m01) we had earlier today.  Could you try re-running it again so that I can see what happens?

ashmaroli:
> Could you try re-running it again so that I can see what happens?

Perhaps I don't have the required privileges, or perhaps it is just bugs. But once I click the rerun button. the only appreciable changes are:
- Flash message stating that rerun requested successfully
- Button text changes to `Cancel check suite`
- Check status indicator updates to *yellow*

However when I consequently click the `Cancel check suite`, only the flash message is updated and rendered. The button text remains unchanged.

In both situations, no change is discernible in the VM panels.

ethomson:
Thanks for the information, we're investigating why that particular check run is misbehaving.  In the meantime, I pushed an update to make sure that checks run were working more generally.

ashmaroli:
@ethomson Thanks for the trigger. All check runs in the Actions suite have passed successfully.

mattr-:
@jekyllbot: merge +dev

PR #7846, opened 2019-10-06
Rubocop version upgrade

### LINKED ISSUE #7836: Update dependency constraint to allow for rubocop v0.75.0
Hey there! :wave:

The constraint I found was `~> 0.72.0`, and the latest version available is `0.75.0`.

donquxiote:
This is a 🐛 bug fix.

## Summary
Upgrades the rubocop dependency to 0.75.0, the latest version, so others who pull down the project can use the latest.

Also adds some inline disabling to resolve rubocop issues that appear to be intentionally written in for readability.

Verified all tests pass locally with `script/cibuild`

## Context

Resolves #7836


donquxiote:
using rubocop 0.75.0 triggers a few formatting warnings, which I've found and can add the `rubcop:disable` comment for as part of this PR, but I wasn't sure how liberal y'all like to be with those comments.

ashmaroli:
> I've found and can add the `rubcop:disable` comment

@donquxiote We prefer having a `.rubocop_todo.yml` generated instead. I shall however leave it to you to figure out how to get the RuboCop-TODO file generated and what recommendations within it can be immediately tackled.
IMHO, there's no point in simply updating to the latest RuboCop version if we don't act on its recommendations with earnest.

donquxiote:
After running rubocop with the `auto-gen-config` I've added the rubocop_todo that it generated. However using rubocop's inherit_from function in the config broke the rubocop check when I ran it locally. So I added the line from the `rubocop_todo` into the regular `.rubocop.yml`

ashmaroli:
> So I added the line from the `rubocop_todo` into the regular `.rubocop.yml`

Since *the only entry* in the todo file is now excluded via the main config file, there is no point in having a todo file, is there..?

However, I see some whitespace changes as well. If those were made to appease the `Layout/SpaceAroundOperators` cop, then please revert those changes and disable the cop via the TODO file because it appears to have a bug. The cop is configured to `AllowForAlignment` by default. So it doesn't work as advertised.
Having it entered in the todo file will remind us to check again with a future RuboCop version.

Thank you.

donquxiote:
You had a good point about the rubocop_todo file. That file is now present and the rubocop.yml should inherit from it. I also removed those spacing corrections and moved those to the todo file as well.

ashmaroli:
Thank you very much @donquxiote 

@jekyllbot: merge +dev

PR #7856, opened 2019-10-11
Doc: minor fix, should be greater or equal to min version [ci skip]

crispgm:

## Summary

Minor fix on 3 to 4 migrating guide. It should be `>=` 2.4 (ruby min version) instead of `>`.


ashmaroli:
Thank you @crispgm 
@jekyllbot: merge +docs

PR #7860, opened 2019-10-14
Update README.markdown

bradykondek:


ashmaroli:
Thank you for submitting this change @bradykondek but I feel it is not necessary as static site generators have no say in determining domains.

PR #7865, opened 2019-10-18
Include _config.yml in a new theme's gemspec

bigelowcc:
This is a 🐛 bug fix.

## Summary

#7304 introduced loading the `_config.yml` from a theme, but the gemspec doesn't include this file.

Currently a site will correctly include values from the theme's `_config.yml` locally, but not once the theme is packaged and pulled from a gem repository, as the file doesn't exist.

This PR adds `_config.yml` to the gemspec generated when you run `jekyll new-theme`.

## Context

#7304 introduced this feature.

## Notes

- I haven't added a test for this change as there currently doesn't seem to be a test around gem packaging
- All existing themes will still encounter this problem unless they amend their gemspec files to include `_config.yml`

### DIFF
--- a/features/step_definitions.rb
+++ b/features/step_definitions.rb
@@ -239,9 +239,14 @@ end
 
 When(%r!^I decide to build the theme gem$!) do
   Dir.chdir(Paths.theme_gem_dir)
-  File.new("_includes/blank.html", "w")
-  File.new("_sass/blank.scss", "w")
-  File.new("assets/blank.scss", "w")
+  [
+    "_includes/blank.html",
+    "_sass/blank.scss",
+    "assets/blank.scss",
+    "_config.yml"
+  ].each do |filename|
+    File.new(filename, "w")
+  end
 end
 
 #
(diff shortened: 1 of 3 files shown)


ashmaroli:
Thank you for submitting this @bigelowcc.
I should've foreseen this and documented appropriately. Better late than never.

Regarding tests, we currently cover *this aspect* via Cucumber features.

As for the proposed change, the feature is currently limited to just loading **`_config.yml`** at the root of the theme-gem. `_config.yaml` will be ignored.

ashmaroli:
Not merging this right away since we need to decide if Jekyll should bundle `_config.yml` by default.
Theme authors should ideally opt-in voluntarily to bundle the config file in the gem..

mattr-:
If you're packaging a theme as a gem, we should make it easiest for the user to use the gem by default. Theme authors should opt-out.

@jekyllbot: merge +minor

ashmaroli:
> Theme authors should opt-out.

Note to myself: *Ensure that this is mentioned in the v4.1.0 release-post since novice users won't realize they're bundling their config file right away.*

PR #7870, opened 2019-10-20
Document sorting two documents by their `date`

citelao:

## Summary

Collections sort by `date` if both documents have that key, otherwise they sort by `path`. This was giving me problems, so I figured I'd at least correct the documentation.

## Context

N/A

## How tested

I based this change off experienced behavior and code inspection.

- Collections *fall back to* `Document#<=>` if `sort_by` is missing or non-sortable (and call `docs.sort!` if the `sort_by` isn't defined for either):
     https://github.com/jekyll/jekyll/blob/dd7d03eccdefd7bc4227e577dd409fde52fc7eb8/lib/jekyll/collection.rb#L240
- `Document#<=>`  tries to compare date, falling back to `path`:
    https://github.com/jekyll/jekyll/blob/dd7d03eccdefd7bc4227e577dd409fde52fc7eb8/lib/jekyll/document.rb#L336-L342
Thanks!

### DIFF
--- a/docs/_docs/collections.md
+++ b/docs/_docs/collections.md
@@ -127,7 +127,7 @@ help you control the output url for the entire collection.
 
 By default, two documents in a collection are sorted by their `date` attribute when both of them have the `date` key in their front matter. However, if either or both documents do not have the `date` key in their front matter, they are sorted by their respective paths.
 
-You can control this sorting via the collection's metadata:
+You can control this sorting via the collection's metadata.
 
 ### Sort By Front Matter Key
 
--- a/docs/_docs/collections.md
+++ b/docs/_docs/collections.md
@@ -125,7 +125,9 @@ help you control the output url for the entire collection.
 
 ## Custom Sorting of Documents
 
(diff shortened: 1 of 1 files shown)


docs/_docs/collections.md
```diff
@@ -125,7 +125,7 @@ help you control the output url for the entire collection.
 
 ## Custom Sorting of Documents
 
-By default, documents in a collection are sorted by their paths. But you can control this sorting via the collection's metadata.
+By default, documents in a collection are sorted by their `date`s (if both have that front-matter key) or by their paths (if one or both do not have the `date` key). But you can control this sorting via the collection's metadata.
```
ashmaroli: How about being a little more verbose? What say?

```suggestion
By default, two documents in a collection are sorted by their `date` attribute when both of them have the `date` key in their front matter. However, if either or both documents do not have the `date` key in their front matter, they are sorted by their respective paths.

You can control this sorting via the collection's metadata:
```

citelao: Sounds good!

devmoses: It's more clear now. Good job!

DirtyF:
@jekyllbot: merge +docs

PR #7872, opened 2019-10-22
change collection example from sequence to mapping

itaysk:

## Summary

The example for using collections uses a yaml sequence instead of mapping. I did not see this style being used anywhere with collections, and skimming over https://github.com/jekyll/jekyll/blob/2736589ba19ab8390168be58480df4f359745464/lib/jekyll/configuration.rb it seems like the parsed collections configuration is expected to be a hashmap rather then array.

### DIFF
--- a/docs/_docs/collections.md
+++ b/docs/_docs/collections.md
@@ -13,7 +13,7 @@ example here's a collection of staff members:
 
 ```yaml
 collections:
-  - staff_members
+  staff_members:
 ```
 
 You can optionally specify metadata for your collection in the configuration:


ashmaroli:
@itaysk Perhaps the documentation is not clear about the fact, but `collections` *can be a sequence or a mapping* based on *the presence of metadata*...

For example, consider the following two different configurations:
```yaml
collections:
  - recipes
  - travels
  - movies
```
```yaml
collections:
  recipes:
    output: true
  travels:
    output: true
    foo: bar
  movies:
    output: false
```
Both of above are valid usage and will register the same no. of collections for the site, but has a slightly different behavior nevertheless:
- With a *sequence*, individual documents of all collections (except `posts`) are not written to the destination folder. However, they allow iteration in templates e.g. `{% for doc in site.travels %}...`
- With a *mapping*, you have control over which collection's documents are not to be written to the destination and which are to be written by setting the `output` metadata. (*Posts will always get written.*). From Jekyll 4.0 onwards, you can use an `order` metadata to manually sort documents to your desire..

itaysk:
got it, thanks for clarifying (IMO it would be nice adding this clarification to the doc)

ashmaroli:
Feel free to submit a PR to improve https://jekyllrb.com/docs/collections/

PR #7873, opened 2019-10-22
clarify _config.yml/collections type

itaysk:

## Summary
Clarifying the expected type of the `collections` field.

## Context
based on comment from: https://github.com/jekyll/jekyll/pull/7872

### DIFF
--- a/docs/_docs/collections.md
+++ b/docs/_docs/collections.md
@@ -15,14 +15,18 @@ example here's a collection of staff members:
 collections:
   - staff_members
 ```
-
-You can optionally specify metadata for your collection in the configuration:
+In this case `collections` is defined as a squence (i.e array) with no additional metadata defined for each collection.  
+You can optionally specify metadata for your collection by defining `collections` as a mapping (i.e hashmap) instead of sequence, and then defining additional fields in it:
 
 ```yaml
 collections:
   staff_members:
     people: true
 ```
+
+<div class="note">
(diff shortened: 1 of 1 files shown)


docs/_docs/collections.md
```diff
@@ -15,15 +15,19 @@ example here's a collection of staff members:
 collections:
   - staff_members
 ```
-
-You can optionally specify metadata for your collection in the configuration:
+In this case `collections` is defined as a squence (i.e array) with no additional metadata defined for each collection.
```
DirtyF:
```suggestion
In this case `collections` is defined as a sequence (i.e array) with no additional metadata defined for each collection.  
```

docs/_docs/collections.md
```diff
@@ -15,15 +15,19 @@ example here's a collection of staff members:
...
+You can optionally specify metadata for your collection by defining `collections` as a mapping (i.e hashmap) instead of sequence, and then defining additional fields in it:
 
 ```yaml
 collections:
   staff_members:
     people: true
 ```
 
+<div class="note">
+  <p>When defining a collection as a sequence, it's pages will not be rendered by default. To enable this, <code>output: true</code> must be specified on the collection, which requires defining the collection as a mapping. For more information, see the section <a href="#Output">Output</a></p>
```
DirtyF:
```suggestion
  <p>When defining a collection as a sequence, it's pages will not be rendered by default. To enable this, <code>output: true</code> must be specified on the collection, which requires defining the collection as a mapping. For more information, see the section <a href="#output">Output</a></p>
```

DirtyF:
@jekyllbot: merge +docs

PR #7878, opened 2019-10-26
Update item_property to recognize integers

### LINKED ISSUE #7827: Number-like strings are converted to floats by certain Jekyll Filters
## My Environment

| Operating System |      All Platforms      |
| `jekyll`         |  **`master`** and `v4.0.0`    |

---

## Expected Behaviour

Expected whole number strings to be rendered verbatim.

## Current Behavior

filter-bug

summerisgone:
This is a 🐛 bug fix

- [x]  I've added tests

- [x] no need to update documentation

- [x] The test suite passes locally

## Summary

Method `item_property` aggressively parses floats from values. In order to properly render integers I added integer check before float check. I guess better solution would be type check, but I'm afraid it might break other things.

## Context

Fixes #7827

--
P.S.
Thanks for wonderful software, glad to bring my 5 cents!

### DIFF
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -371,10 +371,15 @@ module Jekyll
     # rubocop:disable Performance/RegexpMatch
     # return numeric values as numbers for proper sorting
     def parse_sort_input(property)
-      number_like = %r!\A\s*-?(?:\d+\.?\d*|\.\d+)\s*\Z!
-      return property.to_f if property.to_s =~ number_like
-
-      property
+      float_like = %r!\A\s*-?(?:\d+\.?\d*|\.\d+)\s*\Z!
+      integer_like = %r!\A\s*-?\d+\s*\Z!
+      if property.to_s =~ integer_like
+        property.to_i
+      elsif property.to_s =~ float_like
+        property.to_f
+      else
+        property
+      end
     end
     # rubocop:enable Performance/RegexpMatch
 
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -8,11 +8,6 @@ module Jekyll
     include GroupingFilters
     include DateFilters
 
-    FLOAT_LIKE = %r!\A\s*-?(?:\d+\.?\d*|\.\d+)\s*\Z!
-    INTEGER_LIKE = %r!\A\s*-?\d+\s*\Z!
-    private_constant :FLOAT_LIKE
-    private_constant :INTEGER_LIKE
-
     # Convert a Markdown string into HTML output.
     #
     # input - The Markdown String to convert.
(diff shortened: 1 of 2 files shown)


parkr:
Two quick questions:

1. Does the Liquid library have a method for casting to numbers that we could use instead of our own implementation?
2. Are the Windows failures expected?

ashmaroli:
> 2. Are the Windows failures expected?

Windows failures are only on GH Actions: There's a PR to patch that: #7885

ashmaroli:
> 1. Does the Liquid library have a method for casting to numbers that we could use instead of our own implementation?

Liquid has a utility function:
https://github.com/Shopify/liquid/blob/57c9cf64ebc777fe5e92d4408d31a911f087eeb4/lib/liquid/utils.rb#L48-L63
```ruby
    def self.to_number(obj)
      case obj
      when Float
        BigDecimal(obj.to_s)
      when Numeric
        obj
      when String
        /\A-?\d+\.\d+\z/.match?(obj.strip) ? BigDecimal(obj) : obj.to_i
      else
        if obj.respond_to?(:to_number)
          obj.to_number
        else
          0
        end
      end
    end
```

summerisgone:
@parkr @ashmaroli 
> Does the Liquid library have a method for casting to numbers that we could use instead of our own implementation?
Actually it is not required to cast to string or to any type at all, since data comes in typed manner from yaml files. But `item_property` function is used for filtering, sorting and grouping. I think in-place solution might fit better then moving cast to templates, since it would have less side effects.

ashmaroli:
> it is not required to cast to string or to any type at all

@summerisgone The reason `#parse_sort_filter` exists is to allow *sorting* of stringified integers.
For example, Ruby would sort `["15", "20", "10", "120"]` into `["10", "120", "15", "20"]` which is not acceptable. But if we were to cast those strings to Integer type and then sort the array, the result would be `[10, 15, 20, 120]`, which would in turn result in the expected results for *filtering, sorting or grouping*.

summerisgone:
@ashmaroli what is your suggestion, to import `to_number` from Liquid utils instead of `parse_sort_input`? Shall I update the PR?
Consider it won't parse spaces, but I think it might be OK 🤷‍♂️

ashmaroli:
@summerisgone Its a public method. So you can call it directly. For example:
```ruby
Liquid::Utils.to_number(input)
```
However, you may need to add additional tests to ensure expected results for various objects.

summerisgone:
@ashmaroli I tried to use `Liquid::Utils.to_number`, but it does convert all non-numbers to 0, which is not desired when sorting strings. I think it is not drop-in replacement for `parse_sort_input` unfortunately

summerisgone:
@ashmaroli what else blocks the merge?

ashmaroli:
@summerisgone The change looks okay.
I'm just wondering if there's a way to go about this without stringifying `property`..
But since the diff shows that `property` is already being stringified on `master`, you can rest assured that this will get merged.
Optimizations if any will be via a separate pull request.

mattr-:
@jekyllbot: merge +minor

PR #7885, opened 2019-11-01
Split action steps to avoid using `&&` on Windows

ashmaroli:
The step currently fails on `master` with the following message:

    gem update --system --no-document && gem update bundler --no-document
                                      ~~
    The token '&&' is not a valid statement separator in this version.
      CategoryInfo          : ParserError: (:) [], ParseException
      FullyQualifiedErrorId : InvalidEndOfLine


ashmaroli:
@jekyllbot: merge +dev

PR #7887, opened 2019-11-01
fixed grammatical error (it's --> its)

ArthurZey:

## Summary
Before: When defining a collection as a sequence, it's pages will not be rendered by default.
After: When defining a collection as a sequence, its pages will not be rendered by default.
Justification: "its" is possessive; "it's" is "it is".


DirtyF:
@jekyll: merge +docs

PR #7910, opened 2019-11-26
Upgrade rake to use version 13

### LINKED ISSUE #7900: Update dependency constraint to allow for rake v13.0.1
Hey there! :wave:

I noticed that the constraint you have for rake doesn't allow for the latest version to be used.

The constraint I found was `~> 12.0`, and the latest version available is `13.0.1`.

thetizzo:

## Summary

This PR updates the version number for `rake` in the Gemfile to use `~> 13.0` in order to solve issue #7900 

I have not added any tests but running `script/cibuild` passes locally.

## Context

Resolves #7900

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ gemspec :name => "jekyll"
 # refinements introduced in i18n-1.3.0
 gem "i18n", "~> 1.2.0" if RUBY_ENGINE == "jruby"
 
-gem "rake", "~> 12.0"
+gem "rake", "~> 13.0"
 
 group :development do
   gem "launchy", "~> 2.3"


ashmaroli:
Hello @thetizzo.
Thank for submitting this PR.

The current change forces upgrade to Rake 13. That is unnecessary.
Instead it'd be better if you would just *relax the version constraint*:
```ruby
gem "rake", ">= 12.0", "< 14.0"
```

thetizzo:
@ashmaroli I considered relaxing the requirements in that way but my thought was that since rake is only referenced in the Gemfile and not the gemspec then this won't effect users of Jekyll, just developers of Jekyll so it would be alright to just bump it to the next version.

Since others have since approved this PR, I'm not sure I should make this change at this point.  Let me know if this changes needs to be made and I can make it.

mattr-:
This change is fine as-is, from point of view. Just waiting on CI to pass and then it'll merge.

DirtyF:
@jekyllbot: merge +dev

PR #7912, opened 2019-11-28
add front matter dashes to _includes/default.html

davidhullster:
so newbies don't get lost trying to understand why their includes don't work


ashmaroli:
Includes don't need front matter dashes to work..
Therefore closing this.

PR #7913, opened 2019-11-28
say where to find jekyll plugins

noraj:


DirtyF:
We have some links on https://jekyllrb.com/resources/, but yeah we should add some links here too.

noraj:
Also I spotted there are duplicate and non-synced resources:

image
image

DirtyF:
I guess we could store those links in a data file, and create some includes, to avoid that. 😄

noraj:
@DirtyF I added a data file.

PR #7914, opened 2019-11-29
Fix typo in css/scss styles filename

chbrandt:
Inside `/assets/css` directory, files -- eg, `styles` -- go with `css` extension, not `scss`


chbrandt:
My mistake. Closing it.

PR #7918, opened 2019-12-02
docs: add options for watch and force polling

patrikerikss0n:

## Summary

Added config options for regeneration (watch) and force polling to the documentation. We use jekyll in a combination with docker, npm and webpack and we like to keep all build (serve) commands in the config instead of the command line. The only command we run is `jekyll serve --config fake_config.dev.yml` and let the config(s) take care of the rest.

### DIFF
--- a/docs/_docs/configuration/options.md
+++ b/docs/_docs/configuration/options.md
@@ -171,6 +171,7 @@ class="flag">flags</code> (specified on the command-line) that control them.
         <p class="description">Enable auto-regeneration of the site when files are modified.</p>
       </td>
       <td class="align-center">
+        <p><code class="option">watch: BOOL</code></p>
         <p><code class="flag">-w, --[no-]watch</code></p>
       </td>
     </tr>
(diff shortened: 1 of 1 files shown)


docs/_docs/configuration/options.md
```diff
@@ -171,6 +171,7 @@ class="flag">flags</code> (specified on the command-line) that control them.
         <p class="description">Enable auto-regeneration of the site when files are modified.</p>
       </td>
       <td class="align-center">
+        <p><code class="option">watch: BOOL</code></p>
```
DirtyF: According to https://github.com/Zequez/jekyll/blob/master/lib/jekyll/commands/build.rb#L75 this should be `force_polling`

patrikerikss0n: Sorry, did you want me to edit something? `force_polling` is also set in this PR.

DirtyF:
@jekyllbot: merge +docs

PR #7925, opened 2019-12-05
File extension should be "markdown" not "markup".

BillRaymond:
See the Summary section in this pull request for details.

## Summary

This is an update to the POSTS documentation. It states that "MARKUP" is the file extension when it should be "MARKDOWN". Also, the two examples show ".md" with the extension, but when I test (at least on GitHub Pages), the post will not show unless I use a ".markdown" extension, so suggest changing that as well.

### DIFF
--- a/docs/_docs/posts.md
+++ b/docs/_docs/posts.md
@@ -24,12 +24,12 @@ YEAR-MONTH-DAY-title.MARKUP
 ```
 
 Where `YEAR` is a four-digit number, `MONTH` and `DAY` are both two-digit
-numbers, and `MARKUP` is the file extension representing the format used in the
+numbers, and `MARKDOWN` is the file extension representing the format used in the
 file. For example, the following are examples of valid post filenames:
 
 ```
-2011-12-31-new-years-eve-is-awesome.md
-2012-09-12-how-to-write-a-blog.md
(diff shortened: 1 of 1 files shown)


ashmaroli:
@BillRaymond These are not mistakes but intentional usage.
In `YEAR-MONTH-DAY-title.MARKUP`, the individual words are placeholders and should not be taken literally.
For example, a file literally named `YEAR-MONTH-DAY-title.markdown` **is not a valid post**.

The reason we chose `MARKUP` is because the extension can be any of the Markdown extensions (`.markdown`, `.mkdown`, `.mkdn`, `.mkd`, `.md`) if the content is in Markdown.
Even HTML files can also be posts. i.e., `2019-12-05-hello-world.html` is a valid post filename.

You're free to test the above info and make changes to this pull request to improve the quality of the documentation.

BillRaymond:
Yesterday, I was creating blog postings with an MD extension and every build failed, with GitHub sending me an email. I changed to .markdown and all was good. Today, I rename back to .md and all is good, so go figure. Anyway, I apologize for not understanding that MARKUP is the right term, but as a brand new user, MARKDOWN makes a lot more sense (as a reader of the docs). I will close. Thank you for the review though.

PR #7928, opened 2019-12-05
Add an option to easily disable disk-cache

### LINKED ISSUE #7804: feat: Disable disk caching, or set cache directories outside of source
## Summary

This is a simple request to allow me to disable disk caching and/or set the cache directories outside of the source directory.

## Motivation

There are certain scenarios where a disk cache is not required and can be a nuisance. This can't be achieved unless implemented in core, as far as I can tell?

1. In production, I have no need of a persistent disk cache as the build is always run in a fresh virtual machine in the cloud. There is no disk persistence between builds.
2. In development, I don't need disk caching. I'm changing so much of the structure of the site while I develop, there is no point in a cache and, because I'm worried the cache will interfere with development, I end up cleaning it constantly.
3. Also for development, while I can .gitignore the .jekyllcache folder, this isn't possible with certain other cloud sync/collaboration tools. My development folder is cloud synced and the sync engine goes haywire with all the cache files.

## Guide-level explanation

Either or both of these would solve the issues noted above:

1. Offer a build configuration option to disable disk caching
2. Remove the check that forces the cache_dir to be within the source dir

As an example of a build directive:

* `jekyll serve --no_cache`

As an example of how the cache directory could live outside of the source folder:
* ~/
    * dev/
        * my-site/
            * source/
            * public/
            * _config.yml → `cache_dir: ~/.jekyll-cache/my-site`
            * etc
    * .jekyll-cache/
        * my-site/

## Drawbacks

I am not clear why the cache_dir was forced to be within the source_dir

ashmaroli:
  - [x] I've added tests
  - [x] I've adjusted the documentation
  - [x] The test suite passes locally

## Summary

Running in `safe` mode just to disable Jekyll 4.0's disk-caching seems like an overkill.
Plugins would need to be modified to be compatible with `safe` mode, whitelist gems...

With this change one can disable disk-cache in non-*safe* mode via either `--disable-disk-cache` CLI switch or `disable_disk_cache: true` configuration.

## Context

Closes #7804 
/cc @Convincible

### DIFF
--- a/docs/_docs/configuration/options.md
+++ b/docs/_docs/configuration/options.md
@@ -43,11 +43,33 @@ class="flag">flags</code> (specified on the command-line) that control them.
     <tr class="setting">
       <td>
         <p class="name"><strong>Safe</strong></p>
-        <p class="description">Disable <a href="/docs/plugins/">custom plugins, and ignore symbolic links</a>.</p>
+        <p class="description">
+          Disable <a href="/docs/plugins/">custom plugins</a>, caching to disk
+          and ignore symbolic links.
+        </p>
       </td>
       <td class="align-center">
         <p><code class="option">safe: BOOL</code></p>
         <p><code class="flag">--safe</code></p>
+      </td>
+    </tr>
+    <tr class="setting">
+      <td>
+        <p class="name">
+          <strong>Disable Disk Cache</strong>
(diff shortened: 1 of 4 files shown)


DirtyF:
@ashmaroli 👍 for the implementation, tests should pass though

mattr-:
@jekyllbot: merge +minor

PR #7931, opened 2019-12-09
Include path in jekyll new commands (Usage docs)

MSVO:

## Summary
The command `jekyll new` to create a new jekyll site outputs an error `jekyll 4.0.0 | Error:  You must specify a path.`. The correct command is `jekyll new <path>`, where `<path>` should be specified by the user.

## Context
I noticed the issue while reading the docs, from where an *improve this page* link took me here.

### DIFF
--- a/docs/_docs/usage.md
+++ b/docs/_docs/usage.md
@@ -7,8 +7,8 @@ The Jekyll gem makes a `jekyll` executable available to you in your terminal.
 
 You can use this command in a number of ways:
 
-* `jekyll new some/path` - Creates a new Jekyll site with default gem-based theme at `./some/path`. The directories will be created as necessary.
-* `jekyll new some/path --blank` - Creates a new blank Jekyll site scaffold at path `./some/path`.
+* `jekyll new PATH` - Creates a new Jekyll site with default gem-based theme at `./some/path`. The directories will be created as necessary.
+* `jekyll new PATH --blank` - Creates a new blank Jekyll site scaffold at path `./some/path`.
 * `jekyll build` or `jekyll b` - Performs a one off build your site to `./_site` (by default)
 * `jekyll serve` or `jekyll s` - Builds your site any time a source file changes and serves it locally
 * `jekyll doctor` - Outputs any deprecation or configuration issues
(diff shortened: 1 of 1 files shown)


docs/_docs/usage.md
```diff
@@ -7,8 +7,8 @@ The Jekyll gem makes a `jekyll` executable available to you in your terminal.
 
 You can use this command in a number of ways:
 
-* `jekyll new` - Creates a new Jekyll site with default gem-based theme
-* `jekyll new --blank` - Creates a new blank Jekyll site scaffold
+* `jekyll new .` - Creates a new Jekyll site in present working directory with default gem-based theme. The `.` can be replaced with a path. The path will be created if it does not exist.
```
ashmaroli: Please recommend `jekyll new PATH` instead.
`jekyll new .` will throw an error if the current directory is not empty. In that situation, one will need to use `--force`:
```
jekyll new . --force
```

```suggestion
* `jekyll new some/path` - Creates a new Jekyll site with default gem-based theme at `./some/path`. The directories will be created as necessary.
```

docs/_docs/usage.md
```diff
@@ -7,8 +7,8 @@ The Jekyll gem makes a `jekyll` executable available to you in your terminal.
...
+* `jekyll new --blank .` - Creates a new blank Jekyll site scaffold in the present working directory.
```
ashmaroli: Similarly, here as well..
```suggestion
* `jekyll new some/path --blank` - Creates a new blank Jekyll site scaffold at path `./some/path`.
```

DirtyF: Let's make it consistent with the previous line (either `jekyll new [path] --blank` or `jekyll new PATH --blank`)

ashmaroli: I had suggested `jekyll new some/path` in the previous line, so `jekyll new some/path --blank` here..
@MSVO If you choose to go with the placeholder suggestion, opt for `jekyll new PATH` and
`jekyll new PATH --blank` for consistency with output from command `jekyll help new`

DirtyF:
@MSVO Thanks for the clarification 🙏 

@jekyllbot: merge +docs

ashmaroli:
@DirtyF Your last commit did not include complete changes. The lines now read:
```
`jekyll new PATH` - Creates a new Jekyll site with default gem-based theme at `./some/path`.
```
Please push a correction a.s.a.p..
Thanks.

PR #7940, opened 2019-12-20
docs: remove watch option in config

patrikerikss0n:

## Summary

In PR #7918 I added config options for `watch` and `force_polling` to the docs. However, after running this, it seems that jekyll says that using `watch` from config is deprecated and it should be used from the CLI instead:

```bash
Configuration file: _config.dev.yml
Deprecation: Auto-regeneration can no longer be set from your configuration file(s). Use the --[no-]watch/-w command-line option instead.
```

Perhaps the `watch` part of #7918 shouldn't have been added at all and the preferred way is using `--watch` (?).

This PR removes the `watch` config option.


ashmaroli:
Thank you for the correction @patrikerikss0n 
@jekyllbot: merge +docs

PR #7944, opened 2019-12-23
Update 01-setup.md

otherpaco:

## Summary

If you do not remove or rename index.markdown jekyll starts to render the basic theme/site the moment you use front matter in your index.html file (step 2 ff)

### DIFF
--- a/docs/_docs/step-by-step/01-setup.md
+++ b/docs/_docs/step-by-step/01-setup.md
@@ -52,6 +52,8 @@ is completely optional but it's a great habit to get into. You can learn more
 about using Git by reading through the
 [Git Handbook](https://guides.github.com/introduction/git-handbook/).
 
+Before you start, rename `index.markdown` to `index.markdown.bak` to prevent jekyll from rendering the basic theme. 
+
 Let's add your first file. Create `index.html` in the root with the following
 content:
 


ashmaroli:
I don't see the point here.
Why not just disable the theme instead?

otherpaco:
@ashmaroli I just started with jekyll and ran in the problem that step 2 of the tutorial is already broken.
Of course disable the theme would be the better way. I will add a sentence in Step 1.
Thanks for the tip

ashmaroli:
> step 2 of the tutorial is already broken.

Please elaborate. What error message did you get..?

otherpaco:
@ashmaroli disbaling the theme is not doing the trick. Seems like you have to remove the index.markdown or rename it to render the index.html.
Just try it yourself, follow step 1 and 2 of the tutorial.
If things are unclear I am happy to open an issue

otherpaco:
@ashmaroli there is no error message just use jekyll 4.0 and follow step 1 and 2 of the tutorial. As long as index.markdown exists Jekyll renders the default site instead of the hello world! downcase.

otherpaco:
@ashmaroli nevermind, my bad I wasnt following step 1 exactly and installed jekyll with the basic theme. I am sorry thank you for your support

PR #7950, opened 2019-12-26
Allow extensionless document in a strict site

ashmaroli:
- This is a 🐛 bug fix.
- I've added tests

## Background

Jekyll allows Documents in a custom collection to be without an extension (even if Markdown content in it won't get *converted*).
But in Jekyll 4.0, such documents output an error:
```
Error: could not read file source/_foo/bar: undefined method `gsub!' for nil:NilClass
```
But building proceeds since the error is rescued away.
However, if the site has been configured with `strict_front_matter: true`, the build aborts on error.

## Summary

Ensure a non-nil `slug` for extensionless document in a custom collection and not abort builds with `strict_front_matter: true`.
Extensionless document inside `_posts` continue to remain *unread*.


ashmaroli:
> why we allow extensionless documents in collection

@DirtyF I tested down until Jekyll 3.6..
 It doesn't look like if we ever *explicitly supported* extensionless documents in a collection.. :man_shrugging:

DirtyF:
We should rather aim at removing those discrepancies and align on posts behavior for extensionless documents.

ashmaroli:
> aim at removing those discrepancies and align on posts behavior for extensionless documents

Could you clarify with an example scenario..?

DirtyF:
I haven't got a specific scenario in mind, where you would need to have extensionless files in a collection, do you?

DirtyF:
@jekyll: merge +fix

PR #7961, opened 2020-01-03
Leave a footprint for every dev-server instance

### LINKED ISSUE #7956: feat: Warn user if running server in same environment where trying to build
## Summary

Situation: User is trying to do a production build passing in environment `JEKYLL_ENV=production bundle exec jekyll build`

--> BUT the user also has a development server running (say in VSCode from a week ago 😨).

Observed Behavior:

Jekyll will do the build. . . but it will prefer the development server's environment instead of the environment variable being passed in.

## Motivation

This happened to me on a side project and it was very frustrating. I didn't have any optics around why the system wasn't behaving as I would expect it to. . . it just felt like broken software.

I imagine that beginners are the folks most likely to get burned by this as they pull up some development server (as per the docs) and then try to do a production build in a different tab.

#DoItForTheBeginners

Preferred Behavior:

If a user attempts to run a `jekyll build` and we observe that in their system they also have a running development server then it should give a warning of some sort that declares that they are clobbering the environment and that they should go turn off the development server (perhaps with a tip like: You can see if you have a local development server running with a [command](https://stackoverflow.com/questions/3855127/find-and-kill-process-locking-port-3000-on-mac) like `netstat -vanp tcp | grep 4000` 

An implementation could work if we create a lockfile of some sort when the development server is running?

## Drawbacks

Time == money

ashmaroli:

## Summary

**Proof-of-concept**
Warn a user if there's an existing instance being served already.

## Context

Resolves #7956 

/cc @7hoenix In case you want to take this for a test-run:
```ruby
gem "jekyll", git: "https://github.com/jekyll/jekyll.git", ref: "refs/pull/7961/head"
```

### DIFF
--- a/lib/jekyll/commands/build.rb
+++ b/lib/jekyll/commands/build.rb
@@ -29,6 +29,8 @@ module Jekyll
           options = configuration_from_options(options)
           site = Jekyll::Site.new(options)
 
+          Footprint.check(options)
+
           if options.fetch("skip_initial_build", false)
             Jekyll.logger.warn "Build Warning:", "Skipping the initial build." \
                                " This may result in an out-of-date site."
@@ -91,3 +93,5 @@ module Jekyll
     end
   end
 end
+
+require_relative "build/footprint"
--- a/lib/jekyll/commands/build/footprint.rb
+++ b/lib/jekyll/commands/build/footprint.rb
@@ -1,0 +1,83 @@
+# frozen_string_literal: true
+
+module Jekyll
+  module Commands
+    class Build
+      class Footprint
+        class << self
+          def check(options)
+            @source ||= File.expand_path options["source"]
+            setup_registry
+            return if registry.empty?
+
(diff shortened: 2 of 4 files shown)


RMKoRn:
I can't pay you I'm a one-man business I'm sorry

ashmaroli:
@RMKoRn Who asked you to pay??

RMKoRn:
Ashwin MaroliREADME.md

ashmaroli:
Jekyll is a free software released under the MIT License. You don't have to pay to use it.

RMKoRn:
I just received a bill from these people I'm not paying it
https://secure.2co.com/myaccount/

ashmaroli:
Jekyll (and the Jekyll Org on GitHub) does not send anyone bills.
So I'm not even going to bother opening that link.

The Issue Tracker (or open Pull Requests) of this repository is not the place to talk about all this..

PR #7967, opened 2020-01-08
Clear cached Liquid template scope before render

### LINKED ISSUE #7811: Jekyll 4.0 gets confused about page variables
Hi, congrats on launching Jekyll 4! Very much enjoying the speed boost. 

While upgrading our site, I was syncing `_site` to a local empty git repo to diff changes. One thing I noticed, was that Jekyll seems to leak variables between pages.

Here's code I had in my default layout:

```html
            {%if page.sidebar and page.sidebar != ""%}
              {%assign sidebarClass = page.sidebar%}
              {%capture sidebar%}
                {%include {{page.sidebar}}.html%}
              {%endcapture%}
            {%endif%}

            <!-- We're not including the sidebar directly, so the CRM can fill its contents too -->
            <aside class="sidebar {{sidebarClass}}">
              <!-- SIDEBAR_START -->
              {{sidebar}}
              <!-- SIDEBAR_STOP -->
            </aside>
```

Now some pages would set `sidebarInclude: about` to indicate they'd like to have the 'about sidebar' included by the default layout.

After upgrading to 4.0, pages that didn't want the about sidebar loaded, got one anyway, until I added these two lines at the top:

```html
            {%assign sidebarClass = ""%}
            {%assign sidebar = ""%}
```

It would appear that maybe caching makes variable scope leak over to different pages(?). This is what my very limited understanding from Jekyll/Ruby would lead me to believe, but it could be something else entirely.

ashmaroli:
- This is a 🐛 bug fix.
- *No unit-tests.*

## Summary

Since `Liquid::Templates` are now cached, ensure that every render starts with a clean context.

## Context

Resolves #7811

### DIFF
--- a/lib/jekyll/liquid_renderer/file.rb
+++ b/lib/jekyll/liquid_renderer/file.rb
@@ -18,6 +18,10 @@ module Jekyll
       end
 
       def render(*args)
+        # clear assigns to `Liquid::Template` instance prior to rendering since
+        # `Liquid::Template` instances are cached in Jekyll 4.
+        @template.instance_assigns.clear
+
         measure_time do
           measure_bytes do
             measure_counts do
--- a/lib/jekyll/liquid_renderer/file.rb
+++ b/lib/jekyll/liquid_renderer/file.rb
@@ -18,9 +18,7 @@ module Jekyll
       end
 
       def render(*args)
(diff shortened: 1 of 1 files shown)


DirtyF:
@jekyllbot: merge +fix

PR #7974, opened 2020-01-15
Remove post-install message from gemspec

ashmaroli:
The post-install message meant for Jekyll 4.0.0 is not necessary for the next release.


ashmaroli:
> Not sure why tests are failing

We have a couple of flaky tests that pass when re-run..

ashmaroli:
@jekyllbot: merge +dev

PR #7979, opened 2020-01-20
Fix YAML representation of `group_by` result

etcadinfinitum:

## Summary

There was a small syntax issue in the YAML example block for the `group_by` operator; one of the inner lists wasn't closed with a square bracket. In the interest of accuracy, I am submitting a fix.

## Context

No corresponding issues that I was able to find.


ashmaroli:
Thank you very much @etcadinfinitum 
@jekyllbot: merge +doc

PR #7981, opened 2020-01-22
Reduce iterations by using `ary.each_with_object`

ashmaroli:
- This is a 🐛 bug fix.

## Summary

- Replaces instances of `array.map.compact`, `array.map.reject`, `array.map.select` with `array.each_with_object(obj) { |item, result| result << item if condition }`
- Fixes some comments to correctly reflect the implementation.

### DIFF
--- a/lib/jekyll/drops/drop.rb
+++ b/lib/jekyll/drops/drop.rb
@@ -79,8 +79,7 @@ module Jekyll
         end
       end
 
-      # Generates a list of strings which correspond to content getter
-      # methods.
+      # Generates a list of strings which correspond to content getter methods.
       #
       # Returns an Array of strings which represent method-specific keys.
       def content_methods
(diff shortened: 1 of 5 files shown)


mattr-:
I don't consider this a bug fix. It doesn't fix anything. The code already works.

Is it safe to say that this is for allocation reduction instead? If so, do you have numbers?

ashmaroli:
Reduction in allocations is negligible..
```diff
--- master branch https://travis-ci.org/jekyll/jekyll/jobs/639383898
+++ PR branch     https://travis-ci.org/jekyll/jekyll/jobs/640489916

- Total allocated: 379.64 MB (3794027 objects)
+ Total allocated: 379.53 MB (3792145 objects)
  Total retained:  19.62 MB (104595 objects)
```

ashmaroli:
Micro-benchmarking doesn't show any significant improvements either .. :man_shrugging: :thinking:

PR #7985, opened 2020-01-28
Move --baseurl to build command options

aaronadamsTO:
It looks like `--baseurl` works on `build` as well as `serve`, possibly since this PR was merged:

https://github.com/jekyll/jekyll/pull/5135

I just tried it myself and sure enough, `jekyll build --baseurl /test` successfully transforms URLs in Jekyll 4.0.0.

As such, this PR moves Base URL to Build Command Options.

### DIFF
--- a/docs/_docs/configuration/options.md
+++ b/docs/_docs/configuration/options.md
@@ -310,6 +310,16 @@ class="flag">flags</code> (specified on the command-line) that control them.
         <p><code class="flag">--strict_front_matter</code></p>
       </td>
     </tr>
+    <tr class="setting">
+      <td>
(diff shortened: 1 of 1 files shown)


DirtyF:
@aaronadamsTO please note that you can perfectly run `jekyll serve --baseurl /test`

aaronadamsTO:
Agreed @DirtyF, and as per the doc page:

> the `serve` sub-command can accept any of the options for the `build` sub-command, which are then applied to the site build which occurs right before your site is served.

So per the current format of the doc, moving the command to the `build` section is done to indicate it works with both `build` and `serve`.

DirtyF:
@jekyllbot: merge +docs

PR #7987, opened 2020-01-30
Quicker categories for documents without superdirs

ashmaroli:
## Summary

When a document's special directory (e.g. `_posts` for a post) is located at the root of the site's `collections_dir`, its `categories_from_path` is always empty.

So it's wasteful to process an empty string to get an empty array eventually.


ashmaroli:
@jekyllbot: merge +dev

PR #7988, opened 2020-01-31
Return relative_url if site.url is an empty string

ashmaroli:
## Summary

Currently, when `site.config["url"]` is an empty string, `:absolute_url` proceeds to prepend the result from `:relative_url` with the empty string and then normalize the result via Addressable.

Instead, this PR proposes to return the result from `:relative_url` (which is already normalized via Addressable) in this scenario.

## Coverage

The expected result is already covered by a test:
https://github.com/jekyll/jekyll/blob/f3b4cad057a9632d4129d0f0529ec5bdbad8abce/test/test_filters.rb#L421-L428

### DIFF
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -44,10 +44,11 @@ module Jekyll
         return input if Addressable::URI.parse(input.to_s).absolute?
 
         site = @context.registers[:site]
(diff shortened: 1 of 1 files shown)


ashmaroli:
@jekyllbot merge +dev

ashmaroli:
@jekyllbot merge +dev

DirtyF:
@jekyllbot merge +dev

PR #7991, opened 2020-02-02
Generate drop content methods once per class

ashmaroli:
## Summary

Calling `self.class.instance_methods` per instance generates (and allocates) a new array of Symbols on each call.
This PR proposes to move the concerned logic into the drop's singleton class and then reference the resulting singleton class attribute in each class instance.


ashmaroli:
### Memory Profile Summary

```diff
--- master https://travis-ci.org/jekyll/jekyll/jobs/644965617
+++ PR     https://travis-ci.org/jekyll/jekyll/jobs/645161398

- Total allocated: 424.44 MB (4021894 objects)
+ Total allocated: 422.24 MB (4013791 objects)
- Total retained:  19.78 MB (107009 objects)
+ Total retained:  19.78 MB (107032 objects)

ashmaroli:
> Is the `Drop` class a public class?

Yes, `Jekyll::Drops::Drop` is a `public` class.

> could it be re-opened and have methods added to it

Adding methods to the `Jekyll::Drops::Drop` class via monkey-patching is redundant because the `#content_methods` explicitly filter out `Jekyll::Drops::Drop.instance_methods` on `master` as well.

However, a `Jekyll::Drops::Drop` subclass (e.g. `Jekyll::Drops::DocumentDrop`) could be monkey-patched by a plugin.. which would be broken by the proposed change here. :disappointed: 
Good Catch! @mattr-

PR #7996, opened 2020-02-06
Update 03-front-matter.md

dgolant:

## Summary

I've added a short clarifying statement in the getting started guide that I think will prevent StackOverflow questions like [these](https://stackoverflow.com/questions/29662793/using-liquid-tags-in-a-jekyll-page-not-working)


DirtyF:
@jekyllbot: merge +docs

PR #8008, opened 2020-02-15
Adds a 'precision' liquid filter to allow management of numbers

pallan:

## Summary

Adds a `precision` filter to liquid to allow users to define the precision of a number being displayed. This would allow users to pad numbers with trailing zeroes if needed.

## Context

There are no open issues that I could find. I needed this for a local project and did not see any way to do it so I created a custom plugin and thought I would contribute it upstream.

### DIFF
--- a/docs/_data/jekyll_filters.yml
+++ b/docs/_data/jekyll_filters.yml
@@ -274,6 +274,14 @@
 
 #
 
+- name: Precision
+  description: Define the precision of a number
+  examples:
+    - input: '{{ "1.2" | precision: 3 }}'
+      output: '1.200'
+
+#
+
 - name: Array Filters
   description: >-
     Push, pop, shift, and unshift elements from an Array.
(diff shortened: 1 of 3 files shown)


DirtyF:
We already have a [`round` Liquid filter](https://shopify.dev/docs/liquid/reference/filters/math-filters#round).

ashmaroli:
For the record, I gave my approval because, the `precision` filter formats a given *number-like string* or number with padding *as required*.
`{{ 4.5612 | round: 5 }}` is not equal to `{{ 4.5612 | precision: 5 }}`
`{{ 1.58 | round: 3 }}` is not equal to `{{ 1.58 | precision: 3 }}`
`EUR {{ 12 | round: 2 }}` is not equal to `EUR {{ 12 | precision: 2 }}`

@pallan Since there is conflict here, try submitting this filter to Liquid directly. If they accept, well and good. Jekyll will be able to use it as well. However if they reject, we'll *probably* reconsider..

mattr-:
I'd prefer this gets proposed for inclusion in Liquid as well. If they reject the filter, then please resubmit it here and we'll reconsider adding it. 

Thanks for contributing!

PR #8013, opened 2020-02-19
Optimize markdown parsing with Kramdown by reusing the options and parser objects

ashmaroli:
- Tested by existing coverage.

## Summary

`Kramdown::Document` is designed to set up the necessary options and a parser (based on the provided options) to initiate conversion of a Markdown content string.

But Jekyll uses the same set of options (and therefore the parser as well) to convert *all of its Markdown documents*. Therefore, having kramdown run the set up tasks for each document is wasteful.

This PR proposes to subclass `Kramdown::Document` and move the set up tasks from the `:initialize` method to a `singleton_menthod`. This approach has the set up tasks run for just the *first Markdown document contents* converted, cache the resulting objects and subsequently provide these objects to consequent conversions.

### DIFF
--- a/lib/jekyll/converters/markdown/kramdown_parser.rb
+++ b/lib/jekyll/converters/markdown/kramdown_parser.rb
@@ -15,10 +15,13 @@ module Kramdown
       # rubocop:disable Naming/MemoizedInstanceVariableName
       def setup(options)
         @cache ||= {}
-        @cache[:id] ||= options.hash
 
         # reset variables on a subsequent set up with a different options Hash
-        reset! unless @cache[:id] == options.hash
+        unless @cache[:id] == options.hash
+          @options = @parser = nil
+          @cache[:id] = options.hash
+        end
+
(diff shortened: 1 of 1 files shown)


ashmaroli:
### Memory Profile Summary

```diff
--- master https://travis-ci.org/jekyll/jekyll/jobs/652560205
+++ PR     https://travis-ci.org/jekyll/jekyll/jobs/652608908

- Total allocated: 396.41 MB (3625375 objects)
+ Total allocated: 387.31 MB (3599878 objects)
- Total retained:  19.80 MB (107263 objects)
+ Total retained:  19.80 MB (107275 objects)
```

mattr-:
I went and looked at the Kramdown code. It's too simple to require further changes. Would be nice to see something like this pushed upstream if they'll agree to it.

mattr-:
@jekyllbot: merge +minor

ashmaroli:
Awesome!! Thank you very much @mattr- :tada:

PR #8014, opened 2020-02-19
Update unit tests for Kramdown-based converter

ashmaroli:
## Summary

This is an attempt to reduce the flakiness of our `test/test_kramdown.rb`

### DIFF
--- a/test/test_kramdown.rb
+++ b/test/test_kramdown.rb
@@ -4,10 +4,24 @@ require "helper"
 require "rouge"
 
 class TestKramdown < JekyllUnitTest
+  def fixture_converter(config)
+    site = fixture_site(
+      Utils.deep_merge_hashes(
+        {
+          "markdown" => "kramdown",
+        },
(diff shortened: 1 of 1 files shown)


test/test_kramdown.rb
```diff
@@ -4,10 +4,24 @@
 require "rouge"
 
 class TestKramdown < JekyllUnitTest
+  def fixture_converter(config)
+    site = fixture_site(
+      Utils.deep_merge_hashes(
+        {
+          "markdown" => "kramdown"
```
DirtyF:
```suggestion
          "markdown" => "kramdown",
```

ashmaroli: I had already pushed the change. But GitHub never picked it up due to the partial outage..

ashmaroli:
Let's see if tests continue to pass (without reruns) on `master` as well..
@jekyllbot: merge +dev

PR #8015, opened 2020-02-20
Configure kramdown toc_levels as array by default

ashmaroli:
## Summary

When kramdown is given the option `toc_levels` as a string range (`"1..6"`), kramdown *parses* the string into an Array but in the process, generates *intermediate objects* of `Range`, `MatchData` and `Array`.
Since this process repeats for every *kramdown document* converted by kramdown, Jekyll can help *reduce* the work done by passing an Array *by default.*

*This is a backwards-compatible change.*


mattr-:
@jekyllbot: merge +dev

PR #8016, opened 2020-02-20
Add Visual Studio Code Development Container

MarkBennett:
## Summary

Add files to create a Visual Studio Code development container which can be used to quickly get new contributors up and building the app. To use this you need to have [Visual Studio Code](https://code.visualstudio.com/) and the [Remote Development Extension Pack](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.vscode-remote-extensionpack) installed.

The development environment includes `gcc`, `ruby`, and `nodejs`. I've verified that `script/cibuild` runs without errors. I've also added a section to the installation page describing how to setup VS Code to use this environment.

From start to finish setting up a new development box is entirely automatic and takes about 5 minutes. Once you're done you can open a new terminal in the development container using the same shortcut keys as you would when using VS Code normally.

The editor files are all in `/.devcontainter` directory which can be ignored by other editors and team members.

## Context

I haven't been able to find any issues related to this, but please add them if I missed them.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,43 @@
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
+
+FROM ruby:2
+
+# Avoid warnings by switching to noninteractive
+ENV DEBIAN_FRONTEND=noninteractive
+
+# This Dockerfile adds a non-root user with sudo access. Use the "remoteUser"
+# property in devcontainer.json to use it. On Linux, the container user's GID/UIDs
+# will be updated to match your local UID/GID (when using the dockerFile property).
+# See https://aka.ms/vscode-remote/containers/non-root-user for details.
+ARG USERNAME=vscode
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
+# Configure apt and install packages
+RUN apt-get update \
+    && apt-get -y install --no-install-recommends apt-utils dialog 2>&1 \
+    # Verify git, process tools installed
+    && apt-get -y install git openssh-client iproute2 procps lsb-release \
+    #
+    # Install ruby-debug-ide and debase
+    && gem install ruby-debug-ide \
+    && gem install debase \
+    #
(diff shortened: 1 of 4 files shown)


ashmaroli:
Technically, this is not a *documentation* change. It is an enhancement to the repository's development workflow.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,55 @@
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
+
+FROM ruby:2
```
mattr-: This is not specific enough. We don't support Ruby 2.7 yet, for example.

MarkBennett:
> Technically, this is not a _documentation_ change. It is an enhancement to the repository's development workflow.

Ok. I wasn't sure so flagged it as both. Thanks @ashmaroli.

MarkBennett:
> I'm not really sure how I feel about this. We provide zero help to people using other editors and I honestly wish we could decouple the actual code base from it's development environment and avoid the question of editor preference all together.

Fair point, on other projects I've worked on it's usually an all or nothing approach. I couldn't find anything specific about adding editor configuration in other issues, but it might be worth adding that to the docs for future if it's decided to keep them out.

Just sharing as I found it useful for me.

.devcontainer/Dockerfile
MarkBennett: Where do we specify the supported versions? Gemspec just lists > 2.4.

https://github.com/jekyll/jekyll/blob/master/jekyll.gemspec#L33

ashmaroli:
Please move the information added to the `docs/_docs/installation.md` to 
`docs/_docs/contributing.md` instead since that info has nothing to do with how Jekyll is to be installed / used.

MarkBennett:
> Please move the information added to the `docs/_docs/installation.md` to
> `docs/_docs/contributing.md` instead since that info has nothing to do with how Jekyll is to be installed / used.

Docs moved @ashmori. Thanks for the feedback.

.devcontainer/Dockerfile
mattr-: hmmm. maybe we just shouldn't worry about it then. :smiley:

mattr-:
@jekyllbot: merge +dev

MarkBennett:
Awesome! Thanks for the quick reviews @mattr- and @ashmaroli! :)

ashmaroli:
You're welcome @MarkBennett

PR #8017, opened 2020-02-21
docs: improve highlighting of code blocks

EricFromCanada:

## Summary

After setting the default highlight language to `plaintext` so my local build matches what GitHub Pages generates, I found and fixed some inconsistencies in how code blocks were being formatted and highlighted. In some places I changed which language a code block was highlighted as (e.g. `html` → `liquid`) to adjust which parts the reader should pay attention to. I also made a bunch of whitespace fixes. 

Each category of changes are grouped into separate commits, but I can squash if needed.


DirtyF:
@jekyllbot: merge +docs

PR #8021, opened 2020-02-24
remove leading slash from Sass file location

iamgeef:

## Summary

Removes the leading slash from the `First create a Sass file at /assets/css/styles.scss with the following content:` line

## Context

Paths that start with a slash denote the root of the file system, assuming the user is in the site directory the path shouldn't have a leading slash.
Previous steps that have a 'create a file' direction omit the leading slash.


ashmaroli:
Thank you @iamgeef 
@jekyllbot: merge +doc

PR #8023, opened 2020-02-24
Bypass disk-access to read or write scenario stats

ashmaroli:
## Summary

Experiment to see if this speeds up Cucumber Test suite on AppVeyor.


ashmaroli:
RESULT: The tests took *more than a minute* longer..
**`FAIL`**

PR #8026, opened 2020-02-25
[Docs] Fix asset link ref in step-by-step tutorial

dmiedev:
After creating posts in the step 8 of the step by step tutorial, Jekyll will look for `/YYYY/MM/DD/assets/css/styles.css` as `_layouts/post.html` requires the `default` layout which references `assets/css/styles.css`. This results in `ERROR '/YYYY/MM/DD/assets/css/styles.css' not found.`. 
This can be solved by this change.

## Summary
Replacement of the line in the code snippet of the step 7 of the step by step tutorial.

### DIFF
--- a/docs/_docs/step-by-step/07-assets.md
+++ b/docs/_docs/step-by-step/07-assets.md
@@ -70,7 +70,7 @@ Open `_layouts/default.html` and add the stylesheet to the `<head>`:
   <head>
     <meta charset="utf-8">
     <title>{{ page.title }}</title>
-    <link rel="stylesheet" href="assets/css/styles.css">
+    <link rel="stylesheet" href="/assets/css/styles.css">
   </head>
   <body>
     {% include navigation.html %}
--- a/docs/_docs/step-by-step/07-assets.md
+++ b/docs/_docs/step-by-step/07-assets.md
@@ -70,7 +70,7 @@ Open `_layouts/default.html` and add the stylesheet to the `<head>`:
   <head>
     <meta charset="utf-8">
     <title>{{ page.title }}</title>
-    <link rel="stylesheet" href="{{ 'assets/css/styles.css' | relative_url }}">
+    <link rel="stylesheet" href="/assets/css/styles.css">
   </head>
   <body>
     {% include navigation.html %}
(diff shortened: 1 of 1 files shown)


ashmaroli:
@dmiedev Instead of prefixing a slash, prefer using the `relative_url` filter:
```diff
- <link rel="stylesheet" href="assets/css/styles.css">
+ <link rel="stylesheet" href="{{ 'assets/css/styles.css' | relative_url }}">
```

dmiedev:
Explanation of the filter in the step may be needed?...

ashmaroli:
Good idea! That'd help novice users a lot!
Also, I'd like it if you could update the title of this PR to better reflect the changes in the PR.

dmiedev:
I'd like to get help with the documentation supplementation from someone who knows how to explain the usage of the filter because I'm still a newcomer to Jekyll.

ashmaroli:
@dmiedev Jekyll has two closely related *URLFilters*: `relative_url` and `absolute_url`.
- `relative_url` converts given string to a URL relative to the root of the generated site. It automatically incorporates the optionally configured `baseurl` of the current session. For example, if one has `baseurl: "/blog"` in their config file, then `{{ 'foo.html' | relative_url }}` will result in `"/blog/foo.html"`.
- `absolute_url` converts given string to an absolute URI based on the configured `url` and `baseurl` of the current session. For example, if one has both `url: "https://www.example.com"` and
`baseurl: "/blog"` in their config file, then `{{ 'foo.html' | absolute_url }}` will result in `"https://www.example.com/blog/foo.html"`.

*The URLFilters are capable of bit more. But that's for users to find out on their own..*

---

Feel free to understand the filter and adapt the relevant information into one suitable for the current pull request.

dmiedev:
@ashmaroli Could you please explain why the `relative_url` filter is preferable to `/`?

ashmaroli:
Thank you @dmiedev 
@jekyllbot: merge +doc

dmiedev:
@ashmaroli So the filter isn't needed? I'm confused

ashmaroli:
> So the filter isn't needed?

Perhaps recommending the filter to novice users in this tutorial wasn't such a nice idea. I decided to ship the basic fix instead of wasting time convincing to opt for the *better* fix.

dmiedev:
@ashmaroli Okay 
You wouldn't have to convince me, I asked because I wanted to understand the filter better. 
But maybe filter recommendation is really out of place.

dmiedev:
Noticed the issue was caused by #8021

ashmaroli:
Yes, I missed the faux pas during review..

PR #8029, opened 2020-02-27
Corrected command to modify PATH

mattpennathe3rd:

## Summary

The existing command to modify the path will not persist after closing and reopening Terminal. This new command should be permanent.


mattr-:
@jekyllbot: merge +docs

PR #8032, opened 2020-02-27
Docs: Render full contents of just the latest post

ashmaroli:

## Summary

The older posts are just rendered as references to their respective page where the full content is already rendered:

news-item-archive

This has two benefits:
- Jekyll doesn't have to render the full post contents twice &mdash; once on the index page and then again on the post's URL.
- Users need not scroll through the full contents of every post at both `/news/` and `/releases/`


mattr-:
@jekyllbot: merge +docs

PR #8042, opened 2020-03-02
Remove dev dependencies from new theme-gem gemspec

ashmaroli:

## Summary

There is no explicit dependency on Bundler or Rake while developing a theme-gem.
As Bundler is now included with Ruby itself, there wouldn't be a need for the mention in the gemspec going forwards.


DirtyF:
> As Bundler is now included with Ruby i

only from 2.6, we still support 2.5 right?

ashmaroli:
> we still support 2.5 right?

Right. And we still support 2.4 too.
By *Bundler is now included with Ruby..*, I meant that users on newer Ruby versions are going to be  in conflict with their default Bundler version and the numbers in the gemspec here.

DirtyF:
@jjekyll: merge +minor

PR #8044, opened 2020-03-04
Simplify Jekyll::Hooks.trigger logic

ashmaroli:
## Summary

- Use `Hash#dig` to traverse nested hashes and check if the result is `nil`.
- Guard against *empty* object as well so that we need not consequently duplicate it via `:sort_by`.


mattr-:
@jekyllbot: merge +dev

PR #8048, opened 2020-03-07
Non-deprecated `vendor/bundle` path configuration

riccardoporreca:

## Summary

The documented [Bundler configuration](https://jekyllrb.com/tutorials/using-jekyll-with-bundler/#configure-bundler) to install gems under `./vendor/bundle/` refers to a deprecated usage of `bundle install`

```sh
$ bundle install --path vendor/bundle
[DEPRECATED] The `--path` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set path 'vendor/bundle'`, and stop using this flag
```

### DIFF
--- a/docs/_tutorials/using-jekyll-with-bundler.md
+++ b/docs/_tutorials/using-jekyll-with-bundler.md
@@ -41,7 +41,7 @@ other gems on your system. If you skip this step, Bundler will install your
 dependencies globally on your system.
 
 ```sh
-bundle install --path vendor/bundle
+bundle config --local set path 'vendor/bundle'
 ```
 
 <div class="note info">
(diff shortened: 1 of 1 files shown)


DirtyF:
@jekyllbot: merge +docs

riccardoporreca:
@DirtyF, happy to contribute!

Just one note on `--local`: If omitted the configuration is not stored in a project-specific `./.bundle/config` as documented in the note, but in the global user-specific `~/.bundle/config`. I could have been more explicit about it in the PR.
```sh
$ bundle config --help
```
```
Executing bundle config set <name> <value> will set that configuration to the value specified for all bundles  executed  as
       the current user. The configuration will be stored in ~/.bundle/config. If name already is set, name will be overridden and
       user will be warned.
```

PR #8055, opened 2020-03-15
Remove version-constraint relaxation for i18n gem

ashmaroli:
## Summary

The relaxation was introduced via https://github.com/jekyll/jekyll/commit/1684905ec793be16ebb3ca4e81406aff5cbdbd10 to allow use of activesupport 4.2.10 on Ruby 2.2.

Since Rubies older than v2.4.0 are not relevant for Jekyll 4.0, the constraint-relaxation can be removed.


ashmaroli:
@jekyllbot: merge +dev

PR #8057, opened 2020-03-15
chore: simplify require for Jekyll::VERSION

DirtyF:
> the `require_relative` method comes handy, It will load a file relative to the current location. — https://piotrmurach.com/articles/writing-a-ruby-gem-specification/


DirtyF:
@jekyllbot: merge +dev

PR #8060, opened 2020-03-19
Update 09-collections.md

kluepfel:
Added hint to restart the server after the second configuration change

## Summary

Added a hint to restart the jekyll server after the second configuration change. I stopped the tutorial on one day, and picked it up the next, and this had me stumped for some time, until I remembered "wasn't there something about having to restart the server after you change the config?"

This should fix that.


DirtyF:
@jekyllbot: merge +docs

PR #8063, opened 2020-03-20
Remove extra paragraph tags

kieranbarker:

## Summary

I've removed the paragraph tags around `{{ post.excerpt }}`. This tag already outputs paragraph tags, so wrapping it in another set causes invalid HTML output.


DirtyF:
@jekyllbot: merge +docs

PR #8067, opened 2020-03-22
Reduce Pathname objects from front matter defaults

ashmaroli:
## Summary

Instead use simple String methods as alternative.
*This can be significant improvement for sites with numerous default front matter scopes in their config file.*


ashmaroli:
@jekyllbot: merge +dev

PR #8073, opened 2020-03-24
Improve documentation on usage of Jekyll commands

### LINKED ISSUE #8065: docs: Finding the switches for serve under "BUILD > Command line usage" seems unintuitive
## Motivation

The docs on the website for `jekyll serve` are hidden away under [configuration options](https://jekyllrb.com/docs/configuration/options/#serve-command-options) in the BUILD section. I found them by using the search tool.

Serving and building seem separate in my mind, even if one depends on the other. Hence, finding the switches for serving under a section for build's configuration options felt unintuitive.

## Suggestion

Call the section BUILD & SERVE as those are the two most likely options you'll use (`new` is already covered in the quick start), or perhaps COMMAND LINE USAGE and then have sections like `Overview` (which could [stay the same](https://jekyllrb.com/docs/usage/) yet its title would make more sense), `new`, `build` and `serve` with their more extended instructions and something like `More…` for the rest, or split those up too.

What I'm saying is, I'm stupid and need all the help I can get to find things, but that's a good test for documentation :)

Thanks for all the hard work on the project.

Regards,
iain

yb66:

docs: The aim is to show what the commands/options are without needing to run the cli client or even have Jekyll installed

## Summary

I've allowed a bit of duplication between the CLI section and the
 Configuration section by moving shared stuff into the includes dir.

The docs/more_usage.md could serve as a start for expanding on those options.

Also added a couple of missing config options to the tables.

## Context

I was running Jekyll in a Docker container and this would've been
easier for me than spinning up another container and running a
 shell to get them, but finding the CLI options needed a search.

resolves #8065


ashmaroli:
@yb66 Thank you very much for proposing to improve the documentation.
I've not looked into it deeply but would like to know if you could minimize the HTML table code by having a template render the info contained in a data file.

(No worries if you did try that route and faced difficulty..)

yb66:
@ashmaroli I didn't try templating it, I'll give that a go, probably tomorrow now.

Regards,
iain

DirtyF:
I don't see the added value of maintaining all those pages, when people can use `jekyll help`

yb66:
@DirtyF So I should spin up another Docker container just to get the help because you can't be bothered to document your own app?

Okay.

PR #8069, opened 2020-03-22
Use `layout.path` when rendering the Liquid layout

ashmaroli:
- This is a 🐛 bug fix.

## Summary

So that Jekyll's `LiquidRenderer` can correctly differentiate between layout from source vs layout from theme.

The bug can be identified in the `--profile` output:
```
| Filename                                     | Count |  Bytes |  Time |
+----------------------------------------------+-------+--------+-------+
| _posts/2020-03-20-welcome-to-jekyll.markdown |     1 |  1.68K | 0.181 |
| _layouts/default.html                        |     4 | 20.33K | 0.032 |
| minima-2.5.1/_includes/head.html             |     4 |  7.22K | 0.023 |
| feed.xml                                     |     1 |  4.00K | 0.007 |
| minima-2.5.1/_includes/header.html           |     4 |  3.96K | 0.004 |
| minima-2.5.1/_includes/footer.html           |     4 |  4.65K | 0.003 |
| _layouts/post.html                           |     1 |  2.45K | 0.003 |
| minima-2.5.1/_includes/social.html           |     4 |  1.56K | 0.002 |
| _layouts/home.html                           |     1 |  0.38K | 0.002 |
| _layouts/page.html                           |     1 |  0.67K | 0.000 |
+----------------------------------------------+-------+--------+-------+
| TOTAL (for 10 files)                         |    25 | 46.88K | 0.257 |
```
After fix:
```
| Filename                                     | Count |  Bytes |  Time |
+----------------------------------------------+-------+--------+-------+
| _posts/2020-03-22-welcome-to-jekyll.markdown |     1 |  1.68K | 0.150 |
| minima-2.5.1/_layouts/default.html           |     4 | 20.33K | 0.021 |
| minima-2.5.1/_includes/head.html             |     4 |  7.22K | 0.013 |
| feed.xml                                     |     1 |  4.00K | 0.012 |
| minima-2.5.1/_includes/header.html           |     4 |  3.96K | 0.004 |
| minima-2.5.1/_includes/footer.html           |     4 |  4.65K | 0.003 |
| minima-2.5.1/_includes/social.html           |     4 |  1.56K | 0.002 |
| minima-2.5.1/_layouts/post.html              |     1 |  2.45K | 0.002 |
| minima-2.5.1/_layouts/home.html              |     1 |  0.38K | 0.001 |
| minima-2.5.1/_layouts/page.html              |     1 |  0.67K | 0.000 |
+----------------------------------------------+-------+--------+-------+
[...]

PR #8075, opened 2020-03-25
Improve path normalization in liquid_renderer

ashmaroli:
- This is a 🐛 bug fix. 
- I've added tests

## Summary

Merging #8069 exposed a bug in `Jekyll::LiquidRenderer`. The table output when building with
`--profile` isn't robust to handle absolute paths of layouts from plugins (like `jekyll-redirect-from`).

So, this PR revamps the path normalization logic into a new private instance method.

### DIFF
--- a/lib/jekyll/liquid_renderer.rb
+++ b/lib/jekyll/liquid_renderer.rb
@@ -5,11 +5,6 @@ require_relative "liquid_renderer/table"
 
 module Jekyll
   class LiquidRenderer
-    extend Forwardable
-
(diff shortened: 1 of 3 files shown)


ashmaroli:
**Update:** *Introduced a `Theme#basename` getter so that `LiquidRenderer` need not allocate a new string for every file from a theme repository (when the Gemfile points to a local theme repo / remote theme repo on GitHub / when using the `jekyll-remote-theme` plugin).*

DirtyF:
@jekyll: merge +fix

PR #8081, opened 2020-03-29
Add default front matter for tutorials collection

ashmaroli:
## Summary

Replace repetition with front matter defaults


DirtyF:
@jekyllbot: merge +docs

PR #8083, opened 2020-03-30
Reduce array allocations from `StaticFile#path`

ashmaroli:
## Summary

`Jekyll::StaticFile#path` involves duplicating arrays via `Array#Compact`. So multiple calls to the unmemoized method would result in unnecessary Array allocations.

### DIFF
--- a/lib/jekyll/static_file.rb
+++ b/lib/jekyll/static_file.rb
@@ -40,11 +40,13 @@ module Jekyll
 
     # Returns source file path.
     def path
-      # Static file is from a collection inside custom collections directory
-      if !@collection.nil? && !@site.config["collections_dir"].empty?
-        File.join(*[@base, @site.config["collections_dir"], @dir, @name].compact)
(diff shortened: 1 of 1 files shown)


ashmaroli:
In theory, a resource's (irrespective of page, document, static-file, layout, etc) `path` shouldn't change. Additionally, for static-files, there is no equivalent writer (`:path=`). So, it can be assumed to be a *read-only* attribute.

If at all something is mutating the `path` attribute, then that code should be fixed.

ashmaroli:
Should I freeze the attribute value and see if something breaks / fails..?

DirtyF:
As you see fit, just want to make sure we don't ship any regression, or that we don't miss any test here 😅

ashmaroli:
*Note: I'll unfreeze the attribute value after some testing..*

DirtyF:
@jekyllbot: merge +fix

PR #8100, opened 2020-04-06
Debug reading Page and Layout objects

ashmaroli:
- This is a 🙋 feature or enhancement.

## Summary

Currently only `Document` objects are shown to be read when Jekyll builds with `--verbose` switch.
With this change, all classes that include the `Convertible` module (mainly `Page`, `Layout`), except `PageWithoutAFile`, will have their `relative_path` output when building with `--verbose` mode.

PR #8101, opened 2020-04-06
Remove leading slashes from relative paths

ashmaroli:
## Summary

Strip leading slashes from core primitive objects (pages / layouts / static files) consistently.

### DIFF
--- a/lib/jekyll/layout.rb
+++ b/lib/jekyll/layout.rb
@@ -42,7 +42,7 @@ module Jekyll
         @base_dir = site.source
         @path = site.in_source_dir(base, name)
       end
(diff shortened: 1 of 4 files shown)


ashmaroli:
@DirtyF We don't need this strictly. When I exposed the relative paths of layouts via #8100, I came across this minor inconsistency. Relative paths of pages and documents do not have a leading slash. But layouts and static files do.
Apparently, we can't strip the leading slash for static files until the next major version. So that is out of the window.

ashmaroli:
Currently not a breaking change because I've reverted the changes to static files.

PR #8104, opened 2020-04-09
Cache contents of included file read into memory

ashmaroli:
## Summary

Cache contents of included file read into memory.

This could result in some nice improvements in memory usage in sites:
-  that use `{% include_relative %}` to include a *Markdown page* &mdash; no need to re-read the Markdown file more than twice (Once by a `Reader` subclass and once more by `IncludeRelativeTag`).
- that use `{% include %}` to include *large* includes in multiple contexts.


ashmaroli:
Superseded by #8158

PR #8037, opened 2020-02-28
docs: improvements for note boxes

EricFromCanada:
This is a 🔦 documentation change.

## Summary

Notes currently are marked up either as a single paragraph with a [Kramdown IAL](https://kramdown.gettalong.org/syntax.html#block-ials) or as straight HTML if the note has a title or multiple paragraphs. I fixed the CSS to make their text styles match and converted the single-paragraph HTML notes to IAL form, and fixed several other inconsistencies. 

## Context

Mostly a sequel to #8017.

I also have on hand CSS and markup to allow converting titled or multi-paragraph notes to IAL form, if there's interest.

PR #8108, opened 2020-04-10
docs: convert note boxes and other HTML to Kramdown

EricFromCanada:

## Summary

This adds CSS that allows notes with titles to be generated with a [Kramdown IAL](https://kramdown.gettalong.org/syntax.html#block-ials) and converts the remaining notes from HTML, except for [this one](https://jekyllrb.com/docs/upgrading/3-to-4/) which contains a code block. This also converts any remaining HTML-based links and code/strong/em spans to Kramdown.

## Context

A follow-up to #8037.


DirtyF:
Thanks for cleaning things up, I feel like those notes could be includes to avoid using `strong` and `br` to mimick a note title and write CommonMark compatible Markdown.

`{% include note class="" title="" %}` (would have to test how this would impact build time)

ashmaroli:
@EricFromCanada Sometimes the line-lengths are short, sometimes long.
Could we stick to one length consistently..?

EricFromCanada:
@DirtyF The difficulty with switching to pure Markdown is its lack of support for any way to tag a block or span with attributes that can be turned into CSS classes, as Kramdown IALs currently do. (There is a proposal for a [common attribute syntax](https://talk.commonmark.org/t/consistent-attribute-syntax/272/137) extension for CommonMark out there, though.) So, to have an include generate a block for a note with the requisite classes on the opening tag, it'd need to either recieve the entire contents of the note as a parameter, or only return the opening tag (and optional title) and rely on the page calling it to supply the closing `</p>` or `</div>`.

@ashmaroli Were you thinking 80 characters per line, 100, or no hard line breaks within paragraphs?

ashmaroli:
> Were you thinking 80 characters per line, 100, or no hard line breaks within paragraphs?

I was going for consistency with the line-length of the said file in this PR.
In a future PR, we could have a site-wide realignment to be within 120 chars per line followed by hard line breaks.

PR #8112, opened 2020-04-13
Add details in warning message to reveal --future switch

dee-see:

## Summary

The warning that says "Skipping: POST has a future date" doesn't do anything to suggest that there's a way around that. `jekyll serve --help` does show that the `--future` switch exists, but I thought that adding it to the message would be helpful.

## Context

Every 6 months I write a post for a future date, run `bundle exec jekyll s` to verify it locally, see that my post isn't posted and wonder if there's a way around it. This would help my brain. ;)

Not sure if you're really interested in this, but throwing it out there just in case. Thanks!

### DIFF
--- a/lib/jekyll/readers/post_reader.rb
+++ b/lib/jekyll/readers/post_reader.rb
@@ -76,7 +76,7 @@ module Jekyll
     def publishable?(doc)
       site.publisher.publish?(doc).tap do |will_publish|
         if !will_publish && site.publisher.hidden_in_the_future?(doc)
-          Jekyll.logger.warn "Skipping:", "#{doc.relative_path} has a future date"
+          Jekyll.logger.warn "Skipping:", "#{doc.relative_path} has a future date, run jekyll with --future to publish anyway"
         end
       end
     end


lib/jekyll/readers/post_reader.rb
```diff
@@ -76,7 +76,7 @@ def processable?(doc)
     def publishable?(doc)
       site.publisher.publish?(doc).tap do |will_publish|
         if !will_publish && site.publisher.hidden_in_the_future?(doc)
-          Jekyll.logger.warn "Skipping:", "#{doc.relative_path} has a future date"
+          Jekyll.logger.warn "Skipping:", "#{doc.relative_path} has a future date, run jekyll with --future to publish anyway"
```
DirtyF, line 79:
```suggestion
          Jekyll.logger.warn "Skipping:", "#{doc.relative_path} has a future date. You can pass  --future option if you want to publish posts with a future date."
```

DirtyF: I like the idea of providing docs from the CLI, but here this might be too verbose and not a so common use case.

@jekyll/core Do we really need to enhance our warning here?

PR #8113, opened 2020-04-13
Config include trailing slash

### LINKED ISSUE #7951: Include configuration syntax for hidden folders
## Motivation

Hi, I've struggled (like a lot of persons it seems) with setting up github pages with shpinx. In particular, the way `include` works on folders/subfolders.

To solve the problem, I went to the documentation and looked how it was done in the default configuration [here](https://jekyllrb.com/docs/configuration/default/). And either the syntax is not the same for include or exclude, either there is a typo in the documentation:

To exclude a folder, it seems like one needs to write:
```
exclude: ["folder_name/"]
```
But it did not work to include a folder (specifically a folder starting with an underscore).
After some testing and looking around, I found that the syntax for  `include` seems to be:
```
include: ["_special_folder_name"]
```
 

## Suggestion

It is quite frustrating to have to guess that I find. Maybe an update of the documentation would be needed ?

A great way to improve the doc would be to give examples, at least for that case, but also maybe for and command/option (the default configuration is quite helpful for that, but it wasn't enough for me this time).

Thanks !

billykong:
This is a 🐛 bug fix.

- I've added tests

## Summary

`include` in the configuration file should include directory when trailing slash is present, e.g.  
```
# in _config.yml
include:
  - _pages/
```
should include the directory `_pages/` if exists.   
  
The current behavior in the `master` is that including `_pages` will include the directory while `_pages/` does not.  
  
This PR check that `_pages/` matches a directory named `_pages` but not file.

## Context

Resolves #7951

### DIFF
--- a/test/test_entry_filter.rb
+++ b/test/test_entry_filter.rb
@@ -151,11 +151,10 @@ class TestEntryFilter < JekyllUnitTest
     end
 
     should "match directory only if there is trailing slash" do
-      data = ["../_glob_include_test/", "_not_dir/"]
-      assert @filter.glob_include?(data, "../_glob_include_test")
-      assert @filter.glob_include?(data, "../_glob_include_test/include_me.txt")
-      assert !@filter.glob_include?(data, "../_not_dir")
+      data = ["../_glob_include_test/_is_dir/", "../_glob_include_test/_not_dir/"]
+      assert @filter.glob_include?(data, "../_glob_include_test/_is_dir")
+      assert @filter.glob_include?(data, "../_glob_include_test/_is_dir/include_me.txt")
+      assert !@filter.glob_include?(data, "../_glob_include_test/_not_dir")
     end
-
   end
 end
--- a/test/test_entry_filter.rb
+++ b/test/test_entry_filter.rb
@@ -151,10 +151,10 @@ class TestEntryFilter < JekyllUnitTest
     end
 
     should "match directory only if there is trailing slash" do
-      data = ["../_glob_include_test/_is_dir/", "../_glob_include_test/_not_dir/"]
-      assert @filter.glob_include?(data, "../_glob_include_test/_is_dir")
-      assert @filter.glob_include?(data, "../_glob_include_test/_is_dir/include_me.txt")
       assert !@filter.glob_include?(data, "../_glob_include_test/_not_dir")
+      data = ["../source/_glob_include_test/_is_dir/", "../source/_glob_include_test/_not_dir/"]
+      assert @filter.glob_include?(data, "../source/_glob_include_test/_is_dir")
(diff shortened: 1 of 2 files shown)


billykong:
Hi @DirtyF, is there any problem with this pull request that needs to be fixed or something I should do? It is my first time going through the pull request workflow in OSS. Please feel free to let me know if I missed something.

DirtyF:
@billykong nope, we're good, we just need to take the time to review your submission.

test/test_entry_filter.rb
```diff
@@ -149,5 +149,12 @@ class TestEntryFilter < JekyllUnitTest
       assert @filter.glob_include?(data, "vendor/bundle/jekyll/lib/page.rb")
       assert @filter.glob_include?(data, "/vendor/ruby/lib/set.rb")
     end
+
+    should "match directory only if there is trailing slash" do
+      data = ["../_glob_include_test/_is_dir/", "../_glob_include_test/_not_dir/"]
```
ashmaroli, line 154: our test site is in `test/source`. Please place the two `_glob_include_test` files in `test/source`.

test/test_entry_filter.rb
```diff
@@ -149,5 +149,12 @@ class TestEntryFilter < JekyllUnitTest
...
+      assert @filter.glob_include?(data, "../_glob_include_test/_is_dir")
+      assert @filter.glob_include?(data, "../_glob_include_test/_is_dir/include_me.txt")
+      assert !@filter.glob_include?(data, "../_glob_include_test/_not_dir")
```
ashmaroli, line 157: We prefer `refute something` instead of `assert !something`

billykong:
Thanks @DirtyF and @ashmaroli

ashmaroli:
@billykong Can you test separately if the issue you're resolving here can be reproduced with Jekyll 3.8.6..? (Want to see if this is a regression in Jekyll 4.0)..
Thanks.

billykong:
> can be reproduced with Jekyll 3.8.6..?

Hi @ashmaroli , it can be reproduced in Jekyll 3.8.6

billykong:
Tested with Jekyll 3.8.6.  
  
With Trailing slash directory:  
jekyll-3 8 6-with-trailing-slash-marked  
The `_pages` directory is not generated.
  
Without Trailing slash directory:  
jekyll-3 8 6-without-trailing-slash-marked
The `_pages` directory is generated.

ashmaroli:
Thanks for testing with Jekyll 3.8.6 @billykong, So it is certain now that the bug existed prior to Jekyll 4.0 and Jekyll 3.8.6 :+1:

ashmaroli:
Thanks @billykong 
@jekyllbot: merge +bug

PR #8123, opened 2020-04-15
Added zlib1g-dev and sudo

bongiozzo:
gem update doesn't succeed without zlib1g-dev


docs/_docs/installation/windows.md
```diff
@@ -56,19 +56,19 @@ which hosts optimized versions of Ruby for Ubuntu.
...
 ```sh
-gem update
+sudo gem update
 ```
 
 Now all that is left to do is install Jekyll.
 
 ```sh
-gem install jekyll bundler
+sudo gem install jekyll bundler
```
yg, line 71: Is `sudo` is needed?

PR #8131, opened 2020-04-23
Allow disabling import of theme configuration

### LINKED ISSUE #8114: bug: variables used in theme configuration are showing up when not overridden
## My Environment

| Operating System |    ruby v2.5.0 - Gentoo        |
| `jekyll`         |  4.0.0    |
| `github-pages`   | N/A     |

---

## Expected Behaviour

The information added under _config.yml was the only information jekyll used. 

## Current Behavior

Having default variables filled, under the theme configuration file, results in those items showing up, if the user does not override them with their own.

### Step 1. Update `Gemfile` to include jekyll v4, and minima latest master
```
gem "jekyll", ">= 4.0.0"
gem "minima", git: "https://github.com/jekyll/minima"
```
### Step 2. Update `_config.yml` with new minima settings:
```
 minima:
   social_links:
     mastodon:
       - username: myusernamehere
         instance: mastodon.social
```
### Step 3. 
```
rm Gemfile.lock
bundle install && bundle update 
bundle exec jekyll serve --watch
```
### Result: 
Github, & Twitter show up, even though we only specified Mastodon.

image

## Regression Testing:
This does not happen when used with `jekyll v3.8.6`

ashmaroli:
- I've added tests.

## Summary

So that related bugs are not activated out-of-the-box.. 

## Context

Resolves #8114

### DIFF
--- a/features/theme_configuration.feature
+++ b/features/theme_configuration.feature
@@ -11,6 +11,17 @@ Feature: Bundling Config file with Theme gems
     And the _site directory should exist
     And I should see "aero" in "_site/index.html"
 
+  Scenario: Disabling import of theme configuration entirely
+    Given I have a configuration file with:
+      | key                 | value      |
+      | theme               | test-theme |
+      | ignore_theme_config | true       |
+    And I have an "index.md" page that contains "{{ site.test_theme.skin }}"
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should not see "aero" in "_site/index.html"
+
   Scenario: A pre-configured theme with valid config file overriding Jekyll defaults
     Given I have a configuration file with "theme" set to "test-theme"
     And I have an "index.md" page that contains "{{ site.baseurl }}"
--- a/features/theme_configuration.feature
+++ b/features/theme_configuration.feature
@@ -3,16 +3,24 @@ Feature: Bundling Config file with Theme gems
   I want to be able to pre-configure my gemified theme
   In order to make it easier for other Jekyllites to use my theme
 
-  Scenario: Enabling import of theme configuration
-    Given I have a configuration file with:
-      | key                 | value      |
-      | theme               | test-theme |
-      | import_theme_config | true       |
+  Scenario: Easy onboarding with a pre-configured theme
+    Given I have a configuration file with "theme" set to "test-theme"
     And I have an "index.md" page that contains "{{ site.test_theme.skin }}"
     When I run jekyll build
     Then I should get a zero exit status
     And the _site directory should exist
     And I should see "aero" in "_site/index.html"
+
+  Scenario: Disabling import of theme configuration entirely
+    Given I have a configuration file with:
+      | key                 | value      |
+      | theme               | test-theme |
+      | ignore_theme_config | true       |
+    And I have an "index.md" page that contains "{{ site.test_theme.skin }}"
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should not see "aero" in "_site/index.html"
 
   Scenario: A pre-configured theme with valid config file overriding Jekyll defaults
     Given I have a configuration file with "theme" set to "test-theme"
--- a/features/theme_configuration.feature
+++ b/features/theme_configuration.feature
@@ -3,24 +3,16 @@ Feature: Bundling Config file with Theme gems
   I want to be able to pre-configure my gemified theme
   In order to make it easier for other Jekyllites to use my theme
 
-  Scenario: Easy onboarding with a pre-configured theme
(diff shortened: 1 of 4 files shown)


DirtyF:
Can you explain how you see a theme config play with a site config?

I would expect the theme to have his own config, my site to have its config, and my site's config to override the theme's config if I want to override some theme's config. 
A clear convention before adding options.

ashmaroli:
Yes, that *was exactly* the intended workflow.
Only that are some unforeseen stuff that comes into play.
For example, say a user's theme's config resolves to:
```
{"foo"=>{"alpha"=>"beta"}}
```
And their site config includes the following pair:
```
{"foo"=>{"lorem"=>"ipsum"}}
```
The user would expect `"foo"` to just have a single attribute (`"lorem"`). But, in reality, the resulting config has `"foo"` with two attributes (`"alpha"` and `"lorem"`). The user has to then try and  workaround this by explicitly setting `foo.alpha => nil`.
For a feature meant to ease onboarding, this additional overhead does not bode well.

Another unforeseen issue is that the resulting config (after overriding theme-config with user-end config) is a Ruby `Hash` instance rather than a `Jekyll::Configuration` (subclass of Ruby `Hash`) instance. So effectively, the query `site.config.is_a?(Jekyll::Configuration)` is `false` which was an unintended breaking change and another v4.0 regression.

The easiest route out I saw was to remove the feature entirely since it is just one-version old. But I guess SemVer principles expects us to carry the baggage until a v5.0. So the next route would be to either provide an option to disable this *non-essential feature* or add more technical debt (to buoy a *non-essential feature*) by patching the bugs as they are reported.

DirtyF:
When there's no overlap between the theme and the site's config, do we get a `Jekyll::Configuration`?

ashmaroli:
> When there's no overlap between the theme and the site's config

We get a `Jekyll::Configuration` if nothing is done irrespective of overlap. 
Consider the following lines on `master`:
https://github.com/jekyll/jekyll/blob/18f04c61adc8ebfa116361e527f7d167928b737f/lib/jekyll/site.rb#L434-L436
The argument `config` to the above private method is *always* an `Jekyll::Configuration` instance.
<br/>
https://github.com/jekyll/jekyll/blob/18f04c61adc8ebfa116361e527f7d167928b737f/lib/jekyll/site.rb#L441
The result from `SafeYAML.load_file` is always a `Hash`
<br/>
https://github.com/jekyll/jekyll/blob/18f04c61adc8ebfa116361e527f7d167928b737f/lib/jekyll/site.rb#L450
And finally `Hash` is returned by the utility method
https://github.com/jekyll/jekyll/blob/18f04c61adc8ebfa116361e527f7d167928b737f/lib/jekyll/utils.rb#L28-L30
https://github.com/jekyll/jekyll/blob/18f04c61adc8ebfa116361e527f7d167928b737f/lib/jekyll/utils.rb#L41-L47

lib/jekyll/site.rb
```diff
@@ -432,6 +432,8 @@ def collections_path
     private
 
     def load_theme_configuration(config)
+      return config unless config["import_theme_config"] == true
```
mattr-, line 435: The logic here seemed a bit backwards and the variable name used here seems wrong based on the docs.

```suggestion
      return config if config["ignore_theme_config"] == true
```

features/theme_configuration.feature
```diff
@@ -3,8 +3,11 @@ Feature: Bundling Config file with Theme gems
   I want to be able to pre-configure my gemified theme
   In order to make it easier for other Jekyllites to use my theme
 
-  Scenario: Easy onboarding with a pre-configured theme
-    Given I have a configuration file with "theme" set to "test-theme"
+  Scenario: Enabling import of theme configuration
+    Given I have a configuration file with:
+      | key                 | value      |
+      | theme               | test-theme |
+      | import_theme_config | true       |
```
mattr-, line 10:
```suggestion
      | ignore_theme_config | true       |
```

But this will break this test. It seems like we'd want to keep this the way it is and write a new test for ignoring the theme configuration.

ashmaroli:
Thanks for the review @mattr- 
I started this off with an *opt-out* stand and switched to an *opt-in* in https://github.com/jekyll/jekyll/pull/8131/commits/a6f932e52e1755177b46e107a2ae32bbd177180e But I forgot to update the documentation.
Either ways, which stand would be the most apt for Jekyll 4.1 and Jekyll 4.0.1..? *Opt-in* or *opt-out*?

mattr-:
Opt out is the least surprising from my point of view, as the feature was already enabled. We don't have numbers but I imagine the number of people that run into this is low enough that we can just point them to the docs when they bring it up and that's good enough for now. we can always revisit the default in 5.0

ashmaroli:
@jekyllbot: merge +minor

PR #8135, opened 2020-04-24
Added note about OS specific installation instructions.

aredshaw:


aredshaw:
I spent hour trying to install Jekyll on my Ubuntu system because I followed the quickstart page. I then went through the troubleshooting instructions and then stackoverflow. I wish I knew that installing on Ubuntu was different than what the quickstart instructions mentioned.

DirtyF:
@jekyll: merge +docs

PR #8138, opened 2020-04-25
Escape regex characters in paths to match

### LINKED ISSUE #8137: The tests in test_convertible.rb trying to match the file paths in the output will fail if path contains special characters
Hi there,

I'm part of the team which packages jekyll for Debian. In our case the path where the source is contains a `+` sign. So `assert_match(%r!#{File.join(@base, name)}!, out)` in `test/test_new_command.rb` fails. because the plus sign is not escaped. The solution is using `Regexp.escape()` as shown in this patch:

<https://salsa.debian.org/ruby-team/jekyll/-/blob/master/debian/patches/0022-escape-regex-characters-in-match.patch>

dleidert:
Our paths contain the + sign, which if not escaped changes the match regex
and the tests fail.

Fixes #8137

### DIFF
--- a/test/test_convertible.rb
+++ b/test/test_convertible.rb
@@ -32,7 +32,7 @@ class TestConvertible < JekyllUnitTest
         assert_equal({}, ret)
       end
       assert_match(%r!YAML Exception!, out)
-      assert_match(%r!#{File.join(@base, name)}!, out)
+      assert_match(%r!#{Regexp.escape(File.join(@base, name))}!, out)
     end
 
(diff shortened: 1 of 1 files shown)


dleidert:
Hm. The issue on appveyor is not related to this patch.

ashmaroli:
@dleidert We've stopped testing with Ruby 2.4 on AppVeyor on our `master`. If you could *rebase* to the latest master, the tests should run fine.
Thanks.

ashmaroli:
Thank you @dleidert 
@jekyllbot: merge +dev

PR #8140, opened 2020-04-25
Use platforms instead of install_if

### LINKED ISSUE #8139: The created Gemfile (jekyll new) should use platforms instead if install_if
Hi there,

I'm part of the Debian maintainers for Jekyll running Debian Linux. When I run `jekyll new foo/ --skip-bundle`, move into the created directory and run `jekyll b` I get an error that the gems, which are for Win/MingW/JRuby platforms (tz* and wdm) cannot be found although they are not even necessary on my platform. I had a discussions with bundler's upstream author in <https://lists.debian.org/debian-ruby/2020/04/msg00062.html> about it and it seems better suited to replace the `install_if` by `platforms` in the created `Gemfile`. I propose the patch in <https://salsa.debian.org/ruby-team/jekyll/-/blob/master/debian/patches/0024-use-platforms-in-default-gemfile.patch>.

dleidert:
With :install_if bundler will check for the gems even if they are not required
on our platform. To prevent users from having this issue use :platforms
instead: https://lists.debian.org/debian-ruby/2020/04/msg00062.html

fixes #8139

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -82,7 +82,7 @@ group :jekyll_optional_dependencies do
 
   # Windows and JRuby does not include zoneinfo files, so bundle the tzinfo-data gem
   # and associated library
-  install_if -> { RUBY_PLATFORM =~ %r!mingw|mswin|java! } do
+  platform :jruby, :mswin, :mingw, :x64_mingw do
     gem "tzinfo", "~> 1.2"
     gem "tzinfo-data"
   end
--- a/Gemfile
+++ b/Gemfile
@@ -74,7 +74,7 @@ group :jekyll_optional_dependencies do
   gem "rdoc", "~> 6.0"
   gem "tomlrb", "~> 1.2"
 
-  platform :ruby, :mswin, :mingw, :x64_mingw do
+  platforms :ruby, :mswin, :mingw, :x64_mingw do
     gem "classifier-reborn", "~> 2.2"
     gem "liquid-c", "~> 4.0"
     gem "yajl-ruby", "~> 1.4"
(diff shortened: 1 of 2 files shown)


ashmaroli:
@dleidert We also have a similar usage in our dev Gemfile. Could you patch those lines as well..?
Thanks.
https://github.com/jekyll/jekyll/blob/2bc1e04620ce22154ef423acba223eecdc379f13/Gemfile#L83-L88

Gemfile
```diff
@@ -82,7 +82,7 @@ group :jekyll_optional_dependencies do
 
   # Windows and JRuby does not include zoneinfo files, so bundle the tzinfo-data gem
   # and associated library
-  install_if -> { RUBY_PLATFORM =~ %r!mingw|mswin|java! } do
+  platform :jruby, :mswin, :mingw, :x64_mingw do
```
DirtyF, line 85:
```suggestion
  platforms :jruby, :mswin, :mingw, :x64_mingw do
```

DirtyF: Is it platform or platforms?

dleidert: I just copied it from the section above in Gemfile (line 77). It seems both are used widely, but the manual page lists it as `platforms`.

dleidert: Pushed another commit

ashmaroli:
@DirtyF Do you think this patch qualifies to be backported for v4.0.x..?

DirtyF:
@ashmaroli Not if we release 4.1 sooner than later.

DirtyF:
@jekyll: merge +fix

PR #8144, opened 2020-04-27
Remove `gem install bundler` from quick-start

searls:
Since Bundler has shipped with Ruby for a while, I think the copy-paste snippet that Jekyll shares with folks visiting the home page should probably not include a bundler install step, especially considering how fraught it can be to manage multiple bundler versions (post-Bundler 2) for more novice Rubyists.

### DIFF
--- a/docs/pages/index.html
+++ b/docs/pages/index.html
@@ -46,7 +46,7 @@ permalink: /
         <p class="line">
           <span class="path">~</span>
           <span class="prompt">$</span>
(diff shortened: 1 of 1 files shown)


DirtyF:
Bundler is a part of Ruby’s standard library since Ruby 2.6, Ruby 2.4 has just gone EOL, we'll need to keep mentioning  installing bundler until 2.6 is the minimal require version.

PR #8150, opened 2020-04-30
Clarify `bundle config` in Bundler tutorial

mkasberg:

## Summary

* Add `--local` option to the `bundle config set` command in the tutorial.
* Clarify what the config command does and why you might want to use it.
* Clarify the differences between this tutorial and the Quickstart.

## Context

I originally wrote this Bundler tutorial back in 2018. At that time, running
`bundle install --path vendor/bundle` was the correct way to install
gems to a custom directory with Bundler. Since then, Bundler released
[version 2.1](https://github.com/rubygems/bundler/blob/master/CHANGELOG.md#210pre1-august-28-2019),
which deprecates breaking changes for Bundler 3. One of these changes is
the deprecation of the `install --path` option, which has been replaced
by `bundle config set path foo`. The tutorial was updated for this in #8048.

This PR builds upon clarifies that change. We use `--local` to avoid
system-wide user preference changes (which is implied by the info box
beneath). In addition, I wanted to clarify the reasons you might want to
follow this approach, after reviewing this tutorial with a fresh set of
eyes.

### DIFF
--- a/docs/_tutorials/using-jekyll-with-bundler.md
+++ b/docs/_tutorials/using-jekyll-with-bundler.md
@@ -53,7 +53,7 @@ gems in the `./vendor/bundle/` project subdirectory. The advantage of doing this
 is that bundler will install gems within your project folder instead of the
 location used by `gem install`. This can help you avoid permissions errors you
 might otherwise get during gem installation, depending how you installed Ruby.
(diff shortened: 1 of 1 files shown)


DirtyF:
@jekyll: merge +docs

PR #8158, opened 2020-05-04
Initialize include-files as Jekyll objects

ashmaroli:
## Summary

This PR optimizes the processing of *include files*.

Currently, include files are *located and read into memory* for each call to the tag's `render` method. Though the parsed include templates are cached, they're cached to a given Liquid context. So, the file warrants a re-parsing when the context changes.

With the proposed changes, include files are initialized into new `Jekyll::Inclusion` core objects. The actual file-contents and its parsing are deferred until necessary (during the `render` calls) and subsequently cached until next regeneration.
To maintain backwards-compatibility with Ruby API, the rendering is now handled by a new (temporary) subclass of `Tags::IncludeTag`.

*Does not need documentation updates or tests since there are no discernible changes to either Ruby API or API exposed for Liquid templates*.


DirtyF:
Manual testing does not show any build time improvement on our docs.

ashmaroli:
### Memory profile (on AppVeyor)
```diff
--- master https://ci.appveyor.com/project/jekyll/jekyll/builds/32641386/job/nvd9jag59n0u2y3b
+++ PR     https://ci.appveyor.com/project/jekyll/jekyll/builds/32685467/job/vexmf42xsrltwa0p

- Total allocated: 357.98 MB (2873621 objects)
- Total retained:  19.76 MB (109571 objects)
+ Total allocated: 351.86 MB (2853141 objects)
+ Total retained:  19.78 MB (109651 objects)
```

DirtyF:
I'll defer to @mattr- if it's worth the "future" breaking change, hard to evaluate the gain from the only memory consumption results on large sites.

ashmaroli:
> the "future" breaking change

Technically, it need not progress into a breaking change if we or the maintainers at the time don't want to, but I placed the comment to deter plugin authors to have a plausible *less-messy* option.

lib/jekyll/inclusion.rb
```diff
@@ -0,0 +1,60 @@
...
+      when site.theme && %r!#{site.theme.includes_path}/(?<rel_path>.*)!o
+        PathManager.join dir_path, Regexp.last_match[:rel_path]
+      when %r!#{site.in_source_dir(site.config["includes_dir"], "/")}(?<rel_path>.*)!o
+        PathManager.join dir_path, Regexp.last_match[:rel_path]
+      else
+        path.sub(site.in_source_dir("/"), "")
+      end
+    end
+    # rubocop:enable Metrics/AbcSize
+  end
```
mattr-, line 59: `compute_relative_path` as a concept is more important than `dir_path`. I'd prefer to see `compute_relative_path` ordered above `dir_path`. I struggled to understand why `dir_path` was necessary since it wasn't used until we got to `compute_relative_path`

lib/jekyll/inclusion.rb
```diff
@@ -0,0 +1,60 @@
...
+      "#{self.class} #{relative_path.inspect}"
+    end
+    alias_method :to_s, :inspect
+
+    private
+
+    attr_reader :site
+
+    def dir_path
+      if site.theme && path.start_with?(site.theme.includes_path)
```
mattr-, line 40: Is there a better way to express this concept? Why is the theme important? Why is the include path important? What purpose does having this code here serve vs. it being somewhere else? Is there a better place to put this?

ashmaroli: > Why is the theme important?

There are two possible sources to load an include file from: `site.source` and `site.theme.root`.

If a site is using a theme, there's a 50% chance that an include file is loaded from the theme. Include files in such situations would have a long path e.g. `/users/jo/blog/vendor/bundle/ruby/2.4.0/gems/my_theme-0.1.0/_includes/header.html`.

**Conventionally**, the `relative_path` in such situations would be `path` without the preceding
`/users/jo/blog/vendor/bundle/ruby/2.4.0/gems/my_theme-0.1.0/`. At this point, the relative path would be identical to the `relative_path` for an include file from `site.source`.

To eliminate this confusion, I decided to tag include files from the theme with the theme's dirname. Consequently, the above include file will now have a `relative-path` &mdash;
`my_theme-0.1.0/_includes/header.html` &mdash; instantly discernible when calling `#inspect`

> Why is the include path important?

If we know that `path` starts with `"#{site.theme.root}/_includes", then `dir_path` is without a doubt, `<theme-gem-dirname>/_includes`. Instead of using String substitution to remove the unneeded substrings and allocate duplicates for multiple include files from the theme, we just use a cached string object via `PathManager`.

> Is there a better place to put this?

IMHO, placing it here is best so that it is `private`. Elsewhere it would need to be `public` and would carry overhead for future changes.

lib/jekyll/inclusion.rb
```diff
@@ -0,0 +1,60 @@
...
+  class Inclusion
+    attr_reader :name, :path
+
+    def initialize(site, base, name)
+      @site = site
+      @name = name
+      @path = PathManager.join(base, name)
+    end
+
+    def relative_path
```
mattr-, line 13: I don't see where this is used outside this class and suggest that we make it private.

lib/jekyll/inclusion.rb
mattr-: You took my questions a bit too literally, I believe. In fact, the first question is the most important one:

> Is there a better way to express this concept?

I asked this question because these private methods are the only places where we're worried about handling a theme. In cases like these, I always find it's good to step back and wonder "is there something special that requires these special cases here or have I missed another abstraction?"

tl;dr: the special casing we have here for themes is a smell and I'd like to encourage you to find a way to eliminate the smell with another abstraction. That would probably be best done in a separate PR that would then put this PR on hold.

ashmaroli:
@mattr- I've dropped the idea of using a `relative_path` for these instances entirely.

ashmaroli:
@jekyllbot: merge +dev

PR #8164, opened 2020-05-07
Add console command to the Jekyll CLI

jaredcwhite:

## Summary

This adds the ability to run `jekyll console` or `jekyll c` and get an IRB session where you can inspect and interact with the loaded `site` object at the point in time before it would render out the site files.

## Context

This has been extraordinarily helpful for me while debugging websites,  plugins, and internal APIs as I've worked on my fork of Jekyll, so I'm hoping it will be useful for the community at large. I'm not sure if there's a way to write a unit test for this command since it's interactive, but if you have any ideas please let me know!


DirtyF:
Thanks Jared, this is a great idea, my first question is should this be a plugin or not? 🤔

mattr-:
I don't think this should be a plugin.

lib/jekyll/commands/console.rb
```diff
@@ -0,0 +1,57 @@
...
+        # TODO: is there a way to add a unit test for this command?
+        # rubocop:disable Style/GlobalVars, Metrics/AbcSize, Metrics/MethodLength
+        def process(options)
+          Jekyll.logger.info "Starting:", "Jekyll v#{Jekyll::VERSION.magenta}" \
+                                      " console…"
+          Jekyll.logger.info "Environment:", Jekyll.env.cyan
+          site = Jekyll::Site.new(configuration_from_options(options))
+          site.reset
+          site.read
+          site.generate
```
ashmaroli, line 33: I feel the user would like to know that some of the build phases have already been run..
For example,
```
Reading files... done!
Running generators... done!
```
In large sites, there will be a significant time delay between `site.reset` and `site.render`. So, we don't want the use reporting back to us that *the console takes a long time to boot..*

That said, IMHO, I'd like an option to start the console with just `Jekyll::Site.new(config)` executed so that I can manually run `site.read` and `site.generate` etc at a later stage..

lib/jekyll/commands/console.rb
```diff
@@ -0,0 +1,57 @@
...
+
+          $JEKYLL_SITE = site
```
ashmaroli, line 35: Is this global variable absolutely necessary?
Could `site` be passed into the `workspace` instead?

lib/jekyll/commands/console.rb
```diff
@@ -0,0 +1,57 @@
...
+          irb = IRB::Irb.new(workspace)
+          IRB.conf[:MAIN_CONTEXT] = irb.context
+          eval("site = $JEKYLL_SITE", workspace.binding, __FILE__, __LINE__)
+          Jekyll.logger.info "Console:", "Now loaded as " + "site".cyan + " variable."
+
+          trap("SIGINT") do
+            irb.signal_handle
+          end
+
+          begin
```
ashmaroli, line 47: Why a `begin..end` block if there's no `rescue` / `ensure`..?

docs/_docs/usage.md
```diff
@@ -11,6 +11,9 @@ You can use this command in a number of ways:
 * `jekyll new PATH --blank` - Creates a new blank Jekyll site scaffold at specified path.
 * `jekyll build` or `jekyll b` - Performs a one off build your site to `./_site` (by default)
 * `jekyll serve` or `jekyll s` - Builds your site any time a source file changes and serves it locally
+* `jekyll console` or `jekyll c` - Opens up an IRB console and lets you
```
ashmaroli, line 14: Looks like this line can be extended bit more in sync with surrounding ones.

lib/jekyll/commands/console.rb
jaredcwhite: Great question—I tried a bunch of things to avoid having to use a global variable and wasn't successful. The challenge is it's setting up the REPL to run within the top-level `main` context, which I think is what we want, but once there I don't have access to any previously-defined local variables like `site`. Any suggestions?

lib/jekyll/commands/console.rb
jaredcwhite: Build phase notifications sounds good. For an option to skip those and only do the new/config, any thoughts? `--skip-build` or `--skip-load` or something like that?

ashmaroli: I like `--blank`.

lib/jekyll/commands/console.rb
ashmaroli: To be honest, I don't have much experience with setting up REPL environments.
Will have to dig into the Ruby docs..

jaredcwhite:
Just pushed up some of the changes suggested by @ashmaroli

parkr:
This is really neat!

PR #8165, opened 2020-05-08
Provide invokables for common drop query keys

ashmaroli:
## Summary

While playing around with profilers, it noticed that **memory allocations** are greater when a `Drop` queried with a key doesn't have a method defined for the key and is therefore forwarded to query `fallback_data`.

So, this PR provides commonly queried keys as methods defined for the drops.

### DIFF
--- a/lib/jekyll/drops/document_drop.rb
+++ b/lib/jekyll/drops/document_drop.rb
@@ -64,6 +64,18 @@ module Jekyll
           result[key] = doc[key] unless NESTED_OBJECT_FIELD_BLACKLIST.include?(key)
         end
       end
+
+      def title
+        @obj.data["title"]
+      end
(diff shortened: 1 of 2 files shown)


ashmaroli:
### Memory Profile

```diff
--- master https://travis-ci.org/github/jekyll/jekyll/jobs/684126964
+++ PR     https://travis-ci.org/github/jekyll/jekyll/jobs/684634377

- Total allocated: 298.04 MB (3049942 objects)
- Total retained:  19.40 MB (108644 objects)
+ Total allocated: 296.24 MB (3004828 objects)
+ Total retained:  19.40 MB (108643 objects)
```

ashmaroli:
@jekyllbot: merge +dev

PR #8168, opened 2020-05-09
chore(ci): cache dependencies

DirtyF:

## Summary

Test caching bundler dependencies to speed up GitHub Actions builds.


DirtyF:
@jekyll: merge +dev

ashmaroli:
@DirtyF This is breaking the action-builds on Windows platform (and now on `master` too).. Looks like the caching mechanism isn't very smart especially since we do not check `Gemfile.lock` into version control..

I vote to revert this change..

PR #8170, opened 2020-05-10
Site: Filter through just the docs collection

ashmaroli:

## Summary

`{{ site.documents }}` returns *all of the documents* in the `site`.
However, include files `_includes/docs_contents.html` and `_includes/docs_contents_mobile.html` are just used to render the navigation for documents in the `docs` collection.

In other words, it is wasteful having to loop through the entire set of *documents* when the current scope requires just a subset.


DirtyF:
@jekyll: merge +site

PR #8171, opened 2020-05-10
Add find filters to optimize where-first chains

ashmaroli:
- I've added tests.

## Summary

Introducing Liquid filters `find` and `find_exp` to optimize scenarios involving `first` chained to `where` and `where_exp` filter respectively.

### Purpose

`where` (`where_exp`) evaluates given conditions against *all items* in a given array and chaining the `first` filter returns a single object from the pool. This could be wasteful when the array is large with numerous items and the first winner is near the starting end.

i.e. say the array is `["apple", "lemonade", "spade", "sunglasses", "umbrella", "cola"]`
`{{ array | where: "type", "drink" | first }}` and `{{ array | find: "type", "drink" }}` are equivalent and return the same object (`"lemonade"`) but the latter stops processing after two items.

### DIFF
--- a/benchmark/find-filter-vs-where-first-filters.rb
+++ b/benchmark/find-filter-vs-where-first-filters.rb
@@ -1,0 +1,49 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require 'benchmark/ips'
+require_relative '../lib/jekyll'
+
+puts ''
+print 'Setting up... '
+
+SITE = Jekyll::Site.new(
+  Jekyll.configuration({
+    "source"             => File.expand_path("../docs", __dir__),
+    "destination"        => File.expand_path("../docs/_site", __dir__),
+    "disable_disk_cache" => true,
+    "quiet"              => true,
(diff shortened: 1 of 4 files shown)


DirtyF:
Is that something that already exists in Liquid? What's the expected gain?

ashmaroli:
> Is that something that already exists in Liquid?

As of now, there is no `find` filter in upstream Liquid.

> What's the expected gain?

For the general user-base? Depends on how it is used.
For Jekyll's docs site, profiles for `PR#8172` should shed some light.

ashmaroli:
I've added a benchmark script to compare the performance of the `find` filter vs `where` + `first` filters to illustrate the possible boost.

ashmaroli:
> If liquid don't have it, why should we?

Liquid has a different implementation of `where` filter from ours. So even if they were to add a `find` filter, it would be based on their version of `where`. That could have a different outcome for our users.

DirtyF:
@jekyll: merge +minor

PR #8177, opened 2020-05-14
Check if entry is a directory once per enumerator

ashmaroli:
## Summary

Because testing `File.directory?(entry)` for every `pattern` in `enumerator` (`site.exclude` and `site.include`) is redundant and can hamper performance.

Therefore, test `entry` once per enumerator and stash the result in a local variable which is tallied against `enumerator`'s items when needed.


ashmaroli:
@jekyllbot: merge +fix

Teja7048:
yes

PR #8178, opened 2020-05-14
Ease discovery of CLI commands (in their entirety)

dotnetCarpenter:

## Summary

After spending hours trying to figure out a solution to my issue, I stumbled upon [a SO answer](https://stackoverflow.com/a/48292461/205696) that mentions the `--config` flag, which is paramount to solving my use-case.

Then I found it is hidden away in the documentation, as the second section under *Configuration Options*, that starts out with explaining options in `_config.yml`.

I expected all CLI commands, options and arguments to be documented under *Command Line Usage* or via `jekyll --help`, like most other terminal programs. In lieu of that I have tried my best to ease the discovery of the CLI commands (in their entirety).

It is unfortunate that CLI usage is split in the documentation between CLI usage and `_config.yml` options but to fix that, I think a bigger restructure of the docs are necessary.

### DIFF
--- a/docs/_docs/usage.md
+++ b/docs/_docs/usage.md
@@ -5,17 +5,22 @@ permalink: /docs/usage/
 
 The Jekyll gem makes a `jekyll` executable available to you in your terminal.
 
-You can use this command in a number of ways:
+The `jekyll` program has several commands but the structure is always:
+
+```
+jekyll command [argument] [option]
+```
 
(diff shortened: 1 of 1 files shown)


DirtyF:
@jekyll: merge +docs

dotnetCarpenter:
@DirtyF Thanks! You are so fast. I made #8179 because I thought that more discussion was required. But I am very happy with this quick fix and I think it will help a lot of others as well.

I'm gonna close #8179

Teja7048:
yes

PR #8188, opened 2020-05-18
Initialize static files' data hash only if needed

ashmaroli:
## Summary

Instead of initializing a static file's `data` hash at the time the static file is read-in, initialize the `data` hash at a later time if required.

*This can improve performance of sites with numerous static files and default front matter in the config file.*


DirtyF:
@jekyll: merge +minor

PR #8196, opened 2020-05-20
Improve documentation on tags and categories

### LINKED ISSUE #8195: docs:
## Motivation

I think the documentation is overall in low quality for beginners as it is rather short and lacks a lot of information. Especially the tags and category section leaves a lot of questions. It just states 

"difference between categories and tags is a category can be part of the URL for a post whereas a tag cannot."

But it seems like the category has to be part or will be part of the url instead of can be and it still doesnt explain the overall difference at all. It still doesnt help me to understand how tags ans categories work and when to use them. It seems like Categories can be used to generate a site where all posts in a category are listed but theres no information on the advantage or disadvantage of tags. As a former WP user this is really confusing

## Suggestion

Extend and Clarify

ashmaroli:

## Summary

Rewrite the entire section by introducing tags first and then introducing categories as an extended feature-set of tags.

## Context

Closes #8195

### DIFF
--- a/docs/_docs/posts.md
+++ b/docs/_docs/posts.md
@@ -120,34 +120,34 @@ you wish to access the currently-rendering page/posts's variables (the
 variables of the post/page that has the `for` loop in it), use the `page`
 variable instead.
 
-## Categories and Tags
+## Tags and Categories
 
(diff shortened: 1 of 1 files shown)


DirtyF:
@jekyll: merge +docs

PR #8201, opened 2020-05-22
Build docs site with GitHub Actions

iBug:
This is essentially a rework of #8126.

Key points and notable features:

- Use `GITHUB_TOKEN` provided by GitHub Actions for authentication
    - Reference: [GitHub Help article on `GITHUB_TOKEN`][1]
    - There used to (before February 2020) be an issue where the above token cannot trigger GitHub Pages builds. This is no longer the case. ([Reference][2])
- Cache installation from Bundler
    - The cache depends on both `Gemfile` and `jekyll.gemspec` from the repository root. As Jekyll is always in rapid development, I chose to always run `bundle install`, regardless of whether there's a cache hit.
    - Use a global environment variable to specify Ruby version ([Reference][5])
- Generate prettier commit messages. Example:

    ```text
    Deploy docs from 0123456789abcdef0123456789abcdef

    iBug: Build docs site with GitHub Actions
    ```

- Good Bash script and Git practices, e.g.
    - Well-delimited variables (`${VAR}` instead of `$VAR`)
    - Quoted strings (`cmd "$var"` instead of `cmd $var`)
    - Use `git push <remote> +<refspec>` instead of `git push -f`
        - See [this Stack Overflow answer][3] and the comments below it

**Note**: Please add a commit to the `gh-pages` branch *before* merging this PR. As explained in the commit message of e8eb528e4aa6c7783565ff16c5e64cd15958f80a, GitHub Pages currently has a bug where it won't deploy properly on branches containing only one single commit. I am working around this issue by adding commits on top of the existing branch in many of my projects, and they're all working smoothly, so I'm carrying the same technique to this PR.

---

On a side note, previous failures from #8126 are (most likely) caused by [an ongoing incident of GitHub][4]. Specifically:

> We have identified the source causing elevated errors as well as **occasional stale data** on GitHub.com. We are working on remediation.

As old `GITHUB_TOKEN`s are supplied to Actions runs (they expire 60 minutes after creation), `git push` failed to authenticate. This should not happen normally so it'll be settled automatically after GitHub recovers.

---

Please feel free to share any concerns you have with my implementation. I'll be eager to answer them.

  [1]: https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token
  [2]: https://github.community/t5/GitHub-Actions/Github-action-not-triggering-gh-pages-upon-push/m-p/31283/highlight/true#M751
  [3]: https://stackoverflow.com/a/39389793/5958455
  [4]: https://www.githubstatus.com/incidents/6tcfpztf6j9m
  [5]: https://github.community/t5/GitHub-Actions/How-to-set-and-access-a-Workflow-variable/m-p/53516/highlight/true#M8860

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -5,6 +5,9 @@ on:
     branches:
       - master
 
+env:
+  RUBY_VERSION: 2.7
+
 jobs:
   deploy_docs:
     if: "!contains(github.event.commits[0].message, '[ci skip]')"
@@ -13,12 +16,22 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-ruby@v1
         with:
-          ruby-version: 2.7
+          ruby-version: ${{ env.RUBY_VERSION }}
+      - name: Setup cache for Bundler
+        id: cache
+        uses: actions/cache@v2
+        with:
+          path: vendor/bundle
+          key: ${{ runner.os }}-bundler-${{ env.RUBY_VERSION }}-${{ hashFiles('Gemfile') }}-${{ hashFiles('jekyll.gemspec') }}
+          restore-keys:
+            - ${{ runner.os }}-bundler-${{ env.RUBY_VERSION }}-${{ hashFiles('Gemfile') }}-
+            - ${{ runner.os }}-bundler-${{ env.RUBY_VERSION }}-
       - name: Set up dependencies
         run: |
           gem update --system --no-document
           gem update bundler --no-document
-          bundle install --jobs 4 --retry 3
+          bundle install --path=vendor/bundle --jobs 4 --retry 3
+          bundle clean
       - name: Build site
         run: bundle exec jekyll build --source docs --destination docs/_site --verbose --trace
         env:
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -35,7 +35,8 @@ jobs:
       - name: Build site
         run: bundle exec jekyll build --source docs --destination docs/_site --verbose --trace
         env:
-          JEKYLL_PAT: ${{ secrets.GITHUB_TOKEN }}
+          # For jekyll-github-metadata
+          JEKYLL_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: Deploy to GitHub Pages
         run: |
           SOURCE_COMMIT="$(git log -1 --pretty="%an: %B" "$GITHUB_SHA")"
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -28,8 +28,6 @@ jobs:
             - ${{ runner.os }}-bundler-${{ env.RUBY_VERSION }}-
       - name: Set up dependencies
         run: |
-          gem update --system --no-document
-          gem update bundler --no-document
           bundle install --path=vendor/bundle --jobs 4 --retry 3
           bundle clean
       - name: Clone target branch
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -38,7 +38,9 @@ jobs:
           REMOTE_REPO="https://${GITHUB_ACTOR}:${{ secrets.GITHUB_TOKEN }}@github.com/${GITHUB_REPOSITORY}.git"
 
           echo "Publishing to ${GITHUB_REPOSITORY} on branch ${REMOTE_BRANCH}"
-          git clone --depth=1 --branch="${REMOTE_BRANCH}" --single-branch --no-checkout "${REMOTE_REPO}"
+          rm -rf docs/_site/
+          git clone --depth=1 --branch="${REMOTE_BRANCH}" --single-branch --no-checkout \
+            "${REMOTE_REPO}" docs/_site/
       - name: Build site
         run: bundle exec jekyll build --source docs --destination docs/_site --verbose --trace
         env:
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,45 @@
+name: Build and deploy Jekyll documentation site
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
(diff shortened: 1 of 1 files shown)


iBug:
CC @ashmaroli @parkr

MichaelCurrin:
This workflow is heavy. It would be abstracted in a simple workflow like this one:

https://github.com/BryanSchuetz/jekyll-deploy-gh-pages#example

As recommended by DirtyF in another thread https://github.com/github/pages-gem/issues/651#issuecomment-541442887

Or see this approach which is what the Jekyll docs recommend.

https://jekyllrb.com/docs/continuous-integration/github-actions/#setting-up-the-action

I added the article by the way and my demo listed at the end works now using `GITHUB_TOKEN` not the `JEKYLL_PAT` as in the tutorial.

MichaelCurrin:
Oh I see in the original PR 

> (Use an in-house module instead of third-party action for greater control over the process)

What specific control is needed that can't be done with the third party ones and options or additional build steps?

ashmaroli:
> What specific control is needed that can't be done with the third party ones

Third-party services are always tailored for the generic userbase.
That is why there are always provisions to customize a procedure.

Having a completely in-house procedure for something as simple as setting up a shell script and a YAML config file will allow
control over the overall log output and commit messages.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Set up dependencies
+        run: |
+          gem update --system --no-document
+          gem update bundler --no-document
+          bundle install --path=vendor/bundle --jobs 4 --retry 3
+          bundle clean
+      - name: Build site
+        run: bundle exec jekyll build --source docs --destination docs/_site --verbose --trace
+        env:
+          JEKYLL_PAT: ${{ secrets.GITHUB_TOKEN }}
```
MichaelCurrin, line 38: What's the significance of this line? There is no commit steps here like in `REMOTE_REPO` setup below.

Does this have to do with `github-pages` plugin and getting GH metadata? I know that plugin uses `GITHUB_TOKEN` and gives build warnings locally if that is not set, so wondering if `JEKYLL_PAT` works for that?

MichaelCurrin:
> > What specific control is needed that can't be done with the third party ones
> 
> Third-party services are always tailored for the generic userbase.
> That is why there are always provisions to customize a procedure.
> 
> Having a completely in-house procedure for something as simple as setting up a shell script and a YAML config file will allow
> control over the overall log output and commit messages.

Thanks I am onboard with this approach then. There is more control, but at the cost of having to maintain boilerplate instead of having it maintained in a separate action (which means less control and a different kind of risk of commands going out of date).

.github/workflows/docs.yml
iBug: Looks like I mixed things up. I can't recall when the knowledge of `JEKYLL_PAT` emerged in my head, but I just went through stuff again and am now confident it's a mistake. I've changed this into `JEKYLL_GITHUB_TOKEN` which is wanted by the `jekyll-github-metadata` plugin.

iBug:
I've got some new updates, namely issues with GitHub Pages deployment. Please kindly check the commit message of e8eb528e4aa6c7783565ff16c5e64cd15958f80a and the PR description, which I've updated to explain what I'm doing.

MichaelCurrin:
@iBug Thanks for updating the token as suggested. The other commits and message make sense.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,62 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-ruby@v1
+        with:
+          ruby-version: ${{ env.RUBY_VERSION }}
+      - name: Setup cache for Bundler
+        id: cache
+        uses: actions/cache@v2
+        with:
+          path: vendor/bundle
+          key: ${{ runner.os }}-bundler-${{ env.RUBY_VERSION }}-${{ hashFiles('Gemfile') }}-${{ hashFiles('jekyll.gemspec') }}
```
mattr-, line 25: I think the `Gemfile.lock` file would be the better thing to hash here. Dependencies can be updated without updating the Gemfile, and I'd feel like we'd want to invalidate the cache in those cases. Maybe this doesn't matter for building the site though. I'm fine if we want to leave it as is.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,62 @@
...
+          restore-keys:
+            - ${{ runner.os }}-bundler-${{ env.RUBY_VERSION }}-${{ hashFiles('Gemfile') }}-
+            - ${{ runner.os }}-bundler-${{ env.RUBY_VERSION }}-
+      - name: Set up dependencies
+        run: |
+          gem update --system --no-document
+          gem update bundler --no-document
```
mattr-, line 32: I don't see a reason for these for the site build. Removing them will also make the build faster.

mattr- (changes requested):
I'd like to see the `gem` invokations go away before this gets merged.

mattr- (approved):
LGTM.

Very cool to see us use this for site building. :tada:

DirtyF:
@jekyll: merge +docs

iBug:
@DirtyF Hi, in case you missed it, my first message included this:

> **Note**: Please add a commit to the `gh-pages` branch *before* merging this PR.

You may want to create the `gh-pages` branch (an empty commit works) before re-running [this failed first job](https://github.com/jekyll/jekyll/runs/1387006936?check_suite_focus=true).

DirtyF:
@iBug so the job is running now, but it's still running Jekyll 3.9.0. Can't deactive GiHub Pages in the repository settings with a gh-pages branch 🤷

iBug:
@DirtyF Are you sure it's still 3.9? The presence of a `.nojekyll` on branch `gh-pages` should skip the GH part of building (again). You can switch the Pages source from master branch to gh-pages branch in the repository settings.

DirtyF:
I fixed it, jekyllrb.com is now bult with current master, thanks @iBug 👍

PR #8203, opened 2020-05-23
allow sites to use plugins stored in theme's plugins_dir

orchardcc:

## Summary

A remote theme may have custom plugins stored in the `plugins_dir`. However, those custom theme plugins are currently not being added as a dependency so sites can't make use of those custom theme plugins. 

## Context

Fixes benbalter/jekyll-remote-theme#75

### DIFF
--- a/lib/jekyll/plugin_manager.rb
+++ b/lib/jekyll/plugin_manager.rb
@@ -95,10 +95,11 @@ module Jekyll
     #
     # Returns an Array of plugin search paths
     def plugins_path
+      theme_plugins_path = site.in_theme_dir(Jekyll::Configuration::DEFAULTS["plugins_dir"])
       if site.config["plugins_dir"].eql? Jekyll::Configuration::DEFAULTS["plugins_dir"]
-        [site.in_source_dir(site.config["plugins_dir"])]
+        [site.in_source_dir(site.config["plugins_dir"]), theme_plugins_path].compact
       else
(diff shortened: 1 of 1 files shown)


ashmaroli:
Sorry @orchardcc, this feature will ***never be implemented in the Core*** for security reasons.
If you wan't to use plugins in a theme repository, you'll have to *manually copy them over* to your site.
Or if that can't be done, ask the theme's author to gemify those Ruby plugins and add the most necessary (ones that are absolutely needed for the theme to work for all users) *as the theme's runtime dependencies*.

orchardcc:
Greetings @ashmaroli, thanks for the prompt reply. I assume you mean security reasons with respect to GitHub Pages? From my understanding, the function `require_plugin_files` of the `plugins_manager.rb` file only adds files in the `plugins_dir` if safe mode is set to false. And since GitHub Pages builds and deploys sites with safe mode set to true, none of the plugins in the `plugins_dir` would be added. This change is for those who build sites locally (where safe mode may be false) and then push the built site to the repo or web server.

ashmaroli:
Security reasons can refer to any environment the Jekyll site is built &mdash; GitHub Pages, Docker container, local system, whatever.
Providing the means to load arbitrary Ruby code in a theme-gem (which is meant to be a vehicle for HTML / Liquid templates) is not a responsible move.
I already gave you an option of gemifying the plugin code and having that gem added as runtime dependency of the theme. That's as far Jekyll Core will be going regarding this feature.

PR #8213, opened 2020-05-28
Add atom:link rel="self" to RSS feed template

kslstn:
The [W3C validator](https://validator.w3.org/feed/check.cgi) says:
> This feed is valid, but interoperability with the widest range of feed readers could be improved by implementing the following recommendations.

With the added line, this should be fixed.

### DIFF
--- a/docs/_tutorials/convert-site-to-jekyll.md
+++ b/docs/_tutorials/convert-site-to-jekyll.md
@@ -437,6 +437,7 @@ layout: null
     <channel>
         <title>{{ site.title }}</title>
         <link>{{ site.url }}</link>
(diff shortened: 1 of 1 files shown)


ashmaroli:
@kslstn You do realize that you're updating a tutorial, don't you?
If you were not intending on fixing a real-world scenario, feel free to ignore this comment.

kslstn:
Yeah, since the template seems perfect except for that missing line, I thought it’d be useful to add.

DirtyF:
note: most of the time, you'd want to use jekyll-feed plugin to insert a link to your feed

PR #8220, opened 2020-05-31
test/test_site.rb: Changed absolute path "/tmp" to call to temp_dir()

noproblema:
…ir() to pass test on Windows

This is a 🐛 bug fix.

## Summary

test/test_site.rb uses absolute path "/tmp" and compares that with the output of DirectoryHelpers.temp_dir(). That makes two tests to fail in windows.

## Context

I have not been able to find any issue

### DIFF
--- a/test/test_site.rb
+++ b/test/test_site.rb
@@ -31,16 +31,16 @@ class TestSite < JekyllUnitTest
     end
 
     should "have an array for plugins if passed as a string" do
-      site = Site.new(site_configuration("plugins_dir" => "/tmp/plugins"))
+      site = Site.new(site_configuration("plugins_dir" => temp_dir("plugins")))
       array = [temp_dir("plugins")]
       assert_equal array, site.plugins
     end
 
     should "have an array for plugins if passed as an array" do
-      site = Site.new(site_configuration(
-                        "plugins_dir" => ["/tmp/plugins", "/tmp/otherplugins"]
-                      ))
       array = [temp_dir("plugins"), temp_dir("otherplugins")]
+      site = Site.new(site_configuration(
+                        "plugins_dir" => array
+                      ))
       assert_equal array, site.plugins
     end
 


ashmaroli:
Thank you for the suggestion, but it looks like you misunderstood the test here.
The test-site is configured with a relative path string like a user would do. The helper method `temp_dir()` adapts the result for Unix-based and Windows.

noproblema:
Hi Ashwin. Thanks for your reply. Perhaps you are right and I misunderstood the test, but still these tests will always fail in windows. 

The reason is that they check the assertion "/tmp/something" == temp_dir("something"). That will be true in linux but not in windows.There is not a relative path there. Is an absolute path: "/tmp", that is the same path returned by temp_dir() in linux.

The failure occurs in two tests:
"have an array for plugins if passed as a string" and "have an array for plugins if passed as an array" that check if a plugins path is correctly processed both when passed as a string or as array. In the two cases the paths provided are using temp_dir() but the result is checked against the absolute path "/tmp"

Best






On Sunday, May 31, 2020, 3:43:10 PM GMT+2, Ashwin Maroli <notifications@github.com> wrote: 







Thank you for the suggestion, but it looks like you misunderstood the test here.
The test-site is configured with a relative path string like a user would do. The helper method temp_dir() adapts the result for Unix-based and Windows.

ashmaroli:
> but still these tests will always fail in windows

Are you developing on Windows? If yes, have you tried running the test instead of speculating that they will fail..?
We run the tests on Linux and Windows via Travis CI and AppVeyor CI respectively. So if the tests were to ever fail, we'd have caught them.

> check the assertion `"/tmp/something" == temp_dir("something")`

This is why I said you misunderstood the test.
The two places you've changed `"/tmp/something"` are lines that are being passed to `Jekyll::Site.new`. There's no point in passing `D:/tmp/something` to that constructor and testing the correct result.
Assertion in the test is in later steps, and it always involves the use of helper method `temp_dir`

noproblema:
Yes, I have run the tests when testing other issue. That is the reason that I found them.

That is the result of running script/test on a unmodified clone of master
==========================================================
Finished tests in 180.360901s, 5.0122 tests/s, 8.9432 assertions/s.


Failure:
TestSite#test_: configuring sites should have an array for plugins if passed as an array.  [/home/yo/work/jekyll-noproblema/test/test_site.rb:44]
+++ actual
@@ -1 +1 @@
-["/home/yo/work/jekyll-noproblema/tmp/plugins", "/home/yo/work/jekyll-noproblema/tmp/otherplugins"]          
+["/tmp/plugins", "/tmp/otherplugins"]


Failure:
TestSite#test_: configuring sites should have an array for plugins if passed as a string.  [/home/yo/work/jekyll-noproblema/test/test_site.rb:36]
+++ actual
@@ -1 +1 @@
-["/home/yo/work/jekyll-noproblema/tmp/plugins"]
+["/tmp/plugins"]


904 tests, 1613 assertions, 2 failures, 0 errors, 0 skips
===========================================================

As you an see the tests fails because are testing equality between

1st case ["/home/yo/work/jekyll-noproblema/tmp/plugins", "/home/yo/work/jekyll-noproblema/tmp/otherplugins"] AND ["/tmp/plugins", "/tmp/otherplugins"]
 
2nd case ["/home/yo/work/jekyll-noproblema/tmp/plugins"] AND ["/tmp/plugins"]

I don't understand how your tests for windows can pass.

noproblema:
Sorry for the big letters. I think that markdown did it for me.

noproblema:
That is bash in windows. But I think that in windows the result will be the same.

ashmaroli:
Here's how it runs at my end on Windows `cmd.exe`:
```
bash script/test test/test_site.rb

+ ruby -S bundle exec ruby -I test test/test_site.rb
# -------------------------------------------------------------
# SPECS AND TESTS ARE RUNNING WITH WARNINGS OFF.
# SEE: https://github.com/Shopify/liquid/issues/730
# SEE: https://github.com/jekyll/jekyll/issues/4719
# -------------------------------------------------------------

# Running tests with run options --seed 6953:

......................................................

Finished tests in 99.417755s, 0.5432 tests/s, 0.9354 assertions/s.


54 tests, 93 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for Unit Tests to ... 2683 / 3908 LOC (68.65%) covered.

real    1m57.309s
user    0m0.015s
sys     0m0.000s
```

ashmaroli:
@noproblema Could you run the following (either directly via `irb` or indirectly via a script file) in your bash on windows?
```ruby
require 'jekyll'
puts Jekyll::VERSION
puts Jekyll::Utils::Platforms.windows?
```
It should output `true` in either case.

noproblema:
There it is:


~/work/jekyll-noproblema$ bundle exec irb
irb(main):001:0> require 'jekyll'
=> true
irb(main):002:0> puts Jekyll::VERSION
4.1.0
=> nil
irb(main):003:0> puts Jekyll::Utils::Platforms.windows?
26
=> nil



On Tuesday, June 2, 2020, 6:14:08 PM GMT+2, Ashwin Maroli <notifications@github.com> wrote: 







@noproblema Could you run the following (either directly via irb or indirectly via a script file) in your bash on windows?

require 'jekyll'
puts Jekyll::VERSION
puts Jekyll::Utils::Platforms.windows?

It should output true in either case.

ashmaroli:
I see.. and what does running `File.read("/proc/version")` in `irb` return?

noproblema:
irb(main):002:0> File.read("/proc/version")
=> "Linux version 4.4.0-17763-Microsoft (Microsoft@Microsoft.com) (gcc version 5.4.0 (GCC) ) #864-Microsoft Thu Nov 07 15:22:00 PST 2019\n"






On Tuesday, June 2, 2020, 8:51:17 PM GMT+2, Ashwin Maroli <notifications@github.com> wrote: 







I see.. and what does running File.read("/proc/version") in irb return?

PR #8222, opened 2020-06-01
Disable page excerpts by default

### LINKED ISSUE #8216: page.excerpt behaviour changed between 4.0.1 and 4.1.0
Hello, version 4.1.0 introduced an unexpected change in behaviour for `page.excerpt` per [pull request 7642](https://github.com/jekyll/jekyll/pull/7642).

I was using page.excerpt as a meta description entry:

`<meta name="description" content="{% if page.excerpt %}{{ page.excerpt | strip_html | strip_newlines | truncate: 160 }}{% else %}{{ site.description }}{% endif %}">`

In an example of an excerpt not being defined for a page, this would have evaluated to the site.description, but now I'm getting markup generated for me that breaks because of HTML being included.

I would recommend the new implementation approach things similarly by stripping html, stripping newlines, and even maybe truncating.

ashmaroli:
- This is a 🐛 bug fix.
- I've added tests.

## Summary

Jekyll 4.1.0 changed `{{ page.excerpt }}` for `Page` objects from `nil`.
This patch allows to opt-in via a config setting.

## Context

Resolves #8216

### DIFF
--- a/docs/_docs/pages.md
+++ b/docs/_docs/pages.md
@@ -35,6 +35,11 @@ If you have a lot of pages, you can organize them into subfolders. The same subf
 
 You might want to have a particular folder structure for your source files that changes for the built site. With [permalinks](/docs/permalinks) you have full control of the output URL.
 
+## Excerpts for pages
+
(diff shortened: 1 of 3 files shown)


ashmaroli:
@jekyllbot: merge +bug

PR #8230, opened 2020-06-06
Fix special treatment for 'page 1' in docs of pagination

fleaz:

## Summary
The example code in the documentation has a section for the corner case of page 1 because Jekyll won't build a pagination URL for page 1. I'm not sure why the cretor used `previous_page_path` for page 1 but this results in wrong hrefs in the pagination. If you are e.g. on page 5 the link behind the 1 would be to page 4. Simply using `href="/"` for page 1 solves this issue afaik.

### DIFF
--- a/docs/_docs/pagination.md
+++ b/docs/_docs/pagination.md
@@ -154,7 +154,7 @@ page with links to all but the current page.
     {% if page == paginator.page %}
       <em>{{ page }}</em>
     {% elsif page == 1 %}
-      <a href="{{ paginator.previous_page_path | relative_url }}">{{ page }}</a>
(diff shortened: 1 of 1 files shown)


ashmaroli:
@fleaz As much as your argument making sense to me, I suggest that you keep the `relative_url` intact to be in sync with rest of the code.
```
<a href="{{ '/' | relative_url }}">{{ page }}</a>
```
is a better change because it is *baseurl-aware*..

fleaz:
This is understandable. Updated my PR accordingly :)

ashmaroli:
Thanks @fleaz 
@jekyllbot: merge +doc

PR #8234, opened 2020-06-07
Don't generate excerpts for non-html pages

ashmaroli:
- This is a 🐛 bug fix. 
- I've added tests.

## Summary

I don't see a major use-case for *generating excerpts* for non-html pages (mainly assets) even with `page_excerpts: true`.

Should there be a need, we'll have to enable it via another config option in a future minor release so that only those willing, gets to bear the performance burden.


ashmaroli:
@jekyllbot: merge +fix

kylebarbour:
I can't test this at the moment, but this should resolve #7009. Hoping to be able to test this in the next several weeks.

PR #8235, opened 2020-06-07
Replace deprecated 'show' command with 'info'

tanant:

## Summary

Suggests using the `info` command, rather than the `show` command when trying to locate the theme.

It's a super minor thing, but when running through the quickstart/first steps, I noticed the suggested command throws a deprecation warning. Did a quick check, and looks like as at bundler 2.1.0.pre.1 `bundler info GEM` has feature parity with `bundle show GEM` so it seems like we should update the docs accordingly.

### DIFF
--- a/docs/_docs/structure.md
+++ b/docs/_docs/structure.md
@@ -36,7 +36,7 @@ A basic Jekyll site usually looks something like this:
   </p>
   <br />
   <p>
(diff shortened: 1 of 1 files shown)


ashmaroli:
Thank you @tanant 
@jekyllbot: merge +doc

PR #8250, opened 2020-06-13
Fix missing yaml front matter colon and adjust/add clarifying language.

jaybe-jekyll:
Fix missing yaml front matter colon and adjust/add clarifying language.


DirtyF:
@jekyll: merge +docs

PR #8257, opened 2020-06-20
Skip require when math_engine set to nil

### LINKED ISSUE #8256: kramdown setting `math_engine: nil` breaks kramdown_parser
## My Environment

| Operating System |  macOS Catalina          |
| `jekyll`         | 4.1.0     |
| `github-pages`   | No     |

---

## Expected Behaviour

In `_config.yml`, I have the following block of kramdown settings:
```
kramdown:
  math_engine: nil
```

This option should be accepted, as on the [kramdown documentation](https://kramdown.gettalong.org/rdoc/Kramdown/Options.html), to not use a math engine to convert math blocks. I know this worked at least as recent as jekyll version 3.4.3.

## Current Behavior

If `math_engine: nil` is set, running `jekyll build` gives a Dependency Error; in [kramdown_parser](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/converters/markdown/kramdown_parser.rb#L131), it looks for a hypothetical gem `kramdown-math-nil` which of course does not exist. The fix may be a simple as adding `&& math_engine != "nil"` to the previous if statement, which works on my machine.

## Code Sample

To replicate, create a new Jekyll site with `jekyll new .`, then append the following to your `_config.yml`:
```
kramdown:
  math_engine: nil
```
Then, run `jekyll build` and observe the aforementioned error.

zaprice:
This is a 🐛 bug fix.

## Summary

This pull request resolves the crash caused when using kramdown setting `math_engine: nil`.
`kramdown_parser` no longer throws an error by attempting to require the non-existent gem `kramdown-math-nil`.

## Context

Closes #8256

### DIFF
--- a/lib/jekyll/converters/markdown/kramdown_parser.rb
+++ b/lib/jekyll/converters/markdown/kramdown_parser.rb
@@ -127,7 +127,8 @@ module Jekyll
 
           # `mathjax` emgine is bundled within kramdown-2.x and will be handled by
           # kramdown itself.
-          if (math_engine = @config["math_engine"]) && math_engine != "mathjax"
+          if (math_engine = @config["math_engine"]) &&
+              math_engine != "mathjax" && math_engine != "nil"
             Jekyll::External.require_with_graceful_fail("kramdown-math-#{math_engine}")
           end
         end


ashmaroli:
Please explain why you are defining `math_engine: nil` ?
If you must disable it explicitly, please use the value `null` instead.

We don't wish to specially handle string `'nil'` as a valid value.

PR #8265, opened 2020-06-24
Fix grammar in documentation section

HonkingGoose:

## Summary

Change the word 'is' to 'are', because when the subject is plural, we use 'are'.

### DIFF
--- a/docs/_docs/usage.md
+++ b/docs/_docs/usage.md
@@ -19,7 +19,7 @@ Typically you'll use `jekyll serve` while developing locally and `jekyll build` 
 
 For a full list of options and their argument, see [Build Command Options](/docs/configuration/options/#build-command-options).
 
-Here is some of the most common commands:
+Here are some of the most common commands:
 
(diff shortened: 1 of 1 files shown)


ashmaroli:
Thank you for catching and correcting the issue, @HonkingGoose.
For future reference, I recommend providing a short and meaningful commit message for the sake of our Git History.

DirtyF:
@jekyll: merge

HonkingGoose:
@ashmaroli I agree that a good commit message is important! I was not sure what the right commit message should be for one-word changes such as this...

Something like:
> "Fix grammar in documentation"

Or:
> "Fix grammar, change word "is" to "are"

Or:
> "Change word from "is" to "are"

What would you prefer in future?

ashmaroli:
I believe `"Fix grammar in documentation"` summarizes the type of change and the area of code being changed, perfectly &mdash; The change is not in one of the `lib/` files, or `test/` files..

PR #8272, opened 2020-06-28
Add comment re: `verbose: true` not functional; use -V or --verbose

jaybe-jekyll:
`verbose: true` within `_config.yml` does not produce verbosity or expected results.

Added documentation comment and suggested using `-V` or `--verbose` from command line.

Should/does `verbose: true` do anything?

### DIFF
--- a/docs/_docs/configuration/default.md
+++ b/docs/_docs/configuration/default.md
@@ -61,7 +61,7 @@ paginate_path       : /page:num
 timezone            : null
 
 quiet               : false
-verbose             : false
+verbose             : false # does nothing; use -V or --verbose at command line
 defaults            : []
 
 liquid:


DirtyF:
If it does nothing let's just remove it. It was added to make sure `quiet` wins over `verbose`in https://github.com/jekyll/jekyll/commit/9c03fc3f275da9a1e893d9e46048143ac17fe8b9)

jaybe-jekyll:
> It was added to make sure `quiet` wins over `verbose`in [9c03fc3](https://github.com/jekyll/jekyll/commit/9c03fc3f275da9a1e893d9e46048143ac17fe8b9)
> 
> It only works for jekyll serve.

`verbose: true` does not produce expected results of `-V` or `--verbose` for `serve` or otherwise.

It would be nice if `verbose: true` did produce results as if specifying `--verbose` via command line.

ashmaroli:
While your claim that `verbose: true` in a config file doesn't yield expected results is correct, the proposed change in this PR is technically incorrect.

Try running the following Ruby script from the root of your site:
```ruby
require 'bundler/setup'
require 'jekyll'

Jekyll::Commands::Build.process({
  "source"      => File.expand_path(".", __dir__),
  "destination" => File.expand_path("./_site", __dir__),
  "verbose"     => true,
  "plugins"     => [],
})

**Comment by jaybe-jekyll** [comment]
> While your claim that `verbose: true` in a config file doesn't yield expected results is correct, the proposed change in this PR is technically incorrect.
> 
> Try running the following Ruby script from the root of your site:
> 
> ```ruby
> require 'bundler/setup'
> require 'jekyll'
> 
> Jekyll::Commands::Build.process({
>   "source"      => File.expand_path(".", __dir__),
>   "destination" => File.expand_path("./_site", __dir__),
>   "verbose"     => true,
>   "plugins"     => [],
> })
> ```

I don't really understand your replay and what value it provides or what solution it promotes.

Using `verbose: true` within `_config.yml` as documented does not produce `--verbose` output.

I recommend either making `verbose: true` from `_config.yml` produce the expected verbose that the documentation infers, or remove or update the documentation to reflect reality.

**Comment by DirtyF** [comment]
> Using verbose: true within _config.yml as documented does not produce --verbose output.
> 
> I recommend either making verbose: true from _config.yml produce the expected verbose results that the documentation infers, or remove or update the documentation to reflect reality.
> 

There are two different `verbose`: the `verbose` config option is different as the `verbose` option from the command.

The `verbose` config option will produce a verbose output for the webrick server, e.g:

```
[2020-06-28 22:59:47] INFO  WEBrick 1.6.0
[2020-06-28 22:59:47] INFO  ruby 2.7.1 (2020-03-31) [x86_64-darwin19]
[2020-06-28 22:59:47] DEBUG WEBrick::HTTPServlet::FileHandler is mounted on /.
[2020-06-28 22:59:47] DEBUG unmount .
[2020-06-28 22:59:47] DEBUG Jekyll::Commands::Serve::Servlet is mounted on /site.
[2020-06-28 22:59:47] DEBUG Rack::Handler::WEBrick is mounted on /admin.
[2020-06-28 22:59:47] DEBUG Rack::Handler::WEBrick is mounted on /_api.
```

The CLI option will display the full detail about the rendering phase and works with the `build` command.
Naming things is hard, we'll consider how to better distinguish those two _verbose_ modes.

Use the command line option if you're expecting details on the rendering phase.

**Comment by jaybe-jekyll** [comment]
> > Using verbose: true within _config.yml as documented does not produce --verbose output.
> > I recommend either making verbose: true from _config.yml produce the expected verbose results that the documentation infers, or remove or update the documentation to reflect reality.
> 
> There are two different `verbose`: the `verbose` config option is different as the `verbose` option from the command.
> 
> The `verbose` config option will produce a verbose output for the webrick server, e.g:
> 
> ```
> [2020-06-28 22:59:47] INFO  WEBrick 1.6.0
> [2020-06-28 22:59:47] INFO  ruby 2.7.1 (2020-03-31) [x86_64-darwin19]
> [2020-06-28 22:59:47] DEBUG WEBrick::HTTPServlet::FileHandler is mounted on /.
> [2020-06-28 22:59:47] DEBUG unmount .
> [2020-06-28 22:59:47] DEBUG Jekyll::Commands::Serve::Servlet is mounted on /site.
> [2020-06-28 22:59:47] DEBUG Rack::Handler::WEBrick is mounted on /admin.
> [2020-06-28 22:59:47] DEBUG Rack::Handler::WEBrick is mounted on /_api.
> ```
> 
> The CLI option will display the full detail about the rendering phase and works with the `build` command.
> Naming things is hard, we'll consider how to better distinguish those two _verbose_ modes.
> 
> Use the command line option if you're expecting details on the rendering phase.

Look forward to adapting to better distinguish the difference. It is confusing; this issue arose in a community setting and required time and digging to decipher reality unfortunately.

After redrafting the PR a few times, I understand your hesitancy to further comment within the configuration default.md documentation. However, closing the topic without accomplishing productive change is disappointing.

The challenge, as you've stated, is that adding `-V` or `--verbose` at command line will enable both `verbose: true` and `--verbose` when *serving* (or *building* if with the --verbose switch), providing for both verbose webrick web server output (while serving), *and* rendering phase feedback (for building *or* serving).

In contrast, only setting `verbose: true` in _config.yml produces webrick web server feedback during serving while no rendering phase feedback for serving *or* building.

That is confusing and costs people time and frustration which warrants clarification.

Perhaps a short comment for the _config.yml `verbose:` parameter-

```yaml
verbose: false # true provides webrick web server verbosity during jekyll serve.
```

... along with an update to the [documentation](https://jekyllrb.com/docs/configuration/options/#build-command-options) for `-V, --verbose` to differentiate the current-

```
Verbose output
Print verbose output.
```

to something such as...

```
Verbose output
Prints verbose output of rendering phase during `jekyll build`.
Prints verbose output of rendering phase *and* webrick web server during `jekyll serve`.
Also see [Default Configuration](https://jekyllrb.com/docs/configuration/default/) `verbose:` setting.
```

ashmaroli:
> I don't really understand your reply and what value it provides or what solution it promotes.

@jaybe-jekyll I was just pointing out the fact that even if `verbose: true` has no effect when set in a config file, it works as expected if one were to use the Ruby API directly (e.g. in a Ruby script or plugin) and therefore this pull request isn't entirely correct either.
Could you open an issue ticket regarding this grievance?
We'll try to address it in a future release.

Additionally, if you want the docs to explicitly state that the setting in the config file doesn't output the expected debug information, I'll push a commit directly.

PR #8274, opened 2020-06-30
Update pointer to special permalink variables for collections

jaybe-jekyll:
Update pointer to special permalink variables for collections


ashmaroli:
Thanks @jaybe-jekyll 
@jekyllbot: merge +doc

PR #8275, opened 2020-07-01
Fix url filters, caching key was independent on baseurl

torrocus:
This is a 🐛 bug fix.

## Summary

PR #7793 was optimizing URL filters using cache. But there didn't take into account the possibilities of different `baseurl` in one project. This case occurs when many languages ​​are used.
Now cache key is depend on `baseurl` and input path.
In addition, I added tests for these cases.

## Context

It's not related to any issue in this repository.
But working with Jekyll plugins it's related to https://github.com/kurtsson/jekyll-multiple-languages-plugin/issues/168

### DIFF
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -16,11 +16,12 @@ module Jekyll
                 else
                   (@context.registers[:cached_absolute_url] ||= {})
                 end
-        cache[input] ||= compute_absolute_url(input)
+        unique_cache_key = unique_input_path(input)
+        cache[unique_cache_key] ||= compute_absolute_url(input)
 
         # Duplicate cached string so that the cached value is never mutated by
         # a subsequent filter.
-        cache[input].dup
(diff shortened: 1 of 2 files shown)


ashmaroli:
@torrocus I understand that a change introduced in v4.1.0 broke a plugin but I don't understand this concept of multiple baseurl values for the same project. Please elaborate..

ashmaroli:
The proposed implementation here would result in unnecessary allocation of arrays and strings for all users irrespective of whether they use the plugin.
So as of now, my stand is that since the plugin mutates the value of `site.config["baseurl"]`, the plugin should implement its own override for URL Filter cache...

/cc @jekyll/core

torrocus:
@ashellwig Perhaps I did not quite express myself correctly. The point is that multi-language plugins build pages for each language separately. And then the `baseurl` in the configuration changes. That's why when I debugged, I saw `baseurl` change.

Ok, I will understand that the patch should be in the plugin.

Issue #8246, opened 2020-06-12
markdownify filter seems to ignore the {% link %} tag in front matter


beep:
## My Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Operating System | macOS 10.15.4|
| `jekyll`         | Latest     |
| `github-pages`   | No     |

---

## Expected Behaviour

I have an `intro` field in my posts’ front matter, which I use as a summary on archive pages, and on lists of related posts. I’ve noticed that if that field contains the `{% link %}` tag, that tag’s not parsed by the `markdownify` filter.

Here’s an example `intro` field:

```
---
title: Post title.
intro: Here’s a _brief_ summary of this post, which follows up on [this other post]{% link _posts/2020-01-01-previously.md %}).
…
---
```

Assuming the following syntax in my template:

```
{{ post.intro | markdownify }}
```

I’d expect it would render the following markup:

```
<p>Here’s a <em>brief</em> summary of this post, which follows up on <a href="/blog/previously/">this other post</a>.</p>
```

## Current Behavior

Currently, the above front matter renders as:

```
<p>Here’s a <em>brief</em> summary of this post, which follows up on <a href="{% link _posts/2020-01-01-previously.md %}">this other post</a>.</p>
```

It’s possible I’ve misunderstood or missed something. But thank you for your help!

ashmaroli:
To understand this behavior, you need to first be aware of the [Order of Interpretation](https://jekyllrb.com/tutorials/orderofinterpretation/) in a Jekyll build process.
Secondly, the `markdownify` filter just *converts* Markdown text into HTML. It does not interpret Liquid constructs.

DirtyF:
@ashmaroli What would it take for Jekyll to also render Liquid in front matter?

ashmaroli:
> What would it take for Jekyll to also render Liquid in front matter?

A new filter at minimum.
The simplest one would just take the given string, parse it via `Liquid::Template` and render with current context, easy to implement but may become difficult to maintain because unlike the converter filters &mdash; `markdownify`, `scssify` and `sassify`, the new filter would be *context-sensitive* and therefore prone to issues based on the use-case...

ashmaroli:
Alternatively, one can also try using the `capture` tag and then pass the result to `markdownify` filter.
The OP's scenario can be modified as:
```
{% capture intro %}{{ post.info }}{% endcapture %}
{{ intro | markdownify }}

beep:
[...]

PR #8286, opened 2020-07-06
docs: Liquid is not rendered in front matter

DirtyF:
Following the discussion in https://github.com/jekyll/jekyll/issues/8246#issuecomment-654295625
with @beep and @ashmaroli


ashmaroli:
@DirtyF [My comment](https://github.com/jekyll/jekyll/issues/8246#issuecomment-654063924) in the linked discussion was only a suggestion and hadn't been tested. I don't think it should be documented as is..
(The [Liquid documentation](https://shopify.github.io/liquid/tags/variable/#capture) itself only notes the use of Liquid variables inside the `capture` block).

Were you able to get the desired output on modifying the failing scenario?

PR #8291, opened 2020-07-08
Add a note on the rendering process in the docs

ashmaroli:

## Summary

The *Order of Interpretation* tutorial touches upon some of the quirks of Jekyll's rendering process.
But since tutorials are authored from a user's point-of-view and because they're not a readily accessible part of the documentation site, I propose to add a dedicated section to the main documentation.

## Context

Closes #8286 
Related discussion in #8246

### DIFF
--- a/docs/_data/docs_nav.yml
+++ b/docs/_data/docs_nav.yml
@@ -9,6 +9,7 @@
   docs:
     - link: /docs/usage/
     - link: /docs/configuration/
+    - link: /docs/rendering-process/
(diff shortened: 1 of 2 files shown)


ashmaroli:
> link to plugins and the guide/tutorial from this page

Good idea. We can expand this document to mention the *pre_render* and *post_render* hooks then..
But I feel that's best for a future PR.

DirtyF:
@jekyll: merge +docs

PR #8292, opened 2020-07-08
Test that Liquid expressions are not deeply evaled

ashmaroli:
## Summary

This is to test and confirm that Liquid expressions are only evaluated superficially.


DirtyF (approved):

DirtyF:
@jekyll: merge +dev

PR #8293, opened 2020-07-09
Enable Rubocop accessor grouping, fix existing offenses

torrocus:
This is refactoring.

## Summary

Enables Rubocop `Style/AccessorGrouping` rule.
And fixes existing offenses related to accessor grouping.
Accessors with comments still have them, but in some cases I shortened them to the essential context.

## Context

The goal of this PR is to to use more strict rules with Rubocop.
Consequently, this is to ensure greater readability and uniformity of the code.

### DIFF
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -5,9 +5,14 @@ module Jekyll
     include Convertible
 
     attr_writer :dir
-    attr_accessor :site, :pager
-    attr_accessor :name, :ext, :basename
-    attr_accessor :data, :content, :output
+    attr_accessor :basename,
+                  :content,
+                  :data,
+                  :ext,
+                  :name,
+                  :output,
+                  :pager,
+                  :site
 
     alias_method :extname, :ext
 
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -5,14 +5,7 @@ module Jekyll
     include Convertible
 
     attr_writer :dir
-    attr_accessor :basename,
-                  :content,
-                  :data,
-                  :ext,
-                  :name,
-                  :output,
-                  :pager,
-                  :site
+    attr_accessor :basename, :content, :data, :ext, :name, :output, :pager, :site
 
     alias_method :extname, :ext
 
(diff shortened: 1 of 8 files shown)


lib/jekyll/page.rb
```diff
@@ -5,9 +5,14 @@ class Page
...
-    attr_accessor :name, :ext, :basename
-    attr_accessor :data, :content, :output
+    attr_accessor :basename,
+                  :content,
+                  :data,
+                  :ext,
+                  :name,
+                  :output,
+                  :pager,
+                  :site
```
ashmaroli, line 15:
```suggestion
    attr_accessor :basename, :content, :data, :ext, :name, :output, :pager, :site
```

lib/jekyll/site.rb
```diff
@@ -2,15 +2,22 @@
...
+
+    attr_reader :cache_dir,
+                :config,
+                :dest,
+                :filter_cache,
+                :includes_load_paths,
+                :liquid_renderer,
+                :profiler,
+                :regenerator,
+                :source
```
ashmaroli, line 20:
```suggestion
    attr_reader :cache_dir, :config, :dest, :filter_cache, :includes_load_paths,
                :liquid_renderer, :profiler, :regenerator, :source
```

torrocus:
@ashmaroli 
I will present a different point of view. Maybe it will convince you.

Alphabetical order means that we subconsciously know where to look for a given attribute. 
The alphabet has been with us since childhood and is everywhere 
(ex. phone book, dictionary, list of countries, cities, languages on the forms).
Our brain loves patterns, and the alphabet is one of them.

I understand that longer lists pushes rest of the class structure downwards.
But this is not a problem with attributes, but a problem with the whole class.
Personally, I think classes should be small.
And most importantly they should do one thing at a time (SOLID principles).

If each attribute is on a separate line, adding or removing attributes changes only 1 line.
This is a big advantage, because the git history is preserved for the other attributes.

In addition, most editors have the ability to sort multiple lines, so keeping order is easy.
What editor are you using?

DirtyF:
@jekyll: merge +dev

PR #8294, opened 2020-07-09
Replace nested conditional with guard clauses

torrocus:
This is refactoring.

## Summary

I wanted to remove the `lib/jekyll/tags/include.rb` file from the exclusion list in `.rubocop_todo.yml`.
But when I looked there, I did a little refactoring.
Now this code is more readable.

## Context

### DIFF
--- a/lib/jekyll/tags/include.rb
+++ b/lib/jekyll/tags/include.rb
@@ -253,8 +253,7 @@ module Jekyll
       end
 
       def page_path(context)
-        page = context.registers[:page]
-        site = context.registers[:site]
+        page, site = context.registers.values_at(:page, :site)
         return site.source unless page
 
         site.in_source_dir File.dirname(resource_path(page, site))
--- a/lib/jekyll/tags/include.rb
+++ b/lib/jekyll/tags/include.rb
@@ -252,8 +252,6 @@ module Jekyll
         Array(page_path(context)).freeze
       end
 
-      private
-
       def page_path(context)
         page = context.registers[:page]
         site = context.registers[:site]
@@ -262,6 +260,8 @@ module Jekyll
         path = resource_path(page, site)
         site.in_source_dir File.dirname(path)
       end
+
+      private
 
       def resource_path(page, site)
         path = page["path"]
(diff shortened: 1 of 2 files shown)


lib/jekyll/tags/include.rb
```diff
@@ -252,21 +252,21 @@ def tag_includes_dirs(context)
         Array(page_path(context)).freeze
       end
 
+      private
+
```
ashmaroli, line 256: You should not change an existing public method into a private method unless you plan to bump the major version

torrocus: You may be right, although this method is only called within this class. There are also no separate tests or documentation for this method.

lib/jekyll/tags/include.rb
```diff
@@ -253,20 +253,20 @@ def tag_includes_dirs(context)
...
-            end
-          resource_path.sub!(%r!/#excerpt\z!, "")
-          site.in_source_dir File.dirname(resource_path)
-        end
+        page = context.registers[:page]
+        site = context.registers[:site]
+        return site.source unless page
+
+        path = resource_path(page, site)
+        site.in_source_dir File.dirname(path)
```
ashmaroli, line 261: The local variable assignment can be done away with..
```suggestion
        site.in_source_dir File.dirname(resource_path(page, site))
```

torrocus: Sometimes it improves readability. Although I could do it like this:
```
directory_path = File.dirname resource_path(page, site)
site.in_source_dir directory_path
```

ashmaroli:
If you're willing to, the first two lines could be squashed as well:
```diff
- page = context.registers[:page]
- site = context.registers[:site]
+ page, site = context.registers.values_at(:page, :site)

torrocus:
> ```diff
> - page = context.registers[:page]
> - site = context.registers[:site]
> + page, site = context.registers.values_at(:page, :site)
> ```
It makes sense. We don't need to refer to twice `context.registers`. Done.

ashmaroli:
@jekyllbot: merge +fix

ashmaroli:
Thank you @torrocus

lib/jekyll/tags/include.rb
bamaswade: Ok

bamaswade (commented):
site.in_source_dir

bamaswade (commented):
lib/jekyll/tags/include.rb

PR #8296, opened 2020-07-10
Update RuboCop to-do file

torrocus:
This is refactoring.

## Summary

Remove the `lib/jekyll/site.rb` file from the exclusion list in `rubocop_todo.yml`.
This exclusion no longer affected.
I noticed this by accident when I wanted to improve this code.

## Context


ashmaroli:
The recommended way to update the RuboCop TODO is by running:
```
bundle exec rubocop --auto-gen-config
```
That'll update the version and the timestamp in the file.

ashmaroli:
@jekyllbot: merge +dev

PR #8304, opened 2020-07-14
Add hydejack to showcase

qwtel:

 
## Summary

Add [Hydejack](https://hydejack.com/) to showcase

## Context

To be honest, I can see why you wouldn't want to include themes in the showcase, but I think I've pushed the envelope further with this one than practically any other theme, so you might want to consider it. Thanks!

### DIFF
--- a/docs/_data/showcase.yml
+++ b/docs/_data/showcase.yml
@@ -239,3 +239,9 @@
     - marketing-site
     - documentation
     - software
+- name: Hydejack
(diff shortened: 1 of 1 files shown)


DirtyF:
@qwtel Nice one, but as you get the showcase is not about listing themes, but big websites using Jekyll to show where it's used in production. I'd invite you to submit your theme to the platforms listed on https://jekyllrb.com/resources/

🙏

PR #8320, opened 2020-07-29
Clarify target of subordinate clause

chbrown:
The parenthetical clause providing examples ("such as css and images [...] and so forth") applies to "Every other directory and file", not the exceptions ("those listed above").

---

- [x] I read the contributing document at https://jekyllrb.com/docs/contributing/

## Summary

_Very_ minor grammatical correction. When I initially read the original wording, I was like, "Wait, what? That makes no sense / that seems like the opposite of what I'd expect." But then I realized if I mentally shifted the modifier clause to the beginning of the sentence, it made total sense. 

This PR corrects the attachment point of the modifier clause (but more economically than my mental shift).

### DIFF
--- a/docs/_docs/structure.md
+++ b/docs/_docs/structure.md
@@ -194,8 +194,8 @@ An overview of what each of these does:
       </td>
       <td>
         <p>
-          Every other directory and file except for those listed above—such as
-          <code>css</code> and <code>images</code> folders,
+          Except for the special cases listed above, every other directory and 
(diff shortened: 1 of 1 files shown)


DirtyF:
@jekyll: merge +docs

chbrown:
P.S. Very nice PR experience! (My first in Jekyll.) Minimal impact ⇒ minimal fuss, as it should be. Plus, seeing it deployed on the live website so quickly is quite satisfying 😄 

👍 👍

PR #8331, opened 2020-08-09
Allow webrick[headers][Cache-Control] to be 'unset' using _config.yml (and expand test cases in Windows)

pccasto:

## Summary

This change allows a user to cause WEBrick to send headers without any Cache-Control field, by setting a configuration value to `nil`. Previously, the best that could be achieved was to cause the value of Cache-Control to be an empty string, but there was no way to send a header without that field at all.

As a necessary part of testing (in my Windows environment), this pull request changes the Gemfile, and removes a skip instruction in the `test_commands_serve.rb`. This has the added benefit of allowing four cases which were previously skipped on Windows to run. The change to the Gemfile allows the EventMachine gem to be properly sourced for a Windows environment. I included it in the pull request, because the testing was closely coupled to the feature change. However, neither of those changes is essential for cases where testing is done outside of Windows.

## Context

The need for this came about because of the behavior of the `netlify/cli` tool. 

Using the tool `netlify dev` in a development environment with Jekyll starts a npm based web server that acts as a front-end proxy to Jeckyll/WEBrick. The code in netlify-cli and its associated npm proxy module (http-party/node-http-proxy) does not allow the proxying server to set a value for a field in a header when the originating server has already set a value for that particular field. This makes it challenging to test header configuration settings at the `netlify` level. See: https://github.com/netlify/cli/issues/990.

While netlify/cli and http-party/node-http-proxy may change in the future, there may be other situations where a user may not wish to send out one of the default WEBrick header values as specified by Jekyll (currently only the one). This change gives the user control over that capability on the Jekyll side.

This feature could be extended to other aspects of configuration (e.g. Content-Type), but is currently limited to just the static default setting.

---
The existing test `test_commands_service` had several tests skipped while running in a Windows environment. That section seemed to be the most appropriate place to add the tests for this change. By installing the EventMachine gem with the `--platform ruby` option, or by including it in the Gemfile as it is now specified, EventMachine actions can work on Windows.

I know that change goes beyond the scope of this feature, but because it was intertwined with the testing (in my main environment), I included it as part of this pull request.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -31,6 +31,8 @@ group :test do
   gem "test-theme-symlink", :path => File.expand_path("test/fixtures/test-theme-symlink", __dir__)
 
   gem "jruby-openssl" if RUBY_ENGINE == "jruby"
+
+  gem 'eventmachine', '1.2.7', git: 'https://github.com/eventmachine/eventmachine', tag: 'v1.2.7' if Gem.win_platform?
 end
 
 #
--- a/Gemfile
+++ b/Gemfile
@@ -31,8 +31,7 @@ group :test do
   gem "test-theme-symlink", :path => File.expand_path("test/fixtures/test-theme-symlink", __dir__)
 
   gem "jruby-openssl" if RUBY_ENGINE == "jruby"
-
-  gem 'eventmachine', '1.2.7', git: 'https://github.com/eventmachine/eventmachine', tag: 'v1.2.7' if Gem.win_platform?
+  gem "eventmachine", "1.2.7", git: "https://github.com/eventmachine/eventmachine", tag: "v1.2.7" if Gem.win_platform?
 end
 
 #
--- a/docs/_docs/configuration/webrick.md
+++ b/docs/_docs/configuration/webrick.md
@@ -21,7 +21,7 @@ with Chrome's aggressive caching when you are in development mode.
 
 
 You can override the default `Cache-Control` by setting it to a value of your
-own choosing including the empty string. If you don't want the WEBrick server
+own choosing (a caching instruction, or an empty string). If you don't want the WEBrick server
 to send any `Cache-Control` header at all, set the value to nil. That configuration
 is useful when Jekyll/WEBrick is behind a proxy server, such as `netlify dev`.
 
--- a/docs/_docs/configuration/webrick.md
+++ b/docs/_docs/configuration/webrick.md
@@ -18,3 +18,16 @@ Jekyll provides by default `Content-Type` and `Cache-Control` response
 headers: one dynamic in order to specify the nature of the data being served,
 the other static in order to disable caching so that you don't have to fight
 with Chrome's aggressive caching when you are in development mode.
+
+
+You can override the default `Cache-Control` by setting it to a value of your
+own choosing including the empty string. If you don't want the WEBrick server
+to send any `Cache-Control` header at all, set the value to nil. That configuration
+is useful when Jekyll/WEBrick is behind a proxy server, such as `netlify dev`.
+
+```yaml
+# File: _config.yml
+webrick:
+  headers:
+    Cache-Control: nil
+```
--- a/lib/jekyll/commands/serve/servlet.rb
+++ b/lib/jekyll/commands/serve/servlet.rb
@@ -193,7 +193,9 @@ module Jekyll
         def set_defaults
           hash_ = @jekyll_opts.fetch("webrick", {}).fetch("headers", {})
           DEFAULTS.each_with_object(@headers = hash_) do |(key, val), hash|
+            # use the default value unless the user specifies a value, and that value is not nil
             hash[key] = val unless hash.key?(key)
+            hash.delete(key) if hash[key].nil?
           end
         end
       end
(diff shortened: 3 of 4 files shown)


Gemfile
```diff
@@ -31,6 +31,7 @@ group :test do
   gem "test-theme-symlink", :path => File.expand_path("test/fixtures/test-theme-symlink", __dir__)
 
   gem "jruby-openssl" if RUBY_ENGINE == "jruby"
+  gem "eventmachine", "1.2.7", git: "https://github.com/eventmachine/eventmachine", tag: "v1.2.7" if Gem.win_platform?
```
ashmaroli, line 34: What is the difference between this and using the released version (`v1.2.7`) directly?

docs/_docs/configuration/webrick.md
```diff
@@ -18,3 +18,16 @@ Jekyll provides by default `Content-Type` and `Cache-Control` response
 headers: one dynamic in order to specify the nature of the data being served,
 the other static in order to disable caching so that you don't have to fight
 with Chrome's aggressive caching when you are in development mode.
+
```
ashmaroli, line 21: Unnecessary blank line.
```suggestion
```

docs/_docs/configuration/webrick.md
```diff
@@ -18,3 +18,16 @@ Jekyll provides by default `Content-Type` and `Cache-Control` response
...
+
+You can override the default `Cache-Control` by setting it to a value of your
+own choosing (a caching instruction, or an empty string). If you don't want the WEBrick server
```
ashmaroli, line 24: Inconsistent line-lengths on this paragraph.
If the previous line ended at `your`, why is this line ending after three more words?

docs/_docs/configuration/webrick.md
```diff
@@ -18,3 +18,16 @@ Jekyll provides by default `Content-Type` and `Cache-Control` response
...
+to send any `Cache-Control` header at all, set the value to nil. That configuration
+is useful when Jekyll/WEBrick is behind a proxy server, such as `netlify dev`.
+
+```yaml
+# File: _config.yml
+webrick:
+  headers:
+    Cache-Control: nil
```
ashmaroli, line 32: *Attention:* YAML treats this `nil` as a string `"nil"`
You should instead use `null` or `~` or leave undefined / empty.

ashmaroli:
Thank you for submitting a complete pull request with documentation and tests included. However, some changes need to be made.
I have not screened the tests yet, may ask for changes there too if needed.

Gemfile
pccasto: If done without this approach (by just listing `gem "eventmachine"` in the Gemfile), the package will be downloaded and built locally as EventMachine-mingw32. The native extension versions does not work in windows. It builds/installs, but fails in use.  This is why you've had to comment out the tests in Windows in the past. 

It is not about the specific version, it is about getting a pure ruby version of the gem. The eventmachine gem will work on windows.  It just has to be the version built with --platform ruby.  Unfortunately trying to set the platform to ruby using bundler or in the Gemfile is still problematic.  See https://github.com/rubygems/rubygems/issues/3347. 

The approach used here causes the gem to be built as if this were done an the command line: `gem install eventmachine --platform ruby `. The only difference is a bit more network traffic, since there isn't a locally built copy - the gem is pulled from github.

```
$ bundle install | grep eventmachine
Using eventmachine 1.2.7 from https://github.com/eventmachine/eventmachine (at v1.2.7@9cb13a8)
```

Someone else came up with this incantation (https://www.dixis.com/posts/171). It was a good find that helps solve/work-around this issue until such time as bundler and Gemfiles can accommodate specifying --platform ruby.

It would also be possible to manually install  ```gem install eventmachine --platform ruby``` and then update the Gemfile to use eventmachine. But that requires manual work, and would not be something that would work for the automated tests.

docs/_docs/configuration/webrick.md
pccasto: Ouch.  How could I have forgotten that...  And my unit tests explicitly set the value to nil in a hash (which makes perfect sense in Ruby), rather than actually reading from YAML. So my use of nil in the hash worked, but it will not when actually read.

I'm going to go back and do a bit more end-to-end local testing, and then update documentation accordingly.

I'll have an updated pull request this evening (about 12 hours from now in my TZ).

pccasto:
> Thank you for submitting a complete pull request with documentation and tests included. However, some changes need to be made.
> I have not screened the tests yet, may ask for changes there too if needed.

The tests manage to prove that the code properly handles the situation when the Cache-Controls is nil/null at the time this method is called. However, when going back to do some end-to-end tests, I found there were places where someplace in the Jekyll/WEBrick flow that the Cache-Control header was being set, in spite of the configuration in the yaml file. 

The primary page winds up being handled correctly, but all of the subsequent pages/images that are pulled in by the primary page are not.  I need to move this to WIP and evaluate the code paths that would cause this.

ashmaroli:
ProTip: Instead of prefixing the PR title with `WIP:`, convert the PR into a *draft PR*.
Convert it back when you're sure there's nothing more left to be done.

PR #8344, opened 2020-08-16
Remove Redcarpet from docs

lime:

## Summary

I've removed the Redcarpet section from the Markdown configuration docs.

## Context

Redcarpet support was removed in v4.0 (#6987). The docs were updated in a separate PR (#6990), but that was merged into the [docs-40 branch](https://github.com/jekyll/jekyll/tree/docs-40) and never made it into master.

### DIFF
--- a/docs/_docs/configuration/default.md
+++ b/docs/_docs/configuration/default.md
@@ -73,9 +73,6 @@ liquid:
 rdiscount:
   extensions        : []
 
-redcarpet:
-  extensions        : []
-
 kramdown:
   auto_ids          : true
   entity_output     : as_char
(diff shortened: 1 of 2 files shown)


ashmaroli:
Golly! Thank you for catching the issue and submitting the fix.
The `docs-40` branch has conflicts now so can't be easily merged. So this will do.

docs/_docs/configuration/default.md
```diff
@@ -73,9 +73,6 @@ liquid:
 rdiscount:
   extensions        : []
```
lime, line 75: While we're at it, the RDiscount example config could probably also be removed.

```suggestion
```

ashmaroli: That's not needed in this PR. I'll merge the original `docs-40` branch after rebasing it so that the original contributions find their way to `master`

ashmaroli:
Oops! I mean't to hide your comment. Not delete it..
My bad!

ashmaroli:
@lime I've manually added you as the co-author of the commit on the `docs-40` branch as an attribution for this PR.
Thank you once again.

PR #8347, opened 2020-08-17
Supress warning issued for redirect pages

### LINKED ISSUE #8346: Warning issued for redirect pages
With the merge of #8342 warnings are issued when a site uses the `jekyll-redirect-from` plugin:
```
Conflict: The URL 'jekyll/docs/_site/conduct/index.html' is the destination for the following pages: redirect.html, redirect.html
```
This should have been caught during the review, though.
Since it slipped through, the way forward would be to **have an explicit check for a redirect-page** and not issue the warning since the use of same permalink is deliberate.

/cc @SeekingMeaning

SeekingMeaning:
This is a 🐛 bug fix.

## Summary

This updates `lib/jekyll/commands/doctor.rb` to skip redirect pages (from the `jekyll-redirect-from` plugin) when checking for permalink conflicts

## Context

Closes #8346

### DIFF
--- a/features/permalinks.feature
+++ b/features/permalinks.feature
@@ -151,3 +151,19 @@ Feature: Fancy permalinks
     And the _site directory should exist
     And I should see "Conflict: The URL '" in the build output
     And I should see "amazing.html' is the destination for the following pages: awesome.md, cool.md" in the build output
+
+  Scenario: Use the same redirect twice
+    Given I have a configuration file with "plugins" set to "[jekyll-redirect-from]"
+    And I have an "index.html" file with content:
+    """
+    ---
+    redirect_from:
+      - redirect.html
+      - redirect.html
+    ---
+    """
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should not see "Conflict: The URL '" in the build output
+    And I should not see "redirect.html' is the destination for the following pages: redirect.html, redirect.html" in the build output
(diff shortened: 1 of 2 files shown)


ashmaroli:
I think having a separate `:allow_used_permalink?` method should be safe.
If there are users facing this issue with another plugin, we can ask them to write a local plugin that redefines the said `:allow_used_permalink?` method to suit their need..
```ruby
def allow_used_permalink?(item)
  defined?(JekyllRedirectFrom) && item.is_a?(JekyllRedirectFrom::RedirectPage)
end
```
Additionally, this change definitely needs a test. You can add another Cucumber scenario like before.

SeekingMeaning:
> I think having a separate `:allow_used_permalink?` method should be safe.

Just making sure, where should this method be added?

ashmaroli:
> where should this method be added?

It can be a private class method in `Jekyll::Commands::Doctor`

ashmaroli:
Thank you for the quick turnarounds @SeekingMeaning 
I would like the cucumber scenario to more resemble a real-world use-case.

`redirect_from ["redirect.html", "redirect.html"]` doesn't seem clear enough to me.

If you're able to change it, great! Otherwise, I shall update this branch when I can come up with a nice scenario.

ashmaroli (approved):

ashmaroli:
Thank you @SeekingMeaning
@jekyllbot: merge +minor

PR #8350, opened 2020-08-17
chore(release): v4.1.2 :gem: security

DirtyF:
Patch release to force kramdown 2.3.0 to avoid [CVE-2020-14001](https://github.com/advisories/GHSA-mqm2-cgpr-p4m6)

### DIFF
--- a/History.markdown
+++ b/History.markdown
@@ -1,6 +1,12 @@
-## HEAD
+## 4.1.2 / 2020-08-17
 
-  * Update default.html (#8305)
+### Bug Fixes
+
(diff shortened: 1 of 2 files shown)


ashmaroli:
@DirtyF I don't see the need to force kramdown-2.3.x.
Users on v4.x or v3.9 just need to run `gem update kramdown` or `bundle update kramdown`.

However, if we must ship a v4.1.2, then it has to be based off a `4.1-stable`. The one minor enhancement on `master` is not ready to be shipped.

DirtyF:
@ashmaroli I'll let you handle it then.

PR #8353, opened 2020-08-24
Update default.md with info requested in #8314

### LINKED ISSUE #8314: plugins_dir documentation and verbose output
## Motivation

It's not clear how to use the config key `plugins_dir`, causing people to hit issues e.g. #5517, #5527, [and here](https://github.com/jekyll/jekyll/issues/7842#issuecomment-664391036).

## Suggestion

1. Clarify that directory strings must be relative to CWD not source directory.
2. Document that this key can take an array of strings, and plugins will then be loaded in this order

Furthermore when providing incorrect strings, even under `--verbose` output, we do not get useful error messages explaining why plugins have not loaded.

shannonxtreme:

## Summary

 - Added a `note info` div to clarify CWD vs site source for dir keys.
 - Added an inline comment to plugins_dir key to clarify load order in array of strings.

## Context

  Resolves #8314

### DIFF
--- a/docs/_docs/configuration/default.md
+++ b/docs/_docs/configuration/default.md
@@ -7,7 +7,7 @@ Jekyll runs with the following configuration options by default. Alternative
 settings for these options can be explicitly specified in the configuration
 file or on the command-line.
 
-<div class note info>
+<div class="note info">
   <h5>Be aware of directory paths</h5>
   <p>
     Make directory path values in configuration keys like `plugins_dir` relative to the current working directory, not the site source. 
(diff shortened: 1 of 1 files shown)


ashmaroli:
Adding the source code below for reference:
https://github.com/jekyll/jekyll/blob/cac8d6ca5facf490c8c3e8843d8998b0645e3481/lib/jekyll/plugin_manager.rb#L82-L103

ashmaroli (approved):

ashmaroli:
Thank you for improving our documentation @shannonxtreme

DirtyF (approved):

DirtyF:
@jekyll: merge +docs

PR #8354, opened 2020-08-24
Clarify description of `safe` option

### LINKED ISSUE #8321: docs: safe option wording is ambiguous
## Motivation

[`safe` option wording is ambiguous](https://jekyllrb.com/docs/configuration/options/)

> Disable custom plugins, caching to disk and ignore symbolic links.

What does Disable apply to?

## Suggestion

Avoid ambiguous commas. Use short clear sentences.

- Disable custom plugins. Disable disk caching. Ignore symbolic links.
- Disable custom plugins, disk caching, symbolic links.

shannonxtreme:

## Summary

Change `Safe` option description to improve clarity and specificity. 

## Context

Resolves #8321

### DIFF
--- a/docs/_docs/configuration/options.md
+++ b/docs/_docs/configuration/options.md
@@ -44,7 +44,9 @@ class="flag">flags</code> (specified on the command-line) that control them.
       <td>
         <p class="name"><strong>Safe</strong></p>
         <p class="description">
-          Disable <a href="/docs/plugins/">custom plugins</a> and caching to disk. Ignore symbolic links.
+          Disable <a href="/docs/plugins/">plugins</a>, caching to disk
+          and ignore symbolic links.  </br>
+          Whitelisted plugins are not disabled.
         </p>
       </td>
       <td class="align-center">
--- a/docs/_docs/configuration/options.md
+++ b/docs/_docs/configuration/options.md
@@ -44,8 +44,7 @@ class="flag">flags</code> (specified on the command-line) that control them.
       <td>
         <p class="name"><strong>Safe</strong></p>
         <p class="description">
-          Disable <a href="/docs/plugins/">custom plugins</a>, caching to disk
-          and ignore symbolic links.
+          Disable <a href="/docs/plugins/">custom plugins</a> and caching to disk. Ignore symbolic links.
         </p>
       </td>
       <td class="align-center">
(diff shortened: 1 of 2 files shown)


ashmaroli:
While we're at it, I think it'd be better if we also clarify that it is not just *custom plugins* that are disabled, but also plugin-gems that have not been listed under the `whitelist` key.

shannonxtreme:
Does this mean all plugins not in the whitelist, regardless of whether "custom" or a gem, are disabled for safe mode?

ashmaroli:
> Does this mean all plugins not in the whitelist, regardless of whether "custom" or a gem, are disabled for safe mode?

Technically, yes.
But there is a catch, though. This screening is only applicable for plugins loaded via the config file.
If one lists a gem under the `:jekyll-plugins` group in the Gemfile, it will be loaded irrespective of the `safe` option.
The above observation is just for your information. It need not be included in this PR to avoid unnecessary confusion.

shannonxtreme:
@ashmaroli I added a clarification and removed "custom" from plugins to make it less confusing. The gist of it is that unless whitelisted, plugins are disabled. Didn't add anything about gems in `:jekyll-plugins` but you can modify the line to

 "Whitelisted plugins as well as gems listed in `:jekyll-plugins` on the Gemfile are not disabled." 

if you want.

docs/_docs/configuration/options.md
```diff
@@ -44,8 +44,9 @@ class="flag">flags</code> (specified on the command-line) that control them.
       <td>
         <p class="name"><strong>Safe</strong></p>
         <p class="description">
-          Disable <a href="/docs/plugins/">custom plugins</a>, caching to disk
-          and ignore symbolic links.
+          Disable <a href="/docs/plugins/">plugins</a>, caching to disk
+          and ignore symbolic links.  </br>
+          Whitelisted plugins are not disabled.
```
ashmaroli, line 49: How about rephrasing to the following?
```suggestion
          Disable <a href="/docs/plugins/">non-whitelisted plugins</a>, caching to disk
          and ignore symbolic links.
```

shannonxtreme: I like it. The only reason I had it separate was to link "whitelisted" to the config page for whitelisting.

ashmaroli: I don't think we have a dedicated section on the `whitelist` configuration..
If that's indeed the case, I think you can add a *note* to the page at `/docs/plugins/` informing the visitor about the whitelist..
(*it can be part of this PR itself*)

ashmaroli (approved):

ashmaroli:
Thank you @shannonxtreme 
@jekyllbot: merge +doc

PR #8355, opened 2020-08-24
Run server with Autoreload

DespertaWeb:
Show how to serve the web with autoreload rightaway:

`bundle exec jekyll s --livereload
`

## Summary

Described how to add autoreload mode when running the server:
`bundle exec jekyll s --livereload` 

## Context
At the quick start guide is not explained how to run in autoreload mode. And this is a must feature always.

### DIFF
--- a/docs/_docs/index.md
+++ b/docs/_docs/index.md
@@ -33,6 +33,9 @@ cd myblog
 5. Build the site and make it available on a local server.
 ```
 bundle exec jekyll serve
+
+// with autoreload
+bundle exec jekyll s --livereload
 ```
 6. Browse to [http://localhost:4000](http://localhost:4000){:target="_blank"}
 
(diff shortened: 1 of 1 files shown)


ashmaroli:
Personally, I don't think the quickstart section has to recommend livereloading straight away..

DespertaWeb:
> Personally, I don't think the quickstart section has to recommend livereloading straight away..

Well, there's no one interested on a tecnology that forces you to manually reload. And unless they explicitly tell you that autoreload is possible, you most likely will abandon for good that technology.

At least that is exactly what happened to me. I've been about to try out another tech 'cause I didn't wanted to be CMD+R every 5 secons like in the old good times. Now all techs just autoreload by just running the server.

DirtyF (changes requested):
👍 This common option is hidden deep in the configuration table for now.

The current edit breaks the rendering, I suggest we add a note instead.

DespertaWeb:
> The current edit breaks the rendering

This is only for DEV environment, shouldn't be used in PROD.

Anyway adding a note sound super to me ^^

DirtyF (approved):
@jekyll: merge +docs

DirtyF:
@jekyll:  merge +docs

PR #8358, opened 2020-08-27
Simplifying the Git post-receive hook-example

tnt:
In a bare git repo (which is what we dealing with here) the [$GIT_DIR](https://git-scm.com/docs/githooks/)-variable points to what we want to clone here. (It would actually also work with a non-bare repo though `$GIT_DIR` points to the repo-root's `.git` subdirectory in this case.)

So omitting `$GIT_REPO` and using `$GIT_DIR` instead makes one less adjustment necessary.

This is a 🔨 code refactoring of some code-example in the docs.

## Context

Just used it and had to do one additional edit before it worked because I overlooked the $GIT_REPO variable. 

I you think it is not worth thinking about it to much or conducting en extended discussion about it. So don't hesitate to just drop this pr without further notice before it creates headaches to anyone. It is meant as a quick improvement but not to waste your time.

### DIFF
--- a/docs/_docs/deployment/automated.md
+++ b/docs/_docs/deployment/automated.md
@@ -43,12 +43,11 @@ installed on the server:
 export GEM_HOME=$HOME/gems
 export PATH=$GEM_HOME/bin:$PATH
 
-GIT_REPO=$HOME/myrepo.git
 TMP_GIT_CLONE=$HOME/tmp/myrepo
 GEMFILE=$TMP_GIT_CLONE/Gemfile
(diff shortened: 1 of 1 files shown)


ashmaroli (approved):
More info at:
https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables

DirtyF (approved):

ashmaroli:
Thank you for improving our docs @tnt 
@jekyllbot: merge +doc

PR #8359, opened 2020-08-27
Add note about rebooting system after installation

### LINKED ISSUE #8277: Win10, Ruby 2.7.1 installation
Hi,

I'm on Windows 10,  with Ruby 2.7.1 and I still have no clue how to properlly install Jekyll.
Here's what I do (following https://jekyllrb.com/docs/installation/windows/):
1. I install Ruby from the installer
2. check to install MSYS2 as well
3. Run ridk install at the end of installation, which asks me to choose what to do next:
 (MSYS base / MSYS ubdate or MSYS + MINGW) 
and this is already not documented.

The link to RubyInstaller Documentation doesn't help at all (https://github.com/oneclick/rubyinstaller2#using-the-installer-on-a-target-system), as it doesn't contain anything relevant to what I'm doing at this stage.

I choose the third option, to have everything installed and proceed.

In step 3. I run gem install jekyll bundler, but it fails, jekyll is not installed, and there is absolutely no hint what to do with that. The docs jus assume that everything works nicely. Well, it doesn't.
The bundler installs fine, but jekyll doesn't and I have no clue what to do next.

shannonxtreme:

## Summary

Added a note info class to line 31 that also contains a link to a new RubyInstaller issue.

## Context

Resolves #8277


larskanis (approved):
LGTM. I'll release a new series of RubyInstaller versions tomorrow. They should fix the recent issues with MSYS2. Nevertheless I think this note is a good addition.

DirtyF (approved):

ashmaroli:
@larskanis Should the system be rebooted or just the command line program?

larskanis:
@ashmaroli The system should be rebooted. This ensures that the command line gets the latest system PATH variable and that all MSYS2 processes have been terminated.

Background: MSYS2 uses shared memory to implement the `fork()` call on Windows. This is quite tricky and doesn't work across distinct MSYS2 versions. As a result MSYS2 executables terminate with strange error messages. Even if the user thinks, that all MSYS2 applications have been closed, there are some background processes like GPG or SSH agents, that are still running and can prohibit execution of other MSYS2 apps. I implemented some countermeasures in recent RubyInstaller, but the ultimate way is still a reboot.

ashmaroli:
Thank you for the clarification @larskanis

ashmaroli:
Thank you @shannonxtreme 
@jekyllbot: merge +doc

PR #8361, opened 2020-08-27
Adding handling of collections_dir

alexander-morris:
I painstakingly discovered that when the specified directory's name contains '_' or is exactly the word 'collections' then _posts will not be read while contained within it.

Just hoping to share it to help others avoid the same pitfall. :)

### DIFF
--- a/docs/_docs/configuration/options.md
+++ b/docs/_docs/configuration/options.md
@@ -32,6 +32,16 @@ class="flag">flags</code> (specified on the command-line) that control them.
     </tr>
     <tr class="setting">
       <td>
+        <p class="name"><strong>Collections Directory</strong></p>
+        <p class="description">Change the direction where collections are stored. *Cannot be 'collections', and cannot contain '_' if _posts/ is inside it.</p>
+      </td>
+      <td class="align-center">
+        <p><code class="option">source: collections_dir</code></p>
+        <p><code class="flag">-s, --source collections_dir</code></p>
+      </td>
+    </tr>
+    <tr class="setting">
+      <td>
         <p class="name"><strong>Site Destination</strong></p>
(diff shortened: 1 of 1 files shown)


ashmaroli (changes requested):
The proposed changes cannot be accepted as they are incorrect:
- **`collections_dir` can be named `"collections"` and can contain `_`**.
Just that it cannot start with the characters `.` / `_` / `#` / `~`
- `collections_dir` does not have an associated command-line switch. So, there is no reason to document the setting in this table or in this document entirely

alexander-morris:
Hey @ashmaroli - thanks for the quick reply. 

I can definitely move this note elsewhere and remove the command-line switch.

I'm not sure about the 'collections' title though. In my case, when it is called 'collections', the _posts directory within it is not read.

ashmaroli:
> In my case, when it is called 'collections', the _posts directory within it is not read.

We already have a test for it:
https://github.com/jekyll/jekyll/blob/badc9c2d30e6b34a97ccdb0f20b1635e1aea575d/features/collections_dir.feature#L6-L18

MichaelCurrin:
My comments are similar to those above.

The change looks out of place as I would expect a 1:1 mapping between sections on the left and arguments on the right. This extra row is not a new argument but just a note on an _existing_ argument.

I suggest rather update this section:

```
Site Source

Change the directory where Jekyll will read files
```

to say:

```
Site Source

Change the directory where Jekyll will read files.

If you have a collections directory e.g. `my_collection`, it must be inside this source directory. It may not be named 'collections' and may not start with `.`,  `_`, `#` or `~`.
```

I don't think if the `_posts` part is relevant. The name rule stands alone without caring about `_posts` is inside it (which is not an antipattern)

PR #8363, opened 2020-08-28
Only enable BOM encoding option on UTF encodings

### LINKED ISSUE #7972: warning: BOM with non-UTF encoding ISO-8859-1 is nonsense
## My Environment

| Operating System |  Linux/Debian |
| `jekyll`         | Latest     |
| `github-pages`   | No     |

---

## Expected Behaviour

No BOM warning is produced when files are read

## Current Behavior

When using `encoding: ISO-8859-1` in `_config.yml`, the following errors occur when building a jekyll site, once for every time a file is read:
- `.../vendor/bundle/ruby/2.5.0/gems/jekyll-4.0.0/lib/jekyll/convertible.rb:41: warning: BOM with non-UTF encoding ISO-8859-1 is nonsense`
- `.../vendor/bundle/ruby/2.5.0/gems/jekyll-4.0.0/lib/jekyll/tags/include.rb:179: warning: BOM with non-UTF encoding ISO-8859-1 is nonsense`

Inserting `puts file_read_opts(context)` on line 179 of `include.rb` shows `{:encoding=>"bom|ISO-8859-1"}`. Hard-coding the second argument to `File.read()` to be `{:encoding=>"ISO-8859-1"}` fixes the warning. I'm not sure where the `bom` bit is coming from or whether `File.read()` ought to be happy with it being there, but that appears to be the source of the problem.

(Sorry, I wish I could use utf-8 encoding too :smile: )

LinuxMercedes:
This is a 🐛 bug fix.

## Summary

Ruby does not allow the BOM option on non-UTF encodings, warning that it is "nonsense". To remove this warning, we change the BOM option check here to match how Ruby checks for compatible encodings.

Relevant bits from the Ruby source are the [test](https://github.com/ruby/ruby/blob/ruby_2_7/io.c#L5693-L5699) and the [`utf_prefix` string](https://github.com/ruby/ruby/blob/ruby_2_7/io.c#L5454).

## Context

Fixes #7972.

### DIFF
--- a/lib/jekyll/utils.rb
+++ b/lib/jekyll/utils.rb
@@ -303,12 +303,15 @@ module Jekyll
     # and a given param
     def merged_file_read_opts(site, opts)
       merged = (site ? site.file_read_opts : {}).merge(opts)
-      if merged[:encoding] && !merged[:encoding].start_with?("bom|")
+
+      # always use BOM when reading UTF-encoded files
+      if merged[:encoding]&.downcase&.start_with?("utf-")
         merged[:encoding] = "bom|#{merged[:encoding]}"
       end
-      if merged["encoding"] && !merged["encoding"].start_with?("bom|")
+      if merged["encoding"]&.downcase&.start_with?("utf-")
         merged["encoding"] = "bom|#{merged["encoding"]}"
       end
+
       merged
     end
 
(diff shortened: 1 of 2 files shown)


ashmaroli (approved):
LGTM!

ashmaroli:
Thanks @LinuxMercedes
@jekyllbot: merge +fix

PR #8389, opened 2020-09-14
Stash documents `write?` attribute in a variable

ashmaroli:

## Summary

Instead of checking `collection&.write? && site.publisher.publish?(self)` on every call to the `:write?` method, compute the result once per instance and reuse the result on subsequent calls.

## Context

`document.write?` is invoked via various core classes (e.g. `Site`, `Regenerator`) and in-house plugins (e.g. `jekyll-mentions`, `jemoji`)

### DIFF
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -352,7 +352,9 @@ module Jekyll
     # method returns true, and if the site's Publisher will publish the document.
     # False otherwise.
     def write?
-      collection&.write? && site.publisher.publish?(self)
+      return @write_p if defined?(@write_p)
+
+      @write_p = collection&.write? && site.publisher.publish?(self)
     end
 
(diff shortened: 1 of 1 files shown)


DirtyF (approved):
Good catch 🥇

ashmaroli:
Thanks :)
@jekyllbot: merge +fix

DirtyF:
@ashmaroli I hadn't build Jekyll website locally for a few months, last time it was taking 6.2s on my machine, current `master` is now at 3.6s. 🚀 👏

ashmaroli:
The recent optimizations aren't major ones. Just minor bottlenecks I discovered when profiling various sites in the wild.
Our docs site is too small to get a noticeable improvement from these changes..
Thanks for the encouragement anyways. :smiley: 
Cheers :clinking_glasses:

PR #8392, opened 2020-09-16
Reduce string allocations from generating doc URLs

ashmaroli:

## Summary

Generating URLs from a drop (current route for all documents of a collection) results in a lot of string allocations because of use of non-mutating methods to manipulate strings.

This pull request replaces those non-mutating methods that mutate a duplicated version of long-term strings and mutate temporary strings.

### DIFF
--- a/lib/jekyll/url.rb
+++ b/lib/jekyll/url.rb
@@ -94,7 +94,8 @@ module Jekyll
 
     def generate_url_from_drop(template)
       template.gsub(%r!:([a-z_]+)!) do |match|
-        pool = possible_keys(match.sub(":", ""))
+        name = Regexp.last_match(1)
+        pool = name.end_with?("_") ? [name, name.chomp!("_")] : [name]
 
         winner = pool.find { |key| @placeholders.key?(key) }
         if winner.nil?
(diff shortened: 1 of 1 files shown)


lib/jekyll/url.rb
```diff
@@ -94,7 +94,8 @@ def possible_keys(key)
 
     def generate_url_from_drop(template)
       template.gsub(%r!:([a-z_]+)!) do |match|
-        pool = possible_keys(match.sub(":", ""))
+        name = Regexp.last_match(1)
+        pool = name.end_with?("_") ? [name, name.chomp!("_")] : [name]
```
ashmaroli, line 98: Notes:
- `Regexp.last_match(1)` here is equivalent to `match.sub(":", "")` but without the extra allocation.
- The `#possible_keys` public method does not mutate the argument `key` but this inlined version modifies the string in-place.

lib/jekyll/url.rb
```diff
@@ -107,15 +108,17 @@ def generate_url_from_drop(template)
         value = "" if value.nil?
         replacement = self.class.escape_path(value)
 
-        match.sub(":#{winner}", replacement)
-      end.squeeze("/")
+        match.sub!(":#{winner}", replacement)
+      end
```
ashmaroli, line 112: Notes:
- Calling `String#squeeze` here is redundant *with respect to the bigger picture (`Jekyll::URL.new.to_s`)* because the result from this method is passed onto `Jekyll::URL#sanitize_url` that removes extra `"/"` from input strings.
- The local `match` variable value is not further required. So it can be mutated.

DirtyF:
```diff
- Total allocated: 300.38 MB (3084063 objects)
- Total retained:  20.75 MB (113527 objects)

+ Total allocated: 300.31 MB (3082539 objects)
+ Total retained:  20.75 MB (113529 objects)
```

```diff
- Docs site build:  6.03 seconds
+ Docs site build: 3.969 seconds
```

ashmaroli:
Its at times like these that I wish we could revive Utterson from its ashes..

ashmaroli:
### Third Party Repo Profile Summary

```diff
--- master
+++ PR

- Total allocated: 296.75 MB (2461825 objects)
- Total retained:  43.43 MB (112686 objects)
+ Total allocated: 295.53 MB (2435049 objects)
+ Total retained:  43.43 MB (112687 objects)
```

lib/jekyll/url.rb
ashmaroli: Sadly, It is not a private method. So we'll have to bear the tech debt until v5.0

DirtyF (approved):

ashmaroli:
@jekyllbot: merge +fix

PR #8393, opened 2020-09-16
Check default front matter scope against symbols

ashmaroli:

## Summary

The `type` attribute of Jekyll's core convertible classes are Symbols. i.e. Pages are `:pages`, documents are the parent collection-label symbolized, etc

(The exception being `Jekyll::StaticFile` and `Jekyll::Layout` which have `type` as `nil`).

Coercing a Symbol (or `nil`) to String allocates the string to memory on each call. Since this coercion in the code below is only for comparison, we can instead coerce the string values to Symbols and avoid allocating temporary strings:
https://github.com/jekyll/jekyll/blob/7e0f907c6452da3e1364be9c0c873c03067e4db0/lib/jekyll/frontmatter_defaults.rb#L159-L161


ashmaroli:
### Third Party Repo Profile Summary

```diff
--- master
+++ PR

- Total allocated: 296.75 MB (2461825 objects)
+ Total allocated: 295.90 MB (2440529 objects)
  Total retained:  43.43 MB (112686 objects)
```

DirtyF (commented):
I don't see any clear gain here: https://github.com/jekyll/jekyll/runs/1129387000?check_suite_focus=true#step:7:18

ashmaroli:
I've restarted the check-run. Let's see if that makes any difference :wink:

ashmaroli:
The new check run has different timestamps:
```diff
--- https://github.com/jekyll/jekyll/runs/1129387000?check_suite_focus=true#step:7:18
+++ https://github.com/jekyll/jekyll/runs/1129680368?check_suite_focus=true#step:7:18

-                    done in 32.912 seconds.
-                    done in 36.238 seconds.
-                    done in 34.763 seconds.
+                    done in 29.318 seconds.
+                    done in 30.575 seconds.
+                    done in 30.011 seconds.
```
But the memory usage remains the same.

The takeaway is that the *build* job is an unreliable metric :sweat_smile:

ashmaroli:
@jekyllbot: merge +fix

PR #8400, opened 2020-09-18
Reduce allocation from normalize_whitespace filter

ashmaroli:

## Summary

Chaining multiple non-mutating methods result in unnecessary duplication.
The solution is to duplicate input once and mutate the result thereafter.


DirtyF (approved):
LGTM

ashmaroli:
Thanks.
@jekyllbot: merge +fix

PR #8401, opened 2020-09-18
Check if site is in incremental mode optimally

ashmaroli:

## Summary

`Jekyll::Site#incremental?` takes an optional argument which is an empty Hash by default:
https://github.com/jekyll/jekyll/blob/f8c72089dd9591d1654f10dedd819536d2674c7c/lib/jekyll/site.rb#L367-L369

Since the empty hash is a default parameter, every call to the method allocates an empty Hash.
Therefore, calling the method while rendering an *include file* can allocate numerous Hashes unnecessarily.

The solution is then to access the config value directly.

### DIFF
--- a/lib/jekyll/tags/include.rb
+++ b/lib/jekyll/tags/include.rb
@@ -201,7 +201,7 @@ module Jekyll
         @site.inclusions[file] ||= locate_include_file(file)
         inclusion = @site.inclusions[file]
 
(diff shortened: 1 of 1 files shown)


DirtyF (approved):

ashmaroli:
Thank you @DirtyF :) 
@jekyllbot: merge +fix

PR #8404, opened 2020-09-20
Enable `Performance/ChainArrayAllocation` cop

ashmaroli:
## Summary

RuboCop Performance has a cop that allows detecting method chains that duplicate starting array with each chain segment.
But the cop is disabled by default.

This pull request enables the cop and fixes offenses reported by it by switching non-mutating methods with mutating counterparts.

### DIFF
--- a/lib/jekyll/drops/drop.rb
+++ b/lib/jekyll/drops/drop.rb
@@ -85,12 +85,8 @@ module Jekyll
       # Returns an Array of strings which represent method-specific keys.
       def content_methods
         @content_methods ||= (
-          self.class.instance_methods \
-            - Jekyll::Drops::Drop.instance_methods \
-            - NON_CONTENT_METHODS
-        ).map(&:to_s).reject do |method|
-          method.end_with?("=")
-        end
+          self.class.instance_methods - Jekyll::Drops::Drop.instance_methods - NON_CONTENT_METHODS
+        ).map!(&:to_s).tap { |result| result.reject! { |method| method.end_with?("=") } }
       end
 
       # Check if key exists in Drop
--- a/lib/jekyll/filters/url_filters.rb
+++ b/lib/jekyll/filters/url_filters.rb
@@ -76,13 +76,16 @@ module Jekyll
 
         parts = [sanitized_baseurl, input]
         Addressable::URI.parse(
-          parts.compact.map { |part| ensure_leading_slash(part.to_s) }.join
+          parts.map! { |part| ensure_leading_slash(part.to_s) }.join
         ).normalize.to_s
       end
 
       def sanitized_baseurl
         site = @context.registers[:site]
-        site.config["baseurl"].to_s.chomp("/")
+        baseurl = site.config["baseurl"]
+        return "" if baseurl.nil?
+
+        baseurl.to_s.chomp("/")
       end
 
       def ensure_leading_slash(input)
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -147,7 +147,11 @@ module Jekyll
 
     # The path to the page source file, relative to the site source
     def relative_path
-      @relative_path ||= File.join(*[@dir, @name].map(&:to_s).reject(&:empty?)).sub(%r!\A/!, "")
+      @relative_path ||= begin
+        parts = [@dir, @name]
+        parts.map!(&:to_s).reject!(&:empty?)
+        File.join(*parts).tap { |result| result.sub!(%r!\A/!, "") }
+      end
     end
 
     # Obtain destination path.
--- a/lib/jekyll/renderer.rb
+++ b/lib/jekyll/renderer.rb
@@ -36,9 +36,7 @@ module Jekyll
     #
     # Returns Array of Converter instances.
     def converters
-      @converters ||= begin
-        site.converters.select { |c| c.matches(document.extname) }.tap(&:sort!)
-      end
+      @converters ||= site.converters.select { |c| c.matches(document.extname) }.tap(&:sort!)
     end
 
     # Determine the extname the outputted file should have
--- a/lib/jekyll/renderer.rb
+++ b/lib/jekyll/renderer.rb
@@ -36,7 +36,9 @@ module Jekyll
     #
     # Returns Array of Converter instances.
     def converters
-      @converters ||= site.converters.select { |c| c.matches(document.extname) }.sort
+      @converters ||= begin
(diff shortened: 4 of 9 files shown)


RamboCacing666:
> ## Summary
> RuboCop Performance has a cop that allows detecting method chains that duplicate starting array with each chain segment.
> But the cop is disabled by default.
> 
> This pull request enables the cop and fixes offenses reported by it by switching non-mutating methods with mutating counterparts.

CONTRIBUTING.md

DirtyF (approved):
Tests pass, looks like it's safe to mutate in those methods.
The gain is hard to notice in our test suite.

ashmaroli:
Thanks for the review @Dirty :)
When you say *The gain is hard to notice..*, if you meant speed-wise, then yeah, this is a micro-optimization. Any speed gain will be shadowed by a drain elsewhere. However, if you meant allocation-wise, our docs site is a small sample. The third-party repo workflow shows an approx. gain of 2MB.

That said, I'll wait to see if @parkr or @mattr- sees something we missed or overlooked

lib/jekyll/drops/drop.rb
```diff
@@ -85,12 +85,8 @@ def []=(key, val)
...
       def content_methods
         @content_methods ||= (
-          self.class.instance_methods \
-            - Jekyll::Drops::Drop.instance_methods \
-            - NON_CONTENT_METHODS
-        ).map(&:to_s).reject do |method|
-          method.end_with?("=")
-        end
+          self.class.instance_methods - Jekyll::Drops::Drop.instance_methods - NON_CONTENT_METHODS
+        ).map!(&:to_s).tap { |result| result.reject! { |method| method.end_with?("=") } }
```
parkr, line 89:
```suggestion
        ).map!(&:to_s).reject! { |method| method.end_with?("=") }
```

What's the reason to use `#tap` here? I think we can remove it now.

lib/jekyll/filters/url_filters.rb
```diff
@@ -76,13 +76,16 @@ def compute_relative_url(input)
...
         ).normalize.to_s
       end
 
       def sanitized_baseurl
         site = @context.registers[:site]
-        site.config["baseurl"].to_s.chomp("/")
+        baseurl = site.config["baseurl"]
+        return "" if baseurl.nil?
+
+        baseurl.to_s.chomp("/")
```
parkr, line 88: site.config["baseurl"] isn't going to change so we could memoize this whole method, right?

lib/jekyll/page.rb
```diff
@@ -147,7 +147,11 @@ def path
 
     # The path to the page source file, relative to the site source
     def relative_path
-      @relative_path ||= File.join(*[@dir, @name].map(&:to_s).reject(&:empty?)).sub(%r!\A/!, "")
+      @relative_path ||= begin
+        parts = [@dir, @name]
+        parts.map!(&:to_s).reject!(&:empty?)
+        File.join(*parts).tap { |result| result.sub!(%r!\A/!, "") }
```
parkr, line 153: This method is definitely less readable to me than the previous iteration. Could we do something like:

```ruby
def relative_path
  @relative_path ||= Addressable::URI.normalize_path("#{@dir}/#{@name}").sub(%r!\A/!, "")
end
```

Here's it in action:

```ruby
>> @dir = nil
=> nil
>> @name = "index"
=> "index"
>> Addressable::URI.normalize_path("#{@dir}/#{@name}").sub(%r!\A/!, "")
=> "index"
>> File.join(*[@dir, @name].map(&:to_s).reject(&:empty?)).sub(%r!\A/!, "")
=> "index"
>> @dir = "docs/some_subdir"
=> "docs/some_subdir"
>> Addressable::URI.normalize_path("#{@dir}/#{@name}").sub(%r!\A/!, "")
=> "docs/some_subdir/index"
>> File.join(*[@dir, @name].map(&:to_s).reject(&:empty?)).sub(%r!\A/!, "")
=> "docs/some_subdir/index"
```

lib/jekyll/renderer.rb
```diff
@@ -36,7 +36,9 @@ def layouts
     #
     # Returns Array of Converter instances.
     def converters
-      @converters ||= site.converters.select { |c| c.matches(document.extname) }.sort
+      @converters ||= begin
```
parkr, line 39: We don't need the `begin` here. I think we can do `\` instead and get rid of the `begin ... end`

parkr (commented):
This is looking good! I left a couple comments about some optimizations we might be able to make here.

lib/jekyll/filters/url_filters.rb
ashmaroli: Yes, we could memoize this. Great idea! :+1:

lib/jekyll/drops/drop.rb
ashmaroli: Since [`Array#reject!`](https://ruby-doc.org/core-2.4.0/Array.html#method-i-reject-21) can return `nil` if the input array was changed, calling `#tap` ensures that the input array is returned.

lib/jekyll/page.rb
ashmaroli: Okay. I'll revert this change.
I won't be using Addressable to generate the path since it generates a lot of intermediate objects..

I'll benchmark it and verify..

parkr: I guess I'm looking to remove the need for the array. Maybe we do `"#{@dir}/#{@name}"` and remove the leading slash and remove all double slashes?

lib/jekyll/drops/drop.rb
parkr: Gotcha! Thanks for the explanation – the bang (!) methods are oddities. 😄

lib/jekyll/page.rb
ashmaroli: @parkr Understood. I'll open a new PR using `Jekyll::PathManager.join` and ping you.
(A dedicated PR allows more of associated discussion and adding more tests without bloating the scope of this PR..)

ashmaroli: Handling in #8408

parkr (approved):

ashmaroli:
Requires #8408 to be reviewed and merged first...

lib/jekyll/page.rb
ashmaroli: Resolved via #8408

ashmaroli:
@jekyllbot: merge +fix

PR #8406, opened 2020-09-21
Reduce allocations from rendering item as liquid

ashmaroli:

## Summary

Instead of using `#map` recursively, iterate through using `#each_with_object`.


DirtyF (approved):

DirtyF:
@jekyll: merge +dev

PR #8407, opened 2020-09-22
Clarify gem installation instructions

m-naumann:

## Summary

I made clearer that not Ruby itself but RubyGems packages should not be installed as root.

## Context

Small change, no issue opened.


DirtyF (approved):
👌

Issue #6448, opened 2017-10-19
Document differences between `highlight` tag and Markdown syntax highlighting


pathawks:
As pointed out in #6446, using the `{% highlight %}` tag will result in output that is different from using triple backticks in Markdown.

* If this is intentional, it should be documented.
* If this is not intentional, it should be fixed.

I kind of think this is _not_ intentional, as our redcarpet parser has code to try to normalize highlighted code:
https://github.com/jekyll/jekyll/blob/28af8bd85f832ddea1a761f0d109ddcc2ea5f903/lib/jekyll/converters/markdown/redcarpet_parser.rb#L5-L13

I don't think it would be too difficult to add some code to our Markdown processors that intercepted triple backtick code blocks and highlighted them as if they had used the `{% highlight %}` tag, but perhaps this is out of scope?

(I could have _sworn_ that we did this at one point, but I can't find any evidence of that now. Must have been a fever dream.)

/cc: @jekyll/build @jekyll/stability

DirtyF:
I tend to favor Markdown over Liquid when writing technical documentation, but using one or the other flavor shouldn't change the final output.

> Usually a <figure> is an image, illustration, diagram, code snippet, etc., that is referenced in the main flow of a document, but that can be moved to another part of the document or to an appendix without affecting the main flow. — [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure)

This is a choice we should leave to our users, and thus remove the default` <figure>` tag for code snippets.

pathawks:
Related: #3738 #3738

zachattack:
Just a drifter, but I would like this to be unified also. I am seeing the style to reset blink which I think is a symptom of the liquid vs markdown implementations for highlighting.
![2017-11-08 13 54 51](https://user-images.githubusercontent.com/210152/32571286-83532ace-c48c-11e7-9924-2cff19fc67c4.gif)

DirtyF:
After reading the story behind the differences in previous issues:

* `{% highlight %}` will output a `<figure>` element (because why not?)
* triple fenced code blocks in Markdown won't.

As we don't want to break existing websites, and strongly adhere to semver, this behavior is not gonna change before next major version.

We can add a note to the documenttion for now and refactor our code snippets in v4.0 so that they all generate the same output.

pathawks:
Too, I suppose, different MarkDown processors could render code blocks differently. We should just be very clear about the behavior of the `highlight` tag 👍🏼

[...]

PR #8412, opened 2020-09-25
Introduce new Liquid tag for rendering code blocks

### LINKED ISSUE #8411: Highlighted code with line numbers renders as non-conforming HTML
[An issue](https://github.com/pmarsceill/just-the-docs/issues/286) raised by users of of the [Just the Docs](https://pmarsceill.github.io/just-the-docs/) theme is closely related to Jekyll issue https://github.com/jekyll/jekyll/issues/4432. It would be much appreciated if Jekyll could fix the cause of the issue, since the available [work-arounds](https://pmarsceill.github.io/just-the-docs/docs/ui-components/code/line-numbers/#workarounds) are tedious to use.

## Validation errors

The Jekyll documentation states that [line numbers can be added to highlighted code snippets](https://jekyllrb.com/docs/liquid/tags/#line-numbers), giving the following example:
```markdown
{% highlight ruby linenos %}
def foo
  puts 'foo'
end
{% endhighlight %}
```
`jekyll new` currently renders the above as:
```html
<figure class="highlight"><pre><code class="language-ruby" data-lang="ruby"><table class="rouge-table"><tbody><tr><td class="gutter gl"><pre class="lineno">1
2
3
</pre></td><td class="code"><pre><span class="k">def</span> <span class="nf">foo</span>
  <span class="nb">puts</span> <span class="s1">'foo'</span>
<span class="k">end</span>
</pre></td></tr></tbody></table></code></pre></figure>
```
(The differences from the rendered HTML reported in https://github.com/jekyll/jekyll/issues/4432 are not significant for this issue.)

[HTML validation](https://validator.w3.org) reports that the above HTML is non-conforming:

> **Error**: **Element [`table`](https://html.spec.whatwg.org/multipage/#the-table-element) not allowed as child of element [`code`](https://html.spec.whatwg.org/multipage/#the-code-element) in this context. (Suppressing further errors from this subtree.)**

Similar validation errors are reported when using fenced code snippets in Markdown with kramdown's [option to add line numbers globally](https://github.com/mmistakes/minimal-mistakes/issues/1437#issuecomment-354797503).

The HTML produced for highlighted code with line numbers is also problematic when using the [Jekyll layout for compressing HTML](http://jch.penibelst.de), which [relies on `pre` tags not being nested](https://github.com/penibelst/jekyll-compress-html/issues/71#issuecomment-172069842) (and conforming HTML does not allow nested `pre` elements).

To produce conforming HTML from highlighted code with line numbers, it seems that users currently need to resort to tedious [work-arounds](https://pmarsceill.github.io/just-the-docs/docs/ui-components/code/line-numbers/#workarounds) supported by [Liquid code](https://github.com/pmarsceill/just-the-docs/blob/master/_includes/fix_linenos.html).

## Styling

In the blog created by `jekyll new`, the code snippet in the welcome message appears as:
Screenshot 2020-09-24 at 17 28 13
When the `linenos` option is added, it appears as:
Screenshot 2020-09-24 at 17 31 43
To improve the formatting (removing the table margins when tables are children of code elements), theme styling needs to target the non-conforming HTML produced by Jekyll, and hope that browsers display it consistently (see this [Just the Docs example](https://pmarsceill.github.io/just-the-docs/docs/ui-components/code/line-numbers/#examples)).

## How to fix?

The origin of the non-conforming HTML appears to be the following [Ruby code](https://github.com/jekyll/jekyll/blob/fd8a7e7e3b92f87a247dcc51bbd99c977a1e5d83/lib/jekyll/tags/highlight.rb#L98):
```ruby
      def add_code_tag(code)
        code_attributes = [
          "class=\"language-#{@lang.to_s.tr("+", "-")}\"",
          "data-lang=\"#{@lang}\"",
        ].join(" ")
        "<figure class=\"highlight\"><pre><code #{code_attributes}>"\
        "#{code.chomp}</code></pre></figure>"
      end
```
That produces conforming HTML only when `#{code.chomp}` is phrasing content. I am not a Ruby programmer, but presumably it would be easy to test if `#{code.chomp}` is a table element, and suppress the wrapping in `pre` and `code` elements; and perhaps the `#{code_attributes}` are superfluous?

Of course, changing the HTML produced from highlighted code with line numbers will inevitably break theme styling that relies on the non-conforming nesting currently produced by Jekyll.

ashmaroli:
- This is a 🙋 feature.

## Summary

The existing tag `{% highlight %}` renders non-conforming HTML when the tag is configured to render line numbers as well.
In order to permit backwards-compatibility, a new tag is being introduced to resolve the situation.

Notable characteristics:
- **name:**  `codeblock`
- **type:**  Liquid Block
- **usage:** `{% codeblock <lang> [<options>] [annotated] %}...{% endcodeblock %}`
- The tag always renders line-numbers (non-configurable).
- The tag is hard-coded to highlight code with `rouge` only and utilizes the `Rouge::Formatters::HTMLLineTable` formatter.
- The `<options>` (of the form `key=value` or `key = value`) can be used to configure the `Rouge::Formatters::HTMLLineTable` formatter.
- If the `annotated` attribute is present, the code-block contains an extra `<div>` element rendering the value of `<lang>`.

## Context

Resolves #8411

### DIFF
--- a/lib/jekyll/tags/code_block.rb
+++ b/lib/jekyll/tags/code_block.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+require "rouge"
+
+module Jekyll
+  module Tags
+    class CodeBlock < Liquid::Block
+      EMPTY_OPTIONS  = {}.freeze
+      HTML_FORMATTER = Rouge::Formatters::HTML.new
+      MARKUP_SYNTAX  = %r!\s*([\w-]+)(?:\s+(.*?))?\s+(annotated)?!.freeze
+      OPTIONS_SYNTAX = %r!
+        ([\w-]+)\s*=\s*
+        (?:"([^"\\]*(?:\\.[^"\\]*)*)"|'([^'\\]*(?:\\.[^'\\]*)*)'|([\w.-]+))
+      !x.freeze
+
+      private_constant :HTML_FORMATTER, :MARKUP_SYNTAX, :EMPTY_OPTIONS, :OPTIONS_SYNTAX
+
+      def initialize(tag_name, markup, tokens)
+        super
+
+        if markup =~ MARKUP_SYNTAX
+          @lang, @options, @annotated = Regexp.last_match.captures
+        else
+          @lang = "plaintext"
+          @options = nil
+          @annotated = nil
+        end
+      end
+
+      def render(context)
+        code = super.to_s
+        code.strip!
+
+        prefix    = context["highlighter_prefix"] || ""
+        suffix    = context["highlighter_suffix"] || ""
+        options   = @options ? parse_options(context) : EMPTY_OPTIONS
+        formatter = Rouge::Formatters::HTMLLineTable.new(HTML_FORMATTER, options)
+        lexer     = Rouge::Lexer.find_fancy(@lang, code) || Rouge::Lexers::PlainText
+        body      = wrap_formatted(formatter.format(lexer.lex(code)), options[:caption])
+
+        "#{prefix}#{body}#{suffix}"
+      end
+
+      private
+
+      def parse_options(context)
+        {}.tap do |options|
+          @options.scan(OPTIONS_SYNTAX) do |key, d_quoted, s_quoted, variable|
+            value = if d_quoted
+                      d_quoted.include?('\\"') ? d_quoted.gsub('\\"', '"') : d_quoted
+                    elsif s_quoted
+                      s_quoted.include?("\\'") ? s_quoted.gsub("\\'", "'") : s_quoted
+                    elsif variable
+                      context[variable]
+                    end
+
+            options[key.to_sym] = value
+          end
+        end
+      end
+
+      def wrap_formatted(code, caption)
+        buffer = +""
+        buffer << %(<div class="highlighter language-#{@lang} highlighter-rouge">)
+        buffer << %(  <div data-code-block-lang="#{@lang}">#{code}</div>) if @annotated
+        buffer << code
+        buffer << %(  <div class="code-block-caption">#{caption}</div>) if caption
+        buffer << "</div>"
+        buffer
+      end
+    end
+  end
+end
+
+Liquid::Template.register_tag "codeblock", Jekyll::Tags::CodeBlock


ashmaroli:
/cc @pdmosses if you would like to take this on a test-drive and provide feedback.
To test, point Gemfile to this pull request:
```ruby
# Gemfile
gem 'jekyll', github: 'jekyll/jekyll', ref: 'refs/pull/8412/head'
```

pdmosses:
@ashmaroli Thanks for the quick work on this issue, and for the test-drive opportunity!

The removal of the `pre` and `code` element wrapping has addressed the reported validation error, as expected. 

However, when there is more than one `{% codeblock <lang> [<options>] [annotated] %}...{% endcodeblock %}` on the same page, HTML validation reports errors on duplicate `id`s, e.g.:
```
Error: Duplicate ID line-1.
From line 67, column 97; to line 67, column 127
e"><tbody><tr id="line-1" class="lineno"><td cl
```
Try uploading an HTML file rendered from the following markdown to https://validator.w3.org/nu/#file:
```
{% codeblock ruby %}
def print_hi(name)
  puts "Hi, #{name}"
end
print_hi('Tom')
{% endcodeblock %}

{% codeblock ruby %}
def foo
  puts 'foo'
end
{% endcodeblock %}

```

pdmosses:
As more general feedback, I don't find the tag `codeblock` suggestive of automatic line numbers, nor that it is a variant of `highlight`. 

Perhaps `codeblock` could have the same interface as `highlight`, with a `linenos` option? When used without `linenos` it would be equivalent to using `highlight`.

An alternative might be to introduce a new option (say `line-numbers`) that renders the same as the current `codeblock`, and make the `linenos` option deprecated.

ashmaroli:
Thank you for the quick feedback, @pdmosses.

> HTML validation reports errors on duplicate `id`s

Ouch! I did not forsee this when I implemented the formatter in Rouge.. Hmm.., looks like I'll have to implement a custom Rouge formatter in Jekyll directly as well.. That'll present the opportunity for greater control over the result..

> I don't find the tag `codeblock` suggestive of automatic line numbers

Agreed. I don't like the name either. I could come up with an apt name yet be short for easier usage.
Do you have any suggestions?

> Perhaps `codeblock` could have the same interface as `highlight`, with a `linenos` option?

It could. However, the intention here was to limit computation to just the needful.
`highlight` is the superior version that caters to a wider user-base.

> An alternative might be to introduce a new option (say `line-numbers`) for `highlight`

Having a new tag altogether allows large-scale changes:
- Replace `<figure>` wrapper tags with `<div>` wrapper similar to the markup rendered by kramdown.
- Do away with having `<pre><code>` wrappers entirely without incurring end-user confusion / end-user maintenance overhead.
- Appease users who prefer existing non-conforming markup to avoid having to adjust stylesheets and scripts; and users who prefer conforming markup.

That said, thank you very much for your suggestions. Much appreciated.

pdmosses:
@ashmaroli indeed, a new tag has significant advantages. Uniformity with the HTML kramdown generates from fenced code would be particularly useful.

However, I still think it would be good for the new tag to omit line numbers by default. That would make it more analogous to `highlight`. And it would avoid users forgetting to change the end tag when switching line numbers on or off (as I did myself when changing from `highlight` to `codeblock` for the test drive...). The option to include them could then be `linenos`, or you might follow kramdown/Rouge and use `line_numbers`. Having to use `highlight` to avoid line numbers could also be error-prone unless the remaining options of the two tags are consistent.

The tag name `codeblock` isn't bad, but `code` would be shorter 😃 and equally suggestive...

ashmaroli:
Summary of recent changes:
- **name:** `rougify`
- **usage:** `{% rougify <lang> [<options>] %}...{% endrougify %}
- notes:
  - Takes optional `[<options>]` only as key-value pairs similar to Jekyll's `{% include %}` tag.
  - Line numbers are not rendered by default. Requires passing `linenos = true` as option.
  - Similarly, `annotation = true` to render a language-label above the code-block.
  - A caption for the code-block can be rendered below the code-block when passed as `caption = "My Caption"` option.
  - The formatter that renders the table with line-numbers can be customized by passing relevant options.

ashmaroli:
@pdmosses I've tweaked the new tag a bit. (Details in [this comment](https://github.com/jekyll/jekyll/pull/8412#issuecomment-699608528)). Are you willing to take it for another test-drive?

pdmosses:
@ashmaroli the updated tag looks good (and validates OK).

 However, the rendered HTML isn't as close as I was expecting to that produced by kramdown with the options:
```yaml
kramdown:
  syntax_highlighter_opts:
    block:
      line_numbers: true
```
This source:
````markdown
ROUGIFY:

{% rougify ruby linenos=true %}
def foo
  puts 'foo'
end
{% endrougify %}

KRAMDOWN:

```ruby
def foo
  puts 'foo'
end
```
````
rendered as:
```html
<p>ROUGIFY:</p>

<div class="language-ruby highlighter-rouge">
    <div class="highlight">
        <table class="rouge-line-table">
            <tbody>
                <tr data-line-id="line-1" class="lineno">
                    <td class="rouge-gutter gl" style="-moz-user-select: none;-ms-user-select: none;-webkit-user-select: none;user-select: none;">
                        <pre>1</pre>
                    </td>
                    <td class="rouge-code">
                        <pre>
                            <span class="k">def</span>
                            <span class="nf">foo</span>
                        </pre>
                    </td>
                </tr>
                <tr data-line-id="line-2" class="lineno">
                    <td class="rouge-gutter gl" style="-moz-user-select: none;-ms-user-select: none;-webkit-user-select: none;user-select: none;">
                        <pre>2</pre>
                    </td>
                    <td class="rouge-code">
                        <pre>
                            <span class="nb">puts</span>
                            <span class="s1">'foo'</span>
                        </pre>
                    </td>
                </tr>
                <tr data-line-id="line-3" class="lineno">
                    <td class="rouge-gutter gl" style="-moz-user-select: none;-ms-user-select: none;-webkit-user-select: none;user-select: none;">
                        <pre>3</pre>
                    </td>
                    <td class="rouge-code">
                        <pre>
                            <span class="k">end</span>
                        </pre>
                    </td>
                </tr>
            </tbody>
        </table>
    </div>
</div>

<p>KRAMDOWN:</p>

<div class="language-ruby highlighter-rouge">
    <div class="highlight">
        <pre class="highlight">
            <code>
                <table class="rouge-table">
                    <tbody>
                        <tr>
                            <td class="rouge-gutter gl">
                                <pre class="lineno">1
                                2
                                3
                                </pre>
                            </td>
                            <td class="rouge-code">
                                <pre>
                                    <span class="k">def</span>
                                    <span class="nf">foo</span>
                                    <span class="nb">puts</span>
                                    <span class="s1">'foo'</span>
                                    <span class="k">end</span>
                                </pre>
                            </td>
                        </tr>
                    </tbody>
                </table>
            </code>
        </pre>
    </div>
</div>
```
Both the class and the structure of the rendered tables look quite different. 

Moreover, the `rougify` version seems somewhat bigger, which might be an issue for sites with a lot of long code listings.

Perhaps issue #6448 is relevant also for line numbers?

ashmaroli:
Thank you for your feedback @pdmosses 


> the rendered HTML isn't as close as I was expecting..  Both the class and the structure of the rendered tables look quite different.

Yes, this difference was intentional / by design &mdash; each line of code and the associated line number is on a separate table-row; the class name being different so users can style the new table without affecting the ones from `{% highlight %}` or `kramdown`.

pdmosses:
@ashmaroli thanks for the explanation. Using a different class name is surely the best when the table structure is different. But I was hoping the table structure could be the same as for kramdown's code fences, to *allow* themes to use exactly the same CSS for styling code with line numbers, regardless of whether the HTML is rendered by `rougify` or kramdown. 

The new tag is rendering *exactly* the same HTML as kramdown for code without line numbers, thus addressing #6448 for that case. Are there advantages of introducing the new table structure for code with line numbers that outweigh the disadvantage of differing from kramdown?

pdmosses:
Thanks for the documentation preview, which looks good.

Following the instructions there for [Stylesheets for syntax highlighting](https://deploy-preview-8412--jekyllrb.netlify.app/docs/liquid/tags/#stylesheets-for-syntax-highlighting), I added `native.css` to the site created by `jekyll new`. The result of `rougify` without line numbers looks fine – but with the `linenos=true` option I get this:

Screenshot 2020-09-30 at 19 00 08

Is there a CSS file somewhere that one can add to eliminate the superfluous table margins in the rougified code, and produce results like those shown in [the docs](https://deploy-preview-8412--jekyllrb.netlify.app/docs/liquid/tags/#enhanced-code-blocks) when using `rougify`? If so, I think it would be helpful to add a link to it.

ashmaroli:
@pdmosses Thank you for the feedback and for being very accommodating.

Regarding your query on the styles, unfortunately, the only solution is to manually define styles for the currently rendered markup.

I understand that the theme you're using is Minima which is what you get when you run `jekyll new`. However, since the theme is a separate project from Jekyll, the theme may only get updated once the implementation here gets stabilized, accepted and merged / released publicly.

That said, I can always push a branch to my fork of the Minima repository with necessary changes that you can use via the 
`jekyll-remote-theme` plugin or by pointing your Gemfile directly to the branch..
Stay tuned :)

pdmosses:
@ashmaroli In fact I'm not intending to use the Minima theme myself for highlighting code on websites: issue #8411 is theme-independent, and it was easiest to illustrate it using Minima. 

I understand that Minima will not get updated straight away with styling for the HTML rendered by the new tag. However, `jekyll new` creates a site based on Minima; surely it could add a style file that customises the existing Minima styling to give sensible layout for rougified code with line numbers? 

Then the bit in the Jekyll welcome message about "Jekyll also offers powerful support for code snippets" could be expanded with illustrations of the use of `rougify`. Otherwise, new users might follow the link to the Jekyll docs, find the section about enhanced code blocks, and be quite disappointed when the example with line numbers in the docs doesn't produce a sensible result...

I don't have much (recent) experience with CSS, but I managed to customise Minima with a file `rouge-line-table.scss` (suggestions for improvement are welcome):
```css
table.rouge-line-table {
  padding: ($spacing-unit / 3) ($spacing-unit / 2);
  margin: 0;
  border: 0;
  border-collapse: separate;
  -webkit-border-vertical-spacing: 0;

  td,
  pre {
    padding: 0;
    margin: 0;
    background-color: #eef;
    border: 0;
  }

  td.gl {
    width: 4ex;
    padding-right: $spacing-unit;
    text-align: right;
  }
}
```
That produces results like this:

Screenshot 2020-10-01 at 16 28 51

ashmaroli:
@pdmosses That's a nice attempt.
If you're interested, I have pushed a branch of my take to my fork of Minima. You can try it by editing your Gemfile as:
```ruby
gem "minima", github: "ashmaroli/minima", branch: "rougify-lineno-styles-for-v2.x"
```

pdmosses:
@ashmaroli The styling using your fork of Minima looks very good! (I'll edit out my clumsy attempt above.)

BTW, the new tag renders the following HTML for a line number:

```html
<td class="rouge-gutter gl" style="-moz-user-select: none;-ms-user-select: none;-webkit-user-select: none;user-select: none;"><pre>1</pre></td>
```

Wouldn't it be cleaner to omit the `style` from the HTML, and leave it to the theme/user styling to specify whether to exclude line numbers when selecting? That might also make the HTML for long listings significantly smaller...

ashmaroli:
> The styling using your fork of Minima looks very good!

Thank you.

>  (I've edited out my clumsy attempt above.)

I did not mean to belittle your attempt in any way. Just sharing..

> Wouldn't it be cleaner to omit the `style` from the HTML, and leave it to the theme/user

Yes, it would. I shall remove it.

PR #8408, opened 2020-09-24
Compute relative_path of pages using PathManager

ashmaroli:
- This is a 🔨 code refactoring change.
- I've added tests.

## Summary

`File.join` is sensitive to `nil` arguments. So we need to coerce arguments to String to be safer.
Using `Jekyll::PathManager.join` on the other hand allows optimized execution for same results.

## Context

Prior discussions: https://github.com/jekyll/jekyll/pull/8404#discussion_r494402547

PR #8415, opened 2020-09-27
Handle nil argument to Jekyll.sanitized_path

ashmaroli:
- This is a 🐛 bug fix.
- I've added tests.

## Summary

Bail early if the questionable path to be prefixed is `nil`.

## Context

Discovered while implementing #8408


DirtyF (approved):

ashmaroli:
@jekyllbot: merge +fix

PR #8418, opened 2020-09-28
Fix test suite compatibility with JRuby

ashmaroli:
## Summary

Lock to an older version of `rubocop-ast` on JRuby since the `v0.7.0` has a dependency on a native extension not compatible with JRuby.


ashmaroli:
While this is a valid solution for the time being, if a future version of `rubocop` requires at least `rubocop-ast-0.7.0`, we'll have to lock rubocop version as well on JRuby.

DirtyF (approved):
Thanks

DirtyF:
@jekyll: merge +dev

PR #8427, opened 2020-10-08
Use Regexp#match? when MatchData is not required

ashmaroli:

## Summary

`Regexp#match?` is the faster alternative from Ruby 2.4 when the `MatchData` object is not consequently utilized.

While the first build after running `jekyll clean` won't have files in the dest-dir, consequent builds will have the directory populated. So this line could get executed numerous times especially in a site with numerous pages and files


DirtyF (approved):
All green, jekyll site seems to compile a little bit faster, memory allocation is still mostly Kramdown and Nokogiri.

ashmaroli:
> memory allocation is still mostly kramdown and nogikiri

Allocations from Nokogiri will reduce when they release `v1.11.0`..

ashmaroli:
@jekyllbot: merge +fix

PR #8430, opened 2020-10-08
chore(deps): bump Rubocop to 0.93.0

DirtyF:
Failed to fix the safe navigation warnings on convertible and page 😊


ashmaroli:
On `master` the `Convertible#validate_permalink!` method is just trying to weed out empty permalinks:
https://github.com/jekyll/jekyll/blob/01d0b78abb557d67c66c2be65d30fcfa07c89974/lib/jekyll/convertible.rb#L71-L75

So that can be simplified into:
```ruby
    def validate_permalink!(filename)
      if self.data["permalink"] == ""
        raise Errors::InvalidPermalinkError, "Invalid permalink in #{filename}"
      end
    end
```
The other case where `self.data["permalink"].to_s.empty?` would be true is if `self.data["permalink"]` is `nil` which is a valid value. So it should be ignored.

Regarding `Jekyll::Page`, the patch would be:
```diff
     def excerpt
       return @excerpt if defined?(@excerpt)

-      @excerpt = data["excerpt"]&.to_s
+      @excerpt = data["excerpt"]
     end
 
     private
 
     def generate_excerpt
       return unless generate_excerpt?
 
-      data["excerpt"] ||= Jekyll::PageExcerpt.new(self)
+      data["excerpt"] ||= Jekyll::PageExcerpt.new(self).to_s
```

ashmaroli (approved):

ashmaroli:
Thank you @DirtyF 
@jekyllbot: merge +dev

PR #8437, opened 2020-10-12
Require Ruby version 2.5+ in gemspec

toshimaru:
## Summary

Require Ruby version 2.5+

## Context

Jekyll doesn't work for Ruby v2.4 since it requires rubygems v2.7+.

```console
$  ruby -v
ruby 2.4.10p364 (2020-03-31 revision 67879) [x86_64-darwin19]

$ cat Gemfile
source 'https://rubygems.org'
ruby '2.4.10'
gem 'jekyll', '4.1.1'

$ bundle install
Fetching gem metadata from https://rubygems.org/..........
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...
Bundler found conflicting requirements for the RubyGems version:
  In Gemfile:
    RubyGems (= 2.6.14.4)

    jekyll (= 4.1.1) was resolved to 4.1.1, which depends on
      RubyGems (>= 2.7.0)
```

### DIFF
--- a/jekyll.gemspec
+++ b/jekyll.gemspec
@@ -29,7 +29,7 @@ Gem::Specification.new do |s|
   s.rdoc_options     = ["--charset=UTF-8"]
   s.extra_rdoc_files = %w(README.markdown LICENSE)
 
-  s.required_ruby_version     = ">= 2.4.0"
+  s.required_ruby_version     = ">= 2.5.0"
   s.required_rubygems_version = ">= 2.7.0"
 
   s.add_runtime_dependency("addressable",           "~> 2.4")


ashmaroli:
@toshimaru Before you spend more time on this, Jekyll won't be dropping support for Ruby 2.4 for the time being..
Perhaps when we decide to ship a v5.0...

(*Yes, I'm aware that Ruby 2.4 is EOL now.. and Ruby 2.5 will EOL in about 6 months from now..*)

toshimaru:
> Jekyll won't be dropping support for Ruby 2.4 for the time being..

@ashmaroli 
I've updated the description, I'd say Jekyll does not work for Ruby v2.4 since it requires rubygems 2.7.0+. (`required_rubygems_version = ">= 2.7.0"`) 🤔

ashmaroli:
> I'd say Jekyll does not work for Ruby v2.4

Could you post the entire contents of the above Gemfile here?

ashmaroli:
The problem is perhaps in the Rubygems version you are using..
Run `gem update --system` to update your Rubygems version.

toshimaru:
I've created the Pull Request since the default rubygem bundled in Ruby v2.4 doesn't work for Jekyll v4.
But yes, If we run `gem update --system`, it works without an issue.

Okay, Jekyll still supports Ruby v2.4. let's close this. 👍

PR #8444, opened 2020-10-22
Use Ruby 2.7.1 in GitHub Actions

ashmaroli:
## Summary

So that the profiling results
- can expose any deprecation warning not fixed yet.
- better reflect allocations by using newer Ruby versions.

*(Not using Ruby 2.7.2 just in case the deliberate breaking change in Ruby not abort the workflow prematurely)*

### DIFF
--- a/.github/workflows/third-party.yml
+++ b/.github/workflows/third-party.yml
@@ -25,7 +25,7 @@ jobs:
     - name: Set up Ruby
       uses: actions/setup-ruby@v1
       with:
-        ruby-version: 2.6.x
+        ruby-version: 2.7.1
     - name: Set up Dependencies Cache
       uses: actions/cache@v1
       with:


ashmaroli:
### Third-party Repository Memory Profile

```diff
--- master https://github.com/jekyll/jekyll/runs/1286705193
+++ PR     https://github.com/jekyll/jekyll/runs/1294108138

- Total allocated: 201.77 MB (1169764 objects)
- Total retained:  40.76 MB (130449 objects)
+ Total allocated: 193.21 MB (1180812 objects)
+ Total retained:  45.00 MB (122172 objects)
```

Strangely, while the **number of allocated objects increased**, the **size of memory allocated decreased**.
and while the **number of retained objects decreased**, the **size of memory retained increased**.

DirtyF (approved):

DirtyF:
@jekyll: merge +dev

PR #8450, opened 2020-10-29
docs: init project

spinningrachel:
Document update based on my experience using the tutorial.

## Context

Document update based on my experience using the tutorial.
I also recommend adding the attached screen capture directly where I updated the text.
gemfile


DirtyF (approved):
Thanks for the edit, amended as you can use bundler to generate the `Gemfile`

PR #8451, opened 2020-11-01
Memoize defaults computed for Convertibles

ashmaroli:

## Summary

Classes that mixin the `Jekyll::Convertible` module (primarily `Jekyll::Page`) doesn't have their `to_liquid` method memoized.
Therefore, every call to the `#to_liquid` method (esp. via the Liquid Filters `where`, `find`, `sort`) has their default front matter (which never changes for a given `site.config` hash) freshly computed.

### DIFF
--- a/lib/jekyll/convertible.rb
+++ b/lib/jekyll/convertible.rb
@@ -117,7 +117,6 @@ module Jekyll
           hsh[attribute] = send(attribute)
         end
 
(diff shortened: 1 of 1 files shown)


DirtyF (approved):
No noticeable change on a small website with defaults.

ashmaroli:
Thanks @DirtyF 
@jekyllbot: merge +fix

PR #8453, opened 2020-11-04
Reduce array allocations from merging categories

ashmaroli:

## Summary

Call *set union* on `data["categories"]` only if it is an Array instead of defaulting to a temporary empty array.


DirtyF (approved):

ashmaroli:
Thanks :)
@jekyllbot: merge +fix

PR #8454, opened 2020-11-04
Replace empty read-only hashes with a constant

ashmaroli:
## Summary

Use a constant in order to reduce allocations from multiple method calls


ashmaroli:
Not worth the added complexity..

PR #8456, opened 2020-11-05
docs: installation: fedora: add additional dependencies

parzival3:
On Fedora, add:
- openssl-devel 
- redhat-rpm-config

as default dependecies, there is no reason to not list them in the installation section since the build step depends on them.

## Summary
Add default build dependency on Fedora.


DirtyF (approved):
Thanks, I'll trust your on this  👍

DirtyF:
@jekyll: +docs

PR #8458, opened 2020-11-05
Memoize destination of pages, documents and staticfiles

ashmaroli:

## Summary

Because multiple calls to the `:destination` method with the same dest_dir (generally `@site.dest`) will result in unnecessary computations and allocations.


DirtyF (approved):
The test suite does not show a significant gain. :shrug:

ashmaroli:
> does not show a significant gain

Hence the label, `micro-optimization` :wink:
@jekyllbot: merge +fix

PR #8465, opened 2020-11-11
Allow triggering `:post_convert` events atomically

ashmaroli:
- This is a 🐛 bug fix.
- I've added tests.

## Summary

`Jekyll::Document` instances allows triggering all registered hook events for owner `:documents` as well as an individual collection:
https://github.com/jekyll/jekyll/blob/11ff8aa0dd610fa26689dc8d2d42decd3bba79ee/lib/jekyll/document.rb#L390-L393
But this support was not extended for the `post_convert` hook event on `master` (*hook event to be introduced in v4.2*).

This pull request fixes the above discrepancy.


DirtyF (approved):

DirtyF:
@jekyllbot: merge +fix

PR #8472, opened 2020-11-16
Explicitly return nil after site process phase

ashmaroli:
- This is a 🐛 bug fix.
- I've added tests.

## Summary

The various phases within `site.process` has `Returns nothing.` in the Ruby code documentation.
Yet, those phases when called individually returns something other than nothing.

This pull request fixes above by explicitly returning `nil`.

## Context

When toying with the site build process phases via IRB, various phases may output large amount of data in site with numerous pages and documents.


DirtyF (approved):
0️⃣

PR #8477, opened 2020-11-19
Make --limit-posts option faster

emad-elsaid:

## Summary

When `_posts` directory has too many files (mine 60k). running `jekyll server --limit-posts 10` will take around 16 seconds, I expected limiting the posts to makes it take shorter time.

After digging into the code I found that the post reader reads all files all the time, then limit them, that takes too long and we don't need to read more than the limit.

So my solution was to pass down the limit to the part that reads the posts from the directory and limit that.

This makes `jekyll server --limit-posts 10` in the same directory takes 3 seconds that's 5.3x faster.

### DIFF
--- a/lib/jekyll/reader.rb
+++ b/lib/jekyll/reader.rb
@@ -4,7 +4,7 @@ module Jekyll
   class Reader
     attr_reader :site, :limit
 
-    def initialize(site, limit)
+    def initialize(site, limit = 0)
       @site = site
       @limit = limit
     end
--- a/lib/jekyll/reader.rb
+++ b/lib/jekyll/reader.rb
@@ -132,14 +132,14 @@ module Jekyll
     # subfolder - The String representing the directory to read.
     #
     # Returns the list of entries to process
-    def get_entries(dir, subfolder, limit)
+    def get_entries(dir, subfolder, limit = 0)
       base = site.in_source_dir(dir, subfolder)
       return [] unless File.exist?(base)
 
       entries = Dir.chdir(base) { filter_entries(Dir["**/*"], base) }
       entries.delete_if { |e| File.directory?(site.in_source_dir(base, e)) }
 
-      entries_limit = [entries.length, limit].min
+      entries_limit = entries.length < limit ? entries.length : limit
       limit.positive? ? entries.last(entries_limit) : entries
     end
 
--- a/lib/jekyll/reader.rb
+++ b/lib/jekyll/reader.rb
@@ -2,7 +2,7 @@
 
 module Jekyll
   class Reader
-    attr_reader :site, :limit
+    attr_reader :site
 
     def initialize(site, limit = 0)
       @site = site
@@ -144,6 +144,8 @@ module Jekyll
     end
 
     private
+
+    attr_reader :limit
 
     # Internal
     #
--- a/lib/jekyll/reader.rb
+++ b/lib/jekyll/reader.rb
@@ -140,7 +140,7 @@ module Jekyll
       entries.delete_if { |e| File.directory?(site.in_source_dir(base, e)) }
 
       entries_limit = [entries.length, limit].min
-      limit.positive? ? entries[-entries_limit, entries_limit] : entries
+      limit.positive? ? entries.last(entries_limit) : entries
     end
 
     private
--- a/lib/jekyll/reader.rb
+++ b/lib/jekyll/reader.rb
@@ -2,10 +2,11 @@
 
 module Jekyll
   class Reader
-    attr_reader :site
+    attr_reader :site, :limit
 
-    def initialize(site)
+    def initialize(site, limit)
       @site = site
+      @limit = limit
     end
 
     # Read Site data from disk and load it into internal data structures.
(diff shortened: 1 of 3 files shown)


lib/jekyll/reader.rb
```diff
@@ -131,12 +132,15 @@ def filter_entries(entries, base_directory = nil)
...
     # Returns the list of entries to process
-    def get_entries(dir, subfolder)
+    def get_entries(dir, subfolder, limit)
       base = site.in_source_dir(dir, subfolder)
       return [] unless File.exist?(base)
 
       entries = Dir.chdir(base) { filter_entries(Dir["**/*"], base) }
       entries.delete_if { |e| File.directory?(site.in_source_dir(base, e)) }
+
+      entries_limit = [entries.length, limit].min
```
ashmaroli, line 142: The check for positive `limit` in the next line can be hoisted to enclose this line as well and avoid allocating an array when `limit == 0`.

ashmaroli (changes requested):
Overall, very good first contribution, @emad-elsaid!
However, there are some issues that need to be addressed, notably inconsistencies with making the new parameter optional.

Also, I don't see a value in exposing `:limit` as a *public* getter method especially since it is the same as `site.limit_posts` . Let us have it as a `private` `attr_reader` for now and upgrade in future if requested for.

Additionally, this needs some tests to confirm nothing's broken. You're free to choose between Cucumber based tests and Minitest based tests.

lib/jekyll/reader.rb
```diff
@@ -131,12 +132,15 @@ def filter_entries(entries, base_directory = nil)
...
+      limit.positive? ? entries[-entries_limit, entries_limit] : entries
```
ashmaroli, line 143: I believe this line can be refactored to use `:last(n)`:
```suggestion
      limit.positive? ? entries.last(entries_limit) : entries
```
But, do benchmark the two routes and see if there's a benefit from making the change..

lib/jekyll/reader.rb
emad-elsaid: to avoid this allocation I'll use trinary operator I see it's used here https://github.com/jekyll/jekyll/blob/master/lib/jekyll/site.rb#L480

ashmaroli: Fair enough.

emad-elsaid:
I'm not sure about the tests, while developing this some tests covered the `--limit-posts` and made sure the correct posts are in the result https://github.com/jekyll/jekyll/blob/master/features/site_configuration.feature#L294 

and other tests where using the classes I changed, which complained about the new argument so I added the default value for it.

So the case I'm covering here is reducing the time when there are large number of posts while using `--limit-posts` so I'm not sure how to ensure this with a test so it wouldn't regress.

ashmaroli:
Okay. I've confirmed that the test coverage is already adequate. So, getting a pass status from CI will suffice here.

ashmaroli:
The following line
https://github.com/jekyll/jekyll/blob/3e3d9bce64d576b8d59271548fb4c2e280b93658/lib/jekyll/site.rb#L182

needs to be removed now since the limiting is handled as part of the `site.reader.read` call..

emad-elsaid:
I think if I removed `limit_posts!` this will happen:
- a site with 10 posts and 10 drafts
- limit number of posts to 5
- the reader will limit posts to last 5 and limit drafts to 5
- return 10 posts+drafts instead of total 5
so what current code does is after returning the limited files per directory it limit the sum of them
when I tried to remove it a test failed: 
```
Failure:
TestGeneratedSite#test_: generating limited posts should generate only the specified number of posts.  [/........../jekyll/test/test_generated_site.rb:92]
Minitest::Assertion: Expected: 5
  Actual: 8

914 tests, 2023 assertions, 1 failures, 0 errors, 0 skips
```

one other thing, there is a test in travis that fails but when I run it on my machine it passes, so not sure why this is happening
```
cucumber features/site_configuration.feature:294 # Scenario: Limit the number of posts generated by most recent date
```

ashmaroli:
Drafts and posts are documents of the same collection (`posts`), so they'll end up within `site.posts.docs`.
However, computing the limit and filtering twice does not seem ideal to me.

ashmaroli:
> there is a test in travis that fails but when I run it on my machine it passes

Not sure about this. It passes successfully at my end as well and on AppVeyor CI..
I've restarted the failing jobs on Travis to confirm the behavior.

ashmaroli:
After much thought, I've come to the conclusion that this change is not backwards-compatible.
It is a change in behavior that is markedly different from the long-standing documented feature.

Perhaps in the next major version..

emad-elsaid:
> Drafts and posts are documents of the same collection (`posts`), so they'll end up within `site.posts.docs`.
> However, computing the limit and filtering twice does not seem ideal to me.

Yeah in this case then the current implementation to limit the number of posts in the method that returns posts in a directory needs to be improved, I'll try to find another after we read and merge all directories then limit it and before reading the content of the files themselves.

lib/jekyll/reader.rb
```diff
@@ -131,16 +132,21 @@ def filter_entries(entries, base_directory = nil)
...
     # Returns the list of entries to process
-    def get_entries(dir, subfolder)
+    def get_entries(dir, subfolder, limit = 0)
       base = site.in_source_dir(dir, subfolder)
       return [] unless File.exist?(base)
 
       entries = Dir.chdir(base) { filter_entries(Dir["**/*"], base) }
       entries.delete_if { |e| File.directory?(site.in_source_dir(base, e)) }
+
+      entries_limit = entries.length < limit ? entries.length : limit
```
shushugah, line 142: this can be rewritten has `entries_limit = [entries.length, limit].min`
`

ashmaroli: @shushugah The use of `Array#min` was suggested to be replaced in a previous review to avoid unnecessary allocations.

emad-elsaid:
it's been here and I don't have time to get back to it. so I'll close this.

PR #8483, opened 2020-11-27
Echo PATH to .zshenv for newer macOS's

danielnemenyi:
Apple switched the default shell to Zsh with Catalina, so adding the ruby PATH in .bash_profile is increasingly now the legacy approach, and will not have effect for newer macOS users, preventing them from continuing with the Jekyll installation instructions.

This commit adds instructions to echo the PATH to .zshenv (the Zsh equivalent of .bash_profile) if the user is on Catalina or newer, as well as to check which version of macOS is running for those who are unsure.

## Summary

I've updated the macOS installation documentation to make them work for users of Catalina and newer versions of macOS, for whom the default shell is no longer bash but is now zsh, and so for whom adding the ruby PATH to .bash_profile will not have any effect. 

I've added this change to both the global and local installation instructions.

### DIFF
--- a/docs/_docs/installation/macos.md
+++ b/docs/_docs/installation/macos.md
@@ -30,7 +30,14 @@ brew install ruby
 Add the brew ruby path to your shell configuration:
 
 ```bash
+# MacOS Catalina (10.15) users and newer
+echo 'export PATH="/usr/local/opt/ruby/bin:$PATH"' >> ~/.zshenv
+
+# MacOS Mojave (10.14) users and older
 echo 'export PATH="/usr/local/opt/ruby/bin:$PATH"' >> ~/.bash_profile
+
+# Unsure which version of macOS you are using? Try
+sw_vers -productVersion 
 ```
 
 Relaunch your terminal and check your Ruby setup:
--- a/docs/_docs/installation/macos.md
+++ b/docs/_docs/installation/macos.md
@@ -105,7 +105,14 @@ ruby -v
 Append your path file with the following, replacing the `X.X` with the first two digits of your Ruby version:
 
 ```bash
+# MacOS Catalina (10.15) users and newer
+echo 'export PATH="$HOME/.gem/ruby/X.X.0/bin:$PATH"' >> ~/.zshenv
+
+# MacOS Mojave (10.14) users and older
 echo 'export PATH="$HOME/.gem/ruby/X.X.0/bin:$PATH"' >> ~/.bash_profile
+
+# Unsure which version of macOS you are using? Try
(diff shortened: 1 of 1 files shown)


MichaelCurrin:
Changes look good. Though I think the emphasis is too much on the OS. In the instructions for checking what your OS is, I think it is more useful to check. Especially since when I upgraded to Catalina, it was optional to change to ZSH ie the default shell was Bash until I deliberately changed it. I assume new machines start from ZSH though 

Example content 
---

Check what your default shell is.

```sh
echo $SHELL
```
This was Bash historically, but from macOS Catalina onwards the default is ZSH.

If ZSH is your default, use `.zshenv` below.
If Bash is your default, edit `.bash_profile` instead.

```
echo 'export ... '>> ~/.zshenv
echo 'export ... '>> ~/.bash_profile
```

Also I put the lines side by side instead of separate code blocks.

DirtyF (approved):
Thanks : 🙏

danielnemenyi:
Thanks! This was my very first pull request to a public open source project :))

Must say I do agree with @MichaelCurrin's change though - it does make more sense to base this on $SHELL rather than the macOS version (as I proposed and as merged to master), for the reason he says?

monfresh:
I'm curious as to why `.zshenv` was chosen over `.zshrc`. On a fresh Catalina installation, the file `~/.zshenv` does not exist, but `~/.zshrc` does. I think `~/.zshrc` is the file that is most commonly referenced by other tools. I think it would be confusing for users to end up with 2 different files to manage. Thoughts?

MichaelCurrin:
The existing doc used `~/.bash_profile` - and `~/.zshenv` is the ZSH equivalent in purpose from what I've read.
In the same way `.zshrc` and `.bashrc` are equivalent.

The `.zshenv` file will get created by `>> ~/.zshenv` if it does not exist. So it means a 2nd file but if that is "more correct" maybe that is okay. 

Similarly, a `.bash_profile` files does not come with Debian, Ubuntu or LinuxLite (Ubuntu-based) based on my experience and reading. But a `~/.profile` does exist (and it is _not_ read if `.bash_profile` exists)  is
n terms of existence, on two Debian-like OS's I got a `.profile` and not `.bash_profile`. Based on that, the existence

DirtyF:
Based on https://apple.stackexchange.com/questions/51036/what-is-the-difference-between-bash-profile-and-bashrc I went with `.bash_profile` and .`zshrc` (because they are the defaults on macOS)

https://jekyllrb.com/docs/installation/macos/

Thank you everyone for your input. 🙏

PR #8485, opened 2020-11-29
Reduce allocations from computing item property

ashmaroli:
- This is an :zap: optimization change.

## Summary

The `:to_liquid` method for `Convertible` objects (mainly `Jekyll::Page` and subclasses) is not memoized. Therefore, calling `Jekyll::Page#to_liquid` multiple times results in unnecessary Hash allocations.

This change optimizes iteration through `site.pages` when `item_property` of each page is computed.

## Possible use-cases

Filtering `site.pages` with either of `sort`, `where`, `find` filters multiple times.

PR #8489, opened 2020-11-30
Optimize Page#dir with a private method

ashmaroli:
- This is an :zap: optimization change

## Summary

In the same boat as #8485, because `Jekyll::Page#to_liquid` isn't memoized, `Jekyll::Page#dir` get called for every call to `#to_liquid`.

Unfortunately, `#dir` can't be directly memoized either because `@dir` is already defined in `Jekyll::Page#initialize` and has different semantics in comparison to the result from calling `#dir`. Using a new instance variable name is risky because `Jekyll::Page` is a popular ancestor for Generator based plugins.

Therefore, move the allocating logic into a new private and memoized method.

### DIFF
--- a/lib/jekyll/page.rb
+++ b/lib/jekyll/page.rb
@@ -64,12 +64,7 @@ module Jekyll
     #
     # Returns the String destination directory.
     def dir
-      if url.end_with?("/")
-        url
-      else
-        url_dir = File.dirname(url)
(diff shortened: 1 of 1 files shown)


DirtyF (approved):
LGTM, don't see any build time improvement in the test suite though.

ashmaroli:
@jekyllbot: merge +fix

PR #8497, opened 2020-12-04
Stash attribute hash for Liquid computed for pages

ashmaroli:

## Summary

Since `Jekyll::Page#to_liquid` isn't memoized, the attribute hash for Liquid gets computed on every call.
Therefore compute the attribute hash on first call and stash it for consequent reuse.


DirtyF (approved):
💾

ashmaroli:
@jekyllbot: merge +fix

PR #8503, opened 2020-12-06
Generate pages for every tag and category in site

ashmaroli:

## Summary

Proof-of-Concept for *built-in* generation of taxonomy pages.

Disabled by default, the pages are generated only when the user explicitly configures to do so:
```yaml
# _config.yml

taxonomy_pages:
  tags: enabled
  categories: enabled
```
- Each generated page is hard-coded to be **an empty HTML source file**, with access to the `site` instance and the list of (posts) associated with given tag or category.
- Tag pages are of type `tag` and category pages of type `category`, which can be used to define default front matter via the config file key `defaults`, similar to pages, posts and other documents.
- User / theme will have to provide the necessary layout to render individual pages.
- Index pages for all tags and all categories in a site are not generated and will have to be provided by the user if desired.

### DIFF
--- a/lib/jekyll.rb
+++ b/lib/jekyll.rb
@@ -191,5 +191,6 @@ require_all "jekyll/converters/markdown"
 require_all "jekyll/drops"
 require_all "jekyll/generators"
 require_all "jekyll/tags"
+require_all "jekyll/taxonomy"
 
 require "jekyll-sass-converter"
--- a/lib/jekyll/generators/taxonomy_manager.rb
+++ b/lib/jekyll/generators/taxonomy_manager.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+module Jekyll
+  class TaxonomyManager < Generator
+    safe :true
+
+    def initialize(config)
+      @config = config["taxonomy_pages"]
+    end
+
+    def generate(site)
+      if @config["tags"] == "enabled"
+        site.tags.each do |tag, docs|
+          site.pages << Taxonomy::Page.new(site, "tag", tag, docs)
+        end
+      end
+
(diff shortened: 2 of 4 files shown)


ashmaroli:
@DirtyF What do you think..? Good enough to proceed with test coverage and documentation..?

DirtyF:
This looks interesting.
Could we come up with a small demo site that demonstrates the feature? 🙏 
It would make it way easier to see how this works (and it would document the feature at the same time).

I am particularly interested to see how much control do we have over the generated pages, I want to generate tags pages for posts for instance and categories pages for a collection only. Is that possible?

ashmaroli:
Thanks for the feedback @DirtyF. Regarding your queries:

> I want to generate tags pages for posts for instance and categories pages for a collection only. Is that possible?

Out-of-the-box, Jekyll only handles tags and categories for posts.
For extending support to other collections, one will have to use the `jekyll-tagories` plugin at least v5.0 (to maintain existing behavior but allow users to opt-in to extended functionality).

That said, I have not yet added the fine-grain control to support the given use-case.
Good idea!

> Could we come up with a small demo site that demonstrates the feature?

We could. Especially with GitHub Actions.
But for now, I'll demo this feature using our docs site at `jekyllrb.com`.
Thanks for the idea.

DirtyF:
OK so this is only intended for posts by default, to generate a page like https://jekyllrb.com/news/releases/ for instance?

ashmaroli:
Yes. That's right.
But instead of having a page just for the `release` category, there would be a similar page for **all categories** easily.
(*It is currently an all-or-none feature.*)

ashmaroli:
Dropping this since `jekyll-archives` already provides this functionality..

PR #8519, opened 2020-12-22
Add new tag relative_link that produces links relative to the current…

white-gecko:

So far I have not added the tests and updated the documentation as it first needs to be decided, if such a feature request might be accepted.

## Summary

This pull request introduces a new tag called `relative_link` that creates a link to a post or page relative to the current page. (It also works for the pretty permalink setting.

## Context

As became apparent in the discussion of https://github.com/jekyll/jekyll/issues/6360 and https://github.com/jekyll/jekyll/pull/6362 there is a high need for such a feature. It is necessary:
- for serving a static site from the filesystem (not in the focus of jekyll but if you never start, it will never get into the focus)
- if the page might later be moved to a different base path (https://github.com/jekyll/jekyll/pull/6362#issuecomment-379078036)
- if the target server environment is not clear at the time building the page e.g.:
  - when serving the page from IPFS the base path can not be determined beforhand (https://github.com/jekyll/jekyll/pull/6362#issuecomment-386450745)
  - when a page is served from the same directory to different urls e.g. when a page is served both to github pages but also under a personal domain.

Also the issue raised with #6362 (https://github.com/jekyll/jekyll/pull/6362#issuecomment-358023115) that it does not work for pretty permalink setting is solved in this pull request.

By using a tag, just as the `link`, which is called `relative_link` there should also be no problem with name confusion.

### DIFF
--- a/lib/jekyll/tags/relative_link.rb
+++ b/lib/jekyll/tags/relative_link.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+module Jekyll
+  module Tags
+    class RelativeLink < Liquid::Tag
+      include Jekyll::Filters::URLFilters
+
+      class << self
+        def tag_name
+          "relative_link"
+        end
+      end
+
+      def initialize(tag_name, relative_path, tokens)
+        super
+
+        @relative_path = relative_path.strip
+      end
+
+      def relativize_url(item)
+        page_dir = Pathname(@context.registers[:page]["dir"])
+        site_relative_url = relative_url(item)
+        return Pathname(site_relative_url).relative_path_from(page_dir).to_s
+      end
+
+      def render(context)
+        @context = context
+        site = context.registers[:site]
+        relative_path = Liquid::Template.parse(@relative_path).render(context)
+        relative_path_with_leading_slash = PathManager.join("", relative_path)
+
+        site.each_site_file do |item|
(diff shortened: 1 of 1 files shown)


ashmaroli:
@white-gecko Hello. I'm willing to *consider / look into* this proposal if you will provide *complete test coverage* for the tag.
Basically, the tests need to cover at minimum the following scenarios:
- link to an external stylesheet **from** a nested page (e.g, `permalink: /about/company/`) and nested post (e.g. one that uses categories in the URL: `/movies/mcu/2019/11/08/avengers/`).

white-gecko:
Sounds fair. I don't have the time right now to do that. If anyone else wants to contribute to the pull-request, you are welcome!

fauno:
Hi, if it's any help, we've developed a plugin that does this: https://rubygems.org/gems/jekyll-relative-urls

white-gecko:
@fauno thank you for the hint. Your approach seems to be different to this one. You might want to propose a competing pull request and we can see which one fits better into the code base.

white-gecko:
It is not yet test coverage but an example showing that the scenarios mentioned by @ashmaroli are covered: https://github.com/white-gecko/jekyll-example-use-relative_link

If anybody wants to make proper jekyll tests from it, you are welcome

PR #8521, opened 2020-12-23
Read-Only JSON API Tutorial

izdwuut:
This is a 🔦 documentation change. 

## Summary

I've added a tutorial on building Read-Only JSON APIs. It mostly covers creating a custom plugin with generators, converters and hooks. This is a detailed step-by-step guide. 

I think it's great that you allow such contributions. I was about to put it behind a paywall on Medium, but if you add the tutorial to the docs, it could reach far more people.

[Preview](https://deploy-preview-8521--jekyllrb.netlify.app/tutorials/read-only-json-api/) 👀

PR #8525, opened 2020-12-27
Add a disclaimer to tutorials involving Ruby code

ashmaroli:
## Summary

Tutorials utilizing just Liquid logic is considered safe since Liquid is inherently non-evaling. However, tutorials with Ruby code to be used as plugins cannot be deemed safe without proper review and test-coverage.
Therefore, place a disclaimer to indirectly state that the user discretion is recommended.

The disclaimer is not rendered by default. Tutorials with Ruby code need to explicitly enable it via the front matter:
```
plugin_disclaimer: true
```

### DIFF
--- a/docs/_layouts/tutorials.html
+++ b/docs/_layouts/tutorials.html
@@ -27,6 +27,13 @@ layout: default
           {% endif %}
         </header>
         {{- content -}}
+        {% if page.plugin_disclaimer %}
+          <div class="disclaimer-ribbon">
+            Disclaimer: The Jekyll Core Team does not endorse the Ruby code in this tutorial and is
+            not liable to resolve any bugs therein or issues arising otherwise from the use of the
+            provided example(s) in production builds.
+          </div>
+        {% endif %}
         {%- include section_nav_tutorials.html -%}
       </article>
     </div>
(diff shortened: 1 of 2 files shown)


sa-orlov (approved):

sa-orlov (approved):

chrisfinazzo:
> ## Summary
> Tutorials utilizing just Liquid logic is considered safe since Liquid is inherently non-evaling. However, tutorials with Ruby code to be used as plugins cannot be deemed safe without proper review and test-coverage.
> Therefore, place a disclaimer to indirectly state that the user discretion is recommended.
> 
> The disclaimer is not rendered by default. Tutorials with Ruby code need to explicitly enable it via the front matter:
> 
> ```
> plugin_disclaimer: true
> ```

While testing this in the context of #8521, I noticed that it wasn't using the CSS I expected. Seems like this should have the regular "note" class applied...

Screen Shot 2020-12-29 at 11 05 37 AM

ashmaroli:
> the regular "note" class applied..

@chrisfinazzo I chose to go with the bronze+chocolate-brown palette so that it is subtle and doesn't pop right out at the visitor (like the sky-blue or red note boxes on other pages) especially since the info within is not directly related to **the topic of the page**.
Whatsay? Change it anyways?

DirtyF (approved):

ashmaroli:
@jekyllbot: merge +doc

PR #8533, opened 2021-01-05
Remove GitHub Pages content which is in GitHub docs

BenWhetton:

## Summary

Much of the contents of the Jekyll GitHub Pages guide is duplicated on the GitHub docs page. I imagine they have been considerably updated since the Jekyll GitHub Pages guide was last modified. Given that the GitHub docs seem to be pretty high quality to me, I think it's probably best to link to them rather than duplicating their contents.

I'm not entirely sure how much of the Jekyll GitHub Pages guide needs to be kept:
- The "github-pages gem" section seems to be duplicated [in the GitHub docs](https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages/creating-a-github-pages-site-with-jekyll) (see steps 8-9) so I've removed it.
- I can't find the information from the "Project Page URL Structure" section in the GitHub docs (I am not familiar enough with this to judge whether it is still accurate/useful so I haven't touched it).
- I believe the "Deploying Jekyll to GitHub Pages" section is also pretty much duplicated [here](https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages/about-github-pages#types-of-github-pages-sites) so we might want to remove that too.

In addition, all the existing links to the GitHub docs were broken so I've fixed them.

The blog post linked in the second paragraph is fairly outdated now. It recommends using custom css and html files rather than a jekyll theme. [The suggested css file has some formatting issues as well](https://github.com/jmcglone/jmcglone.github.io/issues/20). I have removed the link to it since I believe the GitHub docs are now a better reference which is more likely to be maintained.

It's possible that I have missed something about the purpose of this page and you have good reasons for keeping the duplicated content. If this is the case, I'll happily update this PR with just the blog post removal and link fixes.

Note: the GitHub guides are not perfect, there are a few missing steps when describing how to use bundler, for example. The [Jekyll bundler guide](https://jekyllrb.com/tutorials/using-jekyll-with-bundler/#initialize-bundler) is more complete. I intend to submit a PR to fix this and suggest that they might prefer to link to the Jekyll guide.

## Context

Not related to an open issue. I'm happy to make one if that would be useful.

### DIFF
--- a/docs/_docs/github-pages.md
+++ b/docs/_docs/github-pages.md
@@ -15,40 +15,20 @@ simply because Jekyll treats files without front matter as static assets.
 So if you only need to push generated HTML, you're good to go without any
 further setup.
 
-Never built a website with GitHub Pages before? [See this marvelous guide by
-Jonathan McGlone](http://jmcglone.com/guides/github-pages/) to get you up and
-running. This guide will teach you what you need to know about Git, GitHub, and
-Jekyll to create your very own website on GitHub Pages.
+The [GitHub Pages Documentation](https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages)
+is comprehensive and includes a [a guide to setting up a GitHub Pages site using
+Jekyll](https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages/setting-up-a-github-pages-site-with-jekyll).
+We recommend following this guide.
 
-##  The github-pages gem
-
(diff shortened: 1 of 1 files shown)


DirtyF (approved):
Agreed, I don't see for us the need to maintain documentation for GitHub Pages

DirtyF:
@jekyll: merge +docs

PR #8524, opened 2020-12-25
Add webrick as a dependency

argilo:
This is a 🐛 bug fix.

## Summary

As of Ruby 3.0, WEBrick is no longer bundled with Ruby, so it needs to be listed as a dependency of Jekyll.

## Context

Fixes #8523

PR #8536, opened 2021-01-06
Add webrick as a dependency for tests

pvalena:
This is a 🐛 bug fix.

## Summary

Add webrick into test dependencies, because it's used in tests, and it's not shipped with Ruby 3.0.


ashmaroli:
Related: #8524

ashmaroli:
Superseded by #8524

PR #8537, opened 2021-01-07
Update Actions Page in the Docs

izdwuut:

## Summary

Since GitHub [changed](https://github.blog/changelog/2020-10-01-the-default-branch-for-newly-created-repositories-is-now-main/) the default branch from `master` to `main`, I think it would be nice to reflect the changes in the docs. I believe this applies to more pages as well, but I only changed it in [Actions](https://jekyllrb.com/docs/continuous-integration/github-actions/). 

Alternatively we can include a note that the default branch is now `main` and leave the docs as-is. I think it would be easier.

## Context

Doesn't apply.

### DIFF
--- a/docs/_docs/continuous-integration/github-actions.md
+++ b/docs/_docs/continuous-integration/github-actions.md
@@ -32,8 +32,8 @@ The first and foremost requirement is a Jekyll project hosted at GitHub. Choose 
 project or follow the [Quickstart]({{ '/docs' | relative_url }}) and push the repository to GitHub
 if it is not hosted there already.
 
-We're only going to cover builds from the `master` branch in this page. Therefore, ensure that you
-are working on the `master` branch. If necessary, you may create it based on your default branch.
+We're only going to cover builds from the `main` branch in this page. Therefore, ensure that you
+are working on the `main` branch. If necessary, you may create it based on your default branch.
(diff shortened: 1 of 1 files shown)


ashmaroli:
I'm really not in favor of this change..

DirtyF:
Hey @ashmaroli could you explain why?

DirtyF (changes requested):
A notice about the `main` branch is enough, it's a [pretty recent change](https://github.blog/changelog/2020-10-01-the-default-branch-for-newly-created-repositories-is-now-main/), and only new repositories have that nomenclature.

ashmaroli:
I'm not in favor of the change because the documentation as it stands does not lose relevance or meaning irrespective of what GitHub deems to be the default branch of a repository. Nowhere in the page do we *enforce the need to run the action on the default branch*. It could be any branch. `master` is just an example name.

izdwuut:
@ashmaroli A reasonable explanation. I have to admit, I didn't put too much thought to this PR. It was a simple action-reaction, like: "the new default branch is `main`, right? So it should be changed".

Now, I don't think that it's really worth it. More confusion is not needed. 

I think we can mutually agree on closing the PR.

ashmaroli:
Thank you @izdwuut, :smiley: 
I'll let you close this yourselves.

izdwuut:
Yay!

PR #8538, opened 2021-01-07
style: enable new cops

DirtyF:
This is a 👨🏻‍🎤  style enhancement.

## Summary

- Bump to [Rubocop 1.8.0](https://github.com/rubocop-hq/rubocop/releases/tag/v1.8.0) 🆙 
- Enable new cops 👮🏻 
- Add extensions for minitest, rake and rspec 💅🏻 
- Lint Robucop config file ✅ 
- Auto-correct ✨

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,7 +2,10 @@
 inherit_from: .rubocop_todo.yml
 
 require:
+  - rubocop-minitest
   - rubocop-performance
+  - rubocop-rake
+  - rubocop-rspec
   - ./rubocop/jekyll
 
 Jekyll/NoPutsAllowed:
@@ -21,6 +24,7 @@ AllCops:
     - script/**/*
     - vendor/**/*
     - tmp/**/*
+  NewCops: enable
 
 Layout/BeginEndAlignment:
   Enabled: true
@@ -30,14 +34,14 @@ Layout/EmptyLinesAroundAttributeAccessor:
   Enabled: true
 Layout/EndAlignment:
   Severity: error
-Layout/HashAlignment:
-  EnforcedHashRocketStyle: table
-Layout/IndentationWidth:
-  Severity: error
 Layout/FirstArrayElementIndentation:
   EnforcedStyle: consistent
 Layout/FirstHashElementIndentation:
   EnforcedStyle: consistent
+Layout/HashAlignment:
+  EnforcedHashRocketStyle: table
+Layout/IndentationWidth:
+  Severity: error
 Layout/LineLength:
   Exclude:
     - !ruby/regexp /features\/.*.rb/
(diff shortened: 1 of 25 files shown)


test/test_filters.rb
```diff
@@ -940,12 +940,12 @@ def to_liquid
 
         results = @filter.where(hash, "featured", "true")
         assert_equal 2, results.length
-        assert_equal 9.2, results[0]["rating"]
-        assert_equal 4.7, results[1]["rating"]
+        assert_in_delta(9.2, results[0]["rating"])
```
ashmaroli, line 943: Can we revert these changes?
I feel that `assert_equal` is more intuitive than `assert_in_delta` ...

.rubocop.yml
```diff
@@ -21,6 +24,7 @@ AllCops:
     - script/**/*
     - vendor/**/*
     - tmp/**/*
+  NewCops: enable
```
ashmaroli, line 27: I'm not so sure we should enable new cops without reviewing their effects first..
CI could fail with an offense unrelated to a commit or PR.

.rubocop.yml
```diff
@@ -127,7 +142,6 @@ Lint/UselessTimes:
 Lint/Void:
   Exclude:
     - lib/jekyll/site.rb
-
```
ashmaroli, line 130: Can we have these separators between Cop departments back? Or will they get removed automatically with autocorrect again?

ashmaroli (commented):
Just some minor thoughts.. :+1:  otherwise..

.rubocop.yml
DirtyF: YAMLint 🤷🏻

.rubocop.yml
DirtyF: It was kind of the point fix style issues as early as possible, a PR should embrace our styling guide. No strong opinion on this. It was just easier than to add all Minitest default cops in the Rubocop File, I was hoping it would lead to less maintenance work.

ashmaroli: > a PR should embrace our styling guide.

I agree. But I'm talking about offenses unrelated to changes in the PR..

Say a contributor has RuboCop 1.8 installed locally and has prepared their pull request after ensuring `script/fmt` reports no offense. However, by the time they *actually opens the pull request* on GitHub, RuboCop has released a v1.9.0 with many new cops that get enabled automatically and report offenses on code not related to the pull request but fails the CI build nevertheless.

test/test_filters.rb
DirtyF: > Use assert_in_delta if comparing floats. Assertion passes if the expected value is within the delta of actual value.

This is intended to compare floats,  it's what we're doing here, something you have to get used to maybe, but once you know it, it makes the test as intuitive as relying on a generic equal.

DirtyF:
@jekyllbot: merge +dev

PR #8569, opened 2021-02-02
Highlight variables

### LINKED ISSUE #8290: feat: Pass language variable to {% highlight %}
## Summary
Highlighting code in template where a file is linked in the markdown front matter can currently happen only if the language of the code is known when writing the template. I propose that variables should be passed to the `{% highlight %}` tag, like it is for the conditional tags, though it may be more clear if it was in a format that resembles the `{% include %}` tag.

## Motivation
I am building a [portfolio template](https://github.com/urishx/Jekyll-portfolio) based on Minima, where the projects in a portfolio are stored in markdown files, gathered as a collection. I want to pass a code url (I'm using the [jekyll-remote-include](https://github.com/netrics/jekyll-remote-include) plugin for that), and the language to highilght it in, as some are C++, some are JS, and some are other languages still. I am including the links in the file's front matter, and would like to pass the language as well, in the following syntax:
```yaml
array:
  - code: https://raw.githubusercontent.com/username/repository/branch/cppflie.ino
    lang: cpp
  - code: https://raw.githubusercontent.com/username/repository/branch/pythonfile.py
    lang: python
```
And in my template I want to write:
```liquid
{% for i in array %}
    {% capture codeurl %}
        {% remote_include {{ i.code }} %}
    {% endcapture %}
    {% highlight i.lang %}
        {{ codeurl | strip }}
    {% endhighlight %}
{% endfor %}
```

The current implementaion requires the template to define in advance a language, and I believe it would be beneficial to pass the language using a liquid variable for creating templates which can be agnostic to the language of the file type to be highlighted. 

I believe it is in line with the use of conditionals, and may allow easier building of templates which include several code excerpts, in different languages, for example in a tutorial about API usage in different languages. \
I also think that it may be more clear to the user if the syntax can be similar to the `{% include %}` tag, but the regex expression the paramaters are tested against in the `initialize` function does not include curly braces, and I'm not sure if adding them may break something down the line or not.

*Not sure, but I'll try. New to Ruby, please excuse :)*

As far as I can tell, the `{% highlight %}` tag is defined in the file [highlight.rb](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/tags/highlight.rb) and its documentation is in [tags.md](https://github.com/jekyll/jekyll/blob/master/docs/_docs/liquid/tags.md). I believe that this feature will only affect these two files.

In the file [highlight.rb](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/tags/highlight.rb), the parameters passed are tested for validity in the `initialize` function, downcased, then saved to the global variable `@lang`. It is then used in the `render_rouge` function. \
I propose to test for the exsitance of a fitting lexer in context, in the `render` function:
```ruby
begin
    ::Rouge::Lexer.find((context[@lang]).downcase) != nil
    @lang = (context[@lang]).downcase
rescue
    @lang = @lang.downcase
end
```
In the `initalized`, it is then only needed to change `@lang = Regexp.last_match(1).downcase` to `@lang = Regexp.last_match(1)`. \
In order for that to work, the import of Rouge should be moved to the top (from the current location in `render_rouge`) and be global.

In the file [tags.md](https://github.com/jekyll/jekyll/blob/master/docs/_docs/liquid/tags.md) the following can then be added for documentation (based on [include.md](https://raw.githubusercontent.com/jekyll/jekyll/master/docs/_docs/includes.md)):
>### Using variables names for the language
>
>The name of the language you for the code to be highlighted can be specified as a variable instead of specifying the language directly in the template. For example, suppose you defined a variable in your page's front matter like this:
>
>```yaml
>---
>title: My page
>my_code: footer_company_a.html
>my_lang: liquid
>---
>```
>
>You could then reference that variable in your >highlight:
>
>{% raw %}
>```liquid
>{% if page.my_variable %}
>  {% capture my_code %}
>    {% include {{ page.code }} %}
>  {% endcapture %}
>  {% highlight page.my_lang %}
>    {{ my_code | strip }}
>  {% endhighlight %}
>{% endif %}
>```
>{% endraw %}
>
>In this example, the capture will store the include file `_includes/footer_company_a.html`, then the highlight will would match the display to match the syntax of `liquid`.

## Drawbacks

1. It's a corner case (though in a tag included in Jekyll core).
1. Rouge should be imported either as global or both in `render` and `render_rouge`.
1. In the method I suggested above each time highlight is called the lexer is searched at least twice.
1. The syntax would be better if it was similar to include.

## Unresolved Questions

I don't know what I don't know...

UriShX:

## Summary

This feature adds the ability to pass in variables to Jekyll's `{% highlight %}` tag, instead of hard-coded parameters, while maintaining compatibility with the existing usage of the tag.

Suggested usage of this feature could be:
- Using the tag in a layout, allowing different languages and line numbering options for each to be specified in the front matter of each rendered file. For example, see [my portfolio layout](https://github.com/UriShX/Jekyll-portfolio/blob/master/_layouts/portfolio.html#L130) and a [markdown file rendered by that layout](https://raw.githubusercontent.com/UriShX/Jekyll-portfolio/master/_portfolio/raspi-vid-on-btn-press.md).
- Highlighting a different language then the main language of the code, for eg. - a HTML page embedded in a C program for an embedded IoT device, such as [this Arduino file](https://raw.githubusercontent.com/UriShX/portfolio/master/Roboclaw_control_over_ESP32_with_AP_for_control/roboclaw_esp32_w_AP_and_config.ino).

I've added tests to `test/test_tags.rb`, and updated the documentation to include passing variables. I've also added a new Cucumber, `features/highlight_w_vars.feature`.

## Context
Resolves #8290

### DIFF
--- a/docs/_docs/liquid/tags.md
+++ b/docs/_docs/liquid/tags.md
@@ -2,6 +2,7 @@
 title: Tags Filters
 permalink: "/docs/liquid/tags/"
 ---
+
 All of the standard Liquid
 [tags](https://shopify.github.io/liquid/tags/control-flow/) are supported.
 Jekyll has a few built in tags to help you build your site. You can also create
@@ -27,6 +28,7 @@ and 100% compatible with stylesheets for Pygments.
 To render a code block with syntax highlighting, surround your code as follows:
 
 {% raw %}
+
 ```liquid
 {% highlight ruby %}
 def foo
@@ -34,6 +36,7 @@ def foo
 end
 {% endhighlight %}
 ```
+
 {% endraw %}
 
 The argument to the `highlight` tag (`ruby` in the example above) is the
@@ -57,6 +60,7 @@ numbers. For instance, the following code block would include line numbers next
 to each line:
 
 {% raw %}
+
 ```liquid
 {% highlight ruby linenos %}
 def foo
@@ -64,7 +68,52 @@ def foo
 end
 {% endhighlight %}
 ```
+
 {% endraw %}
+
+### Variables for language and line numbers
+
+{: .note}
+To pass a language variable to the `highlight` tag, the variable **must** be named `lang`, as in {% raw %}{{ page.lang }}{% endraw %}. The `linenos` variable can be called by any other name.
+
+Both language and the line numbering parameter `linenos` can be passed to the `highlight` tag as variables, instead of specific paramters. \
+For example, suppose you defined a variable in your page's front matter like this:
+
+```yaml
+---
+title: My page
+lang: ruby
+lines: true
+---
+
+```
+
+You could then reference those variables in your link:
+
+{% raw %}
+
+```liquid
+{% highlight {{ page.lang }} {{ page.lines }} %}
+def foo
+  puts 'foo'
+end
+{% endhighlight %}
+```
+
+{% endraw %}
+
+In this example, the `highlight` tag would render the code in the same manner as in the Line Numbers example, above. However, a different page may turn off the line numbering, eg.:
+
+```yaml
+---
+title: My page
+lang: ruby
+lines: false
+---
+
+```
+
+Would render (using the same layout) the same highlighted code block without line numbering.
 
 ### Stylesheets for syntax highlighting
 
@@ -93,28 +142,32 @@ To link to a post, a page, collection item, or file, the `link` tag will generat
 You must include the file's original extension when using the `link` tag. Here are some examples:
 
 {% raw %}
+
 ```liquid
 {% link _collection/name-of-document.md %}
 {% link _posts/2016-07-26-name-of-post.md %}
 {% link news/index.html %}
 {% link /assets/files/doc.pdf %}
 ```
+
 {% endraw %}
 
 You can also use the `link` tag to create a link in Markdown as follows:
 
 {% raw %}
+
 ```liquid
 [Link to a document]({% link _collection/name-of-document.md %})
 [Link to a post]({% link _posts/2016-07-26-name-of-post.md %})
 [Link to a page]({% link news/index.html %})
 [Link to a file]({% link /assets/files/doc.pdf %})
 ```
+
 {% endraw %}
 
 The path to the post, page, or collection is defined as the path relative to the root directory (where your config file is) to the file, not the path from your existing page to the other page.
 
-For example, suppose you're creating a link in `page_a.md` (stored in `pages/folder1/folder2`) to `page_b.md` (stored in  `pages/folder1`). Your path in the link would not be `../page_b.html`. Instead, it would be `/pages/folder1/page_b.md`.
+For example, suppose you're creating a link in `page_a.md` (stored in `pages/folder1/folder2`) to `page_b.md` (stored in `pages/folder1`). Your path in the link would not be `../page_b.html`. Instead, it would be `/pages/folder1/page_b.md`.
 
 If you're unsure of the path, add {% raw %}`{{ page.path }}`{% endraw %} to the page and it will display the path.
 
@@ -129,14 +182,17 @@ The name of the file you want to link can be specified as a variable instead of 
 title: My page
 my_variable: footer_company_a.html
 ---
+
 ```
 
 You could then reference that variable in your link:
 
 {% raw %}
+
 ```liquid
 {% link {{ page.my_variable }} %}
 ```
+
 {% endraw %}
 
 In this example, the `link` tag would render a link to the file `footer_company_a.html`.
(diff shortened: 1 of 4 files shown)


docs/_docs/liquid/tags.md
```diff
@@ -57,15 +60,61 @@ numbers. For instance, the following code block would include line numbers next
...
 ```
+
+{% endraw %}
+
+### Variables for language and line numbers
+
+{: .note}
+To pass a language variable to the `highlight` tag, the variable **must** be named `lang`, as in {% raw %}{{ page.lang }}{% endraw %}. The `linenos` variable can be called by any other name.
+
+Both language and the line numbering parameter `linenos` can be passed to the `highlight` tag as variables, instead of specific paramters. \
```
MichaelCurrin, line 79: Add line break instead of \. And fix typo

```suggestion
Both the language and the line numbering parameter (`linenos`) can be passed to the `highlight` tag as variables, instead of the standard approach of fixed parameters.

```

docs/_docs/liquid/tags.md
```diff
@@ -57,15 +60,61 @@ numbers. For instance, the following code block would include line numbers next
...
+{% highlight {{ page.lang }} {{ page.lines }} %}
+def foo
+  puts 'foo'
+end
+{% endhighlight %}
+```
+
 {% endraw %}
 
+In this example, the `highlight` tag would render the code in the same manner as in the Line Numbers example, above. However, a different page may turn off the line numbering, eg.:
```
MichaelCurrin, line 105:
```suggestion
In this example, the `highlight` tag would render the code in the same manner as in the Line Numbers example, above. However, a different page may turn off the line numbering

e.g.
```

docs/_docs/liquid/tags.md
```diff
@@ -57,15 +60,61 @@ numbers. For instance, the following code block would include line numbers next
...
 end
 {% endhighlight %}
 ```
+
+{% endraw %}
+
+### Variables for language and line numbers
+
+{: .note}
+To pass a language variable to the `highlight` tag, the variable **must** be named `lang`, as in {% raw %}{{ page.lang }}{% endraw %}. The `linenos` variable can be called by any other name.
```
MichaelCurrin, line 77: I don't think this statement is true about _must_ use "lang".

Variable names don't matter. the value matters


Have you tried using another name and see what happens?

```
---
language: ruby
---
{% highlight {{ page.language }} %}
...
```

docs/_docs/liquid/tags.md
```diff
@@ -146,25 +202,31 @@ In this example, the `link` tag would render a link to the file `footer_company_
 If you want to include a link to a post on your site, the `post_url` tag will generate the correct permalink URL for the post you specify.
 
 {% raw %}
+
```
MichaelCurrin, line 205: Unnecessary style changes to add lines.

features/highlight_w_vars.feature
```diff
@@ -0,0 +1,37 @@
...
+      {% assign lines = 'linenos' %}
+      {% highlight {{ page.lang }} {{ lines }} %}
+      RewriteEngine On
+      RewriteCond %{REQUEST_FILENAME} !-f
+      RewriteCond %{REQUEST_FILENAME} !-d
+      RewriteRule ^(.*)$ index.php [QSA,L]
+      {% endhighlight %}
+
+      """
+    And I have a "_config.yml" file with content:
```
MichaelCurrin, line 29:
```suggestion
      And I have a "_config.yml" file with content:
```
Check your indentation

MichaelCurrin (changes requested):
Doc changes

docs/_docs/liquid/tags.md
UriShX: Thanks for your comments.

In order to differentiate between the `lang` and `linenos` variables I used a regexp that catches only variable names that end in 'lang'.

Testing the Regexp [here](https://rubular.com/r/fCGFAF7dCFA90E) with `{{ page.language }}` catches the variable in the parameters (linenos) var. But I think it would be useful to catch all similar variables (lang / language / lang_a / lang1 etc.) so that several top-level language variables could be specified for the page, like so:
```text
---
lang1: cpp
lang2: html
---
{% highlight {{ page.lang1 }} %}
/* code */
{% endhighlight %}
...
{% highlight {{ page.lang2 }} %}
<!-- markup -->
{% endhighlight %}
```
I tried to accommodate that by expanding the regexp a bit, and would appreciate testing. [Here](https://rubular.com/r/riHdOEUdsaaFAT) is the revised regexp.

MichaelCurrin: Oh okay that makes sensitive but sounds like an anti-pattern in programming. I didn't know the tag could even know the name of the field passed to it, I thought it would use the evaluated value only.

I would have expected something like you pass one parameter and that is only lang, or pass two and that is lang and then linenos, exactly in that order only (no  regex needed then as you have positional parameters).

If there was a concern about the need to use no land but still set linenos, I would have expected  calling as `{% highlight nil true %}`. Or using named parameters - if tags support this like filters do? `{% highlight linenos=true %}` and  `{% highlight {{ page.lang }} linenos=true %}`

PR #8574, opened 2021-02-05
Added hash_add_item and hash_remove_item filters, plus docs and tests

JonTheNiceGuy:

## Summary
This adds filters to add and remove items from a hash (similar to the push and pop filters for arrays).

## Context
I've tested this in my own context, however, I'm struggling to get the `cibuild` to run. I've put a [comment](https://talk.jekyllrb.com/t/problems-running-script-cibuild/5608) about this on [Jekyll Talk](https://talk.jekyllrb.com), and am happy to have this considered to be "WIP" until that's resolved, if required. In the interim, I've got a local plugin until this is adopted.

### DIFF
--- a/docs/_data/jekyll_filters.yml
+++ b/docs/_data/jekyll_filters.yml
@@ -340,6 +340,18 @@
 
 #
 
+- name: Hash Filters
+  description: >-
+    hash_add_item and hash_remove_item add and remove items from a Hash.
+    These are <strong>NON-DESTRUCTIVE</strong>, i.e. they do not mutate the Hash.
+  examples:
+    - input: '{{ data.store | hash_add_item: "key", "value" }}'
+      output: '{"existing" => "data", "key" => "value"}'
+    - input: '{{ data.store | hash_remove_item: "existing" }}'
+      output: '{}'
+
+#
+
 - name: Inspect
   description: Convert an object into its String representation for debugging.
   examples:
--- a/lib/jekyll/filters.rb
+++ b/lib/jekyll/filters.rb
@@ -355,6 +355,20 @@ module Jekyll
       new_ary
     end
 
+    def hash_add_item(hash, key, value)
+      return hash unless hash.is_a?(::Hash)
+
+      hash[key] = value
+      hash
+    end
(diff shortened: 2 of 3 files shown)


JonTheNiceGuy:
Looks like my worry about `cibuild` not running locally were unfounded!

ashmaroli:
Hello @JonTheNiceGuy,
Thank you for submitting a complete pull request. :)

I like the idea. However, if you look at the existing filters, they generally do not *mutate* the given object.
So in that sense, this will be a pair of anomaly.

JonTheNiceGuy:
Apologies! Should I use this form:

```
new_hash = hash.dup
new_hash.do.stuff
new_hash
```

? If so, I'll fix this later.

ashmaroli:
> Should I use this form..

Yes, please.
The downside unfortunately, is increased memory allocation.

Additionally, you may also consider the following (optional):
- how about naming the filters as `inject` and `eject`.
- should the filter continue execution if provided `key` is `nil` or `false` or an empty string?
- you could use `Object#tap` to avoid assigning a variable separately?
    ```ruby
    def filter
      [...] # guard clauses
      input.dup.tap { |hash| do_something(hash) }
    end
    ```

JonTheNiceGuy:
> how about naming the filters as inject and eject.

I've added two new `hash_inject_item` and `hash_eject_item` methods which are mutating, and updated `hash_add_item` and `hash_remove_item` to be non-mutating. Docs and tests are added.

> should the filter continue execution if provided key is nil or false or an empty string?

Honestly, I don't know how to abort execution 😞  I'm really rather new to Ruby and Jekyll-dev, so I've made my PR small, and if you think there's some merit in improving my PR later, then it's something that could be incorporated.

> you could use Object#tap to avoid assigning a variable separately?

I have no idea what this is 😄 ! As mentioned before, I'm very new to Ruby!

One other thing I was thinking - is it worth having a mutating version of the array functions, again perhaps naming them `inject` and `eject` too?

JonTheNiceGuy:
Anything needed to progress this?

PR #8578, opened 2021-02-08
Action configuration to use "main" as default repo

jondarrer:

## Summary

**Updates the action configuration to use "main" as default repository name**

The default GitHub branch when creating a repository is now "main" and no longer "master", but I have included a comment for those with older repositories.

https://github.com/github/renaming
https://www.zdnet.com/article/github-to-replace-master-with-main-starting-next-month/


MichaelCurrin:
I like the idea of using `main` to match new GH default.

But I would not add the part about the date. Yes GH used master before that date. But if you run git init locally you still get master

So the comment can more generic and maybe even before the config and not inline. Saying


    Under branches, section below, make sure to reference your default branch. Typically this is "main" or "master".
    ```yaml 
    ...
       branches:
          - main
    ```

MichaelCurrin (changes requested):
Leave as just main and move comment above. Remove date reference.

jondarrer:
> Leave as just main and move comment above. Remove date reference.

@MichaelCurrin Thanks for carefully reviewing the change.

I originally inlined the comment to guard against blinkered copy and paste, but you're right that it doesn't live there.

I'll also remove the date, which you point out is misleading as that is the date GitHub made their change, and git clients may still use master as the default branch.

MichaelCurrin:
Great, thanks.

For interest, I recently discovered how to change git's default branch. According to git docs, the default is still master.

```sh
$ git config --global init.defaultBranch main
```

```sh
$ git init my-app
$ cd my-app
$ git branch
main
```

https://michaelcurrin.github.io/dev-cheatsheets/cheatsheets/version-control/git/commands/init.html

jondarrer:
@MichaelCurrin Upon further reflection, I'm thinking to keep it as `master`, as the rest of the document does in fact explicitly refer to this as the default branch 5 times throughout. And it does address the default branch question early on within the [Workspace setup](https://github.com/jekyll/jekyll/blob/master/docs/_docs/continuous-integration/github-actions.md#workspace-setup) section:

```md
We're only going to cover builds from the `master` branch in this page. Therefore, ensure that you
are working on the `master` branch. If necessary, you may create it based on your default branch.
```

I should have read things through a little more closely to begin with 🔎.

DirtyF (approved):

PR #8579, opened 2021-02-08
Use latest Jekyll-action configuration

jondarrer:

## Summary

The most recent version of the jekyll-action GitHub Action, 2.1.0, has changed the _environment variable_ `JEKYLL_PAT` to a _parameter_ called `token`.

https://github.com/marketplace/actions/jekyll-actions#deprecation

Adding this change should make it easier for anyone setting up the GitHub Action. Using the latest version of the action brings it into line with current documentation.

### DIFF
--- a/docs/_docs/continuous-integration/github-actions.md
+++ b/docs/_docs/continuous-integration/github-actions.md
@@ -113,9 +113,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: helaili/jekyll-action@2.1.0
+      - uses: helaili/jekyll-action@v2
         with:
-          token: ${{ secrets.JEKYLL_PAT }}
+          token: ${{ secrets.GITHUB_TOKEN }}
 ```
 
 {% endraw %}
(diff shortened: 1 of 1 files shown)


MichaelCurrin:
Hi. Thanks for picking up the deprecation. 

When I wrote this article, the GH token approach was no supported.

Can you make additional changes to the same page? There is a bullet point mention of JEKYLL_PAT which can come out. And the second time it is mentioned is in the Permissions section - that entire section can be removed now since one no longer needs to generate a custom token. GH generates the GH_TOKEN value automatically.

Another benefit of GH_TOKEN is its security - it scoped to one repo only while JEKYLL_PAT gives access to edit _all_ repos of a user.
Not that you have to mention that, but just for context.

MichaelCurrin (changes requested):
Remove all references to JEKYLL_PAT and creating a custom token.

When it comes to use GH_TOKEN, make sure to add a note around that saying one must copy and paste code as is (ie don't substitute to your own value in there). And mention that GH Actions will generate and use GH_TOKEN on a workflow run, so one does not need to generate it manually.

MichaelCurrin:
By the way, I just made the update in the demo repo which is linked to in tutorial 

It uses this new token flow

https://github.com/MichaelCurrin/jekyll-actions-quickstart/blob/v1.3.0/.github/workflows/jekyll.yml

jdvp:
Yeah, I think the entire "Providing Permissions" section can be removed. It might also be good to update again to the latest version (2.2.0)

For me, it also didn't work unless I specifically gave the target branch too, but unsure if anyone else has had that problem. Might want to include it as well.
```
  - uses: helaili/jekyll-action@2.2.0
    with:
      token: ${{ secrets.GITHUB_TOKEN }}
      target_branch: 'gh-pages'
```

DirtyF (approved):

DirtyF:
@jekyll merge +docs

PR #8585, opened 2021-02-18
Load Jekyll plugins from BUNDLE_GEMFILE location

bglw:
This is a 🐛 bug fix.

  - ✅ &nbsp;I've added tests 🎉 
  - ✅ &nbsp;The test suite passes locally 🎉 

## Summary

Before telling bundler to load plugins in the `:jekyll_plugins` group, `plugin_manager.rb` checks for the existence of a Gemfile in the current working directory.  
In some setups, the Gemfile may live elsewhere and be referenced using the `BUNDLE_GEMFILE` environment variable. In this case, the Gemfile check will fail and the `:jekyll_plugins` group will not be loaded.  
Bundler is aware of the file (since the gems still resolve), so all we need to do is expand the check for a Gemfile to include the BUNDLE_GEMFILE environment variable.

## Reproduction

For a given Jekyll site with plugins referenced in the Gemfile and **not** in the _config.yml, run the following in the parent directory:
```
export BUNDLE_GEMFILE=src/Gemfile 
bundle install
bundle exec jekyll build --source src
```
The plugins in the Gemfile plugins group will not be loaded.

## Context

I haven't found an existing issue or PR referencing this.

### DIFF
--- a/lib/jekyll/plugin_manager.rb
+++ b/lib/jekyll/plugin_manager.rb
@@ -46,7 +46,7 @@ module Jekyll
     end
 
     def self.require_from_bundler
-      if !ENV["JEKYLL_NO_BUNDLER_REQUIRE"] && File.file?("Gemfile")
+      if !ENV["JEKYLL_NO_BUNDLER_REQUIRE"] && gemfile_exists?
         require "bundler"
 
         Bundler.setup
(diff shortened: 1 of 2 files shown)


DirtyF (approved):
LGTM, thanks, this needs another pair or eyes from our Core Team.

DirtyF:
@jekyll: merge +fix

PR #8586, opened 2021-02-18
Relax dependency version constraints

oturpe:

This is a 🐛 bug fix.

- I've added tests (if it's a bug, feature or enhancement) **N/A**
- I've adjusted the documentation (if it's a feature or enhancement) **Self documenting**
- The test suite passes locally (run `script/cibuild` to verify this) **Ran with success until prompt "Your user account isn't allowed to install to the system RubyGems ... enter your password and install the bundled gems to RubyGems using sudo.", test run terminated on such unreasonable request.*

## Summary
During Fedora 34 development, Jekyll 4.2.0 has been tested against dependency versions outside of current definitions. This  commit relaxes constraints to allow some versions that were found compatible:

* `mercenary` was 0.3.6 in Fedora, works correctly. Nothing in [changelist][1] indicates problems, either.
* `terminal-table` was 1.8.0 in Fedora, then updated directly to 3.0.0. Both were tested and found to work well. The [changelist][2] also shows that breaking changes between terminal-table 1 and 3 are either deprecation of very old Ruby runtimes or tiny changes to terminal output.

[1]: https://github.com/jekyll/mercenary/blob/master/History.markdown
[2]: https://github.com/tj/terminal-table/blob/master/History.rdoc

## Context
Not related to existing issue


jekyll.gemspec
```diff
@@ -41,10 +41,10 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency("kramdown",              "~> 2.3")
   s.add_runtime_dependency("kramdown-parser-gfm",   "~> 1.0")
   s.add_runtime_dependency("liquid",                "~> 4.0")
-  s.add_runtime_dependency("mercenary",             "~> 0.4.0")
```
voxik, line 44: This downgrades the version from `>= 0.4.0 && < 0.5.0` to `>= 0.3.6 && < 0.4.0`. More about the `~>` pessimistic constraint can be found here:

https://guides.rubygems.org/patterns/#pessimistic-version-constraint

oturpe: Thank you for pointing this out! I mixed this with some other semver specification language, npm's probably. Changed to `"~> 0.3", ">= 0.3.6"` and tested that it can select 0.4.0.

voxik: Yes, this could work, if also 0.5 is acceptable for upstream

oturpe: Hmm, that was not my intent either. Simply to allow 0.3.6 and not change anything else. I am pessimistic about my ability to use the pessimistic constraint correctly. Better to stay with standard mathematical operators here. One last try: `">= 0.3.6", "< 0.5"`.

DirtyF (approved):
Thanks, I'll trust our test suite here 😅

PR #8590, opened 2021-02-26
Added max ruby version

jp26jp:
[Jekyll doesn't work with ruby 3.0.0](https://github.com/github/pages-gem/issues/752#issuecomment-764758292)

## Summary

Added max version of Ruby.

### DIFF
--- a/docs/_docs/index.md
+++ b/docs/_docs/index.md
@@ -14,7 +14,7 @@ tweak the site's look and feel, URLs, the data displayed on the page, and more.
 
 Jekyll requires the following:
 
-* Ruby version **{{ site.data.ruby.min_version }}** or higher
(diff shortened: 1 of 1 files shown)


ashmaroli:
You can use Jekyll with Ruby 3.0 if you want. Just ensure that the `webrick` gem is included in your Gemfile (or global environment if there's no Gemfile). Try running the following:
```
bundle add webrick
```

Refer https://github.com/jekyll/jekyll/issues/8523 for additional discussion if the above isn't sufficient.

jp26jp:
That should be added somewhere in the docs then.

bamaswade:
> 

lib/jekyll/tags/include.rb

PR #8587, opened 2021-02-18
Add documentation for Sass configuration options

seshrs:
This is a 🔦 documentation change.

## Summary

This PR adds documentation about Sass configuration options provided by [`jekyll-sass-converter`](https://github.com/jekyll/jekyll-sass-converter). Although a couple of these options were briefly explained in the Assets docs, the docs were not comprehensive. Also, as noted in #8570, it would be nice to point out that Sass load paths are resolved relative to the site's `source` path.

## Context

Closes #8570.

PR #8614, opened 2021-03-24
Use official Ruby setup GH action

seshrs:
This is a CI fix (housekeeping).

## Summary

I noticed that the CI results in #8587 mentioned that [`actions/setup-ruby`](github.com/actions/setup-ruby) is no longer maintained and that we should migrate to using [`ruby/setup-ruby`](github.com/ruby/setup-ruby) instead.

This PR updates jobs to use `ruby/setup-ruby`. I'm not super-familiar with Ruby and had to guess which steps were obsolete with the new action. Hopefully, the GH Actions results on this PR will help evaluate whether I got it right! 😅 

~EDIT: Haha that didn't work. Will try debugging later this week.~


seshrs:
Oh I see some overlap with #8610. Will wait for that to merge first.

.github/workflows/ci.yml
```diff
@@ -80,24 +80,13 @@ jobs:
...
-    - name: 'Update Rubygems and Bundler'
-      run: |
-        gem update --system --no-document
-        gem update bundler --no-document
-    - name: Set up bundle
-      run: |
-        bundle config path vendor/bundle
-        bundle install --jobs 4 --retry 3
+        bundler-cache: true
+      env:
```
MichaelCurrin, line 87: For this and the other  config, the docs recommend setting env vars at the job level rather than the step level.

https://github.com/ruby/setup-ruby#bundle-config

I don't know enough about the action to know if it makes a difference, but if i am uncertain I prefer to follow the instructions.

MichaelCurrin: Oh nevermind. I think it is just the action author doesn't know you can set the values at the step level.

MichaelCurrin (approved):

MichaelCurrin:
Looks good but I see there are conflicts to fix and this is WIP.

MichaelCurrin (changes requested):
Fix conflicting file.

seshrs:
@MichaelCurrin yes you were right, the bundle env variables had to be declared at the job-level so that other steps use the correct bundle config! Thanks for pointing that out 😃 

Given that the `Third-Party Repository Profiling` workflow completed successfully, I think we have more confidence that the `docs` workflow will also work when this PR is merged with `master`.

MichaelCurrin:
I recommend making a copy of your existing branch and take out the `on` restriction by branch, so you can see this run in your branch on your fork.

Then you can delete that branch later

MichaelCurrin:
You can find your branch marked with passing runs on the Actions tab 

https://github.com/jekyll/jekyll/actions

seshrs:
Hi @MichaelCurrin, sorry I dropped the ball for a couple of weeks 😅 

Is there some further action needed from me to push this PR forward?

MichaelCurrin:
No PR changes are outstanding. This just needs a review and merging by a maintainer - a reviewer with write access

parkr (approved):
Looks good to me!

I wasn't able to fully validate the bundler-cache option but I'm assuming we'll be able to continue benefitting from the Actions cache.

MichaelCurrin:
I didn't know about the option when I last reviewed this, but I saw the docs and comment on bundler cache flag here

https://github.com/ruby/setup-ruby#single-job


That means that in this PR, we the action already handles Bundle, bundle install and cache. And so the cache action and the gem install steps are unnecessary and should be taken out. To make the workflow lighter. And also to make the build short and not create conflicts potentially by having two caches and two gem install steps.

DirtyF:
@jekyll: merge +dev

PR #8618, opened 2021-03-26
Update include tag to be more permissive

parkr:
## Summary

Our `{% include %}` tag presently isn't permissive enough to handle many valid filenames, include `foo@bar.html`. This `@` symbol is used by systems within the Node ecosystem and the `include` tag is often used to selectively include HTML/JS/CSS from within a Node package in a Jekyll site. 

This does _not_ include any of the reserved characters: https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words

It adds support for the following characters:

- `@`
- `#`
- `~`
- `+`
- `-`
- `(` and `)`

## Context

This PR continues the work that @Convincible did in #7096.

We'd like to be able to use this in GitHub Pages, so I'll plan a back port to Jekyll 3.x as well.

### DIFF
--- a/lib/jekyll/tags/include.rb
+++ b/lib/jekyll/tags/include.rb
@@ -13,7 +13,7 @@ module Jekyll
       !mx.freeze
 
       FULL_VALID_SYNTAX = %r!\A\s*(?:#{VALID_SYNTAX}(?=\s|\z)\s*)*\z!.freeze
-      VALID_FILENAME_CHARS = %r!^[\w/.\-()+~\#@{}\[\]]+$!.freeze
+      VALID_FILENAME_CHARS = %r!^[\w/.\-()+~\#@]+$!.freeze
       INVALID_SEQUENCES = %r![./]{2,}!.freeze
 
       def initialize(tag_name, markup, tokens)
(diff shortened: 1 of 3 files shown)


mattr- (approved):
I'm fine with the additions as is, though I agree about both `{` and `}` being on the bubble for inclusion.

Thanks for prepping tests for this! :heart:

parkr:
@mattr- Thank you! I removed `[` `]` `{` and `}` for now. Easier to add them later than to try to remove them.

parkr:
> **Continuous Integration / Style Check (Ruby 2.5) (pull_request) ** Failing after 4m — Style Check (Ruby 2.5)

@ashmaroli Hi! Are these style checks failing on `master`? I am happy to fix them up but I didn't change anything about these failures in my file that I noticed... Thanks!!

DirtyF:
@parkr yes styles are also failing for master, will give it a look.

DirtyF:
@jekyll: merge +minor

parkr:
@DirtyF i can't remember - is there an easy way to backport this to 3.x? I can't remember if @jekyllbot can do it for us or not.

DirtyF:
@parkr I never done a backport but according to https://github.com/parkr/auto-reply/issues/31 it's still manual.

PR #7253, opened 2018-09-18
Don't reset site.url to localhost:4000 by default

ashmaroli:
Local tests show that the local preview server is based on `config["host"]` and `config["port"]` values..
So even if `config["url"] == https://jekyllrb.com`, the site will be served at the default setting: `//127.0.0.1:4000`

This simplifies users' workflow by **not having to `build` the site again with `JEKYLL_ENV=production`**
(unless their templates are designed to render differently for the two ENV vars, which is a different story altogether)

*P.S. The above observation was made on Windows. Please confirm for Unix platforms, if possible.*

PR #8620, opened 2021-03-29
Revert #7253: "Don't reset site.url to localhost:4000 by default"

benlk:
This is a 🐛 bug fix.
I have not run the CI suite.

## Summary

When running with `bundle exec jekyll serve` and no other options, the `absolute_url` filter will now use the local server URL instead of the URL from the config (like `benlk.com`) when that URL is the default `localhost:4000`.

## Context

This PR is designed to fix the regression introduced in https://github.com/jekyll/jekyll/pull/7253 and described in my comments there: https://github.com/jekyll/jekyll/pull/7253#issuecomment-776359879

If you don't want to check the links, here's a summary: 

- https://github.com/jekyll/jekyll/pull/7253 changed the behavior of `jekyll serve` to use the site URL from the site's config when serving a local development version of the site, if the localhost host and port were not explicitly specified. This meant that `site.url` and the `absolute_url` filter would use the production URL when generating and serving the site locally. The documentation on jekyllrb.org was not updated to reflect this change in behavior, so people who innocently ran `jekyll serve` and expected assets to load from their local development server would instead see:
    - images loaded from the production server, or 404 if not present on the production server
    - stylesheets loaded from the production server without changes from edits made locally, or 404 if not present on the production server
    - scripts loaded from the production server without changes from edits made locally, or 404 if not present on the production server
    - font files failing to load from production servers not configured to allow CORS requests for font files from `localhost:4000`, or 404 if not present on the production server
- That change did not affect cases where the local server URL was explicitly set via `bundle exec jekyll serve --host localhost --port 4000` to be `localhost:4000`, or when it was set to be any other combination of hosts or ports. That change only affected the default `jekyll serve`, which is the local development action recommended in the Jekyll docs.

This PR does update documentation with the recommendations from discussion in https://github.com/jekyll/jekyll/pull/7253

Reverts #7253

### DIFF
--- a/docs/_docs/usage.md
+++ b/docs/_docs/usage.md
@@ -23,8 +23,8 @@ Here are some of the most common commands:
 
 * `jekyll new PATH` - Creates a new Jekyll site with default gem-based theme at specified path. The directories will be created as necessary.
 * `jekyll new PATH --blank` - Creates a new blank Jekyll site scaffold at specified path.
-* `jekyll build` or `jekyll b` - Performs a one off build your site to `./_site` (by default).
-* `jekyll serve` or `jekyll s` - Builds your site any time a source file changes and serves it locally.
+* `jekyll build` or `jekyll b` - Performs a one off build your site to `./_site`, using the production configuration (by default).
+* `jekyll serve` or `jekyll s` - Builds your site any time a source file changes and serves it locally, using placeholders for the site URLs to enable it to be served locally (by default). Should not be used to generate a site copy that will be deployed to a live website.
 * `jekyll clean` - Removes all generated files: destination folder, metadata file, Sass and Jekyll caches.
 * `jekyll help` - Shows help, optionally for a given subcommand, e.g. `jekyll help build`.
 * `jekyll new-theme` - Creates a new Jekyll theme scaffold.
--- a/docs/_docs/usage.md
+++ b/docs/_docs/usage.md
@@ -23,8 +23,8 @@ Here are some of the most common commands:
 
 * `jekyll new PATH` - Creates a new Jekyll site with default gem-based theme at specified path. The directories will be created as necessary.
 * `jekyll new PATH --blank` - Creates a new blank Jekyll site scaffold at specified path.
-* `jekyll build` or `jekyll b` - Performs a one off build your site to `./_site`, using the production configuration (by default).
-* `jekyll serve` or `jekyll s` - Builds your site any time a source file changes and serves it locally, using placeholders for the site URLs to enable it to be served locally (by default). Should not be used to generate a site copy that will be deployed to a live website.
+* `jekyll build` or `jekyll b` - Performs a one off build your site to `./_site` (by default).
+* `jekyll serve` or `jekyll s` - Builds your site any time a source file changes and serves it locally.
 * `jekyll clean` - Removes all generated files: destination folder, metadata file, Sass and Jekyll caches.
 * `jekyll help` - Shows help, optionally for a given subcommand, e.g. `jekyll help build`.
 * `jekyll new-theme` - Creates a new Jekyll theme scaffold.
--- a/docs/_docs/step-by-step/01-setup.md
+++ b/docs/_docs/step-by-step/01-setup.md
@@ -77,10 +77,15 @@ called `_site`.
 * `jekyll serve` - Does `jekyll build` and runs it on a local web server at `http://localhost:4000`, rebuilding the site any time you make a change.
 
 {: .note .info}
-When you're developing a site, use `jekyll serve`. To force the browser to refresh with every change, use `jekyll serve --livereload`. If there's a conflict or you'd like Jekyll to serve your development site at a different URL, use the `--host` and `--port` arguments, as described in the [serve command options]({{ '/docs/configuration/options/#serve-command-options' | relative_url }}).
+When you're developing a site, use `jekyll serve`. To force the browser to refresh with every change, use `jekyll serve --livereload`.
+If there's a conflict or you'd like Jekyll to serve your development site at a different URL, use the `--host` and `--port` arguments,
+as described in the [serve command options]({{ '/docs/configuration/options/#serve-command-options' | relative_url }}).
 
 {: .note .warning}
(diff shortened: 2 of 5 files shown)


benlk:
Before merging, this PR should also update the usage docs https://github.com/jekyll/jekyll/edit/master/docs/_docs/usage.md to describe the intended use of build and serve in their default configurations, and how to override those configurations to account for the use that https://github.com/jekyll/jekyll/pull/7253 wanted to satisfy.

mattr-:
> Before merging, this PR should also update the usage docs https://github.com/jekyll/jekyll/edit/master/docs/_docs/usage.md to describe the intended use of build and serve in their default configurations, and how to override those configurations to account for the use that https://github.com/jekyll/jekyll/pull/7253 wanted to satisfy.

I agree with this comment. If this something you'll do soon? It doesn't seem productive to do anything else with this PR if it's not complete.

benlk:
@mattr- It's something that I'd prefer to leave to a maintainer, but I can draft some language and add it to the PR if you'd like.

mattr-:
> @mattr- It's something that I'd prefer to leave to a maintainer, but I can draft some language and add it to the PR if you'd like.

I would appreciate it if you would draft some language and include it in the PR. Thanks! :heart:

benlk:
@mattr- I've updated the docs in this PR to better differentiate between build and serve; I'm open to revisions in the language: https://github.com/jekyll/jekyll/pull/8620/commits/b298914c7a3e4dcaf39135c3547911ba7f9d9e4c

However, this doesn't address the latter part:
> and how to override those configurations to account for the use that #7253 wanted to satisfy.

Is the current admonition to see the full list of commands and build options okay for that purpose?

> For a full list of options and their argument, see [Build Command Options](/docs/configuration/options/#build-command-options).

docs/_docs/usage.md
```diff
@@ -23,8 +23,8 @@ Here are some of the most common commands:
 
 * `jekyll new PATH` - Creates a new Jekyll site with default gem-based theme at specified path. The directories will be created as necessary.
 * `jekyll new PATH --blank` - Creates a new blank Jekyll site scaffold at specified path.
-* `jekyll build` or `jekyll b` - Performs a one off build your site to `./_site` (by default).
-* `jekyll serve` or `jekyll s` - Builds your site any time a source file changes and serves it locally.
+* `jekyll build` or `jekyll b` - Performs a one off build your site to `./_site`, using the production configuration (by default).
```
ashmaroli, line 26: There is no *production* configuration technically. `jekyll serve` explicitly tests if the `env == "development"` which is the default environment. `jekyll build` does no such checks.

```suggestion
* `jekyll build` or `jekyll b` - Performs a one off build your site to `./_site` (by default).
```

benlk: Perhaps this phrasing, to remove the "production" word while still explaining _which_ configuration is used to do the build?

```suggestion
* `jekyll build` or `jekyll b` - Performs a one off build your site to `./_site`, using the configuration from `_config.yml` (by default).
```

ashmaroli: Perhaps you're misunderstanding or perhaps overthinking.
Stating `... configuration from '_config.yml ...' is redundant because there is no other configuration.

`jekyll build` executes a build in the context of the current directory with default settings. To override default settings, one may either use CLI flags like `--source`, `--config`, `--incremental`, etc or use their counterparts (if available) directly in the config file.
`jekyll serve` is just a superset of `jekyll build`. **It does everything `jekyll build` does verbatim but additionally overrides `site.config["url"]` and starts up a local webserver for the files in the destination directly.**

To summarize figuratively, the two commands can be see as:

```ruby
j_build = run_build_process
j_serve = j_build + override_site_url + start_server
```

benlk: > jekyll serve is just a superset of jekyll build. It does everything jekyll build does verbatim but additionally overrides site.config["url"] and starts up a local webserver for the files in the destination directly.

`jekyll serve` _does not currently override_ `site.config["url"]`. This has been the case since https://github.com/jekyll/jekyll/pull/7253 . My PR reverts that PR, and fixes documentation, for the reasons described in this PR's description.

I'm being explicit about _where_ the build settings for `jekyll build` come from, and what the intended use case of `jekyll build` is in contrast to `jekyll serve`. The reason I am being so painstakingly explicit is because this whole exercise comes from someone believing that `jekyll serve` built deployable, production-site-ready code into `/_site` like `jekyll build` does.

The revised line is meant to be read in contrast with the next line in the diff:

```
* `jekyll serve` or `jekyll s` - Builds your site any time a source file changes and serves it locally, using placeholders for the site URLs to enable it to be served locally (by default). Should not be used to generate a site copy that will be deployed to a live website.
```

I'm trying to prevent future occurrences of a misconception which has already occurred and which has already led to negative impacts on Jekyll users. For further reading, please see https://github.com/jekyll/jekyll/pull/7253#issuecomment-777653085

ashmaroli: > `jekyll serve` _does not currently override_ `site.config["url"]`

Yes. I agree. But this PR is undoing that behavior. So the documentation should be based on the future mindset.
Anyways, all this back-n-forth is not breaking the stalemate.
I'd like to ship this change via v4.2.1 in the near future.

So here's my recommendation:
- Please undo the changes you have currently made to `docs/_docs/usage.md`. Let's keep those summaries concise.
- Find another suitable place (`_docs/usage.md` included) to document the bullet-points raised in [#7253 (comment)](#7253#issuecomment-777786105)
    > - describe the effects of `serve` upon the variables it treats differently
    > - state that `serve` serves from the same build directory used by `jekyll build`, and encourage users to run `jekyll build` before deploying files from that directory to a production server.

benlk: These changes have been removed in https://github.com/jekyll/jekyll/pull/8620/commits/bf406f55ea6c48a655311a55a4475149f15c90cf

The bullet-points raised have been addressed in changes to the tutorial section. I thought about tacking them on at the bottom of the usage section, but that breaks the flow of the page, and the information would be better contained in the bullet points. Thus, addenda to the tutorial.

ashmaroli (approved):

mattr- (approved):

PR #8669, opened 2021-05-11
Increase h1 rule specificity in 404.html

kevinnls:

## Summary
In the temple 404 page
change CSS selector `h1` to `.container > h1` to avoid clashing with h1 elements in other parts of the page


DirtyF:
Thanks, but I don't feel this change is really necessary, those are basic inline styles local to the page, people are free to come up with their custom code for their 404 page. It's just meant as a default placeholder, not production code.

PR #8676, opened 2021-05-16
Revert "style: run rubocop -a"

kmk3:
This is a 🐛 bug fix.

## Summary

This reverts commit 5631170d443c01f85e5c59198bb0847ffdac4af1.

It fixes a warning when using Ruby 3.0, but brings about another one
when using Ruby 2.5, which is what ci uses.  From the logs of that
commit's workflow run[1]:

    +style-check | RUBY=2.5
    +style-check | --> RUN script/fmt
    +style-check | RuboCop 1.12.1
    +style-check | The following cops were added to RuboCop, but are not configured. Please set Enabled to either `true` or `false` in your `.rubocop.yml` file.

    +style-check | Please also note that can also opt-in to new cops by default by adding this to your config:
    +style-check |   AllCops:
    +style-check |     NewCops: enable
    +style-check | Minitest/AssertWithExpectedArgument: # (new in 0.11)
    +style-check |   Enabled: true
    +style-check | Performance/MapCompact: # (new in 1.11)
    +style-check |   Enabled: true
    +style-check | Performance/RedundantEqualityComparisonBlock: # (new in 1.10)
    +style-check |   Enabled: true
    +style-check | Performance/RedundantSplitRegexpArgument: # (new in 1.10)
    +style-check |   Enabled: true
    +style-check | For more information: https://docs.rubocop.org/rubocop/versioning.html
    +style-check | Inspecting 134 files
    +style-check | .C....................................................................................................................................

    +style-check | Offenses:

    +style-check | lib/jekyll/cache.rb:170:23: C: Security/MarshalLoad: Avoid using Marshal.load.
    +style-check |       value = Marshal.load(cached_file)
    +style-check |                       ^^^^

    +style-check | 134 files inspected, 1 offense detected

    +style-check | Try running `script/fmt -a` to automatically fix errors
    +style-check |
    +style-check | Command /bin/sh -c script/fmt failed with exit code 1
    +style-check |
    +style-check | ERROR: Command exited with non-zero code: RUN script/fmt

[1] https://github.com/jekyll/jekyll/runs/2586877102

## Context

Relates to #8654.

PR #8675, opened 2021-05-14
Spell check action for markdown documentation

jsoref:
This is a 🙋 feature or enhancement.

## Summary

Add the [check-spelling action](https://github.com/marketplace/actions/check-spelling
) limited to `docs/` directory `.md` files.

## Context

Based on @DirtyF https://github.com/jekyll/jekyll/pull/8661#issuecomment-841496512

Note: as currently configured, pushes get a comment, and PRs get another, this may feel like duplication, but it enables contributors to enable actions in their forks and get feedback *before* creating a PR.

I'm still looking into ways to address this duplication.

<details><summary>It's software, as with everything else, so there are bugs</summary>

* There's one glitch where the comment has a duplicated word, it's fixed in the `prerelease` branch, but generally I prefer for public projects to use released versions).
* This version has a flag guarded feature for updating PRs, but it shouldn't be used in public repositories as it has a bug where it'll comment prematurely for things not addressed to the bot (this is also fixed in `prerelease`).

There's one more feature that I'm working on which I'd like to finalize before my next release (hopefully in a week or so, I failed to notice where I was in the calendar, which is very ironic, as I've been literally counting the days).
</details>

### Landing advice
For any fast moving repository, the initial merge tends to be a bit rocky. Between when #8661 was written and when it was merged, two additional typos were added just to the documentation directory.

The person merging this should enable actions in their fork and test the merged result (the easiest thing to do is make this branch the repository's default and create a PR from [jekyll/jekyll](https://github.com/jekyll/jekyll) `master` into their repository (with `spell-check` as the default) -- you can use [check-spelling/jekyll](https://github.com/check-spelling/jekyll) as that repository if you like, as it's set up (but I won't be able to make changes to it, so you're probably better served using your own fork). If it gets a ✅ from check-spelling, then it's safe to merge. If it gets a ❌ , there will be a comment identifying the words that are unrecognized. They can be added to `expect.txt` if they're intended, or fixed if they're errors.

It's best to do that before merging. If you merge with a ❌ then anyone who uses that landing point for their branch point will get slightly confusing ❌ feedback on their branch. (Their PR could still report a ✅ if `master` fixed because of the configuration of the workflow, but the confusion is confusing.)

---
Sadly, I won't be able to provide much help between now and Wednesday. Generally I'm fairly responsive (w/in 2 days).

If you have a communications channel that you use and want help w/ the landing, let me know (again, I can help from Wednesday). Have a good weekend/week.

PR #8661, opened 2021-05-03
Spelling

jsoref:
This is a 🐛 bug fix.

## Summary

This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

## Context

The misspellings have been reported at https://github.com/jsoref/jekyll/commit/117f54927e27e1178428174c60aea7719fa0ecea#commitcomment-50300464

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/jekyll/commit/beb09ef03f42a45765ee6f12855bd70443ccd2c5

Note: this PR does not include the action. If you're interested in running a spell check on every PR and push, that can be offered separately.

PR #8677, opened 2021-05-16
Update expect to cover docs/_posts

jsoref:
This is a 🐛 bug fix.

## Summary

check-spelling is currently checking `docs/` (including `_posts/`) but when its original corrections were landed, they ignored `_posts/`, so certain additional items needed to be added to `expect.txt`

## Context

This applies https://github.com/jekyll/jekyll/pull/8676#issuecomment-841849062 (which was introduced with the landing of #8675)


kmk3:
So I started writing #8678 before this was created and ended afterwards; I'll
just copy part of what I said there:

> Suggestion: Ignore the `docs/_posts` directory to avoid having to start
> adding/maintaining contributor names on the dictionary.

On a related note, while I see the value in spell checking code and
documentation (which are usually the most frequently changed parts), I noticed
that in #8661 there were spelling changes done to the files in `docs/_posts`.
Personally, I would have refrained from doing so, as many of them are
old/historical publications (e.g.: one of the files changed is from 2013),
whose textual content I would consider rather "set in stone" at this point in
time.

DirtyF (approved):

DirtyF:
@jekyll: merge +dev

PR #8680, opened 2021-05-17
fix(security):  CVE-2021-28834

DirtyF:
We should force Kramdown to >=2.3.1 to avoid https://github.com/advisories/GHSA-52p9-v744-mwjj

### DIFF
--- a/jekyll.gemspec
+++ b/jekyll.gemspec
@@ -38,7 +38,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency("i18n",                  "~> 1.0")
   s.add_runtime_dependency("jekyll-sass-converter", "~> 2.0")
   s.add_runtime_dependency("jekyll-watch",          "~> 2.0")
-  s.add_runtime_dependency("kramdown",              "~> 2.3")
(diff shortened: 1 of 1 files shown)


mattr- (approved):

jekyll.gemspec
```diff
@@ -38,7 +38,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency("i18n",                  "~> 1.0")
   s.add_runtime_dependency("jekyll-sass-converter", "~> 2.0")
   s.add_runtime_dependency("jekyll-watch",          "~> 2.0")
-  s.add_runtime_dependency("kramdown",              "~> 2.3")
+  s.add_runtime_dependency("kramdown",              "~> 2.3.1")
```
parkr, line 41: This locks kramdown to 2.3.x, rather than 2.x. I think you want something like this:

```suggestion
  s.add_runtime_dependency("kramdown",              "~> 2.3", ">= 2.3.1")
```

DirtyF:
@jekyll: merge +fix

PR #8681, opened 2021-05-18
Improve the "Markdown Options" Docs

mkasberg:
This is a 🔦 documentation change.

## Summary

It seems the Markdown Options docs have become out of date, as noted in
https://github.com/jekyll/jekyll/issues/8593. The docs most likely
became out of date when Kramdown was updated to v2, as described in
these release notes:
https://jekyllrb.com/news/2020/08/05/jekyll-3-9-0-released/.

Initially, I intended to fix the specific issue about `gfm_quirks` noted
in #8593, but after digging a little deeper to figure out the correct
way to use that option, it became apparent that other parts of the
Markdown Options Jekyll docs were also outdated and the whole page
needed to be updated to be more accurate and more useful for modern
versions of Jekyll.

So, with that in mind, here's an overview of the changes I'm proposing:

* Adjust the page headings. `Kramdown` should be an `h2` and the
  Kramdown sub-headings should be `h3`.
* Document modern config for the GFM parser via `input: GFM`.
* Link to the CodeRay docs and provide an example of the correct
  `syntax_highlighter` config with `syntax_highlighter_opts`.
* Link to the options supported by Kramdown and provide an example of
  their usage.
* Remove the list of all the relatively advanced Kramdown options
  supported by Jekyll, directing users instead to the Kramdown docs.

At a high level, this is a move away from attempting to document every
supported Kramdown option on this page and a move toward linking to
up-to-date documentation for the various gems we support. The old docs
were outdated, and not as useful as the docs provided by individual gems
because the Jekyll docs didn't provide thorough descriptions as the
[Kramdown docs](https://kramdown.gettalong.org/options.html) do. I think
my proposed change captures a good balance of showing example usage in
Jekyll while still relying on the gems we depend on to provide thorough
descriptions of the supported options.

## Context

Fixes #8593

Issue #8593, opened 2021-03-05
docs: update markdown options to kramdown 2


tordans:
I think the docs at
- https://github.com/jekyll/jekyll/blob/master/docs/_docs/configuration/markdown.md
- https://jekyllrb.com/docs/configuration/markdown/

are outdated.

At least the option `gfm_quirks:true` does not work. It says:

```
       Jekyll Feed: Generating feed for posts
  Conversion error: Jekyll::Converters::Markdown encountered an error while converting '404.md':
                    Invalid type TrueClass for option gfm_quirks
bundler: failed to load command: jekyll (/Users/tobiasjordans/.asdf/installs/ruby/3.0.0/bin/jekyll)
```

This is in line with https://jekyllrb.com/news/2020/08/05/jekyll-3-9-0-released/ which I read as: GFM will not work out of the box.

However, `bundle add kramdown-parser-gfm` and using 
```
kramdown:
  input: GFM
```
from https://www.markdownguide.org/tools/jekyll/ does work.

It might be, that the page still references kramdown 1.x, not the new 2.x?

## Suggestion

Update the docs.

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the latest <code>3.x-stable</code> or <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, please consider building it first as a plugin. Jekyll 3 introduced [hooks](http://jekyllrb.com/docs/plugins/#hooks) which provide convenient access points throughout the Jekyll build pipeline whereby most needs can be fulfilled. If this is something that cannot be built as a plugin, then please provide more information about why in order to keep this issue open.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

tordans:
@jekyllbot its still an issue.

mkasberg:
I noticed this was tagged `help-wanted`, so I took a look. This was a bit more complex than I was expecting, but I think I've come up with a good solution in #8681.

A few notes specific to this issue:

`gfm_quirks: true` is actually incorrect usage. If you want to use GFM syntax, you can do so like this:

```yaml
kramdown:
  input: GFM
```

[...]

PR #8684, opened 2021-05-19
Improve GitHub Flavored Markdown Docs

mkasberg:

## Summary

I recently merged #8681 to improve the Jekyll docs for markdown options.
While researching that, I noted that the `kramdown-parser-gfm` gem is
required by default in Jekyll, but I didn't dig further as to why.

After #8681 merged, I was curious about why `kramdown-parser-gfm` is
required by default, and I followed up on digging into it. As it turns
out, this GFM parser is required by default because Jekyll uses GFM by
default. This is clear from the [default config
docs](https://jekyllrb.com/docs/configuration/default/) and the code:

https://github.com/jekyll/jekyll/blob/76517175e700d80706c9139989053f1c53d9b956/lib/jekyll/configuration.rb#L67-L72

Although this is outlined in the default configuration docs, other parts
of the docs don't do a good job making it clear that GFM is actually the
default behavior. My recent docs change in #8681 made this problem worse
since I used a bunch of config in my examples that's actually just
default config.

In this PR, I'm addressing that problem by modifying my update to the
Markdown Options docs to clearly note that GFM is the default parser.
I'm also updating a tutorial that implied GFM isn't the default. I've
searched this repo for "GFM" and I'm pretty confident I've found all the
problematic docs.

## Context

This follows up on #8681, which was originally created in response to #8593.

### DIFF
--- a/docs/_docs/configuration/markdown.md
+++ b/docs/_docs/configuration/markdown.md
@@ -9,22 +9,23 @@ available.
 
 Kramdown is the default Markdown renderer for Jekyll, and often works well with no additional configuration. However, it does support many configuration options.
 
-### GitHub Flavored Markdown
+### Kramdown Processor
 
-Kramdown supports GitHub Flavored Markdown (GFM). To use GFM with Kramdown in Jekyll, add the following to your configuration.
-
-```yaml
-kramdown:
-  input: GFM
-```
-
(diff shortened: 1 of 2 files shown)


DirtyF (approved):
Thanks @mkasberg 🙏

DirtyF:
@jekyll: merge +docs

PR #8718, opened 2021-06-27
Inject livereload script using `location.protocol` instead of `http:`

kaysond:
This is a 🐛 bug fix.

## Summary

If you run `jekyll serve --livereload` behind an ssl-terminating reverse proxy (or, presumably, using jekyll's own ssl options), the livereload script gets loaded over `http`. Browsers like Firefox will reject this mixed content load.

Instead of forcing the injection to use `http`, this change uses `location.protocol` which dynamically chooses `http` or `https` depending on the URL in the browser.

This API is well supported: https://caniuse.com/mdn-api_location_protocol

### DIFF
--- a/lib/jekyll/commands/serve/servlet.rb
+++ b/lib/jekyll/commands/serve/servlet.rb
@@ -101,7 +101,7 @@ module Jekyll
           @template ||= ERB.new(<<~TEMPLATE)
             <script>
               document.write(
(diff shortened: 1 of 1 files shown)


mattr- (approved):

DirtyF (approved):

DirtyF:
@jekyll: merge +fix

PR #8725, opened 2021-07-02
Implements sort_order attribute for collections

Convincible:

## Summary

Allows the sort order of collections to be reversed from the _config.yml setting `sort_order`.

This is a minimal implementation that bolts onto the existing approach. However the collection is sorted (even no sorting, i.e. default/alphabetically), the user can simply reverse this.

This is particularly useful for date ordering, where ascendant and descendant are equally valid sort orders depending on the circumstance.

## Context

Picks up #3735 where @parkr suggested a simple implementation of @goodtouch's suggestion to allow collection sorting to be more configurable.

### DIFF
--- a/lib/jekyll/collection.rb
+++ b/lib/jekyll/collection.rb
@@ -247,6 +247,8 @@ module Jekyll
 
         # Finally restore the `docs` array with just the Document objects themselves
       end.map!(&:last)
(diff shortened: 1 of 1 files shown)


fauno:
It would be cool if posts collection wasn't sorted by date in `Jekyll::Drops::SiteDrop`

PR #8727, opened 2021-07-02
Add 'webrick' warning note to "Quickstart" Docs

miguelbrandao:

## Summary

For Ruby 3.0.0 and higher, `bundle exec jekyll serve` will fail because of a missing dependency: `webrick`.

The documentation doesn't mention this problem, while it states that all Ruby versions above 2.4.0 are supported. This is confusing to new users who will probably install the latest Ruby version and thus will be unable to move past step 5 of the [Quickstart](https://jekyllrb.com/docs/).

This PR aims to put a band-aid in this situation by adding a warning at the end of the instructions, allowing affected users to quickly identify and solve the problem.

image

## Context

I came across this problem when starting a new site and getting a bunch of errors on step 5 of the Quickstart.
A quick search lead me to [this discussion](https://github.com/github/pages-gem/issues/752) which explained what was happening and provided a working solution.

### DIFF
--- a/docs/_docs/index.md
+++ b/docs/_docs/index.md
@@ -41,6 +41,9 @@ bundle exec jekyll serve
 ```
 6. Browse to [http://localhost:4000](http://localhost:4000){:target="_blank"}
 
+{: .note .warning}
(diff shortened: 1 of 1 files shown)


hendramanu:
Agree. It should included in docs. Almost all Jekyll template I found are getting webrick error.

DirtyF (approved):

DirtyF:
@jekyll: merge +docs

PR #8778, opened 2021-08-21
Add warning about `webrick` to 01-setup.md

kronaemmanuel:

## Summary

Since Ruby version 3.0.0 or higher do not come with [webrick](https://github.com/github/pages-gem/issues/752). There is already a warning on the [Quickstart](https://jekyllrb.com/docs/) page. I have added a similar warning on `01-setup.md`.

Alternatively, a warning can also be added on line 34 to put in `gem "webrik"` if using Ruby version >= 3.0.0. Lemme know if you want to do it this way, I'll do the changes.

## Context

Related to: [Webrick issue on pages-gem repo](https://github.com/github/pages-gem/issues/752)

However, I couldn't find an issue already open on this repo. Should I open a new issue for this?

### DIFF
--- a/docs/_docs/step-by-step/01-setup.md
+++ b/docs/_docs/step-by-step/01-setup.md
@@ -76,6 +76,9 @@ before we can view it. Run either of the following commands to build your site:
 called `_site`.
 * `jekyll serve` - Does `jekyll build` and runs it on a local web server at `http://localhost:4000`, rebuilding the site any time you make a change.
 
+{: .note .warning}
+If you are using Ruby version 3.0.0 or higher, running the above commands [may fail](https://github.com/github/pages-gem/issues/752). You may fix it by adding `webrick` to your dependencies: `bundle add webrick`
+
 {: .note .info}
 When you're developing a site, use `jekyll serve`. To force the browser to refresh with every change, use `jekyll serve --livereload`. 
 


ashmaroli:
@kronaemmanuel Thank you for opening this PR.
I'd like to inform you that we have already added `webrick` as a dependency of Jekyll. Therefore, it will be automatically installed and loaded from next release onwards.

Consequently, whether this change is necessary or not is up for debate.

kronaemmanuel:
@ashmaroli,
Thankyou for your reply. I guess writing a warning just for people for people who are using Ruby version 3.0.0 and greater but are using Jekyll version lower than the latest release (3.9.1 i guess) wouldn't be practical. They can find what they need to do just by googling i guess.

Please feel free to close this PR then.

evanwill:
@ashmaroli it seems like this is still an issue in new release 4.2.1? 
I am running into the `webrick` issue a lot with people getting started--> they install ruby 3 (since it is most recent stable), then Jekyll and Bundler, then get a very confusing error when trying to jekyll serve. Very hard to new learners to debug.

ashmaroli:
@evanwill `v4.3.0` onward will include `webrick` as a dependency. Older versions wont.

evanwill:
Thanks @ashmaroli , looking forward to v4.3!

iuriguilherme:
It's been over a year and I ran into this issue, so I'm adding to the debate that this change was necessary one year ago, as it's still necessary now.

I can add to the debate my opinion that when 100% of installs use 4.3 it'll be no longer necessary.

My environment (WSL2 Ubuntu LTS):

```
$ ruby --version
ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-linux-gnu]
$ gem --version
3.3.5
$ bundle --version
Bundler version 2.3.16
$ jekyll --version
jekyll 4.2.2
```

dogweather:
Ok, a year later — could we document this problem or release 4.3.x?

Jekyll breaks for everyone who tries it. They have to land here and search for workarounds in the comments.

ashmaroli:
Closing this as no longer relevant.

PR #8801, opened 2021-09-06
WIP: Wasteful entryfilter

### LINKED ISSUE #7727: Wasteful allocations in EntryFilter
Under certain circumstances. the following methods can result in wastage of resources:
https://github.com/jekyll/jekyll/blob/62959527dd870691107739724c23b51ba03828f3/lib/jekyll/entry_filter.rb#L45-L53

For example:
```ruby
entry = 'about.md'

entry == File.basename(entry)     # => true
entry.eql? File.basename(entry)   # => true
entry.equal? File.basename(entry) # => false. (Same value but different objects)
```
In such situations, testing if an enumerator contains either `entry` or `File.basename(entry)` is simply a repetition involving an extra allocation when the first check fails.

IMO, this can be optimized. But I'm not able to see a simple solution.

fauno:
This is a 🔨 code refactoring.

## Summary

Work in progress for Wasteful allocations in EntryFilter 

## Context

Resolves #7727

### DIFF
--- a/lib/jekyll/entry_filter.rb
+++ b/lib/jekyll/entry_filter.rb
@@ -67,7 +67,7 @@ module Jekyll
     end
 
     def excluded?(entry)
-      glob_include?(site.exclude - site.include, relative_to_source(entry)).tap do |excluded|
+      glob_include?(excluded, relative_to_source(entry)).tap do |excluded|
         if excluded
           Jekyll.logger.debug(
             "EntryFilter:",
@@ -75,6 +75,10 @@ module Jekyll
           )
         end
       end
+    end
+
+    def excluded
+      @excluded ||= site.exclude - site.include
     end
 
     # --
--- a/lib/jekyll/entry_filter.rb
+++ b/lib/jekyll/entry_filter.rb
@@ -53,7 +53,7 @@ module Jekyll
     end
 
     def special?(entry)
-      SPECIAL_LEADING_CHAR_REGEX.match?(entry) || (entry.include?(File::Separator) && SPECIAL_LEADING_CHAR_REGEX.match?(File.basename(entry)))
+      SPECIAL_LEADING_CHAR_REGEX.match?(File.basename(entry))
     end
 
     def backup?(entry)
--- a/lib/jekyll/entry_filter.rb
+++ b/lib/jekyll/entry_filter.rb
@@ -22,6 +22,7 @@ module Jekyll
       base_dir
     end
 
+    # NOTE: Shouldn't this be a private method?
     def relative_to_source(entry)
       File.join(
         base_directory, entry
--- a/lib/jekyll/entry_filter.rb
+++ b/lib/jekyll/entry_filter.rb
@@ -53,13 +53,11 @@ module Jekyll
     end
 
     def included?(entry)
-      glob_include?(site.include, entry) ||
-        glob_include?(site.include, File.basename(entry))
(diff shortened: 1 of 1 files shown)


fauno:
Agh, sorry I'm too used to Gitlab and thought it was the same for marking a PR as draft (you add "WIP:" or "Draft:" to its title)

lib/jekyll/entry_filter.rb
```diff
@@ -76,6 +71,10 @@ def excluded?(entry)
       end
     end
 
+    def excluded
```
ashmaroli, line 74: Presently, I don't see a need for this to be exposed as a public method.
So, please make it private.

lib/jekyll/entry_filter.rb
```diff
@@ -99,13 +98,14 @@ def symlink_outside_site_source?(entry)
     # Check if an entry matches a specific pattern.
     # Returns true if path matches against any glob pattern, else false.
     def glob_include?(enumerator, entry)
-      entry_with_source = PathManager.join(site.source, entry)
+      current_directory = PathManager.join(site.source, base_directory)
```
ashmaroli, line 101: In the initializer, we duplicate given absolute path and strips `site.source`  from it  (scope for optimization).
And here, we're going to recreate the absolute path and cache it? Why not initialize a new instance variable containing the absolute path?

ashmaroli (commented):
The failing unit test is legit considering how `_includes/tmp` is indeed [a symlink pointing outside](https://github.com/jekyll/jekyll/blob/v4.2.0/test/source/_includes/tmp) the test-site-source

lib/jekyll/entry_filter.rb
fauno: Done

fauno:
> The failing unit test is legit considering how `_includes/tmp` is indeed [a symlink pointing outside](https://github.com/jekyll/jekyll/blob/v4.2.0/test/source/_includes/tmp) the test-site-source

Yes, but it fails when it stops being treated as a special field ("tmp" doesn't start with a special character, the directory does).  I'll check if it's a side effect of my changes though

lib/jekyll/entry_filter.rb
fauno: Right! I'm tempted to change `derive_base_directory` to always be an absolute path, but it'll merit a deprecation warning and other changes (specially in `excluded?` and `relative_to_source`).  Do you think I should go that way or just add a new instance var?

Thanks for reviewing :)

fauno:
> > The failing unit test is legit considering how `_includes/tmp` is indeed [a symlink pointing outside](https://github.com/jekyll/jekyll/blob/v4.2.0/test/source/_includes/tmp) the test-site-source
> 
> Yes, but it fails when it stops being treated as a special field ("tmp" doesn't start with a special character, the directory does). I'll check if it's a side effect of my changes though

The `symlink?` method returns false for the `_include/tmp` symlink because the current dir for `File.symlink?` is the project root instead of `test/source/`, so I think it's another candidate for using absolute paths.

lib/jekyll/entry_filter.rb
ashmaroli: new instance variable to bypass all that deprecation. Something along the lines of `@absolute_base_dir`

fauno:
Didn't abandon this, just can't make time to keep working on it!

PR #8815, opened 2021-09-17
Propagate _data folder from theme

mgerzabek:
All files from _data folder within theme are accessible in consumer project. The behaviour is exactly the same as for overrides of _includes and _layouts. Data files with the same name in consumer project override data files within remote theme.

A similar PR was originally created for @benbalter [jekyll-remote-theme](https://github.com/benbalter/jekyll-remote-theme/pull/89) which didn’t get the attention of the maintainers.

On 23rd of July @parkr [said](https://github.com/benbalter/jekyll-remote-theme/pull/89#issuecomment-885787446)

> I think this could be a really useful feature for all Jekyll themes. Would you consider submitting the change to the Jekyll project upstream?

So this my try on getting it into Jekyll upstream. And hopefully this way I can get what I wanted first, being able to access _data folder content in my remote themes :-)

  - I didn't add tests because I’m new to ruby and happy enough to got the code working. If someone could point me to where and how to add tests, I’d happily learn this…
  - I didn’t adjusted the documentation simply because I’m not sure where to add that piece of documentation. Please point me to the corresponding locations and I’ll do it…
  - The test suite passes locally… but, I get the message `lib/jekyll/reader.rb:14:5: C: Metrics/AbcSize: Assignment Branch Condition size for read is too high. [<2, 26, 0> 26.08/23]` where I don’t really know how to go about that.

## Prior Art

- [`jekyll-data`](https://rubygems.org/gems/jekyll-data) &mdash; plugin authored by @ashmaroli.
- **#5470**  &mdash; prior attempt by @ashmaroli to incorporate this feature in Core.


ashmaroli:
@mgerzabek Have you tried using the `jekyll-data` plugin also referenced in your prior pull request?
There was an attempt to add this functionality to Core but I pulled the plug because unlike includes and layouts, data files were not strictly referenced by their filenames.
Data files provide a high degree of flexibility.

Consider the following object: `site.data.movies.english`. Will you be able to determine with certainty how the object originated?
Was it 
- via `_data/movies/english{.y(a)ml, .json, .csv, .tsv}` or 
- via `english:` key map in `_data/movies{.y(a)ml, .json, .csv, .tsv}`
..?

mgerzabek:
> Have you tried using the `jekyll-data` plugin also referenced in your prior pull request?

No, as I’d been only been aiming at making it work in junction with `jekyll-remote-theme`. 

So, regarding your question, I suppose there’d be a `_data/movies{.y(a)ml, .json, .csv, .tsv}` file with a key `english` as my implementation is only looking into the `_data` folder. No recursive directories allowed here.

ashmaroli:
> I’d been only been aiming at making it work in junction with `jekyll-remote-theme`.

Please try using the plugins in tandem.
```yaml
# in _config.yml
plugins:
  - jekyll-remote-theme
  - jekyll-data
```

> to remove the recursive directory walking since it’s good enough to distribute data files within _data alone. This would make your second option the only possible way..

That is a good alternative I had not considered. The plugin I wrote is recursive within themes as well for parity with behavior for data files at source. I see your current implementation is recursive as well.
However, that doesn't mean there's no scope for further developments. If you can improve the implementation, add adequate test coverage, covering various use-case scenarios , I'll reconsider adding to Core.

*Do note that if your primary need is for deploying with GitHub Pages, don't have high hopes. GHP is locked to Jekyll 3.9. This feature when ready will only go into Jekyll 4.x*

mgerzabek:
> Do note that if your primary need is for deploying with GitHub Pages, don't have high hopes. GHP is locked to Jekyll 3.9. This feature when ready will only go into Jekyll 4.x

I understand and thank you for your clear words! 

This makes me rethink my initial considerations that where based on GHP via `jekyll-remote-theme` being able to also mirror the `_data` folder from the theme… if there is anyone out there who wants this PR in Jekyll 4+ Core, please vote with +1. If there are more than 50 votes until 30th of September 2021, I’d improve the implementation and add the tests to comply to @ashmaroli requirements.

Otherwise I’d just leave this alone since, as I mentioned previously, I’m new to Ruby and the extra work means a lot of efforts beside my day to day job.

(This BTW somehow makes me feel the original PR for @benbalter `jekyll-remote-theme` was the right location anyway…)

ashmaroli:
> if there is anyone out there who wants this PR in Jekyll 4+ Core, please vote with +1. If there are more than 50 votes until 30th of September 2021,

@mgerzabek This could sound a bit obtuse, but *the chances of interested users seeing the comment above by Sep 30th is very slim*. To increase the chances of visibility, you may want to consider utilizing platforms such as [Jekyll Talk](https://talk.jekyllrb.com/), [Reddit](https://reddit.com/r/jekyll) or Twitter.

> feel the original PR for `jekyll-remote-theme` was the right location anyway…

Technically, no. It wasn't the right location.
`jekyll-remote-theme` is meant to do one thing: *Download theme repo from GitHub and allow users to consume the downloaded theme as a local theme.*
What Jekyll does with the theme is not the plugin's primary concern.

However, since your primary need is to get this feature working on GHP and `jekyll-remote-theme` is a plugin revolving around themes, you could make an argument.

> I’m new to Ruby and the extra work means a lot of efforts beside my day to day job.

My humble request would be to consider trying to migrate from GHP and use GitHub Actions to deploy your site.
You don't need prior Ruby programming experience for that. Once set up you can just copy your implementation here to a `*.rb` file inside `_plugins` folder at source and Jekyll (all versions) will pick it up.

mgerzabek:
> _the chances of interested users seeing the comment above by Sep 30th is very slim_.

Right, so I‘ll edit the deadline to be 31st of October and do a post on [Jekyll Talk](https://talk.jekyllrb.com/t/propagate-data-folder-from-theme/6437)

michaelbach:
+1

pjteddie13 (approved):

parkr:
I'm comfortable reading YAML/JSON from `_data/`. Any other file types (like icons, css, JS, etc) should be placed in assets/ instead.

Have we decided how we want precedence to work? We should make sure our docs are clear for this, preferably with an example.

👍

datapolitical:
One of my major use cases is a package of json files that goes along with my theme that is kept in _data/drinks. These get read and processed into pages by Jekyll-datapage-gen.￼￼￼

If you don't want to scan all subfolders recursively for yml/json, you could allow explicit specification of folders to be scanned￼￼￼ in _config.yml

ashmaroli:
@parkr Please refer https://github.com/jekyll/jekyll/pull/8815#issuecomment-921709318 and let us know if the concern can be handled without impacting maintainability

datapolitical:
> @parkr Please refer [#8815 (comment)](https://github.com/jekyll/jekyll/pull/8815#issuecomment-921709318) and let us know if the concern can be handled without impacting maintainability

I guess I’m a little confused, doesn’t that ambiguity already exist with the _data folder itself? Doesn’t exist with the use of Jekyll-data?

What makes it any worse putting it into core?

ashmaroli:
> Doesn’t exist with the use of your plugin Jekyll-data?

Yes, It does exist with `jekyll-data`.
 
> What makes it any worse putting it into core?

Having issues with a plugin is one thing. Bringing a known issue into Core is another.
A plugin can be disabled / abandoned (by the user) if it is causing complications. Making changes to the Core comes with lot of overhead (adherence to semver, provide additional feature to enable/disable functionality, interference with other moving parts in Core, affect on users fine with published API, etc..)

datapolitical:
That makes sense. So if local data always overrides theme￼ ￼I don't see how it would cause confusion.￼

I do think it makes sense to allow users to be explicit about which folders to include if they want.

In my case, the theme would come with 500 or so files and a user may only want a handful of them￼￼.

If they create a _data/drinks folder and place three new yml files inside, default behavior would be to include all the others as well, and some/many may not want that. Giving people a way to say "not this folder" or "only this folder" makes sense.￼

lib/jekyll/readers/theme_data_reader.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Jekyll
+  class ThemeDataReader < Jekyll::DataReader
+    attr_reader :site, :content
+
+    def read(dir)
+      return unless site.theme&.data_path
+
+      base = site.in_theme_dir(dir)
```
parkr, line 10: I believe theme directories are standardized and not configurable. I would feel more comfortable releasing this with only one name for the theme data dir. Our intention was always to make it so a theme could be a Jekyll site, so I'd recommend the default `_data`. So `site.in_theme_dir(site.theme.data_path)`, I think?

lib/jekyll/readers/theme_data_reader.rb
```diff
@@ -0,0 +1,31 @@
...
+      read_data_to(base, @content)
+      Jekyll::Utils.deep_merge_hashes(@content, @site.data)
+    end
+
+    def read_data_to(dir, data)
+      return unless File.directory?(dir) && !@entry_filter.symlink?(dir)
+
+      entries = Dir.chdir(dir) do
+        Dir["*.{yaml,yml,json,csv,tsv}"]
```
parkr, line 19: The superclass also includes reading subdirectories. If we're not going to read subdirectories here, then let's leave a comment saying this only reads one layer deep and why we chose that.

lib/jekyll/readers/theme_data_reader.rb
```diff
@@ -0,0 +1,31 @@
...
+      end
+
+      entries.each do |entry|
+        path = @site.in_theme_dir(dir, entry)
```
parkr, line 23: This is the main difference between the superclass and this class, so maybe we can use `@source_dir` here instead, or use a `@source_dir_proc` that allows us to pass the method `in_source_dir()` or `in_theme_dir()` around.

lib/jekyll/reader.rb
```diff
@@ -19,6 +19,7 @@ def read
       @site.data = DataReader.new(site).read(site.config["data_dir"])
       CollectionReader.new(site).read
       ThemeAssetsReader.new(site).read
+      ThemeDataReader.new(site).read(site.config["data_dir"])
```
parkr, line 22: Themes should definitely only read from `_data` so they're standardized (we do this for all other directories I believe). That way if a user customizes their data dir, then they don't break their usage of the theme. Let's do `read(site.theme&.data_path)` here. Ideally we just use `DataReader` here too.

parkr (commented):
I think we should add this feature. I'd love to be able to reduce the differences between the ThemeDataReader and the Jekyll::DataReader classes so we can reuse the logic in the latter. 

Tests are super thin for the DataReader class, so it would be great to add some more tests for that class using our fixture theme with some data files. 😄 

Do you need help with any of this?

lib/jekyll/reader.rb
mgerzabek: > I think we should add this feature. I'd love to be able to reduce the differences between the ThemeDataReader and the Jekyll::DataReader classes so we can reuse the logic in the latter.

Then the main concern of @ashmaroli wouldn’t be addressed, since the DataReader recursively reads everything within `_data`. But since this holds also true for `_layouts` or `_includes` from end user POV I suggest it’s the desired behaviour.

To use the `DataReader` instead of the clone the merging of the data must happen within the `Reader`. That would mean a new method should do the handling of reading and merging the data.

Does this sound okay?

lib/jekyll/reader.rb
```diff
@@ -16,9 +16,19 @@ def read
...
+    end
+
+    # Read and merge the contents of _data folders. Data from Site takes
+    # precedence over data from theme.
+    #
+    # Returns nothing.
+    def read_data
+      @site.data = DataReader.new(site).read(site.config["data_dir"])
+      theme_data = DataReader.new(site).read(site.theme&.data_path)
+      Jekyll::Utils.deep_merge_hashes(theme_data, @site.data)
```
ashmaroli, line 31: @mgerzabek I doubt this works like you expect it to..
`@site.data` isn't being mutated.

You should either reassign `@site.data` or use `Utils.deep_merge_hashes!`

I don't see tests.
You can modify the local `test-theme` for this and add tests (similar to ones existing for layouts, includes and assets in a theme).

mgerzabek: Okay, @ashmaroli sorry for the noise… assigned the merged hashes now in read method.

> I don't see tests.

I’m (not so) new to Ruby (anymore), but have no idea on how to write tests. If you could point me to a good primer to testing in Ruby especially in context of Jekyll, I’d be very thankful. I’ll look into testing over the weekend.

ashmaroli: @mgerzabek This will need two types of tests.
But first update our [test-theme](https://github.com/jekyll/jekyll/tree/master/test/fixtures/test-theme).
Then try reading through this [unit test file](https://github.com/jekyll/jekyll/blob/master/test/test_theme_assets_reader.rb) and extrapolate for data files in theme.
Finally, add scenarios to this [Cucumber feature](https://github.com/jekyll/jekyll/blob/master/features/theme.feature) (more important than adding a unit test for this pull request).

lib/jekyll/reader.rb
```diff
@@ -16,9 +16,19 @@ def read
...
+    end
+
+    # Read and merge the contents of _data folders. Data from Site takes
+    # precedence over data from theme.
+    #
+    # Returns merged data hash.
+    def read_data
+      site_data = DataReader.new(site).read(site.config["data_dir"])
+      theme_data = DataReader.new(site).read(site.theme&.data_path)
+      Jekyll::Utils.deep_merge_hashes(theme_data, site_data)
```
parkr, line 31: With our change to setting `@site.data` here, let's make the theme a bit more conditional. I'll read directly into `@site.data`, then guard the remainder of the method with a check to see if the theme exists.

```suggestion
      @site.data = DataReader.new(site).read(site.config["data_dir"])
      return unless site.theme&.data_path
      
      theme_data = DataReader.new(site, in_source_dir: site.method(:in_theme_dir)).read(site.theme&.data_path)
      @site.data = Jekyll::Utils.deep_merge_hashes(theme_data, @site.data)
```

This will require some changes to the `DataReader` class as well.

```diff
diff --git lib/jekyll/readers/data_reader.rb lib/jekyll/readers/data_reader.rb
index 6fbc047fa..8e81254aa 100644
--- lib/jekyll/readers/data_reader.rb
+++ lib/jekyll/readers/data_reader.rb
@@ -4,11 +4,12 @@ module Jekyll
   class DataReader
     attr_reader :site, :content
 
-    def initialize(site)
+    def initialize(site, in_source_dir: nil)
       @site = site
       @content = {}
       @entry_filter = EntryFilter.new(site)
-      @source_dir = site.in_source_dir("/")
+      @in_source_dir = in_source_dir || @site.method(:in_source_dir)
+      @source_dir = in_source_dir.call("/")
     end
 
     # Read all the files in <dir> and adds them to @content
@@ -18,7 +19,7 @@ module Jekyll
     # Returns @content, a Hash of the .yaml, .yml,
     # .json, and .csv files in the base directory
     def read(dir)
-      base = site.in_source_dir(dir)
+      base = @in_source_dir.call(dir)
       read_data_to(base, @content)
       @content
     end
@@ -38,7 +39,7 @@ module Jekyll
       end
 
       entries.each do |entry|
-        path = @site.in_source_dir(dir, entry)
+        path = @in_source_dir(dir, entry)
         next if @entry_filter.symlink?(path)
 
         if File.directory?(path)
```

parkr (commented):
Thanks for your patience on this one! I left fairly detailed comments with changes I'd propose. Let me know if you have any questions! Outstanding items:

1. Write tests
    1. Site without a theme
    2. Site with a theme without _data
    3. Site with a theme with empty _data directory
    4. Site with a theme with data at root of _data, and in a subdirectory
2. Write docs for our website (these can be done in a separate PR, e.g. https://jekyllrb.com/docs/themes/#understanding-gem-based-themes needs to be updated)

mgerzabek:
> I left fairly detailed comments with changes I'd propose.

Thank you @parkr for the Jekyll/Ruby coaching. This was very insightful to me. I love the ambiguation at method level for initializing the reader. This is far advanced stuff I never would think about. Also the guarding of blocks is a rather new concept to me that combined with an unless seems very helpful on language level. Thank you!

As a next step I’ll prepare the tests.

features/theme.feature
```diff
@@ -41,6 +41,19 @@ Feature: Writing themes
...
+    And I have an _data directory
+    And I have an "_data/greetings.yml" file with content:
+      """
+      foo: "Hello! I’m foo. And who are you?"
+      """
+    And I have an "index.html" page that contains "{{ site.data.greetings.foo }}"
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "Hello! I’m foo. And who are you?" in "_site/index.html"
```
ashmaroli, line 55: This test is unsatisfactory.
You have the same data file with same content at source and in the theme.
How are you asserting the case that the result you've got is from the file in theme and not from the file at source.

Additionally, you should prove that theme data has been overridden by adding a negative assertion:
```gherkin
And I should not see ...
```

datapolitical: Where are you seeing he has the same file at source? The point is that he's testing whether data from the theme is picked up.￼

Perhaps you could be a little more clear.￼￼

mgerzabek: Sorry folks, I’m currently working on provding some tests. Not quite there.

And to get an idea on how this works I just started with the scenario as suggested by @ashmaroli. So, yes, the current addition only tests if the data from the theme was read. But as I couldn’t find how to provide those scenarios that @parkr suggested within the `theme.feature` I‘m actually reading through the `TestThemeAssetsReader` to understand how I can create those tests...

ashmaroli: @mgerzabek I'm getting mixed signals here.
You're referring to *scenarios*, *features* and `TestThemeAssetsReader` in the same breath. The first two, and the latter belong to two different types of testing interfaces.
*Scenarios*, *features*, etc are terminologies related to Cucumber tests used for *behavior-driven-development*.
`TestThemeAssetsReader` is part of unit-test suite used for *test-driven-development*.

Which aspect are you planning on tackling first..?

mgerzabek: Thank you @ashmaroli. As I said, I’m completely new to testing with Ruby so don’t know anything. Hence my pledge for some links on the how to and my relatively confused first steps…

As you’ve seen I was able to add a _scenario_ to `theme.feature`. I ran the tests on command line with `./script/test` and got a result that looked okay to me… nothing broken, right? But I’m completely overwhelmed on getting data into these kind of tests other than stating them in sentences… so any hints warmly welcomed!

`TestThemeAssetsReader` is a unit test, a concept that I’m rather familiar with from Java and PHP. I have no clue about setting up new test so I’m just reading the other tests, which is helpful. ATM I have a sceleton which serves as kind of a working plan for the missing pieces.

```ruby
# frozen_string_literal: true

require "helper"

class TestThemeDataReader < JekyllUnitTest

  context "site without a theme" do
    setup do
      @site = fixture_site("theme" => nil)
      @reader = Reader.new(@site)
      @reader.read_data
      assert @site.data.greetings
    end

    should "should read data from source" do

    end
  end

  context "site with a theme without _data" do
    should "should read data from source" do
      site = fixture_site("theme" => "test-theme-skinny")
    end
  end

  context "site with a theme with empty _data directory" do
    should "should read data from source" do
      site = fixture_site("theme" => "test-theme")
    end
  end

  context "site with a theme with data at root of _data" do
    should "should sophistical merge keys" do
      site = fixture_site("theme" => "test-theme")
    end
  end

  context "site with a theme with data at root of _data and in a subdirectory" do
    should "should sophistical merge keys" do
      site = fixture_site("theme" => "test-theme")
    end
  end
end

```

But I wonder how I would setup a **site** where I could overwrite the data from the **theme**. I found a `test/source` directory at the same level as `test/fixture`. This folder contains a `_data` directory. Can I put data files for unit tests there?

Again, any hints warmly welcomed…

ashmaroli:
@mgerzabek Like I have stated in the [comment](https://github.com/jekyll/jekyll/pull/8815#discussion_r743615797) above,
*cucumber tests* are more important for this pull request. Unit tests being secondary.

Since you're lost, I'll guide you towards finishing the Cucumber tests first..

You have already started with a new scenario in `theme.feature`, Great!
To test that you need to run `script/cucumber` instead of `script/test`.

- So, go ahead and and first run `script/cucumber features/theme.feature:44`. It will pass. (*false-positive, though..*)
- Now make a change to the string assigned to **`foo:` in the data file within the `test-theme` fixture**.
- Then rerun `script/cucumber features/theme.feature:44`. Does it fail? Or pass? If your "assumptions" outlined in the scenario earlier today are solid, then the test should've failed..

mgerzabek:
Thank you @ashmaroli. Now I know how start the cucumber tests… and in the meantime I scanned through the introduction on cucumber on cucumber.io.

If I understand correctly anything that I put into the scenario, like the data file in line 4 will be in the **site**. So with the `greetings.yml` already in `test/fixtures/test-theme/_data` and the site.data.freetings in line 4 the test has to pass in any case, since the contents from the **site** overwrite the contents from the **theme** by definition.

```cucumber
  Scenario: A theme with data
    Given I have a configuration file with "theme" set to "test-theme"
    And I have an _data directory
    And I have an "_data/greetings.yml" file with content:
      """
      foo: "Hello! I’m foo. And who are you?"
      """
    And I have an "index.html" page that contains "{{ site.data.greetings.foo }}"
    When I run jekyll build
    And I should see "Hello! I’m foo. And who are you?" in "_site/index.html"
```

Spoken the other way around when I change the key `foo` to something different, the test will also succeed. And that is the desired behaviour, isn’t it?

ashmaroli:
> anything that I put into the scenario, like the data file in line 4 will be in the site.

Yes, that's correct!
Specifically, within `site.source`.

> And that is the desired behaviour, isn’t it?

Yes. It is but you would have got the same result if your theme-data implementation was faulty as well. In other words the test in current state is unreliable. (If you have the time, see if the test fails if you don't mutate `@site.data` like it is being mutated now.)
Hence my [earlier comment from today](https://github.com/jekyll/jekyll/pull/8815#discussion_r746573732)
To fix, first assert the expected with different data..

ashmaroli:
The latest change to the test is in the right direction. Proceed with  the scenarios suggested by @parkr.

mgerzabek:
Gotcha!

```cucumber

  Scenario: A theme with data overridden by site data
    Given I have a configuration file with "theme" set to "test-theme"
    And I have an _data directory
    And I have an "_data/greetings.yml" file with content:
      """
      foo: "Hello! I’m foo. And who are you?"
      """
    And I have an "index.html" page that contains "{{ site.data.greetings.foo }}"
    When I run jekyll build
    And I should see "Hello! I’m foo. And who are you?" in "_site/index.html"
    And I should not see "Hello! I’m bar. What’s up so far?" in "_site/index.html"
```
So should I now provide more _scenarios_ or prepare a unit test where I already prepared a skeleton?

ashmaroli:
> So should I now provide more _scenarios_ or prepare a unit test where I already prepared a skeleton?

If you can come up with more scenarios, all the more welcome!

Your unit test skeleton is fine. Just needs some assertions and refutes within those contexts.
Unit tests won't be elaborate. So you needn't test both initial data and overridden data in the same context. Yes, you may add data files to the `_data` directory within `test/source/`..

mgerzabek:
> Your unit test skeleton is fine.

Well, then… with the last commit I added unit tests and the corresponding data for all scenarios requested by @parkr. Now the integration of `_data` folder for **themes** should be fine.

Thank you @ashmaroli and @parkr for your patience and the many helpful hints to get this feature into _Jekyll_!

Finally I added the proposed documentation change like pointed out by @parkr. Any feedback welcomed. Keep in mind I’m not native in English so many formulation errors are forseen.

features/theme.feature
```diff
@@ -41,6 +41,28 @@ Feature: Writing themes
     And I should see "I'm in the project." in "_site/index.html"
     And I should see "<span class=\"sample\">include.html from test-theme</span>" in "_site/index.html"
 
+  Scenario: A theme without data
+    Given I have a configuration file with "theme" set to "test-theme"
```
ashmaroli, line 45: This is not correct.
`test-theme` has data files within it..

features/theme.feature
```diff
@@ -41,6 +41,28 @@ Feature: Writing themes
...
+    And I have an "index.html" page that contains "{{ site.data.greetings.foo }}"
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "Hello! I’m bar. What’s up so far?" in "_site/index.html"
+
+  Scenario: A theme with data overridden by site data
```
ashmaroli, line 52: Prefer `source` for clarity:
```suggestion
  Scenario: A theme with data overridden by data in source directory
```

features/theme.feature
```diff
@@ -41,6 +41,28 @@ Feature: Writing themes
...
+    Given I have a configuration file with "theme" set to "test-theme"
+    And I have an _data directory
+    And I have an "_data/greetings.yml" file with content:
```
ashmaroli, line 55: typo.
```suggestion
    And I have a "_data/greetings.yml" file with content:
```

lib/jekyll/readers/data_reader.rb
```diff
@@ -4,11 +4,12 @@ module Jekyll
...
     attr_reader :site, :content
 
-    def initialize(site)
+    def initialize(site, in_source_dir: nil)
       @site = site
       @content = {}
       @entry_filter = EntryFilter.new(site)
-      @source_dir = site.in_source_dir("/")
+      @in_source_dir = in_source_dir || @site.method(:in_source_dir)
+      @source_dir = @in_source_dir.call("/") if @in_source_dir
```
ashmaroli, line 12: Where is this variable `@source_dir` used?

test/test_theme_data_reader.rb
```diff
@@ -0,0 +1,87 @@
...
+
+require "helper"
+
+class TestThemeDataReader < JekyllUnitTest
+
+  context "site without a theme" do
+    setup do
+      @site = fixture_site("theme" => nil)
+      @reader = Reader.new(@site)
+      @reader.read_data
```
ashmaroli, line 11: `@reader` isn't used subsequently.
Therefore, call the reader directly:
```suggestion
      @site.reader.read_data
```

test/test_theme_data_reader.rb
```diff
@@ -0,0 +1,87 @@
...
+      assert_equal "Hello! I’m foo. And who are you?", @site.data["greetings"]["foo"]
+      assert_equal "Dairy", @site.data["categories"]["dairy"]["name"]
+    end
+  end
+
+  context "site with a theme without _data" do
+    setup do
+      @site = fixture_site("theme" => "test-theme-skinny")
+      @reader = Reader.new(@site)
+      @reader.read_data
```
ashmaroli, line 26: Same as above..

test/test_theme_data_reader.rb
```diff
@@ -0,0 +1,87 @@
...
+      assert_equal "Hello! I’m foo. And who are you?", @site.data["greetings"]["foo"]
+      assert_equal "Dairy", @site.data["categories"]["dairy"]["name"]
+    end
+  end
+
+  context "site with a theme with empty _data directory" do
+    setup do
+      @site = fixture_site("theme" => "test-theme-w-empty-data")
+      @reader = Reader.new(@site)
+      @reader.read_data
```
ashmaroli, line 41: same as above.

test/test_theme_data_reader.rb
```diff
@@ -0,0 +1,87 @@
...
+      assert @site.data["greetings"]
+      assert @site.data["categories"]["dairy"]
+      assert @site.data["cars"]
+    end
+
+    should "should sophistical merge keys" do
+      assert_equal "Hello! I’m foo. And who are you?", @site.data["greetings"]["foo"]
+      refute_equal "Hello! I’m bar. What’s up so far?", @site.data["greetings"]["foo"]
```
ashmaroli, line 64: `refute`-ing earlier has more impact..
```suggestion
      refute_equal "Hello! I’m bar. What’s up so far?", @site.data["greetings"]["foo"]
      assert_equal "Hello! I’m foo. And who are you?", @site.data["greetings"]["foo"]
```

test/test_theme_data_reader.rb
```diff
@@ -0,0 +1,87 @@
...
+    setup do
+      @site = fixture_site("theme" => "test-theme")
+      @reader = Reader.new(@site)
+      @reader.read_data
+      assert @site.data["greetings"]
+      assert @site.data["categories"]["dairy"]
+      assert @site.data["cars"]
+    end
+
+    should "should sophistical merge keys" do
```
ashmaroli, line 62: I don't see what's sophisticated about `["greetings"]["foo"]`..
Please rephrase.
```suggestion
    should "should merge nested keys" do
```

test/test_theme_data_reader.rb
```diff
@@ -0,0 +1,87 @@
...
+      assert_equal "Hello! I’m foo. And who are you?", @site.data["greetings"]["foo"]
+      refute_equal "Hello! I’m bar. What’s up so far?", @site.data["greetings"]["foo"]
+      assert_equal "Mercedes", @site.data["cars"]["manufacturer"]
+    end
+  end
+
+  context "site with a theme with data at root of _data and in a subdirectory" do
```
ashmaroli, line 69: This is not clear. What is your intention?

test/test_theme_data_reader.rb
```diff
@@ -0,0 +1,87 @@
...
+    setup do
+      @site = fixture_site("theme" => "test-theme")
+      @reader = Reader.new(@site)
+      @reader.read_data
+      assert @site.data["greetings"]
+      assert @site.data["categories"]["dairy"]
+      assert @site.data["cars"]
+    end
+
+    should "should sophistical merge keys" do
```
ashmaroli, line 79:
```suggestion
    should "should merge nested keys" do
```

ashmaroli (dismissed):

lib/jekyll/readers/data_reader.rb
mgerzabek: In line 58 where a debug statement states the location of the read data file. Not sure if we can drop this.

test/test_theme_data_reader.rb
mgerzabek: Dairy in source/_data is
```yml
name: Dairy
products:
- name: cheese
  price: 5.3
- name: milk
  price: 2.5
```
Dairy in **theme** is
```yml
name: Cheese Dairy
products:
- name: spread cheese
  price: 1.2
- name: cheddar cheese
  price: 4.5
```
The test looks for the union of products under the `name` of _Dairy_.

lib/jekyll/readers/data_reader.rb
ashmaroli: I see..
The variable is only used to get a *relative path* of the data file.
That said, we can't remove it or modify it to maintain compatibility with any plugins subclassing the DataReader.

So, best leave it as is in current state.

features/theme.feature
mgerzabek: Is it also `And I have a _data directory`? in line 54?

lib/jekyll/readers/data_reader.rb
mgerzabek: Okay, then we’re done here…

test/test_theme_data_reader.rb
```diff
@@ -0,0 +1,82 @@
...
+      assert @site.data["greetings"]
+      assert @site.data["categories"]["dairy"]
+      assert @site.data["cars"]
+    end
+
+    should "should merge nested keys" do
+      refute_equal "Cheese Dairy", @site.data["categories"]["dairy"]["name"]
+      @site.data["categories"]["dairy"]["products"].each do |product|
+        assert_includes "|spread cheese|,|cheddar cheese|,|cheese|,|milk|", "|#{product["name"]}|"
+      end
```
ashmaroli, line 78: I don't think this expectation is correct.
Could you update the test to be clearer..?

For example,

```suggestion
      expected_names = ["spread cheese", "cheddar cheese", "cheese", "milk"]
      product_names  = @site.data["categories"]["dairy"]["products"].map { |product| product["name"] }
      expected_names.each do |expected_name|
        assert_includes product_names, expected_name
      end
```

ashmaroli:
Okay @mgerzabek, now that you are comfortable with adding tests, I want you to explicitly test the following scenario using either Cucumber or Minitest. The scenario is made up of the following key points:
- A theme that supports rendering social media icons based on `site.data`
- The theme has a data file consisting of key-value pairs for each supported social medium. For example, 
    ```yml
    github: octocat
    twitter: jekyllrb
    facebook: GitHub
    ```
- The icons are rendered using an include file named `social.html`:
    ```
    <ul class="social-media">
      {% assign social = site.data.social %}
      {% if social.github %}<li>{{ social.github }}</li>{% endif %}
      {% if social.twitter %}<li>{{ social.twitter }}</li>{% endif %}
      {% if social.facebook %}<li>{{ social.facebook }}</li>{% endif %}
    </ul>
    ```
- An end-user of the theme desires to render only one icon: for his GitHub Account.
   Therefore, **assert** that Jekyll renders the GitHub icon; and **refute** rendering of twitter and facebook icons.

test/test_theme_data_reader.rb
ashmaroli: This is still not resolved. The title is confusing and doesn't convey your intention.

mgerzabek:
> I want you to explicitly test the following scenario using either Cucumber or Minitest

@ashmaroli I don’t see this test as a typical use case for the proposed addition. I don’t want a second way of providing _configuration_ data, but to add something that’s IMHO needed. Hence my sample in the docs with the `i18n` files.

Your request is better covered by some switches in `_config.yml`. That’s the way [minima](https://github.com/jekyll/minima/blob/master/_config.yml) is doing it and I don’t see a benefit in mirroring this behaviour with data files. I see it as anti-pattern.

**But you’re right**, there needs to be a test that covers the assertions I made in the docs. So this is what I added as unit and cucumber tests… just to get a little more into both of them ;-)

HTH

ashmaroli:
> That’s the way [minima](https://github.com/jekyll/minima/blob/master/_config.yml) is doing it and I don’t see a benefit in mirroring this behaviour with data files. I see it as anti-pattern.

I know. That's why I asked for it to be explicitly tested (serves as an ineffective form of documentation as well) because there's nothing stopping a theme-author from following the *anti-pattern*.

Regarding the docs, I'll take a look during the final round of reviews. For now,:
- New features won't be introduced in patch versions. So, this feature will only be in `v4.3.0`. Not `v4.2.x`.
- Break long lines hard (I prefer at 120 chars, but not strict)
- Explicitly document the antipattern to discourage the emergence of issues that I have been concerned about from the very beginning of this PR.

mgerzabek:
> I know. That's why I asked for it to be explicitly tested (serves as an ineffective form of documentation as well) because there's nothing stopping a theme-author from following the _anti-pattern_.

I suspected you drove me down this lane to exactly get clear about it, thanks… **gg**

> Explicitly document the antipattern to discourage the emergence of issues that I have been concerned about from the very beginning of this PR.

Is the note block in line 168 enough?

Thanks again for helping out @ashmaroli

ashmaroli:
> Is the note block in line 168 enough?

Ah..! I didn't see that earlier. Nice!
Please add a line to the note saying that bundling behavior-modifying theme-data is not only **an antipattern** whose use is discouraged, but it is solely upto the theme-author to ensure that every provided data can be easily overridden by the user if they desire to.

Thank you for being patient with me, @mgerzabek. Its been a roller-coaster ride. However, I believe that we have thought of most possible use-cases here.
There are a couple of style issues with the Ruby code that can be corrected by running `script/fmt -a`.

Additionally, https://github.com/jekyll/jekyll/blob/c90d083ef527df817b4be9ae79cb577865c97659/lib/jekyll/theme_builder.rb#L5
needs to be updated as well

Also, do check #7554 for possible conflicts depending on whether that gets approved and merged before this..

mgerzabek:
Okay @ashmaroli thanks for the ride :grin:

I corrected the style issues. This is one thing I learned to love about Ruby. You run the checker and mostly everything gets corrected out of the box. Magic!

I added your proposed sentence to the docs, cross-checked the child-theme PR, which IMHO goes far down another rabbit whole and so I think, we’re done.

ashmaroli:
@mgerzabek You have deleted `test/source/_includes/tmp` in the commit https://github.com/jekyll/jekyll/pull/8815/commits/1e97b9f66b19c47c6ebd3ae438c3f090e8ec6d61
Please undelete that test file (Its a symlink needed for testing our `{% include ... %}` tag)

mgerzabek:
> Please undelete that test file (Its a symlink needed for testing our `{% include ... %}` tag)

Okay, done. But counter intuitive to your request ist the last line in `.gitignore` in project root. It excludes all files within temporary directore (line 28: `tmp/*`).

When I delete the file, then run the testsuite the file is created anyway. Not sure about that.

mgerzabek:
> Please undelete that test file (Its a symlink needed for testing our `{% include ... %}` tag)

Arrgh, too fast. Sorry!

Where does this symlink point to? Do I have to recreate the symlink via terminal or is there a way I could use github to recreate it? I’m not sure how to proceed.

ashmaroli:
> Where does this symlink point to?

@mgerzabek The commit https://github.com/jekyll/jekyll/commit/a2b6aa263ecff76247ca039f66430f3bd52dd791 and the associated PR has all the info to answer your questions. Give it a try. Don't waste your time though. I'll have a go at it before the final review.

mgerzabek:
>  I'll have a go at it before the final review.

@ashmaroli as I don’t get the point, thank you for jumping in!

ashmaroli:
> > I'll have a go at it before the final review.
> 
> as I don’t get the point...

The point being you were struggling and sometimes, it's easier just doing it instead of giving instructions that the other party can easily follow..
You may sync your local branch by running:
```
git pull origin theme_w_data_folder
```

mgerzabek:
> it's easier just doing it instead of giving instructions that the other party can easily follow..

Thank you!

ashmaroli:
> So, is there anything left to do?

I'm not so pleased with the documentation. However, I'll defer sorting that and anything remaining to @parkr

mgerzabek:
> I'm not so pleased with the documentation.

Well, @ashmaroli if you can tell me what should be improved I‘d jump in… _not so pleased_ is kind of vague though.

Anyways, @parkr [said](https://github.com/jekyll/jekyll/pull/8815#pullrequestreview-799413078), the docs could be addressed in another PR as well… so why not proceed from here?

ashmaroli:
> if you can tell me what should be improved I‘d jump in…

You did tell us that you're not a native English speaker and hence the docs may not be up to expected level. It so happens that I'm not a native English speaker as well. So, my suggestions to improve the docs may be shot down by a consequent review.

Let us wait for @parkr's inputs. If he doesn't get the time by Tuesday, I'll ping another maintainer.

In the meantime, you could make the following edits (not mandatory, but highly recommended):
- Reduce line-lengths to 120 chars.
- Rephrase everything to third-person voice. (i.e `your theme` to `a theme`, `she`/`he` to `they`, *et cetra..*) 
- Use regular punctuation. (i.e. `it’s` to `it's`), `themes` to `theme's`, *et cetra..*

parkr (approved):

parkr:
This looks great! I think further improvements could be made in a follow-up PR as needed as long as CI returns green.

ashmaroli:
Thanks @mgerzabek 
@jekyllbot: merge +minor

mgerzabek:
Thanks @ashmaroli!

parkr:
Well done @mgerzabek and thanks @ashmaroli for the review!

PR #8821, opened 2021-09-23
Remove the warning about auto-regeneration on Windows

### LINKED ISSUE #8820: feat: Remove (or allow suppression of) WSL warning
## Summary

When running `jekyll serve -w` on Linux in WSL, I see this warning:

```
Auto-regeneration may not work on some Windows versions.
Please see: https://github.com/Microsoft/BashOnWindows/issues/216
If it does not work, please upgrade Bash on Windows or run Jekyll with --no-watch.
```

Looking at the [source](https://github.com/jekyll/jekyll/blob/6855200ebda6c0e33f487da69e4e02ec3d8286b7/lib/jekyll/commands/build.rb#L77-L85), it appears there's no way for me to suppress this message.

## Motivation

This message is based on a bug that was fixed in 2016. I'm not sure whether there are any users it would still be relevant for, assuming people are taking advantage of their free Windows 10 upgrades (and let's all hope they are, because of security issues). The versions of Windows 10 where this bug would apply are no longer supported by Microsoft.

Some options include:

1. Simply remove the message. The likelihood of it being relevant to anybody is vanishingly small.
2. Attempt to detect the versions of WSL for which this is applicable and only show the message then.
3. Provide us a way to opt out of the message (via an environment variable, ideally)

## Drawbacks

- For option 1, I don't have the stats handy to know how many jekyll users are using an old enough version of WSL to be affected by this bug. I'm using a gut feel here to say it's likely next to none.
- For option 2, I haven't done the research to determine if you could easily detect this situation. The referenced bug would imply that simply looking for the existence of `/proc/sys/fs/inotify/` would likely be good enough. It may also be difficult or impossible to get an old enough version of Windows 10 to verify this fix.
- For option 3, although I would personally hunt this env var down and set it (😁) it would still be pretty hidden, unless you also mention the env var in the warning message.

Obviously option 1 is the simplest the implement. Option 2 feels like the next best option, but the most work (because of verification, not coding). Option 3 is a fairly simple and safe workaround that only requires a bit of code and doesn't depend on verification.

Doing nothing is the safest option, but the saddest one for me personally. 😉

mattr-:

## Summary

This removes the warning about auto regeneration on Windows potentially having problems. According to the info in #8820, the bug we're calling out is four years old and I don't think that it's in our best interest to continue support installations using WSL that are running software that old. I'd rather remove this warning and handle a few extra questions from folks in these scenarios and encourage them to upgrade instead.

## Context

Fixes #8820
Potential 4.2 backport candidate?

### DIFF
--- a/lib/jekyll/commands/build.rb
+++ b/lib/jekyll/commands/build.rb
@@ -73,17 +73,6 @@ module Jekyll
         #
         # Returns nothing.
         def watch(site, options)
-          # Warn Windows users that they might need to upgrade.
-          if Utils::Platforms.bash_on_windows?
-            Jekyll.logger.warn "",
-                               "Auto-regeneration may not work on some Windows versions."
-            Jekyll.logger.warn "",
-                               "Please see: https://github.com/Microsoft/BashOnWindows/issues/216"
-            Jekyll.logger.warn "",
-                               "If it does not work, please upgrade Bash on Windows or "\
-                               "run Jekyll with --no-watch."
-          end
-
           External.require_with_graceful_fail "jekyll-watch"
           Jekyll::Watcher.watch(options, site)
         end


ashmaroli:
AFAICT aren't only bug-fixes, security-fixes and removal of regressions backported in Jekyll?
Since this doesn't fall into any of those categories, I don't think this qualifies for a backport..
*(Besides, you consider this to be an enhancement as well, ergo fit for minor-version bump.)*

I don't see a need for documentation or tests on this one.
Shall I consider it complete and ready for review?

mattr-:
> Shall I consider it complete and ready for review?

yes please 😄

ashmaroli (approved):

ashmaroli:
Thanks @mattr- :smiley: 
@jekyllbot: merge +minor

PR #8822, opened 2021-09-24
Update troubleshooting.md

fanatiicx:

Hello, I added a couple of steps I took to solve a "timeout error" I experienced during installation.

Please let me know if I should create an issue and add the steps there instead of directly here in future (first time contributing to the docs here). Thank you!

### DIFF
--- a/docs/_docs/troubleshooting.md
+++ b/docs/_docs/troubleshooting.md
@@ -206,6 +206,14 @@ you don't have a proper JavaScript runtime. To solve this, either install
 `execjs` and `therubyracer` gems, or install `nodejs`. Check out
 [issue #2327](https://github.com/jekyll/jekyll/issues/2327) for more info.
 
+### Timeout Error
+
+If you get a timeout error after `bundle install`, check [Is It Up](http://isitup.org/rubygems.org) to see that the RubyGems website is up.
+
+If it is, open the Gemfile in your site's folder and change the first line from `https://www.rubygems.org` to `http://www.rubygems.org`
+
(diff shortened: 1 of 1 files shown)


docs/_docs/troubleshooting.md
```diff
@@ -206,6 +206,14 @@ you don't have a proper JavaScript runtime. To solve this, either install
 `execjs` and `therubyracer` gems, or install `nodejs`. Check out
 [issue #2327](https://github.com/jekyll/jekyll/issues/2327) for more info.
 
+### Timeout Error
+
+If you get a timeout error after `bundle install`, check [Is It Up](http://isitup.org/rubygems.org) to see that the RubyGems website is up.
+
+If it is, open the Gemfile in your site's folder and change the first line from `https://www.rubygems.org` to `http://www.rubygems.org`
```
mattr-, line 213: I don't think recommending folks switch away from `https` to `http` is a good thing to do here. Would you be willing to remove this line?

mattr- (commented):
Do you experience this timeout error consistently? I'm hesitant to include this in the documentation as it's not really a problem with Jekyll but with pieces outside our control. I'd like some more time to think about this and hope you'll consider the suggestion I've left in the mean time.

docs/_docs/troubleshooting.md
fanatiicx: @mattr- No problem, I understand that it was probably more of a network than Jekyll issue! I experienced the timeout error consistently and had to do a bit of Googling to see how to complete the installation.

Issue #8826, opened 2021-09-29
Jekyll 4.2.1 gem has files with CRLF line endings, earlier had LF


oturpe:
## My Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Operating System | Fedora Rawhide           |
| `jekyll`         | 4.2.1     |
| `github-pages`   | No     |

---

## Expected Behaviour

I am the maintainer for the Jekyll package in Fedora.
Earlier Jekyll gems contained files with LR line endings.
The build script for creating the Fedora package
worked correctly with this setup.
My expectation when starting packaging 4.2.1
was that the line endings would still be LF.
 
## Current Behavior
In 4.2.1 gem, files use CRLF linebreaks.
This broke the build script in various places:
* Patches applied with the `rpm` `%patch` facility fail
with error "different line endings"
* Running `exe/jekyll --help` (to generate some docs) fails
since the shebang is interpreted incorrectly
due to CR being there:
`/usr/bin/env: ‘ruby\r’: No such file or directory`

The script can be made to succeed
by converting line endings as needed.
Still, I am afraid this change can also have
other unexpected consequences.
The line endings are a brittle topic,
switching them in any direction
can cause all sorts of trouble.

I am not aware of any specific reason
to use LF instead of CRLF in gems.
Still, since there was I change,
I would first like to confirm that this was intentional
before adjusting the Fedora build script to handle this.

## Code Sample

Snippet 1: Failing to apply a patch:
```
$ fedpkg mockbuild
...lots of output...
+ echo 'Patch #0 (0000-jekyll-commands-remove-bundle-install-step-for-new.patch):'
  Patch #0 (0000-jekyll-commands-remove-bundle-install-step-for-new.patch):
+ /usr/bin/patch --no-backup-if-mismatch -f -p1 --fuzz=0
patching file lib/jekyll/commands/new.rb
Hunk #1 FAILED at 13 (different line endings).
Hunk #2 FAILED at 132 (different line endings).
2 out of 2 hunks FAILED -- saving rejects to file lib/jekyll/commands/new.rb.rej
patching file test/test_new_command.rb
error: Bad exit status from /var/tmp/rpm-tmp.vA9u1a (%prep)
...more lines about terminating the build...
```

Snippet 2: Failing to execute `exe/jekyll`:
```
$ fedpkg mockbuild
...lots of output...
[...]

PR #8830, opened 2021-09-30
Add a workflow to build gems consistently

ashmaroli:
## Summary

Use GH Actions to **build** gem.
(Push to Rubygems.org manually)

With this Ruby version and platform remains consistent when triggered by different maintainers.

## Context

As part of responding to #8826 (which will be closed after `v4.2.2` is shipped and a release post is published).
This will be back ported to `4.2-stable` and `3.9-stable`.


ashmaroli:
Since this workflow will not be triggered for the current context, you may view the run on my fork at https://github.com/ashmaroli/jekyll/actions/runs/1291299525 for analysis.

mattr- (commented):
What do you think about evolving this action so that it can be used to release the gem? We could use the push of the tag to trigger this workflow instead of the commit message, and then set up the actions secrets to push the gem.

We might want to create a jekyllbot user on rubygems so that the authentication is not tied to our accounts as well.

I really like the idea of getting the process of making a release off our local machines. Thanks for getting this started.

ashmaroli:
> What do you think about evolving this action so that it can be used to release the gem?

Our release process for backports is not straight-forward and involves switching branches multiple commits. Automating that is definitely beneficial but I didn't do that in the initial phase to have greater control n err on the safer side.


> We might want to create a jekyllbot user on rubygems so that the authentication is not tied to our accounts as well.

:+1:

parkr:
I setup a jekyllbot RubyGems account. I'm all for automating this, especially if it were as simple as adding a Git tag. I have added a secret called `RUBYGEMS_GEM_PUSH_API_KEY` to the org for use in our gem repos. Right now jekyllbot doesn't have access to anything, but it'd be cool to validate with a gem that's less important (maybe like a Jekyll-import or something?). What do you think?

ashmaroli:
> it'd be cool to validate with a gem that's less important (maybe like a Jekyll-import or something?)...

We could use https://github.com/jekyll/jekyll-test-gem-plugin for this purpose.
Since it is not consumer-oriented and is intuitively named, we could play around with it without being concerned of side-effects.

parkr (approved):
We can run this once we give jekyllbot the appropriate permissions for the gem.

ashmaroli:
@jekyllbot: merge +dev

PR #8844, opened 2021-10-05
Add info to `bundle add webrick` for `livereload`

tordans:
`ruby 3.0.2` did raise an error when jekyll is started with `livereload`. Adding webrick was required to solve it.

See also https://talk.jekyllrb.com/t/load-error-cannot-load-such-file-webrick/5417/6

I think, adding this to https://jekyllrb.com/docs/configuration/options/#serve-command-options is a good place to inform users about this issue. Let me know if there are other places, that should receive this note.

### DIFF
--- a/docs/_data/config_options/serve.yml
+++ b/docs/_data/config_options/serve.yml
@@ -11,7 +11,7 @@
 
 
 - name: Live Reload
-  description: Reload a page automatically on the browser when its content is edited.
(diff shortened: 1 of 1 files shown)


ashmaroli:
We have already added `webrick` as a dependency for the next minor version release.
https://github.com/jekyll/jekyll/blob/8af22d0852be034d1b75ce9194ed24e4900ace33/jekyll.gemspec#L49

Additionally, there's a note regarding `webrick` at https://jekyllrb.com/docs/#instructions

PR #8858, opened 2021-10-29
Allow Configurable Converters on CSV

MichaelCordingley:

## Summary

Lets the user specify which converters to pass to `CSV`, defaulting to `nil` if not present in config. This is specifically so that numeric values in a CSV file can be parsed as something other than strings.

### DIFF
--- a/docs/_docs/datafiles.md
+++ b/docs/_docs/datafiles.md
@@ -148,3 +148,30 @@ author: dave
 {% endraw %}
 
 For information on how to build robust navigation for your site (especially if you have a documentation website or another type of Jekyll site with a lot of pages to organize), see [Navigation]({{ '/tutorials/navigation/' | relative_url }}).
+
+## CSV/TSV Parse Options
+
+The way that Ruby parses CSV and TSV files can be customized with the `csv_reader` and `tsv_reader`
+configuration options. Each configuration key exposes the same options:
+
+`converters`: What [CSV converters](https://ruby-doc.org/stdlib-2.6.1/libdoc/csv/rdoc/CSV.html#Converters) should be
+              used when parsing the file. This will need to be set with `numeric` if you want number fields in the
+              CSV to be parsed into numeric types rather than into strings. By default, this list is empty.
+`encoding`:   What encoding the files are in. Defaults to the site `encoding` configuration option.
+`headers`:    Boolean field for whether to parse the first line of the file as headers. When `false`, it treats the
+              first row as data. Defaults to `true`.
+
+Examples:
+
+```yaml
+csv_reader:
+    converters:
+      - numeric
+      - datetime
+    headers: true
+    encoding: utf-8
+tsv_reader:
+    converters:
+      - all
+    headers: false
+```
--- a/docs/_docs/datafiles.md
+++ b/docs/_docs/datafiles.md
@@ -151,7 +151,7 @@ For information on how to build robust navigation for your site (especially if y
 
 ## CSV/TSV Parse Options
 
-The way that Ruby parses CSV and TSV files can be customized with the `csv_reader` and `tsv_reader`
+The way Ruby parses CSV and TSV files can be customized with the `csv_reader` and `tsv_reader`
 configuration options. Each configuration key exposes the same options:
 
 `converters`: What [CSV converters](https://ruby-doc.org/stdlib-2.6.1/libdoc/csv/rdoc/CSV.html#Converters) should be
--- a/docs/_docs/datafiles.md
+++ b/docs/_docs/datafiles.md
@@ -155,8 +155,8 @@ The way Ruby parses CSV and TSV files can be customized with the `csv_reader` an
 configuration options. Each configuration key exposes the same options:
 
 `converters`: What [CSV converters](https://ruby-doc.org/stdlib-2.5.0/libdoc/csv/rdoc/CSV.html#Converters) should be
-              used when parsing the file. This will need to be set with `numeric` if you want number fields in the
-              CSV to be parsed into numeric types rather than into strings. By default, this list is empty.
+              used when parsing the file. Available options are `integer`, `float`, `numeric`, `date`, `date_time` and
+              `all`. By default, this list is empty.
 `encoding`:   What encoding the files are in. Defaults to the site `encoding` configuration option.
 `headers`:    Boolean field for whether to parse the first line of the file as headers. When `false`, it treats the
               first row as data. Defaults to `true`.
--- a/docs/_docs/datafiles.md
+++ b/docs/_docs/datafiles.md
@@ -154,7 +154,7 @@ For information on how to build robust navigation for your site (especially if y
 The way Ruby parses CSV and TSV files can be customized with the `csv_reader` and `tsv_reader`
 configuration options. Each configuration key exposes the same options:
 
-`converters`: What [CSV converters](https://ruby-doc.org/stdlib-2.6.1/libdoc/csv/rdoc/CSV.html#Converters) should be
+`converters`: What [CSV converters](https://ruby-doc.org/stdlib-2.5.0/libdoc/csv/rdoc/CSV.html#Converters) should be
               used when parsing the file. This will need to be set with `numeric` if you want number fields in the
               CSV to be parsed into numeric types rather than into strings. By default, this list is empty.
 `encoding`:   What encoding the files are in. Defaults to the site `encoding` configuration option.
--- a/lib/jekyll/readers/data_reader.rb
+++ b/lib/jekyll/readers/data_reader.rb
@@ -59,14 +59,9 @@ module Jekyll
 
       case File.extname(path).downcase
       when ".csv"
-        CSV.read(path,
-                 :headers  => true,
-                 :encoding => site.config["encoding"]).map(&:to_hash)
+        CSV.read(path, **csv_config).map { |row| convert_row(row) }
       when ".tsv"
-        CSV.read(path,
-                 :col_sep  => "\t",
-                 :headers  => true,
-                 :encoding => site.config["encoding"]).map(&:to_hash)
+        CSV.read(path, **tsv_config).map { |row| convert_row(row) }
       else
         SafeYAML.load_file(path)
       end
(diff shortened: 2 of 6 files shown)


MichaelCordingley:
Oh. Wow. Dumb mistake of the day. Hold on a sec.

MichaelCordingley:
@ashmaroli There we go. I also had to move a few things around to make the linter happy.

MichaelCordingley:
I'm going to make a few more edits to make this more extensible.

MichaelCordingley:
@ashmaroli I see some PRs that have been sitting for multiple years here and my own has sat for a month already. Do you know if this project is accepting PRs?

I've attempted to contribute to enough projects where the PRs sit unremarked for years until I finally just close them, only to _then_ receive a message from the maintainers surprised that I closed it.

ashmaroli:
@MichaelCordingley Yes, this project is open to contributions.
Regarding whether a PR gets immediate attention depends on many things, like:
- whether the change is valuable to the user-base in general and not to just niche cases.
- a maintainer is able to alot time and energy to understand, review, and finalize a merge.

MichaelCordingley:
@ashmaroli Awesome. Thanks! And for this PR, I'm wide open to altering my approach. The big thing is just to be able to have numeric values from data files.

MichaelCordingley:
@ashmaroli How long does it typically take for a PR to receive consideration?

ashmaroli:
@MichaelCordingley Apologies for letting this go stale.

Okay, this PR definitely needs some test coverage.

The existing implementation and the proposed implementation *appears* to be going in very different directions (even though it may not be so in reality). Jekyll already has a concept of *converters*. For a maintainer / contributor stumbling across this implementation at a future date is going to end up confused.
Therefore, please include a comment above the private method(s) clarifying what `:converters` refers to here.

These methods are going to be called for every `.csv` / `.tsv` files in a site. Therefore, generating `config` and consecutively calling `config.fetch` for each file is not optimal, especially when the result is going to remain the same irrespective of the CSV / TSV file paths.

MichaelCordingley:
@ashmaroli I'll definitely make sure it's covered with tests, though I may need assistance doing so. I believe I had difficulty getting them to run last time I worked on this.

Before I get to reworking the code, it sounds as if you're saying that there's a conflict in terminology that I should resolve. Other than that, is there anything you'd want changed from an interface perspective?

MichaelCordingley:
@ashmaroli Updated.

lib/jekyll/readers/data_reader.rb
```diff
@@ -76,5 +71,47 @@ def sanitize_filename(name)
...
+    def tsv_config
+      @tsv_config ||= read_config("tsv_reader", { col_sep: "\t" })
+    end
+
+    # @param config_key [String]
+    # @param overrides [Hash]
+    # @return [Hash]
+    # @see https://ruby-doc.org/stdlib-2.6.1/libdoc/csv/rdoc/CSV.html#Converters
+    def read_config(config_key, overrides = {})
+      reader_config = (config[config_key] || {}).transform_keys(&:to_sym)
```
ashmaroli, line 92: I don't see the need to symbolize the keys.
Why can't we just use the string keys?

ashmaroli:
Thanks for making the changes.
There are some style issues you need to resolve, though. (Please check failing test).
I've additionally left a comment regarding the use of symbol keys..

lib/jekyll/readers/data_reader.rb
MichaelCordingley: Updated to use string keys.

I was trying to keep to the Ruby convention of using symbols for hash keys and other identifiers. I'm vaguely annoyed that there's even a distinction in the language, but here we are.

MichaelCordingley:
All fixed up.

test/test_data_reader.rb
```diff
@@ -14,4 +14,47 @@ class TestDataReader < JekyllUnitTest
...
+    should "parse TSV normally" do
+      assert_equal @parsed, @reader.read_data_file(File.expand_path("fixtures/sample.tsv", __dir__))
+    end
+  end
+
+  context "with csv options set" do
+    setup do
+      reader_config = {
+        "csv_converters" => [:numeric],
+        "headers" => false,
```
ashmaroli, line 37:
```suggestion
        "csv_converters" => [:numeric],
        "headers"        => false,
```

lib/jekyll/readers/data_reader.rb
```diff
@@ -76,5 +71,45 @@ def sanitize_filename(name)
...
+
+    # @return [Hash]
+    def tsv_config
+      @tsv_config ||= read_config("tsv_reader", { :col_sep => "\t" })
+    end
+
+    # @param config_key [String]
+    # @param overrides [Hash]
+    # @return [Hash]
+    # @see https://ruby-doc.org/stdlib-2.6.1/libdoc/csv/rdoc/CSV.html#Converters
```
ashmaroli, line 90:
```suggestion
    # @see https://ruby-doc.org/stdlib-2.5.0/libdoc/csv/rdoc/CSV.html#Converters
```

lib/jekyll/readers/data_reader.rb
```diff
@@ -76,5 +71,45 @@ def sanitize_filename(name)
...
+      @config ||= site.config
+    end
+
+    # @param row [Array, CSV::Row]
+    # @return [Array, Hash]
+    def convert_row(row)
+      return row.to_hash if row.instance_of? CSV::Row
+
+      row
+    end
```
ashmaroli, line 113: Let's compact this using a ternary expression.
```suggestion
    def convert_row(row)
      row.instance_of?(CSV::Row) ? row.to_hash : row
    end
```

parkr (approved):
Sounds good. Might be good to include some documentation as well.

MichaelCordingley:
I don't mid documenting, once I know for sure it's going in and in what form.

ashmaroli:
> once I know for sure it's going in and in what form..

@MichaelCordingley Please elaborate.

MichaelCordingley:
Fantastic. Where do I write it? In this PR or against a separate place?

MichaelCordingley:
Looks like it should be somewhere in https://github.com/jekyll/jekyll/tree/master/docs/_docs, so I'll author against there unless I hear otherwise from you.

ashmaroli:
> Where do I write it? In this PR

Yes, to the same branch. Information related to data files in Jekyll is principally accessed at https://jekyllrb.com/docs/datafiles/. *Improve that page* with info related to this PR at an appropriate location.

MichaelCordingley:
Documented.

MichaelCordingley:
Also rebased and force pushed to clean up the history.

docs/_docs/datafiles.md
```diff
@@ -148,3 +148,28 @@ author: dave
...
+## CSV/TSV Parse Options
+
+The way that Ruby parses CSV and TSV files can be customized with the `csv_reader` and `tsv_reader`
+configuration options. Each configuration key exposes the same options:
+
+`converters`: What [CSV converters](https://ruby-doc.org/stdlib-2.6.1/libdoc/csv/rdoc/CSV.html#Converters) should be
+              used when parsing the file. This will need to be set with `numeric` if you want number fields in the
+              CSV to be parsed into numeric types rather than into strings. By default, this list is empty.
+`encoding`:   What encoding the files are in. Defaults to the site `encoding` configuration option.
+`headers`:    Boolean field for whether to parse the first of of the file as headers. When `false`, it treats the first
```
parkr, line 161: First of of the file? Seems like two "of"s there, one might need to be "line" to make this sentence make sense.

docs/_docs/datafiles.md
```diff
@@ -148,3 +148,28 @@ author: dave
...
+              row as data. Defaults to `true`.
+
+```yaml
```
parkr, line 164: Maybe call this block out as examples, not as the default values.

parkr (approved):
Nice! Thanks for adding docs. Two comments.

docs/_docs/datafiles.md
MichaelCordingley: 🤦

MichaelCordingley: Fixed.

docs/_docs/datafiles.md
MichaelCordingley: Like so?

docs/_docs/datafiles.md
```diff
@@ -148,3 +148,30 @@ author: dave
 {% endraw %}
 
 For information on how to build robust navigation for your site (especially if you have a documentation website or another type of Jekyll site with a lot of pages to organize), see [Navigation]({{ '/tutorials/navigation/' | relative_url }}).
+
+## CSV/TSV Parse Options
+
+The way that Ruby parses CSV and TSV files can be customized with the `csv_reader` and `tsv_reader`
```
ashmaroli, line 154: `that` doesn't seem to quite fit here..
```suggestion
The way Ruby parses CSV and TSV files can be customized with the `csv_reader` and `tsv_reader`
```

docs/_docs/datafiles.md
```diff
@@ -148,3 +148,30 @@ author: dave
...
+configuration options. Each configuration key exposes the same options:
+
+`converters`: What [CSV converters](https://ruby-doc.org/stdlib-2.6.1/libdoc/csv/rdoc/CSV.html#Converters) should be
```
ashmaroli, line 157: Like I mentioned in a prior commit message, recommend the cuurent minimum supported Ruby.
```suggestion
`converters`: What [CSV converters](https://ruby-doc.org/stdlib-2.5.0/libdoc/csv/rdoc/CSV.html#Converters) should be
```

docs/_docs/datafiles.md
```diff
@@ -148,3 +148,30 @@ author: dave
...
+              used when parsing the file. This will need to be set with `numeric` if you want number fields in the
+              CSV to be parsed into numeric types rather than into strings. By default, this list is empty.
```
ashmaroli, line 159: Instead of cutting off, lets suggest available options.
For example,
```
Supported values are `all`, `datetime`, `numeric`, etc
```
(if more available or better link to external source).

ashmaroli (changes requested):
Requires some amendments to the documentation.
Plus, you need to fix our `check-spelling` action failures.

P.S. Don't bother squashing / rebasing. We squash commits during merge.

docs/_docs/datafiles.md
MichaelCordingley: Good call.

docs/_docs/datafiles.md
MichaelCordingley: Added.

MichaelCordingley:
> Plus, you need to fix our check-spelling action failures.

It looks like it's taking issue with the link to the Ruby docs.

MichaelCordingley:
I think that gets everything. 🤞

ashmaroli (approved):

ashmaroli:
Thanks @MichaelCordingley
@jekyllbot: merge +minor

PR #8877, opened 2021-11-12
Improve readability of CI logs

ashmaroli:
## Summary

- Running CI via an Earthfile on GitHub Actions results in extremely noisy log outputs. Therefore, run directly. See https://github.com/jekyll/jekyll/runs/4188128787?check_suite_focus=true for an example of a noisy output. The `Earthfile` however  has not been removed to maintain compatibility with the original intention behind introducing the medium.
-  Rewrote the workflow file to improve readability of CI job names and step names.
- Additionally, lock JRuby builds to `http_parser-0.6.0` because newer versions do not have a Java counterpart.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,60 +3,66 @@ name: Continuous Integration
 on:
   push:
     branches:
-    - master
-    - /.*-stable/
+      - master
+      - "*-stable"
   pull_request:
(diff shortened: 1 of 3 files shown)


mattr- (approved):

ashmaroli:
@jekyllbot: merge +fix

PR #8880, opened 2021-11-14
Support both tzinfo v1 and v2. Support non-half hour offsets.

### LINKED ISSUE #8516: Update dependency constraint to allow for tzinfo v2.0.4
Hey there! :wave:

I noticed that the constraint you have for tzinfo doesn't allow for the latest version to be used.

The constraint I found was `~> 1.2`, and the latest version available is `2.0.4`.

philr:

## Summary

This pull request changes `WinTZ.calculate` to support use with tzinfo v2 as well as the currently supported v1. Instead of determining the offset from the difference between local time and UTC, it now uses the defined offset returned by tzinfo (in a v1 and v2 compatible way).

The offset minutes calculation has also been changed to allow for time zones that don't use hour or half hour offsets (e.g. Australia/Eucla's UTC+08:45).

The new `gemfile_contents` and Windows documentation have updated to allow use with either tzinfo v1 or v2.

Unit tests for WinTZ have been added. This required adding tzinfo and tzinfo-data to the `:test` group in the Gemfile.

## Context

Resolves #8516.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,6 +32,8 @@ group :test do
   gem "test-theme", :path => File.expand_path("test/fixtures/test-theme", __dir__)
   gem "test-theme-skinny", :path => File.expand_path("test/fixtures/test-theme-skinny", __dir__)
   gem "test-theme-symlink", :path => File.expand_path("test/fixtures/test-theme-symlink", __dir__)
+  gem "tzinfo", ">= 1", "< 3"
+  gem "tzinfo-data"
 
   if RUBY_ENGINE == "jruby"
     gem "http_parser.rb", "~> 0.6.0"
@@ -80,13 +82,6 @@ group :jekyll_optional_dependencies do
     gem "classifier-reborn", "~> 2.2"
     gem "liquid-c", "~> 4.0"
     gem "yajl-ruby", "~> 1.4"
-  end
-
-  # Windows and JRuby does not include zoneinfo files, so bundle the tzinfo-data gem
-  # and associated library
-  platforms :jruby, :mswin, :mingw, :x64_mingw do
-    gem "tzinfo", "~> 1.2"
-    gem "tzinfo-data"
   end
 end
 
--- a/docs/_docs/installation/windows.md
+++ b/docs/_docs/installation/windows.md
@@ -121,21 +121,13 @@ While 'new' blogs created with Jekyll v3.4 and greater, will have the following 
 sites *will* have to update their `Gemfile` (and installed gems) to enable development on Windows:
 
 ```ruby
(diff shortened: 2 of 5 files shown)


ashmaroli:
@philr Thank you very much for submitting this PR.
Everything LGTM on a quick read-through. If its not asking too much, could you please add a Cucumber based test affirming the non-half hour offset behavior in `features/site_configuration.feature`.

This module is currently used only for builds on Windows platform. So, does **native zoneinfo on non-Windows OS / platforms** correctly set non-half hour offset out-of-the-box? If not, could the `WinTZ` module be re implemented as a platform-agnostic TZ handler? (Mainly asking w.r.t a long-existing issue that got closed due to inactivity; #7550)

philr:
I've added a scenario to test the non-half hour offset behaviour.

Native zoneinfo will handle the non-half hour offset correctly.

mattr- (approved):
LGTM

ashmaroli:
@mattr- Requesting a re-review for this one.
I updated the branch to:
  - use `tzinfo-data` gem and `tzinfo` gem only on Windows (and JRuby)
  - additional Appveyor test-matrix entries to explicitly test with `tzinfo v1` and `tzinfo v2`. (*Downside being increased CI runtime*).

Gemfile
```diff
@@ -32,8 +32,6 @@ group :test do
   gem "test-theme", :path => File.expand_path("test/fixtures/test-theme", __dir__)
   gem "test-theme-skinny", :path => File.expand_path("test/fixtures/test-theme-skinny", __dir__)
   gem "test-theme-symlink", :path => File.expand_path("test/fixtures/test-theme-symlink", __dir__)
-  gem "tzinfo", ">= 1", "< 3"
```
philr: I moved tzinfo into the test group because the new TestWinTz tests run on all platforms. The tests are still passing on non-Windows because tzinfo is being brought in through another dependency.

ashmaroli: The `WinTZ` utility module is only used in Windows:
https://github.com/jekyll/jekyll/blob/f51ccbfad3e2df46a714a3c50b80f2556529c774/lib/jekyll.rb#L133-L139

So as of now, testing the module on non-Windows platforms is redundant.
That said, I was hoping to use the `WinTZ` in all platforms in a separate PR, after merging this.

mattr- (approved):

ashmaroli:
Totally forgot about this.
Thank you very much @philr. :pray: 

@jekyllbot: merge +minor

parkr:
Backporting to v3.9.x in https://github.com/jekyll/jekyll/pull/9280.

PR #8883, opened 2021-11-16
highlight tag support variable lang

pythias:

## Summary

Allow send variable to `highlight` tag

```markdown
---
examples:
  - lang: php
    code: > 
      
      function test() {
        echo date('Y-m-d');
      }
  - lang: ruby
    code : >

      def foo
        puts 'foo'
      end
---

{% for example in page.examples %}

{% highlight example.lang %}
{{ example.code }}
{% endhighlight %}

{% endfor %}
```

### DIFF
--- a/lib/jekyll/tags/highlight.rb
+++ b/lib/jekyll/tags/highlight.rb
@@ -34,18 +34,19 @@ module Jekyll
         prefix = context["highlighter_prefix"] || ""
         suffix = context["highlighter_suffix"] || ""
         code = super.to_s.gsub(LEADING_OR_TRAILING_LINE_TERMINATORS, "")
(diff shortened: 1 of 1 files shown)


pythias (commented):

## Summary

Allow send variable to `highlight` tag

```markdown
---
examples:
  - lang: php
    code: > 
      
      function test() {
        echo date('Y-m-d');
      }
  - lang: ruby
    code : >

      def foo
        puts 'foo'
      end
---

{% for example in page.examples %}

{% highlight example.lang %}
{{ example.code }}
{% endhighlight %}

{% endfor %}
```

pythias (commented):
When an file has differents highlighter languages, the @lang will be set to the last one. So the var lang should be an param pass to the function

PR #8890, opened 2021-11-25
docs: Fix of a bug in the pagination section

swagar:
There is a little bug in the laste code snippet in the pageination section. The link for the first page does not work when the `paginate_path` property points to a subfolder. In that case `{{ '/' | relative_url }}` points to the base url and ignores the subfolder.

I made a little demo which you can find [here](https://swagar.github.io/wrong-page1-link/blog/page2/). There you find the link for the first page twice. One with the old link und one with the new and working link.

You can finde the source code [here](https://github.com/swagar/wrong-page1-link/blob/master/blog/index.html#L38)

### DIFF
--- a/docs/_docs/pagination.md
+++ b/docs/_docs/pagination.md
@@ -154,7 +154,7 @@ page with links to all but the current page.
     {% if page == paginator.page %}
       <em>{{ page }}</em>
     {% elsif page == 1 %}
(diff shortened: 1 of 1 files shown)


ashmaroli (approved):
lgtm!

ashmaroli:
Thanks @swagar
@jekyllbot: merge +doc

PR #8891, opened 2021-11-26
add attr data-lang to <figure> tag

hsayed21:
**code refactoring:**

## Summary
by default  when adding fenced block code, a rouge highlighter will render 

\```bash
 code
\``` 
to
```html
<div class="language-bash highlighter-rouge">
          <div class="highlight">
               <pre class="highlight">
                  <code> code here..... </code>
               </pre>
          </div>
<div>
```
the problem here if you want to add language label, easily using kramdown block inline
after end code add
\```html
\```
{: lang="HTML" }

will add attribute to parent div will be \<div lang="HTML" class="language-bash highlighter-rouge">

using CSS style content: attr(lang)
you can improve label CSS style 

but problem with **liquid highlight tag**
{% highlight bash %}
......code
{% endhighlight %}

will render \<figure> tag but without lang name

### DIFF
--- a/lib/jekyll/tags/highlight.rb
+++ b/lib/jekyll/tags/highlight.rb
@@ -104,7 +104,7 @@ module Jekyll
           "class=\"language-#{@lang.to_s.tr("+", "-")}\"",
           "data-lang=\"#{@lang}\"",
         ].join(" ")
-        "<figure class=\"highlight\"><pre><code #{code_attributes}>"\
+        "<figure class=\"highlight\" data-lang=\"#{@lang}\"><pre><code #{code_attributes}>"\
         "#{code.chomp}</code></pre></figure>"
       end
     end


ashmaroli:
@hsayed21 I understand your proposal, but IMO, this is just duplication.
If you look at line#105 in the diff, you'll see that the markup already contains `data-lang` attribute, albeit attached to
`figure > pre > code` instead of `figure` directly.

hsayed21:
@ashmaroli 
I understand, but for some reason, it is not possible to `figure` tag to access attribute data-lang of `code` tag

hsayed21:
Using CSS or Sass

BenSouchet:
@hsayed21 Just for info, the `:has()` css selector ( https://caniuse.com/css-has ) will one day be available on major browsers, for info it's currently in active development in Google Chrome (see https://bugs.chromium.org/p/chromium/issues/detail?id=669058 ).

With this selector you will be able to to exactly what you want, do something like `figure:has(pre>code[lang="bash"]) {...}`.

hsayed21:
> Indeed!, it does what is required, but I think it is not supported in most browsers yet

parkr (commented):
I'm ok with this. The one big downside is that folks with CSS selectors for data-lang will have 2 elements for the same code block they're matching against.

PR #8905, opened 2021-12-14
Update resources.md

ngvoicu:


mattr-:
:wave: Appreciate your contribution and recognize that it's hard to start a new business in general, and perhaps doubly so for a Jekyll theme. However, the other sites listed contain more than a single theme and I'd like to continue that tradition of listing sites that are theme aggregators rather than offering a single theme. Thanks!

PR #8907, opened 2021-12-16
Remove text on GITHUB_TOKEN which is now built-in

letmaik:
Just a documentation update.

### DIFF
--- a/docs/_docs/continuous-integration/github-actions.md
+++ b/docs/_docs/continuous-integration/github-actions.md
@@ -129,7 +129,7 @@ The above workflow can be explained as the following:
 - We specify our selected **action** and **version number** using `helaili/jekyll-action@2.0.5`.
   This handles the build and deploy.
 - We set a reference to a secret **environment variable** for the action to use. The `GITHUB_TOKEN`
-  is a _Personal Access Token_ and is detailed in the next section.
+  is available out-of-the-box in GitHub Actions.
 
 Instead of using the **on.push** condition, you could trigger your build on a **schedule** by
 using the [on.schedule] parameter. For example, here we build daily at midnight by specifying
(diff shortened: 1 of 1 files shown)


ashmaroli:
I agree with your take that there's no explicit need to *create* `GITHUB_TOKEN`. But creating a `PAT` grants freedom on the scope.
I'll approve this PR if you will add a link to the documentation on the restricted scope-flexibility of `GITHUB_TOKEN` and consequently recommend using a PAT for additional flexibility.

letmaik:
Could you go into more detail on how a PAT allows to limit the scope further?  `GITHUB_TOKEN` is scoped to the current repo and can be further limited through

```yaml
    permissions:
      contents: write
```

A PAT cannot be scoped to a single repo. The thing I don't fully understand is how the `public_repo` OAuth scope compares to the `contents: write` scope of `GITHUB_TOKEN`.

ashmaroli:
> Could you go into more detail on how a PAT allows to limit the scope further?

I think you *misread* my comment. I was referring to the *limited scope* of the auto-generated `GITHUB_TOKEN` w.r.t to a PAT currently recommended by the live Jekyll Documentation..

Anyways, the GitHub docs on the relevant info are at:
- [about `GiTHUB_TOKEN`](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret)
- [granting additional permissions via PAT](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#granting-additional-permissions)

letmaik:
> > Could you go into more detail on how a PAT allows to limit the scope further?
> 
> I think you _misread_ my comment. I was referring to the _limited scope_ of the auto-generated `GITHUB_TOKEN` w.r.t to a PAT currently recommended by the live Jekyll Documentation..

I was caught up by "consequently recommend using a PAT for additional flexibility". I can't think of any scenario where it makes sense to recommend using a PAT for github pages.

docs/_docs/continuous-integration/github-actions.md
```diff
@@ -129,7 +129,7 @@ The above workflow can be explained as the following:
 - We specify our selected **action** and **version number** using `helaili/jekyll-action@2.0.5`.
   This handles the build and deploy.
 - We set a reference to a secret **environment variable** for the action to use. The `GITHUB_TOKEN`
-  is a _Personal Access Token_ and is detailed in the next section.
+  is available out-of-the-box in GitHub Actions.
```
ashmaroli, line 132:
```suggestion
  is a secret token automatically initialized at the start of every workflow run.
  More information can be found in [GitHub documentation](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret). 
```

ashmaroli (approved):

ashmaroli:
> I can't think of any scenario where it makes sense to recommend using a PAT for github pages.

Fair enough.
I'll merge when the CI run finishes.
Thanks @letmaik

PR #8918, opened 2021-12-29
Lock gem `psych` to `v3.x`

ashmaroli:
## Summary

`psych-4.x` default in Ruby 3.1 breaks our Cucumber based tests.

## Context

```
(::) failed steps (::)

Tried to load unspecified class: Date (Psych::DisallowedClass)
(eval):2:in `date'
/home/runner/work/jekyll/jekyll/features/step_definitions.rb:166:in `/^I have a configuration file with "(.*)" set to "(.*)"$/'
/home/runner/work/jekyll/jekyll/features/step_definitions.rb:175:in `block (2 levels) in <top (required)>'
/home/runner/work/jekyll/jekyll/features/step_definitions.rb:174:in `each'
/home/runner/work/jekyll/jekyll/features/step_definitions.rb:174:in `/^I have a configuration file with:$/'
features/site_configuration.feature:211:in `I have a configuration file with:'
```

### DIFF
--- a/jekyll.gemspec
+++ b/jekyll.gemspec
@@ -43,7 +43,11 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency("liquid",                "~> 4.0")
   s.add_runtime_dependency("mercenary",             ">= 0.3.6", "< 0.5")
   s.add_runtime_dependency("pathutil",              "~> 0.9")
+
+  # Ruby 3.1.0 shipped with `psych-4.0.3` which caused some of our Cucumber-based tests to fail.
+  # TODO: Remove lock once we implement a way to use Psych 4 without breaking anything.
   s.add_runtime_dependency("psych",                 "~> 3.3")
+
   s.add_runtime_dependency("rouge",                 "~> 3.0")
   s.add_runtime_dependency("safe_yaml",             "~> 1.0")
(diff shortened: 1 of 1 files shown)


mattr-:
If we switch to unsafe loading, can we avoid locking Psych? If we can't avoid locking Psych, would you mind adding a `TODO` comment of some sort explaining why we locked it?

fauno:
Is it related to https://github.com/jekyll/jekyll/pull/8772?

ashmaroli:
> If we switch to unsafe loading, can we avoid locking Psych?

I'm not so confident of making the switch to *unsafe loading*, especially for a minor release Jekyll 4.3.0. Perhaps we'll get enough feedback by the time Jekyll 4.4.0 or Jekyll 5.0 materializes..?

Ruby 3.0.0 shipped with `psych-3.3.0` as a default gem and based on the CI results here, the locking doesn't introduce regressions even with Ruby 3.1.0 which shipped with `psych-4.0.3` instead.

Therefore, my take is that *locking* is the optimal route now.


> Is it related to #8772?

Not really @fauno. That PR doesn't explicitly use Psych 4.0 instead relying on the default version shipped with the Ruby version in use.

mattr- (approved):
Appreciate the comment here. Let's :shipit:

mattr-:
@jekyllbot: merge +dev

PR #8924, opened 2022-01-03
Add `jekyll console` command

ashmaroli:

## Summary

Allows running `jekyll console` / `jekyll c` to start an `irb` session preloaded with a site instance that can be accessed via `site`.
The site instance is actually a subclass of `Jekyll::Site`, that responds to additional methods outputting additional console messages.

## Context

*Super enhanced* version of #8164
Closes #8164

### DIFF
--- a/lib/jekyll/commands/console.rb
+++ b/lib/jekyll/commands/console.rb
@@ -81,13 +81,7 @@ module Jekyll
         end
 
         def start_console
-          print "\nInitializing console "
-
-          60.times do
-            print "."
-            sleep 0.03
-          end
-          log "\n\n"
+          log "\nInitializing console..\n\n"
 
           CONSOLE_DEPS.each { |dep| require dep }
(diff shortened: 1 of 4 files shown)


lib/jekyll/commands/console.rb
```diff
@@ -0,0 +1,115 @@
...
+        end
+
+        def start_console
+          print "\nInitializing console "
+
+          60.times do
+            print "."
+            sleep 0.03
+          end
+          log "\n\n"
```
mattr-, line 90: What do you think about removing the printing of the 60 dots here? It really don't do anything other than be an artificial delay. Does it have some other purpose?

ashmaroli: You're right in stating that it doesn't do anything but create an artificial delay.
I intended it to be an animation just to spice things up slightly.
I'll remove it.

mattr- (approved):

PR #8938, opened 2022-01-17
Fix response header for WASM assets during `jekyll serve`

### LINKED ISSUE #8936: [Bug]: `jekyll serve` sets the wrong content type for .wasm files
### Operating System

Ubuntu 20.04

### Ruby Version

ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-linux]

### Jekyll Version

jekyll 4.2.1

### GitHub Pages Version

N/A. The issue occurs when serving my site locally via `bundle exec jekyll serve`.

### Expected Behavior

When serving a site with `jekyll serve`, asset files with the extension `.wasm` should be served with the Content-Type `application/wasm`.

IMPORTANT: It can't be `application/wasm; charset=utf-8`! It needs to be `application/wasm` in order for WebAssembly to work.

### Current Behavior

At the moment, I'm seeing the Content-Type of the .wasm asset coming back as `application/wasm; charset=utf-8`.

IIRC, earlier, it was coming back as `text/html; charset=utf-8`, which is even worse. I've been tinkering locally, so it's possible that I've halfway fixed it to return `application/wasm; charset=utf-8`, but even so, it seems like something deep in the internals of Jekyll or WEBrick is automatically adding the `;charset=utf-8` part.

NOTE: This issue was originally raised 3 years ago: https://github.com/jekyll/jekyll/issues/7743  but there was no response and your bot automatically closed it as stale after 5 months. The problem persists, though. Please don't close this without responding!

ashmaroli:
- This is a 🐛 bug fix.

## Summary

`.wasm` files cannot be served with response header `application/wasm; charset=utf-8`.
Therefore, don't inject UTF-8 charset for `application/wasm` mime-type.

## Context

Resolves #8936
/cc @daveyarwood

### DIFF
--- a/lib/jekyll/commands/serve/servlet.rb
+++ b/lib/jekyll/commands/serve/servlet.rb
@@ -185,6 +185,10 @@ module Jekyll
           key = res.header.keys.grep(%r!content-type!i).first
           typ = res.header[key]
 
+          # WASM require the response header to be exactly "application/wasm". Even a trailing
+          # semicolon is illegal.
+          return if typ == "application/wasm"
+
           unless %r!;\s*charset=!.match?(typ)
             res.header[key] = "#{typ}; charset=#{@jekyll_opts["encoding"]}"
           end


parkr:
This seems like a bandaid that gets around the real problem: binary types don't have a charset. We use mime-db (or did, anyway) to generate the mime.types file, and it does include if there's a charset. Perhaps we could write this such that we add the charset for types which support it? https://github.com/jshttp/mime-db#data-structure https://github.com/jshttp/mime-db/blob/master/db.json 

Only text-based files have a charset, all others are binary, so if we can solve this for all binary formats then we can also solve this for wasm.

ashmaroli:
> Only text-based files have a charset, all others are binary...

[insert *now why didn't I think of that* emoji]
Excellent point @parkr. So, should *Jekyll's* `mime.types` include a new cell containing charset data for applicable types?

PR #8947, opened 2022-01-23
Document yes/no/true/false collection value type inference

PrinsFrank:

## Summary

Add conversion of yes/no/true/false to true/false values to documentation.

## Context

When working with a list of alpha-2 country codes in a collection, I noticed that the following collection:
```
country_codes
 - NO
```
was converted to: 
```
country_codes
 - false
```
I understand converting true and false values to their respective types, but have not seen converting 'yes' and 'no' before, so I added this to the documentation.

### DIFF
--- a/docs/_docs/collections.md
+++ b/docs/_docs/collections.md
@@ -46,6 +46,12 @@ information, see the section <a href="#output">Output</a>.
   <p>If you specify a directory to store all your collections in the same place with <code>collections_dir: my_collections</code>, then you will need to move your <code>_drafts</code> and <code>_posts</code> directory to <code>my_collections/_drafts</code> and <code>my_collections/_posts</code>. Note that, the name of your collections directory cannot start with an underscore (`_`).</p>
 </div>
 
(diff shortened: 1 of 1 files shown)


ashmaroli:
@PrinsFrank Thank you for proposing this addition.
To throw some additional light on the matter, the reason behind this is that, **YAML**, considers these inputs as Boolean true or false.
Therefore, this is applicable to not just documents in collections, but to all places YAML is / may be used. For example, front matter of standalone pages, YAML data files, config files written in YAML, etc.

PrinsFrank:
@ashmaroli That is the case for [YAML 1.1](https://yaml.org/type/bool.html), but not [1.2 that was published on 2009-07-21](https://yaml.org/spec/1.2.0/#id2602744). I've only worked with more recent implementations so that's why I was confused. I guess this is a deeper issue in underlying libs, but I don't have the experience in ruby and gems to fix this. I will close this for now is it is not really breaking.

PR #8948, opened 2022-01-28
Recommend using `actions/cache` in GitHub Actions documentation

detohm:
## Summary
Update the document of Github Actions to add cache action step for reducing the build times. Please refer to the original document of [jekyll-action](https://github.com/helaili/jekyll-action). The cache step is also included for performance improvement.

Therefore, I think this change would benefit all bloggers who are using or begin to use Github Actions to publish thier sites.

There are changes below,
- Update example of workflow .yml code
- Update code description


ashmaroli (approved):

ashmaroli:
Thank you @detohm
@jekyllbot: merge +doc

PR #8965, opened 2022-02-10
Fix response header for content served via `jekyll serve`

### LINKED ISSUE #8936: [Bug]: `jekyll serve` sets the wrong content type for .wasm files

ashmaroli:
- This is a 🐛 bug fix.

## Summary

Use `script/vendor-mimes` to **generate a JSON version of the data** used to generate `lib/jekyll/mime.types` and use that JSON during `jekyll serve` to determine whether `charset=` has to be injected to the `"content-type "` in the response-header.

- Enhance `script/vendor-mimes` to generate **`lib/jekyll/commands/serve/mime_types.json`**.
- Enhance `script/vendor-mimes` to log statements during runtime.
- Update servlet to use generated JSON file during `jekyll serve`.
- Update `lib/jekyll/mime.types` with current data at upstream repository.

## Context

Closes #8938
Resolves #8936


script/vendor-mimes
```diff
@@ -2,16 +2,40 @@
...
+# ----
+
+json = begin
+  log_info "Reading remote data.."
+  URI.open("https://raw.githubusercontent.com/jshttp/mime-db/master/db.json").read
+rescue StandardError => e
+  log_error "Error reading remote data!"
+  log_error e.message
+  log_error "Aborting."
+  exit
```
parkr, line 30: nit: prefer `abort` when an error occurs

script/vendor-mimes
```diff
@@ -25,11 +49,22 @@ data.each do |mime, meta|
...
+config = File.expand_path "../lib/jekyll/mime.types", __dir__
 File.write(config, output)
+log_info "Done! See: #{config.inspect.white}"
+
+# --- Generate JSON file from mime-type data ----
+puts
+
+log_info "Dumping sanitized hash to file as JSON.."
+json_file = File.expand_path "../lib/jekyll/commands/serve/mime_types.json", __dir__
+File.write(json_file, JSON.pretty_generate(data) + "\n")
```
parkr, line 69: Saving the whole database means we have superfluous data. That bloats the size of the gem and it increases start-up time. I'd love to see the pre-processing for the file happen here. Filter the DB to a JSON object whose keys are MIME types and the value is the desired charset, like:

```json
{
    "text/plain": "utf-8"
    // ...
}
```

Then in the servlet, you can query the object by the MIME type like you're doing now, but its value is the charset rather than an object potentially containing the charset.

If this doesn't dramatically reduce the size of the json file then we can keep it as-is.

lib/jekyll/commands/serve.rb
```diff
@@ -194,6 +194,7 @@ def webrick_opts(opts)
             :JekyllOptions      => opts,
             :DoNotReverseLookup => true,
             :MimeTypes          => mime_types,
+            :MimeTypesJSON      => File.read(File.expand_path("serve/mime_types.json", __dir__)),
```
parkr, line 197: If we decide to follow my suggestion to reduce the scope of this file, then perhaps we can rename this to something like `:MIMETypeCharsets` and call the file something like `mime_type_charsets.json` to match.

parkr (commented):
This is great, thank you! My primary feedback is to consider slimming down the database to just the data we need. More detailed comment left in-line.

script/vendor-mimes
ashmaroli: I am on board with the idea.
Since the servlet only uses the JSON object to inject the charset, it makes sense to remove the surrounding cruft.

script/vendor-mimes
ashmaroli: I'm vetoing this because using `abort` results in Ruby 2.7 spitting out the stack trace which I don't want for a dev-dependency-script.

parkr: The point of `abort` is to exit with the status code != 0 so that it's clear there was an error. This is a UNIX convention going back to the beginnings of UNIX. You can `echo $?` to see if the command failed and exiting with a non-zero status code will cause `&&` statements to fail.

`Kernel#exit` allows you to pass in an argument to change the exit status:

```suggestion
  exit 1
```

script/vendor-mimes
```diff
@@ -23,13 +48,27 @@ data.each do |mime, meta|
...
 strlen = mimes.keys.max_by(&:length).length
 output = ""
 output << "# Woah there. Do not edit this file directly.\n"
 output << "# This file is generated automatically by script/vendor-mimes.\n\n"
 mimes = mimes.sort_by { |k,v| k }
 output << mimes.map { |mime,extensions| "#{mime.ljust(strlen)} #{extensions.sort.join(" ")}" }.join("\n")
 
+config = File.expand_path "../lib/jekyll/mime.types", __dir__
 File.write(config, output)
+log_info "Done! See: #{config.inspect.white}"
```
parkr, line 66: For the other file, you have `log_success` when you've completed. Should you use log_success here, too?

Screen Shot 2022-02-11 at 10 08 05 AM

script/vendor-mimes
```diff
@@ -23,13 +48,27 @@ data.each do |mime, meta|
...
+
+# --- Generate JSON file from charset_data ----
+puts
+
+log_info "Dumping miimetype-charset mapping as JSON.."
```
parkr, line 71:
```suggestion
log_info "Dumping mimetype-charset mapping as JSON.."
```

script/vendor-mimes
```diff
@@ -23,13 +48,27 @@ data.each do |mime, meta|
   next if extensions.empty?
   mimes[mime] = [] if mimes[mime].nil?
   mimes[mime].concat extensions
+
+  # Extract mime-types with "charset" metadata
+  charset_data[mime] = meta["charset"] if meta.key?("charset")
```
parkr, line 53: We went from (current behavior) every MIME type gets the default charset to (proposed behavior) only a dozen or so get the default charset. I think we're missing some here. How would you feel about adding all `text/*` MIME types here?
```suggestion
  charset_data[mime] = meta["charset"] if meta.key?("charset")
  charset_data[mime] ||= "default" if mime.start_with?("text/")
```

parkr (commented):
Thanks for updating! I like the smaller list. Oddly... it feels like we're missing some MIME types like text/html which are clearly text-based and would likely . 🤔 Left a comment inline.

script/vendor-mimes
ashmaroli: It was initially. =)
But the green in the middle kinda felt like it broke the flow for me.. :)

Kay-kay2019: :)

script/vendor-mimes
Kay-kay2019: Thank you so very much I appreciate  everyone's help

script/vendor-mimes
Kay-kay2019: 👍

parkr (approved):

parkr:
The DB provides the default charset that IANA specified for a MIME type. This code was introduced to fix the bug where a lot of text documents don't have a default charset (which generally means it uses ASCII). I think we found a good solution for adding the charset to files with characters (text files mostly) and remove the charset header from files without characters like WASM. Great work! 👍

parkr:
@jekyllbot: merge +fix

PR #8972, opened 2022-02-20
Prefer "switch" term instead of "flag"

EmilySeville7cfg:
## Summary

Flag [may be understood](https://superuser.com/questions/1070059/is-there-a-difference-between-a-command-line-flag-and-a-command-line-option) as an option without any value turning on/off some feature.


ashmaroli:
@EmilySeville7cfg In the StackExchange thread you've linked to, there's an answer that sheds some light on both `switch` and `flag`.
I therefore would like to see both used in our docs with an explanation preface that removes the ambiguity of their usage in the entire document.

PR #8983, opened 2022-03-03
Use array for plugins_dir instead of string

EmilySeville7cfg:
## Summary

Use array of strings for `plugins_dir` instead of string default value.

### DIFF
--- a/docs/_docs/configuration/default.md
+++ b/docs/_docs/configuration/default.md
@@ -19,7 +19,7 @@ file or on the command-line.
 source              : .
 destination         : ./_site
 collections_dir     : .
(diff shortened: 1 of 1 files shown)


mattr-:
The default for `plugins_dir` is the string `_plugins`, so it makes sense to keep it the same way here since that would match the default.

As part of the options help output for the Jekyll binary, we also indicate that `plugins_dir` could be an array. 

A little bit more context about the problem you're trying to address would be helpful here. Thanks! 🙇‍♂️

ashmaroli:
Closing as use of string `_plugins` is perfectly valid.

Issue #8784, opened 2021-08-23
Can't use Jekyll on macOS Monterey (Apple Silicon M1)


rizwan95:
I am unable to use Jekyll on macOS Monterey. I even tried using arch -x86_64 but it gives me the following error 
`arch: posix_spawnp: bundle: Bad CPU type in executable`

  - I have updated ruby. 
  - I have done bundle install
  - I have ran these commands ->  bundle update ffi sassc,  gem install ffi, arch -x86_64 gem install ffi   
  

## My Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Operating System |     macOS Monterey  (M1 chip)     |
| `jekyll`         | Latest     |

---

## Current Behavior
Getting the following error 
```
/Users/<user>/.gem/ruby/3.0.0/gems/ffi-1.15.3/lib/ffi/library.rb:145:in `block in ffi_lib': Could not open library '/opt/homebrew/lib/ruby/gems/3.0.0/gems/sassc-2.4.0/ext/libsass.bundle': dlopen(/opt/homebrew/lib/ruby/gems/3.0.0/gems/sassc-2.4.0/ext/libsass.bundle, 0x0005): tried: '/opt/homebrew/lib/ruby/gems/3.0.0/gems/sassc-2.4.0/ext/libsass.bundle' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64e')), '/usr/local/lib/libsass.bundle' (no such file), '/usr/lib/libsass.bundle' (no such file) (LoadError)
	from /Users/<user>/.gem/ruby/3.0.0/gems/ffi-1.15.3/lib/ffi/library.rb:99:in `map'
	from /Users/<user>/.gem/ruby/3.0.0/gems/ffi-1.15.3/lib/ffi/library.rb:99:in `ffi_lib'
	from /opt/homebrew/lib/ruby/gems/3.0.0/gems/sassc-2.4.0/lib/sassc/native.rb:13:in `rescue in <module:Native>'
	from /opt/homebrew/lib/ruby/gems/3.0.0/gems/sassc-2.4.0/lib/sassc/native.rb:10:in `<module:Native>'
	from /opt/homebrew/lib/ruby/gems/3.0.0/gems/sassc-2.4.0/lib/sassc/native.rb:6:in `<module:SassC>'
	from /opt/homebrew/lib/ruby/gems/3.0.0/gems/sassc-2.4.0/lib/sassc/native.rb:5:in `<top (required)>'
[...]

PR #8993, opened 2022-03-11
Simplify macOS installation docs

monfresh:

## Summary

Instead of providing various options for installing Ruby, it's better to
stick to a single reliable solution: using a version manager. This
drastically simplifies this doc and makes it more maintainable.

Over the past 10 years, I've maintained detailed tutorials for properly
installing Ruby on a Mac. I've helped thousands of people, and I've seen
the same issues over and over.

## Context

Per @ashmaroli's request here:

https://github.com/jekyll/jekyll/issues/8784#issuecomment-1060267860

I've linked to my Jekyll installation guide and I pledge to keep it up
to date, as I have with my other Ruby on Mac guides over the past ten
years.

Here's a collection of tweets from grateful people who have benefited
from my guides and script for setting up a proper Ruby environment on a
Mac:

https://twitter.com/monfresh/timelines/1480261956988596228

### DIFF
--- a/docs/_docs/installation/macos.md
+++ b/docs/_docs/installation/macos.md
@@ -61,8 +61,8 @@ This will take a few minutes, and once it's done, configure your shell to
 automatically use `chruby`:
 
 ```sh
-echo "source /usr/local/opt/chruby/share/chruby/chruby.sh" >> ~/.zshrc
-echo "source /usr/local/opt/chruby/share/chruby/auto.sh" >> ~/.zshrc
+echo "source $(brew --prefix)/opt/chruby/share/chruby/chruby.sh" >> ~/.zshrc
+echo "source $(brew --prefix)/opt/chruby/share/chruby/auto.sh" >> ~/.zshrc
 echo "chruby ruby-{{ site.data.ruby.current_version }}" >> ~/.zshrc
 ```
 
--- a/docs/_docs/installation/macos.md
+++ b/docs/_docs/installation/macos.md
@@ -66,7 +66,6 @@ echo "source $(brew --prefix)/opt/chruby/share/chruby/auto.sh" >> ~/.zshrc
 echo "chruby ruby-{{ site.data.ruby.current_version }}" >> ~/.zshrc
 ```
 
-If you're on an M1 Mac, replace `/usr/local` with `/opt/homebrew`.
 If you're using Bash, replace `.zshrc` with `.bash_profile`. If you're not sure, 
 read this external guide to 
 [find out which shell you're using](https://www.moncefbelyamani.com/which-shell-am-i-using-how-can-i-switch/).
--- a/docs/_docs/installation/macos.md
+++ b/docs/_docs/installation/macos.md
@@ -83,7 +83,7 @@ Next, read that same external guide for important notes about
 
 ## Install Jekyll
 
-After installing Ruby with chruby, install Jekyll:
+After installing Ruby with chruby, install the latest Jekyll gem:
 
 ```sh
 gem install jekyll
--- a/docs/_docs/installation/macos.md
+++ b/docs/_docs/installation/macos.md
@@ -3,155 +3,28 @@ title: Jekyll on macOS
 permalink: /docs/installation/macos/
 ---
 
-## Install Command Line Tools
-To install the command line tools to compile native extensions, open a terminal and run:
+## Supported macOS versions
 
-```sh
-xcode-select --install
-```
+- Monterey (macOS 12)
+- Big Sur (macOS 11)
+- Catalina (macOS 10.15)
 
-### set SDKROOT (only macOS Catalina or later)
-Starting on macOS Catalina (10.15) the headers used for Ruby have been moved from their previous location which results in some gems, including Jekyll to fail installation. This can be solved by setting SDKROOT in your shell configuration to the value provided by xcrun.
-
-```ssh
-export SDKROOT=$(xcrun --show-sdk-path)
-```
+Older macOS versions might work, but we don't officially support them.
 
 ## Install Ruby
 
-Jekyll requires **Ruby v{{ site.data.ruby.min_version }}** or higher.
-macOS Big Sur 11.x ships with Ruby 2.6.3. Check your Ruby version using `ruby -v`.
+To install Jekyll on macOS, you need a proper Ruby development environment. While macOS comes preinstalled with Ruby, we don't recommend using that version to install Jekyll. This external article goes over the various reasons [why you shouldn't use the system Ruby](https://www.moncefbelyamani.com/why-you-shouldn-t-use-the-system-ruby-to-install-gems-on-a-mac/).
 
-If you're running a previous version of macOS, you'll have to install a newer version of Ruby. Installation with [Homebrew](https://brew.sh) is simple if you're only planning to use Ruby for Jekyll. Install with a version manager such as [asdf](https://asdf-vm.com/), [chruby](https://github.com/postmodern/chruby), [rbenv](https://github.com/rbenv/rbenv), or [rvm](https://rvm.io/) if you need to switch among Ruby versions (instructions for rbenv are below). See the guide [Install Ruby on Mac](https://mac.install.guide/ruby/index.html) for details and recommendations.
+Instead, you'll need to install a separate and newer version of Ruby using a version manager such as [asdf](https://asdf-vm.com/), [chruby](https://github.com/postmodern/chruby), [rbenv](https://github.com/rbenv/rbenv), or [rvm](https://rvm.io/). Version managers allow you to easily install multiple versions of Ruby, and switch between them.
 
-### With Homebrew {#brew}
-To run the latest Ruby version you need to install it through [Homebrew](https://brew.sh).
-
-```sh
-# Install Homebrew
-/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
-
-# Install Ruby
-brew install ruby
-```
-
-Add the brew ruby and gems path to your shell configuration:
-
-```bash
-# If you're using Zsh
-echo 'export PATH="/usr/local/opt/ruby/bin:/usr/local/lib/ruby/gems/3.0.0/bin:$PATH"' >> ~/.zshrc
-
-# If you're using Bash
-echo 'export PATH="/usr/local/opt/ruby/bin:/usr/local/lib/ruby/gems/3.0.0/bin:$PATH"' >> ~/.bash_profile
-
-# Unsure which shell you are using? Type
-echo $SHELL
-```
-
-Relaunch your terminal and check your Ruby setup:
-
-```sh
-which ruby
-# /usr/local/opt/ruby/bin/ruby
-
-ruby -v
-{{ site.data.ruby.current_version_output }}
-```
-
-You're now running the current stable version of Ruby!
-
-### With rbenv {#rbenv}
-
-People often use [rbenv](https://github.com/rbenv/rbenv) to manage multiple
-Ruby versions. This is very useful when you need to be able to run a given Ruby version on a project.
-
-```sh
-# Install Homebrew
-/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
-
-# Install rbenv and ruby-build
-brew install rbenv
-
-# Set up rbenv integration with your shell
-rbenv init
-
-# Check your installation
-curl -fsSL https://github.com/rbenv/rbenv-installer/raw/main/bin/rbenv-doctor | bash
-```
-
-Restart your terminal to apply your changes.
-Next, you can install the Ruby version you want. Let's install the latest stable version:
-
-```sh
-rbenv install {{ site.data.ruby.current_version }}
-rbenv global {{ site.data.ruby.current_version }}
-ruby -v
-{{ site.data.ruby.current_version_output }}
-```
-
-That's it! Head over to [rbenv command references](https://github.com/rbenv/rbenv#command-reference) to learn how to use different versions of Ruby in your projects.
+We recommend `chruby` because it's the most reliable and easiest to use. Follow this external guide for [installing Jekyll on a Mac](https://www.moncefbelyamani.com/how-to-install-jekyll-on-a-mac-the-easy-way/) and setting up a proper Ruby environment with `chruby`.
 
 ## Install Jekyll
 
-After installing Ruby, install Jekyll and Bundler.
-
-### Local Install
-
-Install the bundler and jekyll gems:
+After installing Ruby with chruby, install Jekyll and Bundler:
 
 ```sh
-gem install --user-install bundler jekyll
-```
-
-Get your Ruby version:
-
-```sh
-ruby -v
-{{ site.data.ruby.current_version_output }}
-```
-
-Append your path file with the following, replacing the `X.X` with the first two digits of your Ruby version:
-
-```bash
-# If you're using Zsh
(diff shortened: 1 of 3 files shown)


docs/_docs/installation/macos.md
```diff
@@ -3,155 +3,28 @@ title: Jekyll on macOS
...
+## Supported macOS versions
 
-```sh
-xcode-select --install
-```
+- Monterey (macOS 12)
+- Big Sur (macOS 11)
+- Catalina (macOS 10.15)
 
-### set SDKROOT (only macOS Catalina or later)
```
monfresh, line 13: In the past ten years of installing Ruby on a Mac on every OS X and macOS release, I have never had to set `SDKROOT`. This is absolutely not needed.

docs/_docs/installation/macos.md
```diff
@@ -3,155 +3,28 @@ title: Jekyll on macOS
...
-{: .note .warning}
-We recommend not installing Ruby gems globally to avoid file permissions problems and using `sudo`.
-
-#### On Mojave (10.14)
-
-Because of SIP Protections in Mojave, run:
-
-```sh
-sudo gem install bundler
-sudo gem install -n /usr/local/bin/ jekyll
```
monfresh, line 146: We should never ask people to install gems with `sudo`. When installing Ruby with a version manager, `sudo` is never needed.

docs/_docs/installation/macos.md
```diff
@@ -3,155 +3,28 @@ title: Jekyll on macOS
...
-gem env
-```
-
-{: .note .info}
-Every time you update Ruby to a version in which the first two digits change, update your path to match.
-
-### Global Install
-
-{: .note .warning}
-We recommend not installing Ruby gems globally to avoid file permissions problems and using `sudo`.
```
monfresh, line 138: File permissions problems are only an issue when trying to install gems using the system Ruby. When installing Ruby with a reliable version manager like `chruby`, `sudo` is never needed, and there aren't any file permissions problems.

Since we are discouraging using the system Ruby, this section is no longer needed.

docs/_docs/installation/macos.md
```diff
@@ -3,155 +3,28 @@ title: Jekyll on macOS
...
+Instead, you'll need to install a separate and newer version of Ruby using a version manager such as [asdf](https://asdf-vm.com/), [chruby](https://github.com/postmodern/chruby), [rbenv](https://github.com/rbenv/rbenv), or [rvm](https://rvm.io/). Version managers allow you to easily install multiple versions of Ruby, and switch between them.
 
-That's it! Head over to [rbenv command references](https://github.com/rbenv/rbenv#command-reference) to learn how to use different versions of Ruby in your projects.
+We recommend `chruby` because it's the most reliable and easiest to use. Follow this external guide for [installing Jekyll on a Mac](https://www.moncefbelyamani.com/how-to-install-jekyll-on-a-mac-the-easy-way/) and setting up a proper Ruby environment with `chruby`.
 
 ## Install Jekyll
 
-After installing Ruby, install Jekyll and Bundler.
-
-### Local Install
```
monfresh, line 98: This entire local install section only applies when using the system Ruby. When using a version manager like `chruby`, it takes care of setting the `PATH`, and therefore we can eliminate this whole section, and the complexity of having to set the `PATH` and remembering to update it to match the Ruby version.

docs/_docs/installation/macos.md
```diff
@@ -3,155 +3,28 @@ title: Jekyll on macOS
...
+Older macOS versions might work, but we don't officially support them.
 
 ## Install Ruby
 
-Jekyll requires **Ruby v{{ site.data.ruby.min_version }}** or higher.
-macOS Big Sur 11.x ships with Ruby 2.6.3. Check your Ruby version using `ruby -v`.
-
-If you're running a previous version of macOS, you'll have to install a newer version of Ruby. Installation with [Homebrew](https://brew.sh) is simple if you're only planning to use Ruby for Jekyll. Install with a version manager such as [asdf](https://asdf-vm.com/), [chruby](https://github.com/postmodern/chruby), [rbenv](https://github.com/rbenv/rbenv), or [rvm](https://rvm.io/) if you need to switch among Ruby versions (instructions for rbenv are below). See the guide [Install Ruby on Mac](https://mac.install.guide/ruby/index.html) for details and recommendations.
-
-### With Homebrew {#brew}
```
monfresh, line 27: While Homebrew is an option for installing Ruby, I don't recommend mentioning it because eventually it will cause people confusion and headaches when they need to work with different projects that use different versions of Ruby.

Given that installing Ruby with a version manager is just as easy, and it has the additional benefit of supporting multiple versions at the same time, we should favor it. People prefer to be given clear instructions that will serve them in the long term, as opposed to making them choose between various options that they don't fully understand.

Furthermore, the Homebrew gems path that needs to be added to the `PATH` depends on the version of Ruby that was installed with Homebrew, which means these instructions below only work if someone installed Ruby 3.0.x. Sure, we could update these instructions to say to replace `3.0` with the first 2 digits of the Ruby version, but why make things more complicated than necessary?

docs/_docs/installation/macos.md
```diff
@@ -3,155 +3,28 @@ title: Jekyll on macOS
...
-## Install Command Line Tools
-To install the command line tools to compile native extensions, open a terminal and run:
+## Supported macOS versions
 
-```sh
-xcode-select --install
-```
+- Monterey (macOS 12)
+- Big Sur (macOS 11)
+- Catalina (macOS 10.15)
```
monfresh, line 10: These are the OSes officially supported by Homebrew, and since we're using Homebrew to install a version manager, we have to mirror what Homebrew supports.

docs/_docs/installation/macos.md
```diff
@@ -3,155 +3,28 @@ title: Jekyll on macOS
 permalink: /docs/installation/macos/
 ---
 
-## Install Command Line Tools
```
monfresh, line 6: Homebrew automatically installs the command line tools if they're not already there, and since we're installing Ruby via `chruby`, which in turn gets installed via Homebrew, this step is redundant.

mattr-:
Every single comment you've left on this PR is a criticism of what was here before. Why is that necessary? We can see for ourselves the changes you've made and if we have questions, we'll ask them.

Based on your attitude towards other contributors prior work, as well as the fact that this update links to your website rather than providing actual content, I'm closing this. I appreciate the time and effort that went into making these changes but I don't think this update is for us.

monfresh:
@mattr- I'm sorry my comments came off as criticisms. That was not at all the intention. Most of them are facts to provide the reasoning behind removing those sections. When I make changes, I like to provide details that explain why. I was just trying to be thorough rather than just removing a lot of content without explanation.

As for linking to my website, that's because @ashmaroli specifically asked me to here: https://github.com/jekyll/jekyll/issues/8784#issuecomment-1060267860

He specifically said to just link to my site:

> You need not port the entire contents of your blog post to our documentation, just a section-heading, a brief preface and a link to your blog post will do. (Yes, do note the navigation to an external third-party site (your blog)).

The reason why he asked me to do this is because my guide has helped a lot of people who have been stuck trying to install Jekyll on Mac.

ashmaroli:
> as well as the fact that this update links to your website rather than providing actual content, I'm closing this.

@mattr- The author opened this pull request as a response to my request https://github.com/jekyll/jekyll/issues/8784#issuecomment-1060267860 

So, I will shoulder the blame for that action apologize for making that call.

mattr-:
> When I make changes, I like to provide details that explain why. I was just trying to be thorough rather than just removing a lot of content without explanation.

I appreciate the attention to detail. In the future, it's more beneficial to comment on the additions (IOW the green bits in a PR diff), rather than the subtractions (the red parts). This is a small thing but it makes a big difference because the comment is on your change, why it's there, and it makes the comments about your work, rather than someone else's work.

Also, I'm going to apologize because I didn't pay close enough attention to the words in the PR description and the linked comments.

If we can fix the issues with the spelling action, let's reset things and review this a bit more objectively.

monfresh:
The spell check is complaining about my name `moncefbelyamani` and `preinstalled`. The latter is a valid word according to the Dictionary app on my Mac.

What is the proper file to add these exceptions to? Is it `.github/actions/spelling/expect.txt`?

monfresh:
I added those exceptions to `expect.txt` and the Spell Check action is passing now.

monfresh:
Also, thanks for reopening the PR and your constructive criticism. I can see how my changes came off with the wrong tone. I apologize. I'm really passionate about this topic and can get carried away sometimes. 

I thought about it some more, and I think I have a good middle ground. I propose to add the basic instructions for installing Ruby with `chruby`, similar to the current `rbenv` instructions, and then link to my more comprehensive guide for further setup and details, and for those who might have already tried to set up a dev environment but are running into issues.

One of the reasons I favor `chruby` is because you can use gems right after installing them, and it installs gems separately in each version of Ruby. Whereas with `rbenv`, you sometimes have to restart the terminal (or open a new tab) for things to be recognized. This is a common source of confusion. 

For example, after installing a Ruby version and setting it with `rbenv global`, it doesn't always recognize it right away. Similarly, after installing a gem like Rails, `rails -v` doesn't work right away; you need to restart the terminal. With `chruby`, this isn't an issue that I can recall. This doesn't seem to be an issue when installing Jekyll with `rbenv`, but I prefer to recommend tools that will work consistently across various gems so people aren't surprised.

monfresh:
@mattr- I just pushed a commit that adds instructions for installing Ruby with chruby, and I also added a note about adding `webrick` if using Ruby 3.x, which is another common issue people run into.

I think it looks much better now than what I originally had. I'd love to hear your thoughts.

docs/_docs/installation/macos.md
```diff
@@ -3,156 +3,84 @@ title: Jekyll on macOS
...
-echo 'export PATH="$HOME/.gem/ruby/X.X.0/bin:$PATH"' >> ~/.bash_profile
+Next, read that same external guide for important notes about [setting and switching between Ruby versions with chruby](https://www.moncefbelyamani.com/how-to-install-xcode-homebrew-git-rvm-ruby-on-mac/#how-to-install-different-versions-of-ruby-and-switch-between-them).
 
-# Unsure which shell you are using? Type
-echo $SHELL
-```
+## Install Jekyll
 
-Check that `GEM PATHS:` points to your home directory:
+After installing Ruby with chruby, install Jekyll and Bundler:
```
ashmaroli, line 69: Lets not recommend installing `bundler` since it is available with Ruby installations nowadays and Jekyll doesn't really need the *latest* version of Bundler either.
```suggestion
After installing Ruby with chruby, install latest Jekyll gem:
```

docs/_docs/installation/macos.md
```diff
@@ -3,156 +3,84 @@ title: Jekyll on macOS
...
 
 ```sh
-gem env
+gem install bundler jekyll
```
ashmaroli, line 72: See above comment.

```suggestion
gem install jekyll
```

docs/_docs/installation/macos.md
```diff
@@ -3,156 +3,84 @@ title: Jekyll on macOS
...
-{: .note .info}
-Every time you update Ruby to a version in which the first two digits change, update your path to match.
-
-### Global Install
-
-{: .note .warning}
-We recommend not installing Ruby gems globally to avoid file permissions problems and using `sudo`.
-
-#### On Mojave (10.14)
+### Note about Jekyll and Ruby 3.x
```
ashmaroli, line 75: Let us drop this section entirely.
The docs already have this covered in a different page. No need to reinstate here, mostly because `webrick` gem will be installed automatically as a Jekyll dependency from the next minor version release onwards.

ashmaroli (commented):
If you're able to edit this page on your local system, I recommend limiting line-lengths to 120 chars so that text-editors that do not soft wrap lines do not force extended horizontal scrolling.

monfresh:
Thanks for the feedback. I just pushed the changes. I typically set the line length to 80 characters for readability, but I just realized that my Sublime Text settings were just wrapping at 80 characters, and not actually setting each line to 80 chars.

If you prefer 120, I'm happy to make the lines longer.

docs/_docs/installation/macos.md
```diff
@@ -3,156 +3,84 @@ title: Jekyll on macOS
...
-
-### Local Install
-
-Install the bundler and jekyll gems:
+This will take a few minutes, and once it's done, configure your shell to automatically use `chruby`:
 
 ```sh
-gem install --user-install bundler jekyll
+echo "source /usr/local/opt/chruby/share/chruby/chruby.sh" >> ~/.zshrc
+echo "source /usr/local/opt/chruby/share/chruby/auto.sh" >> ~/.zshrc
```
mattr-, line 65: What do you think about using `brew --prefix` here instead?
```suggestion
echo "source $(brew --prefix)/opt/chruby/share/chruby/chruby.sh" >> ~/.zshrc
echo "source $(brew --prefix)/opt/chruby/share/chruby/auto.sh" >> ~/.zshrc
```

monfresh: Yes! I had the same thought but forgot to make the change. I'll also remove the next line.

ashmaroli:
> If you prefer 120, I'm happy to make the lines longer.

There is no hard preference. Our source already contains documents with lines at 80 chars.
So, all good.

monfresh:
Sounds good. Any other changes you'd like me to make? And should I squash the commits?

ashmaroli (approved):
I'm not a macOS user. Therefore, will take this at face-value.

ashmaroli:
> And should I squash the commits?

The commits will be squashed automatically at the time of merging.

monfresh:
Hello. Any idea when this might be merged?

mattr- (approved):

mattr-:
@jekyllbot: merge +docs

mattr-:
Sorry! I thought this was merged already after the first approval.

monfresh:
No worries. Thanks for merging!

bamr87:
Have these new instructions been tested? I followed these new steps on a new install of Monterey 12.4 and it took forever to install and still didn't work...

monfresh:
@bamr87 I think the issue might be with this line in the instructions, which is not up to date with the latest version of Ruby:
```
echo "chruby ruby-3.1.1" >> ~/.zshrc
```
Try this instead:

```
echo "chruby ruby-3.1.2" >> ~/.zshrc
```

To verify if `chruby` was properly installed, try running `chruby`. It should show all versions that are present on your machine. You can switch to one of those versions like this:

```
chruby 3.1.2
```
Then when you do `ruby -v`, it should show 3.1.2

I will update the version in the Jekyll docs.

I keep [my tutorial](https://www.moncefbelyamani.com/how-to-install-xcode-homebrew-git-rvm-ruby-on-mac/) up to date, and I will try to remember to update the Jekyll docs as well.

bamr87:
@monfresh - Thanks for the info. I should have paid more attention to the version, and that seemed to have been the issue. 

It would be helpful to note in these instructions that the zsh profile should be updated for every new version. 

Also, it might be worth noting that the hostname of a user's computer will also change. I don't know exactly why, but my first run through updated my hostname to the computer ID, and this second run through updated it to something slightly different. See screenshot.

image

monfresh:
Yes, I will make a note in the instructions about updating the shell profile. Thanks for the suggestion!

These kinds of issues are the reason why I built the [Ruby on Mac](https://www.rubyonmac.dev) script that takes care of everything for you. It automatically updates the shell profile (whether you're using zsh, bash, or fish) with whatever version got installed so that they are always in sync. And it comes with a `rom` alias that you can run regularly to keep everything up to date.

As for the hostname, I very much doubt it has anything do with installing Ruby. This is the first time in 10 years of helping people install Ruby on their Mac that I've heard of this issue.

PR #9008, opened 2022-03-24
Enhance `script/default-site` to increase coverage

ashmaroli:
## Summary

Primarily, build the default site in `safe` mode as well.
Secondarily, improve readability of log messages by adding some newlines and colorizing step labels.

### DIFF
--- a/lib/jekyll/entry_filter.rb
+++ b/lib/jekyll/entry_filter.rb
@@ -82,7 +82,7 @@ module Jekyll
     #   since we use Pathutil#in_path? now.
     # --
     def symlink?(entry)
-      site.safe && File.symlink?(entry) && symlink_outside_site_source?(entry)
+      site.safe && symlink_outside_site_source?(entry)
     end
 
     # --
--- a/lib/site_template/_config.yml
+++ b/lib/site_template/_config.yml
@@ -33,6 +33,10 @@ github_username:  jekyll
 theme: minima
 plugins:
   - jekyll-feed
+  - jekyll-seo-tag
+whitelist:
+  - jekyll-feed
+  - jekyll-seo-tag
 
 # Exclude from processing.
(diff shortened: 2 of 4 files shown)


jlachlan:
Don't use any custom fonts like "Sticks Unicode" people are using fonts to parse the css auto size instead of writing code to parse tables based on linguistics or knowledge of database. I'm seeing Linux Dameon based fonts sized to move data between logic and it doesn't work... it just is endless error correction on mx 

Can someone like remove the random fonts that firmware dependent from the containers ... idk all the Apache font or history of all fonts ever used in html instead of using html to code the font... like why are we using fonts in css and html and not headers and document identifiers ? It's just loading firmware to root not actually containing anything !

jlachlan:
Is it so the UI is more fun on the coder space? Delete custom fonts from your terminals bc spacing is bits and Boolean when transmitting and receding crosstalk your gonna get decoherence from the signals and false errors from the code and then just write code below to do what thing before was supposed to do. 

Can Apache 3.0 have its own font we buy from Adobe for $20 ! 

Think that would solve some issues if we use adobe font in adobe or oracle in SQL... 

Overlay rendering of fonts via layers like maps don't change the base layer of the system. 

It's not even like we have to type on those 10 inch matrix green font terminals. Even in excel no one writes the formatting in the data. How?? 

No more bold Asus Stick dependent motherboard fonts loading firmware in containers in a word doc. 

GitHub I'm talking to you... no more Robinhood slock machine stuff in the coding UI. It's boring bc it's a nerd version of legal docs they have line numbers and logic too. 

These containers would be epic for legal work if they actually worked to switch between referenced laws which have very easy to lookup numbers and are already organized.

Why are we making apps that organize our virtual desk with fake post it's instead of cars that argue your speeding tickets and drive you around and order your food 😂

PR #9011, opened 2022-03-26
Move JSON schema to Jekyll repo

EmilySeville7cfg:
[JSON schema][schema] is moved from schemastore.

[schema]: https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/jekyll.json

### DIFF
--- a/schemas/jekyll.json
+++ b/schemas/jekyll.json
@@ -1657,5 +1657,5 @@
       },
       "additionalProperties": false
     }
-  },
+  }
 }
--- a/schemas/jekyll.json
+++ b/schemas/jekyll.json
@@ -1,0 +1,1662 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$comment": "https://jekyllrb.com/docs/configuration/",
+  "title": "Jekyll static site generator config file schema",
+  "definitions": {
+    "global-permalink": {
+      "description": "The global permalink format\nhttps://jekyllrb.com/docs/permalinks/#global",
+      "type": "string",
+      "default": "date",
+      "examples": [
+        "/:year",
+        "/:short_year",
+        "/:month",
+        "/:i_month",
+        "/:short_month",
+        "/:day",
+        "/:i_day",
+        "/:y_day",
+        "/:w_year",
+        "/:week",
+        "/:w_day",
+        "/:short_day",
+        "/:long_day",
+        "/:hour",
+        "/:minute",
+        "/:second",
+        "/:title",
+        "/:slug",
+        "/:categories",
+        "/:slugified_categories",
+        "date",
+        "pretty",
+        "ordinal",
+        "weekdate",
+        "none",
+        "/:categories/:year/:month/:day/:title:output_ext",
+        "/:categories/:year/:month/:day/:title/",
+        "/:categories/:year/:y_day/:title:output_ext",
+        "/:categories/:year/:week/:short_day/:title:output_ext",
+        "/:categories/:title:output_ext"
+      ],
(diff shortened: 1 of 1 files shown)


ashmaroli:
@EmilySeville7cfg Please clarify your intention / motivation for this pull request.
What are we going to achieve with this..? How is the schema going to be consumed?

EmilySeville7cfg:
My motivation is to enable hints for Jekyll site config file. The first thing user need to perform is: install [`Yaml`][yaml] extension. After it when config file is created hints are shown via `Ctrl`+`Space`:

image

I think it's better to place this schema here because Jekyll developers knows better how Jekyll must work. I can fix schema errors and enhance it, maintainers can assign such issues for me. ;)

[yaml]: https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml

ashmaroli:
Hmm. I feel this should be a separate project and has no legit place inside **jekyll** gem which is an archive of Ruby code.
I'll at other maintainers to weigh in before reaching a decision.

EmilySeville7cfg:
I don't think it must be put inside gem. 😄 Schema can exist just in repo.

ashmaroli:
> I don't think it must be put inside gem. smile Schema can exist just in repo.

Our end-users consume Jekyll using the gem. They're not exposed to the repository until they manually clone this repo.
So how is Visual Studio Code going to detect / use the schema then?

EmilySeville7cfg:
> So how is Visual Studio Code going to detect / use the schema then?

It is not a problem because:

- schema stored here
- reference to it is stored in schemastore repo

It doesn't matter where schema is stored - it just must be accessible to be consumed by VS Code YAML extension.

mattr-:
Is schemastore not meant to be the authoritative version of the schema? I was pretty excited to see that PR merged there. I'm not sure what the guidelines are there.

EmilySeville7cfg:
> Is schemastore not meant to be the authoritative version of the schema? I was pretty excited to see that PR merged there. I'm not sure what the guidelines are there.

We prefer to place schemas for the specific repos (like this one) instead of storing everything in one place. In cases when it is not possible we store schema in schemastore.

ashmaroli:
Thanks for the additional info, @EmilySeville7cfg

However, **"I"** still think the schema should be a separate project.
Maintaining it is outside the scope of this repo even if *you* were to be made a Jekyll Core maintainer.

*Some* of the config keys are recent additions and therefore may confuse users. For example, `ignore_theme_config` and certain *permalink* placeholders introduced in Jekyll 4.0 have no relevance for a user deploying via GitHub Pages which is locked to Jekyll 3.
Having the schema in a separate project would allow:
- hosting separate schema(s) for supported version(s).
- develop scripts to automate generation of schema by parsing / running a *released* Jekyll rubygem.
- develop extension(s) to VSC that would parse user's `Gemfile` / `Gemfile.lock` to load / enable relevant Jekyll schema.

EmilySeville7cfg:
For several tasks (not just supporting JSON schema) I agree - a separate repo is better.

mattr-:
I'd rather store the schema with the code so that we don't have to go multiple places to make things line up in the future should we change the configuration schema or another schema that we add at a later date. I'm not sure how projects like schemastore manage different versions of schema based on the version of the project being used, but I'm not sure we need to solve that problem.

EmilySeville7cfg:
> I'm not sure how projects like schemastore manage different versions of schema based on the version of the project being used, but I'm not sure we need to solve that problem.

It specifies `version` mappings inside `version` key in `catalog.json`:

```json
    {
      "name": "AnyWork Automation Configuration",
      "description": "AnyWork Automation Configuration used to configure automation scripts on AnyWork",
      "fileMatch": [
        ".awc.yaml",
        ".awc.yml",
        ".awc.json",
        ".awc.jsonc",
        ".awc"
      ],
      "url": "https://json.schemastore.org/anywork-ac-1.1.json",
      "versions": {
        "1.0": "https://json.schemastore.org/anywork-ac-1.0.json"
      }
    },
```

`url` is the version used by default.

PR #9012, opened 2022-03-26
Release Jekyll v3.9.2

parkr:
Headline feature is Ruby 3.x support for the Jekyll 3.9.x series!


ashmaroli (approved):
Not sure if we should highlight 3.x compatibility when the user will still have to manually add `webrick` to their Gemfile.
Also, the release post has a **future timestamp** from *now*. Not an issue if the PR is merged at or after that time.

Fukhemet (commented):
Thanks

PR #9013, opened 2022-03-26
Add Jekyll 3.9.2 release post to 'master' branch

parkr:
This publishes the release post from #9012 to the `master` branch so that it can be published to jekyllrb.com


ashmaroli (approved):
Not sure if we should highlight 3.x compatibility when the user will still have to manually add `webrick` to their Gemfile.
**Caution:**, the release post has a **future timestamp** from *now*. Won't be an issue if the PR is merged at or after that time, though else Jekyll won't render the release post.

parkr:
@jekyllbot: merge +docs

PR #9015, opened 2022-03-28
Check symlink outside site_source without Pathutil

ashmaroli:

## Summary

`pathutil` gem appears to be unmaintained.
Therefore, emulate its existing use in `lib/jekyll/entry_filter.rb` using Ruby's built-in methods.

## Test Coverage

Relies on existing test defined as: 
https://github.com/jekyll/jekyll/blob/e695c1e24b298994ecb84daa74fba3d20959cba9/test/test_entry_filter.rb#L89-L93

### DIFF
--- a/lib/jekyll/entry_filter.rb
+++ b/lib/jekyll/entry_filter.rb
@@ -86,14 +86,10 @@ module Jekyll
     end
 
     # --
-    # NOTE: Pathutil#in_path? gets the realpath.
-    # @param [<Anything>] entry the entry you want to validate.
-    # Check if a path is outside of our given root.
+    # Check if given path is outside of current site's configured source directory.
     # --
     def symlink_outside_site_source?(entry)
-      !Pathutil.new(entry).in_path?(
-        site.in_source_dir
-      )
+      !File.realpath(entry).start_with?(site.in_source_dir)
     end
 
     # Check if an entry matches a specific pattern.


parkr (approved):

lib/jekyll/entry_filter.rb
```diff
@@ -86,14 +86,10 @@ def symlink?(entry)
...
-    # NOTE: Pathutil#in_path? gets the realpath.
-    # @param [<Anything>] entry the entry you want to validate.
-    # Check if a path is outside of our given root.
+    # Check if given path is outside of current site's configured source directory.
     # --
     def symlink_outside_site_source?(entry)
-      !Pathutil.new(entry).in_path?(
-        site.in_source_dir
-      )
+      !File.realpath(entry).start_with?(site.in_source_dir)
```
parkr, line 92: One alternative here might be to check if a constructed path and the real path are not equal?

```ruby
site.in_source_dir(entry) != File.realpath(entry)
```

parkr (commented):
LGTM, I thought of an alternative for you to consider but your original solution is likely still best.

lib/jekyll/entry_filter.rb
ashmaroli: The downside to this is that it results in consumption of additional resources.
`site.in_source_dir(entry)` is more expensive (both in terms of allocations and cpu) than `site.in_source_dir`.

A *better* alternative to *my original proposal here* would be to check directly against `site.source` instead:
```suggestion
      !File.realpath(entry).start_with?(site.source)
```
But since `EntryFilter#symlink_outside_site_source?` method is mostly invoked *only if* `site.safe && File.symlink?(entry)` returns `true`, the difference between `site.source` and `site.in_source_dir` is negligible.

parkr: Sounds good. :shipit:

ashmaroli:
@jekyllbot: merge +dev

PR #9024, opened 2022-04-03
JSON Schema hints for _data files

EmilySeville7cfg:
## Summary

This PR adds JSON Schema hints and validation for _data files.

image

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,13 @@
+{
+  "yaml.schemas": {
+    "schemas/config_options/options.json": "docs/_data/config_options/*.yml",
+    "schemas/docs_nav.json": "docs/_data/docs_nav.yml",
+    "schemas/jekyll_filters.json": "docs/_data/jekyll_filters.yml",
+    "schemas/jekyll_variables.json": "docs/_data/jekyll_variables.yml",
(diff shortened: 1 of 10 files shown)


ashmaroli:
Okay, since you've already proposed `schemas/jekyll.json` for general use in a different PR, it would be best if you organize the current set of schemas under `schemas/jekyllrb.com/` for ease of maintenance.
Also, please add a README to that folder detailing how one should proceed with when creating a new data file for `jekyllrb.com`

ashmaroli:
@EmilySeville7cfg Will you be able to write a script to automate generation / modification of these schemas?
(Not an issue if you're not comfortable coding).

EmilySeville7cfg:
I can write a script. Is it right that I need to generate valid JSON schema from .yml configs?

ashmaroli:
> I can write Bash/Python script.

Awesome! Bash will be sufficient.
You may place the script inside directory `script` and make a note of it in the Readme at `schemas/jekyllrb.com/`.


> Is it right that I need to generate valid JSON schema from .yml configs

Not sure I understand the ".yml configs" you're referring to, but yes the data files inside `docs/` are in YAML and any new ones in future will be in YAML (because its the easiest).

PR #9027, opened 2022-04-09
Update WSL instructions

rtfmoz2:
To save many hours of tracking down version dependancies.


rtfmoz2:
Tested with against a fresh install of Ubuntu under WSL2 after sudo apt-get update & upgrade.

docs/_docs/installation/windows.md
```diff
@@ -49,23 +49,30 @@ Your terminal should now be a Bash instance. Next, update your repository lists
...
+source ~/.$(basename $SHELL)rc
+```
 Next, install Ruby. To do this, let's use a repository from [BrightBox](https://www.brightbox.com/docs/ruby/ubuntu/),
 which hosts optimized versions of Ruby for Ubuntu.
 
 ```sh
 sudo apt-add-repository ppa:brightbox/ruby-ng
 sudo apt-get update
-sudo apt-get install ruby2.5 ruby2.5-dev build-essential dh-autoreconf
+sudo apt-get install ruby ruby-dev build-essential dh-autoreconf
```
ashmaroli, line 65: Would be best to add a comment here to note what Ruby version will be installed.
I don't think BrightBox supports Ruby 3.0 and newer..
What version got installed in your test images?

docs/_docs/installation/windows.md
```diff
@@ -49,23 +49,30 @@ Your terminal should now be a Bash instance. Next, update your repository lists
...
+sudo apt-get install libssl-dev zlib1g-dev libreadline-dev libffi-dev
```
ashmaroli, line 66: I understand why `libffi-dev` would be needed but please clarify why other packages would be needed.

ashmaroli (commented):
Needs some minor clarification.

*P.S. You need not squash or rebase your commits. It'll be automatically handled during merge.*

docs/_docs/installation/windows.md
rtfmoz2: installs ruby 2.7, will add comment

docs/_docs/installation/windows.md
rtfmoz2: Gem update fails without the packages added.
zlib fails to update without dev package added.
readline has the the same issue. Without dev package gem update cannot complete.

erikkjellgren:
Shouldn't the WSL2 (assuming Ubuntu) installation be exactly the same as the Ubuntu installation https://jekyllrb.com/docs/installation/ubuntu/ ?

I am using WSL2, I have just been using the Ubuntu installation for the WSL2. The behaviour I have encountered is exactly the same as installing on native ubuntu.

PR #9038, opened 2022-04-20
Improve documentation on generator-based plugins

### LINKED ISSUE #9037: [Docs]: SamplePlugin Has Many Major Errors
## Motivation

The `SamplePlugin::CategoryPageGenerator` class from the [`Generator` example page](https://jekyllrb.com/docs/plugins/generators/) is badly, totally, completely broken.

## Suggestion

Either fix it or delete it.

  * super is not called
  * Most of the variables defined have no purpose

ashmaroli:

## Context

Resolves #9037
/cc @mslinn

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -434,6 +434,7 @@ nodejs
 noifniof
 nokogiri
 notextile
+onboarding
 onclick
 onebox
 oneclick
--- a/docs/_docs/plugins/generators.md
+++ b/docs/_docs/plugins/generators.md
@@ -79,10 +79,7 @@ The following example is a more complex, advanced-level generator that generates
 as a whole, it would be best (but not a requirement) to be acquainted with the implementation of `Jekyll::Page` before proceeding ahead.
 
 In this example, the aim of the generator is to create **a page** for each category registered in the `site`. The pages are virtual i.e.
-created at runtime. So their content, front matter and other attributes ideally need to be designed by the plugin itself. However, since a
-plugin cannot determine the presentational aspects of the render (the desired HTML) which is primarily dictated by the site's theme, the
-example plugin will only provide the backend data (`page.linked_docs`) that is expected to be consumed by the user's Liquid template.<br/>
-In other words, *the generated pages will be empty* unless linked to a layout (which is outside the scope of current document.)
+created at runtime. So their content, front matter and other attributes ideally need to be designed by the plugin itself.
 * The pages are *intended* to render a list of all documents under a given category. So the crux of the generator will be to loop through
 `site.categories` and initialize new *objects*, one per registered category.
 * Having the ability to configure the pages via [front matter defaults](/docs/configuration/front-matter-defaults/) would be awesome! So
--- a/docs/_docs/plugins/generators.md
+++ b/docs/_docs/plugins/generators.md
@@ -90,9 +90,10 @@ module SamplePlugin
   class CategoryPageGenerator < Jekyll::Generator
     safe true
 
-    # For every `category` registered under current site, *generate* an instance of
-    # `SamplePlugin::CategoryPage` and store them inside `site.pages` which is an array
-    # of all standalone pages in the current site.
+    # For every `category` registered under current site, *generate* an
+    # instance of `SamplePlugin::CategoryPage` and store them inside
+    # `site.pages` which is an array of all standalone pages in the
+    # current site.
     def generate(site)
       site.categories.each do |category, documents|
         site.pages << CategoryPage.new(site, category, documents)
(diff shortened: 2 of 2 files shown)


ashmaroli:
@mslinn You may access the rendered preview at https://deploy-preview-9038--jekyllrb.netlify.app/docs/plugins/generators/

wU7700n890 (approved):
Ok

mslinn:
Wow, that happened fast. I have been reviewing the pull request since it was created less than an hour ago. I would like to offer feedback, is there any chance my feedback could be incorporated at this point?

ashmaroli:
> is there any chance my feedback could be incorporated at this point?

Yes. The approval you're seeing can be disregarded.

mslinn:
Thank you for taking swift action. I am pleasantly surprised.

These are my initial comments. Some of them exceed the scope of this pull request and should be handled as separate topics. I put it all together here for context.

## Sample Program Comments

  * The program should be given a name for convenience. I call it `category_generator.rb`, but that name is arbitrary.
  * When I tried the `category_generator.rb` plugin, it generated many directories, all with one file within (`index.html`). Instead, the program should either:
    * Generate one directory with many files in it, plus an `index.html`.
    * Generate one file, with one subhead for every category.
  * All the generated `index.html` files are empty, so this plugin is a no-op.
  * `safe` is well documented as a CLI flag, but there is no user-level mention of the `safe` property defined by [`Jekyll::Plugin`](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/plugin.rb#L53-L63). The source code comments are not helpful to the reader (`safety` is not mentioned anywhere, and is therefore undefined): 
    ```
    # Get or set the safety of this plugin. When called without an argument
    # it returns the safety. When an argument is given, it will set the
    # safety.
    #
    # safe - The Boolean safety (default: nil).
    #
    # Returns the safety Boolean.
    ```
    The following should be added to `plugin.rb` comments and user-level documentation:
    * What is this property for, and does it have any relationship with the cli flag of the same name?
    * What are best practices for using this property?
    * When should this property be set, and when should it not be set?
    * What consequences might arise through misuse of this property?
  * Instantiating a new `Page` by setting variables is inherently fragile. Clearly, `Page` was not intended to be exposed to the public because:
    *  It has no module- or class-level documentation
    * Implementation details are completely exposed, hence the redundant variable settings required. If the implementation changes, for example a variable is renamed, plugins built this way would all break. 
  * It would be much better to use a constructor method to create a `Page`. `Page` should be properly commented, and at a minimum, the setter for `name` should be rewritten to also set the values of `basename`, `ext` and `url_placeholders`. This would provide a more sensible interface for outsiders to manipulate `Page`s.
  * Putting a trailing comma after `'linked_docs' => posts,` would encourage the use of that writing style.
  * An example of an include that uses `linked_docs`, showing source code, expanded HTML and the rendered page would provide a more complete example. Alternatively, that code should be deleted.
  * `data.default_proc` is an advanced concept that is introduced without motivation, background documentation or working examples. If this is worth mentioning, and I think it is, then it should be discussed properly; alternatively, the code should be deleted.
  * The comment "`Look up front matter defaults scoped to type categories`" leads to a discussion about scopes. This topic is not sufficiently well documented in the Jekyll documentation; however, this example attempts to build on those concepts. If not properly explained, this code should be deleted.
  * `"Override the URL template used in `Jekyll::Page` to suit our use-case."` - no use case is provided. 
  * The `url_placeholders` hash of properties overrides the defaults set in [`page.rb`](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/page.rb). The comment in that file (`Returns a hash of URL placeholder names (as symbols) mapping to the desired placeholder replacements. For details see "url.rb"`) increases the cognitive load for a plugin writer. Since it appears that plugin writers are expected to be familiar with the source code in `page.rb` (although that is never actually stated in black-and-white), that file should be well commented; the purpose of this hash should be expressed in a straightforward manner.

## Documentation Comments
  * The example should show one of the generated files.
  * The term "use case" is not usually hyphenated.
  * Since this example requires the leveraging of [`page.rb`](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/page.rb), providing a link to the source for that class would help the reader.
  * "Having the ability to configure the pages via [front matter defaults](https://github.com/jekyll/jekyll/blob/6b4db589c90071da1894ecfd697a1c646e6a8e27/docs/configuration/front-matter-defaults) would be awesome! So assigning a particular type to these pages would be beneficial." Why is this true? What benefits would be provided? The explanation at the end ("The generated pages if set up to be managed via front matter defaults, the following example will assign a layout to all generated pages with type, :categories:") does not provide a compelling reason to write code in such an oblique manner. Why not just set `data['layout'] = "category_page"`? Seems like unnecessary complexity, again increasing the reader's cognitive load for no benefit.

mslinn:
```
Implementing this
# allows the user to alter front matter aspects (e.g. `layout`,
# `permalink`, etc) from the user's config file without editing
# the plugin code simply by setting the `scope.type` to
# `categories`.
```

Readers would be best served by showing them a snippet of `_config.yml` that illustrates exactly what you mean. Easy for you, not so easy for some others. 

Even better: apply the `default` layout to the generated files, so they won't be empty. Most Jekyll sites have a layout called `default`, because the out-of-the-box experience leads them down that path. Once a potential plugin programmer sees empty `index.html` files, they will likely become less motivated to consider generators. Why? Because the example does not illustrate business value. 

Consider how a programmer might have to justify to their [pointy-haired-boss](https://en.wikipedia.org/wiki/Pointy-haired_Boss) why they should spend time learning to write Jekyll generators... only to encounter a demo that provides no business value. First-time readers likely lack the ability to add content to the generated pages. As you have said in so many words, the scope of this example is not comprehensive, and it would be impossible to attempt to achieve a completely comprehensive example. We need to help budding plugin developers make the case to their superiors why learning this technology would be a good investment. Producing technical documentation is actually an aspect of marketing.

In general, reducing barriers to entry increases technology adoption by enhancing the joy of working with the technology to quickly achieve business goals, instead of struggling against unnecessarily high cognitive load for what appears to be (on the surface) no additional business value. 

**Tech docs should be a welcoming and supportive experience.**

Please populate the generated files with content.

ashmaroli:
> The plethora of directories persists. Now they all live under `categories/`. Those intermediate directories should not exist..

@mslinn The directories are dependent on your permalink template.
That is outside the scope of the discussion here and am going to ignore it.

PR #9079, opened 2022-06-22
Add note for Ruby versions 3.0 and higher

erikhowe:
When running Jekyll serve for the first time, users who are using Ruby 3.0 or higher will receive an error message. If they copy the first line of the trace, they'll find a Github issue where the solution is the first post (https://github.com/jekyll/jekyll/issues/8523), but it would improve developer experience if the requirement were part of the tutorial


ashmaroli:
Closed as no longer valid.

PR #9098, opened 2022-08-04
Fixed xz dependency

Neurone:

## Summary

While installing `ruby`, `ruby-install` depends on `xzcat` command. If `xz` is not installed, you get an error when you launch `ruby-install ruby`. This PR fixes that.

## Context

N/A


zxz054321 (approved):
Same experience. `brew install chruby ruby-install xz` will fix the problem.

parkr (approved):

ashmaroli:
Thanks @Neurone

PR #9132, opened 2022-09-23
Support jekyll-sass-converter 3.x

ntkme:

## Summary

jekyll-sass-converter 3.x release will come soon. This PR relaxes version requirements so that users will be able to choose between 2.x and 3.x. 

On API level there is no change, thus it remains compatible with current jekyll.

However, there are a few behavior changes espeically in regard to how sass import works that may break a small fraction of users if they didn't lock dependency version: https://github.com/jekyll/jekyll-sass-converter#migrate-from-2x-to-3x

## Context

https://github.com/jekyll/jekyll-sass-converter/pull/140


ashmaroli:
Thanks @ntkme
@jekyllbot: merge +minor

PR #9135, opened 2022-09-28
Unlock Psych dependency

### LINKED ISSUE #9122: Update dependency constraint to allow for psych v4.0.5
Hey there! :wave:

I noticed that the constraint you have for psych doesn't allow for the latest version to be used.

The constraint I found was `~> 3.3`, and the latest version available is `4.0.5`.

yboulkaid:
## Summary

In https://github.com/jekyll/jekyll/pull/8918 we locked the `psych`
dependency because the new version shipped with ruby 3.1 was breaking
our cucumber-based tests.

This PR allows for us to relax that dependency by making our tests use
`SafeYAML` instead of `YAML`.

## Context

We are using `YAML.load`, which works in psych 3.x but raises a
`Psych::DisallowedClass` when trying to parse a date with psych v4.x.

This PR fixes this by making our tests use `SafeYAML` instead of `YAML`,
as we are doing in the rest of the codebase.

Closes #9122.

### DIFF
--- a/features/step_definitions.rb
+++ b/features/step_definitions.rb
@@ -164,7 +164,6 @@ Given(%r!^I have a configuration file with "(.*)" set to "(.*)"$!) do |key, valu
       {}
     end
   config[key] = SafeYAML.load(value)
-
   Jekyll.set_timezone(value) if key == "timezone"
   File.write("_config.yml", YAML.dump(config))
 end
--- a/features/step_definitions.rb
+++ b/features/step_definitions.rb
@@ -163,12 +163,7 @@ Given(%r!^I have a configuration file with "(.*)" set to "(.*)"$!) do |key, valu
     else
       {}
     end
-
-  if Gem::Version.new(Psych::VERSION) >= Gem::Version.new("4.0.0")
-    config[key] = YAML.load(value, permitted_classes: [Date])
-  else
-    config[key] = YAML.load(value)
-  end
+  config[key] = SafeYAML.load(value)
 
   Jekyll.set_timezone(value) if key == "timezone"
   File.write("_config.yml", YAML.dump(config))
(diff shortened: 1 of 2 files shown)


features/step_definitions.rb
```diff
@@ -163,7 +163,13 @@
     else
       {}
     end
-  config[key] = YAML.load(value)
+
+  if Gem::Version.new(Psych::VERSION) >= Gem::Version.new("4.0.0")
+    config[key] = YAML.load(value, permitted_classes: [Date])
+  else
+    config[key] = YAML.load(value)
+  end
```
yboulkaid, line 171: As suggested in the last part of the PR description, this could also be:

```suggestion
  config[key] = YAML.unsafe_load(value)
```

ashmaroli:
At the outset, our test suite only flagged the use of `YAML.load` in our Cucumber helper, but what if allowing psych 4 breaks some plugins downstream?
The Jekyll Core uses SafeYAML extensively. So what if we were to switch the use of `YAML.load` with its SafeYAML counterpart?

yboulkaid:
I like the idea of using SafeYAML everywhere, I'll try to have a patch out later today 👍

yboulkaid:
I have now updated the PR to use `SafeYAML` rather than depend on psych 👍 

@fauno I would say that this is not directly related to #8772 as it is smaller in scope. It does however affect it as it introduces a new place where `SafeYAML` has to be swapped for psych's `YAML`.

yboulkaid:
CI is failing on ruby 2.6 Windows with:

```
Failure:
TestSite#test_: configuring sites should configure cache_dir.  [C:/projects/jekyll/test/test_site.rb:81]
Minitest::Assertion: Expected false to be truthy.
Skipped:
```

I can't tell if this is a flake or if it's related to this though? 🤔

ashmaroli (approved):
LGTM! :+1:

ashmaroli:
Thanks @yboulkaid 
@jekyllbot: merge +fix

PR #9125, opened 2022-09-19
Bump RuboCop to `1.36.x`

ashmaroli:
Additionally configure some new cops.

Closes #9117

PR #9136, opened 2022-09-28
Disable pending cops when running rubocop

yboulkaid:
## Summary

In #9125 we bumped the rubocop version to 1.36, which introduces new
cops. Unconfigured cops output a warning when running `script/fmt`,
which was noisy.

This PR makes rubocop run with the `--disable-pending-cops` option
to make the script output cleaner.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -126,6 +126,8 @@ Lint/NestedPercentLiteral:
     - test/test_site.rb
 Lint/NoReturnInBeginEndBlocks:
   Enabled: true
+Lint/NonAtomicFileOperation:
+  Enabled: false
 Lint/NumberedParameterAssignment:
   Enabled: true
 Lint/OrAssignmentToConstant:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -374,7 +374,7 @@ Style/GuardClause:
 Style/HashAsLastArrayItem:
   Enabled: true
 Style/HashConversion:
-  Enabled: true  
+  Enabled: true
 Style/HashEachMethods:
   Enabled: true
 Style/HashExcept:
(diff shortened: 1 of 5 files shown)


ashmaroli:
I am on the fence regarding this.

I refrained from enabling certain cops (ones concerned with patterns and numbered parameters) in the aforementioned pull request because they were related to Ruby 3 syntax which Jekyll has not yet adopted.
The remaining ones revolving minitest and rspec seemed noisy.

yboulkaid:
> I refrained from enabling certain cops (ones concerned with patterns and numbered parameters) in the aforementioned pull request because they were related to Ruby 3 syntax which Jekyll has not yet adopted.

Aha I see. Is there a problem enabling them even if Jekyll's syntax is still following ruby 2.x? They will not be triggered until someone tries the new syntax, right?

> The remaining ones revolving minitest and rspec seemed noisy.

Do you mean noisy as entries in rubocop.yml, the cop recommendations themselves?

I am not very opinionated when it comes to which cops should be enabled or not (especially since I'm new to the codebase). The main point of this PR is to have a warning-free `script/fmt`. Another way to achieve this can be to run rubocop with `--disable-pending-cops`, do you think it's a more fitting solution?

ashmaroli:
> Is there a problem enabling them even if Jekyll's syntax is still following ruby 2.x?

Personally, I feel enabling unnecessary cops means redundancy.

> Another way to achieve this can be to run rubocop with `--disable-pending-cops` which disables all the new cops

I like this alternative. Especially since that is your primary goal with this PR.
Let us go with adding this flag to the shell script and updating the PR title and description to reflect the intention.

yboulkaid:
> Personally, I feel enabling unnecessary cops means redundancy.

I buy the argument 👍 

> I like this alternative. Especially since that is your primary goal with this PR.
> Let us go with adding this flag to the shell script and updating the PR title and description to reflect the intention.

Done :)

ashmaroli (approved):

ashmaroli:
Thanks @yboulkaid
@jekyllbot: merge +dev

PR #9138, opened 2022-09-30
Allow line highlighting via the highlight_lines Rouge feature

lylo:

## Summary

Support for highlighting lines of code was added to Rouge in https://github.com/rouge-ruby/rouge/pull/1426/files. This PR adds support to Jekyll for this feature.

Screenshot 2022-09-30 at 14 29 09


ashmaroli:
Thank you for submitting this proposal, @lylo.
This looks good to me at first glance save for a small correction to the documentation. I suggest you specify the need for "double-quoted" value instead of just a quoted value since our parser isn't super-flexible as of now.

lylo:
> Thank you for submitting this proposal, @lylo. This looks good to me at first glance save for a small correction to the documentation. I suggest you specify the need for "double-quoted" value instead of just a quoted value since our parser isn't super-flexible as of now.

Good spot, thanks @ashmaroli. I've updated this in 7cfd1fae4e26f3a9c885394ee164b4820c1f903c

lylo:
Hey 👋 I just wondered whether there was anything I could do to push this PR forward, or is it just a matter of waiting for someone on core to find it and take a look?

parkr:
I could have sworn we had this feature like 8 years ago. Was it a Pygments-only feature that Rouge only recently included?

ashmaroli:
@lylo Pinging you to inform you that this enhancement will land in the next minor release (v4.4.0).
I do have some issues the overall code in the branch but it is not your fault since you were just acting based on surrounding code. I will get back to this after Jekyll v4.3.1 ships.

lylo:
Thanks @ashmaroli 🙏

ashmaroli (approved):
LGTM!

lib/jekyll/tags/highlight.rb
```diff
@@ -81,20 +81,32 @@ def render_pygments(code, _context)
...
+
         lexer = ::Rouge::Lexer.find_fancy(@lang, code) || Rouge::Lexers::PlainText
         formatter.format(lexer.lex(code))
       end
 
+      def line_highlighter_formatter(formatter)
+        ::Rouge::Formatters::HTMLLineHighlighter.new(
+          formatter,
+          :highlight_lines      => @highlight_options[:highlight_lines].map(&:to_i),
+          :highlight_line_class => @highlight_options[:highlight_line_class] || "hll"
```
parkr, line 95: I don't see a test for `highlight_line_class` in the PR diff... did I miss that? When I read it, I thought it was for selecting a Ruby class within Rouge, but looking at it again I think it's to add a CSS class to... some element in the output. A test would help illuminate this and ensure we don't break it in the future. Can it take a list of classes?

parkr (approved):
LGTM. One quick comment about testing the class customization.

lib/jekyll/tags/highlight.rb
ashmaroli: @parkr I think [these lines in the Cucumber test](https://github.com/jekyll/jekyll/pull/9138/files#diff-460702aaee67a2a9b221cdee16477ef0fe3202c7eef2f8ee18e9714a3b965543R60-R61) covers the `hll` bit removing the shadow of doubt:
```gherkin
And I should see "<span class=\"hll\"><span class=\"k\">module</span> <span class=\"nn\">Jekyll</span>" in "_site/index.html"
And I should see "<span class=\"hll\">  <span class=\"k\">module</span> <span class=\"nn\">Tags</span>" in "_site/index.html"
```
@lylo Please add **two supersets** of the current scenario with custom classnames (one for valid unquoted input, the other for invalid *quoted* input)

lylo: @ashmaroli @parkr thanks for picking this up. I'm new to cucumber so I'm not entirely sure what you mean by "two supersets" but I've added new scenarios to cover both the custom line highlighting class, as well as syntax validation. See 97e60d1d8b0b239b1fa9358655dfac92ed270a4c and 56d6d8115a1c8d532a0cd50190ed8775e0b303cc. If there's a more succinct way, please do let me know and I'll be happy to change.

ashmaroli:
@lylo Thank you for adding the new features. Needs some minor edits:
- The `Scenario` keyword for last scenario is intended incorrectly.
- I feel testing for `"Liquid Exception: Syntax error"` would be better.

Since you mentioned "succinct", I would like to request you to make the following change to *cucumber scenarios you have added*:
- There really is no need to have the kramdown code block with triple backticks and the configuration entry for `kramdown.input`. They are principally unrelated to your changes.
(I didn't want to burden you with this earlier since you were  only repeating the pattern in existing code, but since the diff has gotten bigger now, It's best not adding more noise.)

ashmaroli:
P.S. to above comment 👆:
I have edited the `features/highlighting.feature` on the base branch (`master`: 94028e7) to give a better idea of what I meant by "noise" in the earlier comment.

ashmaroli:
Why have you renamed the sample file to "test.rb" from "index.html"?

lib/jekyll/tags/highlight.rb
lylo: This has now been addressed. Apologies for the confusion with the renamed test file there, that was my ignorance - not paying attention.

I've also updated `highlight.rb` so that it throws a syntax error rather than bailing with an obscure (to the user) Ruby error https://github.com/jekyll/jekyll/pull/9138/commits/2dbb1812397cc19165f2bc095f84035cc0c3838e#diff-d38972f992a69af4316684fc12c85a72ca6332c30d73359334eb614fdba826b9R92

lylo:
> Why have you renamed the sample file to "test.rb" from "index.html"?

You can ignore that, it was reverted in a later commit

lylo:
> Buddy, you have made things harder for yourself by cherry-picking my commit from the `master`. Now you will have to undo that action before this gets merged...... The best route is to rebase to current `master` now.
> 
> Locally run in terminal:
> 
> ```
> git fetch upstream master
> git rebase upstream/master
> ```

Sorry, I appear to have made a complete mess of this. I'll need to come back to it when I can actually focus time on it, potentially start a fresh PR.

lylo:
@ashmaroli I have a clean branch locally I can put a fresh PR up for if you prefer. I think that would be best in this circumstance, but I'll defer to you. Apologies for the additional work I've created here.

ashmaroli:
@lylo No worries, I have reset the branch to its former state, but sanitized to remove the cherry-picked commit.
When you get the time, reset your *local* copy before you work any further on this pull request.

To reset your local branch to cleanly mirror the *current* remote version, run:
```sh
git checkout master
git branch -D support-line-highlighting
git fetch upstream refs/pull/9138/head:support-line-highlighting
git checkout support-line-highlighting
```

ashmaroli:
Couple of issues with the validation for proposed features:
- Incorrect expectation -- Tag should expect a (double-)quoted string of numbers, *not an array* for *hl_lines*. Array of integers is a result of preliminary internal processing, not user-given.
- Insufficient -- *hl_class* gets flagged when quoted only if the string consists of non-numbers. What if the parser doesn't catch it? For example: `highlight_line_class="2 3 4"`?

lylo:
> * Incorrect expectation -- Tag should expect a (double-)quoted string of numbers, _not an array_ for _hl_lines_. Array of integers is a result of preliminary internal processing, not user-given.

Fixed in 03a68ea9aa7ca4fcc25748b73da79504f4dead25

> * Insufficient -- _hl_class_ gets flagged when quoted only if the string consists of non-numbers. What if the parser doesn't catch it? For example: `highlight_line_class="2 3 4"`?

I've addressed this in 289421a8de06849d38101c6631d35c5730eaa52d by relaxing the rules of the parser. So while the correct format for declaring `highlight_line_class` is still:

```liquid
{% highlight ruby highlight_lines="1 2" highlight_line_class=one_two %}
```

This change will be forgiving and do the expected thing if somebody did try and pass in a double-quoted list of integers, so this:

```liquid
{% highlight ruby highlight_lines="1 2" highlight_line_class="1 2" %}
```

Will result in a class name of `"1 2"`

I'm happy to be less forgiving and raise a syntax error instead if you prefer. 

The main problem I'm having here is the overarching strict parsing rules and regexes. It could really do with a full overhaul. I'm just trying to work within it's constraints.

ashmaroli:
@lylo Found multiple issues with the latest implementation. I guess you couldn't give 100% focus into this due to the various moving parts and inputs from my end. I therefore took the liberty of cleaning the branch up directly instead of first pointing out all the issues.
Do run `git pull origin  support-line-highlighting` to update your local copy.

ashmaroli:
@parkr Requesting a fresh round of review on this pull request. Thanks.

ashmaroli:
Hello @lylo, It just dawned on me that there's a possibility that users may get confused by the proposed line-highlighter parameter names.

### Tl;dr

Can you think of better prefixes for the new proposed options?

### Details

The tag itself is named `highlight` to convey the notion of *syntax-highlighting*.
Now if we were to ship options named `highlight_lines`, users may think it's a directive to apply syntax-highlighting only to the declared lines. Similar being the confusion with `highlight_class`.

My brain came up with using `mark_*` instead (akin to the HTML tag `mark`), but I feel its too technical and may not be intuitive for lay-users.

lylo:
@ashmaroli I used `highlight_lines` and `highlight_line_class` because this is what's used in Rouge itself, but it's a fair question to ask – internal class names in a dependency doesn't necessarily make sense from a user perspective. 

If we're avoiding `highlight` I think `mark_lines` and `mark_class` would be good alterntives. I can't think of anything better right now. `emphasize` is another word, but I think mark is simpler. 

Happy to change if you agree?

ashmaroli:
@lylo I understand your perspective and that the current implementation is simpler because of it. I guess we will have to settle for `mark_*` then because `emphasis` has a different meaning in HTML and it's a lot more keystrokes than `mark`.
I am sorry that your first contribution to Jekyll got a bit more complicated now.

lib/jekyll/tags/highlight.rb
```diff
@@ -23,7 +23,7 @@ def initialize(tag_name, markup, tokens)
 
             #{markup}
 
-            Valid syntax: highlight <lang> [linenos]
+            Valid syntax: highlight <lang> [linenos] [highlight_lines="3 4 5"] [highlight_line_class=class_name]
```
parkr, line 26: Is there a strong reason to customize the highlight line class? Couldn't a user simply apply the styles they want to `.hll`?

lib/jekyll/tags/highlight.rb
```diff
@@ -23,7 +23,7 @@ def initialize(tag_name, markup, tokens)
...
           MSG
```
parkr, line 27: Could be an interesting opportunity to link out to the jekyllrb.com page for this tag so users have a direct line to a helpful webpage with more information on the matter.

parkr (approved):
LGTM. I don't have a strong opinion on the naming – whatever you feel is best for a non-technical user. Thanks!

lib/jekyll/tags/highlight.rb
lylo: I was having the same thought today! I think we should remove this option for the time being. If we find it's a problem for people in the future, we can always add it. Feels like a premature optimisation right now. 

I'll fix.

lylo: Removed in 8a202c50a6205fc0dda441ec8a52a686aac9a987

lib/jekyll/tags/highlight.rb
lylo: Fixed in 8a202c50a6205fc0dda441ec8a52a686aac9a987

lylo:
> @lylo I understand your perspective and that the current implementation is simpler because of it. I guess we will have to settle for `mark_*` then because `emphasis` has a different meaning in HTML and it's a lot more keystrokes than `mark`.
> I am sorry that your first contribution to Jekyll got a bit more complicated now.

It's no problem, I'm enjoying the journey and learning a lot.

In 8a202c50a6205fc0dda441ec8a52a686aac9a987 I've changed this to `mark_lines`. I've also removed the class option as per suggestion from @parkr which simplifies the code.

ashmaroli:
Thank you @lylo.
If you're interested, you may submit another PR to initialize the upstream `HTMLTable` formatter using keywords instead of a Ruby Hash.

@jekyllbot: merge +minor

ashmaroli:
@jekyllbot: merge +minor

lylo:
> Thank you @lylo. If you're interested, you may submit another PR to initialize the upstream `HTMLTable` formatter using keywords instead of a Ruby Hash.

Hey @ashmaroli, do you mean changing from:

```ruby
Rouge::Formatters::HTMLTable.new(
  formatter,
  :css_class    => "highlight",
  :gutter_class => "gutter",
  :code_class   => "code"
)
```

to:

```ruby
Rouge::Formatters::HTMLTable.new(
  formatter,
  css_class: "highlight",
  gutter_class: "gutter",
  code_class: "code"
 )
```

ashmaroli:
@lylo I did mean that, but I have already made the changes to `master`.

PR #9144, opened 2022-10-03
Mention dirs in config are relative to source

remino:

### DIFF
--- a/docs/_docs/configuration/default.md
+++ b/docs/_docs/configuration/default.md
@@ -10,7 +10,7 @@ file or on the command-line.
 <div class="note info">
   <h5>Be aware of directory paths</h5>
   <p>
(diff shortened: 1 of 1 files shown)


ashmaroli:
Okay, since we are updating documentation, we might as well shed as much light as possible.
Below are what users should know, especially when starting out with Jekyll. Feel free to use the points below to construct a wholesome text copy:

- Jekyll calculates paths relative to the `source`.
- By default, `source` is CWD or current working directory.
- Users have two alternatives to customize `source`. First route is via command-line using the flag `--source`, `-s`. The other route is to define a `source` key in the config file(s).
- The two routes above work similarly except for one important detail: When using CLI flag(s) `--source` , `-s` to override `source`, Jekyll looks for **the config file at that location.** However, if `source` has been overridden via the config file, then the config file **is read from the CWD**.
- Jekyll only accepts paths that are below the configured `source` path. Any path *outside* or *above* the `source` is ignored. The only exception to this is theme-gem and its contents.
- Paths mentioned in config file should be relative to source and point to a descendant in the source directory.

Issue #1126, opened 2013-05-18
Cucumbers tests need to be more clear around timezones


jpiasetz:
The cucumbers tests need a clean up around time-zones. Some of the asserts are implicitly imply date times that would have different time-zones.

See #1090

PR #9161, opened 2022-10-18
Add explicit ISO date, time and timezone test for post dates

bcr:

## Summary

I know you love ISO 8601. I do too. I wanted to make sure if I use that format in my post `date` that it would work. It does. This test will enforce that it continues to work :).

## Context

May be part of #1126

### DIFF
--- a/features/post_data.feature
+++ b/features/post_data.feature
@@ -58,17 +58,17 @@ Feature: Post data
     Then the _site directory should not exist
     And I should see "Document '_posts/2016-01-01-test.md' does not have a valid date in the YAML front matter." in the build output
 
-  Scenario: Use post.date variable with ISO
+  Scenario: Use post.date variable with ISO date and time and timezone
     Given I have a _posts directory
     And I have a _layouts directory
     And I have the following post:
       | title     | date                      | layout | content                 |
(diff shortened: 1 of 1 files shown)


bcr:
I happen to be participating in hacktoberfest also, if you add the label `hacktoberfest-accepted` to this PR I think that helps me. No pressure, I just happened to encounter this during hacktoberfest. Enjoying Jekyll.

bcr:
Hm. The tests pass on my machine. I will dig into it some more. It may be the case that because I used the same base filename as a previous test that something is cross-threaded? Like some torn state between the two tests?

My successful run is here https://reports.cucumber.io/reports/d023abfc-1786-4d9c-aae8-296525d37693

bcr:
I'm thinking that the CI timezone differs from the post timezone, so the final directory is different. Now the question is whether or not this is a bug. Digging.

bcr:
Yeah, I think the essence is that the TZ for the date for the file is the local TZ of whoever generates it, not the TZ in the post.

ashmaroli:
Hello @bcr,
v4.3.0 is slated for release on Thursday and soon after, the codebase will move into v5 development mode. v5 will include breaking changes and won't be shipping anytime soon..

Issue #8842, opened 2021-10-03
ERROR:  While executing gem ... (ArgumentError)     wrong number of arguments (given 4, expected 1)


sriveravi:
Error when executing

>> gem install jekyll bundler

ERROR:  While executing gem ... (ArgumentError)     wrong number of arguments (given 4, expected 1)

I have found the solution, to uninstall psych gem

>> gem uninstall psych

https://stackguides.com/questions/68899508/gem-install-wrong-number-of-arguments-given-4-expected-1

Also, it may be worthwhile to update the WSL installation instructions. They do not have the information to make the gem directory local so that you can install without sudo. I followed the ubuntu instructions.

Info:

gem -v
2.7.6.2

ruby -v
ruby 2.5.8p224 (2020-03-31 revision 67882) [x86_64-linux-gnu]

Running on Windows 10, WSL2 with Ubuntu 20.04.
Surface pro 3.

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the latest <code>3.x-stable</code> or <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, please consider building it first as a plugin. Jekyll 3 introduced [hooks](http://jekyllrb.com/docs/plugins/#hooks) which provide convenient access points throughout the Jekyll build pipeline whereby most needs can be fulfilled. If this is something that cannot be built as a plugin, then please provide more information about why in order to keep this issue open.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

LayersOfAbstraction:
You are definitely not alone. I get the same problem before I even hit `update gem`. I should have never ported WSL2 to a bigger drive. Think I will just get a bigger SDD if I find workaround. Also that SO solution results in more problems. If I run 
```
gem uninstall psych 
```
I will just get the following warning.

```
rdoc-6.4.0 depends on psych (>= 4.0.0)
If you remove this gem, these dependencies will not be met.
Continue with Uninstall?
```
Doesn't sound like an option. And yes I tried that. Just made the problem worse. Reinstalling WSL2 did not help.

ashmaroli:
@LayersOfAbstraction Are you on Ruby 3.1?
`psych-4.x` wasn't installed by default until Ruby 3.1.

[...]

PR #9162, opened 2022-10-20
Wsl2 install doc update

### LINKED ISSUE #9160: [Docs]: Ruby version 2.5 doesn't work for Bash on Windows
## Motivation

I think I found a part in the documentation that isn't correct. On the page "docs/installation/windows/#installation-via-bash-on-windows-10" it says that you should install ruby2.5. But after that on my machine the command `gem install jekyll bundler` crashed.

## Suggestion

I could solve the problem by using ruby2.6.

LayersOfAbstraction:

## Summary

As support for Ruby 2.5 ended [25 Dec 2017](https://endoflife.date/ruby#:~:text=2.6.10-,2.5,2.5.9,-2.4) I have changed the installation documentation for WSL2 to use Ruby 3.0.1 which will end during Christmas 2025. 

Added dependencies for Ruby to allow us to use the `bundle add webrick` command to ensure we could install Jekyll to a local directory.
## Context
Everything I have suggested is covered in [my blog on how to install Jekyll on WSL2](https://levelup.gitconnected.com/how-to-install-jekyll-on-wsl-2-13c3b285d513) and should contain further links that back up my claims.

 Fixes #9160 and #8842

### DIFF
--- a/docs/_docs/installation/windows.md
+++ b/docs/_docs/installation/windows.md
@@ -51,13 +51,39 @@ Your terminal should now be a Bash instance. Next, update your repository lists 
 sudo apt-get update -y && sudo apt-get upgrade -y
 ```
 
-Next, install Ruby. To do this, let's use a repository from [BrightBox](https://www.brightbox.com/docs/ruby/ubuntu/),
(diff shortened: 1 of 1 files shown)


sc68cal:
Just throwing my two cents in here but I think `rvm` might be a better way to manage this, instead of `rbenv`

Issue #9149, opened 2022-10-06
Incremental Build | Use Hash instead of Modification Time


DER-SSt:
### Motivation
I am trying to use the [incremental build](https://jekyllrb.com/docs/configuration/incremental-regeneration/) in a Gitlab-CI-runner. 
After every push the website should be incrementally rebuild.

### Problem
When the runner checks the project out, every file has the current time as modification time. Since Jekyll detects changes based on the modification time, this leads to a full rebuild every time.

### Proposed Solution
Hash input files and store the hash in `.jekyll-metadata`. 
Ignore the modification time completely.

### Gitlab Runner Configuration
```yaml
ci_build_job:
  cache:
    paths:
      - _site/
      - .jekyll-metadata
  script:
    - stat Gemfile   # prints Modification Time
    - jekyll build --incremental
```

### Show that Modifcation Time is used by jekyll --incremental
Tested with [jekyll/jekyll:4.2.2](https://hub.docker.com/layers/jekyll/jekyll/4.2.2/images/sha256-6859f3af91550c7d5935ac97e82d586dad4c67a0577b5484b9303e41565687ef?context=explore)
```bash
jekyll new inctest
cd inctest/
jekyll build --incremental --profile  # Full build
jekyll build --incremental --profile  # Incremental build

find . -type f -exec touch {} +       # Change Modification Time of all files
jekyll build --incremental --profile  # Full build, should be incremental
```

ashmaroli:
Prior Art towards same goal: #7204

yboulkaid:
I'm willing to take over the work from #7204 and continue working on it, will keep you posted!

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the latest <code>3.x-stable</code> or <code>master</code>/<code>main</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, please consider building it first as a plugin. Jekyll 3 introduced [hooks](http://jekyllrb.com/docs/plugins/#hooks) which provide convenient access points throughout the Jekyll build pipeline whereby most needs can be fulfilled. If this is something that cannot be built as a plugin, then please provide more information about why in order to keep this issue open.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

jekyllbot:
[...]

PR #9166, opened 2022-10-22
Allow using MD5 hashes for incremental builds

yboulkaid:

## Summary

This PR adds the optional ability to use MD5 digests for incremental
builds instead of the file's modification time.

## Context

We are using the file's modification times as the base for deciding
wether or not to include them in the incremental build.

This creates issues, for example when files are pulled to a CI server
and their modification times are reset.

This PR adds the optional ability to use MD5 digests for incremental
builds instead of the file's modification time, to make incremental
generation more reliable.

Credit where credit is due, this work is a continuation of @steinbachr's
work in #7204. I picked up where they left, made some changes to the
code for readability and added some tests.

## Notes/questions:

- @mattr- [suggested][1] to make hashing the default, but I would like 
  to start by having it as an option for now so that people could try it
  out first. Maybe we can then make it the default shortly before
  releasting Jekyll 5.0?

- Metadata files are automatically migrated to the new format on the
  first non-incremental run, so we don't need to do extra work here.

[1]: https://github.com/jekyll/jekyll/pull/7204#issuecomment-612708132
  

This fixes issue #9149
This closes PR #7204

### DIFF
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -41,7 +41,7 @@ module Jekyll
       "lsi"                 => false,
       "excerpt_separator"   => "\n\n",
       "incremental"         => false,
-      "incremental_cache_key" => "mtime",
+      "incremental_key"     => "mtime",
 
       # Serving
       "detach"              => false, # default to not detaching the server
--- a/lib/jekyll/configuration.rb
+++ b/lib/jekyll/configuration.rb
@@ -41,6 +41,7 @@ module Jekyll
       "lsi"                 => false,
       "excerpt_separator"   => "\n\n",
       "incremental"         => false,
+      "incremental_cache_key" => "mtime",
 
       # Serving
       "detach"              => false, # default to not detaching the server
--- a/lib/jekyll/regenerator.rb
+++ b/lib/jekyll/regenerator.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "digest"
+
 module Jekyll
   class Regenerator
     attr_reader :site, :metadata, :cache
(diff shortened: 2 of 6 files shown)


ashmaroli:
Thank you for picking this up @yboulkaid.
Is this ready for review or still under development? If it is, do mark this as a draft pull request.

Some *personal* opinions to consider (not mandatory):
  - I prefer using *polymorphism* over switching behavior based on conditionals at various places.
  - How about splitting this PR into (at least) two..? The first PR would be about *allowing the support for (more than 1) incremental cache implementation*.. i.e., introducing the variable `incremental_cache_key` and an interface to support easy switching between *mtime-based* and *digest-based*..
- The subsequent PR would be about focusing only on the implementation and not about its effect on the `Regenerator` instance or generated metadata file..
- One may wonder why the need for an interface if we plan on supporting only one implementation from v5 onwards. My take is that the interface would allow us to fall back to time-tested mtime-based implementation if desired. Additionally, a savvy user can easily implement a SHA256-based caching as a plugin if he so prefers.

yboulkaid:
> Is this ready for review or still under development? If it is, do mark this as a draft pull request.

This was ready for review but your comments prompted some changes I would like to make, so I'm reconverting this to a draft.

It's especially your last point: if we decide to use the MD5 digest for the v5, then I would rather just switch the implementation right away, without introducing a configuration option. A couple arguments for this:
- The incremental feature is already flagged as experimental in the docs, so this is an an area where backwards compatibility isn't as important as other areas.
- Hashing will always be more reliable than modification time.
- Not introducing an option would allow us to keep the codebase complexity down.

What do you think? If you're onboard with this I can go ahead and make the changes in this PR.

ashmaroli:
@yboulkaid Incremental regeneration may be noted as an experimental feature in the docs but since the behavior has been long standing it is best we don't break SemVer promise in a minor version release. Therefore, introducing a config option is best if you intend to release this in v4.x so that users have an option to fall back onto existing behavior if they notice a performance issue or something else, (highly unlikely but better err on the safer side).

yboulkaid:
Gotcha. I'm okay with waiting for a v5 release for this for the sake of a simpler implementation, so let's do that 👍

I'll simplify this and leave it as a draft until we start development on the v5 branch (I saw that you had started that work but had to revert to to get a couple bugfixes in)

PR #9168, opened 2022-10-24
Run tests in utc

oturpe:

This is a 🐛 bug fix.

## Summary

If on a system where system timezone is not +00,
when tests are launched with 'script/test', test case
"TestFilters#test_: filters jsonify filter should convert drop to json"
fails due to difference in expected and actual value of 'date'.
This difference equals the system timezone offset.
Some other forms of lauching tests do work,
such as 'script/test test/test_filters.rb'.
    
Make the tests run correctly regardless of system timezone
by setting TZ=UTC.

## Context

Not related to a GitHub issue.

### DIFF
--- a/script/test
+++ b/script/test
@@ -34,6 +34,8 @@ else
   rubies=($(script/rubies))
 fi
 
+# Tests only pass when timezone offset is zero.
+TZ=UTC
 
 for ruby in $rubies; do
   if [[ "$ruby" == "jruby" ]]


yboulkaid:
I have also stumbled upon this when running tests locally, that would be nice to get in!

parkr (approved):

ashmaroli:
Thanks @oturpe
@jekyllbot: merge +dev

oturpe:
Thank you!

PR #9169, opened 2022-10-26
minor documentation update: removed wrong description of root directory 🔦

SamLubbers:

## Summary

In this document, they describe the root directory as the directory `where your config file is`. This is true in most cases, when the user hasn't changed the `source` [configuration option](https://jekyllrb.com/docs/configuration/options/). However, when `source` is set to another directory, then the root directory of files is in another directory from the directory of the config file.

Please let me know if the text I have removed adds clarity, or if I should make a reference to the `source` option from the configuration

## Context

I am new to jekyll, I've read through the entire documentation, and I noticed this minor inconsistency

### DIFF
--- a/docs/_docs/liquid/tags.md
+++ b/docs/_docs/liquid/tags.md
@@ -112,7 +112,7 @@ You can also use the `link` tag to create a link in Markdown as follows:
 ```
 {% endraw %}
 
-The path to the post, page, or collection is defined as the path relative to the root directory to the file, not the path from your existing page to the other page.
+The path to the post, page, or collection is not the path from your existing page to the other page, but the path relative to the root directory: where your config file is (if you have not changed the `source` configuration option).
 
 For example, suppose you're creating a link in `page_a.md` (stored in `pages/folder1/folder2`) to `page_b.md` (stored in  `pages/folder1`). Your path in the link would not be `../page_b.html`. Instead, it would be `/pages/folder1/page_b.md`.
 
--- a/docs/_docs/liquid/tags.md
+++ b/docs/_docs/liquid/tags.md
@@ -112,7 +112,7 @@ You can also use the `link` tag to create a link in Markdown as follows:
 ```
 {% endraw %}
 
-The path to the post, page, or collection is defined as the path relative to the root directory (where your config file is) to the file, not the path from your existing page to the other page.
+The path to the post, page, or collection is defined as the path relative to the root directory to the file, not the path from your existing page to the other page.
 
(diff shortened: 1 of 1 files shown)


ashmaroli:
@SamLubbers Thank you for proposing the change, but with that extra bit of context removed, a new user stumbling upon this page directly will not have any clue as to what the root directory is.
So the better alternative would be to concisely state what root dir is with respect to the two scenarios you've mentioned.

SamLubbers:
@ashmaroli thank you for your quick review. I agree with you that it is better to add additional context so that readers understand what the root directory is. I've just tried improving the description. Please let me know if this is good enough.

ashmaroli:
Thank you @SamLubbers for accepting my suggestion. But now that I look at the bigger picture, the explanation seems convoluted.
Since we are clarifying stuff related to the `{% link ... %}` tag, I feel it would be best describing stuff with respect to the `link` tag itself.
Assuming you have the time to spare, do try and understand the following and then reword the whole line / paragraph concisely:
- Let's call the spade, a spade, instead of what it isn't.
- The `link` tag simply looks for requested resource using their `relative_path` attribute. This `relative_path` is exposed to the end-user via Liquid as `path`.
- Unfortunately, the `relative_path` isn't quite consistent for all resources.
- For a standalone *page*, it is the *path to the resource from the `site.source` **without a forward-slash** at the beginning*.
- For a *static file*, it is the *same as for a page **but with a forward-slash** at the beginning*.
- For documents in a user-defined collection, it is the *same as for a page* irrespective of `collections_dir` configuration.
- For posts, it is the *same as for a document in user-defined collection* but changes depending on the use of superdirs to act as categories.

SamLubbers:
@ashmaroli thank you for the detailed explanation. I now see that the detail that is important for the user to know is that when linking to an asset, they must use the forward slash `/` and when linking to a page, they shouldn't. After reading the whole section of `Links` I see that it could be clarified with the following improvement.

- From the subsection `Linking to pages` remove the example link to a post (`{% link _posts/2016-07-26-name-of-post.md %}`), because `Linking to posts` is described in the subsection below
- From the subsection `Linking to pages` remove the example link to an asset, and make a new section `Linking to assets` where we describe that assets must use a forward slash `/`
- Improve the description of the `path` in the `Linking to pages` section: explain that that for pages (and collections) the path is relative to the root directory, and that because of that they must not use the forward slash `/`

Let me know if you also think this would clarify the documentation

ashmaroli:
Hello @SamLubbers, I agree with everything except the last bullet-point.
- Omit the use of `root dir` entirely. It is not a *thing* in Jekyll yet.
- The actual reason relative path of assets needs a forward slash at the beginning is an internal discrepancy.
- You have misunderstood my explanation with respect to collection documents (and posts). Please read it again. If you don't see the difference, I will reword the explanation.

PR #9170, opened 2022-10-26
Revert "Incrementally rebuild when a data file is changed"

### LINKED ISSUE #9165: [Bug]: 4.3 not reading data from `site.<collection>` correctly.
### Operating System

macOS Monterey 

### Ruby Version

ruby 3.0.3p157

### Jekyll Version

jekyll 4.3

### GitHub Pages Version

_No response_

### Expected Behavior

On version 4.2.2, we have a `_data` collection, reading data from here works as expected, for example: `{{site.data.customers}}`. When logging the data, it shows an object with all properties.
image

In collections where we use a `name` property, that property is set correctly on the site, for example when using: `{{include.event.name}}`.
image

### Current Behavior

Seeing odd behavior when reading our _data with `site.data`. When logging the data it shows a Jekyll hash, rather than the actual object: `site.data.customers: #<Jekyll::DataHash:0x000000012cad8db0>`

The `name` property is also being replaced by the filename in all collections. In our `events` collection, we have a `name` property is each file within the collection, for example: `name: Coalesce 2022`. However, the `name` property is replaced with the filename on the site in Jekyll 4.3.0 when using `{{include.event.name}}`. Screenshots below:
Screen Shot 2022-10-21 at 2 54 07 PM

image
image

The same is occurring for another collection `_resource-categories/case-studies.md`:
```
---
name: Case studies
singular: Case study
slug: case-studies
description: ""
---
```
And when logged, the name property is replaced with the file name:
image

### Relevant log output

```shell
Build installed latest Jekyll 4.3.0 version:
9:37:35 AM: Fetching jekyll 4.3.0
9:37:36 AM: Installing jekyll 4.3.0

We noticed these issues after Jekyll 4.3.0 was installed.
Reverting to Jekyll 4.2.2 reverted the issue:

10:10:58 AM: Fetching jekyll 4.2.2
10:10:58 AM: Installing jekyll 4.2.2
```

### Code Sample

_No response_

ashmaroli:
## Summary

While investigating the reason behind #9165, I realized that patching the reported issue would only act as a band-aid and change in behavior elsewhere will exist until someone reports it. The better solution in order to honor Jekyll's SemVer policy is to revert the causal commit.

Reverts jekyll/jekyll#8771

## Context

Resolves #9165


ashmaroli:
@jekyllbot: merge +fix

PR #9174, opened 2022-11-02
Fix Nested tree navigation with recursion example

robertlove:
The current [Nested tree navigation with recursion](https://jekyllrb.com/tutorials/navigation/#scenario-9-nested-tree-navigation-with-recursion) example results in invalid HTML due to incorrect nesting. This PR fixes that.


ashmaroli (approved):

ashmaroli:
Thank you for catching and fixing this @robertlove.
@jekyllbot: merge +doc

PR #9185, opened 2022-11-17
Optimize default front matter using `File.fnmatch?`

ashmaroli:
- This is a :zap: change.
- I've added tests.

## Summary

Instead of using `Dir.glob` to create a list of matching paths and testing *given relative_patch* against each list entry, use `File.fnmatch?` to directly test if *given relative path* matches configured `scope["path"]` with glob pattern sigil (`*`).

## Notes

- With this, there is no need for having *collections_dir* as part of the `scope["path"]`. But for backwards-compatibility, continue accepting `scope["path"]` with collections_dir and strip it away under the hood. *TODO: Deprecate this need in a subsequent pull request.*


mattr- (approved):

test/test_front_matter_defaults.rb
```diff
@@ -53,9 +49,36 @@ class TestFrontMatterDefaults < JekyllUnitTest
...
+            "values" => { "css_class" => "epilson" },
+          },
+        ]
+      )
+      site.read
+      @staff = site.collections["staff"]
+    end
+
+    should "affect the appropriate items only" do
+      item = @staff.docs.sample
```
parkr, line 75: [`Array#sample`](https://ruby-doc.org/core-3.1.1/Array.html#method-i-sample) selects a random item, so if you're expecting to always have the same item come out, then use `#shift` or `#[]`.

test/test_front_matter_defaults.rb
```diff
@@ -53,9 +49,36 @@ class TestFrontMatterDefaults < JekyllUnitTest
...
+      assert_equal "simple", item.data["layout"]
+      assert_nil item.data["css_class"]
 
-    should "call Dir.glob block" do
-      assert_includes @output, "Globbed Scope Path:"
+      item = @staff.files.sample
```
parkr, line 79: Same here about `#sample`

parkr (approved):
Using `#sample` (a method for randomly selecting an array element) in the tests is probably not what you want, but otherwise LGTM.

test/test_front_matter_defaults.rb
ashmaroli: Actually, the intention here is to indeed test a random object in the *homogeneous* array during each run.

parkr: Every doc is the same, or the array is of size 1? I’m still confused about this. If it’s one element, it would be clearer to first assert the size, then assert its contents. If the array has >1 element that you expect to be the same, wouldn’t it be more effective to iterate over all elements and assert each element has the contents you want.

what I’m afraid of is that we’ll add a new doc or file to this collection in the fixture site and these tests will randomly fail rather than consistently fail. I think tests should be deterministic to avoid contributors chasing down confusing errors that only occur sometimes.

parkr (approved):
Thanks, @ashmaroli!

parkr:
Interesting Ruby 3.1 failure... couldn't compile C extensions for liquid-c.

ashmaroli:
Restarted it to see if it was a one-off failure..

ashmaroli:
This could be because of recent release Ruby 3.1.3 or maybe not.
Will merge this after investigation.

ashmaroli:
@parkr Can I backport this change to `4.3-stable` ?

parkr:
> This could be because of recent release Ruby 3.1.3 or maybe not.
Will merge this after investigation.

I took a look at the repo for issues and noticed that there's a 4.1.0 release but the error looks like a 4.0.0 release. Are we able to use the latest 4.x release of liquid-c?

parkr:
> Can I backport this change to 4.3-stable ?

I guess, but why? Is 4.4 a big upgrade from 4.3 such that folks couldn't upgrade easily? We don't need to provide every optimization to every minor branch, but if you think folks are prevented from upgrading then go for it.

ashmaroli:
> Are we able to use the latest 4.x release of liquid-c?

@parkr The latest release of liquid-c probably syncs with Liquid 5.x. I haven't tested though.
Regardless, my take is to lock to Ruby 3.1.2 for now. Handling in #9196

ashmaroli:
@jekyllbot: merge +fix

PR #9039, opened 2022-04-20
Remove misleading totals row from `--profile` table

ashmaroli:
## Summary

The total time printed as part of the build profile table is often *significantly* different from the total time reported in the *Build Summary* table. Therefore, *stop* outputting the info that is more misleading than beneficial.

*(Deferring a "fix" per se, for future)*

## Context

Closes #8796

PR #9186, opened 2022-11-17
Remove totals in profile table properly

ashmaroli:
- This is a 🐛 bug fix.

## Summary

In #9039 I proposed to remove the `TOTALS` row but there were remnants that continued to generate and render info.
The current pull request removes all of such remnants.

## Notes

- Consider backporting this to the `4.3-stable` branch.


ashmaroli:
@jekyllbot: merge +fix

PR #9188, opened 2022-11-18
Optimize rendering of Link tag

ashmaroli:

## Summary

Currently, the `{% link %}` tag iterates through every site file (via `Site#each_site_file`) on every render call.
Optimize that flow by iterating through every site file once, prepend `site.baseurl` to the item's url and stash the result in a Hash for lookup, just before documents and pages get rendered.


test/test_tags.rb
```diff
@@ -7,21 +7,22 @@ def setup
...
     site.read if override["read_all"]
-
     info = { :filters => [Jekyll::Filters], :registers => { :site => site } }
     @converter = site.converters.find { |c| c.instance_of?(converter_class) }
     payload = { "highlighter_prefix" => @converter.highlighter_prefix,
                 "highlighter_suffix" => @converter.highlighter_suffix, }
 
-    @result = Liquid::Template.parse(content).render!(payload, info)
-    @result = @converter.convert(@result)
+    site.send(:set_up_link_registry)
```
mattr-, line 22: The fact that we're having to use `send` here is setting off warning bells in my head. Maybe the link registry should be a separate object?

test/test_tags.rb
```diff
@@ -7,21 +7,22 @@ def setup
     FileUtils.mkdir_p("tmp")
   end
 
+  # rubocop:disable Metrics/AbcSize
```
mattr-, line 10: Rather than disabling the AbcSize warning here, maybe we can extra this into a separate test helper class? what do you think?

mattr- (commented):
Noticed this was in my review queue, so I left a few comments. I like where this idea is going though!

test/test_tags.rb
ashmaroli: A separate helper class sounds good to me. :+1:
I shall open a separate pull request implementing it.

y377:
@ashmaroli hi!
I couldn't find any reason to submit an issue to Jekyll, but now I finally managed to find one, about `{% link %}`

Scenario.
A collection, which is called _books, has three subfolders, namely: Harry Potter, Lord of the Rings and Iron Man
Suppose that document A within Harry Potter (let's define it as A first) has two hyperlinks in the body of A to Lord of the Rings document B and Iron Man document C. Under this one document, or if there are fewer hyperlinks, we can manually add `{% link _books/Lord of the Rings/document B %}` and `{% link _books/Iron Man/document C %}`, if we have just one hyperlink (and under a subfolder), let's use Lord of the Rings as an example, we can use the liquid variable `{% assign mydoc = "_books/Lord of the Rings/" %}`
then the body of the document directly using `[document name]( {% link {{ mydoc }} document B.md %})` and `[document name]( {% link {{ mydoc }} document BB.md %})`, because they have the same path, this is very efficient, but when there are multiple places to add, or multiple documents need to add different paths of hyperlinks to D ,E,F,G,H,K omitted and so on, we can also use liquid `assign` to define a number of different variable names and add them manually one by one, because after all, it can be used.
But a special use case arises; suppose we have thousands of documents, each document has dozens of hyperlinks, if we use regular expressions, the document is modified by regular replacement, you can omit the above manual addition steps, if a subfolder, the variable name is only one, can be very fast processing, but in different subfolders, we need to define the variable name without, at this time regular expression can correctly capture the "document name", but the variable name is the same, regular in the processing can not be well distinguished from their corresponding path, I tried ` {% assign mydoc = "_books/Lord of the Rings/" %}{% assign mydoc = _books/Iron Man / %}`
Also tried `{% assign mydoc = "_books/Lord of the Rings/" or mydoc = _books/Iron Man/ %}`
and `{% assign mydoc = "_books/Lord of the Rings/" and mydoc = _books/Iron Man/ %}`.
After regular capture, it does not distinguish well which document corresponds to which folder, all will appear to build the error.
If you have a better solution, please teach me more, thank you very much!

Translated with www.DeepL.com/Translator (free version)

PR #9200, opened 2022-12-02
Update macos.md

bfindell:
The instructions for ruby installation failed on a Macbook with an M1 chip.  Here is the source of the fix:   

https://www.rubyonmac.dev/how-to-install-ruby-on-macos-12-6-apple-silicon

This is a 🐛 bug fix (or workaround) in the jekyll installation instructions.

### DIFF
--- a/docs/_docs/installation/macos.md
+++ b/docs/_docs/installation/macos.md
@@ -57,6 +57,12 @@ Install the latest stable version of Ruby:
 ruby-install ruby
 ```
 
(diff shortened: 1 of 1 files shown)


monfresh:
Hi. I'm the author of that article. Thanks for this PR, but this is no longer an issue. When following the instructions in the docs, the latest version of Ruby will be installed. Currently, the latest version of Ruby is 3.3.0. Also, the bug that required adding `--enable-shared` affected older Ruby patch versions, and since there are newer patch versions (such as 3.1.4), this workaround is no longer needed.

PR #9195, opened 2022-11-26
Update Jekyll on macOS (macos.md) and current ruby version to 3.1.3 in ruby.yml

masaki9:
This is a 🔦 documentation change.

 I've updated the current ruby version to 3.1.3 in ruby.yml to update the ruby version referenced in the [Jekyll on macOS](https://jekyllrb.com/docs/installation/macos/) guide.

## Summary

Ruby version 3.1.3 was released yesterday, November 24, 2022, and step 2 in the [Jekyll on macOS](https://jekyllrb.com/docs/installation/macos/) guide currently references ruby version 3.1.2. The pull request change updates the current ruby version in ruby.yml to 3.1.3 from 3.1.2 to update the guide.

## Context

Fixes #9194

PR #9202, opened 2022-12-03
Update current Ruby version to 3.1.3

matangover:
This will update the installation instructions that reference `current_version` to make them work properly.


masaki9:
FYI, I already created #9195 with the same change as yours 8 days ago.

matangover:
Thanks

PR #9196, opened 2022-11-27
Lock Ruby in CI to v3.1.2 and bump JRuby to v9.4.0.0

ashmaroli:
## Summary

Our CI appears to break on Ruby 3.1.3 in one of our PRs. Investigate with `master` codebase.

## Context

Ruby 3.1.3 was shipped on 2022-11-24.

PR #9222, opened 2022-12-21
Unpin Ruby v3.1.2 and use v3.1.x instead in CI

ntkme:
## Summary

This partially reverts changes done in https://github.com/jekyll/jekyll/pull/9196

## Context

Ruby 3.1 in the CI was pinned to 3.1.2 due to issues in ruby-builder and setup-ruby action. That is now fixed: https://github.com/ruby/setup-ruby/commit/4b2d1d631efa087f8896c15a0c6023dc2f483198

FYI @ashmaroli @parkr


ntkme:
Never mind, looks like it is not the same issue.

PR #9248, opened 2023-01-09
Patch Liquid 4 for Ruby 3.2

parkr:
This is a 🐛 bug fix.

## Summary

Shim the methods that Ruby 3.2 removed to regain Liquid 4.x compatibility.

## Context

Ruby 3.2 doesn't have the #tainted? and #untaint methods, so it fails with
Liquid 4. The Liquid 5 upgrade has some incompatibilities that break our
tests and potentially users' sites (#9030, https://github.com/Shopify/liquid/issues/1566).
This upgrade to Liquid 5 will need to wait for Jekyll 5 due to this
breakage of these commonly-used filters.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,8 @@ jobs:
             ruby_version: "3.0"
           - label: Ruby 3.1.2
             ruby_version: "3.1.2"
+          - label: Ruby 3.2.0
+            ruby_version: "3.2.0"
           - label: JRuby 9.4.0.0
             ruby_version: "jruby-9.4.0.0"
     steps:
--- a/Gemfile
+++ b/Gemfile
@@ -86,7 +86,7 @@ group :jekyll_optional_dependencies do
 
   platforms :ruby, :mswin, :mingw, :x64_mingw do
     gem "classifier-reborn", "~> 2.2"
-    gem "liquid-c", "~> 4.0"
+    gem "liquid-c", "~> 4.0" if RUBY_VERSION < '3.2'
     gem "yajl-ruby", "~> 1.4"
   end
 
--- a/lib/jekyll/liquid_extensions.rb
+++ b/lib/jekyll/liquid_extensions.rb
@@ -20,3 +20,22 @@ module Jekyll
     end
   end
 end
+
+if RUBY_VERSION >= "3.2"
+
+  module Liquid
+    module StandardFilters
+      def escape(input)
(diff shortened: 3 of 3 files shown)


parkr:
> I was unsure about monkey-patching Ruby Core class myself and decided to wait for a response from the Liquid devs regarding a potential release in the 4.x series.

I haven't gotten much from them so far. They may still be out on vacation for the winter holidays but I figure we should do what we can to improve the situation for our users.

> Instead of patching the Object class globally, why not patch it conservatively using Ruby's Refinement feature?

Sure! Want to edit this branch to that effect? We could also only add the patch if these methods don't already exist to scope it down further.

> I am not really sure if Object#tainted? should return false or just be a no-op. However, since the convention for predicate method definitions is to return a Boolean, I am in agreement with your implementation.

👍 

> How far back are we going to backport this patch? Is 3.9.x still under maintenance?

Yes, I want to patch 3.9.x. This would unblock GitHub Pages from upgrading.

> What's the plan in the event Liquid devs does cut a patch for this in their 4.x series eventually?

I'd revert the changes in that case, since users could patch instead by upgrading liquid. I did notice that liquid-c didn't compile in Ruby 3.2 but that's unrelated to the tainted feature.

ashmaroli:
> We could also only add the patch if these methods don't already exist to scope it down further.

As per https://github.com/Shopify/liquid/pull/1268/commits/065ccbc4aa5b4955dae9743509907bd9ad0db0b9, the only places `[un]taint[?]` is used is in `Liquid::StandardFilters` and `Liquid::Variable`. So, we should be good to go with the current state of this branch.

sc68cal (approved):

parkr:
Your changes LGTM!

markets:
@parkr @ashmaroli a fix in Liquid side has been recently submitted, and it seems to work fine: https://github.com/Shopify/liquid/pull/1668.

ashmaroli:
I guess this can be closed since Shopify/Liquid is going to release the patch themselves. :tada:

ashmaroli:
Superseded by https://github.com/Shopify/liquid/pull/1668

parkr:
I think Liquid v4.0.4 will fix this. Can anyone confirm? https://rubygems.org/gems/liquid/versions/4.0.4

markets:
@parkr I can confirm that works fine on my side (Jekyll plugin): https://github.com/markets/jekyll-timeago/pull/36/commits/1d263e1038029e904216e3e91722cfdb18db4881 (all matrix ✅)

PR #9257, opened 2023-01-15
Rename sass partial created for new blank site

ashmaroli:
- This is a 🐛 bug fix.
- I've added tests.

## Summary

`jekyll-sass-converter-3.0.0` dropped support for importing sass partial named the same as the calling sass file.
Therefore, rename the partial **`_sass/main.scss`** to **`_sass/base.scss`**.

## Context

Resolves #9250

PR #9259, opened 2023-01-17
script/default-site: accept flags for `jekyll new`

parkr:

This is a 🔨 code refactoring.

## Summary

In #9257, @ashmaroli proposed copying `script/default-site` to `script/blank-site` so that we could run `jekyll new --blank` then `jekyll build` the way that `script/default-site` does. I was hesitant to duplicate much of the script into another script, so I was thinking that maybe we could make the common components a single script.

With this, you can `script/default-site --blank` to run `jekyll new --blank`. It still runs the smoke test that says "does a blank site build".

The best outcome would probably be a Cucumber test which is already setup to do roughly this kind of operation (create a site, then run jekyll build, and verify that it built the expected output).

### DIFF
--- a/script/default-site
+++ b/script/default-site
@@ -1,20 +1,23 @@
 #!/usr/bin/env bash
 # Runs the `jekyll new` command and builds the default site as a sanity check
+# Additional flags for `jekyll build` may be passed to this script.
 
 set -e
 
+TMP_SOURCE="tmp/default-site-$$"
+
 echo "$0: setting up tmp directory"
 mkdir -p ./tmp
-rm -Rf ./tmp/default-site
+rm -Rf "./$TMP_SOURCE"
 
 echo "$0: creating new default site"
-bundle exec jekyll new tmp/default-site
-pushd tmp/default-site
+bundle exec jekyll new "$TMP_SOURCE"
+pushd "$TMP_SOURCE"
 
 echo "$0: respecifying the jekyll install location"
 ruby -e "contents = File.read('Gemfile'); File.write('Gemfile', contents.sub(/gem \"jekyll\".*\\n/, 'gem \"jekyll\", path: \"../../\"'))"
 echo "$0: installing default site dependencies"
 BUNDLE_GEMFILE=Gemfile bundle install
 echo "$0: building the default site"
-BUNDLE_GEMFILE=Gemfile bundle exec jekyll build --verbose --profile
+BUNDLE_GEMFILE=Gemfile bundle exec jekyll build --verbose --profile "$@"
 popd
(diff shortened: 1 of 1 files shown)


ashmaroli:
Hmmm.. I'm really not able to wrap my head around the *current diff*. (Your idea is sound. But the implementation is sketchy.)
- I am not able to instantly grok the purpose of using PID-specific tmp dir. Is it avoid deleting a pre-existing non-empty tmp dir?
- I do not see the point of "$@" hook to `jekyll build` incantation. `--blank` is only valid for `jekyll new`. `jekyll build` doesn't recognise `--blank`.

> The best outcome would probably be a Cucumber test..

I agree with this thought.
In fact, there's already a scenario testing `jekyll build --blank` (`features/create_sites.feature:6`). *However, It just checks the presence of specific source files instead of running `jekyll build` and validating those source files.*
IMHO, it'd be best if you can scrap the current proposal and *enhance* the above Cucumber test with additional 3steps (to `cd` into scaffolded blank site, run `jekyll build` and assert a zero exit status).

script/default-site
```diff
@@ -1,20 +1,23 @@
...
+
 echo "$0: setting up tmp directory"
 mkdir -p ./tmp
-rm -Rf ./tmp/default-site
+rm -Rf "./$TMP_SOURCE"
 
 echo "$0: creating new default site"
-bundle exec jekyll new tmp/default-site
-pushd tmp/default-site
+bundle exec jekyll new "$TMP_SOURCE"
```
parkr, line 14: New "$@" should be here - sorry for the confusion @ashmaroli!

parkr:
> I am not able to instantly grok the purpose of using PID-specific tmp dir. Is it avoid deleting a pre-existing non-empty tmp dir?

Yeah, I was thinking that we'd invoke this twice and it'd be nice to not blow away the first one, but I can revert if you prefer! Let me know.

ashmaroli:
> I was thinking that we'd invoke this twice and it'd be nice to not blow away the first one..

If we are indeed continuing down the "Bash script" route (as against the "Cucumber test" route), I feel we can address your primary concern (about duplicated code) using just "Bash functions" to respect DRY and the secondary concern (about keeping directories separate and untouched) by exposing `$@` to `jekyll new`.

```
setup_tmp_dir() { ... }

...

setup_tmp_dir()
[[ $@ ]] && args="$@" || args="default-site"
bundle exec jekyll new "$args"
...
```
```
script/default-site
```
```
script/default-site --blank
```

parkr:
Updated to remove the PID-specific dir.

In general I’m trying to simplify, so adding another file with bash functions doesn’t feel like it simplifies.

ashmaroli (approved):

ashmaroli:
Thanks @parkr
@jekyllbot: merge +dev

PR #9261, opened 2023-01-19
Update default.md

y377:
Request to update the document, this prompt is very suitable for intermediate Jekyll people (if the Jekyll users are divided into primary, intermediate, advanced) to build debugging, in order to build more efficient, more powerful medium and large websites, I can be considered intermediate, but the first time I know this debugging option, And found a '_config.yml' configuration error. I would never have found this useful option if I hadn't debugged the TOC for karmdown.

### DIFF
--- a/docs/_docs/configuration/default.md
+++ b/docs/_docs/configuration/default.md
@@ -85,5 +85,7 @@ kramdown:
   input             : GFM
   hard_wrap         : false
   footnote_nr       : 1
-  show_warnings     : false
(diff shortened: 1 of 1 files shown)


ashmaroli:
Hello @y377.
I cannot accept a full-scale *alternate* suggestion (I mean explaining what the *opposite* of the default value would do) as an inline comment.
Please move that to a separate section, preferably as a `<div class="note info" />` item after the main code-block on the page.

Issue #9268, opened 2023-01-24
[Bug]: i18n dependency version conflict for jekyll 3.9.2 preventing activesupport update to fix CVE-2023-22796


agnostic-apollo:
### Operating System

Ubuntu 21.04

### Ruby Version

ruby 2.7

### Jekyll Version

3.9.2

### GitHub Pages Version

227

### Expected Behavior

The [`activesupport` `6.1.7.1`](https://rubygems.org/gems/activesupport/versions/6.1.7.1) can be used with `jekyll` `3.9.x`.

### Current Behavior

The [`CVE-2023-22796`](https://github.com/advisories/GHSA-j6gc-792m-qgm2) requires `activesupport` to be updated to `6.1.7.1` or ` 7.0.4.1`.

The [`github-pages` `227`](https://rubygems.org/gems/github-pages/versions/227) gem depends on [`jekyll` `= 3.9.2`](https://rubygems.org/gems/jekyll/versions/3.9.2), which depends on `i18n ~> 0.7`.

The [`activesupport` `6.1.7.1`](https://rubygems.org/gems/activesupport/versions/6.1.7.1) depends on `i18n >= 1.6, < 2`, which prevents an update. The [`activesupport` `6.0.6.1`](https://rubygems.org/gems/activesupport/versions/6.0.6.1) was the last version that depended on `i18n >= 0.7, < 2`, which then uses [`i18n` `0.9.5`](https://rubygems.org/gems/i18n/versions/0.9.5) to also satisfy `jekyll`'s `i18n ~> 0.7` requirement.

To fix the issue would require using [`jekyll` `>= 4.0.0`](https://rubygems.org/gems/jekyll/versions/4.0.0), which depends on `i18n >= 0.9.5, < 2` or a new `3.9.x` release with a higher `i18n` dependency version.

Re-post of https://github.com/github/pages-gem/issues/866

CC: @parkr

parkr:
This gem supports the `slugify` filter. https://github.com/jekyll/jekyll/commit/93e3eb06d2

I filed #9269 to fix this.

agnostic-apollo:
Great, thanks!

agnostic-apollo:
Jekyll is also using both `1.2` and `2.0` for test variants for windows.

https://github.com/jekyll/jekyll/blob/ff60d51c9b59d3bb5cc89f6a4f493602b42c5cac/appveyor.yml#L17-L23

ashmaroli:
[...]

PR #9282, opened 2023-01-27
Release :gem: v3.9.3

parkr:
This is a 🚀 release.

## Summary

Add release post, history, and update `version.rb` for :gem: release v3.9.3.

After this is reviewed & approved, I'll forward-port it to `master`.

## Context

Truly closes https://github.com/jekyll/jekyll/issues/9268.


ashmaroli:
Hello @parkr,
I would like to the following changes made:
- Let's not highlight the bit about *non-half hour offsets* --- the module is only used when built using Windows platform; its not a *fix* per se because it was not a feature prior to this. Let it be a sort of an Easter egg for those few who stumble upon it.
- I think the release post should also state that this is partly *a security release as well* because the loosened version-constraints on `i18n` gem was made to allow updating `activesupport` to a more recent version.

Other than the above, **please consider the pull request approved** and proceed to ship the release at your discretion.

parkr:
Removed the bit about non-half-hour offsets. I don't think I need to mention that it's a security release since we didn't patch anything of our own, and none of the dependencies we use had a security vulnerability. This just fixes the mismatch so that folks can upgrade, but not all users of Jekyll will need to upgrade (so I don't want to scare them).

PR #9237, opened 2022-12-29
test: use hash explicitly for Struct initializer for ruby 3.2

mtasaka:
ruby 3.2 changes Struct.new behavior so that keyword_init option is enabled by default. This means that keyword argument passed to the initializer of Struct is now treated differently with ruby3.2.

To avoid behavior change in jekyll test code, use hash explicitly.

Fixes #9236 .

PR #9285, opened 2023-01-30
Backport #9237 to 4.3-stable "test: use hash explicitly for Struct initializer for ruby 3.2"

oturpe:

This is a 🐛 bug fix.

## Summary

Merge request #9237 fixed running tests with Ruby 3.2.
In Fedora, we already use version of Ruby, and want to include the latest Jekyll 4.3.2.
Because of this, I ask to backport #9237 to `4.3-stable`.

This project seems to use extensive automation to manage pull requests.
If you prefer, you can close this pull request and backport the merge request using your preferred method.

## Context

No issue, but I can add one if you prefer to have one for the archives.
The title would be _tests from 4.3-stable fail with Ruby 3.2_.

### DIFF
--- a/test/test_filters.rb
+++ b/test/test_filters.rb
@@ -718,7 +718,7 @@ class TestFilters < JekyllUnitTest
           {
             "name" => name,
             :v     => 1,
-            :thing => M.new(:kay => "jewelers"),
+            :thing => M.new({:kay => "jewelers"}),
             :stuff => true,
           }
         end


ashmaroli:
Hello @oturpe,
We do not bundle the `test/` directory in `jekyll`  gem. So to me, this backport seems redundant *as of now*.
Is there any explicit need for this commit for vendoring Jekyll on Fedora?

oturpe:
The reason I ask for this is this:
In Fedora, we always try to run the project's test cases during package build.
This ensures that the package works correctly using Fedora's dependency versions, compiler flags, patched and so on.
For Jekyll, and many other  Ruby gems, we do the following:
First, we download the official gem for the runtime parts,
then add the test cases from the corresponding Git tag.
(See [here][test-merge] if you are interested in the details.)

For this approach to work, we either carry a downstream patch for Ruby 3.2,
or the tests need to be updated in the stable branch as proposed here.
We already have the downstream patch in place,
so it is not a big issue if you do not want to maintain the tests in stable branches.

[test-merge]: https://src.fedoraproject.org/rpms/rubygem-jekyll/blob/bea8f8aa5f21147844374097191a0a3d115c0b21/f/rubygem-jekyll.spec#_106

ashmaroli:
Thanks for the insight.
Given the fact that since we don't bundle the `tests/` in the gem, I won't be merging this pull request now because even if I were to cut a v4.3.3, the gem will be identical to v4.3.2.
But I won't outright close this either just in case we need to ship a critical bugfix as v4.3.3 .

mattr- (approved):
I'm cutting a 4.3.3 release, so I'll include this PR in that release as well.

PR #9288, opened 2023-02-02
Updated windows.md

Gyrohammer:

## Summary

I updated the ruby versions and gem update command.

## Context
Ruby is now at v2.7.x and some of the packages (gems? New to ruby sorry) Jekyll uses are incompatible with 2.5. I removed 2.5 and installed 2.7 from the same repository listed in the doc. Jekyll then failed install since the latest sass-embedded wasnt supported on RubyGems 3.1.6. To update RubyGems I then had to run `gem update --system` which resolved the issue  and let me install Jekyll. This was all performed on Windows 11 via WSL.

### DIFF
--- a/docs/_docs/installation/windows.md
+++ b/docs/_docs/installation/windows.md
@@ -57,13 +57,13 @@ which hosts optimized versions of Ruby for Ubuntu.
 ```sh
 sudo apt-add-repository ppa:brightbox/ruby-ng
 sudo apt-get update
-sudo apt-get install ruby2.5 ruby2.5-dev build-essential dh-autoreconf
+sudo apt-get install ruby2.7 ruby2.7-dev build-essential dh-autoreconf
 ```
 
 Next, update your Ruby gems:
 
 ```sh
-gem update
+gem update --system
 ```
 
 Install Jekyll:


ashmaroli:
I agree with you, @Gyrohammer.
Even though Jekyll Core codebase is totally compatible with Ruby 2.5, its dependencies need not be so.
Ruby 2.7 itself will be EOL in about 7 weeks. So people just starting out with Ruby / Jekyll tend to install more recent Ruby 3.1 and above.

Are there more changes incoming since you have this as a draft request..?

Gyrohammer:
Hey there! My apologies I've never contributed to a public project before and got cold feet! I thought a draft wouldnt publish but it did. I looked it up after the fact and saw that 2.7 is EOL, I wasnt aware before I published this. I wasnt planning to do anything more with this so the draft was just a mistake on my part, sorry for the confusion.

ashmaroli:
No worries, @Gyrohammer. The very first time is always filled with apprehensions. :)

Regarding the proposed changes, I think `gem update --system` wouldn't be necessary if the user is installing Ruby 3.1 or newer.
Will you be able to check and make changes here if necessary?

PR #9304, opened 2023-02-14
Fix `jekyll serve --detach` with jekyll-sass-converter 3.x

### LINKED ISSUE #9234: [Bug]: serve --detach doesn't detach
### Operating System

macOS 12.6.2

### Ruby Version

ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-darwin21]

### Jekyll Version

jekyll 4.3.1

### GitHub Pages Version

_No response_

### Expected Behavior

I expected following commands to work after jekyll detaches:
``` sh
bundle exec jekyll serve --detach && echo "jekyll detached"
```

### Current Behavior

Jekyll keeps using terminal's stdin, stdout.

``` console
$ bundle exec jekyll serve --detach && echo "jekyll detached"
Configuration file: /path/to/site/_config.yml
To use retry middleware with Faraday v2.0+, install `faraday-retry` gem
            Source: /path/to/site
       Destination: /path/to/site/_site
      Generating...
                    done in 1.411 seconds.
 Auto-regeneration: disabled when running server detached.
    Server address: http://127.0.0.1:4000/
Server detached with pid '77748'. Run `pkill -f jekyll' or `kill -9 77748' to stop the server.
```

If I press Ctrl-C, jekyll produces fatal crash:

- Crash log: [crash.log](https://github.com/jekyll/jekyll/files/10312418/crash.log)
- DiagnosticReports: [ruby-2022-12-28-145100.txt](https://github.com/jekyll/jekyll/files/10312408/ruby-2022-12-28-145100.txt)

### Relevant log output

_No response_

### Code Sample

_No response_

ntkme:
This is a 🐛 bug fix.

## Summary

This PR fixes the issue that `Process.fork` + `Process.detach` would not detach correctly and may crash ruby when subprocess of sass-embedded (part of jekyll-sass-converter) is still running. The problem is that sass-embedded runs a child process and registers an `at_exit` hook to stop the child process, and the `at_exit` hook ran when main process exits but also get inherited by the forked process and would run again when forked process exits.

This PR changes it so that any `at_exit` hook would not run when main process exits, but then would only run when the forked process exits.

## Context

-  Fixes #9234

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -274,12 +274,19 @@ module Jekyll
         def boot_or_detach(server, opts)
           if opts["detach"]
             pid = Process.fork do
+              # Detach the process from controlling terminal
+              $stdin.reopen("/dev/null", "r")
+              $stdout.reopen("/dev/null", "w")
+              $stderr.reopen("/dev/null", "w")
               server.start
             end
 
             Process.detach(pid)
             Jekyll.logger.info "Server detached with pid '#{pid}'.",
                                "Run `pkill -f jekyll' or `kill -9 #{pid}' to stop the server."
+
+            # Exit without running `at_exit` inherited by the forked process.
+            Process.exit! 0
           else
             t = Thread.new { server.start }
             trap("INT") { server.shutdown }
--- a/test/test_commands_serve.rb
+++ b/test/test_commands_serve.rb
@@ -313,4 +313,34 @@ class TestCommandsServe < JekyllUnitTest
       @merc.execute(:serve, "watch" => false)
     end
   end
+
+  context "using --detach" do
+    setup do
+      skip_if_windows "fork is not supported on Windows"
+      skip("fork is not supported by JRuby") if jruby?
+
+      @temp_dir = Dir.mktmpdir("jekyll_serve_detach_test")
+      @destination = File.join(@temp_dir, "_site")
+      Dir.mkdir(@destination) || flunk("Could not make directory #{@destination}")
+    end
+
+    teardown do
+      FileUtils.remove_entry_secure(@temp_dir, true)
+    end
+
+    should "fork into daemon process" do
+      process, output = Jekyll::Utils::Exec.run("jekyll", "serve",
+                                                "--source", @temp_dir,
+                                                "--dest", @destination,
+                                                "--detach")
+
+      re = %r!Server detached with pid '(?<pid>\d+)'!
+
+      assert_match re, output
+      assert_equal 0, process.exitstatus
+
+      pid = re.match(output)["pid"].to_i
+      Process.kill 9, pid
+    end
+  end
 end
--- a/test/test_commands_serve.rb
+++ b/test/test_commands_serve.rb
@@ -340,6 +340,8 @@ class TestCommandsServe < JekyllUnitTest
       assert_equal 0, process.exitstatus
 
       pid = re.match(output)["pid"].to_i
+      assert pid > 1
+
       Process.kill 9, pid
     end
   end


ntkme:
@ashmaroli This is a bug related to jekyll-sass-converter 3.x upgrade. Would you mind take a look?

ashmaroli:
Hello @ntkme,
Is it possible to add a test regarding this? (I am on a Windows machine and therefore cannot test this manually).
Also, curious about the effect when user locks to older version of the sass-converter that doesn't involve `sass-embedded`.

ntkme:
> Is it possible to add a test regarding this? (I am on a Windows machine and therefore cannot test this manually).

Because this involves forking, we will have to write a test to actually launch a real Jekyll process, capture and parse the pid from stdout, and then check main pid has stopped with exit code 0 and new pid is running.

> Also, curious about the effect when user locks to older version of the sass-converter that doesn't involve `sass-embedded`.

After forking, the old main process will exit normally as it has nothing more to do. The old version does not have any `at_exit` hook so this PR basically has no impact to it.

ashmaroli:
> Because this involves forking, we will have to write a test to actually launch a real Jekyll process, capture and parse the pid from stdout, and then check main pid has stopped with exit code 0 and new pid is running.

Yes @ntkme, that is the basic idea. Will you be able to translate that into code? If the current implementation (of the lib code) doesn't have a suitable interface for testing, you may extract the entire conditional branch into a new (preferably private) method for easier testing.

ntkme:
@ashmaroli Test added. Also fixed another issue discovered in test when `--detach` is issued from non-tty, that stdio inherited from main process would cause a deadlock on test process reading as the file descriptors remains open after main process exits.

lib/jekyll/commands/serve.rb
```diff
@@ -274,12 +274,19 @@ def launch_browser(server, opts)
         def boot_or_detach(server, opts)
           if opts["detach"]
             pid = Process.fork do
+              # Detach the process from controlling terminal
+              $stdin.reopen("/dev/null", "r")
+              $stdout.reopen("/dev/null", "w")
+              $stderr.reopen("/dev/null", "w")
```
ashmaroli, line 280: Is it possible to switch this out with the [`Open3`](https://ruby-doc.org/stdlib-2.5.0/libdoc/open3/rdoc/Open3.html) module?

ntkme: If we want to have a simpler or cleaner implementation we have to change the semantics of how this work.

Currently, `--detach` runs these steps:

1. Build the site.
2. Fork into child process and start server.
3. Print the child process PID.
4. Exit main process.

We have several options but they would change the semantics:

The simplest solution is to use `Process.daemon`. It is a single call that does fork and detach stdio then continue rest of the code in forked process. However:
    - If we call `Process.daemon` on main, the problem is that we cannot get and print PID in main process. A typical practice of using this method would be adding a `--pid-file` option to let forked process to write its PID into a file.
    - We can also call `Process.daemon` in the forked process to use it for detach stdio, the problem is that it would fork again and change PID.

Therefore, if we stick with fork based implementation, this is the best we can do unless we agree to change the semantics of how this option works today.

Another option would be replace the fork with a real ruby process that just runs webrick, with two different possible implementations:

1. Serialize webrick options and start webrick with a ruby one-liner using `Process.spawn` (or `open3`).
2. Add an option on Jekyll command line that start webrick without building the site. Spawn a Jekyll process with the same ARGV extended with option that disable the build step, so that it just start the server.

ashmaroli: A huge thank you for giving me a detailed, well-thought out reply, @ntkme 🙏 🙇

Since I don't intend to change much about this, especially since this is particularly unfamiliar territory for me, the current implementation here will do.

ashmaroli:
LGTM!
Requesting a third set of :eyes: from either @mattr- or @parkr for final review.

test/test_commands_serve.rb
```diff
@@ -313,4 +313,34 @@ def serve(opts)
...
+                                                "--source", @temp_dir,
+                                                "--dest", @destination,
+                                                "--detach")
+
+      re = %r!Server detached with pid '(?<pid>\d+)'!
+
+      assert_match re, output
+      assert_equal 0, process.exitstatus
+
+      pid = re.match(output)["pid"].to_i
```
parkr, line 342: Shall we check that this PID isn’t 0, 1, or -1? Don’t want to kill the init process in our test suite…

ntkme: Added an assertion for `pid > 1`.

test/test_commands_serve.rb
```diff
@@ -313,4 +313,36 @@ def serve(opts)
...
+      assert pid > 1
+
+      Process.kill 9, pid
```
parkr, line 345: This will raise if the process has already exited but that's fine since we'd likely want a test failure anyway.

parkr (approved):

parkr:
@jekyllbot: merge +bug

Issue #6430, opened 2017-10-17
parsing error while highlighting java code snippets


Mellivora123:
- [x] I believe this to be a bug, not a question about using Jekyll.
- [ ] I updated to the latest Jekyll (or) if on GitHub Pages to the latest `github-pages`
- [x] I ran `jekyll doctor` to check my configuration
- [ ] I read the CONTRIBUTION file at https://jekyllrb.com/docs/contributing/
- [ ] This is a feature request.

---

- [ ] I am on (or have tested on) ***macOS*** 10+
- [x] I am on (or have tested on) ***Debian/Ubuntu*** GNU/Linux
- [ ] I am on (or have tested on) ***Fedora*** GNU/Linux
- [ ] I am on (or have tested on) ***Arch*** GNU/Linux
- [ ] I am on (or have tested on) ***Other*** GNU/Linux
- [ ] I am on (or have tested on) ***Windows*** 10+

## My Reproduction Steps
create a new post, then add the following code snippet as the post content:  

<pre>
```java
import java.lang.reflect.Field;
import java.util.List;

public class FieldSpy<T> {
    public boolean[][] b = {{ false, false }, { true, true } };
    public String name  = "Alice";
    public List<Integer> list;
    public T val;

    public static void main(String... args) {
        try {
            Class<?> c = Class.forName(args[0]);
            Field f = c.getField(args[1]);
            System.out.format("Type: %s%n", f.getType());
            System.out.format("GenericType: %s%n", f.getGenericType());

        // production code should handle these exceptions more gracefully
        } catch (ClassNotFoundException x) {
            x.printStackTrace();
        } catch (NoSuchFieldException x) {
            x.printStackTrace();
        }
    }
}
```
</pre>

The error message:
> jekyll 3.5.1 | Error:  Liquid syntax error (line 6): Variable '{{ false, false }' was not properly terminated with regexp: /\}\}/

## The Output I Wanted

The normal java syntax highlighting code snipped.

pathawks:
You will need to surround the code with a `{% raw %}` tag to prevent it from being rendered as Liquid.

I believe this is documented. If not, please do open a PR.

Mellivora123:
I am using the [fenced code blocks way](https://help.github.com/articles/creating-and-highlighting-code-blocks/#fenced-code-blocks) to highlight the code snippet. The problem is the parser(the *liquid* parser or the plugin or whatever) can not process `{{` in the code snippet correctly.

[...]

PR #9308, opened 2023-02-21
create new tag `include_raw` to automatically wrap file contents in {% raw %} / {% endraw %}

klandergren:

## Summary

New tag `include_raw` that just overrides `read_file` to wrap the content in raw liquid tags. Does not attempt to mess with the `OptimizedIncludeTag` approach or worry about registering a file path with the liquid renderer that likely shouldn't be rendered directly.

## Context

I want to include a file containing Go code (e.g. `s1 := []int{0, 1, 2}`) that renders without a liquid parse error.

Other users seem to encounter this issue:

* [Is it possible to have Jekyll automatically render code blocks raw?](https://stackoverflow.com/questions/40572336/is-it-possible-to-have-jekyll-automatically-render-code-blocks-raw)
* related to issue #6430
* would affect #6439

I am creating this as a draft PR before doing more (doc updates, tests) to check:
* is this functionality desired?
* if so, is this the preferred approach?
* any other feedback before continuing?

I am also happy to have a maintainer just take this PR "as is" and modify however desired.

Thank you for Jekyll!

### DIFF
--- a/lib/jekyll/tags/include.rb
+++ b/lib/jekyll/tags/include.rb
@@ -268,8 +268,17 @@ module Jekyll
         path.delete_suffix("/#excerpt")
       end
     end
+
+    class IncludeRawTag < IncludeTag
+      def read_file(file, context)
+        contents = File.read(file, **file_read_opts(context))
+        "{% raw %}#{contents}{% endraw %}"
+      end
+    end
+
   end
 end
 
 Liquid::Template.register_tag("include", Jekyll::Tags::OptimizedIncludeTag)
 Liquid::Template.register_tag("include_relative", Jekyll::Tags::IncludeRelativeTag)
+Liquid::Template.register_tag("include_raw", Jekyll::Tags::IncludeRawTag)


ashmaroli:
Hello @klandergren,
Does Liquid actually complain when it encounters the Go code mentioned in the opening comment (`s1 := []int{0, 1, 2}`)? If not, it is a poor choice of an example.
Regarding the implementation itself, recent versions of Jekyll doesn't actually use the `Tags::IncludeTag` directly. Consequently, contents of an include_file are not stashed within Liquid context. Include files are nowadays loaded as `Jekyll::Inclusion` instances.
Regarding whether the maintainers are interested in accepting this feature, the current answer would be in the negative because the added complexity isn't crucial. (Users can already add the `raw` instruction where desired. Why move a non-repeating block of content into a partial and have it rendered via another instruction..?, etc..etc..). That said, we're open to considering this if you can provide a compelling justification.

klandergren:
re: example
You are correct that it was a poor choice as it does not replicate the issue—please accept my apologies. I misremembered how the Liquid syntax error is generated. It requires a `{{`. I meant to use an example like: `coords := []coord{{3, 2}, {1, 2}, {1, 1}}` (where `coord` is a struct defined elsewhere).

re: compelling justification
I have a site which contains a git submodule with go code. I wish to render some of the `.go` files as-is with syntax highlighting. I do not want to edit each included file to contain `raw / endraw` instructions. Additionally, those files may update when the submodule updates.

I believe the only ways to accomplish this currently would be:
* edit the `.go` file directly to add the `raw / endraw` instructions
* not use `include` and copy the contents of the file and embed them directly with `raw / endraw` instructions

Hence the creation of `include_raw`.

I created an [example repository](https://github.com/klandergren/include-raw-use-case) which is an out-of-the-box Jekyll site (e.g. just `bundle exec jekyll serve` to run) to better demonstrate the issue.

Please note:
* this is demo site is built [using the branch of this PR](https://github.com/klandergren/include-raw-use-case/blob/main/Gemfile#L10) to demonstrate `include_raw` usage
* the relevant files are:
  * [main.go](https://github.com/klandergren/include-raw-use-case/blob/main/_includes/golang/main.go), the file we wish to include but not edit with `raw` instruction tags. It contains a line that will generate a Liquid syntax error if included.
  * [2023-02-22-liquid-syntax-error.markdown](https://github.com/klandergren/include-raw-use-case/blob/main/_posts/2023-02-22-liquid-syntax-error.markdown), the post showing existing behavior, with Liquid syntax error info
  * [2023-02-22-include-raw-no-error.markdown](https://github.com/klandergren/include-raw-use-case/blob/main/_posts/2023-02-22-include-raw-no-error.markdown), the post showing usage with `include_raw`
* imagine, if you will, that `_includes/golang` is a submodule that is not easily edited, or too onerous to maintain a version containing files with `raw / endraw` usage

If there is an alternative way to accomplish this goal I am open to it!

re: implementation details with `Jekyll::Inclusion`
Noted. I suggest we revisit after proving / disproving justification.

Thank you, and I apologize again for the waste of time with the poor example choice.

ashmaroli:
> I have a site which contains a git submodule with go code. I wish to render some of the `.go` files as-is with syntax highlighting.

I see. So, the source-content is not only outside `_includes` directory, but outside the current `site.source` entirely. 😞 
I guess you can modify the example site to see if the current implementation in this branch can support your use-case.
(*FYI: there are already plugins created by the community to support include_files outside conventional locations. Search `rubygems.org` to find them.*)

Hmm.. the suggested use-case doesn't have enough gravitas for being part of core codebase.
Maybe you can convince us with your implementation. Include tag is generally slow because of additional Liquid parsing and rendering. However, the `include_raw` tag can be fast because of absence of Liquid parsing and rendering. On that note, the injection of `{% raw %} {% endraw %}` by the proposed tag can be avoided as well.

klandergren:
Thanks for your advice and expertise @ashmaroli—weighing the options I think it is best to close this PR. I released the functionality as a separate gem [jekyll-include-raw](https://github.com/klandergren/jekyll-include-raw). Thanks again for your consideration, and for maintaining Jekyll!

PR #9333, opened 2023-04-04
cmd/doctor: Remove duplicate source paths from conflicting_urls check

iBug:

## Summary

If a single source file is somehow duplicated in Jekyll's build pipeline, it irritates `jekyll doctor` to complain about `Conflict: The following destination is shared by multiple files`, which doesn't make much sense.

This isn't normally possible, but my workaround for a missing feature for the `jekyll-feed` plugin relies on this.

## Context

This PR is one way to fix jekyll/jekyll-feed#390.

### DIFF
--- a/lib/jekyll/commands/doctor.rb
+++ b/lib/jekyll/commands/doctor.rb
@@ -129,7 +129,7 @@ module Jekyll
               next if allow_used_permalink?(thing)
 
               dest_path = thing.destination(site.dest)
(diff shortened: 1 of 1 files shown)


ashmaroli:
Hello @iBug,
Since source files do not get duplicated *normally*, I believe this patch has no place in Jekyll Core.
I suggest you move this patch into *your workaround plugin*.

My stand on this PR is: `wont-merge`.

PR #9334, opened 2023-04-06
Update blank_template for Jekyll

danielhaim1:
## Summary

Updated config.yml, index.md, header.html, main.scss, and default.html to include additional site settings, improved ordering and variable names, and better styling.

## Context

- Updated the config.yml file to include additional site settings and build settings, as well as comments to explain each setting. Added a basic description to the index.md file.
- Improved the header.html file to include a better ordering of the meta tags and added a default_page_title variable to the site settings.
- Improved the main.scss file to include more robust font fallbacks and better variable names.
- Updated the default.html file to include a more robust ordering of meta tags and improved styling links.

### DIFF
--- a/lib/blank_template/_config.yml
+++ b/lib/blank_template/_config.yml
@@ -1,3 +1,36 @@
+# Site Information
+title: "My Jekyll Site" # the name of your site, e.g. ACME Corp.
+description: "This is a Jekyll site" # a brief description of your site
+
+# Build Configuration
 url: "" # the base hostname & protocol for your site, e.g. http://example.com
 baseurl: "" # the subpath of your site, e.g. /blog
-title: "" # the name of your site, e.g. ACME Corp.
+source: "." # the source directory for your site
+destination: "./_site" # the destination directory for your site
+lang: "en-US" # the language code for your site, e.g. en-US
+
+# Excluded Files
+# Add any files or directories that should be excluded from the Jekyll build process here
+exclude:
+  - Gemfile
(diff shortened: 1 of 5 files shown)


ashmaroli:
Thank you for taking the time to propose changes, @danielhaim1.
But the blank template was meant be as non-opinionated / codeless as possible; not inject more code.

Sorry, but rejecting this.

danielhaim1:
@ashmaroli, I suggest we reconsider the decision to reject the updates. These updates do not make the code more complex and actually include improvements to naming conventions and commented-out examples that aim to reduce reliance on documentation.

Currently, there is an inconsistency between the `default.html` and `config.yml` files with regard to the `lang` option. Although `{{ site.lang | default: "en-US" }}` is present in `default.html`, the `lang` option is not included in `config.yml`.

Here are the key points of the update:
- The update addresses this inconsistency and adds basic technical examples to help users better understand the template's functionality.
- `config.yml` is updated to include the `lang` option and other basic options. It clarifies how `{{ site.lang | default: "en-US" }}` is handled using the `site.lang` already used in `default.html`.

Another simple update has been made to the `main.scss` file for better clarity. The file now includes comments to explain its purpose and how it should be used:

Before:
```scss
---
---

@import "base";
```

After:
```scss
---

# Leave to compile

---

@import "base";
```

The comments make it clear that the `---` should be left and all SCSS styles should be imported in their respective files and included through the `@import` directive.

I have made some reduction to the last update. Let me know your thoughts.

PR #9338, opened 2023-04-08
Improve accessibility of the docs

danielhaim1:
## Summary

Improvements made to the docs include changing `<!DOCTYPE HTML>` to `<!DOCTYPE html>` and `en-US` to `en`, adding missing accessibility, documentation and aria-labels, implementing structured data to templates, adding `target="_blank"` to some external links, and adding `defer` attribute for faster page load.

Still making improvements, but I realize I should start chunking these updates down so that the team can review them more efficiently, and to reduce their workload.

This is a 🐛 bug fix.
This is a 🔨 code refactoring.

### DIFF
--- a/docs/_includes/anchor_links.html
+++ b/docs/_includes/anchor_links.html
@@ -11,9 +11,7 @@ const anchorForId = (id) => {
 
 /* Finds all headers of the specified level within the given element, and adds a permalink to each header*/
 const linkifyAnchors = (level, containingElement) => {
-  /* Retrieve all headers with the specified level as an Array*/
   const headers = Array.from(containingElement.getElementsByTagName(`h${level}`));
-  /* For each header with an ID, add a permalink*/
   headers.forEach((header) => {
     if (header.id) {
       header.appendChild(anchorForId(header.id));
--- a/docs/_includes/anchor_links.html
+++ b/docs/_includes/anchor_links.html
@@ -1,33 +1,40 @@
 <script>
-  var anchorForId = function (id) {
-    var anchor = document.createElement("a");
-    anchor.className = "header-link";
-    anchor.href      = "#" + id;
-    anchor.innerHTML = "<span class=\"sr-only\">Permalink</span><i class=\"fa fa-link\"></i>";
-    anchor.title = "Permalink";
-    return anchor;
-  };
+// Creates an anchor element with the given ID and link for the permalink
+const anchorForId = (id) => {
+  const anchor = document.createElement("a");
+  anchor.className = "header-link";
+  anchor.href = `#${id}`;
+  anchor.innerHTML = `<span class="sr-only">Permalink</span><i class="fa fa-link"></i>`;
+  anchor.title = "Permalink";
+  return anchor;
+};
 
-  var linkifyAnchors = function (level, containingElement) {
-    var headers = containingElement.getElementsByTagName("h" + level);
-    for (var h = 0; h < headers.length; h++) {
-      var header = headers[h];
-
-      if (typeof header.id !== "undefined" && header.id !== "") {
-        header.appendChild(anchorForId(header.id));
-      }
+// Finds all headers of the specified level within the given element, and adds a permalink to each header
+const linkifyAnchors = (level, containingElement) => {
+  // Retrieve all headers with the specified level as an Array
+  const headers = Array.from(containingElement.getElementsByTagName(`h${level}`));
+  // For each header with an ID, add a permalink
+  headers.forEach((header) => {
+    if (header.id) {
+      header.appendChild(anchorForId(header.id));
     }
-  };
+  });
+};
 
-  document.onreadystatechange = function () {
-    if (this.readyState === "complete") {
-      var contentBlock = document.getElementsByClassName("docs")[0] || document.getElementsByClassName("news")[0];
-      if (!contentBlock) {
-        return;
-      }
-      for (var level = 1; level <= 6; level++) {
-        linkifyAnchors(level, contentBlock);
-      }
+// Executes the function when the document is ready
+document.onreadystatechange = () => {
+  if (document.readyState === "complete") {
+    // Find the first element with the class "docs" or "news"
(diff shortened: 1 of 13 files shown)


Octolannie9048382 (approved):

danielhaim1:
Resolved Previous Issues and Awaiting Test Results.

- Converted inline comments from single-line (`//`) to block style (`/* */`).
- Amended double quotation marks in `news_item_archive.html` to rectify an error.

danielhaim1:
It appears that the issue has been successfully resolved. This marks a significant improvement in accessibility, complemented by schema integration.

docs/_includes/anchor_links.html
```diff
@@ -1,33 +1,38 @@
...
-        header.appendChild(anchorForId(header.id));
-      }
+/* Finds all headers of the specified level within the given element, and adds a permalink to each header*/
+const linkifyAnchors = (level, containingElement) => {
+  /* Retrieve all headers with the specified level as an Array*/
+  const headers = Array.from(containingElement.getElementsByTagName(`h${level}`));
+  /* For each header with an ID, add a permalink*/
+  headers.forEach((header) => {
+    if (header.id) {
+      header.appendChild(anchorForId(header.id));
```
mattr-, line 19: The comment at the top of the method is great. The comments that duplicate the code in the function aren't as great because the code is straightforward enough to read it and know what it's doing without the comment. What do you think about letting the code stand alone where it's completely obvious?
```suggestion
/* Finds all headers of the specified level within the given element, and adds a permalink to each header*/
const linkifyAnchors = (level, containingElement) => {
  const headers = Array.from(containingElement.getElementsByTagName(`h${level}`));
  headers.forEach((header) => {
    if (header.id) {
      header.appendChild(anchorForId(header.id));
```

docs/_includes/anchor_links.html
```diff
@@ -1,33 +1,38 @@
...
+document.onreadystatechange = () => {
+  if (document.readyState === "complete") {
+    /* Find the first element with the class "docs" or "news" */
+    const contentBlock = document.getElementsByClassName("docs")[0]
+      ?? document.getElementsByClassName("news")[0];
+    /* If there is no content block, exit the function */
+    if (!contentBlock) { return; }
+    /* For each level of headers, add a permalink to each header */
+    for (let level = 1; level <= 6; level++) {
+      linkifyAnchors(level, contentBlock);
```
mattr-, line 34: Similar thing here
```suggestion
    const contentBlock = document.getElementsByClassName("docs")[0]
      ?? document.getElementsByClassName("news")[0];
    if (!contentBlock) { return; }
    for (let level = 1; level <= 6; level++) {
      linkifyAnchors(level, contentBlock);
```

docs/_includes/docs_contents_mobile.html
```diff
@@ -1,15 +1,15 @@
...
+    {% for section in site.data.docs_nav %}
+      <optgroup label="{{ section.title }}">
+        {%- for item in section.docs -%}
+          {% assign page = site.docs | where: "url", item.link | first %}
+          <option value="{{ page.url | relative_url }}">
+            {{- page.menu_name | default: page.title -}}
+          </option>
+        {%- endfor %}
+      </optgroup>
+    {% endfor %}
```
mattr-, line 13: I'm not following the reasoning for the re-indentation work in this PR. We've not indented the liquid in most pages because it's not present in the final site HTML, where as indenting it along with the HTML gives disjointed indentation in the final output.

mattr- (commented):
Would it be possible to break this down further? Some things are good, like the improved accessibility, the better javascript, etc. However, it's still too wide of a PR to be merged as is. 

Ideally, I'd see something like two or three PRs for this.
 - Javascript Improvements
 - Accessibility Improvements
 - Enhanced schema usage for the docs site

The JS and accessibility improvements look like they could go hand in hand and if it makes more sense to combine those two things into a single PR, I think that's fine.

docs/_includes/docs_contents_mobile.html
danielhaim1: Hello @mattr-, the indentation has been adjusted solely to enhance readability in my editor.

docs/_includes/anchor_links.html
danielhaim1: Hey @mattr-, agreed. I've added comments for an easier understanding of the script's purpose. Since I wasn't certain who on the team would review it, I'd typically suggest removing these comments before merging and figured we might be using some sort of compressor that removes these comments anyway.

Following your suggestion.

danielhaim1:
> Would it be possible to break this down further? Some things are good, like the improved accessibility, the better javascript, etc. However, it's still too wide of a PR to be merged as is.
> 
> Ideally, I'd see something like two or three PRs for this.
> 
> * Javascript Improvements
> * Accessibility Improvements
> * Enhanced schema usage for the docs site
> 
> The JS and accessibility improvements look like they could go hand in hand and if it makes more sense to combine those two things into a single PR, I think that's fine.

Absolutely, I concur. I'm planning additional enhancements in a separate PR, so your recommendations are helpful as I'd like to make it as easy as possible for us to review it. Could we proceed with merging this one, and then for the subsequent updates, I'll implement the splitting as recommended? Your input would be appreciated.

mattr- (approved):
Yeah, I think we can get this merged for now. Thanks for improving our accessibility. ❤️ 

@jekyllbot: merge +docs

danielhaim1:
> Yeah, I think we can get this merged for now. Thanks for improving our accessibility. ❤️
> 
> @jekyllbot: merge +docs

🎉

danielhaim1:
> Yeah, I think we can get this merged for now. Thanks for improving our accessibility. ❤️
> 
> @jekyllbot: merge +docs

🎉

heironimus:
@danielhaim1 This PR introduced #9527. Reverting it fixes it. #9529 attempts to fix the issue by just reverting one file, but does not fix it.

PR #9348, opened 2023-04-16
Update convert-site-to-jekyll.md

j15k:
This is a 🐛 bug fix.

## Summary

Fixed a parsing error created by the feed example. Also see https://validator.w3.org/feed/docs/warning/MissingAtomSelfLink.html.

### DIFF
--- a/docs/_tutorials/convert-site-to-jekyll.md
+++ b/docs/_tutorials/convert-site-to-jekyll.md
@@ -426,7 +426,7 @@ layout: null
 ---
 
 <?xml version="1.0" encoding="UTF-8" ?>
-<rss version="2.0">
+<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
 
     <channel>
         <title>{{ site.title }}</title>


docs/_tutorials/convert-site-to-jekyll.md
```diff
@@ -426,7 +426,7 @@ layout: null
 ---
 
 <?xml version="1.0" encoding="UTF-8" ?>
-<rss version="2.0">
+<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
```
parkr, line 429: Is this really an Atom file? Seems like RSS XML should work without making an Atom declaration.

j15k: As https://validator.w3.org/feed/docs/warning/MissingAtomSelfLink.html explains:

> According to the [RSS Advisory Board](http://www.rssboard.org/charter)'s [Best Practices Profile](http://www.rssboard.org/rss-profile-1#namespace-elements-atom-link), identifying a feed's URL within the feed makes it more portable, self-contained, and easier to cache. For these reasons, a feed should contain an atom:link used for this purpose.

parkr (approved):

parkr:
@jekyllbot: merge +docs

PR #9355, opened 2023-04-24
Use FileUtils#remove_entry_secure in StaticFile#write

edward:
Sometimes the `dest_path` in Jekyll::StaticFile#write is a directory. When this happens, `FileUtils.rm(dest_path)` will raise an exception.

Using `FileUtils.remove_entry_secure(dest_path)` instead will not raise an exception and remove (aka “unlink”) file system entries if the file `dest_path` points to is a directory or a non-directory file.

This is a 🐛 bug fix.


parkr:
Why would a static file be a directory?

PR #9187, opened 2022-11-18
Optimize `Site#each_site_file`

ashmaroli:
- This is a 🔨 code refactoring change.

## Summary

`Site#each_site_file` reduced in performance in Jekyll v4.3.0 owing to additional checks to filter duplicate entries of static files from a collection.

This pull request optimizes the regression by avoiding duplicate entries of static files itself thereby removing the need for additional checks and filtering.

## Notes

Consider backporting to the `4.3-stable` branch.

PR #9357, opened 2023-04-29
Revert "Optimize `Site#each_site_file` (#9187)"

DavidS:
This reverts commit b2891a476e1f6a6cfccda7772fb4f5f2f7d166c5.

This is a 🐛 bug fix.

## Summary

Since https://github.com/jekyll/jekyll/pull/9187 static files from collections don't get written out anymore.

## Context

[My gallery plugin](https://github.com/DavidS/cheesy-gallery) injects images and thumbnails to write out as static files into collections. With the optimization from #9187, they do not get rendered anymore. Either this revert, or adding `collections.each_value { |coll| coll.files.each { |doc| yield(doc) if doc.write? } }`  fixes the issue for me.

### DIFF
--- a/lib/jekyll/site.rb
+++ b/lib/jekyll/site.rb
@@ -363,6 +363,7 @@ module Jekyll
       pages.each { |page| yield page }
       static_files.each { |file| yield(file) if file.write? }
       collections.each_value { |coll| coll.docs.each { |doc| yield(doc) if doc.write? } }
+      collections.each_value { |coll| coll.files.each { |doc| yield(doc) if doc.write? } }
     end
 
     # Returns the FrontmatterDefaults or creates a new FrontmatterDefaults


DavidS:
Updated to the second option I've found to avoid the performance impact that #9187 fixed

ashmaroli:
Hello @DavidS
It's unfortunate that the released implementation doesn't play well with your plugin.

Theoretically, the change you have proposed (as per current diff) shouldn't be necessary because the `static_files` reference in ln#364 **is supposed to** be a real-time reference to *all static files* in the site, including those within collections.

So, the bug is either in Jekyll's `site.static_files` or in your generator plugin.

lib/jekyll/site.rb
```diff
@@ -363,6 +363,7 @@ def each_site_file
       pages.each { |page| yield page }
       static_files.each { |file| yield(file) if file.write? }
       collections.each_value { |coll| coll.docs.each { |doc| yield(doc) if doc.write? } }
```
shushugah, line 365: This would have a lower cyclomatic complexity, what do you think? 
```ruby
 collections.each_value do  |coll| 
   if doc.write?
     coll.docs.each(&method(:yield))
     coll.files.each(&method(:yield)) 
  end
end
```

PR #9376, opened 2023-05-30
Document Exclude behavior, Include overrides and the default exclusion list

### LINKED ISSUE #9372: [Docs]: Document whether `exclude` in `_config.yml` replaces or appends to the default exclusions
## Motivation

At the moment it is not clear whether by setting `exclude` to my own list, I am disabling the default exclusions.
If `exclude` replaces the default exclusions, it may be that I do not need some of the `include`s I added for file which were excluded by the default exclusions. Or if I need to repeat those default exclusions to not let undesired files into my build.
If `exclude` appends to the default excludes, the behavior is IMO more intuitive and desirable, but it should be commented so that a user can be sure that is the case.

## Suggestion

Expand the Exclude row in https://jekyllrb.com/docs/configuration/options.
For Exclude, either mention it is added to the default exclusions or replaces them.
Also mention the default exclusions. I know dotfiles are excluded (from the Include row), what else? Is there documentation for this elsewhere that I missed?

I am happy to PR this myself, but I am lacking the information. Can anyone help me determine which is the case with `exclude` and what are the default exclusions?

TomasHubelbauer:

## Summary

I've documented the difference between Exclude in Jekyll 3 and Jekyll 4 as well as the fact that Include can override the Exclude entries.

The exclusion list is taken from a `_config.yml` as generated via `jekyll new`.

image

## Context

Closes #9372.

### DIFF
--- a/docs/_data/config_options/global.yml
+++ b/docs/_data/config_options/global.yml
@@ -43,7 +43,8 @@
     <br />
     In Jekyll 3, the `exclude` configuration option replaces the default exclusion list.
     <br />
-    In Jekyll 4, it adds to it instead and the `include` option can be used to override the default exclusion list entries.
+    In Jekyll 4, user-provided entries get added to the default exclusion list instead and the `include` option can be
+    used to override the default exclusion list entries.
     <br />
     The default exclusions are found in <code>_config.yml</code> as created by <code>jekyll new</code>:
     <ul>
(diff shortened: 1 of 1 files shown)


ashmaroli:
@TomasHubelbauer The term "Include" can be  ambiguous as to what exactly is being intended &mdash; config key `include` / files inside `_includes` / the literal English word "include".
Therefore, readability may be improved by using explicit phrases such as `configuration option include` (where `include` is within single backticks).

TomasHubelbauer:
@ashmaroli great point! I have turned all instances of `Include` and `Exclude` where they refer to the configuration options to inline code spans and improved the prose around them.

ashmaroli (approved):

PR #9390, opened 2023-06-24
Update index.md

Ronakmrup:
Few editorial changes on the language of the page.


mattr-:
I don't like the switch from active to passive voice here, so I won't be merging this PR. Thanks for taking the time to submit a contribution. 🫶

PR #9423, opened 2023-08-13
Mention the front matter requirement in pages.md

IFcoltransG:

## Summary

Specify that markdown files should have front matter, for them to be processed as pages, in case someone is confused why their page isn't showing properly.

## Context

A user just wanting to host a folder of markdown files with a static site may think that 'pages' is the simplest option. Without the docs mentioning front matter, they might not realise that front matter is necessary for those pages to turn into html.


parkr (approved):

parkr:
@jekyllbot: merge +doc

PR #9441, opened 2023-09-04
Test suite uses shoulda-context only.

voxik:

## Summary

Replace shoulda by shoulda-context

## Context

AFAICT, only shoulda-context is used for Jekyll test suite. Therefore replacing `shoulda` by `shoulda-context` should:

1. Remove unneeded dependency (should, shoulda-matchers)
2. Remove ambiguity (because are shoulda-matchers allowed or no?)
3. Improve run time (less files to install, less libraries to load, smaller memory footprint).

Apart from this, I remember times when `shoulda` gem was not really maintained. The stable version still references to `shoulda-matchers 4`, while `shoulda-matchers 5` were released more then 2 years ago.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -48,7 +48,7 @@ group :test_legacy do
   gem "minitest"
   gem "minitest-profile"
   gem "minitest-reporters"
-  gem "shoulda"
+  gem "shoulda-context"
   gem "simplecov"
 end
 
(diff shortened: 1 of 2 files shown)


mattr- (approved):
Thanks!

@jekyllbot: merge +dev

PR #9445, opened 2023-09-13
Configure dependabot to handle rubygems dependencies

mattr-:
This is a 🔨 code refactoring.

## Summary

This updates our dependabot config for jekyll to allow it to bump ruby dependencies. It also attempting to configure update grouping so we can potentially avoid large numbers of PRs that bump individual gems in favor of larger group updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,7 +10,3 @@ updates:
     directory: /
     schedule:
       interval: weekly
-    groups:
-      rubygems:
-        patterns:
-        - '*'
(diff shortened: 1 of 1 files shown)


parkr (approved):
Maybe add jekyll/core as a reviewer?

We can try the groupings for a bit. If we find we often want to accept only part of the group then we can try individual updates. I suspect the first wave of updates will be the biggest.

parkr:
> Maybe add jekyll/core as a reviewer?

Specifically, add as a reviewer of dependabot PRs :)

mattr-:
Good idea. I think I'll also ungroup for now and then once we're in a place where we're not having to update the world all the time, we can attempt grouping again.

mattr-:
Added jekyll/core as a reviewer and did the ungrouping. CI is clean. Going to use `@parkr`'s prior approval and merge it.

@jekyllbot: merge +dev

PR #9235, opened 2022-12-28
Add GitHub Pages helpful notes

fulldecent:
This is a 🔦 documentation change.

## Summary

Adds a helpful setup note for people publishing with GitHub Pages.

## Context

GitHub Pages will build your project using the `github-pages` gem and without any regard for your repository's `Gemfile.lock` file. This PR adds such a note to documentation and advises how you might work with that.

PR #9468, opened 2023-10-21
Print out Gemfile.lock in CI runs

fulldecent:
This is a 🐛 bug fix.

Related: https://github.com/ruby/setup-ruby/issues/537

Related: https://github.com/jekyll/jekyll/pull/9235


mattr-:
I'll close this one since I feel like bundler's output is good enough and #9469 fixes the issue between our version of cucumber and newer versions of activesupport.

PR #9473, opened 2023-10-24
chore: update showcases

SethFalco:

## Summary

When browsing the showcase, there was a broken link. While I was at it, I figured it was worth checking some other links too:

* Replace links to where they redirect if a redirect occurs. (Excluding localization.)
* Remove locale from the URL, so it uses the preferred language of the user.
* Add a trailing slash on base URLs for consistency.
* Remove dead links.

### DIFF
--- a/docs/_data/showcase.yml
+++ b/docs/_data/showcase.yml
@@ -12,7 +12,7 @@
 #     - government
 
 - name: SiteLeaf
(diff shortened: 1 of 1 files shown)


mattr- (commented):
Thank you for updating all of these! :heart:

@jekyllbot: merge +docs

PR #9469, opened 2023-10-22
Keep activesupport at version 7.0.x

mattr-:
This is a 🐛 bug fix.

## Summary

Allowing activesupport to be bumped to version 7.1 with cucumber 5.1.2 causes our CI to fail. This PR keeps activesupport at version 7.0.x to prevent this for now.

## Context

Related: #9235
Related: #9468

PR #9482, opened 2023-11-03
updated gem active active support to 7.1.1

### LINKED ISSUE #9472: Update dependency constraint to allow for activesupport v7.1.1
Hey there! :wave:

I noticed that the constraint you have for activesupport doesn't allow for the latest version to be used.

The constraint I found was `~> 7.0.0`, and the latest version available is `7.1.1`.

kamrandb:

## Context Updated dependency constraint to allow for activesupport v7.1.1 

  Closes #9472 
  Fixes #9472 
  Resolves #9472

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ end
 #
 
 group :test do
-  gem "activesupport", "~> 7.0.0"
+  gem "activesupport", "~> 7.1.1"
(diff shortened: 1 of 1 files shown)


mattr-:
`activesupport` 7.1.1 causes issues with our cucumber test suite, which is why that dependency hasn't been updated yet. it was intentionally pinned in #9469

PR #9392, opened 2023-06-25
Add support for upcoming logger 1.4.3

kou:
This is a 🙋 feature or enhancement.

## Summary

logger 1.4.3 will have ruby/logger#85. It initializes a new instance variable in `Logger#initialize`. `Jekyll::Stevenson < ::Logger` doesn't use `super`. So the new instance variable isn't initialized in `Jekyll::Stevenson`. And it causes an exception:

```text
/tmp/local/lib/ruby/3.3.0+0/logger.rb:385:in `level': undefined method `[]' for nil (NoMethodError)

    @level_override[Fiber.current] || @level
                   ^^^^^^^^^^^^^^^
	from /tmp/local/lib/ruby/gems/3.3.0+0/gems/jekyll-4.3.2/lib/jekyll/log_adapter.rb:45:in `adjust_verbosity'
	from /tmp/local/lib/ruby/gems/3.3.0+0/gems/jekyll-4.3.2/lib/jekyll/configuration.rb:143:in `config_files'
	from /tmp/local/lib/ruby/gems/3.3.0+0/gems/jekyll-4.3.2/lib/jekyll.rb:118:in `configuration'
	from /tmp/local/lib/ruby/gems/3.3.0+0/gems/jekyll-4.3.2/lib/jekyll/command.rb:44:in `configuration_from_options'
	from /tmp/local/lib/ruby/gems/3.3.0+0/gems/jekyll-4.3.2/lib/jekyll/commands/serve.rb:83:in `block (2 levels) in init_with_program'
	from /tmp/local/lib/ruby/gems/3.3.0+0/gems/mercenary-0.4.0/lib/mercenary/command.rb:221:in `block in execute'
	from /tmp/local/lib/ruby/gems/3.3.0+0/gems/mercenary-0.4.0/lib/mercenary/command.rb:221:in `each'
	from /tmp/local/lib/ruby/gems/3.3.0+0/gems/mercenary-0.4.0/lib/mercenary/command.rb:221:in `execute'
	from /tmp/local/lib/ruby/gems/3.3.0+0/gems/mercenary-0.4.0/lib/mercenary/program.rb:44:in `go'
	from /tmp/local/lib/ruby/gems/3.3.0+0/gems/mercenary-0.4.0/lib/mercenary.rb:21:in `program'
	from /tmp/local/lib/ruby/gems/3.3.0+0/gems/jekyll-4.3.2/exe/jekyll:15:in `<top (required)>'
	from /tmp/local/bin/jekyll:25:in `load'
[...]

PR #9510, opened 2023-12-27
Cherry pick logger fix for Ruby 3.3

ntkme:
This PR ports #9392 to 4.3 stable branch. This change is backward compatible with older version of logger. We should  fix this as soon as possible, because Jekyll is completely unusable on Ruby 3.3 right now. In my test this is the only patch required for Ruby 3.3.

This also picks #9469 and #9196, but these are test-only fixes for the CI. 

cc @parkr @ashmaroli Can you please help take a look and perhaps cut a patch release? Thank you.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,8 +24,8 @@ jobs:
             ruby_version: "3.0"
           - label: Ruby 3.1.2
             ruby_version: "3.1.2"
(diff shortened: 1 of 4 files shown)


mattr- (approved):

mattr-:
Looks fine to me. Will get this merged and start running through the release process in a couple of hours.

PR #9522, opened 2024-01-08
Add csv to runtime dependency list

halorrr:

## Summary

This adds csv as a dependency in the gemspec.

## Context

As of Ruby 3.3.0 I was seeing this warning print to the console when running `jekyll serve`:

`/Users/username/.gem/ruby/3.3.0/gems/jekyll-4.3.3/lib/jekyll.rb:28: warning: csv was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add csv to your Gemfile or gemspec. Also contact author of jekyll-4.3.3 to add csv into its gemspec.`

Since this was a fairly simple fix, I figured I'd just throw a PR up for it. I wasn't sure what version to require as a dependency so I just used the latest major version of csv.

I didn't think there was any tests needed for this type of change but let me know if I'm wrong on that.

Notably if we wanted, this could be a conditional dependency on the ruby version wrapped with something like:
```Ruby
if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('3.3')
  s.add_dependency("csv", "~> 3.0")
end
```

### DIFF
--- a/jekyll.gemspec
+++ b/jekyll.gemspec
@@ -32,6 +32,8 @@ Gem::Specification.new do |s|
   s.required_ruby_version     = ">= 2.5.0"
   s.required_rubygems_version = ">= 2.7.0"
 
+  s.add_dependency("csv",                           "~> 3.0")
+
   s.add_runtime_dependency("addressable",           "~> 2.4")
   s.add_runtime_dependency("colorator",             "~> 1.0")
   s.add_runtime_dependency("em-websocket",          "~> 0.5")


mattr- (approved):
Thanks for fixing this warning!

@jekyllbot: merge +dev

dorianmariecom:
Hey, would be nice if this was released somehow, thanks for maintaining jekyll you rock

mmenanno:
@mattr- Just giving a nudge that it looks like this still hasn't been backported yet

parkr:
@mmenanno I created https://github.com/jekyll/jekyll/pull/9616.

PR #9586, opened 2024-04-23
Fixed: Wrong navigation style on the right side of news and docs pages

note4:
Because if clears spaces, li and class are merged into one, so they are not displayed according to the preset CSS style.


mattr- (approved):
Appreciate you fixing these!

@jekyllbot: merge +docs

PR #9589, opened 2024-04-28
Fix inexistent layout warning for the default 404 page

revolter:
This is a 🐛 bug fix.

## Summary

Fixed this warning message:
```
Layout 'default' requested in 404.html does not exist.
```
that you get by default when you create a new Jekyll site.

## Context

Probably introduced by the Minima v3 upgrade:
https://github.com/jekyll/minima?tab=readme-ov-file#base-layout

### DIFF
--- a/lib/site_template/404.html
+++ b/lib/site_template/404.html
@@ -1,6 +1,6 @@
 ---
 permalink: /404.html
-layout: default
+layout: page
 ---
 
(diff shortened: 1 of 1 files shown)


mattr- (approved):
Thank you for the fix!

@jekyllbot: merge +fix

PR #9591, opened 2024-04-28
Fix incorrect and inconsistent variables in the permalinks page

revolter:

## Summary

One entry in this table and all the entries in the other tables already had the colons prefix.


revolter:
I don't think that the checks failing are caused by the changes in this PR, right?

ashmaroli (approved):

ashmaroli:
Thanks @revolter
@jekyllbot: merge +doc

PR #9606, opened 2024-06-01
Acknowledge `livereload_port` from site config too

### LINKED ISSUE #9605: feat: Allow configuring the LiveReload port as an option in the configuration file
## Summary

When you run `jekyll serve` with LiveReload, you can configure most of the LiveReload behaviour as a configuration file option or a command-line flag (see [config documentation](https://jekyllrb.com/docs/configuration/options/#serve-command-options)).

e.g. you can enable LiveReload with the `livereload: BOOL` config option or the `-l, --livereload` command-line flag.

You can only choose the port for LiveReload to listen on with the `--livereload-port PORT` command-line flag. I'd like to be able to set this as an option in my config file.

## Motivation

For Jekyll in general: completeness with the existing options.

For me specifically: I have a couple of Jekyll sites I work on, and I set different server ports in each of their config files so they don't clash, e.g.:

```yml
port: 5959
livereload: true
```

That way I can switch to any project, run `jekyll serve` and get the “right” behaviour: a new Jekyll server with live reloading, and on a port dedicated to this project.

But every project tries to run LiveReload on 35729, and they clash – I have to add it as a command-line flag to my `jekyll serve` command. I’d like it if I could also specify the LiveReload port in the config file, to ensure they don't clash, e.g.

```yml
port: 5959
livereload: true
livereload_port: 35959
```

## Guide-level explanation

Not necessary for this feature.

I assume this would be similar to other config options.

## Drawbacks

Can't think of any.

## Unresolved Questions

I can't think of any.

gouravkhunger:

## Summary

The `livereload_port` option is already supported internally but is hardcoded to `35729`:

https://github.com/jekyll/jekyll/blob/1484c6d6a41196dcaa25daca9ed1f8c32083ff10/lib/jekyll/commands/serve.rb#L55
https://github.com/jekyll/jekyll/blob/1484c6d6a41196dcaa25daca9ed1f8c32083ff10/lib/jekyll/commands/serve.rb#L74

In order to give preference to user defined `livereload_port` instead of the hardcoded value, the above line needed to be moved from the `init_with_program` function to `process`.

## Context

Fixes #9605

### DIFF
--- a/docs/_data/config_options/serve.yml
+++ b/docs/_data/config_options/serve.yml
@@ -34,6 +34,7 @@
 
 - name: Live reload port
   description: Port for LiveReload to listen on.
+  option: "livereload_port: PORT"
   flag: "--livereload-port PORT"
 
 
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -71,7 +71,6 @@ module Jekyll
             end
 
             cmd.action do |_, opts|
-              opts["livereload_port"] ||= LIVERELOAD_PORT
               opts["serving"] = true
               opts["watch"]   = true unless opts.key?("watch")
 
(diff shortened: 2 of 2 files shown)


ashmaroli (approved):

ashmaroli:
Thanks @gouravkhunger
@jekyllbot: merge +minor

PR #9614, opened 2024-06-05
Update current Ruby version renders and macOS docs

### LINKED ISSUE #9613: [Docs]: Add Ruby version recommendation to the Quickstart page
## Motivation

The [Quickstart page](https://jekyllrb.com/docs/) notes that ruby 2.5.0 or higher is require. I checked my ruby version, which was 2.7.8, so there should have been no problem installing bundler and jekyll gem as it said. However, running `gem install jekyll bundler` prints following error.

```
➜ gem install bundler jekyll
Fetching bundler-2.5.11.gem
ERROR:  Error installing bundler:
	The last version of bundler (>= 0) to support your Ruby & RubyGems was 2.4.22. Try installing it with `gem install bundler -v 2.4.22`
	bundler requires Ruby version >= 3.0.0. The current ruby version is 2.7.8.225.

... install other dependencies

ERROR:  Error installing jekyll:
	The last version of google-protobuf (>= 3.25, < 5.0) to support your Ruby & RubyGems was 4.26.1. Try installing it with `gem install google-protobuf -v 4.26.1` and then running the current command again
	google-protobuf requires Ruby version >= 3.0, < 3.4.dev. The current ruby version is 2.7.8.225.
```

Even this error can be resolved simply install proper version of gems as it suggests, it feels something wrong. Someone who is new to jekyll may feel a little frustrated of this error.

As many macOS installations have Ruby 2.6 installed by default, mac users will encounter this kind of errors while doing quickstart.

## Suggestion

How about adding recommended ruby version besides ruby requirment? Like this.

| Ruby version 2.5.0 or higher (But recommand 3.1.3 or higher)

or add links to [os specific guides](https://jekyllrb.com/docs/installation/#guides)

monfresh:

## Summary

This is a documentation change to update the current version of Ruby to the
latest as of today: 3.3.5

As for the macOS docs, I updated the supported macOS versions to be in line with what Homebrew supports. I also removed `xz` from the  prerequisites that need to be installed with Homebrew since it gets installed automatically when you install ruby-install.

### DIFF
--- a/docs/_data/ruby.yml
+++ b/docs/_data/ruby.yml
@@ -1,3 +1,3 @@
 min_version: 2.5.0
-current_version: 3.1.3
-current_version_output: ruby 3.1.3p185 (2022-11-24 revision 1a6b16756e)
+current_version: 3.3.5
+current_version_output: ruby 3.3.5 (2024-09-03 revision ef084cc8f4)
--- a/docs/_docs/installation/macos.md
+++ b/docs/_docs/installation/macos.md
@@ -5,9 +5,7 @@ permalink: /docs/installation/macos/
 
 ## Supported macOS versions
 
-- Ventura (macOS 13)
-- Monterey (macOS 12)
-- Big Sur (macOS 11)
+We match [Homebrew's macOS requirements](https://docs.brew.sh/Installation#macos-requirements), which typically support the last 2 or 3 macOS versions.
 
 Older macOS versions might work, but we don't officially support them.
 
@@ -48,7 +46,7 @@ Jekyll on your Mac, or if you run into any issues, read that guide.
 Install `chruby` and `ruby-install` with Homebrew:
 
 ```sh
-brew install chruby ruby-install xz
+brew install chruby ruby-install
 ```
 
 Install the latest stable version of Ruby (supported by Jekyll):


ashmaroli:
Hello @monfresh, it would be better if you were to bump the JRuby version in a separate pull request.

olleolleolle:
I have since extracted the JRuby version patch version bump in the CI matrix to its own, smaller PR.

ashmaroli:
@monfresh Minimum version will remain Ruby 2.5.0 for v4.x. We'll bump the requirement when we bump the requirement in the gemspec for Jekyll 5.

docs/_data/ruby.yml
```diff
@@ -1,3 +1,3 @@
-min_version: 2.5.0
-current_version: 3.1.3
-current_version_output: ruby 3.1.3p185 (2022-11-24 revision 1a6b16756e)
+min_version: 3.1.6
```
ashmaroli, line 1: This change is not valid for the time being. Suggesting revert.

```suggestion
min_version: 2.5.0
```

monfresh: Ah, ok. Got it. I'll revert and update the commit message since most of it no longer applies.

ashmaroli (approved):

ashmaroli:
Thanks @monfresh
@jekyllbot: merge +doc

ashmaroli:
@jekyllbot: merge +doc

ashmaroli:
@jekyllbot: merge +doc

PR #9646, opened 2024-08-11
Remove broken wdm dependency of new Jekyll site template

### LINKED ISSUE #9615: Cannot create a new jekyll site because of error 'error occurred while installing wdm (0.1.1),'
New to Ruby and Jekyll but generally tech savvy.
Trying to create a new Jekyll site.
Installed Ruby
Followed all the steps in the docs (some problems accessing Jekyll, resolved by editing Windows path and rebooting)

In VSCODE, ran

    gem install jekyll bundler

which ran with no faults.

Tried to create a new jekyll site with

    jekyll new mysite
In VSCODE also

This reported the error `error occurred while installing wdm (0.1.1), and Bundler cannot continue`
I've looked at various solutions to this - since I am new to Ruby I haven't gone into any of its internals.

Ran 

     bundle install
This reported `Gem::Ext::BuildError: ERROR: Failed to build gem native extension.` with quite extensive diagnostics 

Ran

    gem install wdm
This gave the same diagnostics. At the end it states

    Gem files will remain installed in C:/Users/afree/.local/share/gem/ruby/3.3.0/gems/wdm-0.1.1 for inspection.  
    Results logged to C:/Users/afree/.local/share/gem/ruby/3.3.0/extensions/x64-mingw-ucrt/3.3.0/wdm-0.1.1/gem_make.out
  
 I've gone through most of these steps also outside VScode, in both cmd and powershell
 
I am raising this in the Jekyll forum, but it seems to me that if this is the experience of a newbie following the documentation, it's serious enough to bring to the attention of the maintainers. Apologies (and let me know) if this is the wrong protocol.

AeromXundes:

This is a 🐛 bug fix.

## Summary

Removes the broken `wdm` dependency by removing it from the template.
Fixes #9615. Uses https://github.com/jekyll/jekyll/issues/9615#issuecomment-2183060107 as the fix.

### DIFF
--- a/lib/jekyll/commands/new.rb
+++ b/lib/jekyll/commands/new.rb
@@ -96,9 +96,6 @@ module Jekyll
               gem "tzinfo-data"
             end
 
-            # Performance-booster for watching directories on Windows
-            gem "wdm", "~> 0.1.1", :platforms => [:mingw, :x64_mingw, :mswin]
-
             # Lock `http_parser.rb` gem to `v0.6.x` on JRuby builds since newer versions of the gem
             # do not have a Java counterpart.
             gem "http_parser.rb", "~> 0.6.0", :platforms => [:jruby]


ashmaroli:
@AeromXundes I suggest we simply comment that line out instead of removing it completely.
Let the Window user decide if they are able to install the gem successfully.

PR #9664, opened 2024-09-06
Run GitHub Actions only if needed

ashmaroli:
## Summary

Use the `paths` and `paths-ignore` filters to limit triggering GitHub Action runs to those files that are relevant to given workflow

## Context

[GitHub Actions Docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-including-paths)


ashmaroli:
@jekyllbot: merge +dev

PR #9405, opened 2023-07-10
Update macos.md, remove Catalina (macOS 10.15)

fulldecent:
Removes Catalina (macOS 10.15)

As discussed at #9402

PR #9679, opened 2024-09-15
Update mac version support

fulldecent:

## Summary

This removes an old version of macOS that is no longer supported by Homebrew. [Source](https://docs.brew.sh/Installation#macos-requirements).

And it adds the current version.

The Jekyll project does not support Mac versions that Homebrew no longer supports. [Source](https://github.com/jekyll/jekyll/pull/8993#pullrequestreview-908004059).

And my exact PR that did this before was accepted and merged. [Source](https://github.com/jekyll/jekyll/pull/9405).

### DIFF
--- a/docs/_docs/installation/macos.md
+++ b/docs/_docs/installation/macos.md
@@ -5,9 +5,8 @@ permalink: /docs/installation/macos/
 
 ## Supported macOS versions
 
+- Sonoma (macOS 14)
 - Ventura (macOS 13)
-- Monterey (macOS 12)
-- Big Sur (macOS 11)
 
 Older macOS versions might work, but we don't officially support them.
 


ashmaroli:
Hello @fulldecent,
In a still active pull request by @monfresh at #9614, they proposed to remove only `Big Sur (macOS 11)`. Was the support for macOS 12 dropped very recently (after June 2024)?

ashmaroli:
Additionally, why don't we simply link to the official Hombrew requirements section in our docs instead of duplicating the information in docs here and then making changes every once in a while?

monfresh:
Historically, Homebrew has supported the same macOS versions as Apple, which is typically the last 3 versions. According to [endoflife.date](https://endoflife.date/macos), Monterey is still supported by Apple. It seems like Homebrew has changed their policy.

It's also possible that Homebrew updated this recently for the upcoming release of Sequoia, which would make the last 3 versions Sequoia, Sonoma, and Ventura.

I'm fine with updating my PR to link to Homebrew's requirements and mention in parentheses that it's typically the last 3 macOS versions, although older versions might still work.

ashmaroli:
@monfresh If your pull request will be updated to contain just the documentation changes, I shall merge that and close this.

monfresh:
My PR will also have a change to docs/_data/ruby.yml to update the min and current version

ashmaroli:
Thank you @fulldecent but has been superceded by prior art #9614

PR #9680, opened 2024-09-16
Render theme-gem root only in development

ashmaroli:
- This is a 🐛 bug fix.
- I've adjusted the documentation

## Summary

`{{ theme.root }}` rendering the absolute path to theme-gem has no value outside development builds.
So, render the value only when site is built in development mode explicitly.

**`NOT A BREAKING CHANGE`**

Consider backport to `4.3-stable` branch.


ashmaroli:
@jekyllbot: merge +fix

PR #9688, opened 2024-09-20
Configure some new cops

ashmaroli:
## Summary

Add and configure some new cops available via current bundle (RuboCop v1.57 & extensions)


.rubocop.yml
```diff
@@ -54,6 +56,8 @@ Layout/EmptyLinesAroundAttributeAccessor:
   Enabled: true
 Layout/EndAlignment:
   Severity: error
+Layout/EndOfLine:
+  Enabled: false
```
parkr, line 60: Could we do EnforcedStyle: lf to use line feed everywhere? Maybe in a separate PR?

.rubocop.yml
```diff
@@ -41,6 +41,8 @@ Jekyll/NoPutsAllowed:
 # ----------------------------------------
 Gemspec/DeprecatedAttributeAssignment:
   Enabled: true
+Gemspec/DevelopmentDependencies:
+  Enabled: false
```
parkr, line 45: I wonder if this could be a good cleanup in the future? Not sure the value of add_development_dependency anymore in a gemspec.

parkr (approved):

lib/jekyll/plugin_manager.rb
```diff
@@ -83,7 +83,7 @@ def plugin_allowed?(plugin_name)
     # Returns an array of strings, each string being the name of a gem name
     #   that is allowed to be used.
     def whitelist
-      @whitelist ||= Array[site.config["whitelist"]].flatten
+      @whitelist ||= [site.config["whitelist"]].flatten
```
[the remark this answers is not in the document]
ashmaroli: I had similar thoughts too but didn't want to play around with this particular line in this PR (only do cosmetic changes in `lib/**/*` in this PR; Handle specific optimizations in future PRs).

Your points are spot-on regarding allocations. Specifically, current implementation allocates 2 arrays at minimum:
- The outer `[]` allocates a new array irrespective of `site.config["whitelist"]` already being an array.
- The `Array#flatten` duplicates existing array, allocating another array just to handle `site.config["whitelist"]` already being an array.
- Now, there's a sticky situation here. `Array(whitelist_config)` alone won't ensure a flat list of strings. So, we have to call `Array#flatten` (or `Array#flatten! with extra steps).

All of the decision overhead in the third-point is outside the scope of this PR.

.rubocop.yml
```diff
@@ -234,32 +252,70 @@ Minitest/AssertInDelta:
...
 Minitest/RefuteInDelta:
   Enabled: true
 Minitest/RefuteKindOf:
   Enabled: true
+Minitest/RefuteOperator:
+  Enabled: false
 Minitest/RefutePathExists:
   Enabled: true
+Minitest/RefutePredicate:
+  Enabled: false
```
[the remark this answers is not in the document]
ashmaroli: Current:
```ruby
refute @collection.write?
```
After style change:
```ruby
refute_predicate @collection, :write?
```
Doesn't the *current* seem clearer and easier to comprehend faster?
*Maybe this is just a subjective preference.* I'll go with whatever you decide (and apply same for the `assert_predicate` counterpart.)

.rubocop.yml
```diff
@@ -234,32 +252,70 @@ Minitest/AssertInDelta:
...
+Minitest/NonPublicTestMethod:
+  Enabled: true
+Minitest/RedundantMessageArgument:
+  Enabled: true
 Minitest/RefuteInDelta:
   Enabled: true
 Minitest/RefuteKindOf:
   Enabled: true
+Minitest/RefuteOperator:
+  Enabled: false
```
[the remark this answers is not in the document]
ashmaroli: Current:
```ruby
assert before_time <= @site.time
```
After style change (with `AssertOperator` counterpart):
```ruby
assert_operator before_time, :<=, @site.time
```

The new style is objectively harder for humans to parse. So, `disabled` without veto. (applied to counterpart `RefuteOperator` as well).

.rubocop.yml
ashmaroli: We could use `EnforcedStyle: lf` in the current PR as well without affecting CI or including more file changes. All files are already checked out by git with native line-endings automatically.

But for contributors-from-Windows' sake, we may perhaps need to reconfigure `.gitattributes` file.

.rubocop.yml
ashmaroli: I was always in favor of moving dev-deps to Gemfile until I saw your [`commit @ rubocop-jekyll`](https://github.com/jekyll/rubocop-jekyll/pull/27/commits/34338391b2aa35d2c09a5b71d8b77bb3115d3196)

ashmaroli:
Hello @parkr. When you get some time off, please let me know if my replies to your queries are satisfactory.

.rubocop.yml
parkr: I see -- I think it was more that I was too lazy to move them all to the Gemfile. I think managing dev dependencies in the Gemfile in 2024 makes sense, but no need to enforce it now if it's going to be a good amount of work (especially across repos).

.rubocop.yml
parkr: Ok, I think we'd prefer to check in LF instead of CRLF. Can .gitattributes give us a per-repo way of saying "Windows needs CRLF so write them to disk but Git should remove the CR"? As a Windows user (or one in the past), I think you have the best understanding of how this should work.

.rubocop.yml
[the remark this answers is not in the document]
parkr: I see. I think the idea is to have better error messages since `assert boolean_expr` makes test output pretty useless. Perhaps we should enforce that `assert` calls must have a message component, i.e. `assert boolean_expr, message_explaining_expectation`. If it's not already a cop, then no worries.

.rubocop.yml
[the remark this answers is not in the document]
parkr: Replied above, I think `assert/refute predicate, message` would be the best of both worlds, but no worries if there's not a cop for enforcing a message in assert/refute calls.

parkr (approved):

ashmaroli:
Thank you for the discussion, @parkr
@jekyllbot: merge +dev

PR #9690, opened 2024-09-23
Render `data-lang` attr in kramdown code-blocks

ashmaroli:
- I've added tests.

## Summary

Attach `lang` value of highlighted kramdown code-blocks to a `data-lang` HTML attribute.

## Background

Code-blocks rendered using triple backticks (via kramdown) currently have a `language-<LANG>` class attribute on the container element `<div>`. This requires users to depend on JavaScript code to extract just the `<LANG>` value to render language-labels over their code-blocks.

Attaching the `<LANG>` value to a `data-lang` attribute will allow users to render language-labels using plain CSS
`content: attr(data-lang)` in their stylesheet.

## Notes

- `data-lang` attribute is only defined for highlighted code-blocks.
- `data-lang` attribute is not defined for `plaintext` code-blocks.

### DIFF
--- a/lib/jekyll/converters/markdown/kramdown_parser.rb
+++ b/lib/jekyll/converters/markdown/kramdown_parser.rb
@@ -1,6 +1,17 @@
 # frozen_string_literal: true
 
 module Kramdown
+  require "kramdown/converter/html"
+
+  class JekyllKD2HTML < Converter::Html
+    def add_syntax_highlighter_to_class_attr(attr, lang = nil)
+      super
+      return attr if lang.nil? || lang == "plaintext"
+
+      attr["data-lang"] = lang
+    end
+  end
+
   # A Kramdown::Document subclass meant to optimize memory usage from initializing
   # a kramdown document for parsing.
   #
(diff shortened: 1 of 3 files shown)


lib/jekyll/converters/markdown/kramdown_parser.rb
```diff
@@ -1,6 +1,17 @@
 # frozen_string_literal: true
 
 module Kramdown
+  require "kramdown/converter/html"
+
+  class JekyllKD2HTML < Converter::Html
```
mattr-, line 6: What do you think about renaming this to `JekyllHtmlConverter` to keep with a similar naming convention as the other class in this file?

mattr- (commented):
This is looking good to me. I left the one question for you to consider before approval and merge.

parkr (approved):
LGTM, +1 to the question from Matt.

lib/jekyll/converters/markdown/kramdown_parser.rb
ashmaroli: While I did have that exact name (actually it was `JekyllHTMLConverter`) for the class when starting with the patch, the segment `HtmlConverter` felt ambiguous to me. Would someone interpret it as a class for *converting HTML* (like how we have the `sass-converter` plugin).

Regardless, I am open to renaming the class.

parkr: +1 on keeping it clear. My only nit here is that "KD2" is not super clear. We can think of the entire class name with the `Kramdown::` prefix for clarity if you think that would allow the inner class to be less wordy, but I don't mind long class names.

ashmaroli:
I think its better to have this shipped in a separate project instead of Jekyll core.

PR #9697, opened 2024-10-11
Fix missing glob matching pattern support description in the documentation of the `include` and `exclude` configuration options

### LINKED ISSUE #9696: [Docs]: Missing `glob` matching pattern support description in the documentation of the `include` and `exclude` configuration options
## Motivation

Currently in the [Configuration Options | Jekyll • Simple, blog-aware, static sites](https://jekyllrb.com/docs/configuration/options/) documentation there's no mention of the support of [File::Fnmatch](https://ruby-doc.org/core-2.5.5/File.html#method-c-fnmatch-3F) filename glob matching.

One has to dig into [this issue](https://github.com/jekyll/jekyll/issues/948) and [this pull request](https://github.com/jekyll/jekyll/pull/743) to learn the implementation details, we should add proper documentation to avoid these hoops.

## Suggestion

Edit [jekyll/docs/_data/config_options/global.yml at master · jekyll/jekyll](https://github.com/jekyll/jekyll/blob/master/docs/_data/config_options/global.yml) and add the description to the related options.

It should also mention that to exclude a filename(pattern) in all depths the exclude config option should be set in the following fashion:

```yaml
exclude:
  - README.md
  - '**/README.md'
```

brlin-tw:

## Summary

This pull request adds a missing glob pattern support description in the documentation of the `include` and `exclude` site configuration options.

It appears that Regexp is also supported, but I am unable to use it at my end so it will not be addressed in this PR.

## Context

Fixes #9696

### DIFF
--- a/docs/_data/config_options/global.yml
+++ b/docs/_data/config_options/global.yml
@@ -41,15 +41,15 @@
     Exclude directories and/or files from the conversion. These exclusions are relative to the site's source directory
     and cannot be outside the source directory.
     <br />
-    This configuration option also supports Ruby's <a href="https://ruby-doc.org/3.3.5/File.html#method-c-fnmatch"><code>File.fnmatch</code> filename globbing patterns</a> to match multiple entries to exclude.  For example, you can exclude all README.md files in your source tree from being included in your site by specifying the following `exclude` option entries:
-    <ul>
-      <li><code>README.md</code></li>
-      <li><code>**/README.md</code></li>
-    </ul>
-    In Jekyll 3, the `exclude` configuration option replaces the default exclusion list.
+    This configuration option supports Ruby's <a href="https://ruby-doc.org/3.3.5/File.html#method-c-fnmatch">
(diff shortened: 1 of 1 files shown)


brlin-tw:
The following tests failed on my end, but I failed to see how these tests are related to my change:

* TestDocument#test_: a document with a date with timezone should have the expected date.
* TestDocument#test_: a document with a date with timezone should return the expected date via Liquid.
* TestFilters#test_: filters jsonify filter should convert drop to json.

brlin-tw:
For some reason, the unordered list markup broke the .description paragraph:

Screenshot

The corresponding Markdown template seems to be the following file:

<https://github.com/jekyll/jekyll/blob/master/docs/_docs/configuration/options.md>

but I unable to figure out how to fix it so will leave it here.

ashmaroli (approved):

ashmaroli:
Thanks @brlin-tw
@jekyllbot: merge +doc

PR #9711, opened 2024-10-27
Fix layout data overshadowing by subsequent layouts

### LINKED ISSUE #9710: [Bug]: Variables from one layout overwrite the ones in the current layout
### Operating System

Windows 10

### Ruby Version

ruby 3.3.5 (2024-09-03 revision ef084cc8f4) [x64-mingw-ucrt]

### Jekyll Version

jekyll 3.10.0

### GitHub Pages Version

github-pages 232

### Expected Behavior

Having multiple layouts with a frontmatter variable set to different values (for example, `layouts/default2024.html` has `year: 2024` and `layouts/default2025.html` has `year: 2025`), rendering the body of a page that uses a specific layout by selecting it in the frontmatter, for example with `layout: default2024`, should use the appropriate value of the variable.

### Current Behavior

The value of a different layout is used instead of the expected one.

I don't know what determines which of the layouts is used, in the original project it shows 2024 when I expected 2025, but in the example minimal reproduction project I created it shows 2025 when I expected 2024.

### Relevant log output

_No response_

### Code Sample

I have created an example project:
<https://github.com/ror3d/jekyll-layout-bug>

Clone, run `bundle install`, and run `bundle exec jekyll serve`. Visit the server, and notice that in one of the two pages, the year will be wrong.

ashmaroli:
- This is a 🐛 bug fix.
- I've added tests.

## Summary

Stop merging previous layout data into current layout data.

## Context

Fixes #9710

### DIFF
--- a/features/rendering.feature
+++ b/features/rendering.feature
@@ -239,3 +239,29 @@ Feature: Rendering
     And I should not see "content\": \"__Hello World__" in "_site/test.json"
     But I should see "content\": \"<p>about.md</p>" in "_site/test.json"
     And I should see "content\": \"<p><strong>Hello World</strong></p>" in "_site/test.json"
+
+  Scenario: Render layout front matter data
+    Given I have an "alpha.md" page with layout "orchard" that contains "item on sale: {{ layout.item }}"
+    And I have an "beta.md" page with layout "bakery" that contains "item on sale: {{ layout.item }}"
+    And I have a _layouts directory
+    And I have an "_layouts/bakery.html" file with content:
+      """
+      ---
+      item: Carrot Cake
+      ---
+
+      {{ content }}
+      """
+    And I have an "_layouts/orchard.html" file with content:
+      """
+      ---
+      item: Granny Smith Apples
+      ---
+
+      {{ content }}
+      """
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "item on sale: Granny Smith Apples" in "_site/alpha.html"
+    And I should see "item on sale: Carrot Cake" in "_site/beta.html"


ashmaroli:
ping @parkr

parkr:
Hey, I think this is a breaking change since it was a feature that was actually added at one point IIRC. So this would reverse that decision and remove the feature.

ashmaroli:
Hello @parkr, since I am not able to see how exactly this is a breaking-change, (git blame doesn't reveal a concrete usage in the past either), could you please illustrate with an example scenario.

ror3d:
How is this a feature?

ashmaroli:
@ror3d Please look again. There's no mention of this being a **feature** anywhere.

ror3d:
@ashmaroli yes, I meant this as a question to @parkr , how is it a feature that the values from a layout that's not currently in use by a page are overwriting the ones for the layout that is explicitly chosen for that page?

ror3d:
> Hey, I think this is a breaking change since it was a feature that was actually added at one point IIRC. So this would reverse that decision and remove the feature.

@parkr could you clarify how the bug reported in #9710 could be a feature?

parkr:
Hey, so my context on this is that this allows layouts which roll up to all provide some data. Maybe you have a "product" layout which specifies certain relative links or breadcrumbs, and that rolls up to a "default" layout which contains author information etc. If you don't merge the layout data, you don't get all of that metadata available in one place. Sometimes you'd want to access the entire collection of layout metadata all in one place.

This is a "power-user" feature for sure, but it's been there for ~years so removing it could be disruptive. It would be a v5.0 change to remove this since it would be breaking.

One alternative (also breaking technically) is to change the order of the merge.

ror3d:
Right, but there is still a bug in that system, as described in the issue I wrote up #9710 . The values from a layout that is *not* referenced shadow the ones of the layout directly referenced in the document.

That can't be a feature.

Maybe you mean that the PR, by removing the call to `Utils.deep_merge_hashes`, is breaking more complex use cases, but as it stands right now, the call to it seems to not give priority to what it should, or it seems to be merging stuff that it should not have access to.

ashmaroli:
@ror3d Parker's latest comment threw some light on *why* the previous change could be considered a breaking-change, at the same time showing the path for a probable resolution.

I have reinstated merging ancestor-layout-data but with a small change of flipping the order of merged objects. This change should in theory fix your reported issue as well. (Do let me know if it doesn't).
I will adapt Parker's example scenario into a Cucumber scenario in the coming days to test probable disruption.

parkr:
Yeah I think the thing I was thinking of was inheriting (by design) from a parent layout in the tree. But you pointed out that actually siblings (non-ancestors) in the tree were "polluting" the values. That is not right.

PR #9715, opened 2024-11-12
Link to GitHub Pages Dependency versions page instead of explicitly specifying Jekyll version

CookiePLMonster:
## Summary

GitHub Pages gem updated their [shipped Jekyll version](https://pages.github.com/versions/) to 3.10.0. This small PR updates the docs to reflect that.

## Context

Keeps Jekyll docs up to date with the Pages gem.

### DIFF
--- a/docs/_docs/continuous-integration/github-actions.md
+++ b/docs/_docs/continuous-integration/github-actions.md
@@ -18,9 +18,9 @@ environment and gemset.
 
 ### Control over gemset
 
-- **Jekyll version** --- Instead of using the classic GitHub Pages-provided version at `3.9.3`, you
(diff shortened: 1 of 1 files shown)


ashmaroli:
Perhaps a better alternative would be to avoid mentioning a version at all and link to the [GHP Dependency Version page](https://pages.github.com/versions/) instead.

CookiePLMonster:
> Perhaps a better alternative would be to avoid mentioning a version at all and link to the [GHP Dependency Version page](https://pages.github.com/versions/) instead.

That is a good idea! I will update the wording to reflect that.

ashmaroli (approved):

ashmaroli:
Thanks @CookiePLMonster
@jekyllbot: merge +doc

PR #9737, opened 2024-12-27
Fix expected markup per WAI-ARIA requirements

ashmaroli:
## Summary

`role="doc-noteref"` should be an attribute of the `<a/>` tag instead of `<sup/>` in `features/post_excerpts.feature:92`.


ashmaroli:
JRuby error is unrelated. Ignoring..
@jekyllbot: merge +dev

PR #9742, opened 2024-12-30
Add an optional `:rdoc` group of gems

ashmaroli:
## Summary

Move development-dependency `rdoc` and its dependency `psych` into an optional Bundler group named `:rdoc` so that we do not have to deal with issues due to gem `psych` by default.


ashmaroli:
@jekyllbot: merge +dev

PR #9753, opened 2025-01-16
Remove unnecessary runtime dependency

ashmaroli:
## Summary

- gem `mutex_m` is only required within gem `httpclient` which is a development dependency. Therefore, it shouldn't be listed as jekyll's runtime dependency. (*The necessary change has been made upstream via https://github.com/nahi/httpclient/commit/6d804c17cf8f8a58ad6605c15f5e1544ee4fc78f but not released yet.*)
- Add a entry in our HISTORY doc to mention addition of gem `base64` as runtime dependency since that change is actually a minor enhancement instead of an internal fix .

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,11 @@ group :test do
   gem "jekyll_test_plugin"
   gem "jekyll_test_plugin_malicious"
   gem "memory_profiler"
+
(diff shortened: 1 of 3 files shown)


ashmaroli:
@jekyllbot: merge +dev

PR #9758, opened 2025-01-24
Remove unnecessary constraint on gem `mercenary`

ashmaroli:
- This is an 🙋 enhancement.

## Summary

Allow using any version of gem `mercenary` greater than or equals to `v0.3.6` but lesser than `v1.0.0`


ashmaroli:
@jekyllbot: merge +minor

PR #9776, opened 2025-02-15
feat: Allowing post_url tag to receive liquid variables

### LINKED ISSUE #9775: feat: post_url accept liquid variable as input
## Summary

As `{% link {{ somevar }} %}` tag that accepts a Liquid variable as value, allow passing Liquid variables (and thus expansions) to `{% post_url %}`

## Motivation

`post_url` tag works only for constants, treating any argument as plain constant. I can benefit adding metadata pointing to my own posts. For instance, I always create an article about some change in my blog.

So, if I can list for some behavior a list of my own blog posts and show it to me, that would be of great value.

## Guide-level explanation

Let us have a list of blog posts from our own to show, a list of slugs/titles named `my_posts`, with the content:

```yaml
my_posts:
- title: "Post facilmente citável"
  slug: 2024-05-22-citable
- title: "Alpine morreu! Vida longa ao Alpine!"
  slug: 2025-02-03-long-live-alpine
- title: "Fluent Builder com região crítica"
  slug: 2021-12-23-fluent-builder
```

We can iterate through this list and create links:

```liquid
Check these posts for further explanation:

{% for posts in my_posts %}
- [{{ post.title }}]({% post_url {{ post.slug }} %})
{% endfor %}
```

It will deal with the input for `post_url` tag as a Liquid string, just as in the `link` tag.

## Drawbacks

The validation will be at `render` level instead of at initialization level.

It may cause some overhead to run `post_url` as it deals with Liquid rendering to get the post slug.

Users that relies having `{{` in their post urls and link them back using `{% post_url %}` may experience some backlash with this change.

## Unresolved Questions

What will be such overhead?

jeffque:
Adds capability to post_url tag to deal with Liquid variable, solving the Liquid expansion as first step before checking input format.

This is a 🙋 feature or enhancement. Closes #9775 

I've done local tests rendering in my blog, adding the tag with the same code in my plugins.

Some tests that I ran locally:

1st test, running as usual:

```liquid
<!-- sanity check, should work as usual -->
{% post_url 2024-05-22-citable %}
```

Results: rendered as expected

```txt
/blog/2024/05/22/citable
```

2nd test, a non-existing post

```liquid
<!-- sanity check, should fail as usual -->
{% post_url 2024-05-22-pumpkin %}
```

Results: failed with the following message

```none
  Liquid Exception: Could not find post "2024-05-22-pumpkin" in tag 'post_url'. Make sure the post exists and the name is correct. in managerial.md
             Error: Could not find post "2024-05-22-pumpkin" in tag 'post_url'. Make sure the post exists and the name is correct. 
             Error: Run jekyll build --trace for more information.
```

3rd test, a bad format date

```liquid
<!-- sanity check, should fail as usual -->
{% post_url 2024-05-citable %}
```

Results: failed with the following message

```none
  Liquid Exception: Could not parse name of post "2024-05-citable" in tag 'post_url'. Make sure the post exists and the name is correct. Jekyll::Errors::InvalidPostNameError: '2024-05-citable' does not contain valid date and/or title. in managerial.md
             Error: Could not parse name of post "2024-05-citable" in tag 'post_url'. Make sure the post exists and the name is correct. Jekyll::Errors::InvalidPostNameError: '2024-05-citable' does not contain valid date and/or title. 
             Error: Run jekyll build --trace for more information.
```

4th test, passing a liquid variable to tag

```liquid
<!-- feature test, should expand Liquid variable and work as usual -->
{% assign value="2024-05-22-citable" %}
{% post_url {{value}} %}
```

Results: rendered as expected

```txt
/blog/2024/05/22/citable
```

5th test, passing a liquid variable to tag with a non-existing post

```liquid
<!-- feature test, should expand Liquid variable and fail as usual non-existing post -->
{% assign value="2024-05-22-pumpkin" %}
{% post_url {{value}} %}
```

Results: failed with the following message

```none
  Liquid Exception: Could not find post "2024-05-22-pumpkin" (from input "{{value}}") in tag 'post_url'. Make sure the post exists and the name is correct. in managerial.md
             Error: Could not find post "2024-05-22-pumpkin" (from input "{{value}}") in tag 'post_url'. Make sure the post exists and the name is correct. 
             Error: Run jekyll build --trace for more information.
```

6th test, passing a liquid variable to tag with a bad format date

```liquid
<!-- feature test, should expand Liquid variable and fail as usual -->
{% assign value="2024-05-citable" %}
{% post_url {{value}} %}
```

Results: failed with the following message

```none
  Liquid Exception: Could not parse name of post "2024-05-citable" (from input "{{value}}") in tag 'post_url'. Make sure the post exists and the name is correct. Jekyll::Errors::InvalidPostNameError: '2024-05-citable' does not contain valid date and/or title. in managerial.md
             Error: Could not parse name of post "2024-05-citable" (from input "{{value}}") in tag 'post_url'. Make sure the post exists and the name is correct. Jekyll::Errors::InvalidPostNameError: '2024-05-citable' does not contain valid date and/or title. 
             Error: Run jekyll build --trace for more information.
```

## Summary

Expand the input for `post_url` tag as Liquid. Copied the same behavior as seen in [`lib/jekyll/tags/link.rb`](https://github.com/jekyll/jekyll/blob/1f319fb273b6cdf876bc6edd38d7477935cdda8c/lib/jekyll/tags/link.rb#L23)

## Context

Resolves #9775


jeffque:
There were some failures regarding timezone that I believe there was an issue setting my local environment:

```text
Failure:
TestDocument#test_: a document with a date with time but without timezone should have the expected date.  [<path/to/>jekyll/test/test_document.rb:641]
Minitest::Assertion: Expected: "2015/10/01"
  Actual: "2015/09/30"

Failure:
TestDocument#test_: a document with a date with time but without timezone should return the expected date via Liquid.  [<path/to/>jekyll/test/test_document.rb:645]
Minitest::Assertion: Expected: "2015/10/01"
  Actual: "2015/09/30"
```

Also, there was some issue with JSON ordering, but the fields were kept with the correct values, just the ordering that was messed up. I don't believe it is related to any changes that I had made:

```text
Failure:
TestFilters#test_: filters jsonify filter should convert drop to json.  [<path/to/>jekyll/test/test_filters.rb:707]
+++ actual
@@ -1,25 +1,25 @@
 {
+  "output": null,
   "name": "2008-02-02-published.markdown",
   "path": "_posts/2008-02-02-published.markdown",
-  "previous": null,
-  "output": null,
-  "content": "This should be published.
-",
-  "id": "/publish_test/2008/02/02/published",
-  "url": "/publish_test/2008/02/02/published.html",
-  "relative_path": "_posts/2008-02-02-published.markdown",
   "collection": "posts",
   "excerpt": "<p>This should be published.</p>
 ",
-  "draft": false,
   "categories": [
     "publish_test"
   ],
-  "layout": "default",
+  "relative_path": "_posts/2008-02-02-published.markdown",
+  "url": "/publish_test/2008/02/02/published.html",
+  "previous": null,
   "title": "Publish",
+  "tags": [],
+  "content": "This should be published.
+",
+  "id": "/publish_test/2008/02/02/published",
+  "date": "2008-02-02 00:00:00 -0300",
+  "draft": false,
+  "layout": "default",
   "category": "publish_test",
-  "date": "2008-02-02 00:00:00 +0000",
   "slug": "published",
-  "ext": ".markdown",
-  "tags": []
+  "ext": ".markdown"
 }
```

Comparing the obtained output and the expected value:

```
-  "output": null,
+  "output": null,

-  "id": "/publish_test/2008/02/02/published",
+  "id": "/publish_test/2008/02/02/published",

-  "relative_path": "_posts/2008-02-02-published.markdown",
+  "relative_path": "_posts/2008-02-02-published.markdown",

-  "url": "/publish_test/2008/02/02/published.html",
+  "url": "/publish_test/2008/02/02/published.html",

-  "content": "This should be published.
-",
+  "content": "This should be published.
+",

-  "draft": false,
+  "draft": false,

-  "layout": "default",
+  "layout": "default",

-  "date": "2008-02-02 00:00:00 +0000",
+  "date": "2008-02-02 00:00:00 -0300",

-  "previous": null,
+  "previous": null,

-  "tags": []
+  "tags": [],

-  "ext": ".markdown",
+  "ext": ".markdown"
```

The mismatches were mostly due to positioning: like with the `output` field, in the generated JSON was the first field but the expected JSON was the third one. Also, the trailing comma expected in `ext` was due to it being the one before the last field, and the trailing comma not expected the field `tags` due to it being the last field.

Also, the `date` has something to do with my timezone (see related error with timezone previously reported in this comment).

jeffque:
@ashmaroli , can you review this?

ashmaroli:
Hello @jeffque,
Thank you for submitting a pull request to enhance our `post_url` Liquid tag.

I have a couple of misgivings regarding this:
- Parsing and rendering a Liquid construct is expensive (performance-wise and object-allocations-wise). Ergo this *enhancement* will result in *regression* of existing use-cases in the wild i.e. the new implementation will now parse and render given `tag_markup` regardless of whether the markup is a variable.
- The new variable-names and private-method-names does not seem very intuitive and hampers code-readability. (Not a major blocker, especially if you are not a native English-speaker. We can deal with this during final rounds of reviews).

Since the test-suite passes, you should now consider optimizing the implementation especially reduce regressions as much possible.

---

P.S. Personally, I'm not a fan of this syntax (mixing `{{ }}` within `{% %}`), but unfortunately this is the syntax followed by Jekyll's
`{% include %}` and `{% link %}` tags. So, lets roll as-is.
Additionally, there had been talks of deprecating this tag in favor of the `link` tag, among the maintainers in the past. Not sure if that idea has been dropped entirely.

jeffque:
@ashmaroli , so the performance issue already affect `{% link %}`.  I could try as a heuristic detect the presence of `{{` expansions, if absent then does not invoke liquid rendering.

Do you think it makes sense to improve both `{% link %}` and `{% post_url %}` regarding the liquid expansion in this PR? Or should i just target post_url for now and, when proven a okayish performance (with some benchmark sample?) than apply it back to `{% link %}`?

About the names... yes, I am not a native speaker. But even in my mother-tongue I find it difficult to name stuff.

I had no idea that there was some talk about deprecating `post_url`.

ashmaroli:
> the performance issue already affect {% link %}.

To a certain extent, yes.

> I could try as a heuristic detect the presence of `{{` expansions, if absent then does not invoke liquid rendering.

Yes. That would work as a basic start. Other thing(s) you can consider based on your comfort-level with Ruby:
- If there are multiple identical constructs (e.g. multiple `post_url {{ my_var }}`), wouldn't _parsing_ the various constructs repeatedly be wasteful?

Do you think it makes sense to improve both `{% link %}` and `{% post_url %}` regarding the liquid expansion in this PR? Or should i just target `post_url` for now and, when proven a okayish performance (with some benchmark sample?) than apply it back to `{% link %}`?

No. This PR should only be about `{% post_url %}`.

> About the names... yes, I am not a native speaker. But even in my mother-tongue I find it difficult to name stuff.

No issues. Defer for later.

> I had no idea that there was some talk about deprecating `post_url`.

We have not reached a concrete stand regarding this.

Pinging @parkr and @mattr- for comment on confirming this status before the author invests more time in this pull request.

jeffque:
I tried a benchmarked based on https://github.com/jekyll/jekyll/blob/master/benchmark/capture-assign.rb : https://gist.github.com/jeffque/144cca0483411339fdec801ae45fb14e

I am not used to writing benchmarks. If deemed good, I'll add it to the codebase.

I did not factored in time from [`PostComparer`](https://github.com/jekyll/jekyll/blob/f6bab479449c5450333f65c59a784b514a4be53c/lib/jekyll/tags/post_url.rb#L5) to check the significance of the liquid rendering in the whole operation, just a simple variable expansion.

> Yes. That would work as a basic start. Other thing(s) you can consider based on your comfort-level with Ruby:
> 
> - If there are multiple identical constructs (e.g. multiple `post_url {{ my_var }}`), wouldn't parsing the various constructs repeatedly be wasteful?

Yes, it would be wasteful. Also, I am a newbie in ruby, I just do a little more than basic stuff and I do not write idiomatic ruby (my code is tainted with a cursed functional-yish java code with immutable objects).

I am not aware when changes to `my_var` could change the expanded value to apply a useful flyweight here, nor how to properly use Jekyll `context.registers` to avoid not expanding it again in case it should be expanded with the new variable values.

Also, as the input can be hypothetically be filtered and use many variables instead of just one, I just thought that it would be a big effort for, who knows, a little gain?

ashmaroli:
Since it is known that Liquid parsing and rendering will have a certain impact on performance, I don't think there is a value in attaching a benchmark script to this pull request. Having it at your end is beneficial during feature development, though. I suggest removing the attached benchmark (note: We squash commits during merge. So, there's no need to rebase this branch.)

---

Moving on, consider the following code in an include_file:
```liquid
{% for entry in site.data.archives %}
  <a href="{% post_url {{ entry.slug }}.md %}" title="{{ entry.title }}">
    {{ entry.title }}
  </a>
{% endfor %}
```
Lets assume this include_file is used in two different layouts to render a sidebar.

During the *build process*, Liquid will *parse* the above code just *once* regardless of how many *pages* utilize the two layouts linking the include_file. This means that, in this particular scenario, there will only be a single instance of `Jekyll::Tags::PostUrl` class but, **this one instance** will call upon **it's `render()` method** as many times as *the number of pages dependent* on the above include_file via layouts.

*(Just an FYI since your current benchmark does not take this scenario into consideration.)*

jeffque:
@ashmaroli , genial!

I was not aware of that! Thanks for the explanation!

I'll add this shortly, once the object has been rendered the first time, then it should not reprocess it's value.

jeffque:
@ashmaroli , I decided for testing purpose add all my blog posts in all pages. I've listed all the slugs in a data file and in the `default` layout added this:

```liquid
    <ul>
    {% for slug in site.data.brute %}
        <li><a href="{% post_url {{ slug }} %}">{{ slug }}</a></li>
    {% endfor %}
    </ul>
```

In my `site.data.brute` datafile there was listed 102

As you've said, there were lots of reuse of those `post_url` objects created by the `{% post_url %}` (I actually printed when it was rendering, and printing again when was first time). It took 2.701881 seconds to generate my blog.

Just for control purposes, I removed this excerpt and it took 2.606908 seconds.

Just for curiosity, I'll do the same test without the rendering reuse optimization just to check the performance.

jeffque:
Without optimizations: 4.310295 seconds

Brutal!

I'll add an issue to keep track of the optimization in `{% link %}` tag just to not getting it lost.

ashmaroli:
Good to know that, @jeffque.
My example scenario was to give you an idea about probable scope for optimization and fact that Liquid tags are _context-sensitive_; i.e. the rendered output changes (and should change) with respect to given `context`.
In general, the total build times are very flaky due to various factors out of our control, so take care not to overexert yourselves leading to _micro-optimizations_ (won't matter in the overall build process) and _premature optimizations_.

jeffque:
Yes, I messed it up optimizing wrong, trying to fix it.

It least I got to know more about how Jekyll render stuff

lib/jekyll/tags/post_url.rb
```diff
@@ -61,44 +61,86 @@ class PostUrl < Liquid::Tag
...
+      end
+
+      private
+
+      def use_template?
+        content = @orig_post
+        return false if content.nil? || content.empty?
+
+        content.include?("{%") || content.include?("{{")
+      end
```
ashmaroli, line 104: Variable `@orig_post` will always be a string.
So testing for `nil` is redundant.
I'm not on board with allowing `{%` in the markup (because it denotes a Liquid tag which in turn need not always return a string on expansion).
Lastly, if the markup contains `{{` it is definitely not empty. So lets do away with that redundancy.

```ruby
def use_template?
  @orig_post.include?("{{")
end
```
Finally now since this has been distilled to a single line, the method definition itself is unnecessary and can be inlined at point-of-call.

lib/jekyll/tags/post_url.rb
```diff
@@ -61,44 +61,86 @@ class PostUrl < Liquid::Tag
...
+
+      def expand_orig_post
+        return @orig_post if @template.nil?
+
+        @template.render(@context)
+      end
```
ashmaroli, line 110: Ruby allows overstepping `nil` with an operator.
This logic can be replaced inline at call-location with following line:
```ruby
# Bonus: return @orig_post if @template renders into nil or false

@template&.render(context) || @orig_post
```

lib/jekyll/tags/post_url.rb
```diff
@@ -61,44 +61,86 @@ class PostUrl < Liquid::Tag
...
+
+      # if there is liquid rendering besides a simple constant, adds
+      # "(from input \"#{@orig_post}\")" to let maintainer know where and why it failed
+      def determine_post_string(orig_post_expanded)
+        if orig_post_expanded == @orig_post
+          "\"#{orig_post_expanded}\""
+        else
+          "\"#{orig_post_expanded}\" (from input \"#{@orig_post}\")"
+        end
+      end
```
ashmaroli, line 120: Ruby focuses on readability (and aesthetics). The following is equivalent:

```suggestion
      # if there is liquid rendering besides a simple constant, adds
      # `(from input "#{@orig_post}")` giving the user some context.
      def determine_post_string(orig_post_expanded)
        if orig_post_expanded == @orig_post
          orig_post_expanded.inspect
        else
          "#{orig_post_expanded.inspect} (from input #{@orig_post.inspect})"
        end
      end
```

features/post_url_tag.feature
```diff
@@ -14,6 +14,29 @@ Feature: PostUrl Tag
...
+    And the _site directory should exist
+    And I should see "<p><a href=\"/2019/02/04/hello-world.html\">Welcome</a></p>" in "_site/index.html"
+
+    Scenario: A site that is using the defaults for permalink and mentioning two posts via a liquid variable in a for tag
+    Given I have a _posts directory
+    And I have the following post:
+      | title       | date       | content           |
+      | Hello World | 2019-02-04 | Lorem ipsum dolor |
+      | We Meet Again | 2019-02-05 | Alpha beta gamma  |
+    And I have an "index.md" page that contains "{% assign posts = '2019-02-04-hello-world;2019-02-05-we-meet-again' | split: ';' %}{%- for slug in posts -%}[{{slug}}]({% post_url {{slug}} %}){%- endfor -%}"
```
ashmaroli, line 34: We have an alternative definition for long content such as this:

```gherkin
And I have an "index.md" page with content:
  """
  {% assign posts = ... %}
  {% for slug in ... %}
    ...
  {% endfor %}
  """
```

features/post_url_tag.feature
```diff
@@ -14,6 +14,29 @@ Feature: PostUrl Tag
...
+    And I have the following post:
+      | title       | date       | content           |
+      | Hello World | 2019-02-04 | Lorem ipsum dolor |
+    And I have an "index.md" page that contains "{% assign value='2019-02-04-hello-world' %}[Welcome]({% post_url {{value}} %})"
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "<p><a href=\"/2019/02/04/hello-world.html\">Welcome</a></p>" in "_site/index.html"
+
+    Scenario: A site that is using the defaults for permalink and mentioning two posts via a liquid variable in a for tag
```
ashmaroli, line 28: Extra whitespace here:
```suggestion
  Scenario: A site that is using the defaults for permalink and mentioning two posts via a liquid variable in a for tag
```

ashmaroli (commented):
Need some changes to be made

lib/jekyll/tags/post_url.rb
jeffque: TIL, `&.`

lib/jekyll/tags/post_url.rb
```diff
@@ -61,44 +61,86 @@ class PostUrl < Liquid::Tag
...
         site.posts.docs.each do |document|
-          next unless @post.deprecated_equality document
+          next unless post.deprecated_equality document
 
-          Jekyll::Deprecator.deprecation_message(
-            "A call to '{% post_url #{@post.name} %}' did not match a post using the new " \
-            "matching method of checking name (path-date-slug) equality. Please make sure " \
-            "that you change this tag to match the post's name exactly."
-          )
+          Jekyll::Deprecator.deprecation_message(create_deprecation_message(post.name))
```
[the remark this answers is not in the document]
jeffque: Ruby aesthetics never stops to amaze me. I need to learn more from the idiomatic (rather to only be able to appreciate it)

ashmaroli:
@jeffque, the core changes pertaining to your idea look good for the time being. However, there cumulative diff looks a bit too spread out (_Not your fault_).
I have submitted a pull request (`#9829`) to refactor the existing implementation in the `master` branch. It is a widely-scoped pr and may take some time to be reviewed. (_May ultimately be split into multiple pull requests of narrower scopes_).
Once those changes land on `master`, the diff here will be reduced (upon resolving conflicts), improving the readability of the changes as a whole.

Therefore, you may rest assured that this pull request will be merged in due course.

ashmaroli:
Hello @jeffque,
Your branch has been changed to be simpler.
(Of the many changes, I removed the function calls to log additional context regarding the Liquid variable markup currently used, in the event of various errors; The additional message doesn't seem helpful).

Please go through current state of the branch and feel free to leave comments if necessary.

parkr (approved):

jeffque:
@ashmaroli , it is important to, from the error message, recognize where is the`{% post_url %}` that was problematic? 

For instance, this 

```liquid
{% for slug in site.data.brute %}
    <li><a href="{% post_url {{ slug }}-error %}">{{ slug }}</a></li>
{% endfor %}
```

generates this error message:

```txt
      Regenerating: 1 file(s) changed at 2025-06-04 13:34:25
                    _includes/brutal.html
  Liquid Exception: Could not find post "2025-04-28-sponsoring-posts-error" in tag 'post_url'. Make sure the post exists and the correct name is given to the tag. in managerial.md
             Error: Could not find post "2025-04-28-sponsoring-posts-error" in tag 'post_url'. Make sure the post exists and the correct name is given to the tag. 
             Error: Run jekyll build --trace for more information.
```

Which is a little more hard to trace back the original tag.

jeffque:
That been said, I loved how the code evolved! Trying to learn more ruby and upgrade the tools that I use

ashmaroli (approved):

PR #9782, opened 2025-02-27
Correct formatting of defaults scopes in doc

adampayne315:
Copy pasting the example didn't work because the yaml array contains hashes instead of scalar values.

## Summary

Updated doc to fix the yaml defaults config.

### DIFF
--- a/docs/_docs/step-by-step/09-collections.md
+++ b/docs/_docs/step-by-step/09-collections.md
@@ -175,17 +175,20 @@ collections:
     output: true
 
 defaults:
-  - scope:
+  -
+    scope:
       path: ""
       type: "authors"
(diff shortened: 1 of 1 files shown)


ashmaroli:
Hello @adampayne315, as an FYI, the current configuration and previous configuration are equivalent (when parsed) if not absolutely identical (at the outset). So, the current diff seems like an unnecessary change.

adampayne315:
> Hello @adampayne315, as an FYI, the current configuration and previous configuration are equivalent (when parsed) if not absolutely identical (at the outset). So, the current diff seems like an unnecessary change.

Thanks! I tried going through the steps again and it worked.

PR #9784, opened 2025-02-28
Avoid caching resource when called via `include_relative` tag

### LINKED ISSUE #9783: [Bug]: `include_relative` behaviour changes wrt frontmatter depending on included file name
### Operating System

Linux Mint 22.1

### Ruby Version

ruby 3.3.6 (2024-11-05 revision 75015d4c1f) [x86_64-linux]

### Jekyll Version

jekyll 4.4.1

### GitHub Pages Version

_No response_

### Expected Behavior

I believe the expected behaviour of `include_relative` is that the entire contents of the included file will be inserted verbatim; even if the included file starts with a block that looks like YAML frontmatter, the "frontmatter" will not be treated as such and will also be inserted verbatim, e.g.

```
---
---

<p>Inserted text</p>
```

### Current Behavior

The expected behaviour regarding the "frontmatter" does not happen when the following conditions are met:

1. The included file contains liquid tags
2. The included file's filename alphabetically precedes the including file's filename (e.g. `{% include_relative apple.html %}` from inside `index.html`)

When this happens, the "frontmatter" does get parsed, and is not inserted verbatim into the outer file.

### Relevant log output

```shell

```

### Code Sample

Main file, `index.html`:

```
---
---

{% include_relative indew.html %}
{% include_relative indey.html %}
```

First included file, `indew.html`, `"indew" < "index"` so frontmatter not inserted :
```
---
---

{% if true %}hi{% else %}bye{% endif%}
```

Second included file, `indey.html`, `"indey" > "index"` so frontmatter is inserted:
```
---
---

{% if true %}foo{% else %}bar{% endif%}
```

Result:
```
hi --- --- foo
```

Expected result:
```
--- --- hi --- --- foo
```

ashmaroli:
- This is a 🐛 bug fix.
- I've added tests.

## Summary

Initialize a new `LiquidRenderer` instance on every render of `include_relative` tag to bypass caching rendered output of page / document within the singular instance of Jekyll 4's `LiquidRenderer`, especially noticeable when the *including* page / document not only *includes* a page / document that precedes the *including* page itself but also if it *includes* a page / document that follows the *including* page, in the rendering queue.

## Context

Resolves #9783

### DIFF
--- a/features/include_relative_tag.feature
+++ b/features/include_relative_tag.feature
@@ -58,3 +58,22 @@ Feature: include_relative Tag
     Then I should get a zero exit status
     And the _site directory should exist
     And I should see "Welcome back Dear Reader!" in "_site/index.html"
+
+  Scenario: Include multiple files relative to a page at root
+    Given I have an "apple.md" page with foo "bar" that contains "{{ page.path }}, {{ page.foo }}"
+    And I have an "banana.md" page with content:
+      """
+        {% include_relative apple.md %}
+        {% include_relative cherry.md %}
+
+        {{ page.path }}
+      """
+    And I have an "cherry.md" page with foo "lipsum" that contains "{{ page.path }}, {{ page.foo }}"
+    When I run jekyll build
+    Then I should get a zero exit status
+    And the _site directory should exist
+    And I should see "<p>apple.md, bar</p>" in "_site/apple.html"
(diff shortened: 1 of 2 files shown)


mattr- (approved):
@jekyllbot: merge +fix

PR #9813, opened 2025-04-24
Fix logs containing IPv6 URLs

printfn:
This is a 🐛 bug fix.

## Summary

I noticed that if I set the `host` url to an IPv6 address, some of the log output doesn't contain valid URLs.

For example:
```
$ jekyll serve
Configuration file: path/_config.yml
            Source: path
       Destination: path/_site
      Generating...
       Jekyll Feed: Generating feed for posts
                    done in 0.263 seconds.
 Auto-regeneration: enabled for 'path'
LiveReload address: http://:::35729
    Server address: http://:::4000/
```

Note the incorrect live-reload and server addresses. The actual binding seems to work correctly already for IPv6, so I'm only fixing the log output. Double-clicking the correctly formatted URL `http://[::]:4000/` opens up the browser as expected.

```
$ ss -plunt
Netid State  Recv-Q  Send-Q                 Local Address:Port    Peer Address:Port Process
...
tcp   LISTEN 0       4096                            [::]:4000            [::]:*     users:(("ruby3.2",pid=17017,fd=9))
```

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -243,7 +243,7 @@ module Jekyll
         def format_url(ssl_enabled, address, port, baseurl = nil)
           format("%<prefix>s://%<address>s:%<port>i%<baseurl>s",
                  :prefix  => ssl_enabled ? "https" : "http",
-                 :address => address,
+                 :address => address.include?(":") ? "[#{address}]" : address,
                  :port    => port,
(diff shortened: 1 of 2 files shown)


ashmaroli (approved):

ashmaroli:
Thank you @printfn
@jekyllbot: merge +fix

PR #9829, opened 2025-05-26
Improve readability of `post_url` tag

ashmaroli:

## Summary

- Reduce tech-debt carried over from pre-Jekyll-3.0 days.
- Add new (private constant) `Tags::PostComparer::POST_PATH_MATCHER` to be similar to `Document::DATE_FILENAME_MATCHER`; (Soft) deprecate `Tags::PostComparer::MATCHER` in case there is somebody using it (via a plugin) in a subclass.
- ~~Add new `Tags::PostComparer#acceptable?` similar to `Tags::PostComparer#==` but returns a Boolean instead of a `MatchData` via latter.~~
- Extract `raise` calls into dedicated private methods to reduce method-length of `Tags::PostUrl#render`.
- Extract logging deprecation-warning against use of legacy matching method into a dedicated private method ~~to incorporate more context in the log-message dynamically~~.

### DIFF
--- a/features/post_url_tag.feature
+++ b/features/post_url_tag.feature
@@ -129,7 +129,7 @@ Feature: PostUrl Tag
     And I have an "index.md" page that contains "[Movies]({% post_url 2019-02-04-hello-world %})"
     When I run jekyll build
     Then I should get a zero exit status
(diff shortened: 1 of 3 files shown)


ashmaroli:
@jekyllbot: merge +dev

PR #9839, opened 2025-06-10
Add logger as runtime dependency

### LINKED ISSUE #9763: [Bug]: gem `logger` was loaded from the standard library
### Operating System

WSL2 on Windows 11

### Ruby Version

ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-linux]

### Jekyll Version

jekyll 4.4.1

### GitHub Pages Version

_No response_

### Expected Behavior

When I try to get the version of jekyll I expect to see:

```sh
$ bundle exec jekyll -v
jekyll 4.4.1
```

### Current Behavior

But there is a warning that the gem `logger` was loaded from the standard library:

```sh
$ bundle exec jekyll -v
/jekyll/_vendor/ruby/3.4.0/gems/jekyll-4.4.1/lib/jekyll.rb:26: warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add logger to your Gemfile or gemspec to silence this warning.
jekyll 4.4.1
```

I have temporary removed all the gems from my `Gemfile`, but the warning remains. So it doesn't seem to be caused by another gem.
As mentioned in the warning I've added `logger` to the `Gemfile` to silence the warning.

### Relevant log output

```shell

```

### Code Sample

_No response_

quanghuynguyen1902:
## Pull Request: Add logger dependency for Ruby 3.5.0 compatibility

Resolves #9763 by adding logger ~> 1.4 as runtime dependency to prevent Ruby 3.5.0 compatibility warnings

This is a 🐛 bug fix.

## Summary

Adds `logger ~> 1.4` as a runtime dependency to prevent Ruby 3.5.0 compatibility warnings that appear when using Jekyll with Ruby 3.4+.

**Changes:**
- Added `s.add_runtime_dependency("logger", "~> 1.4")` to `jekyll.gemspec`
- Eliminates the warning: "logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0"

**Testing:**
- Verified with Ruby 3.4.4 that the warning no longer appears
- Jekyll functions normally with the explicit logger dependency

### DIFF
--- a/jekyll.gemspec
+++ b/jekyll.gemspec
@@ -44,6 +44,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency("kramdown",              "~> 2.3", ">= 2.3.1")
   s.add_runtime_dependency("kramdown-parser-gfm",   "~> 1.0")
   s.add_runtime_dependency("liquid",                "~> 4.0")
+  s.add_runtime_dependency("logger",                "~> 1.4")
   s.add_runtime_dependency("mercenary",             "~> 0.3", ">= 0.3.6")
   s.add_runtime_dependency("pathutil",              "~> 0.9")
   s.add_runtime_dependency("rouge",                 ">= 3.0", "< 5.0")


taketo1113:
Since Jekyll itself also directly uses `require "logger"`, it might make sense to add logger as a runtime dependency to the jekyll gem as well.
@ashmaroli Would this be acceptable?

https://github.com/jekyll/jekyll/blob/1f319fb273b6cdf876bc6edd38d7477935cdda8c/lib/jekyll.rb#L26

PR #9867, opened 2025-08-25
Fix servlet do_GET header logic

fw42:
When running a jekyll dev webrick server locally (with livereload) while having extra headers configured through the config, responses for certain files don't get the headers applied (for example non-html responses). This PR fixes that.

I originally found this bug by using the following config
```yaml
webrick:
  headers:
    Access-Control-Allow-Origin: "*"
```
in combination with the `jekyll-feed` plugin. When loading the feed (which has content-type xml), I noticed that the headers are missing, but are present for other (html) files.

@ashmaroli @parkr

### DIFF
--- a/lib/jekyll/commands/serve/servlet.rb
+++ b/lib/jekyll/commands/serve/servlet.rb
@@ -160,9 +160,7 @@ module Jekyll
         def do_GET(req, res)
           rtn = super
 
-          if @jekyll_opts["livereload"]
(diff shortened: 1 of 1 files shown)


parkr (approved):
Thanks! I guess we didn't write automated tests for this behavior.

fw42:
@parkr: anyone else I should be pinging for review before this can get merged?

mattr-:
Will try to cut 4.5 soon to get this released but am also super busy at the moment, so please be patient with me. 😄

PR #9885, opened 2025-10-22
Add memoization for Link tag

### LINKED ISSUE #9827: feat: Optimization oportunity for tag {% link %}
## Summary

The `render` method inside the [`Jekyll::Tags::Link`](https://github.com/jekyll/jekyll/blob/84437a5052cd9d5ea09f49cb26dc3e5f6ad5a84e/lib/jekyll/tags/link.rb#L20-L39) will call Liquid rendering many times, even when it had rendered the object before.

## Motivation

While working on #9776 , @ashmaroli noticed a possible performance regression (https://github.com/jekyll/jekyll/pull/9776#issuecomment-2898428293) in my PR.

The current code for rendering a `{% link %}` element suffers from the same thing.

The idea is to avoid rendering an already rendered object, and to avoid calling rendering Liquid unnecessarily.

The current `render` for `{% link %}` is:

```ruby
      def render(context)
        @context = context
        relative_path = Liquid::Template.parse(@relative_path).render(context)
        relative_path_with_leading_slash = PathManager.join("", relative_path)


        site.each_site_file do |item|
          return relative_url(item) if item.relative_path == relative_path
          # This takes care of the case for static files that have a leading /
          return relative_url(item) if item.relative_path == relative_path_with_leading_slash
        end


        raise ArgumentError, <<~MSG
          Could not find document '#{relative_path}' in tag '#{self.class.tag_name}'.

          Make sure the document exists and the path is correct.
        MSG
      end
```

If it is detected that this is first rendering, then do this. Else, just reuse previous value. For instance, something like this:

```ruby
      def render(context)
        @context = context
        return @rendered_value unless @rendered_value.nil?

        @rendered_value = first_render
        return @rendered_value
      end

      private

      def first_render
        relative_path = expand_path
        relative_path_with_leading_slash = PathManager.join("", relative_path)


        site.each_site_file do |item|
          return relative_url(item) if item.relative_path == relative_path
          # This takes care of the case for static files that have a leading /
          return relative_url(item) if item.relative_path == relative_path_with_leading_slash
        end


        raise ArgumentError, <<~MSG
          Could not find document '#{relative_path}' in tag '#{self.class.tag_name}'.

          Make sure the document exists and the path is correct.
        MSG
      end

      def expand_path
        content = @orig_post
        return content if content.nil? || content.empty?
        return content unless content.include?("{%") || content.include?("{{")

        Liquid::Template.parse(content).render(@context)
      end

      def solve_path
        content = @relative_path
        return content if content.nil? || content.empty?
        return content unless content.include?("{%") || content.include?("{{")

        Liquid::Template.parse(content).render(@context)
      end
```

It should not cause any changes to the outcome, this is purely a performance optimization opportunity that needs no change in API.

For example (based on https://github.com/jekyll/jekyll/pull/9776#issuecomment-2900796953), consider the following code in an include_file:

```liquid
{% for entry in site.data.archives %}
  <a href="{% link {{ entry.linkpath }} %}" title="{{ entry.title }}">
    {{ entry.title }}
  </a>
{% endfor %}
```

Lets assume this include_file is used in two different layouts to render a sidebar.

During the _build process_, Liquid will _parse_ the above code just _once_ regardless of how many _pages_ utilize the two layouts linking the include_file. This means that, in this particular scenario, there will only be a single instance of `Jekyll::Tags::Link` class but, **this one instance** will call upon **it's `render()` method** as many times as _the number of pages dependent_ on the above include_file via layouts.

> Note that I, the reporter of this issue, am not an expert in Jekyll and it's rendering process/pipeline. There may be something off with the optimization. Trying to give different contexts for the same tag object may break stuff?

## Drawbacks

Adding a "rendering" instance variable means that each node will have a new field, which will increase a more "long lived" memory.

The wrong optimization will lead to the wrong information being presented, this should be taken extremely carefuly.

mondeja:

## Context

Resolves #9827

### Performance

| Case       | Without Cache | With Cache       |
|------------|---------------|----------------|
| **Best**   | O(n · N)        | O(n)           |
| **Worst**  | O(n · N)        | O(k · N + n)     |

**Where:**

- **n** = number of `{% link %}` renders performed in the page or process.  
- **N** = number of files in the site that the `{% link %}` tag must search through to find a match.  
- **k** = number of distinct paths rendered (k ≤ n).  

**Explanation:**

- **Without Cache:** each render searches linearly through N files → n renders × N files.  
- **With Cache:** each distinct path is searched only once among N files; subsequent accesses are O(1) thanks to the hash cache.

### Memory

| Case       | Without Cache       | With Cache              |
|------------|------------------|------------------------|
| **Memory** | O(n) allocations  | O(k) allocations       |

**Where:**

- **n** = number of `{% link %}` renders performed.  
- **k** = number of distinct file paths rendered (k ≤ n).  

**Explanation:**

- **Without Cache:** Each render allocates new objects for parsing and rendering the Liquid template, repeated n times → memory usage grows roughly linearly with n.  
- **With Cache:** Parsing and rendering still happen per distinct path, but repeated renders of the same path reuse cached results, so memory allocations scale with k instead of n, which can drastically reduce memory consumption if n ≫ k.

---

Summary: in the average case (randomization of 1000 paths with 1/1000 changes to be the same), the memoization is ~47% faster. In the worst case, memoization is 2% slower.

```
$ bundle exec ruby benchmark/link-tag-memoized.rb
Benchmark (average case):
ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-linux-gnu]
       Without cache     1.000 i/100ms
          With cache     1.000 i/100ms
       Without cache      6.646 (± 0.0%) i/s  (150.46 ms/i) -     34.000 in   5.115738s
          With cache      9.791 (± 0.0%) i/s  (102.13 ms/i) -     49.000 in   5.008767s

Comparison:
          With cache:        9.8 i/s
       Without cache:        6.6 i/s - 1.47x  slower

Memory profile for Without cache (average case):
Total allocated: 9560.5078125 KB (128000 objects)
Total retained:  1.078125 KB (1 objects)
Memory profile for With cache (average case):
Total allocated: 8608.3984375 KB (127000 objects)
Total retained:  0.0 KB (0 objects)

Benchmark (worst case):
ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-linux-gnu]
       Without cache     1.000 i/100ms
          With cache     1.000 i/100ms
       Without cache      6.407 (± 0.0%) i/s  (156.07 ms/i) -     33.000 in   5.151846s
          With cache      6.311 (± 0.0%) i/s  (158.45 ms/i) -     32.000 in   5.072271s

Comparison:
       Without cache:        6.4 i/s
          With cache:        6.3 i/s - 1.02x  slower

Memory profile for Without cache (worst case):
Total allocated: 9570.1474609375 KB (128128 objects)
Total retained:  1.078125 KB (1 objects)
Memory profile for With cache (worst case):
Total allocated: 8617.0068359375 KB (127127 objects)
Total retained:  0.0 KB (0 objects)
```

### DIFF
--- a/test/test_tag_link.rb
+++ b/test/test_tag_link.rb
@@ -203,4 +203,45 @@ class TestTagLink < TagUnitTest
       end
     end
   end
+
+  context "memoized links" do
+    setup do
+      @template = Liquid::Template.parse(<<~CONTENT
+        ---
+        title: test memoized link
+        ---
+        {% link {{ file_path }} %}
+      CONTENT
+      )
+
+      # Creamos un solo site que se reutilizará
+      @site = fixture_site(
+        "collections" => {
+          "methods" => { "output" => true },
+        }
+      )
+      CollectionReader.new(@site).read
+    end
+
+    should "use cache for same context" do
+      context = { "file_path" => "_methods/yaml_with_dots.md" }
+      output1 = @template.render!(context, registers: { site: @site })
+      output2 = @template.render!(context, registers: { site: @site })
+
+      assert_match %r!/methods/yaml_with_dots\.html!, output1
+      assert_equal output1, output2
+    end
+
+    should "not share cache between different contexts" do
+      context1 = { "file_path" => "_methods/configuration.md", "whatever" => 123 }
+      context2 = { "file_path" => "_methods/yaml_with_dots.md" }
+
+      output1 = @template.render!(context1, registers: { site: @site })
+      output2 = @template.render!(context2, registers: { site: @site })
+
+      assert_match %r!/methods/configuration\.html!, output1
+      assert_match %r!/methods/yaml_with_dots\.html!, output2
+      refute_equal output1, output2
+    end
+  end
 end
--- a/test/test_tag_link.rb
+++ b/test/test_tag_link.rb
@@ -203,45 +203,4 @@ class TestTagLink < TagUnitTest
       end
     end
   end
-
-  context "memoized links" do
-    setup do
-      @template = Liquid::Template.parse(<<~CONTENT
-        ---
-        title: test memoized link
-        ---
-        {% link {{ file_path }} %}
-      CONTENT
-      )
-
-      # Creamos un solo site que se reutilizará
-      @site = fixture_site(
-        "collections" => {
-          "methods" => { "output" => true },
-        }
-      )
-      CollectionReader.new(@site).read
-    end
-
-    should "use cache for same context" do
-      context = { "file_path" => "_methods/yaml_with_dots.md" }
-      output1 = @template.render!(context, registers: { site: @site })
-      output2 = @template.render!(context, registers: { site: @site })
-
-      assert_match %r!/methods/yaml_with_dots\.html!, output1
-      assert_equal output1, output2
-    end
-
-    should "not share cache between different contexts" do
-      context1 = { "file_path" => "_methods/configuration.md", "whatever" => 123 }
-      context2 = { "file_path" => "_methods/yaml_with_dots.md" }
-
-      output1 = @template.render!(context1, registers: { site: @site })
-      output2 = @template.render!(context2, registers: { site: @site })
-
-      assert_match %r!/methods/configuration\.html!, output1
-      assert_match %r!/methods/yaml_with_dots\.html!, output2
-      refute_equal output1, output2
-    end
-  end
 end
--- a/benchmark/link-tag-memoized.rb
+++ b/benchmark/link-tag-memoized.rb
@@ -1,0 +1,117 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require "benchmark/ips"
+require_relative "../lib/jekyll"
+require "memory_profiler"
+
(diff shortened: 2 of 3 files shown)


test/test_tag_link.rb
```diff
@@ -203,4 +203,45 @@ def render_content_with_collection(content, collection_label)
...
+      context1 = { "file_path" => "_methods/configuration.md", "whatever" => 123 }
+      context2 = { "file_path" => "_methods/yaml_with_dots.md" }
+
+      output1 = @template.render!(context1, registers: { site: @site })
+      output2 = @template.render!(context2, registers: { site: @site })
+
+      assert_match %r!/methods/configuration\.html!, output1
+      assert_match %r!/methods/yaml_with_dots\.html!, output2
+      refute_equal output1, output2
+    end
```
mattr-: The cache is an implementation detail, and these tests aren't really testing that they're using the cache at all. I think it's best to simply remove the tests, as they're not testing anything. What do you think?

PR #7707, opened 2019-06-09
Refactor `EntryFilter#glob_include?`

ashmaroli:
## Summary
- We really don't need `Pathutil` or `Pathname` to be used in `EntryFilter#glob_include?`.
- The array of patterns should ideally be just strings and regexps.
- Handle the most common use-case of *string* patterns and then handle regex patterns and non-string patterns.
- Use `Regexp#match?` from Ruby 2.4 with regex patterns

PR #9886, opened 2025-10-23
Remove `pathutil` dependency

mondeja:

## Summary

Remove the unmaintained and unused `pathutil` dependency.

## Context

The `pathutil` dependency was added in #4859, but the refactors #9015 and #7707 removed all references to it. #9015 forget to remove it as a dependency.

Added a benchmark to see the memory and time needed to load _lib/jekyll.rb_. These benchmarks are showing in my local environment that:

- Before the change, library cold boot memory used was 3074.051 KB
- After the change, is using 2996.818 KB

**Benchmark script:**

```ruby
require "benchmark"
require "objspace"

def measure_time(label)
  t = Benchmark.realtime { yield }
  puts "Time: #{t.round(4)}s"
end

def measure_memory(label)
  GC.start
  mem_before = ObjectSpace.memsize_of_all
  yield
  GC.start
  mem_after = ObjectSpace.memsize_of_all
  mem_used = mem_after - mem_before
  puts "Memory: #{mem_used / 1024.0} KB"
end

def measure(label)
  puts label
  measure_time(label) do
    measure_memory(label) do
      yield
    end
  end
end

measure("lib/jekyll.rb load") do
  load_relative = File.expand_path("../lib/jekyll.rb", __dir__)
  load load_relative
end
```


benchmark/jekyll-lib-cold-boot.rb
mattr-: I think benchmarks like this are good for proving that your change has the intended effect, but I don't see a reason to keep them in the repo. What do you think about removing this file from the repo, but including the benchmark script in the PR (either in the comment or by updating the description)?

mattr- (commented):
Thanks for the cleanup! I left a question.

parkr (approved):

parkr:
CI didn't run!

mattr-:
Closing and then reopening to attempt to get Actions to run on this PR

PR #9889, opened 2025-10-24
Fix flaky test that depends on current minute

mondeja:
This is a test fix.

## Summary

Fix flaky test that depends on current minute number. 

## Context

Sometimes, the Cucumber tests suite fails on GitHub Actions at the scenario `Use site.time variable`. For example, see this run: https://github.com/jekyll/jekyll/actions/runs/17277324576/job/49037269833#step:5:473

I've tracked down the problem and can reproduce it locally executing:

```bash
while bash script/cucumber features/site_data.feature:34; do :; done
```

In one or two minutes at most, the test eventually fails.

It's flaky because it depends on the concordance of the current minute before and after the build. The build takes, in my local environment, from ~150 ms to 1s, so the medium probablity of failure is ~1%.

The fix proposes the usage of a time range with 3 seconds of tolerance. The unused helper functions have been removed.

### DIFF
--- a/features/step_definitions.rb
+++ b/features/step_definitions.rb
@@ -362,8 +362,8 @@ end
 
 #
 
-Then(%r!^I should see today's time in "(.*)" with (\d+) seconds? tolerance$!) do |file, seconds|
-  seconds = seconds.to_i
+Then(%r!^I should see today's time in "(.*)"$!) do |file|
+  seconds = 3
   build_time = Time.now
   content = file_contents(file)
   date_time_pattern = /(\d{4}-\d{2}-\d{2}\s\d+:\d{2}:\d{2})/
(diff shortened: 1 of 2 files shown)


features/step_definitions.rb
```diff
@@ -362,8 +362,22 @@
 
 #
 
-Then(%r!^I should see today's time in "(.*)"$!) do |file|
-  step %(I should see "#{seconds_agnostic_time(Time.now)}" in "#{file}")
+Then(%r!^I should see today's time in "(.*)" with (\d+) seconds? tolerance$!) do |file, seconds|
+  seconds = seconds.to_i
```
mattr-: I think five seconds is a perfectly valid amount of tolerance that doesn't need configuration and we can simply hard code that here. Thoughts? We could even lower the allowed time to something like 2 or 3 seconds.

```suggestion
  seconds = 5
```

mattr- (approved):

PR #9918, opened 2025-12-30
feat(lsi): switch from classifier-reborn to classifier gem

cardmagic:
## Summary
Switch from `classifier-reborn` to the original `classifier` gem (v2.2+) for LSI-based related posts.

## Why?

**Fewer dependencies:** The classifier gem includes a small native C extension for SVD operations, eliminating the need for GSL (GNU Scientific Library) as an external dependency while providing 5-50x faster LSI performance.

**Better for growing sites:** Enables Brand's algorithm for incremental SVD updates. When new posts are added, the LSI index can be updated without rebuilding from scratch—useful for sites with many posts.

**Active maintenance:** The original classifier gem is under active development again, while classifier-reborn hasn't been updated in years.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -84,7 +84,7 @@ group :jekyll_optional_dependencies do
   gem "tomlrb"
 
   platforms :ruby, :mswin, :mingw, :x64_mingw do
-    gem "classifier-reborn", "~> 2.2"
+    gem "classifier", "~> 2.2"
     gem "liquid-c", "~> 4.0"
     gem "yajl-ruby", "~> 1.4"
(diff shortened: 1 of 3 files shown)


agarie:
FWIW I was adapting a few Jekyll sites I maintain to use the `classifier` gem this morning and found out about this PR.

+1 to this 😄

mattr-:
we switched to classifier-reborn because the original classifier gem went unmaintained and the `-reborn` version was maintained. I'll accept a PR to switch things over but this is very low priority for me, personally.

PR #9920, opened 2026-01-02
Add logger to Gemfile for Ruby 4.0

jbourassa:
This is a 🐛 bug fix. (ish)

## Summary

The issue: on Ruby 4.0, `jekyll serve` fails unless I add `logger` to the local Gemfile.

The fix: add `logger` to the gemspec.

## Context
Logger is now a bundled gem, which means it must be added to the Gemfile/gemspec. See [this part of Ruby 4.0 changelog](https://www.ruby-lang.org/en/news/2025/12/25/ruby-4-0-0-released/#:~:text=logger%201%2E7%2E0).

I'm picking the earliest version of Logger available, because we used to rely on the built-in logger which could've been old.


legionerlegion:
сб, 24 янв. 2026 г., 16:41 Taketo Takashima ***@***.***>:

PR #9942, opened 2026-02-24
Add scheduled rebuild for docs site to keep copyright year current

### LINKED ISSUE #9941: Footer copyright year on jekyllrb.com still shows 2025
The footer copyright on jekyllrb.com shows 2025. The template is correct (`site.time | date: '%Y'`), just needs a rebuild to pick up the new year.

hammadxcm:
- I've adjusted the workflow to handle empty commits gracefully

## Summary

The footer on jekyllrb.com displays the copyright year using `{{ site.time | date: '%Y' }}`, which is evaluated at **build time**. The docs workflow (`.github/workflows/docs.yml`) only triggers on pushes to `master` that modify files under `docs/**`. If no docs changes are merged around the turn of the year, the site is never rebuilt and the copyright year goes stale (currently still showing 2025).

This PR makes three targeted changes to the docs workflow:

1. **Add a `schedule` trigger** — A weekly cron (`0 0 * * 1`, every Monday at midnight UTC) ensures the site is rebuilt periodically, keeping `site.time` and any other dynamic content current.

2. **Add a `workflow_dispatch` trigger** — Allows maintainers to manually trigger a rebuild from the Actions tab when needed.

3. **Update the job `if` condition** — Changed from `!contains(github.event.commits[0].message, '[ci skip]')` to `github.event_name != 'push' || !contains(github.event.commits[0].message, '[ci skip]')` so the `[ci skip]` check only applies to `push` events. Schedule and dispatch events always run.

4. **Guard the deploy step against empty commits** — Wraps the `git commit`/`git push` in a `git diff --cached --quiet` check. Without this, scheduled builds that produce no changes would fail because `git commit` exits non-zero on an empty commit. This would cause 51 out of 52 weekly runs to fail.

The footer template itself (`docs/_includes/footer.html`) is correct and idiomatic Jekyll — no changes needed there.

## Context

Fixes #9941

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -6,13 +6,17 @@ on:
       - master
     paths:
       - "docs/**"
+  schedule:
+    # Run weekly to keep dynamic content (e.g. copyright year) current
+    - cron: "0 0 * * 1"
+  workflow_dispatch:
 
 env:
   RUBY_VERSION: 2.7
 
 jobs:
   deploy_docs:
-    if: "!contains(github.event.commits[0].message, '[ci skip]')"
+    if: github.event_name != 'push' || !contains(github.event.commits[0].message, '[ci skip]')
     runs-on: 'ubuntu-latest'
     env:
       BUNDLE_PATH: "vendor/bundle"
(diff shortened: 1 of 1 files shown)


mattr-:
This isn't really an issue, as we haven't published any new content in 2026 that would require a date update, and once we do, the site will update itself to the correct year.

PR #9945, opened 2026-02-26
Fixing incorrectly excluded paths

### LINKED ISSUE #9116: [Bug]: `include` option ignores folders
### Operating System

Windows 11

### Ruby Version

ruby 3.0.3p157 (2021-11-24 revision 3fb7d2cadc) [x64-mingw32]

### Jekyll Version

jekyll 4.2.2

### GitHub Pages Version

_No response_

### Expected Behavior

Given this source folder:

```
excluded/
    included/
      this-file-should-be-included
_config.yml
```

with this  `_config.yml`:

```yaml
exclude:
  - excluded

include:
  - excluded/included
```

When running `bundle exec jekyll build`, I expect the following output:

```
_site/
  excluded/
    included/
      this-file-should-be-included
```

### Current Behavior

The current implementation produces an empty `_site` folder.

Note that if the `include` option points to the file explicitly, the file is included as expected

### Relevant log output

_No response_

### Steps to reproduce issue

- Clone [jekyll-config-include-folder-repro](https://github.com/bblanchon/jekyll-config-include-folder-repro)
- Install site dependencies
- Run `bundle exec jekyll build`

seroperson:
This is a 🐛 bug fix.

## Summary

Fixes cases when `exclude` excludes directories which are declared in `include`:

```
exclude:
  - private
include:
  - private/assets/images
```

Now output directory will contain `private/assets/images`.

## Context

Closes #9116

### DIFF
--- a/lib/jekyll/entry_filter.rb
+++ b/lib/jekyll/entry_filter.rb
@@ -52,7 +52,7 @@ module Jekyll
     end
 
     def included?(entry)
-      glob_include?(site.include, entry) ||
+      glob_include?(site.include, relative_to_source(entry)) ||
         glob_include?(site.include, File.basename(entry))
     end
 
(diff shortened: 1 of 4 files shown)


afurm:
The `glob_include?(site.include, entry)` → `glob_include?(site.include, relative_to_source(entry))` change in `included?` looks correct, but there's a subtle interaction: `relative_to_source(entry)` is also called inside `excluded?` — if `entry` is already a relative path, calling `relative_to_source` again on it could produce an incorrect path (e.g., `foo/../bar` vs `bar`). Has this been tested with entries that already have partial source-relative paths?

seroperson:
> The `glob_include?(site.include, entry)` → `glob_include?(site.include, relative_to_source(entry))` change in `included?` looks correct, but there's a subtle interaction: `relative_to_source(entry)` is also called inside `excluded?` — if `entry` is already a relative path, calling `relative_to_source` again on it could produce an incorrect path (e.g., `foo/../bar` vs `bar`). Has this been tested with entries that already have partial source-relative paths?

I've added a specific test for that

PR #9947, opened 2026-02-27
keep_files option now supports globs

### LINKED ISSUE #9804: feat: Add support for globs in the keep_files option
## Summary

It would be great if users could use globs in the "keep_files" configuration option.

## Motivation

As of now the user needs to know and explicitly list every single file to be preserved. This is verbose and inflexible, e.g. they can't keep all files with the given extension or a prefix.

## Guide-level explanation

Copying this snippet from the "include" configuration option docs and adding it to the "keep_files" docs  will suffice:

> This configuration option supports Ruby's File.fnmatch filename globbing patterns to match multiple entries to include, refer the exclude configuration option for more information. 

Copy the implementation of either "include" or "exclude".

## Drawbacks

It may break backward compatibility in edge cases where an existing configuration lists files that include glob special characters.

## Unresolved Questions

What is life?

seroperson:

## Summary

`keep_files` option now supports globs to get in line with `include` and `exclude`.

## Context

Closes #9804


seroperson:
@mattr- @ashmaroli Hello! Sorry for ping, guys, but would you mind reviewing this? I would not like this to become a stale PR for months ..

mattr-:
The CI builds need investigation/fixing. The spell check build failures can be ignored.

seroperson:
> The CI builds need investigation/fixing. The spell check build failures can be ignored.

Seems like this error also exist in `master`. Yea, I'll fix it here then, and rebase my other PR after this got merged.

seroperson:
Yea, well, seems like minitest 6.x handles plugins in some new way, so they should be registered manually. I reverted back the `--profile` option and manually register this plugin if such API is available.

They're still failing, probably something else got broken since last green build. I'll check this in a few hours.

seroperson:
@mattr- should be good now.

- `--profile` now works for both ruby `2.x` and `3.x`.
- `export TZ=UTC` instead `TZ=UTC` because old variant doesn't propagate into child processes.
- Use `assert_nil` if comparing with `nil`, for some reason using `assert_equal` with `nil` doesn't work in ruby `3.x`.
- Manually excluded `test_coffeescript.rb` test, as `jekyll-coffeescript` produces a little bit different output under ruby `3.x`. This plugin is 5+ years old and probably should be updated. At the end feels like this test fits better in the `jekyll-coffeescript`'s repo.
- Added spelling exclusion.

seroperson:
Well, it was challenging to debug, but flaky `test_site.rb:81` didn't see a cache dir because `@disk_cache_enabled` in `cache.rb` sometimes stuck in `false` state. Whether it fail or not depends on test execution order.

seroperson:
@mattr- Sorry for another ping, but would you mind reviewing this one more time?

seroperson:
@mattr- Friendly reminder =)

mattr-:
@seroperson This is in my review queue and when I'm ready, I'll give it another look. Working to knock some other stuff out first.

PR #9951, opened 2026-03-16
fix: [Docs]: Liquid 4 docs hard to find / Liquid version is unclear

### LINKED ISSUE #9948: [Docs]: Liquid 4 docs hard to find / Liquid version is unclear
## Motivation

The [official Liquid documentation](https://shopify.github.io/liquid/)  (as linked in Jekyll's doc) refers to its version 5.0. However Jekyll uses Liquid 4.0. It makes it hard to understand what is valid or not (e.g Liquid 5 says "The include tag is deprecated, use render").

Also, Jekyll doc doesn't explicitly say that it still uses Liquid 4.0, I had to look up the gemspec from the source.

## Suggestion

Different suggestions : 

- Update Jekyll to use Liquid 5, and updates the Jekyll docs accordingly
- Fork Liquid 4 docs and embed them into Jekyll doc
- At least, put a warning in Jekyll doc regarding the version mismatch issue

JiwaniZakir:
Fixes #9948

Users visiting `docs/_docs/liquid/tags.md` were confused about which version of Liquid Jekyll supports, since the official Liquid documentation describes features (such as the `render` tag) that are unavailable in Jekyll's bundled Liquid 4. The root cause is that no version warning existed on the tags documentation page, leaving users to discover incompatibilities through trial and error. Adds a warning callout block at the top of `docs/_docs/liquid/tags.md` that explicitly states Jekyll uses Liquid 4 and warns that some features in the official Liquid docs may not be available. Verified by reviewing the rendered warning note matches the existing note/warning pattern used throughout the Jekyll docs.

### DIFF
--- a/docs/_docs/liquid/tags.md
+++ b/docs/_docs/liquid/tags.md
@@ -2,6 +2,11 @@
 title: Tags Filters
 permalink: "/docs/liquid/tags/"
 ---
+<div class="note warning">
+  <h5>Jekyll uses Liquid 4</h5>
+  <p>The <a href="https://shopify.github.io/liquid/">official Liquid documentation</a> refers to a newer version of Liquid than the one Jekyll uses. Some features described there (e.g. the <code>render</code> tag) may not be available in Jekyll.</p>
+</div>
+
 All of the standard Liquid
 [tags](https://shopify.github.io/liquid/tags/control-flow/) are supported.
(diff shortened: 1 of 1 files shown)


JiwaniZakir:
Closing this PR as it hasn't received maintainer review. Happy to resubmit if there's interest.

PR #9954, opened 2026-03-30
build: fix broken CI on newer rubies

mattr-:
This is a 🐛 bug fix.

## Summary

This is a minimal set of changes in an attempt to get CI back to green.

## Context

CI has been failing on newer PRs for awhile now. Rather than making contributors fix it in their PRs, we'll fix it centrally.


mattr-:
@jekyllbot: merge +dev

PR #9967, opened 2026-04-17
docs: fix double-space typos in three pages

55728:
## Summary

Remove stray double spaces in three documentation pages. Each fix is a one-character (one-space) diff — no semantic change, no file reorg.

- `docs/_docs/configuration/markdown.md`: `you  need` → `you need`
- `docs/_docs/static_files.md`: `where  you` → `where you`
- `docs/_docs/upgrading/2-to-3.md`: `install our  [jekyll-textile-converter]` → `install our [jekyll-textile-converter]`

## Context

Not related to an open issue. Found while reviewing the docs; confirmed these are not Markdown hard line breaks (all occur mid-line, not at end-of-line) and not inside code blocks.

### DIFF
--- a/docs/_docs/configuration/markdown.md
+++ b/docs/_docs/configuration/markdown.md
@@ -29,7 +29,7 @@ Documentation for Kramdown parsers is available in the [Kramdown docs](https://k
 
 ### Syntax Highlighting (CodeRay)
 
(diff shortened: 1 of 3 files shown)


mattr-:
A visual comparison between the live site and the deploy preview show no difference in the way the pages are rendered though.

55728:
Thanks for reviewing! Good to know that source-only whitespace fixes aren't considered actionable here. Appreciate the quick feedback — looking forward to contributing in other ways.

PR #9969, opened 2026-04-19
fix: Run release-please workflow only in jekyll/jekyll repo

iBug:
This is a 🐛 bug fix.

## Summary

The `release-please.yml` workflow doesn't make sense in forked repositories. Let's restrict its execution to jekyll/jekyll only.

I did not expect a "workflow failed" notification soon after catching up with the upstream in my forked repo:

image

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -11,6 +11,7 @@ permissions:
 
 jobs:
   release-please:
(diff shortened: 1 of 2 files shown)


mattr- (approved):
Didn't think about that since I hardly deal with forks. Thanks for fixing!

PR #9978, opened 2026-05-09
fix(core): avoid mutating document data for fallback dates

### LINKED ISSUE #5931: Issues with lazy initializations using the  ||= operator when iterating document collections
While developing a pagination plugin I came a cross a strange error when users tried to build jekyll using the `--drafts` argument

```
jekyll build --drafts
```

I traced the problem to the use of the `||=` operator in the document.rb class. One example is the [date function](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/document.rb#L60-L62). Using lazy assignment for member variables actually causes issues when iterating over documents that do not have these fields set as the collection iterator is broken as soon as the values are assigned.

This surfaces when using the `--drafts` switch because draft documents may or may not have the `date:` front matter set. 

I encountered this problem when I was simply doing an Array.sort on a list of posts and the problem occurred even before any of my code was called. Namely in the `<=>` compare function in the Document class (this function accesses lazy initialized values).

```
lib/jekyll/document.rb:61:in `date': can't add a new key into hash during iteration (RuntimeError)
```

The use of this `||=` operator should perhaps be revised.

apples-kksk:
## Summary
- avoid writing computed fallback dates into `Document#data` when `Document#date` is read
- cache the fallback date on the document instance instead
- add a regression test for resolving a fallback date while iterating document data

Fixes #5931

## Verification
- `GEM_PATH=/tmp/jekyll-bundle/ruby/3.4.0:/home/molang/.local/share/gem/ruby/3.4.0:/usr/local/lib/ruby/gems/3.4.0:/usr/lib/ruby/gems/3.4.0 TZ=UTC ruby -Ilib:test test/test_document.rb -n '/fallback date/'`
- `GEM_PATH=/tmp/jekyll-bundle/ruby/3.4.0:/home/molang/.local/share/gem/ruby/3.4.0:/usr/local/lib/ruby/gems/3.4.0:/usr/lib/ruby/gems/3.4.0 TZ=UTC ruby -Ilib:test test/test_document.rb`
- `GEM_PATH=/tmp/jekyll-bundle/ruby/3.4.0:/home/molang/.local/share/gem/ruby/3.4.0:/usr/local/lib/ruby/gems/3.4.0:/usr/lib/ruby/gems/3.4.0 /tmp/jekyll-bundle/ruby/3.4.0/bin/rubocop lib/jekyll/document.rb test/test_document.rb`
- `git diff --check HEAD`

### DIFF
--- a/lib/jekyll/document.rb
+++ b/lib/jekyll/document.rb
@@ -84,7 +84,7 @@ module Jekyll
     #
     # Return document date string.
     def date
-      data["date"] ||= (draft? ? source_file_mtime : site.time)
+      data["date"] || fallback_date
     end
 
     # Return document file modification time in the form of a Time object.
@@ -538,6 +538,10 @@ module Jekyll
 
     def generate_excerpt
       data["excerpt"] ||= Jekyll::Excerpt.new(self) if generate_excerpt?
+    end
+
+    def fallback_date
+      @fallback_date ||= (draft? ? source_file_mtime : site.time)
     end
   end
 end
--- a/test/test_document.rb
+++ b/test/test_document.rb
@@ -77,6 +77,16 @@ class TestDocument < JekyllUnitTest
       assert_equal Time.now.strftime("%Y/%m/%d"), @document.to_liquid["date"].strftime("%Y/%m/%d")
     end
 
+    should "not mutate its data when resolving a fallback date" do
+      refute @document.data.key?("date")
+
+      @document.data.each do
+        @document.to_liquid["date"]
+      end
+
+      refute @document.data.key?("date")
+    end
+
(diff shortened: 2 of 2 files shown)


mattr-:
Doesn't follow the PR template and is obviously AI generated in an unsupervised manner.

apples-kksk:
Thanks for the direct feedback.

You’re right that I should have followed the PR template before submitting this, and the PR body did not provide enough concrete context for review. I should have caught that earlier.

I’ll leave this closed. I’ve updated my process so future PRs start from the repository template, include concrete investigation and validation notes, and are checked against the issue requirements before submission.

PR #9981, opened 2026-05-12
fix(core): avoid mutating document data for fallback dates

### LINKED ISSUE #5931: Issues with lazy initializations using the  ||= operator when iterating document collections

apples-kksk:
This is a 🐛 bug fix.

## Summary

Avoid mutating `Document#data` when `Document#date` resolves its fallback value.

`Document#date` previously used `data["date"] ||= ...`, so reading a dateless document's date inserted a new `"date"` key into the document data hash. That can raise `RuntimeError: can't add a new key into hash during iteration` when callers read date-related Liquid data while iterating over `document.data`.

This changes the fallback path to return the computed fallback date without writing it into `data`, while preserving explicit front matter dates.

## Context

Fixes #5931.

The regression test exercises the failure mode from the issue: iterate over `document.data`, read `document.to_liquid["date"]`, and verify that resolving the fallback date does not add a `"date"` key to `document.data`.

No documentation changes are included because this fixes internal mutation behavior without changing the documented `date` API.

Validation run locally:

- `TZ=UTC script/test test/test_document.rb -n '/fallback date/'`: passed, 1 test, 2 assertions, 0 failures.
- `TZ=UTC script/test test/test_document.rb`: passed, 65 tests, 109 assertions, 0 failures.
- `bundle exec rubocop lib/jekyll/document.rb test/test_document.rb`: passed, 2 files inspected, no offenses.
- `BUNDLE_WITH=rdoc EXECJS_RUNTIME=Node TZ=UTC script/cibuild`: passed. This covered full RuboCop, 963 unit tests, 304 cucumber scenarios, and the default-site build.

### DIFF
--- a/test/test_document.rb
+++ b/test/test_document.rb
@@ -80,9 +80,7 @@ class TestDocument < JekyllUnitTest
     should "not mutate its data when resolving a fallback date" do
       refute @document.data.key?("date")
 
-      @document.data.each do
-        @document.to_liquid["date"]
-      end
+      @document.to_liquid["date"]
 
       refute @document.data.key?("date")
     end
--- a/test/test_document.rb
+++ b/test/test_document.rb
@@ -77,6 +77,16 @@ class TestDocument < JekyllUnitTest
       assert_equal Time.now.strftime("%Y/%m/%d"), @document.to_liquid["date"].strftime("%Y/%m/%d")
     end
 
+    should "not mutate its data when resolving a fallback date" do
+      refute @document.data.key?("date")
+
+      @document.data.each do
+        @document.to_liquid["date"]
+      end
+
+      refute @document.data.key?("date")
+    end
+
     should "be jsonify-able" do
       page_json = @document.to_liquid.to_json
       parsed = JSON.parse(page_json)
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -143,12 +143,14 @@ gettalong
 ghp
 ghpages
 giraffeacademy
+gitcms
 githubcom
 gitlab
 gjtorikian
 globbed
 gotcha
 Goulven
+GPT
 gridism
 GSo
 gsubbing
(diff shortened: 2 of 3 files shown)


test/test_document.rb
```diff
@@ -77,6 +77,16 @@ def setup_document_with_dates(filename)
       assert_equal Time.now.strftime("%Y/%m/%d"), @document.to_liquid["date"].strftime("%Y/%m/%d")
     end
 
+    should "not mutate its data when resolving a fallback date" do
+      refute @document.data.key?("date")
+
+      @document.data.each do
+        @document.to_liquid["date"]
+      end
```
mattr-: We don't need to call `to_liquid` for every bit of information in `@document.data` do we?
```suggestion
      @document.to_liquid["date"]
```

mattr- (commented):
Just the one question

test/test_document.rb
apples-kksk: Updated, thanks. The test now calls `@document.to_liquid["date"]` directly.

mattr- (approved):

PR #9982, opened 2026-05-13
feat: support multiple layout output formats

### LINKED ISSUE #3041: Proposal: multiple formats per page using layouts

landeqiming666:
## Summary

This adds support for rendering multiple output files when layouts share the same basename with different extensions. For example, a page using `layout: event` can now render through both `_layouts/event.html` and `_layouts/event.ics`, producing `event.html` and `event.ics`.

The implementation keeps the existing primary layout behavior intact by preferring the HTML-like layout as the regular `site.layouts` entry, while tracking sibling layout variants separately. It also updates writing, cleaning, and incremental regeneration to account for the additional destination paths.

Format-specific layout inheritance is supported as well: if `_layouts/event.ics` declares `layout: wrapper`, Jekyll prefers `_layouts/wrapper.ics` for the iCalendar output when that layout exists.

## Context

Closes #3041.

This follows the original same-basename layout request and keeps the scope focused on layout-driven sibling outputs. A single non-HTML layout without sibling variants keeps the existing behavior and does not create an additional output file.

Tests run locally with Ruby 3.3:

- `bundle exec ruby -Itest test/test_layout_reader.rb`
- `bundle exec ruby -Itest test/test_site.rb --name '/rendering multiple layout formats/'`
- `bundle exec ruby -Itest test/test_document.rb --name '/destination|get the right output_ext/'`
- `bundle exec ruby -Itest test/test_page.rb --name '/destination|url/'`
- `bundle exec rubocop lib/jekyll/cleaner.rb lib/jekyll/convertible.rb lib/jekyll/document.rb lib/jekyll/page.rb lib/jekyll/readers/layout_reader.rb lib/jekyll/regenerator.rb lib/jekyll/renderer.rb lib/jekyll/site.rb test/test_layout_reader.rb test/test_site.rb --format simple`
- `git diff --check`

RuboCop reported no offenses for the inspected files; it also printed the existing new-cop configuration warnings.

### DIFF
--- a/docs/_docs/layouts.md
+++ b/docs/_docs/layouts.md
@@ -99,6 +99,23 @@ The rendered output of this page is:
 </html>
 ```
 
+## Multiple output formats
+
+If multiple layouts share the same basename, Jekyll renders the document once
+for each layout extension. For example, a page with `layout: event` will use
+both `_layouts/event.html` and `_layouts/event.ics` when both files exist.
+
+The regular output keeps the extension assigned by the page or document, such
+as `.html` for Markdown. Additional outputs use the matching layout extension
+and the same URL path. If `event.md` writes `/event.html`, the additional
+iCalendar layout writes `/event.ics`.
+
+Format-specific layouts can inherit from format-specific parents. If
+`event.ics` has `layout: wrapper`, Jekyll will prefer `_layouts/wrapper.ics`
+for the iCalendar output when that layout exists. A single non-HTML layout
+without sibling layouts keeps the existing behavior and does not create an
+extra output file.
+
 ## Inheritance
 
 Layout inheritance is useful when you want to add something to an existing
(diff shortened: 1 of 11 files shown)


landeqiming666:
Hi maintainers, a gentle follow-up on this PR.

It remains mergeable, the latest `commit_lint` run passes (the red entry is an earlier superseded run), and the Netlify preview is available. The change and regression coverage are described in the PR body. Please let me know if any scope or test adjustments would help with review. Thanks!

PR #9984, opened 2026-05-13
docs: Fix wrong paths to avoid opening 404 pages when visiting links

note4:


note4:
The path error causes the wrong link to be generated. Click to open the 404 page.

PR #9985, opened 2026-05-13
feat: add explicit and automatic layout output formats

apples-kksk:

## Summary

Adds secondary layout outputs through the `outputs` front matter key while keeping existing pages unchanged by default.

Explicit output example:

```yaml
layout: event
outputs:
  - ics
```

Automatic same-basename layout example:

```yaml
layout: event
outputs: auto
```

With `_layouts/event.html` and `_layouts/event.ics`, `outputs: auto` writes the normal HTML output plus the matching `.ics` output at the same URL path. Adding `_layouts/event.ics` alone does not change existing `layout: event` pages.

The implementation covers pages, posts, collection documents, pretty permalinks, format-specific layout inheritance, cleanup, incremental regeneration, doctor conflict checks, and post-render hooks for additional outputs.

## Context

Refs #3041.

This keeps the scope to layout-based text outputs. It does not add same-extension variants, custom URL suffixes, MIME handling, binary/PDF generation, converter-level multi-output, or `page.<ext>_url` helpers.

## Relation to #3041

The original issue describes generating multiple outputs from sibling layouts with the same basename. This PR supports that behavior explicitly with `outputs: auto`, and also keeps an explicit subset mode with `outputs: [ics]` plus opt-out via `outputs: false` or `outputs: []`.

For larger sites, `outputs: auto` can be enabled through front matter defaults.

Tests run:

- `ruby -Itest test/test_site.rb --name '/multi-format outputs/'`
- `ruby -Itest test/test_regenerator.rb --name '/single destination path|destination missing|missing output/'`
- `ruby -Itest test/test_layout_reader.rb --name '/non-HTML extension|only a non-HTML layout|site format layout/'`
- `ruby -Itest test/test_page.rb --name '/destination|url/'`
- `ruby -Itest test/test_document.rb --name '/destination|write/'`
- `ruby -Itest test/test_cleaner.rb`
- `ruby -Itest test/test_doctor_command.rb`
- `rubocop` on changed Ruby/test files: no offenses
- `git diff --check`

A broader local `test/test_site.rb test/test_regenerator.rb` run reached unrelated theme-gem setup errors for `test-theme` in this local bundle, after the multi-format and regenerator coverage above had passed.

### DIFF
--- a/docs/_docs/layouts.md
+++ b/docs/_docs/layouts.md
@@ -99,6 +99,37 @@ The rendered output of this page is:
 </html>
 ```
 
+## Multiple output formats
+
+Pages and documents can request additional output formats with the `outputs`
+front matter key. Jekyll keeps the normal `layout` output as the primary output
+and looks for matching layouts with the requested extensions.
+
+For example, this page uses `_layouts/event.html` for its HTML output and
+`_layouts/event.ics` for its calendar output:
+
+```markdown
+---
+title: Meetup
+layout: event
+outputs:
+  - ics
+---
+May meetup details
+```
+
+Jekyll writes the usual HTML file and an additional file with the `.ics`
+extension at the same URL path. With a pretty permalink such as `/events/may/`,
+the additional output is written to `/events/may/index.ics`.
+
+Additional outputs are explicit. Adding `_layouts/event.ics` does not change
+pages that only specify `layout: event`; those pages must also set `outputs`.
+You can use
+[front matter defaults](/docs/configuration/front-matter-defaults/) to enable an
+additional format for many pages at once. Requested output names must be simple
+extension names, such as `ics` or `json`. If the matching layout does not exist,
+Jekyll warns and skips that additional output.
+
 ## Inheritance
 
 Layout inheritance is useful when you want to add something to an existing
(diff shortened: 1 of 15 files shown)


apples-kksk:
Updated this PR to support the original same-basename layout behavior through `outputs: auto`, while keeping existing pages unchanged by default.

`outputs: auto` lets pages/documents opt into automatic sibling layout outputs. `outputs: [ics]` still works for explicit subsets, and `outputs: false` / `outputs: []` can opt out.

I also added coverage for the original #3041-style example, front matter defaults, pretty permalinks, format-specific layout inheritance, cleanup, incremental regeneration, and doctor conflict checks.

docs/_docs/layouts.md
```diff
@@ -99,6 +99,44 @@ The rendered output of this page is:
 </html>
 ```
 
+## Multiple output formats
```
mattr-, line 102: Since the docs site is built and goes live when a PR is merged, this needs a version specifier. However, I can't tell you what the next version is going to be yet, so there's a bit of a chicken and egg problem here at the moment.

mattr- (commented):
Still considering this but leaving an early comment.

docs/_docs/layouts.md
apples-kksk: Thanks, that makes sense. I can add the docs version marker once the target release version is known.

PR #9134, opened 2022-09-26
Relax version constraint to allow Rouge 4.x

ashmaroli:
- This is a 🙋 feature or enhancement.

## Summary

[Rouge 4.0 was released](https://github.com/rouge-ruby/rouge/releases/tag/v4.0.0) that dropped support for Rubies older than 2.7.0

## Context

Closes #9123

PR #10008, opened 2026-08-07
chore: relax version constraint to allow Rouge 5.x

### LINKED ISSUE #10003: Dependency on `rouge` is too strict
Rouge 5.0 released in late May and it doesn't seem to have any breaking changes from 4.x that are relevant to Jekyll 4.4: https://github.com/rouge-ruby/rouge/releases

However, Jekyll has a dependency set as `rouge >= 3.0, < 5.0 is required.`, which makes 4.7 the latest version of Rouge that can be used without conflicts. This restriction is probably too strict and could be loosened.

MsfPablo:
Fixes #10003.

Rouge 5.0 shipped in May 2025 with no changes that affect the way Jekyll uses it, but the gemspec constraint `< 5.0` keeps everyone on 4.7. This widens it to `< 6.0` (mirroring #9134, which did the same for 4.x).

### Ruby 2.7

Rouge 5.x declares `required_ruby_version >= 3.0`, so Bundler automatically keeps Ruby 2.7 users on Rouge 4.x. No change to Jekyll's own `required_ruby_version` is needed, and CI on 2.7 is unaffected.

### Test changes

Two `test_tag_highlight` assertions pinned the exact table-formatter markup, which Rouge 5 changed in three cosmetic ways:

- the gutter cell gained `aria-hidden="true"`
- a leading `<span class="w">` may precede the `<table>`
- the code cell no longer ends with a trailing newline

Both assertions are now regexes that tolerate either form, so the tests hold across the whole supported range rather than only the newest Rouge.

### Verification

`test/test_tag_highlight.rb` passes (22 tests, 48 assertions, 0 failures) against Rouge **3.30.0**, **4.7.0** and **5.1.0**. The rest of the unit suite is unchanged by this PR.

### DIFF
--- a/jekyll.gemspec
+++ b/jekyll.gemspec
@@ -47,7 +47,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency("logger",                "~> 1.3")
   s.add_runtime_dependency("mercenary",             "~> 0.3", ">= 0.3.6")
   s.add_runtime_dependency("pathutil",              "~> 0.9")
-  s.add_runtime_dependency("rouge",                 ">= 3.0", "< 5.0")
+  s.add_runtime_dependency("rouge",                 ">= 3.0", "< 6.0")
   s.add_runtime_dependency("safe_yaml",             "~> 1.0")
   s.add_runtime_dependency("terminal-table",        ">= 1.8", "< 4.0")
   s.add_runtime_dependency("webrick",               "~> 1.7")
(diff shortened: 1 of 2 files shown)


huonw:
Thanks for doing this upgrade.

> a leading `<span class="w">` may precede the `<table>`

Unfortunately, I think this is more than just cosmetic differences in the raw HTML, as the extra spacing is visible in the rendered output: https://github.com/rouge-ruby/rouge/issues/2335

Before:

image

After:

image

